PR #1024, opened 2019-09-12
Remove Word16 range limit for network magic

karknu:


ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
```diff
@@ -111,8 +111,7 @@ nodeToClientCodecCBORTerm = CodecCBORTerm {encodeTerm, decodeTerm}
         CBOR.TInt (fromIntegral networkMagic)
 
       decodeTerm :: CBOR.Term -> Either Text NodeToClientVersionData
-      decodeTerm (CBOR.TInt x) | x >= 0 && x <= 0xffff = Right (NodeToClientVersionData $ fromIntegral x)
-                               | otherwise             = Left $ T.pack $ "networkMagic out of bound: " <> show x
+      decodeTerm (CBOR.TInt x) = Right (NodeToClientVersionData $ fromIntegral x)
```
coot: I'd rather prefer if we extended the limit than remove the check.  An honest node will send a `Word32`, but this decoder will accept any `Int64` value, which then is mapped to `Word32`.

karknu: Exactly what it accepts is dependent on what `Int` happens to be. I'd would prefer that CBOR exposed integers of a fixed hw independent width. I don't see any problem with accepting any `Int64` value that can be mapped to a`Word32` which matches a configured network magic.

"Be liberal in what you accept, and conservative in what you send."

coot: > I'd would prefer that CBOR exposed integers of a fixed hw independent width.
It can only run on `64`-bit machines, so we can be sure it will use `64`-bit integers.

> Be liberal in what you accept, and conservative in what you send
I think it's better if we are strict.  This is not a link redirecting messages and letting the final service decide whether it's valid or not.

@MarcFontaine is (or will be) working on making other decoders to be strict to (expected lengths of encoded tuples to match).

karknu: > This is not a link redirecting messages and letting the final service decide whether it's valid or not.
That sounds like exactly like network-mux does to me.

coot: But this is not in network-mux: this is an application that runs on top of it :)

PR #1027, opened 2019-09-13
low / high mark pipeline strategy

coot:
@edsko, @dcoutts here's just the simple low / high mark strategy.  To express it independently of the protocol I used a recursive type:
```
data MkPipelineDecision where
     MkPipelineDecision
       :: (forall n. Nat n
                  -> BlockNo
                  -> BlockNo
                  -> (PipelineDecision n, MkPipelineDecision))
       -> MkPipelineDecision
```
where
```
data PipelineDecision n where
    Request           :: PipelineDecision Z
    Pipeline          :: PipelineDecision n
    CollectOrPipeline :: PipelineDecision (S n)
    Collect           :: PipelineDecision (S n)
```

It allows to evolve the pipelining strategy together with the protocol.  I haven't look yet how to do the more sophisticated strategy which block fetch is using, but maybe this would work as an initial version.

To keep the pipelining strategy in `ouroboros-network` (as is the case of block fetch), I'd propose to change the `ClientPipelined` module and export an interface which allows to plug any pipe lining strategy, alternatively; alternatively we could provide a client which takes a set of callbacks which are supplied by the consensus layer (I think this would be simpler approach, there's no need for consensus to operate with full generality of the protocol wrapper types).

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Protocol/ChainSync/ExamplesPipelined.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Protocol/ChainSync/ExamplesPipelined.hs
@@ -10,8 +10,10 @@ module Ouroboros.Network.Protocol.ChainSync.ExamplesPipelined
     , Tip
     , chainSyncClientPipelinedMax
     , chainSyncClientPipelinedMin
+    , chainSyncClientPipelinedLowHigh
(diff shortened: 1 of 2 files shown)


coot:
Closing this PR, it is subsumed by #1033

PR #1037, opened 2019-09-17
[#1029] Trace ApplyTxErrs along with rejected txs in Mempool

### LINKED ISSUE #1029: Mempool: Trace rejected tx along with reason
The `TraceMempoolRejectedTxs` constructor should probably look something like this instead:
```
TraceMempoolRejectedTxs
     { _txs          :: [(GenTx blk, ApplyTxErr blk)]
       ...
     }
```

intricate:
Closes #1029.


intricate:
Should we also trace the `ApplyTxErr` in `TraceMempoolRemoveTxs`? I'm not sure whether we're interested in tracing the reason why a once valid transaction from the mempool is no longer valid. I would just assume that it's no longer valid due to changes in the ledger state.

intricate:
> LGTM. Bit of a ptiy that we lost the field names, but I guess it's okay :)

:cry:

PR #1039, opened 2019-09-18
Check that ChainDB is in NF

edsko:
This is WIP, and depends on various commits to dependencies.


edsko:
First commit is #1039.

PR #1040, opened 2019-09-18
Generalize StrictT(M)Var invariant

edsko:
Rather than just returning a `Bool`, we return `Maybe` an error message. This paves the way for more detailed information we get back from the NF check.

Marking this as a draft because still untested.

### DIFF
--- a/io-sim-classes/io-sim-classes.cabal
+++ b/io-sim-classes/io-sim-classes.cabal
@@ -18,6 +18,11 @@ source-repository head
   location: https://github.com/input-output-hk/ouroboros-network
   subdir:   io-sim-classes
 
+flag checktvarinvariant
+  Description: Enable runtime invariant checks on StrictT(M)Var
+  Manual: True
+  Default: False
+
 library
   hs-source-dirs:      src
 
@@ -53,3 +58,5 @@ library
                        -Wno-unticked-promoted-constructors
                        -fno-ignore-asserts
 
+  if flag(checktvarinvariant)
+    cpp-options: -DCHECK_TVAR_INVARIANT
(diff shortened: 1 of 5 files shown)


edsko:
Ok, this works. Ready for review and to be merged.

io-sim-classes/io-sim-classes.cabal
```diff
@@ -18,6 +18,11 @@ source-repository head
   location: https://github.com/input-output-hk/ouroboros-network
   subdir:   io-sim-classes
 
+flag checktvarinvariant
```
mrBliss: When/where do we enable this flag?

Thinking out loud: we want to enable it *always* (?) for `test-consensus` and `test-storage`. We don't want to have to pass it manually via the cmd line.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
```diff
@@ -147,10 +145,22 @@ readTMVar StrictTMVar { tmvar } = Lazy.readTMVar tmvar
...
     Lazy.swapTMVar tmvar a
 
 isEmptyTMVar :: MonadSTM m => StrictTMVar m a -> STM m Bool
 isEmptyTMVar StrictTMVar { tmvar } = Lazy.isEmptyTMVar tmvar
+
+{-------------------------------------------------------------------------------
+  Dealing with invariants
+-------------------------------------------------------------------------------}
+
+checkInvariant :: HasCallStack => Maybe String -> m x -> m x
```
mrBliss: Would it be useful to reuse this function to check other invariants than this one? I mean instead of `assert`, which does not support a custom error message (or a call stack with a depth > 1).

edsko:
It's a good suggestion regarding enabling the flag on CI, but I don't know how to do it. Opened a separate ticket for it https://github.com/input-output-hk/ouroboros-network/issues/1047 .

PR #1042, opened 2019-09-18
Removed BlockStream module in tests

coot:


coot:
This is not even included in the `cabal` file; these are tests for a protocol that was removed.

PR #1044, opened 2019-09-18
[#975] Introduce strict FingerTree

### LINKED ISSUE #975: Introduce strict variant of `FingerTree`
We should introduce a strict variant of `FingerTree` such that we can ensure its elements are in WHNF.

@edsko: 
> folks, I started adding some isNormalForm assertions to the chain DB, but many of them fail. some of those failures might be genuine bugs, but... ChainFragment uses FingerTree, which is not strict. There are some packages with a strict variant, but the paper (http://www.staff.city.ac.uk/~ross/papers/FingerTree.pdf) says that "the structure makes essential use of laziness, though it is also suitable for strict languages that provide a lazy evaluation primitive", so I worry about using a strict variant -- it may affect asymptotic complexity in non-trivial ways.

@dcoutts:
> Right we want it to be strict in its leaves/elements/content
> yeah it'd mess up the asymptotic complexity

@edsko:
> right, that's what I was was worried about. @intricate, perhaps we should put FingerTree behind a module of our own (FingerTree.Strict) that does ensure that the elements are always in WHNF.

intricate:
Closes #975

### DIFF
--- a/ouroboros-network/src/Data/FingerTree/Strict.hs
+++ b/ouroboros-network/src/Data/FingerTree/Strict.hs
@@ -1,0 +1,270 @@
+{-# LANGUAGE BangPatterns               #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+
+-- | Strict variants of 'FingerTree' operations.
+--
+-- TODO: Move this to @cardano-prelude@ as a new module, @Data.FingerTree.Strict@.
+module Data.FingerTree.Strict
+  ( StrictFingerTree
+
+    -- * Construction
+  , empty
+  , singleton
+  , (<|)
+  , (|>)
+  , (><)
+  , fromList
+
+    -- * Deconstruction
+  , null
+
+    -- ** Examining the ends
+  , viewl
+  , viewr
+
+    -- ** Search
+  , SearchResult(..)
+  , search
+
+    -- ** Splitting
+    -- | These functions are special cases of 'search'.
+  , split
+  , takeUntil
+  , dropUntil
+
+    -- * Transformation
+  , reverse
+
+    -- ** Maps
+  , fmap'
+
+  -- Re-export from "Data.FingerTree"
+  , Measured(..)
+  , ViewL (..)
+  , ViewR (..)
+  ) where
+
+import           Prelude hiding (null, reverse)
+
+import           Data.FingerTree (Measured (..), ViewL (..), ViewR (..))
+import qualified Data.FingerTree as FT
+import           Data.Foldable (foldl')
+import           Data.Monoid (Monoid (..))
+import           Data.Semigroup (Semigroup (..))
+import           GHC.Generics (Generic)
+
+infixr 5 ><
+infixr 5 <|
+infixl 5 |>
+
+-- | A @newtype@ wrapper around a 'FT.FingerTree', representing a finger tree
+-- that is strict in its values.
+--
+-- This strictness is not enforced at the type level, but rather by the
+-- construction functions in this module. These functions essentially just
+-- wrap the original "Data.FingerTree" functions while forcing the provided
+-- value to WHNF.
+newtype StrictFingerTree v a = SFT (FT.FingerTree v a)
+  deriving (Eq, Ord, Show)
+
+deriving instance Foldable (StrictFingerTree v)
+deriving instance (Measured v a) => Measured v (StrictFingerTree v a)
+
+{-------------------------------------------------------------------------------
+  Construction
+-------------------------------------------------------------------------------}
+
+-- | /O(1)/. The empty sequence.
+empty :: Measured v a => StrictFingerTree v a
+empty = SFT (FT.empty)
+
+-- | /O(1)/. A singleton sequence.
+singleton :: Measured v a => a -> StrictFingerTree v a
+singleton !a = SFT (FT.singleton a)
+
+-- | /O(n)/. Create a sequence from a finite list of elements.
+-- The opposite operation 'toList' is supplied by the 'Foldable' instance.
+fromList :: (Measured v a) => [a] -> StrictFingerTree v a
+fromList !xs = foldl' (|>) (SFT FT.empty) xs
+
+-- | /O(1)/. Add an element to the left end of a sequence.
+-- Mnemonic: a triangle with the single element at the pointy end.
+(<|) :: (Measured v a) => a -> StrictFingerTree v a -> StrictFingerTree v a
+(!a) <| SFT ft = SFT (a FT.<| ft)
+
+-- | /O(1)/. Add an element to the right end of a sequence.
+-- Mnemonic: a triangle with the single element at the pointy end.
+(|>) :: (Measured v a) => StrictFingerTree v a -> a -> StrictFingerTree v a
+SFT ft |> (!a) = SFT (ft FT.|> a)
+
+-- | /O(log(min(n1,n2)))/. Concatenate two sequences.
+(><) :: (Measured v a)
+     => StrictFingerTree v a -> StrictFingerTree v a -> StrictFingerTree v a
+SFT ft >< SFT ft' = SFT (ft FT.>< ft')
+
+{-------------------------------------------------------------------------------
+  Deconstruction
+-------------------------------------------------------------------------------}
+
+-- | /O(1)/. Is this the empty sequence?
+null :: StrictFingerTree v a -> Bool
+null (SFT ft) = FT.null ft
+
+{-------------------------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


intricate:
I don't think this is entirely correct yet. I still think many of these functions can introduce lazy values to the `FingerTree`.

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
```diff
@@ -0,0 +1,346 @@
+{-# LANGUAGE BangPatterns               #-}
+{-# LANGUAGE CPP                        #-}
+#if __GLASGOW_HASKELL__ >= 706
```
mrBliss: We don't care about compatibility with GHC (and base) versions other than the current one (8.6, in the future 8.8, I expect), so this can go.

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
```diff
@@ -0,0 +1,346 @@
...
+  , split
+  , takeUntil
+  , dropUntil
+
+    -- * Transformation
+  , reverse
+
+    -- ** Maps
+  , fmap'
+  , fmapWithPos
```
mrBliss: Hmm, some of these functions are not used in our codebase. Maybe we can leave them out for now and add them only when they're needed?

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
```diff
@@ -0,0 +1,346 @@
...
+
+-- | /O(n)/. Create a sequence from a finite list of elements.
+-- The opposite operation 'toList' is supplied by the 'Foldable' instance.
+fromList :: (Measured v a) => [a] -> FingerTree v a
+fromList !xs = foldl' (|>) (SFT $ FT.empty) xs
+
+-- | /O(1)/. Add an element to the left end of a sequence.
+-- Mnemonic: a triangle with the single element at the pointy end.
+(<|) :: (Measured v a) => a -> FingerTree v a -> FingerTree v a
+(!a) <| SFT ft = SFT (a FT.<| ft)
```
mrBliss: Are the parentheses around `!a` needed?

intricate: Yeah, they're required here.

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
intricate: Will do.

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
```diff
@@ -0,0 +1,264 @@
...
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+
+-- | Strict variants of 'FingerTree' operations.
+--
+-- TODO: Move this to @cardano-prelude@ as a new module, @Data.FingerTree.Strict@.
+module Ouroboros.Network.StrictFingerTree
```
coot: For me it would make more sense to name this module as `Data.Fingertree.Strict`.

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
```diff
@@ -0,0 +1,264 @@
...
+-- | /O(1)/. Is this the empty sequence?
+null :: FingerTree v a -> Bool
+null (SFT ft) = FT.null ft
+
+{-------------------------------------------------------------------------------
+  Examining the ends
+-------------------------------------------------------------------------------}
+
+-- | View of the left end of a sequence.
+data ViewL s a
```
coot: It's not clear to me why we need `ViewL` and `ViewR`, cannot we just use the originals directly? `fromOriginalViewL` will not be needed.

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
```diff
@@ -0,0 +1,264 @@
...
+
+{-------------------------------------------------------------------------------
+  Search
+-------------------------------------------------------------------------------}
+
+-- | A result of 'search', attempting to find a point where a predicate
+-- on splits of the sequence changes from 'False' to 'True'.
+--
+-- @since 0.1.2.0
+data SearchResult v a
```
coot: The same as for `ViewL`; why not just use the original?

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
```diff
@@ -0,0 +1,264 @@
...
+infixl 5 |>, :>
+
+-- | A @newtype@ wrapper around a 'FT.FingerTree', representing a finger tree
+-- that is strict in its values.
+--
+-- This strictness is not enforced at the type level, but rather by the
+-- construction functions in this module. These functions essentially just
+-- wrap the original "Data.FingerTree" functions while forcing the provided
+-- value to WHNF.
+newtype FingerTree v a = SFT (FT.FingerTree v a)
```
coot: The type should be called `StrictFingerTree`; Warning and error messages will be easier to digest.

ouroboros-network/src/Ouroboros/Network/StrictFingerTree.hs
intricate: You're right. Thanks.

intricate:
Thanks @coot

ouroboros-network/src/Data/FingerTree/Strict.hs
```diff
@@ -0,0 +1,250 @@
...
+reverse (SFT ft) = SFT (FT.reverse ft)
+
+{-------------------------------------------------------------------------------
+  Maps
+-------------------------------------------------------------------------------}
+
+-- | Like 'fmap', but with constraints on the element types.
+fmap' :: (Measured v1 a1, Measured v2 a2)
+      => (a1 -> a2) -> StrictFingerTree v1 a1 -> StrictFingerTree v2 a2
+fmap' f (SFT ft) = SFT (FT.fmap' f ft)
```
mrBliss: I don't think this is strict: each element `x` in the finger tree will be a thunk `f x`.

intricate: Yes, I believe you're right. Currently re-thinking how I should go about this.

coot: I think this is tricky, normally one would get strictness from a bang in the data constructor, which we don't do here.  Adding a bang to result of `f` will not be right (evaluate `f x` when you evaluate `f x`; by strictness `x` is already evaluated). You probably need to re-implement `fmap` without using `FingerTree` `Functor` instance.

intricate: So I went through this with @dcoutts yesterday and he suggested a good solution, at the cost of some performance. The solution is similar to one that @mrBliss suggested on https://github.com/input-output-hk/cardano-ledger/pull/637#discussion_r326649968.

However, the problem with @mrBliss' solution is that the `mappend` implementation for `()` is completely lazy: `_ <> _ = ()`. I didn't know this at the time so I went with another solution.

In this case, I went with this:

```
-- | Like 'fmap', but with constraints on the element types.
fmap' :: (Measured v1 a1, Measured v2 a2)
      => (a1 -> a2) -> StrictFingerTree v1 a1 -> StrictFingerTree v2 a2
fmap' f (SFT ft) = foldMap (\x -> seq x StrictUnit) sft `seq` sft
  where
    sft = SFT (FT.fmap' f ft)

data StrictUnit = StrictUnit

instance Semigroup StrictUnit where
  StrictUnit <> StrictUnit = StrictUnit

instance Monoid StrictUnit where
  mempty = StrictUnit
```

mrBliss:
https://github.com/input-output-hk/cardano-prelude/pull/76 is merged, which also defines `StrictUnit`. Should it export `StrictUnit` so that we can reuse it in this PR? (In follow-up PRs)

edsko:
> input-output-hk/cardano-prelude#76 is merged, which also defines StrictUnit. Should it export StrictUnit so that we can reuse it in this PR? (In follow-up PRs)

No, it should not define `StrictUnit` at all, but rather use `forceElemsToWhnf` directly; that was kind of the point of the PR against prelude, keep that nasty `StrictUnit` stuff isolated in one place.

mrBliss:
> > [input-output-hk/cardano-prelude#76](https://github.com/input-output-hk/cardano-prelude/pull/76) is merged, which also defines StrictUnit. Should it export StrictUnit so that we can reuse it in this PR? (In follow-up PRs)
> 
> No, it should not define `StrictUnit` at all, but rather use `forceElemsToWhnf` directly; that was kind of the point of the PR against prelude, keep that nasty `StrictUnit` stuff isolated in one place.

Oh yes, of course.

PR #1048, opened 2019-09-20
Fix memory leak in ledger DB

edsko:


edsko:
So I tested this with the Byron proxy. Memory profile before this patch:

cardano-byron-proxy

And after:

cardano-byron-proxy

It might be wishful thinking, but it looks to me that the sawteeth in the `before` picture are slanted slightly upwards and the ones in the `after` picture are more even? Perhaps this will remove (some?) of the remaining slower memory leak that @erikd was seeing.

edsko:
By the way, this leak was pointed out by the WIP PR using `noUnexpectedThunks`, which gave a nice invariant violation of

```
Invariant violation: Unexpected thunk in ["(,)","Seq","LedgerDB"]
```

PR #1052, opened 2019-09-23
Make sure Byron header is entirely strict

edsko:
The presence of an `Either` was causing a memory leak.

PR #1053, opened 2019-09-23
export Byron header constructors

avieth:
#1052


avieth:
> Just curious, why export these as patterns...?

Because I can never remember the syntax for importing constructors of associated data families :)

PR #1033, opened 2019-09-16
Update consensus to do pipelining

edsko:
This builds on #1027.

**NOTE:** this changes the instantiation of the ChainSync protocol: the `tip` parameter is now instantiated with the new `Tip blk` type (see `Ouroboros.Consensus.ChainSync`) instead of `Point blk`. The cardano-node and -byron-proxy will have to be updated.

PR #1054, opened 2019-09-23
Block Fetch optimisations

mrBliss:
See the separate commits for more details.

Without these optimisations, enabling pipelining for the ChainSync client (#1033) actually slows it down! This is because the ChainSync client quickly downloads all headers it can (2k), after which the Block Fetch logic has to deal with much larger candidate fragments. It performs multiple filtering passes each time, each of which is linear in the fragment size. The main optimisation in this PR cuts down the filtering work that is required.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
+++ b/ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
@@ -24,8 +24,10 @@ module Ouroboros.Network.BlockFetch.Decision (
 
 import qualified Data.Set as Set
 
-import           Data.List (sortBy, groupBy, transpose)
+import           Data.List (foldl', groupBy, sortBy, transpose)
 import           Data.Function (on)
+import           Data.Maybe (mapMaybe)
+import           Data.Set (Set)
 
 import           Control.Exception (assert)
 import           Control.Monad (guard)
@@ -147,6 +149,7 @@ fetchDecisions
   -> FetchMode
   -> AnchoredFragment header
   -> (Point block -> Bool)
+  -> MaxSlotNo
   -> [(AnchoredFragment header, PeerInfo header extra)]
   -> [(FetchDecision (FetchRequest header), PeerInfo header extra)]
 fetchDecisions fetchDecisionPolicy@FetchDecisionPolicy {
@@ -156,7 +159,8 @@ fetchDecisions fetchDecisionPolicy@FetchDecisionPolicy {
                }
                fetchMode
                currentChain
-               fetchedBlocks =
+               fetchedBlocks
+               fetchedMaxSlotNo =
 
     -- Finally, make a decision for each (chain, peer) pair.
     fetchRequestDecisions
@@ -183,6 +187,7 @@ fetchDecisions fetchDecisionPolicy@FetchDecisionPolicy {
     -- Filter to keep blocks that have not already been downloaded.
   . filterNotAlreadyFetched
       fetchedBlocks
+      fetchedMaxSlotNo
 
     -- Select the suffix up to the intersection with the current chain.
   . selectForkSuffixes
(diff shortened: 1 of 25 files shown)


ouroboros-network/src/Ouroboros/Network/ChainFragment.hs
```diff
@@ -473,14 +474,33 @@ filter :: HasHeader block
...
        -> ChainFragment block
        -> [ChainFragment block]
-filter p = go [] Empty
+filter p = filterWithStop p (const False)
+
+-- | \( O\(n\) \). Same as 'filter', but as soon as the stop condition is
+-- true, the filtering stop and the remaining fragment (starting with the
+-- first element for which the stop condition is true) is the final fragment
+-- in the returned list.
+--
```
mrBliss: Add example:

```
1 2 3 4 5 6
filter odd -> [[1], [3], [5]]
filterWithStop odd (>= 4) -> [[1], [3], [4,5,6]]
```

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -547,6 +564,48 @@ filterNotAlreadyInFlightWithOtherPeers FetchModeBulkSync chains =
             _other                  -> peerFetchBlocksInFlight inflight
         | (_, status, inflight, _) <- chains ]
 
+    -- The highest slot number that is or has been in flight for any peer.
+    maxSlotNoInFlightWithOtherPeers = maximum $ Nothing :
+      [ peerFetchMaxSlotNo inflight | (_, _, inflight, _) <- chains ]
+
+-- | Filter a fragment, but stop the filtering (and add the remaining fragment
```
edsko: Let's say something like: this is intended to an optimized but equivalent form of `filter`. For this to be the case, the following precondition needs to be satisfied: (...).

edsko: And it might be nice to have a test that captures that particular use case (where `filterWithStop` is in fact equivalent to `fitler`, with a non-trivial predicate).

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -547,6 +564,48 @@ filterNotAlreadyInFlightWithOtherPeers FetchModeBulkSync chains =
...
+-- to remove.
+--
+-- PRECONDITION: for all @hdr@ in the chain fragment: if @blockSlot hdr >
+-- maxSlotNo@ then the predicate should not hold for any header after @hdr@ in
+-- the chain fragment.
+--
+-- For example, when filtering out already downloaded blocks from the
+-- fragment, it does not make sense to keep filtering after having encountered
+-- the highest slot number the ChainDB has seen so far, all blocks with a
+-- greater slot number cannot not have been downloaded yet. When the candidate
```
edsko:
```
-- For example, when filtering out already downloaded blocks from the
-- fragment, it does not make sense to keep filtering after having encountered
-- the highest slot number the ChainDB has seen so far: blocks with a
-- greater slot number cannot have been downloaded yet. 
```

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -547,6 +564,48 @@ filterNotAlreadyInFlightWithOtherPeers FetchModeBulkSync chains =
...
+-- fragments get far ahead of the current chain, e.g., @2k@ headers, this
+-- optimisation avoids the linear cost of filtering these headers when we know
+-- in advance they will all remain in the final fragment.
+--
+-- In case the given slot number is 'Nothing', no filtering takes place, as
+-- there should be no matches.
```
edsko: add: "because we haven't downloaded any blocks yet", and make this part of the paragraph before.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -547,6 +564,48 @@ filterNotAlreadyInFlightWithOtherPeers FetchModeBulkSync chains =
...
+--
+-- For example, when filtering out blocks already in-flight for the given
+-- peer, the given @maxSlotNo@ can correspond to the block with the highest
+-- slot number that so far has been in-flight for the given peer. When no
+-- blocks have been in-flight yet, @maxSlotNo@ can be 'Nothing', in which case
+-- no filtering needs to take place, which makes sense, as there are no blocks
+-- to filter out.
```
edsko: Let's add a sentence here explaining that this is conservative, and that the case where blocks (for whatever reason) need to be in flight more than once would okay.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -734,14 +793,15 @@ fetchRequestDecisions
                                              peer)]
   -> [(FetchDecision (FetchRequest header),  peer)]
 fetchRequestDecisions fetchDecisionPolicy fetchMode chains =
-    go nConcurrentFetchPeers0 Set.empty chains
+    go nConcurrentFetchPeers0 Set.empty Nothing chains
   where
-    go !_ !_ [] = []
```
edsko: Let's give this a type signature, which would have made it more obvious that now we have another bang on a maybe..

PR #1055, opened 2019-09-24
Fix memory leak in ChainStateWithEBBs

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/Protocol/WithEBBs.hs
```diff
@@ -1,3 +1,6 @@
+{-# LANGUAGE BangPatterns            #-}
+{-# LANGUAGE DeriveAnyClass          #-}
```
nc6: Why is this needed? Or `DeriveGeneric`?

edsko: It's not yet, but it will in my next PR :) Remnant from when I rebased this on `master`. But we'll need it as soon as I submit the next (big) PR.

PR #1056, opened 2019-09-24
Avoid bang on `Maybe` in registry

edsko:
Not a big deal either way, but might as well avoid the confusion.


edsko:
Actually, let me push another change here, don't merge yet.

PR #1057, opened 2019-09-24
Introduce `Handle` abstraction for `HasFS`

edsko:
This will give us a type that we can provide some instances for,
avoiding the need to propagate up constraints on `h`.


ouroboros-consensus/src/Ouroboros/Storage/FS/API/Types.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveGeneric      #-}
 {-# LANGUAGE RecordWildCards    #-}
 {-# LANGUAGE StandaloneDeriving #-}
```
mrBliss: Mention the new orphans in the comment above `-Wno-orphans`

PR #1058, opened 2019-09-24
Introduce CurrentEBB abstraction

edsko:
This replaces a `Maybe`, which was not strict in its argument (and can
be given a convenient `NoUnexpectedThunks` later).

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
@@ -87,7 +87,7 @@ import           Ouroboros.Storage.EpochInfo (EpochInfo (..))
 import           Ouroboros.Storage.FS.API (HasFS, createDirectoryIfMissing)
 import           Ouroboros.Storage.FS.API.Types (MountPoint (..))
 import           Ouroboros.Storage.FS.IO (ioHasFS)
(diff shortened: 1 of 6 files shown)


ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
```diff
@@ -605,49 +605,51 @@ epochFileParser ImmDbArgs{..} =
...
 
-    extractSlotNoAndEbbHash :: blk -> (SlotNo, Maybe (HeaderHash blk))
+    extractSlotNoAndEbbHash :: blk -> (SlotNo, CurrentEBB (HeaderHash blk))
     extractSlotNoAndEbbHash b =
       -- IMPORTANT: force the slot and the ebbHash, because otherwise we
       -- return thunks that refer to the whole block! See
       -- 'Ouroboros.Consensus.Util.CBOR.readIncrementalOffsets' where we need
       -- to force the decoded value in order to force this computation.
+      --
+      -- TODO: Is this still needed, now that the Maybe is gone..?
```
mrBliss: I think so, note the forcing of the slot (otherwise we keep a `blockSlot b` thunk) in addition to the `ebbHash`.

edsko: Ok, removed.

PR #1061, opened 2019-09-25
Avoid thunks in paths

edsko:
This PR makes 3 changes:

* Make `FsPath` a newtype, using `Text`, and hide its constructor, to avoid any and all thunks inside. This is by far the bulk of this PR, almost of it all mechanical changes though.
* Make the `MockFS` state strict in its fields.
* Make the `PureSimFS` monad strict in the state.


ouroboros-consensus/src/Ouroboros/Storage/FS/API/Types.hs
```diff
@@ -75,7 +82,38 @@ allowExisting openMode = case openMode of
...
+
+instance Show FsPath where
+  show = intercalate "/" . map Strict.unpack . fsPathToList
+
+instance Condense FsPath where
+  condense = show
+
+-- | Constructor for 'FsPath' ensures path is in normal form
+mkFsPath :: [String] -> FsPath
+mkFsPath = fsPathFromList . map Strict.pack
```
mrBliss: What about `OverloadedLists` to avoid changing all occurrences of `[..]` to `mkFsPath [..]`? 😁

ouroboros-consensus/src/Ouroboros/Storage/FS/API/Types.hs
```diff
@@ -75,7 +82,38 @@ allowExisting openMode = case openMode of
...
+
+-- | Split 'FsPath' is essentially @(init fp, last fp)@
+--
+-- Like @init@ and @last@, undefined if empty.
```
mrBliss: undefined -> Nothing

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ImmutableDB/TestBlock.hs
```diff
@@ -274,14 +276,22 @@ shrinkCorruptions :: Corruptions -> [Corruptions]
...
     [ testProperty "TestBlock Binary roundtrip"   prop_TestBlock_Binary
     , testProperty "testBlockEpochFileParser"     prop_testBlockEpochFileParser
     ]
+
+{-------------------------------------------------------------------------------
+  Orphans
+-------------------------------------------------------------------------------}
+
+instance Arbitrary FsPath where
+  arbitrary = undefined
```
mrBliss: ?

ouroboros-consensus/src/Ouroboros/Storage/FS/API/Types.hs
edsko: Good point.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ImmutableDB/TestBlock.hs
edsko: We only use the `shrink` here. It's just for this particular test.

mrBliss: Use `error` with a comment/message stating that, because `undefined` looks like something temporary that was left in by accident.

edsko: Fair enough :)

PR #1071, opened 2019-09-27
Moved Tip to Ouroboros.Network.Block

coot:


ouroboros-network/src/Ouroboros/Network/Block.hs
```diff
@@ -186,6 +190,33 @@ castPoint (Point (At (Point.Block slot hash))) = Point (block slot hash)
 blockPoint :: HasHeader block => block -> Point block
 blockPoint b = Point (block (blockSlot b) (blockHash b))
 
+{-------------------------------------------------------------------------------
+  Tip of a chain
+-------------------------------------------------------------------------------}
+
+-- | Used in chain-sync protoco to advertise the tip of the server's chain.
```
mrBliss: missing l after protoco

mrBliss:
Oh, this type should be used by `Ouroboros.Network.Protocol.ChainSync.Examples*`

coot:
Yes it could, good point, also in some of the test we use a pair instead of a `Tip`; I'll leave it as issue.

coot:
Ref #1072

PR #1077, opened 2019-09-30
Introduce `IOLike` monad

edsko:
This consists of two commits:

* The first introduces `Ouroboros.Consensus.Util.IOLike`, but simply re-exports all the `Control.Monad.Class.*` classes from there; the goal here is to simply re-route the imports, but not maky any real change.
* The second then _hides_ a lot of those classes, instead introducing `IOLike` as a new monad, and then updates the code accordingly.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
@@ -1,13 +1,88 @@
+{-# LANGUAGE FlexibleContexts #-}
+
 module Ouroboros.Consensus.Util.IOLike (
-    module X
+    IOLike
+    -- * Re-exports
+    -- *** MonadThrow
+  , uninterruptibleCancel
+    -- *** MonadSTM
+  , STM
+  , atomically
+  , check
+  , retry
+    -- ** Queues (TODO: Should we have a strict queue?)
+  , TQueue
+  , newTQueue
+  , readTQueue
+  , writeTQueue
+    -- ** StrictTVar
+  , StrictTVar
+  , modifyTVar
+  , readTVar
+  , uncheckedNewTVarM
+  , updateTVar
+  , writeTVar
+    -- *** StrictTMVar
+  , StrictTMVar
+  , putTMVar
+  , readTMVar
+  , takeTMVar
+  , tryTakeTMVar
+  , swapTMVar
+  , uncheckedNewEmptyTMVarM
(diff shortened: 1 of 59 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Util/Random.hs
```diff
@@ -26,7 +26,7 @@ import           Crypto.Random (ChaChaDRG, MonadPseudoRandom, MonadRandom (..),
 import           Data.List (genericLength)
 import           Data.Word (Word64)
 
-import           Control.Monad.Class.MonadSay
+--import ToBeReplaced
```
mrBliss: This can go

ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
```diff
@@ -1,13 +1,88 @@
+{-# LANGUAGE FlexibleContexts #-}
+
 module Ouroboros.Consensus.Util.IOLike (
-    module X
```
mrBliss: Any reason for not simply re-export the modules, except for documentation purposes?

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -204,7 +204,7 @@ deriving instance (TestConstraints blk, Eq   it, Eq   rdr)
 deriving instance (TestConstraints blk, Show it, Show rdr)
                => Show (Success blk it rdr)
 
-run :: forall m blk. (MonadSTM m, MonadAsync m)
+run :: forall m blk. (IOLike m)
```
mrBliss: Get rid of the parens for consistency?

ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
edsko: Because we don't want to export `MonadAsync` and co (that should be `IOLike`). Also, it's nice to see I think at a glance the API we actually depend on.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
edsko: We're not very consistent about parens for single constraints anyway... :/

PR #1080, opened 2019-09-30
Add subdirs for cardano-shell in stack.yaml

intricate:
The `cardano-shell` package was recently moved in to a subdirectory in the repo. This needs to be reflected in `stack.yaml` (for stack users :smile:).

PR #1081, opened 2019-09-30
ChainSync Client: trimming and refactoring 

mrBliss:
* Implement the `docs/ChainSync.md` spec: watch our current chain and trim the
  candidates.

* Introduce `ChainDbView` to abstract over the ChainDB.

* Introduce `Their` and `Our` newtype wrappers to avoid confusing our tip with
  their tip.

* Only store the candidate fragments in the peer map, not the whole candidate
  state. Thread the state of the candidate through manually, without storing
  it in a `StrictTVar`. Split the state up in `NoIntersectionState` and
  `IntersectionState` to be more precise.

* Include our tip and their tip in the `ChainSyncClientException` and add
  field names.

* Fix bug in pipelining: we used our current chain's tip's block number
  instead of the candidate's tip's block number.

* Misc. improvements to the ChainSync client test.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
@@ -2,10 +2,12 @@
 {-# LANGUAGE DeriveAnyClass             #-}
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE DuplicateRecordFields      #-}
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE MultiWayIf                 #-}
 {-# LANGUAGE NamedFieldPuns             #-}
-{-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeFamilies               #-}
@@ -17,8 +19,10 @@ module Ouroboros.Consensus.ChainSyncClient (
   , chainSyncClient
   , bracketChainSyncClient
   , ChainSyncClientException (..)
+  , ChainDbView (..)
   , ClockSkew (..)
-  , CandidateState (..)
+  , Our (..)
+  , Their (..)
     -- * Trace events
   , TraceChainSyncClientEvent (..)
   ) where
@@ -29,14 +33,12 @@ import           Control.Tracer
 import           Data.List (find)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
+import           Data.Maybe (isJust)
 import           Data.Typeable (Typeable)
 import           Data.Void (Void)
 import           Data.Word (Word64)
-import           GHC.Generics (Generic)
 
 import           Control.Monad.Class.MonadThrow
-
-import           Cardano.Prelude (NoUnexpectedThunks)
 
 import           Network.TypedProtocol.Pipelined
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment (..))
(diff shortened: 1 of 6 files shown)


mrBliss:
#1080 needs to be merged first to fix the build.

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -144,225 +236,283 @@ chainSyncClient
...
+          -- Our current chain didn't change, nothing to do
+          return $ Just is
+
+        | isJust (AF.intersectionPoint currentFrag' candidateFrag) ->
+          -- Our current chain changed, but it still intersects with candidate
+          -- fragment, so update the 'currentFrag' field and trim to the
+          -- candidate fragment to the same anchor point.
+          case AF.splitAfterPoint candidateFrag (AF.anchorPoint currentFrag') of
+           Nothing -> error
+             "anchor point must be on candidate fragment if they intersect"
```
edsko: Let's motivate this:

* Before the update to our fragment, both fragments were anchored at the same anchor.
* We still have an intersection. 
* The number of blocks after the intersection cannot have shrunk, but could have increased.
* If it did increase, the anchor point will have shifted up.
* It can't have moved up past the intersection point (because then there would be no intersection anymore).
* This means the new anchor point must be between the old anchor point and the new intersection point.
* Since we know both the old anchor point and the new intersection point exist on their fragment, the new anchor point must also.

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -144,225 +236,283 @@ chainSyncClient
...
+          drainThePipe n $ do
+            currentFrag       <- getCurrentChain
+            currentChainState <- ouroborosChainState <$> getCurrentLedger
+            ourTip            <- Our <$> getOurTip
+            let points = AF.selectPoints (map fromIntegral offsets) currentFrag
+                nis = NoIntersectionState
+                  { currentFrag       = currentFrag
+                  , currentChainState = currentChainState
+                  , currentTip        = ourTip
+                  }
```
edsko: Duplication of initialization logic.

edsko: (Above and below also, since we don't want to do the genesis special case anymore.)

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -376,9 +526,6 @@ chainSyncClient mkPipelineDecision0
       -- the transaction, because we might have to retry the transaction if the
       -- ledger state is too far behind the upstream peer (see below).
       curLedger <- ledgerState <$> getCurrentLedger
```
edsko: Let's document that we don't need this to be consistent with our view of our chain: we use it _only_ to validate _their_ header.

edsko: (Which is true even for the special case below.)

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -144,225 +236,283 @@ chainSyncClient
...
+    -- fragment to the new current chain fragment's anchor point. If not,
+    -- return 'Nothing'.
+    intersectsWithCurrentChain
+      :: IntersectionState blk tip
+      -> STM m (Maybe (IntersectionState blk tip))
+    intersectsWithCurrentChain is@IntersectionState { candidateFrag, currentFrag } = do
+      currentFrag' <- getCurrentChain
+      currentTip'  <- Our <$> getOurTip
+      if
+        | AF.headPoint currentFrag == AF.headPoint currentFrag ->
```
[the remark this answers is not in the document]
edsko: This will have caused us to never trim, so we should strengthen the tests to check that we are trimming (some bound on length? something else?).

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -144,225 +236,283 @@ chainSyncClient
...
-      -- means the candidate is no longer eligible after all.
-      let points = AF.selectPoints (map fromIntegral offsets) curChain
+      let points = AF.selectPoints (map fromIntegral offsets) currentFrag
       return $ SendMsgFindIntersect points $ ClientPipelinedStIntersect
-        { recvMsgIntersectFound    = intersectFound . castPoint
-        , recvMsgIntersectNotFound = intersectNotFound
+        { recvMsgIntersectFound    = \i theirTip ->
+            intersectFound nis (castPoint i) (Their theirTip)
+        , recvMsgIntersectNotFound = \theirTip ->
+            initialIntersectNotFound nis (Their theirTip)
```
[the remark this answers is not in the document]
nfrisby: Just chiming in with a snippet from the write-up I'm drafting.

> C1 and C2 are _wedged_ if their intersection strictly precedes both of the chains' immutable tips: `(C1 ∩ C2 < C1^k) AND (C1 ∩ C2 < C2^k)`.

So "no intersection found" -- if I understand correctly -- implies at least one of those conjuncts, but not necessarily both. I'm still internally shedding over the word choice here, but I'm currently using "tilted" (not as good as wedged :( yet!).

> C1 and C2 are _tilted_ if their intersection strictly precedes exactly one of the chains' immutable tips: `(C1 ∩ C2 < C2^k) XOR (C1 ∩ C2 < C1^k)`. Implies: `(C1^k ≤ C1 ∩ C2 < C2^k) XOR (C2^k ≤ C1 ∩ C2 < C1^k)`.

If two nodes' current chains are wedged, then the nodes will always be wedged. If two nodes are merely tilted, then they could recover if the one with the shorter chain grows far enough along the longer chains suffix (perhaps having to first rolling back some).

It makes sense that a ChainSync client's candidate might not provide enough information for it to distinguish between tilted and wedged. In other words, wedged means both of the peered clients should disconnect, whereas tilted means only one of the clients should disconnect. (I think.)

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -89,45 +98,128 @@ bracketChainSyncClient
...
           tracer
           registry
           getIsInvalidBlock
           (readTVar varCandidate)
-        body varCandidate curChain
+        body varCandidate noIntersection
   where
     register = do
-      (curChain, curChainState) <- atomically $ (,)
+      (currentFrag, currentChainState, ourTip) <- atomically $ (,,)
```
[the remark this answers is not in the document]
mrBliss: I moved this into `findIntersection`, which I call in `initialise`, so we don't even need to do it in `bracketChainSyncClient` :slightly_smiling_face:

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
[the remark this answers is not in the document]
mrBliss: The test now checks whether our fragment and the synched fragment have the same anchor point, and whether it forks within `k` blocks. This catches the missing `'`.

PR #1092, opened 2019-10-01
Check ChainSyncClient for thunks

edsko:
The chain sync client stores a tip, but is polymorphic in that tip, and
if there are thunks in that tip, we might leak. For our tests we are
using the example server from the network layer, which is however not
polymorphic in its `tip` type, but uses a non-strict pair. This PR
replaces that with a strict `ExampleTip`, and pushes that through.
The `NoUnexpectedThunks` test for our chain sync tests now passes.

@coot , are you okay with the changes to the network examples here?

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
@@ -240,11 +240,12 @@ chainSyncClient mkPipelineDecision0 getTipBlockNo tracer cfg btime
                 , getOurTip
                 , getIsInvalidBlock
                 }
-                varCandidate = ChainSyncClientPipelined initialise
+                varCandidate = ChainSyncClientPipelined $
+    continueWithState () $ initialise
   where
     -- | Start ChainSync by looking for an intersection between our current
     -- chain fragment and their chain.
-    initialise :: m (Consensus (ClientPipelinedStIdle Z) blk tip m)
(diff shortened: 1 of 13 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -796,8 +816,16 @@ rejectInvalidBlocks tracer registry getIsInvalidBlock getCandidate =
 -- the state explicit in the types and do the check in 'continueWithState'.
 newtype Stateful m blk tip s st = Stateful (s -> m (Consensus st blk tip m))
 
-continueWithState :: s -> Stateful m blk tip s st -> m (Consensus st blk tip m)
-continueWithState s (Stateful f) = f s
+continueWithState :: NoUnexpectedThunks s
+                  => s -> Stateful m blk tip s st -> m (Consensus st blk tip m)
+#if CHECK_TVAR_INVARIANT
```
mrBliss: Thinking out loud: can't we have a single location where we have this `CPP` condition? Currently, this is the only location in `ouroboros-consensus` where we use it, but might that change?

edsko: Let's come back to that once we have a second location :)

edsko: That said, if you can think of any places where we have additional implicit state..?

PR #1095, opened 2019-10-01
[#1094] Introduce StrictSeq

### LINKED ISSUE #1094: Introduce strict variant of Seq
We should introduce a strict variant of `Seq` from `Data.Sequence` such that we can ensure its elements are in WHNF.

intricate:
Closes #1094

### DIFF
--- a/ouroboros-network/src/Data/Sequence/Strict.hs
+++ b/ouroboros-network/src/Data/Sequence/Strict.hs
@@ -1,0 +1,224 @@
+{-# LANGUAGE BangPatterns               #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PatternSynonyms            #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE ViewPatterns               #-}
+
+-- | Strict variants of 'Seq' operations.
+--
+-- TODO: Move this to @cardano-prelude@ as a new module, @Data.Sequence.Strict@.
+module Data.Sequence.Strict
+  ( StrictSeq (Empty, (:<|), (:|>))
+
+    -- * Construction
+  , empty
+  , singleton
+  , (<|)
+  , (|>)
+  , (><)
+  , fromList
+  , toStrict
+
+    -- * Deconstruction
+    -- | Additional functions for deconstructing sequences are available
+    -- via the 'Foldable' instance of 'Seq'.
+
+    -- ** Queries
+  , null
+  , length
+
+    -- * Scans
+  , scanl
+
+    -- * Sublists
+
+    -- ** Sequential searches
+  , dropWhileL
+  , dropWhileR
+  , spanl
+
+    -- * Indexing
+  , take
+  , drop
+  , splitAt
+  ) where
+
+import           Prelude hiding (drop, length, null, scanl, splitAt, take)
+import           Cardano.Prelude (NoUnexpectedThunks (..), forceElemsToWHNF,
+                     noUnexpectedThunksInValues)
+
+import           Codec.Serialise (Serialise)
+import           Data.Foldable (foldl', toList)
+import           Data.Sequence (Seq)
+import qualified Data.Sequence as Seq
+
+infixr 5 ><
+infixr 5 <|
+infixl 5 |>
+
+infixr 5 :<|
+infixl 5 :|>
+
+{-# COMPLETE Empty, (:<|) #-}
+{-# COMPLETE Empty, (:|>) #-}
+
+-- | A @newtype@ wrapper around a 'Seq', representing a general-purpose finite
+-- sequence that is strict in its values.
+--
+-- This strictness is not enforced at the type level, but rather by the
+-- construction functions in this module. These functions essentially just
+-- wrap the original "Data.Sequence" functions while forcing the provided
+-- value to WHNF.
+newtype StrictSeq a = StrictSeq (Seq a)
+  deriving stock (Eq, Ord, Show)
+  deriving newtype (Foldable, Semigroup, Serialise)
+
+-- | Instance for 'StrictSeq' checks elements only
+--
+-- The internal fingertree in 'Seq' might have thunks, which is essential for
+-- its asymptotic complexity.
+instance NoUnexpectedThunks a => NoUnexpectedThunks (StrictSeq a) where
+  showTypeOf _ = "StrictSeq"
+  whnfNoUnexpectedThunks ctxt = noUnexpectedThunksInValues ctxt . toList
+
+-- | A helper function for the ':<|' pattern.
+--
+viewFront :: StrictSeq a -> Maybe (a, StrictSeq a)
+viewFront (StrictSeq xs) = case Seq.viewl xs of
+  Seq.EmptyL   -> Nothing
+  x Seq.:< xs' -> Just (x, StrictSeq xs')
+
+-- | A helper function for the ':|>' pattern.
+--
+viewBack :: StrictSeq a -> Maybe (StrictSeq a, a)
+viewBack (StrictSeq xs) = case Seq.viewr xs of
+  Seq.EmptyR   -> Nothing
+  xs' Seq.:> x -> Just (StrictSeq xs', x)
+
+-- | A bidirectional pattern synonym matching an empty sequence.
+pattern Empty :: StrictSeq a
+pattern Empty = StrictSeq Seq.Empty
+
+-- | A bidirectional pattern synonym viewing the front of a non-empty
+-- sequence.
+pattern (:<|) :: a -> StrictSeq a -> StrictSeq a
+pattern x :<| xs <- (viewFront -> Just (x, xs))
+  where
+    x :<| xs = x <| xs
+
+-- | A bidirectional pattern synonym viewing the rear of a non-empty
+-- sequence.
+pattern (:|>) :: StrictSeq a -> a -> StrictSeq a
+pattern xs :|> x <- (viewBack -> Just (xs, x))
+  where
+    xs :|> x = xs |> x
+
+{-------------------------------------------------------------------------------
+  Construction
+-------------------------------------------------------------------------------}
+
+-- | \( O(1) \). The empty sequence.
+empty :: StrictSeq a
+empty = Empty
+
+-- | \( O(1) \). A singleton sequence.
+singleton :: a -> StrictSeq a
+singleton !x =  StrictSeq (Seq.singleton x)
+
+-- | \( O(1) \). Add an element to the left end of a sequence.
+-- Mnemonic: a triangle with the single element at the pointy end.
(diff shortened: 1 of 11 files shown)


ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,221 @@
...
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE ViewPatterns #-}
+
+-- | Strict variants of 'Seq' operations.
+--
+-- TODO: Move this to @cardano-prelude@ as a new module, @Data.Sequence.Strict@.
+module Data.Sequence.Strict
+  ( StrictSeq (.., Empty, (:<|), (:|>))
```
mrBliss: I think the `..` (*) and/or the lack of a `COMPLETE` pragma (grep for it, you'll find an example in `AnchoredFragment`) are giving you the warnings about unexhaustive pattern-matches.

(*) Does the raw constructor need to be exported? I would expect not, as it could be used to convert a `Seq` with thunks as elements to a `StrictSeq` without any forcing.

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,221 @@
...
+
+    -- * Construction
+    , empty
```
mrBliss: Can you *de*dent the exports so that they commas line up with the opening and closing parentheses?

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,221 @@
...
+-- This strictness is not enforced at the type level, but rather by the
+-- construction functions in this module. These functions essentially just
+-- wrap the original "Data.Sequence" functions while forcing the provided
+-- value to WHNF.
+newtype StrictSeq a = StrictSeq (Seq a)
+  deriving (Eq, Ord, Show)
+
+deriving instance Foldable StrictSeq
+deriving instance Semigroup (StrictSeq a)
+deriving instance Serialise a => Serialise (StrictSeq a)
```
mrBliss: I think @edsko would like to see `deriving newtype instance` here :slightly_smiling_face:

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,221 @@
...
+  where
+    xs :|> x = xs |> x
+
+{-------------------------------------------------------------------------------
+  Construction
+-------------------------------------------------------------------------------}
+
+-- | \( O(1) \). The empty sequence.
+empty :: StrictSeq a
+empty = StrictSeq (Seq.empty)
```
mrBliss: No parens needed.

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,221 @@
...
+  Scans
+-------------------------------------------------------------------------------}
+
+-- | 'scanl' is similar to 'foldl', but returns a sequence of reduced
+-- values from the left:
+--
+-- > scanl f z (fromList [x1, x2, ...]) = fromList [z, z `f` x1, (z `f` x1) `f` x2, ...]
+scanl :: (a -> b -> a) -> a -> StrictSeq b -> StrictSeq a
+scanl f !z0 (StrictSeq xs) = StrictSeq $ forceElemsToWHNF $ Seq.scanl f z0 xs
+-- scanl f z0 xs = z0 <| snd (mapAccumL (\ x z -> let x' = f x z in (x', x')) z0 xs)
```
mrBliss: Left-over comment?

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,221 @@
...
+
+{-------------------------------------------------------------------------------
+  Sublists
+-------------------------------------------------------------------------------}
+
+-- | \( O(i) \) where \( i \) is the prefix length.  @'dropWhileL' p xs@ returns
+-- the suffix remaining after @'takeWhileL' p xs@.
+dropWhileL :: (a -> Bool) -> StrictSeq a -> StrictSeq a
+dropWhileL p (StrictSeq xs) = StrictSeq (Seq.dropWhileL p xs)
```
mrBliss: BTW, if you're just forwarding to `Seq` without any forcing, you could also use `coerce`, e.g.
```haskell
dropWhileL :: (a -> Bool) -> StrictSeq a -> StrictSeq a
dropWhileL = coerce Seq.dropWhileL
```

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,221 @@
...
+-- | \( O(\log(\min(i,n-i))) \). Split a sequence at a given position.
+-- @'splitAt' i s = ('take' i s, 'drop' i s)@.
+splitAt :: Int -> StrictSeq a -> (StrictSeq a, StrictSeq a)
+splitAt i (StrictSeq xs) = toStrictSeqTuple (Seq.splitAt i xs)
+
+{-------------------------------------------------------------------------------
+  Helpers
+-------------------------------------------------------------------------------}
+
+toStrictSeqTuple :: (Seq a, Seq a) -> (StrictSeq a, StrictSeq a)
```
mrBliss: Doesn't the tuple introduce some laziness again? :thinking:

edsko: I don't  think so. The point is not that we don't return anything without thunks in it. That's not the goal of a strict type. The goal is that _if_ you force a strict seq to WHNF, _then_ all its elements will be, no more.

ouroboros-network/src/Data/Sequence/Strict.hs
intricate: Thanks for informing me of the `COMPLETE` pragmas. I was trying to figure this out.

No, I don't think it needs to be exported. Will fix up.

ouroboros-network/src/Data/Sequence/Strict.hs
intricate: I didn't know that, thanks.

ouroboros-network/src/Data/Sequence/Strict.hs
intricate: Yeah, I don't think this is a problem.

ouroboros-network/src/Data/Sequence/Strict.hs
intricate: I went ahead and used the `deriving newtype (Foldable, Semigroup, Serialise)` syntax.

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,226 @@
...
+
+-- | A @newtype@ wrapper around a 'Seq', representing a general-purpose finite
+-- sequence that is strict in its values.
+--
+-- This strictness is not enforced at the type level, but rather by the
+-- construction functions in this module. These functions essentially just
+-- wrap the original "Data.Sequence" functions while forcing the provided
+-- value to WHNF.
+newtype StrictSeq a = StrictSeq (Seq a)
+  deriving (Eq, Ord, Show)
```
mrBliss: Nitpicking: use `deriving stock` for consistency

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,226 @@
...
+instance NoUnexpectedThunks a => NoUnexpectedThunks (StrictSeq a) where
+  showTypeOf _ = "StrictSeq"
+  whnfNoUnexpectedThunks ctxt = noUnexpectedThunksInValues ctxt . toList
+
+-- | A helper function for the ':<|' pattern.
+--
+viewFront :: StrictSeq a -> Maybe (a, StrictSeq a)
+viewFront (StrictSeq xs) = case Seq.viewl xs of
+                          Seq.EmptyL   -> Nothing
+                          x Seq.:< xs' -> Just (x, StrictSeq xs')
```
mrBliss: Style nazi: I'm not sure what you're aligning these patterns with. I would indent them 4 spaces from the start of the line. Alternatively, indent them 4 (or 2) spaces w.r.t. `case`, but that shifts it so much to the right...

 Same for `viewBack`.

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,226 @@
...
+-- | A helper function for the ':|>' pattern.
+--
+viewBack :: StrictSeq a -> Maybe (StrictSeq a, a)
+viewBack (StrictSeq xs) = case Seq.viewr xs of
+                         Seq.EmptyR   -> Nothing
+                         xs' Seq.:> x -> Just (StrictSeq xs', x)
+
+-- | A bidirectional pattern synonym matching an empty sequence.
+pattern Empty :: StrictSeq a
+pattern Empty = StrictSeq (Seq.Empty)
```
mrBliss: Redundant parens

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,226 @@
...
+StrictSeq s |> (!x) = StrictSeq (s Seq.|> x)
+
+-- | \( O(\log(\min(n_1,n_2))) \). Concatenate two sequences.
+--
+-- @intricate: This probably isn't right.
+(><) :: StrictSeq a -> StrictSeq a -> StrictSeq a
+StrictSeq xs >< StrictSeq ys = StrictSeq (xs Seq.>< ys)
+
+fromList :: [a] -> StrictSeq a
+fromList !xs = foldl' (|>) (StrictSeq Seq.Empty) xs
```
mrBliss: You could use `empty` or `Empty` here

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,226 @@
...
+  where
+    xs :|> x = xs |> x
+
+{-------------------------------------------------------------------------------
+  Construction
+-------------------------------------------------------------------------------}
+
+-- | \( O(1) \). The empty sequence.
+empty :: StrictSeq a
+empty = StrictSeq Seq.empty
```
mrBliss: You could use `Empty` here

ouroboros-network/src/Data/Sequence/Strict.hs
```diff
@@ -0,0 +1,226 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE ViewPatterns #-}
```
mrBliss: I think stylish-haskell would align the `#-}`

PR #1103, opened 2019-10-02
Avoid invariant in `StrictTMVar`, introduce `StrictMVar` (consensus side)

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -0,0 +1,151 @@
...
+-- 'StrictTMVar': if we throw an exception from an STM transaction, the STM
+-- transaction is not executed, and so we would not even be able to provide the
+-- weaker guarantee that we provide for 'StrictMVar'.
+data StrictMVar m a = StrictMVar
+  { invariant :: !(a -> Maybe String)
+    -- ^ Invariant checked whenever updating the 'StrictMVar'.
+  , tmvar     :: !(Lazy.LazyTMVar m a)
+    -- ^ The main TMVar supporting this 'StrictMVar'
+  , tvar      :: !(Lazy.LazyTVar m a)
+    -- ^ TVar for supporting 'readOldValue'
```
mrBliss: There is no `readOldValue`, it is `readMVarSTM`

ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -0,0 +1,151 @@
...
+    --
+    -- This TVar is always kept up to date with the 'LazyTMVar', but holds on
+    -- the old value of the 'LazyTMVar' when it is empty. This is very useful
+    -- to support single writer/many reader scenarios.
+    --
+    -- NOTE: We should always update the 'tmvar' before the 'tvar' so that if
```
mrBliss: Hmm, why lower-case + `''`? This will give Haddock errors. Either use `TMVar`/`TVar` or `@` (or no delimiters).

ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -0,0 +1,151 @@
...
+
+newMVarWithInvariant :: (MonadSTM m, HasCallStack)
+                     => (a -> Maybe String)  -- ^ Invariant (expect 'Nothing')
+                     -> a
+                     -> m (StrictMVar m a)
+newMVarWithInvariant invariant !a =
+    checkInvariant (invariant a) $
+    StrictMVar invariant <$> Lazy.newTMVarM a <*> Lazy.newTVarM a
+
+newEmptyMVar :: MonadSTM m => a -> m (StrictMVar m a)
```
mrBliss: Kind of ugly that it still takes an `a`. Alternatively, scrap the `Empty` variants?

ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -0,0 +1,151 @@
...
+newEmptyMVar = newEmptyMVarWithInvariant (const Nothing)
+
+newEmptyMVarWithInvariant :: MonadSTM m
+                          => (a -> Maybe String)  -- ^ Invariant (expect 'Nothing')
+                          -> a                    -- ^ The initial stale value
+                          -> m (StrictMVar m a)
+newEmptyMVarWithInvariant invariant stale =
```
mrBliss: Also kind of subtle that we're not strict in `a` to allow for `error`. Maybe this pattern should be  documented?

ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
edsko: I agree, but we use them, mostly to block until something becomes ready.

edsko:
Addressed review feedback, force-pushed.

PR #1104, opened 2019-10-02
ChainSyncClient: fix pipelining by using the right clientTipBlockNo

mrBliss:
Instead of using the synced candidate's tip's block number, we were using
their (the server's) chain's tip's block number. This meant that the
difference in the block numbers of the tips was always zero, as we were using
the same block number for both client as server, effectively disabling
pipelining.


edsko:
Looks good, though `pred` needs to worry about the boundary.

Issue #1105, opened 2019-10-02
`cardano-node: Prelude.read: no parse`-crash triggered by `00a51f63`


karknu:
When trying to upgrade `cardano-node` to use a newer version of `ouroboros-network` the node crashed with `cardano-node: Prelude.read: no parse`. I've traced this back to the commit `00a51f63` and reverting it removes the crash.

Output:
```
./dist-newstyle/build/x86_64-linux/ghc-8.6.5/cardano-node-3.0.1.87/x/cardano-node/build/cardano-node/cardano-node --log-config configuration/log-configuration.yaml  --genesis-file configuration/mainnet-genesis.json --genesis-hash 5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb --node-id 0 --topology configuration/topology-proxy-follower.json --host-addr 127.0.0.1 --port 7776 --real-pbft --trace-ip-subscription --trace-mux +RTS -N -A10m -qg -qb -H1024M -M8G
[iohk.cardano.node.CoreId 0.ChainDB:Info:11] [2019-10-02 12:06:31.70 UTC] {"event":{"snapshot":{"kind":"snapshot"},"kind":"TraceLedgerReplayEvent.ReplayFromSnapshot","tip":"TipGen"},"kind":"TraceEvent","tip":"genesis(origin)"}
[iohk.cardano.node.CoreId 0.ChainDB:Info:11] [2019-10-02 12:06:31.70 UTC] {"event":{"immtip":{"kind":"Tip","tip":"genesis(origin)"},"kind":"TraceOpenEvent.OpenedDB","tip":{"kind":"Tip","tip":"89d9b5a@s0"}},"kind":"TraceEvent","tip":"genesis(origin)"}
[iohk.cardano.node.CoreId 0.IpSubscription:Notice:74] [2019-10-02 12:06:31.70 UTC] IPs: Just 0.0.0.0:0 Just [::]:0 [127.0.0.1:7777] Waiting 0.25s before attempting a new connection
[iohk.cardano.node.CoreId 0.IpSubscription:Info:75] [2019-10-02 12:06:31.70 UTC] IPs: Just 0.0.0.0:0 Just [::]:0 [127.0.0.1:7777] Connection Attempt Start, destination 127.0.0.1:7777
[...]

PR #1106, opened 2019-10-02
Restore the original Condense instance for SlotNo

mrBliss:
This addresses #1105: cardano-node breaks because it is using `read` to parse
the `SlotNo` from a trace message, and the tracer is using our `Condense`
instance for `SlotNo`. I warned them that this may cause some accidental
breakage :)


mrBliss:
> `BlockNo` need not technically change, since my problem-causing commit added that instance as opposed to it changing the `SlotNo` instance.
True, I just changed it to be consistent.

Issue #229, opened 2019-02-07
Protocol testing: introduce network latencies


edsko:
Currently we don't use any network latencies. We should introduce latencies (the underlying STM abstraction we use already supports this), ideally in a scriptable way, so that we can generate tests that specify precisely what latency happens when.

mrBliss:
#1131, which addresses this, is blocked by #1147.

PR #1107, opened 2019-10-02
Test for unexpected message delays

nfrisby:
This test confirms that our current test suite is not introducing any unexpected message delays, where _message delays_ is analogous to that from the paper https://eprint.iacr.org/2017/573/20171115:001835.

Issues #229 and #230 will cause this property to fail; they'll have to refine it somehow.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Types.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Types.hs
@@ -207,6 +207,7 @@ data ChainDbEnv m blk = CDB
   , cdbBgThreads      :: !(StrictTVar m [Thread m ()])
     -- ^ The background threads.
   , cdbEpochInfo      :: !(EpochInfo m)
+  , cdbIsEBB          :: !(blk -> Bool)
   } deriving (Generic)
 
 -- | We include @blk@ in 'showTypeOf' because it helps resolving type families
@@ -333,8 +334,10 @@ data TraceOpenEvent blk
 
 -- | Trace type for the various events that occur when adding a block.
 data TraceAddBlockEvent blk
-  = AddedBlockToVolDB    (Point blk)
+  = AddedBlockToVolDB    !(Point blk) !BlockNo !Bool
     -- ^ A block was added to the Volatile DB
+    --
+    -- its point, its block number, whether it's an EBB
 
   | TryAddToCurrentChain (Point blk)
     -- ^ The block fits onto the current chain, we'll try to use it to extend
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Types.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Types.hs
@@ -53,7 +53,7 @@ import           Ouroboros.Network.Block (BlockNo, HasHeader, HeaderHash, Point,
                      SlotNo, StandardHash)
 import           Ouroboros.Network.Point (WithOrigin)
 
-import           Ouroboros.Consensus.Block (BlockProtocol, Header)
+import           Ouroboros.Consensus.Block (BlockProtocol, Header, IsEBB (..))
 import           Ouroboros.Consensus.Ledger.Abstract (ProtocolLedgerView)
 import           Ouroboros.Consensus.Ledger.Extended (ExtValidationError)
 import           Ouroboros.Consensus.Protocol.Abstract (NodeConfig)
@@ -334,10 +334,8 @@ data TraceOpenEvent blk
 
 -- | Trace type for the various events that occur when adding a block.
 data TraceAddBlockEvent blk
-  = AddedBlockToVolDB    !(Point blk) !BlockNo !Bool
+  = AddedBlockToVolDB    !(Point blk) !BlockNo !IsEBB
     -- ^ A block was added to the Volatile DB
-    --
-    -- its point, its block number, whether it's an EBB
 
   | TryAddToCurrentChain (Point blk)
     -- ^ The block fits onto the current chain, we'll try to use it to extend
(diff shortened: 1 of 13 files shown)


ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Types.hs
```diff
@@ -325,8 +326,10 @@ data TraceOpenEvent blk
 
 -- | Trace type for the various events that occur when adding a block.
 data TraceAddBlockEvent blk
-  = AddedBlockToVolDB    (Point blk)
+  = AddedBlockToVolDB    !(Point blk) !BlockNo !Bool
```
nfrisby: I don't like this naked Boolean. Is there a pre-existing `IsEBB`-ish data type?

mrBliss: Not that I know of. There is `BlockOrEBB`, but that contains the `SlotNo` of the block and the `EpochNo` of the EBB. I don't think you can easily compute the `EpochNo` where this event is being traced. So feel free to introduce an `IsEBB`-ish data type (isomorphic to `Bool`).

ouroboros-consensus/test-consensus/Test/Dynamic/General.hs
```diff
@@ -356,3 +359,83 @@ prop_general k TestConfig{numSlots, nodeJoinPlan, nodeTopology} mbSchedule
...
+        , (s, pbnos) <- Map.toList m
+        , (p, bno)   <- Set.toList pbnos
+        ]
+      where
+        -- INVARIANT: these AddBlock events are *not* for EBBs
+        adds = nodeOutputAdds <$> testOutputNodes
+
+        prop1 nid s p bno =
+            counterexample msg $
+            delayOK || noDelay
```
mrBliss: Would it be helpful to use `Property` (`===` and `.||.`) instead of `Bool` here, so the operands of the failing equality are printed?

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Types.hs
nfrisby: I pushed a commit with this message:

```
    consensus: add IsEBB flag type
    
    `IsEBB` is primarily intended for unnamed data constructor arguments. Functions
    and record fields with signatures matching `...IsEBB :: ... -> Bool` are clear
    enough as-is.
```

It adds a flag type `Ouroboros.Consensus.Block.IsEBB`.

ouroboros-consensus/test-consensus/Test/Dynamic/General.hs
nfrisby: I push a refactoring of the property: it cleaned up the point/slot contention and improved the counterexample message. Since one of the comparands is computed, I don't think `===` will be very helpful alone, so I'm just committing to the counterexample.

ouroboros-consensus/test-consensus/Test/Dynamic/General.hs
```diff
@@ -337,3 +360,91 @@ prop_general k TestConfig{numSlots, nodeJoinPlan, nodeTopology} schedule
...
+            -- NOTE This predicate is more general than that specific scenario,
+            -- but we don't anticipate it wholly masking any interesting cases.
+            delayOK1 = 1 == bno
+
+            -- When a slot has multiple leaders, each node chooses one of the
+            -- mutually-exclusive forged blocks and won't fetch any of the
+            -- others until it's later compelled to switch to a chain
+            -- containing one of them
+            --
+            -- TODO This predicate is more general than that specific scenario,
```
nfrisby: I'm tempted to depriortize this TODO, since it'll be superceded by the eventual PR that introduces network latency injections.

ouroboros-consensus/test-consensus/Test/Dynamic/Util.hs
```diff
@@ -233,6 +234,13 @@ consensusExpected k nodeJoinPlan schedule =
   where
     NumBlocks maxForkLength = determineForkLength k nodeJoinPlan schedule
 
+emptyLeaderSchedule :: NumSlots -> LeaderSchedule
+emptyLeaderSchedule (NumSlots t) = LeaderSchedule $
```
mrBliss: Is there an invariant somewhere that this satisfies (which `LeaderSchedule Map.empty` did not)? If so, then it's worth documenting it.

ouroboros-consensus/test-consensus/Test/Dynamic/General.hs
mrBliss: BTW, I just reread this property and I find it more readable :+1:

ouroboros-consensus/test-consensus/Test/Dynamic/Util.hs
nfrisby: I haven't seen such an invariant listed, no. The code seems to be in limbo about it: the invariant is apparently respected everywhere but it is not expected/required anywhere. The `Map.empty` I had here for a moment was the first time I've seen "missing" slots in a leader schedule, so I replaced it. It seemed reasonable to not spoil an apparent invariant without cause.

What do you think - go ahead an introduce the invariant and adjust the type's interface (in a separate PR)?

ouroboros-consensus/test-consensus/Test/Dynamic/General.hs
nfrisby: I amended the comment to mention the latency-inducing issues that we expect will handle it.

nfrisby:
I rebased onto master and also to eliminate `fixup` commits.

nfrisby:
(I belatedly promoted the PR from Draft.)

ouroboros-consensus/test-consensus/Test/Dynamic/Util.hs
mrBliss: I trust your judgment :slightly_smiling_face:

PR #1115, opened 2019-10-03
Improve BF tracing

nfrisby:
I made these changes in order to debug the error that you sent me today. It seems worthwhile to keep them.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/BlockFetchServer.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/BlockFetchServer.hs
@@ -16,9 +16,9 @@ module Ouroboros.Consensus.BlockFetchServer
 import           Data.Typeable (Typeable)
 
 import           Control.Monad.Class.MonadThrow
-import           Control.Tracer (Tracer)
+import           Control.Tracer (Tracer, traceWith)
 
-import           Ouroboros.Network.Block (HeaderHash, StandardHash)
+import           Ouroboros.Network.Block (HeaderHash, Point, StandardHash)
 import           Ouroboros.Network.Protocol.BlockFetch.Server
                      (BlockFetchBlockSender (..), BlockFetchSendBlocks (..),
                      BlockFetchServer (..))
@@ -63,7 +63,7 @@ blockFetchServer
     -> ChainDB m blk
     -> ResourceRegistry m
     -> BlockFetchServer blk m ()
-blockFetchServer _tracer chainDB registry = senderSide
+blockFetchServer tracer chainDB registry = senderSide
   where
     senderSide :: BlockFetchServer blk m ()
     senderSide = BlockFetchServer receiveReq ()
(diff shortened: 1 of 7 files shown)


ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -719,12 +730,6 @@ equatingEither eqA _ (Left  x) (Left  y) = eqA x y
 equatingEither _ eqB (Right x) (Right y) = eqB x y
 equatingEither _ _   (Right _) (Left  _) = False
 
-comparingFst :: Comparing a -> Comparing (a, b)
-comparingFst cmp = cmp `on` fst
-
-equatingFst :: Equating a -> Equating (a, b)
-equatingFst eq = eq `on` fst
-
```
nfrisby: The removal of `comparingFst` and `equatingFst` and the cascade of changes seems a bit heavy-handed, I admit. I'm open to suggestions -- I suspect the related function's implementations could benefit from some newtypes.

ouroboros-consensus/src/Ouroboros/Consensus/BlockFetchServer.hs
```diff
@@ -76,14 +76,16 @@ blockFetchServer _tracer chainDB registry = senderSide
...
         -- blocks back.
-        Left  _  -> SendMsgNoBlocks $ return senderSide
+        Left  _  -> do
+          traceWith tracer (TraceNoBlocks start end)
+          pure $ SendMsgNoBlocks $ return senderSide
         -- When we got an iterator, it will stream at least one block since
         -- its bounds are inclusive, so we don't have to check whether the
         -- iterator is empty.
-        Right it -> SendMsgStartBatch $ sendBlocks it
+        Right it -> pure $ SendMsgStartBatch $ sendBlocks it
```
mrBliss: Nitpick: we're using `return` in this module (and in fact, in most of the consensus codebase) instead of `pure`. Just for consistency, stick with `return`, certainly on line 84 which contains both `pure` and `return` :slightly_smiling_face:

ouroboros-consensus/src/Ouroboros/Consensus/BlockFetchServer.hs
```diff
@@ -106,6 +108,9 @@ blockFetchServer _tracer chainDB registry = senderSide
 
 -- | Events traced by the Block Fetch Server.
 data TraceBlockFetchServerEvent blk
-   -- TODO no events yet. Tracing the messages send/received over the network
-   -- might be all we need?
+  = TraceNoBlocks (Point blk) (Point blk)
```
mrBliss: While you're at it, it would be useful to include a message for starting a batch (and completing one?)

ouroboros-consensus/src/Ouroboros/Consensus/BlockFetchServer.hs
nfrisby: Done

ouroboros-consensus/src/Ouroboros/Consensus/BlockFetchServer.hs
nfrisby: Done

nfrisby:
> , just some minor things.
> Regarding the changes to the BlockFetch logic: maybe @dcoutts can chime in, he wrote the code in the first place.

@dcoutts What do you think? Is it too out-of-hand now? Maybe I should add some newtypes? The delta might look smaller if I try to preserve indentation. Please advise.

nfrisby:
I'm calling this stale.

PR #1117, opened 2019-10-07
Store snapshots of the PBftLedgerView instead of the whole ledger

mrBliss:
The Byron ledger state contains the complete ledger state, i.e.
`CC.Block.ChainValidationState`, but also snapshots of the previous states.
These are used to obtain a `PBftLedgerView` for slots in the past. This meant
that we were keeping track of a bunch of complete ledger states, which are
potentially huge. Fortunately, because we're using immutable data structures
that share the majority of their contents, the overhead was not too great.

Instead of storing whole ledger states, we now only store `PBftLedgerView`s,
which is all we need from these snapshots. In a small test, running the
cardano-byron-proxy for 5 minutes, memory usage shrunk from ~85 MB to ~81 MB.

Important: computing the `PBftLedgerView` given a ledger state should be
strict, otherwise we're still holding onto multiple ledger states.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
@@ -242,9 +242,9 @@ instance (ByronGiven, Typeable cfg, ConfigContainsGenesis cfg, NoUnexpectedThunk
   data LedgerState (ByronBlock cfg) = ByronLedgerState
       { blsCurrent :: !CC.Block.ChainValidationState
         -- | Slot-bounded snapshots of the chain state
-      , blsSnapshots :: !(Seq.StrictSeq (SlotBounded CC.Block.ChainValidationState))
+      , blsSnapshots :: !(Seq.StrictSeq (SlotBounded (PBftLedgerView PBftCardanoCrypto)))
       }
-    deriving (Eq, Show, Generic, NoUnexpectedThunks)
+    deriving (Eq, Show, Generic)
 
   type LedgerError (ByronBlock cfg) = CC.Block.ChainValidationError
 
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -105,15 +106,21 @@ forgePBftFields PBftIsLeader{..} toSign = do
...
 deriving instance PBftCrypto c => NoUnexpectedThunks (PBftLedgerView c)
   -- use generic instance
 
+deriving instance Eq (PBftVerKeyHash c) => Eq (PBftLedgerView c)
 deriving instance Show (PBftVerKeyHash c) => Show (PBftLedgerView c)
 
+instance (Serialise (PBftVerKeyHash c), Ord (PBftVerKeyHash c))
+      => Serialise (PBftLedgerView c) where
+  encode (PBftLedgerView ds) = encode (Bimap.toList ds)
+  decode = PBftLedgerView . Bimap.fromList <$> decode
```
mrBliss: @edsko Should the keys and values of the `Bimap` be forced here?

Note that when deserialising the `PBftLedgerView`s we don't get any sharing between the snapshots.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
```diff
@@ -300,6 +300,16 @@ instance (ByronGiven, Typeable cfg, ConfigContainsGenesis cfg, NoUnexpectedThunk
         where
           slot = convertSlot (CC.Block.cvsLastSlot state)
 
+deriving instance ByronGiven => NoUnexpectedThunks (LedgerState (ByronBlock cfg))
```
nc6: Do you want `instance` rather than `deriving instance` here, if the intention is to use the generic instance? In this case I think it will do the same thing, but in general the pattern elsewhere is to use an empty instance directly where we want to use generics

mrBliss: You're right, will fix.

PR #1118, opened 2019-10-07
Chain Sync Client: reorganise exception type

mrBliss:
* Introduce `ChainSyncClientExceptionType` for the different causes.

* Let `ChainSyncClientException` wrap `ChainSyncClientExceptionType` + our tip
  + their tip, this solves the duplication of having these two tips as fields
  for each constructor.

* No more partial fields.

* TODO what to use for `ourTip` and `theirTip` in `rejectInvalidBlocks`? We
  don't have access to the state in this function.


mrBliss:
Superseded by #1722.

Issue #1006, opened 2019-09-10
Network Magic should be configurable


coot:
There are two places where network magic is set to `0` in `consensus`:
https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L306
https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L317

It should be configurable by the caller (e.g. `cardano-node`). If there is a mismatch, the network layer will terminate the connection right after the connection started.  It's meant to distinguish testnet vs mainnet (or any other -net).

erikd:
> If there is a mismatch, the network layer will terminate the connection right after the connection started.

And provide a `logError` level log message.

mrBliss:
> There are two places where network magic is set to `0` in `consensus`:
> https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L306
> https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L317

Hint: in the future, use *permalinks* when referring to lines of code, because when the files inevitably change, the lines move around. (In this case, it's easy to relocate the lines.)

mrBliss:
Addressed by #1121.

PR #1121, opened 2019-10-08
Make NetworkMagic configurable

mrBliss:
Addresses #1006.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -247,34 +247,38 @@ type NetworkApps peer =
 
 -- | Arguments specific to the network stack
 data RunNetworkArgs peer blk = RunNetworkArgs
-  { rnaIpSubscriptionTracer  :: Tracer IO (WithIPList (SubscriptionTrace Socket.SockAddr))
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -247,34 +247,38 @@ type NetworkApps peer =
...
-  , rnaIpProducers           :: [SockAddr]
+  , rnaIpProducers             :: [SockAddr]
     -- ^ IP producers
-  , rnaDnsProducers          :: [DnsSubscriptionTarget]
+  , rnaDnsProducers            :: [DnsSubscriptionTarget]
     -- ^ DNS producers
+  , rnaNodeToNodeVersionData   :: NodeToNodeVersionData
+    -- ^ Node-to-Node Version Data
+  , rnaNodeToClientVersionData :: NodeToClientVersionData
+    -- ^ Node-to-CLient Version Data
```
coot: Are `rnaNodeTo{Node,Client}VersionData` used? The patch is smaller than I expect.

mrBliss: Oops :man_facepalming:. I force-pushed the remainder of the patch.

mrBliss: > Nitpick: should probably be `Node-to-Client` rather than `Node-to-CLient` smile

Fixed, thanks!

Issue #445, opened 2019-04-15
Chain DB: Protect against unintended file deletion


edsko:
All of the ImmDB, VolDB and LedgerDB will delete files that they try but fail to open. If the user somehow points the node to the wrong directory, and it happens to have a `volatile`, `immutable` or `ledger` subdirectory, we probably don't want to randomly start deleting files ("Cardano deleted all my files!" :-o ). It would be good to somehow protect against this (though not entirely sure how).

mrBliss:
Notes:

* Look at what other DBs do

* We can add a special file with a magic number (network magic, a separate number for different test/mainnets) in the directory. If this special file is corrupt, we can't recover: this will result in a hard failure on start-up. We should do that for each of the three DBs. When that file is created the first time, the directory should be empty.

* Make sure the naming scheme of the files of the different DBs is specific: e.g., `(epoch|index)-%03d.dat` for the Immutable DB. The other two DBs currently use plain numbers for the files,  this should be fixed.

mrBliss:
Feature added in #1123 and enabled in #1244.

PR #1123, opened 2019-10-09
Protect against unintended file deletion

mrBliss:
Addresses #445. The ChainDB deletes invalid files on startup. When
accidentally pointing it to the database directory of another net (testnet vs.
mainnet), it would wipe out that directory, throwing away all blocks. To
protect the user against such accidents, we store the `ProtocolMagicId` in a
file in the database folder and check that upon startup.

See the docstring of `checkProtocolMagicId` for more details.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications    #-}
@@ -23,6 +24,9 @@ module Ouroboros.Consensus.Node
   , NodeArgs (..)
   , NodeKernel (..)
     -- * Internal helpers
+  , checkProtocolMagicId
+  , protocolMagicIdFile
+  , ChainDbCheckError (..)
   , initChainDB
   , mkNodeArgs
   , initNetwork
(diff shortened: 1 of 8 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node/Run/Mock.hs
```diff
@@ -44,6 +46,8 @@ instance ( ProtocolLedgerView (SimpleBlock SimpleMockCrypto ext)
       --  This doesn't matter much
       dummyDate = UTCTime (fromGregorian 2019 8 13) 0
 
+  nodeProtocolMagicId    = \_ _ -> ProtocolMagicId 0
```
mrBliss: Should we provide a different value for the different protocols? We could introduce a type class for this that has an instance for the `SimpleBlock` instantiated to each protocol.

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
```diff
@@ -0,0 +1,132 @@
...
+
+import           Test.Util.FS.Sim.FsTree (FsTree (..))
+import           Test.Util.FS.Sim.MockFS (Files)
+import qualified Test.Util.FS.Sim.MockFS as Mock
+import           Test.Util.FS.Sim.STM (runSimFS)
+
+tests :: TestTree
+tests = testGroup "Node"
+    [ testGroup "checkProtocolMagicId"
+      [ testCase "match"        test_checkProtocolMagicId_match
```
mrBliss: @edsko These are ... **unit** tests. *Ducks*

mrBliss:
Merging this in `cardano-node` will break things for people, because existing database folder don't have the `ProtocolMagicId` in them.

Maybe we should accept folders containing just `immutable`, `ledger`, and `volatile`, but that have no `ProtocolMagicId`? Or just as a temporary measure? Or `cardano-node` could temporarily add some code that does this and remove it after a while?

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -143,6 +165,115 @@ run tracers chainDbTracer rna dbPath pInfo
...
+        res <- CBOR.deserialiseFromBytes fromCBOR <$> hGetAll hasFS h
+        case res of
+          Right (bl, a) | Lazy.null bl -> return $ Right a
+          _ -> return $ Left $ CorruptProtocolMagicId fullPath
+
+    createProtocolMagicIdFile :: ExceptT ChainDbCheckError m ()
+    createProtocolMagicIdFile = lift $
+      withFile hasFS pFile (AppendMode MustBeNew) $ \h ->
+        void $ hPutAll hasFS h $
+          CBOR.toLazyByteString (toCBOR protocolMagicId)
```
mrBliss: Alternatively, we use `show` + `readMaybe` on the underlying `Word32` to get a readable value, which would make debugging easier.

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
edsko: 😮😮🦆🦆

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
mrBliss: Yeah, let's do this. There is precedent for printing `ProtocolMagic` as a decimal number.

edsko:
Will try to rebase this and address the feedback.

PR #1128, opened 2019-10-11
ouroboros-consensus/docs: draft test-consensus-plan.md

nfrisby:
After studying relevant parts of the Ouroboros paper and considering possible test suite designs that would include injecting network latencies, I've drafted this document, which arguably contains at least the beginning of a plan :). The current content is dominated by discussions of the intentional departures from the paper's spec and how that makes it difficult to test the paper's properties when injecting network latencies.

This is a rather dense document, so I think we'll need some synchronous communication. For that purpose, would you skim it until you have a sense of which parts you'd like us to talk about first, and then I'll organize a call with a few slides for that topic? Thanks.

Ideally, I've failed to notice some solutions to the problems listed here and we'll find them together, simplifying the challenges as currently laid out. Otherwise, I'm hoping for feedback on the document's overall organization, its explanations (high- and low-level), reprioritization of its content, reprioritization of the listed tasks, etc. Thanks again.

I wrote this document primarily to organize and vet my thoughts on these issues before discussing them with the consensus team. It'll also serve as a catalyst for us to cooperatively develop a plan. And I think it does eventually belong in the repo to archive our test suite design motivations/decisions.

### DIFF
--- a/ouroboros-consensus/docs/test-consensus/Plan.md
+++ b/ouroboros-consensus/docs/test-consensus/Plan.md
@@ -1,0 +1,629 @@
+# Consensus Testing Plan
+
+This document discusses plans for and obstacles to improving the
+`ouroboros-consensus` integration testing, the `test-consensus` test suite. See
+`ouroboros-consensus/test-consensus/README.md` for a less technical overview.
+
+The core plan is to test both low-level concerns such as closing file handles
+promptly and also the high-level behavior of nodes. The high-level behavior
+consist of both the collective behavior of a node network as well as the
+behavior of each individual node therein. For example, individual nodes should
+never rollback their chain by more than `k` blocks, and groups of nodes should
+reach consensus whenever network latencies, disk failures, adversarial
+behavior, etc allow them to do so.
+
+This current draft focuses on the difficulties involved in determining whether
+a concrete set of specific network latencies allow for consensus. We discuss
+the [specification][sec:specification] of a node, its interface's messages, and
+the interconnect between nodes; then the intentional
+[discrepancies][sec:discrepancies] between that specification and the
+implementation; and finally the [design][sec:design] of the current test suite
+and its planned improvements.
+
+## Specification
+[sec:specification]: #specification
+
+The most obvious specification against which to test the implementation is the
+paper that defines Ouroboros Praos
+<https://eprint.iacr.org/2017/573/20171115:001835>.
+
+  * Figure 4 defines the π_SPoS protocol by specifying the behavior of an honest
+    party: any node that follows those steps and is not offline too long or too
+    often qualifies as an _honest party_.
+
+  * Section 2.1 defines three properties that the authors' prior work
+    established as sufficient conditions for ensuring correct behavior of a
+    node network: Common Prefix, Chain Growth, and Chain Quality.
+
+This section discusses these definitions, and the next section demonstrates
+that they are difficult to use as a specification.
+
+### π_SPoS Protocol
+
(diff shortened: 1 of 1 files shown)


nfrisby:
@mrBliss @edsko @dcoutts @nc6 I hope to get feedback from several of you, but I'm guessing it makes the most sense to start with just one, since the first is doomed to find my "rubber ducky"-level mistakes :(

Would anyone like to go first? If you do the skimming I suggested in the PR description, we can start from there. Thanks!

ouroboros-consensus/docs/test-consensus/Plan.md
```diff
@@ -0,0 +1,629 @@
...
+that is analogous to the delay of a paper message. The test suite therefore
+considers a _message_ to be a pair of a block and a node and such a message's
+_delay_ to be the difference between the slot in which the block was forged and
+the slot in which it was added to the node's chain database. This definition
+preserves the property from the paper that a message delay is the duration (in
+slots) of the communication of a forged chain (headed by the new block) to
+another node in the network.
+
+(Remark. It is possible for different nodes to forge the same EBB in either the
+same or different slots. This particular definition of message delay therefore
```
mrBliss: I thought an EBB for a certain epoch only fits into a certain slot, i.e. the first slot of the next epoch (this slot can *also* contain a regular block). I might be mistaken, but I was under the impression that each (core) node produces every EBB "automically" at the end of each epoch.

ouroboros-consensus/docs/test-consensus/Plan.md
```diff
@@ -0,0 +1,629 @@
...
+_strictly increasing labels_ invariant, and the π_SPoS protocol does not seem
+to guard against that in any way other than by the assumption that the
+recipient will have checked its inbox before the requisite message arrives. In
+other words, the paper assumes that parties are perfectly synchronized (no
+clock skew) and execute the π_SPoS protocol in total isolation _during_ each
+slot, while messages are instead only delivered _between_ slots.
+
+(Remark. With computational delays and/or clock skews, the current
+implementation's node could forge a block with the same slot as its
+predecessor. However, the current test suite has no clock skews etc, and the
```
mrBliss: We don't: https://github.com/input-output-hk/ouroboros-network/blob/9d825e8f541934aa0efffa5039620202c8a4ab0e/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs#L337

ouroboros-consensus/docs/test-consensus/Plan.md
```diff
@@ -0,0 +1,629 @@
...
+In this setup, we are able to additionally check that no final forks are deeper
+than expected, even if they are shallower than k blocks. We compute the
+expected maximum fork depth by processing the leader schedule and node join
+plan as a sequence of commands for an abstract virtual machine.
+
+> Q1 How to determine whether a non-zero _message delay_ should be expected?
+
+All non-zero message delays should be due to late joins and/or the leader
+schedule (via just-in-time downloads), by The Quiescence Assumption. Whenever a
+node add receives a block (i.e. adds it to its ChainDB), we compare the current
```
mrBliss:
```suggestion
node receives a block (i.e. adds it to its ChainDB), we compare the current
```

ouroboros-consensus/docs/test-consensus/Plan.md
```diff
@@ -0,0 +1,629 @@
...
+slot to the maximum of the block's record slot and the slot in which the node
```
mrBliss: `record`?

nfrisby:
Bit- and priority-rot -- I'll come back to this eventually.

Issue #1090, opened 2019-10-01
Testing: test CBOR validity/roundtrip for all types


edsko:
We should have tests that check our CBOR encoders and decoders.

mrBliss:
Addressed in #1129

PR #1129, opened 2019-10-14
Add Byron serialisation roundtrip property tests

mrBliss:
Addresses #1090.

This is a sanity check that we can deserialise the serialised blocks/headers/txs/etc. We use the generators from cardano-ledger for this.


ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
```diff
@@ -0,0 +1,186 @@
...
+import           Test.Tasty.QuickCheck
+
+import qualified Test.Cardano.Chain.Block.Gen as CC.Gen
+import qualified Test.Cardano.Chain.UTxO.Gen as CC.Gen
+
+
+tests :: TestTree
+tests = testGroup "Byron"
+  [ testGroup "Serialisation roundtrips"
+      [ testProperty "roundtrip Block"       prop_roundtrip_Block
```
edsko: Let's make sure that this is exhaustive, and if not, mark the remaining elements as TODO. Might be useful to add a comment to the Byron to say if we add more encoders/decoders, they should have corresponding tests here.

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
```diff
@@ -0,0 +1,186 @@
...
+-------------------------------------------------------------------------------}
+
+roundtrip :: (Eq a, Show a)
+          => (a -> Encoding)
+          -> (forall s. Decoder s a)
+          -> a
+          -> Property
+roundtrip enc dec = roundtrip' enc (const <$> dec)
+
+roundtrip' :: (Eq a, Show a)
```
edsko: Might be useful to mention that if we were to apply this to a transaction with a non-canonical annotation, this test will fail.

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
```diff
@@ -0,0 +1,186 @@
...
+          CC.Block.fromCBORABoundaryHeader <$>
+        hedgehog CC.Gen.genBoundaryHeader
+
+instance Arbitrary ByronHash where
+  arbitrary = ByronHash <$> hedgehog CC.Gen.genHeaderHash
+
+instance Arbitrary (GenTx Block) where
+  arbitrary =
+    mkByronTx . annotateTxAux <$>
+    hedgehog (CC.Gen.genTxAux protocolMagicId)
```
edsko: Might need a more general generator here.

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
```diff
@@ -0,0 +1,186 @@
...
+
+instance Arbitrary (GenTxId Block) where
+  arbitrary = ByronTxId <$> hedgehog CC.Gen.genTxId
```
edsko: Same.

PR #1130, opened 2019-10-14
Karknu/coot/error handling rb

coot:


ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -287,24 +292,29 @@ ipSubscriptionWorker
           peerid
           NodeToNodeProtocols
           IO BL.ByteString x y)
-    -> IO void
+    -> IO ()
```
coot: is removing of `void` here is necessary? or just a consequence of using `void` below

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -287,24 +292,29 @@ ipSubscriptionWorker
...
+  errTracer
   peeridFn
   tbl
+  peerStatesVar
   localIPv4 localIPv6
   connectionAttemptDelay
   ips
   versions
-    = Subscription.ipSubscriptionWorker
+    = void $ Subscription.ipSubscriptionWorker
```
coot: it might be just because of this `void`

ouroboros-network/test/Test/PeerState.hs
```diff
@@ -0,0 +1,476 @@
...
+data TestException2 = TestException2
+  deriving Show
+
+instance Exception TestException2
+
+instance Arbitrary ArbException where
+    arbitrary = oneof
+      [ pure (ArbException TestException1)
+      , pure (ArbException TestException2)
+      -- AsyncException
```
coot: it would be good to uncomment these generators

ouroboros-network/src/Ouroboros/Network/Subscription/Ip.hs
```diff
@@ -37,62 +59,308 @@ data IPSubscriptionTarget = IPSubscriptionTarget {
...
     -- ^ Lookup function, should return expected delay for the given address
+    -> [ErrorPolicy]
+    -> (Time IO -> Socket.SockAddr -> a -> SuspendDecision DiffTime)
     -> IPSubscriptionTarget
-    -> Main IO () x
     -> (Socket.Socket -> IO a)
-    -> IO x
-ipSubscriptionWorker tracer tbl localIPv4 localIPv6 connectionAttemptDelay ips main k = do
-    sVar <- newTVarM ()
+    -> IO Void
```
coot: maybe it can also return `forall void. ... -> IO void`, instead of `Void`.

PR #1133, opened 2019-10-14
Use void in the return type of clients that never finish

mrBliss:


mrBliss:
The clients don't yet have `void` or `Void` as their return type in `NodeNetwork` because changing that file will probably conflict with #1127. I leave the rest up to you @coot & @karknu.

PR #1135, opened 2019-10-14
prevent stale TracingConstraints

nfrisby:
Basic idea: never use `nullTracer` in this module without ``_ `asTypeOf`
showTracing debugTracer``. Similarly for `nullTracers` and
`nullProtocolTracers`.

Without that rule, the `TracingConstraints` synonym had gone stale and almost
certainly would again.


edsko:
Let's leave it in. It's useful.

Issue #1119, opened 2019-10-08
Volatile DB: cleanup


mrBliss:
The VolatileDB needs some cleaning up:
* As noted in #1110, there is not enough abstraction
* Newtypes should be used instead of type aliases
* Data types and fields should be documented
* Function/variable names are too terse and cryptic
* Haddock docstring should be used instead of plain comments
* The code style does not match the rest of the codebase:
  - stylish-haskell should be used
  - Lines should be limited to 80 characters, e.g., `mkInternalState`
  - `let` is used  where `where` would be more readable
* Repeated code patterns should be abstracted into helpers, e.g. `getIsMemberImpl`, `getSuccessorsImpl`, `getPredecessorImpl`, and `getMaxSlotNoImpl` all have the same boilerplate.
* Invariants should be cleared documented (and asserted when applicable)
* Complex code should be split into smaller pieces, e.g., `mkInternalState`.
* The same applies to the q-s-m tests
* The simulated corruptions/errors in the q-s-m tests are too dependent on the implementation and will break whenever the implementation changes (this has blocked refactorings in the past).

**EDIT:** the tests still need to be cleaned up, in particular, they should switch to `Test.Storage.TestBlock`.

mrBliss:
The implementation got a cleaning in #1137, but the tests still need a clean-up.

mrBliss:
About not hard-coding what should happen in case of simulated errors in the q-s-m tests:
* Before you run such a command, make copies of the VolatileDB files
* Run the command
  a. If it fails because of a simulated error, close the VolatileDB, restore the copies of the files, restart in the previous state.
  b. If it doesn't fail, do the same as in a. anyway, because the outcome needs to be deterministic.

In the model, executing a command with a simulated error is a no-op.

Issue #767, opened 2019-07-15
Chain DB: audit for repeatedly opening the same file for reading a block


mrBliss:
While we typically use ImmutableDB iterators to efficiently stream blocks in the ChainDB, sometimes we read blocks one-by-one, in particular when we're reading from the VolatileDB, as it doesn't provide iterator-like functionality. For example, when copying blocks from the VolatileDB to the ImmutableDB, the blocks from the VolatileDB are read one-by-one, which requires opening, seeking, reading, and closing a file for each block.

* Change interface to allow keeping file handles open (VolatileDB + ChainDB)
* Discourage single-shot reads by renaming the function, e.g., `s/getBlock/getBlockInefficiently/`
* Provide the equivalent of `pread` in the FS layer?

mrBliss:
Inspiration: https://github.com/input-output-hk/cardano-sl/commit/3d6b88cd20affff2122086c501dac4dfba0b0a97#diff-65dab1824c1d453ffa204c3b5957bd70R115

mrBliss:
Idea: use a `ResourceRegistry` for open handles?

kderme:
I think using `ResourceRegistry` is a good idea here. Internally Volatiledb will keep track of a cache of open handles and close them with a certain policy, like LRU. The inefficient version of get, will open a new handle, without looking at the cache. Chain db will have to be adapted to use a bracket-style to open the db and will use the modified version of get (which will however look pretty much like the old one on the outside). Do these sound right? @mrBliss

mrBliss:
> I think using `ResourceRegistry` is a good idea here. Internally Volatiledb will keep track of a cache of open handles and close them with a certain policy, like LRU. The inefficient version of get, will open a new handle, without looking at the cache. Chain db will have to be adapted to use a bracket-style to open the Volatiledb and will use the modified version of get (which will however look pretty much like the old one on the outside). Do these sound right? @mrBliss

Reading from the VolatileDB is done by the ChainDB in the following scenarios:
1. When starting up, we construct all possible candidate chains in the VolatileDB. We first construct chains of hashes using `getSuccessors`, without doing any IO. The next step is reading all the headers from the VolatileDB that correspond to these hashes. When there are no forks, those blocks can be read pretty much sequentially. It's at least wasteful to reopen a file for each block (typically `k = 2160` blocks).
[...]

PR #1137, opened 2019-10-15
Cleanup volatile db

kderme:
Addresses most parts of https://github.com/input-output-hk/ouroboros-network/issues/1119

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
@@ -114,8 +114,7 @@ import           Ouroboros.Storage.Util.ErrorHandling (ErrorHandling (..),
                      ThrowCantCatch (..))
 import qualified Ouroboros.Storage.Util.ErrorHandling as EH
 import           Ouroboros.Storage.VolatileDB.API
-import           Ouroboros.Storage.VolatileDB.FileInfo (FileInfo,
-                     FileSlotInfo (..))
+import           Ouroboros.Storage.VolatileDB.FileInfo (FileInfo)
 import qualified Ouroboros.Storage.VolatileDB.FileInfo as FileInfo
 import           Ouroboros.Storage.VolatileDB.Index (Index)
 import qualified Ouroboros.Storage.VolatileDB.Index as Index
@@ -306,7 +305,7 @@ putBlockImpl env@VolatileDBEnv{..} BlockInfo{..} builder =
                     ++ "Current write file not found in Index.")
             (Index.lookup _currentWriteId _currentMap)
         fileInfo' = FileInfo.addSlot bslot _currentWriteOffset
-            (FileSlotInfo (BlockSize bytesWritten) bbid) fileInfo
+            (FileInfo.mkFileSlotInfo (BlockSize bytesWritten) bbid) fileInfo
         currentMap' = Index.insert _currentWriteId fileInfo' _currentMap
         internalBlockInfo' = InternalBlockInfo {
             ibFile       = _currentWritePath
--- a/ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
@@ -88,7 +88,6 @@ module Ouroboros.Storage.VolatileDB.Impl
     , openDBFull
     ) where
 
-import           Control.Arrow (first)
 import           Control.Monad
 import qualified Data.ByteString.Builder as BS
 import           Data.ByteString.Lazy (ByteString)
@@ -118,7 +117,8 @@ import           Ouroboros.Storage.VolatileDB.API
 import           Ouroboros.Storage.VolatileDB.FileInfo (FileInfo,
                      FileSlotInfo (..))
 import qualified Ouroboros.Storage.VolatileDB.FileInfo as FileInfo
-import           Ouroboros.Storage.VolatileDB.Index
+import           Ouroboros.Storage.VolatileDB.Index (Index)
+import qualified Ouroboros.Storage.VolatileDB.Index as Index
 import           Ouroboros.Storage.VolatileDB.Util
 
 {------------------------------------------------------------------------------
@@ -304,10 +304,10 @@ putBlockImpl env@VolatileDBEnv{..} BlockInfo{..} builder =
         fileInfo = fromMaybe
             (error $ "VolatileDB invariant violation:"
                     ++ "Current write file not found in Index.")
-            (Map.lookup _currentWritePath _currentMap)
+            (Index.lookup _currentWriteId _currentMap)
         fileInfo' = FileInfo.addSlot bslot _currentWriteOffset
             (FileSlotInfo (BlockSize bytesWritten) bbid) fileInfo
-        currentMap' = Map.insert _currentWritePath fileInfo' _currentMap
+        currentMap' = Index.insert _currentWriteId fileInfo' _currentMap
         internalBlockInfo' = InternalBlockInfo {
             ibFile       = _currentWritePath
           , ibSlotOffset = _currentWriteOffset
@@ -341,7 +341,7 @@ garbageCollectImpl :: forall m blockId. (IOLike m, Ord blockId)
 garbageCollectImpl env@VolatileDBEnv{..} slot =
     modifyState env $ \hasFS st -> do
       st' <- foldM (tryCollectFile hasFS env slot) st
-        (sortOn fst $ first unsafeParseFd <$> Map.toList (_currentMap st))
+              (sortOn fst $ Index.toList (_currentMap st))
       return (st', ())
 
 -- | For the given file, we check if it should be garbage collected and
@@ -363,9 +363,9 @@ tryCollectFile :: forall m h blockId
 tryCollectFile hasFS@HasFS{..} env slot st@InternalState{..} (fileId, fileInfo) =
     if  | not canGC     -> return st
         | not isCurrent -> do
-            removeFile file
+            removeFile $ filePath fileId
             return st {
-                _currentMap     = Map.delete file _currentMap
+                _currentMap     = Index.delete fileId _currentMap
               , _currentRevMap  = currentRevMap'
               , _currentSuccMap = succMap'
               }
@@ -385,7 +385,6 @@ tryCollectFile hasFS@HasFS{..} env slot st@InternalState{..} (fileId, fileInfo) 
               , _currentSuccMap = succMap'
               }
   where
-    file           = filePath fileId
     canGC          = FileInfo.canGC fileInfo slot
     isCurrent      = fileId == _currentWriteId
     isCurrentNew   = _currentWriteOffset == 0
@@ -444,7 +443,7 @@ nextFile HasFS{..} _err VolatileDBEnv{..} st@InternalState{..} = do
       , _currentWritePath   = file
       , _currentWriteId     = _nextNewFileId
       , _currentWriteOffset = 0
-      , _currentMap         = Map.insert file FileInfo.empty
+      , _currentMap         = Index.insert _nextNewFileId FileInfo.empty
                                 _currentMap
       , _nextNewFileId      = _nextNewFileId + 1
       }
@@ -465,8 +464,8 @@ reOpenFile HasFS{..} _err VolatileDBEnv{..} st@InternalState{..} = do
     -- However the file is open on Append Only, so it should automatically go
     -- to the end before each write.
    hTruncate _currentWriteHandle 0
-   return $ st {
-        _currentMap = Map.insert _currentWritePath FileInfo.empty _currentMap
+   return st {
+        _currentMap = Index.insert _currentWriteId FileInfo.empty _currentMap
       , _currentWriteOffset = 0
       }
 
@@ -501,14 +500,14 @@ mkInternalState
   -> m (InternalState blockId h)
 mkInternalState hasFS err parser n files =
     wrapFsError (hasFsErr hasFS) err $
-      go Map.empty Map.empty Map.empty Nothing [] files
+      go Index.empty Map.empty Map.empty Nothing [] files
   where
     -- | This is the file with the maximum 'FileId' in the db.
     lastFile = safeMaximumOn fst files
-    newFileInfo mp newIndex =
+    newFileInfo curMap newIndex =
       ( filePath newIndex
       , newIndex
-      , Map.insert (filePath newIndex) FileInfo.empty mp
+      , Index.insert newIndex FileInfo.empty curMap
       , FileSize 0 )
 
     truncateOnError Nothing _ _ = return ()
@@ -546,7 +545,7 @@ mkInternalState hasFS err parser n files =
           , _currentRevMap      = currentRevMap
           , _currentSuccMap     = succMap
           , _currentMaxSlotNo   = FileInfo.maxSlotInFiles
-                                    (Map.elems currentMap')
+                                    (Index.elems currentMap')
           }
       where
         (fileToWrite, fdToWrite, currentMap', offset') =
@@ -586,7 +585,7 @@ mkInternalState hasFS err parser n files =
                 slotOffset + unBlockSize blockSize
             fileMap = Map.fromList blocks
             (fileInfo, maxSlotOfFile) = FileInfo.fromParsedInfo blocks
-            newMap = Map.insert file fileInfo currentMap
+            newMap = Index.insert fd fileInfo currentMap
             newMaxSlot = maxSlotList $ catMaybes [maxSlot, maxSlotOfFile]
             -- For each block we need to update the succesor Map of its
             -- predecesor.
--- a/ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
@@ -1,18 +1,18 @@
-{-# LANGUAGE DeriveAnyClass             #-}
-{-# LANGUAGE DeriveGeneric              #-}
-{-# LANGUAGE ExistentialQuantification  #-}
-{-# LANGUAGE FlexibleContexts           #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE LambdaCase                 #-}
-{-# LANGUAGE MultiWayIf                 #-}
-{-# LANGUAGE OverloadedStrings          #-}
-{-# LANGUAGE RankNTypes                 #-}
-{-# LANGUAGE RecordWildCards            #-}
-{-# LANGUAGE ScopedTypeVariables        #-}
-{-# LANGUAGE TupleSections              #-}
--- Volatile on-disk database of binary blobs
+{-# LANGUAGE DeriveAnyClass            #-}
+{-# LANGUAGE DeriveGeneric             #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleContexts          #-}
+{-# LANGUAGE LambdaCase                #-}
+{-# LANGUAGE MultiWayIf                #-}
+{-# LANGUAGE OverloadedStrings         #-}
+{-# LANGUAGE QuantifiedConstraints     #-}
+{-# LANGUAGE RankNTypes                #-}
+{-# LANGUAGE RecordWildCards           #-}
+{-# LANGUAGE ScopedTypeVariables       #-}
+{-# LANGUAGE TupleSections             #-}
+-- | Volatile on-disk database of binary blobs
 --
--- Logic
+-- = Logic
 --
 -- The db is a key-value store of binary blocks and is parametric
 -- on the key of blocks, named blockId. The only constraints are that one must provide
@@ -32,7 +32,7 @@
 -- collection makes the deletion of blocks depend on the number of blocks we insert on
 -- each file, as well as the order of insertion, so it's not deterministic on blocks themselves.
 --
--- Errors
+-- = Errors
 --
 -- On any exception or error the db closes and its Internal State is lost, inluding in memory
 -- indexes. We try to make sure that even on errors the fs represantation of the db remains
@@ -51,12 +51,12 @@
 -- wrapped. All top-level function of this module are safe. You can safely use HasFs calls
 -- in modifyState or wrapFsError actions.
 --
--- Concurrency
+-- = Concurrency
 --
 -- The same db should only be opened once
 -- Multiple threads can share the same db as concurency if fully supported.
 --
--- FS Layout:
+-- = FS Layout:
 --
 -- On disk represantation is as follows:
 --
@@ -83,15 +83,16 @@ module Ouroboros.Storage.VolatileDB.Impl
       -- * tests only
     , VolatileDBEnv(..)
     , InternalState(..)
+    , OpenOrClosed(..)
     , filePath
-    , getInternalState
     , openDBFull
     ) where
 
+import           Control.Arrow (first)
 import           Control.Monad
 import qualified Data.ByteString.Builder as BS
 import           Data.ByteString.Lazy (ByteString)
-import           Data.List (sortOn)
+import           Data.List (find, sortOn)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe
@@ -105,7 +106,7 @@ import           Control.Monad.Class.MonadThrow
 
 import           Ouroboros.Network.Point (WithOrigin)
 
-import           Ouroboros.Consensus.Util (SomePair (..))
+import           Ouroboros.Consensus.Util (safeMaximumOn)
 import           Ouroboros.Consensus.Util.IOLike
 
 import           Ouroboros.Storage.FS.API
@@ -143,15 +144,24 @@ volatileDbIsOpen (VolatileDbOpen _) = True
 volatileDbIsOpen VolatileDbClosed   = False
 
 data InternalState blockId h = InternalState {
-      _currentWriteHandle :: !(Handle h) -- The unique open file we append blocks.
-    , _currentWritePath   :: !FsPath -- The path of the file above.
-    , _currentWriteOffset :: !Word64 -- The 'WriteHandle' for the same file.
-    , _nextWriteFiles     :: ![(FsPath, Word)] -- The path and the size of the next files to write
-    , _nextNewFileId      :: !Int -- The next file name Id.
-    , _currentMap         :: !(Index blockId) -- The content of each file.
-    , _currentRevMap      :: !(ReverseIndex blockId) -- Where to find each block from slot.
-    , _currentSuccMap     :: !(SuccessorsIndex blockId) -- successors for each block.
-    , _currentMaxSlotNo   :: !MaxSlotNo -- Highest ever stored SlotNo
+      _currentWriteHandle :: !(Handle h)
+      -- ^ The only open file we append blocks to.
+    , _currentWritePath   :: !FsPath
+      -- ^ The path of the file above.
+    , _currentWriteId     :: !FileId
+      -- ^ The 'FileId' of the same file.
+    , _currentWriteOffset :: !Word64
+      -- ^ The offset of the same file.
+    , _nextNewFileId      :: !Int
+      -- ^ The next file name Id.
+    , _currentMap         :: !(Index blockId)
+      -- ^ The contents of each file.
+    , _currentRevMap      :: !(ReverseIndex blockId)
+      -- ^ Where to find each block based on its slot number.
+    , _currentSuccMap     :: !(SuccessorsIndex blockId)
+      -- ^ The successors for each block.
+    , _currentMaxSlotNo   :: !MaxSlotNo
+      -- ^ Highest ever stored SlotNo.
     }
   deriving (Generic, NoUnexpectedThunks)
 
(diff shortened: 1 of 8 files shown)


ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -428,25 +441,16 @@ nextFile :: forall h m blockId. IOLike m
...
-            let file = filePath $ _nextNewFileId st
-            hndl <- hOpen file (AppendMode MustBeNew)
-            return $ st {
-                  _currentWriteHandle = hndl
-                , _currentWritePath   = file
-                , _currentWriteOffset = 0
-                , _nextNewFileId      = (_nextNewFileId st) + 1
-                , _currentMap         = Map.insert file FileInfo.empty (_currentMap st)
-            }
-        (file, size) : rest -> do
```
kderme: This was basically dead code, so the _nextWriteFiles field is totally erased. In the past, we supported a number of policies related to files which where not full: We could for example first try to fill these files and then proceed to creating new ones. The `_nextWriteFiles` field kept track of this. However we later deleted these policies and kept the default one (append to an existing non-empty file only if it is the latest in the db).

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -472,106 +476,143 @@ mkInternalStateDB :: (HasCallStack, MonadThrow m, MonadCatch m, Ord blockId)
...
+            [ "Volatile db invariant violation:"
+            , "A file was found with less than "
+            , show n
+            , " blocks, but there are no files parsed."]
+
+        truncateOnError Nothing _ _ = return ()
+        truncateOnError (Just _) file offset =
+            -- the handle of the parser is closed at this point. We need
+            -- to reOpen the file in AppendMode now (parser opens with
+            -- ReadMode).
```
mrBliss: Capitalise the sentence and make it clear what is code by using Haddock markup (`'..'` or `@..@` for non-linked things):
```haskell
-- The handle of the parser is closed at this point. We need
-- to reopen the file in 'AppendMode' now (parser opens with
-- 'ReadMode').

File: ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs [hunk 3]
```diff
@@ -472,106 +476,143 @@ mkInternalStateDB :: (HasCallStack, MonadThrow m, MonadCatch m, Ord blockId)
...
+            --
+            withFile hasFS file (AppendMode AllowExisting) $ \hndl ->
+                hTruncate hasFS hndl (fromIntegral offset)
+
         go :: Index blockId
            -> ReverseIndex blockId
            -> SuccessorsIndex blockId
            -> Maybe (blockId, SlotNo)
-           -> [(FileId, FsPath, FileSize)] -- The relative path and size of the files with less than n blocks, if any found already.
+           -> [(FileId, FsPath, FileSize)] -- Info of files with < n blocks.
```
mrBliss: Even though it is a local function, I would make it a Haddock docstring (`-- ^ ..`).

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -472,106 +476,143 @@ mkInternalStateDB :: (HasCallStack, MonadThrow m, MonadCatch m, Ord blockId)
...
+                  _currentWriteHandle = hndl
+                , _currentWritePath   = fileToWrite
+                , _currentWriteOffset = offset'
+                , _nextNewFileId      = nextNewFileId'
+                , _currentMap         = mp'
+                , _currentRevMap      = revMp
+                , _currentSuccMap     = succMp
+                , _currentMaxSlotNo   = FileInfo.maxSlotInFiles (Index.elems mp')
+            }
+            where
```
mrBliss: `where` indentation

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -472,106 +476,143 @@ mkInternalStateDB :: (HasCallStack, MonadThrow m, MonadCatch m, Ord blockId)
...
+                        (_, Just lst) ->
+                            -- If it's not the last file, we just ignore it and
+                            -- open a new one.
+                            newFileInfo mp $ lst + 1
+
+        updateAndGo mp revMp succMp maxSlot haveLessThanN file restFiles blocks mErr = do
+            truncateOnError mErr file offset
+            newRevMp <- fromEither err $ reverseMap file revMp fileMp
+            go newMp newRevMp newSuccMp newMaxSlot newHaveLessThanN restFiles
+            where
```
mrBliss: `where` indentation

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -472,106 +476,143 @@ mkInternalStateDB :: (HasCallStack, MonadThrow m, MonadCatch m, Ord blockId)
...
+            -- ReadMode).
+            --
+            -- Note that no file is open at this point, so we can safely
+            -- open with AppendMode any file, without the fear of opening
+            -- multiple concurrent writers, which is not allowed.
+            --
+            withFile hasFS file (AppendMode AllowExisting) $ \hndl ->
+                hTruncate hasFS hndl (fromIntegral offset)
+
         go :: Index blockId
```
mrBliss: This looks already better than before, but could still use some improvement. Can you start out with adding simple docstrings to `go` and `updateAndGo`?

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -472,106 +476,143 @@ mkInternalStateDB :: (HasCallStack, MonadThrow m, MonadCatch m, Ord blockId)
...
+                        (Just (fd, wrfile, size), Just lst) | fd == lst->
+                            -- if the file with biggest index is non-empty,
+                            -- then this is the file we're looking for.
+                            (wrfile, lst + 1, mp, size)
+                        (_, Just lst) ->
+                            -- If it's not the last file, we just ignore it and
+                            -- open a new one.
+                            newFileInfo mp $ lst + 1
+
+        updateAndGo mp revMp succMp maxSlot haveLessThanN file restFiles blocks mErr = do
```
mrBliss: With so many arguments, I think a type signature would help me understand it better.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -611,36 +652,51 @@ modifyState VolatileDBEnv{_dbHasFS = hasFS :: HasFS m h, ..} action = do
...
+       => VolatileDBEnv m blockId
+       -> (forall h. InternalState blockId h -> a)
+       -> m a
+getter VolatileDBEnv{..} fromSt = do
+    mSt <- readMVar _dbInternalState
+    case mSt of
+        VolatileDbClosed  -> EH.throwError _dbErr $ UserError ClosedDBError
+        VolatileDbOpen st -> return $ fromSt st
+
+getterStm :: forall m blockId a. IOLike m
```
mrBliss: Can you rename it to `getterSTM`? That fits better with our conventions.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -611,36 +652,51 @@ modifyState VolatileDBEnv{_dbHasFS = hasFS :: HasFS m h, ..} action = do
...
+          => VolatileDBEnv m blockId
+          -> (forall h. InternalState blockId h -> a)
+          -> STM m a
+getterStm VolatileDBEnv{..} fromSt = do
+    mSt <- readMVarSTM _dbInternalState
+    case mSt of
+        VolatileDbClosed  -> EH.throwError' _dbErrSTM $ UserError ClosedDBError
+        VolatileDbOpen st -> return $ fromSt st
```
mrBliss: :+1:

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -611,36 +652,51 @@ modifyState VolatileDBEnv{_dbHasFS = hasFS :: HasFS m h, ..} action = do
...
 
 reverseMap :: forall blockId
```
mrBliss: Can you add a docstring?

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -142,16 +141,23 @@ volatileDbIsOpen :: OpenOrClosed blockId h -> Bool
...
-    , _currentWriteOffset :: !Word64 -- The 'WriteHandle' for the same file.
-    , _nextWriteFiles     :: ![(FsPath, Word)] -- The path and the size of the next files to write
-    , _nextNewFileId      :: !Int -- The next file name Id.
-    , _currentMap         :: !(Index blockId) -- The content of each file.
-    , _currentRevMap      :: !(ReverseIndex blockId) -- Where to find each block from slot.
-    , _currentSuccMap     :: !(SuccessorsIndex blockId) -- successors for each block.
-    , _currentMaxSlotNo   :: !MaxSlotNo -- Highest ever stored SlotNo
+data InternalState blockId h = InternalState
+    { _currentWriteHandle :: !(Handle h)
+    -- ^ The unique open file we append blocks.
```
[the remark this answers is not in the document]
kderme: It highlights that there is a single block we write to, at each time.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -254,68 +260,84 @@ getBlockImpl env@VolatileDBEnv{..} slot = do
...
-                if not (FileInfo.isFull _maxBlocksPerFile fileInfo')
-                then return (st', ())
-                else (\s -> (s,())) <$> nextFile hasFS _dbErr env st'
+putBlockImpl env@VolatileDBEnv{..} binfo@BlockInfo{..} builder =
+  modifyState env $ \hasFS@HasFS{..} st@InternalState{..} ->
+    if Map.member bbid _currentRevMap
+    then return (st, ()) -- putting an existing block is a no-op.
+    else do
+        bytesWritten <- hPut hasFS _currentWriteHandle builder
+        putBlockImpl' env binfo hasFS st bytesWritten
```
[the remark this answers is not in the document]
kderme: @mrBliss I am a bit confused about 4 vs 2 spaces. in which cases, do we have to use 4 but not 2?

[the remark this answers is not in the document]
mrBliss: The guideline in this repository is: use 2 spaces for indentation, unless there's a `where`-clause for the given block, in which case you should indent the `where` keyword 2 spaces and the block above the `where` keyword and the bindings below it with 4 spaces.

So:
```haskell
foo ... =
  bla di bla
```
but
```haskell
foo ... =
    bla di bla
  where
    bla = ..
```
In nested code blocks, the same principle holds.

I intentionally deviate from this guideline in one case: I indent the body of a top-level function (or a big local function) with 4 spaces by default, even without a `where`-clause, so that in case we add one, we don't have to reindent.

kderme:
@mrBliss Thanks for the extensive review!
> It's kind of fishy that we're using `modifyState` in a getter. I see that you just return the original state, but still.
> 
> Can you create a variant of `modifyState` that is read-only? See `withOpenState` from ImmutableDB.
> I don't think you need to do the whole `StateT` thing.

I think this fits well in my next task https://github.com/input-output-hk/ouroboros-network/issues/767.

> I would like to get rid of this one (used currently in the tests). Certainly now you have `getter` and `getterStm`.

I got some errors about escaping parameter when I tried to use the new 'getter' for the 'getInternalState'. I moved it now in testing modules. I keep in mind to rewrite it if I find a better way.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -244,8 +255,8 @@ getBlockImpl :: (IOLike m, Ord blockId)
              => VolatileDBEnv m blockId
              -> blockId
              -> m (Maybe ByteString)
-getBlockImpl env@VolatileDBEnv{..} slot = do
-    modifyState env $ \hasFS@HasFS{..} st@InternalState{..} -> do
+getBlockImpl env@VolatileDBEnv{..} slot =
+    modifyState env $ \hasFS@HasFS{..} st@InternalState{..} ->
```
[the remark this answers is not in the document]
kderme: I think this fits well in my next task https://github.com/input-output-hk/ouroboros-network/issues/767.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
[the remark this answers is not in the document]
mrBliss: I see. I think "sole" or "only" is more accurate in that case.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -228,226 +239,220 @@ isOpenDBImpl VolatileDBEnv{..} = do
...
+        | not isCurrent -> do
+            removeFile $ filePath fileId
+            return st {
+                _currentMap = Index.delete fileId _currentMap
+              , _currentRevMap = currentRevMap'
+              , _currentSuccMap = succMap'
+              }
+        | isCurrentNew  -> return st
+        | otherwise     -> do
+            -- TODO: Rething if we want to reopen after
```
mrBliss: `s/Rething/Rethink/`

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Util.hs
```diff
@@ -74,20 +75,32 @@ wrapFsError :: Monad m
...
 
--- Throws an error if one of the given file names does not parse.
+-- | Parses the 'FsPath' of each filename and zips them together.
+-- If any 'FsPath' does not parse, we return an error.
+parseAllFds :: forall blockId.
+               [FsPath]
+            -> Either (VolatileDBError blockId) [(FileId, FsPath)]
+parseAllFds = mapM go
+  where
+    go :: FsPath -> Either (VolatileDBError blockId) (FileId, FsPath)
```
mrBliss: Rename `go` to `parseFd`. Two reasons: (1) `go` is mostly used for functions that do explicit recursion, which is not the case here, the recursion is handled by `mapM`; (2) we mostly use `go` when we can't think of a good name, but in this case there is a  good candidate :slightly_smiling_face:

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Util.hs
```diff
@@ -74,20 +75,32 @@ wrapFsError :: Monad m
...
+    go file = case parseFd file of
+      Nothing -> Left $ UnexpectedError $ ParserError $ InvalidFilename file
+      Just fd -> Right (fd, file)
```
mrBliss: You could consider using `throwError` instead of `Left` and `return` instead of `Right`. Up to you.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Util.hs
```diff
@@ -74,20 +75,32 @@ wrapFsError :: Monad m
...
-        go fd file = case parseFd file of
-            Nothing  -> Left $ UnexpectedError $ ParserError $ InvalidFilename file
-            Just fd' -> Right $ Just $ maxMaybe fd fd'
+findLastFd = foldM go Nothing
+  where
+    maxMaybe :: Ord a => Maybe a -> a -> a
+    maxMaybe ma a = case ma of
+      Nothing -> a
+      Just a' -> max a' a
+    go :: Maybe FileId -> FsPath -> Either (VolatileDBError blockId) (Maybe FileId)
```
mrBliss: Hmm, this is basically the same `go` as the one in `parseAllFds`. Rename it to `parseFd`, pull out the `maxMaybe`, and promote it to the top-level so it can be shared between `parseAllFds` and `findLastFd`.

Ah, I see `parseFd` already exists. I would unify the two:
```haskell
parseFd :: FsPath -> Either (VolatileDBError blockId) FileId
parseFd file = maybe err Right $
    parseFilename <=< lastMaybe $ fsPathToList file
  where
    err = Left $ UnexpectedError $ ParserError $ InvalidFilename file

    parseFilename :: Text -> Maybe FileId
    parseFilename = readMaybe
                  . T.unpack
                  . snd
                  . T.breakOnEnd "-"
                  . fst
                  . T.breakOn "."
```

Update `unsafeParseFd` by using `either .. id ..`, 
`parseAllFds = mapM $ \f -> (,f) <$> parseFd f`, and
`findLastFd = fmap safeMaximum . mapM parseFd . Set.toList`.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/VolatileDB/TestBlock.hs
```diff
@@ -154,9 +155,18 @@ createFileImpl :: IOLike m
...
 createFileImpl hasFS env = do
-    SomePair _stHasFS st <- Internal.getInternalState env
+    SomePair _stHasFS st <- getInternalState env
     let nextFd = Internal._nextNewFileId st
     let path = Internal.filePath nextFd
     withFile hasFS path (AppendMode MustBeNew) $ \_hndl -> do
         return ()
     return ()
+
+getInternalState :: forall m blockId. IOLike m
```
mrBliss: I was more hoping that we wouldn't need access to the internal state of the VolatileDB in the tests anymore.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -458,120 +463,157 @@ reOpenFile :: forall m h blockId
...
+            newRevMp <- fromEither err $ reverseMap file currentRevMap fileMp
+            go newMp newRevMp newSuccMp newMaxSlot newHaveLessThanN rest
+          where
+            offset = case reverse blocks of
+              [] -> 0
+              (slotOffset, (blockSize,_)) : _ ->
+                -- The file offset is given by the offset of the last
+                -- block plus its size.
+                slotOffset + blockSize
+            fileMp = Map.fromList blocks
```
mrBliss: Use `Map` (like the suggested in my review) instead of `Mp` throughout this function.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -458,120 +463,157 @@ reOpenFile :: forall m h blockId
...
+      , newIndex + 1
+      , Index.insert newIndex FileInfo.empty mp
+      , 0 )
+
+    errorMsg = unwords
+        [ "VolatileDB invariant violation:"
+        , "The impossible happened!"
+        , "A file was found with less than "
+        , show n
+        , " blocks, but there are no files in db!"]
```
mrBliss: You don't need the space after "than " and before " blocks", `unwords` is taking care of them already.

"the" missing before "db!".

I still don't really understand the message: does it mean that a file was found (with less than x blocks?) while we didn't expect to find one?

Also, can you put it closer to where it is used?

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -228,226 +239,220 @@ isOpenDBImpl VolatileDBEnv{..} = do
     mSt <- readMVar _dbInternalState
     return $ volatileDbIsOpen mSt
 
--- closeDB . reOpenDB is a no-op. This is achieved because when we reOpen
--- we try to append on the latest created file.
+-- | Property: @'closeDB' >> 'reOpenDB'@ is a no-op. This is true because
+-- 'reOpenDB` will always append to the last created file.
```
mrBliss: ` should be '

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -458,120 +463,157 @@ reOpenFile :: forall m h blockId
...
+        [ "VolatileDB invariant violation:"
+        , "The impossible happened!"
+        , "A file was found with less than "
+        , show n
+        , " blocks, but there are no files in db!"]
+
+    truncateOnError Nothing _ _ = return ()
+    truncateOnError (Just _) file offset =
+        -- The handle of the parser is closed at this point. We need
+        -- to reOpen the file in 'AppendMode' now (parser opens with
```
mrBliss: `s/reOpen/reopen/`, because `reOpen` makes me think it is referring to a function, but it is not.

mrBliss:
Also, can you fix the warnings produced by `-Widentities`? Only those in the VolatileDB.

kderme:
`withState` and cleaning tests (which includes the `getInternalState`) come in a next pr.

I think I will squash all commits of this pr to a single "VolatileDB cleanup". Not sure if I can do something better at this point.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
kderme: Eventually, I changed a bit the implementation and it is not needed any more.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -146,9 +146,9 @@ volatileDbIsOpen :: OpenOrClosed blockId h -> Bool
 volatileDbIsOpen (VolatileDbOpen _) = True
 volatileDbIsOpen VolatileDbClosed   = False
 
-data InternalState blockId h = InternalState
-    { _currentWriteHandle :: !(Handle h)
-      -- ^ The unique open file we append blocks.
+data InternalState blockId h = InternalState {
+      _currentWriteHandle :: !(Handle h)
+      -- ^ The only open file we append blocks.
```
mrBliss: "we append blocks *to*" or "*to which* we append blocks".

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -374,7 +374,12 @@ tryCollectFile hasFS@HasFS{..} env slot st@InternalState{..} (fileId, fileInfo)
               }
         | isCurrentNew  -> return st
         | otherwise     -> do
-            -- TODO: Rething if we want to reopen after
+            -- We reach this case if we have to garbage collect the current file
+            -- we are appending blocks to. I'm not sure how realistic this case
+            -- is and maybe we want to skip this garbage collection.
+            --
+            -- 'ReOpenFile' techinally truncates the file to 0 offset, so any
```
mrBliss: `s/ReOpenFile/reOpenFile/` `s/techinally/technically/`

Ah, I now understand this case.

> I'm not sure how realistic this case is and maybe we want to skip this garbage collection.

Can you verify that this case is actually triggered in the tests? You can simply let it call `error` and see whether it is thrown. Maybe you'll need to play with the test parameters.

> I'm not sure how realistic this case is and maybe we want to skip this garbage collection.

I would replace this with an explanation:

> For this to happen, a garbage collection would have to be triggered for a slot ...

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -479,21 +478,20 @@ mkInternalStateDB :: (HasCallStack, MonadThrow m, MonadCatch m, Ord blockId)
...
+    dbDir = mkFsPath []
 
     toFsPath :: String -> FsPath
     toFsPath file = mkFsPath [file]
 
 -- | Makes the 'InternalState' by parsing all files.
 --
--- This function may create a new file to append new blocks or use an existing
--- one.
+-- It may create a new file to append new blocks or use an existing one.
```
mrBliss: new blocks *to*

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -506,33 +504,26 @@ mkInternalState hasFS err parser n files =
...
-      , newIndex + 1
       , Index.insert newIndex FileInfo.empty mp
       , 0 )
 
-    errorMsg = unwords
-        [ "VolatileDB invariant violation:"
-        , "The impossible happened!"
-        , "A file was found with less than "
-        , show n
-        , " blocks, but there are no files in db!"]
```
mrBliss: Ah, nice that this is no longer necessary.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -506,33 +504,26 @@ mkInternalState hasFS err parser n files =
...
-        -- 'ReadMode').
-        --
-        -- Note that no file is open at this point, so we can safely
-        -- open with 'AppendMode' any file, without the fear of opening
-        -- multiple concurrent writers, which is not allowed.
-        --
-        withFile hasFS file (AppendMode AllowExisting) $ \hndl ->
-            hTruncate hasFS hndl (fromIntegral offset)
+      -- The handle of the parser is closed at this point. We need
+      -- to reοpen the file in 'AppendMode' now (parser opens with
```
mrBliss: The `ο` in reοpen is a o micron, replace it with an ASCII o :slightly_smiling_face:

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Util.hs
```diff
@@ -75,32 +78,23 @@ wrapFsError :: Monad m
 wrapFsError fsErr volDBErr action =
     tryVolDB fsErr volDBErr action >>= either (throwError volDBErr) return
 
--- | Parses the 'FsPath' of each filename and zips them together.
--- If any 'FsPath' does not parse, we return an error.
+-- | Parses the 'FileId' of each 'FsPath' and zips them together.
+-- If any parsing does not parse, we abort with an error.
```
mrBliss: `If any parsing does not parse, we abort with an error.` -> `When parsing fails, we abort with the corresponding parse error.`

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Util.hs
```diff
@@ -75,32 +78,23 @@ wrapFsError :: Monad m
...
-    maxMaybe ma a = case ma of
-      Nothing -> a
-      Just a' -> max a' a
-    go :: Maybe FileId -> FsPath -> Either (VolatileDBError blockId) (Maybe FileId)
-    go fd file = case parseFd file of
-      Nothing  -> Left $ UnexpectedError $ ParserError $ InvalidFilename file
-      Just fd' -> Right $ Just $ maxMaybe fd fd'
+findLastFd = fmap safeMaximum . mapM parseFd . Set.toList
+
+maxMaybe :: Ord a => Maybe a -> a -> a
```
mrBliss: Is this function still needed? I'd like to get rid of it. Use `safeMaximum` instead.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Util.hs
```diff
@@ -117,12 +111,12 @@ tryVolDB :: forall m blockId a. Monad m
...
-    where
-      fromFS = UnexpectedError . FileSystemError
+  where
+    fromFS = UnexpectedError . FileSystemError
 
-      squash :: Either FsError (Either (VolatileDBError blockId) a)
-               -> Either (VolatileDBError blockId) a
-      squash = either (Left . fromFS) id
+    squash :: Either FsError (Either (VolatileDBError blockId) a)
+              -> Either (VolatileDBError blockId) a
```
mrBliss: Please align the `->` with the `::`

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/VolatileDB/Model.hs
```diff
@@ -49,7 +49,7 @@ import           Ouroboros.Storage.Util.ErrorHandling (ThrowCantCatch)
 import qualified Ouroboros.Storage.Util.ErrorHandling as EH
 import           Ouroboros.Storage.VolatileDB.API
 import qualified Ouroboros.Storage.VolatileDB.Impl as Internal
-import           Ouroboros.Storage.VolatileDB.Util
+import           Ouroboros.Storage.VolatileDB.Util hiding (maxMaybe)
```
mrBliss: So `maxMaybe` is defined here too? Replace it with `safeMaximum` too. If necessary use `max`, which works with two `Maybe a`s.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Util.hs
```diff
@@ -2,12 +2,12 @@
 {-# LANGUAGE LambdaCase          #-}
```
mrBliss: Please add an export list to this module.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
kderme: :man_facepalming: :man_facepalming:

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -611,36 +649,46 @@ modifyState VolatileDBEnv{_dbHasFS = hasFS :: HasFS m h, ..} action = do
...
+getterSTM fromSt VolatileDBEnv{..} = do
+    mSt <- readMVarSTM _dbInternalState
+    case mSt of
+      VolatileDbClosed  -> EH.throwError' _dbErrSTM $ UserError ClosedDBError
+      VolatileDbOpen st -> return $ fromSt st
 
-reverseMap :: forall blockId
-           .  Ord blockId
+-- | For each block found in a parsed file, we insert its 'InternalBlockInfo'.
+-- If the block is already found in the 'ReverseIndex' or is dublicated in the
```
mrBliss: `s/dublicated/duplicated/`

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/VolatileDB/TestBlock.hs
```diff
@@ -75,17 +76,17 @@ myParser hasFs = Parser {
 parseImpl :: forall m h. (MonadThrow m)
           => HasFS m h
           -> FsPath
-          -> m ([(SlotOffset, (BlockSize, BlockInfo BlockId))], Maybe String)
+          -> m ([(SlotOffset, (Word64, BlockInfo BlockId))], Maybe String)
```
mrBliss: Why this change? `BlockSize` is clearer than `Word64` to me.

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -228,226 +240,221 @@ isOpenDBImpl VolatileDBEnv{..} = do
     mSt <- readMVar _dbInternalState
     return $ volatileDbIsOpen mSt
 
--- closeDB . reOpenDB is a no-op. This is achieved because when we reOpen
--- we try to append on the latest created file.
+-- | Property: @'closeDB' >> 'reOpenDB'@  should be a no-op. This is true
+-- because 'reOpenDB` will always append to the last created file.
```
mrBliss:
```suggestion
-- because 'reOpenDB' will always append to the last created file.
```

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
mrBliss: > Can you verify that this case is actually triggered in the tests? You can simply let it call error and see whether it is thrown. Maybe you'll need to play with the test parameters.

@kderme Have you seen this happen?

kderme: Yes, q-s-m triggers this case. I left the comment:
```
-- We reach this case if we have to garbage collect the current file 
-- we are appending blocks to. For this to happen, a garbage
-- collection would have to be triggered for a slot which is bigger
-- than any recently inserted blocks.
```

kderme:
@mrBliss Yes q-s-m triggers this case frequently.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/VolatileDB/TestBlock.hs
mrBliss: I see, ok.

mrBliss:
If you squash your commits titled "Address review" (into the first one, I suppose), we can merge this!

mrBliss:
> If you squash your commits titled "Address review" (into the first one, I suppose), we can merge this!

I meant: squash all *your* commits titled "Address review" into the very first commit ("cleanup Volatile db"). Or, if some of the changes in the "Address review" commits belong to other commits in those.

I don't think that "Address review" commits in general belong in master, we don't need to see the whole history of each PR.

kderme:
@mrBliss I see, I squashed now my commits properly. 

I also checked the diff of the branches before and after the rebase and they are almost identical except for some small improvements.

mrBliss:
Thanks @kderme!

PR #1142, opened 2019-10-16
Configurable Protocol Magic.

karknu:
Use the Protocol Magic already present in Genesis config.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -42,6 +42,7 @@ import           Network.Socket as Socket
 
 import           Ouroboros.Network.Block
 import qualified Ouroboros.Network.Block as Block
+import           Ouroboros.Network.Magic
 import           Ouroboros.Network.NodeToClient as NodeToClient
 import           Ouroboros.Network.NodeToNode as NodeToNode
 import           Ouroboros.Network.Protocol.ChainSync.PipelineDecision
@@ -276,6 +277,8 @@ data RunNetworkArgs peer blk = RunNetworkArgs
     -- ^ IP producers
   , rnaDnsProducers          :: [DnsSubscriptionTarget]
     -- ^ DNS producers
+  , rnaNetworkMagic          :: NetworkMagic
+    -- ^ Network protocol magic, differentiates between mainnet, staging and testnetworks.
   }
 
 initNetwork
@@ -320,10 +323,8 @@ initNetwork registry nodeArgs kernel RunNetworkArgs{..} = do
       (protocolCodecs (getNodeConfig kernel))
       (protocolHandlers nodeArgs kernel)
 
-    -- TODO: network magics should be configurable, this gives an early fail if
-    -- one tries to run testnet vs mainnet.
-    nodeToNodeVersionData   = NodeToNodeVersionData { networkMagic   = 0 }
-    nodeToClientVersionData = NodeToClientVersionData { networkMagic = 0 }
+    nodeToNodeVersionData   = NodeToNodeVersionData { networkMagic   = rnaNetworkMagic }
+    nodeToClientVersionData = NodeToClientVersionData { networkMagic = rnaNetworkMagic }
 
     runLocalServer :: IO ()
     runLocalServer = do
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -248,38 +248,34 @@ type NetworkApps peer =
 
 -- | Arguments specific to the network stack
 data RunNetworkArgs peer blk = RunNetworkArgs
-  { rnaIpSubscriptionTracer    :: Tracer IO (WithIPList (SubscriptionTrace Socket.SockAddr))
+  { rnaIpSubscriptionTracer  :: Tracer IO (WithIPList (SubscriptionTrace Socket.SockAddr))
     -- ^ IP subscription tracer
-  , rnaDnsSubscriptionTracer   :: Tracer IO (WithDomainName (SubscriptionTrace Socket.SockAddr))
-  , rnaMuxTracer               :: Tracer IO (WithMuxBearer (MuxTrace NodeToNodeProtocols))
+  , rnaDnsSubscriptionTracer :: Tracer IO (WithDomainName (SubscriptionTrace Socket.SockAddr))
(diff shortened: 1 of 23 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -215,6 +219,9 @@ class ( Show (ChainState    p)
 newtype SecurityParam = SecurityParam { maxRollbacks :: Word64 }
   deriving (Show, Eq, Generic, NoUnexpectedThunks)
 
+newtype NetworkMagic = NetworkMagic {unNetworkMagic :: Word32 }
```
mrBliss: Missing space after the `{`

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -149,6 +149,7 @@ data PBftParams = PBftParams {
       -- but this implementation follows the specification by fixing that
       -- parameter to the ambient security parameter @k@.
     , pbftSignatureThreshold :: !Double
+    , pbftNetworkMagic :: !NetworkMagic
```
mrBliss: Please align these `=`

ouroboros-consensus/test-util/Test/Util/TestBlock.hs
```diff
@@ -288,6 +288,7 @@ singleNodeTestConfig :: NodeConfig (Bft BftMockCrypto)
 singleNodeTestConfig = BftNodeConfig {
       bftParams   = BftParams { bftSecurityParam = k
                               , bftNumNodes      = 1
+                              , bftNetworkMagic = NetworkMagic 0x0000ffff
```
mrBliss: Please align these `=`

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -289,8 +286,9 @@ initNetwork
   -> NodeArgs    IO peer blk
   -> NodeKernel  IO peer blk
   -> RunNetworkArgs peer blk
+  -> NetworkMagic
   -> IO ()
-initNetwork registry nodeArgs kernel RunNetworkArgs{..} = do
+initNetwork registry nodeArgs kernel RunNetworkArgs{..} magic = do
```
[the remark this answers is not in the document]
karknu: The genesis file is specified when you run `cardano-node`, it will be taken from it. Will provide a PR for it once this is approved.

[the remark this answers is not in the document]
mrBliss: Ah, I missed line 133 in this file, `(networkProtocolMagic cfg)`. That's what I was hoping to see.

Would it make sense to integrate `NetworkMagic` in `RunNetworkArgs`? It is an argument required to run the network layer, right? In that case I would put it there.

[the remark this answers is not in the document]
dcoutts: I think it would make sense to put into the `RunNetworkArgs`.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
dcoutts: Perhaps this type should go in the network lib and we can use it in `NodeToNodeVersionData` etc, without having to `unNetworkMagic` it.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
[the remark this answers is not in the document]
karknu: Sure I can move it into `RunNetworkArgs`.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
karknu: Where do you want to define the type? I don't want one NetworkMagic type for NodeToClient and one NetworkMagic type for NodeToNode.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -324,6 +322,9 @@ initNetwork registry nodeArgs kernel RunNetworkArgs{..} = do
       (protocolCodecs (getNodeConfig kernel))
       (protocolHandlers nodeArgs kernel)
 
+    nodeToNodeVersionData   = NodeToNodeVersionData { networkMagic   = unNetworkMagic $ rnaNetworkMagic }
+    nodeToClientVersionData = NodeToClientVersionData { networkMagic = unNetworkMagic $ rnaNetworkMagic }
```
mrBliss: I would expect the `networkMagic` to have type `NetworkMagic` too.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
karknu: I've moved `NetworkMagic` to a new module, `Ourboros.Network.Magic`.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -171,6 +172,10 @@ class ( Show (ChainState    p)
   -- | We require that protocols support a @k@ security parameter
   protocolSecurityParam :: NodeConfig p -> SecurityParam
 
+  -- | We require that protocols support a magic number used during version
+  -- negotiation.
+  protocolNetworkMagic  :: NodeConfig p -> NetworkMagic
```
dcoutts: I don't see where this is actually used.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -149,6 +150,7 @@ data PBftParams = PBftParams {
       -- but this implementation follows the specification by fixing that
       -- parameter to the ambient security parameter @k@.
     , pbftSignatureThreshold :: !Double
+    , pbftNetworkMagic       :: !NetworkMagic
```
dcoutts: It's also not clear to me that this is used, except to set `protocolNetworkMagic` which also appears to be unused.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
karknu: Discussed on slack. But `protocolNetworkMagic` is used by users of ouroboros-network such as cardano-node and cardano-explorer.

Issue #1140, opened 2019-10-16
Add pread to HasFS


kderme:
At the moment the VolatileDB does not allow concurrent readers, since they are serialized on the MVar of the Internal State. Without this protection and after https://github.com/input-output-hk/ouroboros-network/issues/767, two readers could read from the same handle, which creates a race condition on the file offset. To make this thread safe, the `pread` unix sys-call should be used.

@mrBliss proposed the name `hGetSomeAt` which seems good to me.

Windows seem to provide similar functionality based on https://stackoverflow.com/questions/766477/are-there-equivalents-to-pread-on-different-platforms (I haven't tried yet).

This functionality shouldn't be needed in the ImmutableDB, since there are not concurrent readers of the same handle.

mrBliss:
> This functionality shouldn't be needed in the ImmutableDB, since there are not concurrent readers of the same handle.

It's not necessary, but I would use it in the ImmutableDB too, as it saves us one system call.

kderme:
In order to take full advantage of this design and make reads fully parallel, we may want to use some read locks in the fs layer. At the moment all syscalls are serialized from `MVar` (see `HandleOS`), but a different kind of lock (`Control.Concurrent.ReadWriteLock`?), which allows many parallel pread could be used.

Even without this, we get some benefit in terms of parallelism, however this will make it "more parallel".

mrBliss:
> In order to take full advantage of this design and make reads fully parallel, we may want to use some read locks in the fs layer. At the moment all syscalls are serialized from `MVar` (see `HandleOS`), but a different kind of lock (`Control.Concurrent.ReadWriteLock`?), which allows many parallel pread could be used.
> 
> Even without this, we get some benefit in terms of parallelism, however this will make it "more parallel".

The `StrictMVar` that Edsko introduced supports this already: https://github.com/input-output-hk/ouroboros-network/blob/d3e71facd53da4a2c72e4947f551ddeac45fba41/ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs#L128

kderme:
In the Windows implementation, there is another unfortunate issue with versions. `OVERLAPPED` was supported on a newer version, which katip blocks. https://github.com/haskell/win32/commit/760338e7d289a21281f5a5f62f385f3a3808354f#diff-a99994770dbe04188c1176136aef3d32
[...]

Issue #882, opened 2019-08-05
CI failure for test-storage on mingw32


mrBliss:
Noticed in #879.
```
================================================================================
RUNNING TESTS for ouroboros-consensus-0.1.0.0-test-test-storage-x86_64-pc-mingw32 via wine64
================================================================================
Copying extra test libraries ...
Copying library dependencies...
ghc-pkg: : getDirectoryContents:openDirStream: does not exist (No such file or directory)
ouroboros-storage
  Storage
    HasFS
      HasFS
        q-s-m:
Model {knownHandles = RefEnv
                        {toList = []}
      ,knownPaths = RefEnv {toList =
                      []}
      ,mockFS = MockFS {mockFiles =
                       Folder (Map.fromList [])
                       ,mockHandles = Map.fromList []
                       ,mockNextHandle = Handle 0}}

   == At (Open (PExpPath ["t1"]) (WriteMode MustBeNew)) ==> At (Resp {getResp = Right (WHandle (Reference (Concrete ["t1"])) (Reference (Concrete Opaque)))}) [ 0 ]

Model {knownHandles = RefEnv
                        {toList = [+_FAIL (0.23s)
          *** Failed! Falsifiable (after 50 tests and 29 shrinks):
          Commands
            { unCommands =
                [ Command
                    (At (Open (PExpPath [ "t1" ]) (WriteMode MustBeNew)))
                    (At
                       Resp
                         { getResp =
                             Right
                               (WHandle
                                  (Reference (Symbolic (Var 0))) (Reference (Symbolic (Var 1))))
                         })
                    [ Var 0 , Var 1 ]
                , Command
                    (At (Put (Reference (Symbolic (Var 1))) "\NUL"))
                    (At Resp { getResp = Right (Word64 1) })
                    []
                , Command
                    (At (Close (Reference (Symbolic (Var 1)))))
                    (At Resp { getResp = Right (Unit ()) })
                    []
                , Command
                    (At
                       (Open
                          (PExpRef (Reference (Symbolic (Var 0))))
                          (ReadWriteMode AllowExisting)))
                    (At
                       Resp
                         { getResp =
                             Right
                               (WHandle
                                  (Reference (Symbolic (Var 2))) (Reference (Symbolic (Var 3))))
                         })
                    [ Var 2 , Var 3 ]
                , Command
                    (At (Get (Reference (Symbolic (Var 3))) 1))
                    (At Resp { getResp = Right (ByteString "\NUL") })
[...]

PR #1149, opened 2019-10-21
Concurrent readers

kderme:
This pr addresses a number of issues:
- Keeps open a ReadMode handle for each file in the VolatileDB, so that reading threads don't reopen their own. https://github.com/input-output-hk/ouroboros-network/issues/767
- Use pread syscall to fix the race condition on the file offset. pread is also implemented in Windows (in a hacky way) and related q-s-m tests are added. https://github.com/input-output-hk/ouroboros-network/issues/1140
- read/write locks are used. This allows to maximize read parallelism and take full advantage of the above feautures.
- A Windows related bug is fixed, which failed on ci https://github.com/input-output-hk/ouroboros-network/issues/882

For the read/write locks, I didn't use `readMVarSTM` as discussed, but `RWVar` from concurrent-extras. What I actually want is not write lock, but exclusive locks (when there is a writer, no reader should be allown). If I understand correctly, `StrictMVar` is not designed for this case.

I think ChainDB has some parallel tests, but this pr needs more testing, since VolatileDB lacks parallel q-s-m tests.

Finally this is built on top of https://github.com/input-output-hk/ouroboros-network/pull/1137 and will need rebasing when it is merged.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -241,6 +241,7 @@ library
      build-depends:       Win32
   else
      build-depends:       unix
+                        , unix-bytestring
 
   ghc-options:         -Wall
                        -Wno-unticked-promoted-constructors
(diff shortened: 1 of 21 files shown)


mrBliss:
Can you rebase this PR on top of master now that #1137 is merged?

kderme:
Closing in favor of other smaller prs, which split the same work (https://github.com/input-output-hk/ouroboros-network/pull/1153, https://github.com/input-output-hk/ouroboros-network/pull/1164).

Issue #1120, opened 2019-10-08
Turn mux tracing String tag into a polymorphic variable


coot:
This way we will avoid `Show` instances on `peerid` floating around.

karknu:
Fixed by #1162

PR #1154, opened 2019-10-22
Let MuxTracer log handshake errors

karknu:


ouroboros-network/src/Ouroboros/Network/Socket.hs
```diff
@@ -249,6 +253,7 @@ beginConnection muxTracer handshakeTracer encodeData decodeData acceptVersion fn
         let muxTracer' = Mx.WithMuxBearer (show peerid) `contramap` muxTracer
```
coot: it's outside of this PR, but it would be good to remove the `show peerid` and just store `peerid`. Do we have an issue for that?

karknu: I don't think there is an issue for it, but I remember the comment.

network-mux/src/Network/Mux/Types.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE GADTs                 #-}
```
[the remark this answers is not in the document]
karknu: `ExistentialQuantification` is enough, will change it.

ouroboros-network/src/Ouroboros/Network/Socket.hs
coot: Could you add one so we don't forget about it.

karknu: https://github.com/input-output-hk/ouroboros-network/issues/1120 found it.

PR #1162, opened 2019-10-24
Remove Show constraint for peerid

karknu:


network-mux/src/Network/Mux/Types.hs
```diff
@@ -291,14 +291,14 @@ instance Exception MuxError where
 
 -- Type used for tracing mux events.
 --
-data WithMuxBearer a = WithMuxBearer {
-      wmbTag   :: !String
+data WithMuxBearer peerid a = WithMuxBearer {
+      wmbTag   :: !peerid
```
coot: I would rename it to `wmbPeerId` as well, hopefuly there are not too many places where it is used.

PR #1163, opened 2019-10-24
Replace OuroborosTag.protocolnetworkMagic with RunNode.nodeNetworkMagic

mrBliss:


mrBliss:
I'll submit PRs for cardano-node, cardano-byron-proxy, and cardano-explorer. Note that these do not need to be coordinated, as things will not break, it's just a matter of calling a different function that will produce the same result.

PR #1160, opened 2019-10-23
Update win32 version

kderme:
This pr updates the `Win32` version, given some recent changes on katip and iohk-monitoring-framework.
The new version allows to revert an ugly hack we did to support seeking in windows.

PR #1164, opened 2019-10-24
Add pread syscall and tests

kderme:
First two commits are part of https://github.com/input-output-hk/ouroboros-network/pull/1160
Addresses https://github.com/input-output-hk/ouroboros-network/issues/1140
pread syscall is thread safe, so in a future pr, we will allow concurrent calls with read/write locks.
However in windows read_File always changes the file offset, so write locks will be needed for windows even for pread.


ouroboros-consensus/test-util/Test/Util/FS/Sim/MockFS.hs
```diff
@@ -542,6 +580,56 @@ hGetSome err@ErrorHandling{..} h n =
...
+    withOpenHandleStateRead err h $ \fs OpenHandle{..} -> do
+      file <- checkFsTree err $ FS.getFile openFilePath (mockFiles fs)
+      let fsize = fromIntegral (BS.length file) :: Word64
+      case (openPtr, sign64 o)  of
+        (RW r _ _, Positive o') -> do
+          unless r $ throwError (errNoReadAccess openFilePath "write")
+          let bs = BS.take (fromIntegral n) . BS.drop (fromIntegral o) $ file
+          -- This is the same fsLimitation we get when we seek past the end of
+          -- EOF, in AbsoluteSeek mode.
+          when (o' > fsize) $ throwError (errPastEnd openFilePath)
```
mrBliss: Why do we need to check whether the current offset *of the handle* is past the end of the file?

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -185,27 +187,28 @@ run hasFS@HasFS{..} = go
...
 -------------------------------------------------------------------------------}
 
--- The function 'hGetSome' and 'hPutSome' might perform partial reads/writes,
--- depending on the underlying implementation, see #277. While the model will
--- always perform complete reads/writes, the real IO implementation we are
--- testing /might/ actually perform partial reads/writes. This testsuite will
--- fail when such a partial read or write is performed in the real IO
--- implementation, as these are undeterministic and the model will no longer
--- correspond to the real implementation. See #502 were we track this issue.
+-- The function 'hGetSome', 'hGetSomeAt' and 'hPutSome' might perform partial
```
mrBliss: While you're at it:
```suggestion
-- The functions 'hGetSome', 'hGetSomeAt' and 'hPutSome' might perform partial
```

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -489,6 +493,13 @@ generator Model{..} = oneof $ concat [
          , choose (-1, -10)
          ]
 
+    genOffsetPos :: Gen Int64
+    genOffsetPos = frequency
+         [ (5, return 0)
+         , (3, choose (1, 3))
+         , (2, choose (4, 10))
+         ]
```
mrBliss: Why these values? Why not `NonNegative (Small Word64)`, i.e.`genOffset = getNonNegative . getSmall <$> arbitrary`?

Alternatively, the generator could take the file size into account and generate 0, some valid offset, and offsets >= file size.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -1164,6 +1197,8 @@ tag = C.classify [
           Right $ tagPutSeekGet put (Set.insert h seek)
         Get (Handle h _) n | n > 0 && Set.member h seek ->
           Left TagPutSeekGet
+        GetAt (Handle h _) n _ | n > 0 && Set.member h seek ->
+          Left TagPutSeekGet
```
mrBliss: I don't think this tag applies.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -1175,6 +1210,8 @@ tag = C.classify [
           Right $ tagPutSeekNegGet put (Set.insert h seek)
         Get (Handle h _) n | n > 0 && Set.member h seek ->
           Left TagPutSeekNegGet
+        GetAt (Handle h _) n _ | n > 0 && Set.member h seek ->
+          Left TagPutSeekNegGet
```
mrBliss: This tag neither.

ouroboros-consensus/src-win32/Ouroboros/Storage/IO.hs
```diff
@@ -74,6 +75,18 @@ read fh bytes = withOpenHandle "read" fh $ \h ->
...
     fromIntegral <$> win32_ReadFile h ptr (fromIntegral bytes) Nothing
 
+getCurrentFileOffset :: HANDLE -> IO Int64
+getCurrentFileOffset h = setFilePointerEx h 0 fILE_CURRENT
+
+pread :: FHandle -> Word64 -> Int64 -> IO ByteString
+pread fh bytes pos = withOpenHandle "pread" fh $ \h ->
+  Internal.createUptoN (fromIntegral bytes) $ \ptr -> do
+    initialOffset <- getCurrentFileOffset h
+    _ <- setFilePointerEx h pos fILE_BEGIN
```
mrBliss: What if it fails? Does it throw an error or does it report a special return value?

ouroboros-consensus/src-win32/Ouroboros/Storage/IO.hs
```diff
@@ -74,6 +75,18 @@ read fh bytes = withOpenHandle "read" fh $ \h ->
...
+    n <- fromIntegral <$> win32_ReadFile h ptr (fromIntegral bytes) Nothing
+    _ <- setFilePointerEx h initialOffset fILE_BEGIN
```
mrBliss: Hmm, this is unfortunate.

ouroboros-consensus/test-util/Test/Util/FS/Sim/MockFS.hs
kderme: This is not the offset of the file, but the offset we're reading. It is an error to give an offset bigger than the file size, so I have an fslimitation here.

mrBliss: But `o` is the offset we're reading. `o'` is the current offset of the file, which should not matter for `pread`, I would assume.

ouroboros-consensus/src-win32/Ouroboros/Storage/IO.hs
kderme: The Haskell wrappers take any failing error code and throw it with `throwIO`.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
kderme: Because the tag name has Get instead of GetAt?

I think it tags tests where Seek interacts with Get. GetAt has different properties, but we still want such tests. So maybe I should just add a different tag, like TagPutSeekGetAt?

mrBliss: Ok, I thought the tag was more specific: put something, seek such that the next get reads the thing that was put, then do that get. This wouldn't work the same way for getAt. But I was mistaken.

ouroboros-consensus/test-util/Test/Util/FS/Sim/MockFS.hs
kderme: `o'` is just the absolute value of `o`. The file offset would be given by `RW r _ o`, but we actually ignore the third argument of `RW`. So we don't have scope to the file offset anywhere in this function.

mrBliss: :man_facepalming: Sorry, I misread. Probably because in my mind the offset was already a `Word64` (as I suggested before).

ouroboros-consensus/src-win32/Ouroboros/Storage/IO.hs
kderme: Yes, unfortunately `ReadFile` always moves the file offset based on my tests and [this](https://stackoverflow.com/a/14448717)

mrBliss: Note to self: currently, handles are accessed using `withMVar`, so access is serialised/thread-safe already. In the future, we will try to relax this to allow multiple `pread`s via the same handle, but only on Unix.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -555,9 +574,12 @@ shrinker Model{..} (At cmd) =
           fp :: FsPath
           fp = resolvePathExpr knownPaths pe
 
-      Get      h n  -> At . Get      h <$> shrink n
-      Put      h bs -> At . Put      h <$> shrinkBytes bs
-      Truncate h n  -> At . Truncate h <$> shrink n
+      Get      h n   -> At . Get      h <$> shrink n
+      GetAt    h n o -> do
+        n' <- shrink n
+        return $ At $ GetAt h n' o
```
mrBliss: Why not shrink the offset too?

cabal.project
```diff
@@ -32,7 +32,13 @@ source-repository-package
   type: git
   location: https://github.com/input-output-hk/iohk-monitoring-framework
   subdir:   contra-tracer
-  tag: 1ebe57df3a76a8ae4e3b5a31a9b85132f71534fc
+  tag: 5b91f7a10465ae1215522a39d42df7e970ab8e8f
+
+source-repository-package
+  type: git
+  location: https://github.com/Soostone/katip.git
+  tag: cf60f0e14640ae9702c111bb51aeca7eed50ad48
```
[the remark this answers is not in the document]
kderme: Actually it seems we don't need it in Cabal either. A bit strange, because I remember it was necessary.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -555,9 +575,13 @@ shrinker Model{..} (At cmd) =
           fp :: FsPath
           fp = resolvePathExpr knownPaths pe
 
-      Get      h n  -> At . Get      h <$> shrink n
-      Put      h bs -> At . Put      h <$> shrinkBytes bs
-      Truncate h n  -> At . Truncate h <$> shrink n
+      Get      h n   -> At . Get      h <$> shrink n
+      GetAt    h n o -> do
+        n' <- shrink n
+        o' <- shrink o
```
mrBliss: If `n = 5` and `o = 5`, this won't include `n' = 5, o' = 4` or `n' = 4, o' = 5`.
So instead:
```haskell
GetAt h n o -> At <$>
  [GetAt h n o' | o' <- shrink o] <>
  [GetAt h n' o | n' <- shrink n]
```
(not type-checked, maybe some parentheses are needed)

mrBliss:
Do you plan on merging #1160 first or not?

kderme:
> Do you plan on merging #1160 first or not?

Probably not since it's outdated.

mrBliss:
> > Do you plan on merging #1160 first or not?
> 
> Probably not since it's outdated.

Alright, but don't forget about: https://github.com/input-output-hk/ouroboros-network/pull/1160#issuecomment-547317515

kderme:
> > > Do you plan on merging #1160 first or not?
> > 
> > 
> > Probably not since it's outdated.
> 
> Alright, but don't forget about: [#1160 (comment)](https://github.com/input-output-hk/ouroboros-network/pull/1160#issuecomment-547317515)

Right, I'll try to erase it. I need to install stack on my windows at some point. 
Btw I messed up something while rebasing, let me fix it before merge

stack.yaml
```diff
@@ -61,8 +61,5 @@ extra-deps:
   - tasty-hedgehog-1.0.0.1
   - Unique-0.4.7.6
 
-  # Windows only
-  - Win32-2.5.4.1
-
```
mrBliss: Looking at the Stack snapshot we're using: https://www.stackage.org/lts-13.26, it seems that Win32 is pinned to 2.6.1.0, while you're introducing a `Win32 >= 2.6.2.0` constraint in the cabal file. So I don't think this is buildable on Windows using Stack.

Unfortunately, our CI doesn't catch this.

Can you try building it with Stack on Windows?

That's the only thing blocking this PR for me.

kderme: I get many errors `the stack configuration has no specified version` on windows, if I don't add the 3 dependencies above.

mrBliss: Those three dependencies are already pinned in lts-13.26, so it is strange that you have to add them. I'm thinking too that this is a local problem. I wouldn't feel comfortable adding them.

kderme: @mrBliss do you think this pr is ready now? I can't find what I do wrong in my local machine.

mrBliss: Yeah, let's merge it. At least we know that it doesn't break things on Unix :slightly_smiling_face:.
Building with Cabal works for you on Windows, right?

kderme:
Yes cabal builds fine.

stack.yaml
kderme: Yes cabal builds fine.

PR #1171, opened 2019-10-25
Small wrapper for writing simple DB analyses

edsko:


ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -432,3 +432,18 @@ executable byron-db-converter
    default-language:   Haskell2010
    main-is:            Main.hs
    ghc-options:        -Wall
+
+executable analyse-db
+  hs-source-dirs:      tools/db-analyse
+  main-is:             Main.hs
+  build-depends:       base >=4.12 && <4.13
```
mrBliss: Why the bounds? (I don't see other bounds in this .cabal)

ouroboros-consensus/tools/db-analyse/Main.hs
```diff
@@ -0,0 +1,263 @@
...
+import           Ouroboros.Consensus.Node.Run.Abstract
+import           Ouroboros.Consensus.Node.Run.Byron ()
+import           Ouroboros.Consensus.Util.ResourceRegistry
+
+import           Ouroboros.Storage.ChainDB.API (StreamFrom (..))
+import           Ouroboros.Storage.ChainDB.Impl.ImmDB
+import           Ouroboros.Storage.Common
+import           Ouroboros.Storage.EpochInfo
+import qualified Ouroboros.Storage.ImmutableDB.API as ImmDB
+--import           Ouroboros.Storage.ImmutableDB.Layout
```
mrBliss: Stray comment

ouroboros-consensus/tools/db-analyse/Main.hs
```diff
@@ -0,0 +1,263 @@
...
+
+showSlotBlockNo :: Analysis
+showSlotBlockNo immDB _epochInfo rr =
+    processAll immDB rr go
+  where
+    go :: Either EpochNo SlotNo -> Blk -> IO ()
+    go isEBB blk = putStrLn $ intercalate "\t" [
+                       show isEBB
+                     , show (blockNo   blk)
+                     , show (blockSlot blk)
```
mrBliss: You know I prefer `condense` instead of `show`+ a header line :grin:

edsko: It makes the output less ambiguous. We'd have to do some additional massaging to clearly mark EBBs then. Let's put in the time when that's useful.

PR #1180, opened 2019-10-29
Cleanup the tests of VolatileDB

kderme:
Addresses https://github.com/input-output-hk/ouroboros-network/issues/1119


mrBliss:
This should be orthogonal to #1165.

An important goal is to make the tests less implementation-specific, i.e., the corruption/error simulation should not rely on the implementation.

mrBliss:
> This should be orthogonal to #1165.
> 
> An important goal is to make the tests less implementation-specific, i.e., the corruption/error simulation should not rely on the implementation.

Because of the weird test block used by VolatileDB tests, I could not easily add `IsEBB` to the VolatileDB in #1333 (see https://github.com/input-output-hk/ouroboros-network/pull/1331/commits/0bd4458cd7a83aa79434f7b99d12f1caa0ed375c#diff-bb4db16705ba94291a0f306156f73335R112).

This should be done before #1165, so all commits relating to #1165 should be removed from this PR.

We should use `Test.Storage.TestBlock` for the VolatileDB instead of the current ad-hoc test block.

kderme:
- This pr simplifies and beautifies the tests of the voldb.
- The tests now use the existing `TestBlock`, but I still use my custom Corruptions.

- Simulated errors don't depend that much on the implementation any more. I have followed a different approach from what's described here:

> About not hard-coding what should happen in case of simulated errors in the q-s-m tests:
> 
> * Before you run such a command, make copies of the VolatileDB files
> * Run the command
>   a. If it fails because of a simulated error, close the VolatileDB, restore the copies of the files, restart in the previous state.
>   b. If it doesn't fail, do the same as in a. anyway, because the outcome needs to be deterministic.
> 
> In the model, executing a command with a simulated error is a no-op.

When there is a simulated error for a command which changes the db (like put and gc), I run the same command for a second time without errors. If the first time the command succeeds, the second is a no-op. If it partially or fully fails, the second should bring the db to a consistence state, as if there was no initial failure. 

The Model always succeeds, as if there are no simulated errors.

PR #1183, opened 2019-10-31
`ByronGiven`-free instance for `HasHeader`

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
```diff
@@ -661,20 +685,41 @@ applyByronLedgerBlockOrEBB validationMode
...
+    computeSlot :: CC.Block.ABlockOrBoundary ByteString -> SlotNo
+    computeSlot (CC.Block.ABOBBlock blk) =
+        convertSlot $ CC.Block.headerSlot hdr
+      where
+        hdr = CC.Block.blockHeader blk
+    computeSlot (CC.Block.ABOBBoundary blk) =
+        SlotNo $ CC.Slot.unEpochSlots epochSlots * CC.Block.boundaryEpoch hdr
+      where
+        hdr = CC.Block.boundaryHeader blk
+
```
mrBliss: This should make use `computeHeaderSlot` (that's why that one's top-level).

PR #1184, opened 2019-10-31
Import the ntp-client library part II.

MarcFontaine:
* rename Package to ntp-client
* Use Network.NTP namespace
* Add the package to nix, stack and release.nix
Also:
Make the API more explicit:
Instead  of a TVAR return a record with callback functions.


ntp-client/src/Network/NTP/Client.hs
```diff
@@ -65,6 +56,54 @@ data NtpClientSettings = NtpClientSettings
     }
 
 data NtpClient = NtpClient
+    { ntpGetStatus        :: IO NtpStatus -- Use STM here ?
+      -- ^ Query the current status (non-blocking).
+    , ntpActNow           :: IO ()
+      -- ^ Bypass all internal threadDelays and trigger a new ntp query.
```
coot: Do we need this? Isn't `ntpForcecheck` enough?

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -65,6 +56,54 @@ data NtpClientSettings = NtpClientSettings
...
+    , ntpForceCheck       :: IO NtpStatus
+      -- ^ Perform a ntp query and wait for the results (blocking with timeout).
+    , ntpAbort            :: IO () --todo/discuss
```
coot: do you mean a clean shutdown of the service? As a first step I would test if one can shutdown it with an async exception, e.g. do all the threads are shut down properly?

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: Yes maybe its enought..I Just wanted to keep the existing thing and add the new new Idea so it does not get lost. Also there is no real implementation of ntpForceCheck yet.

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: Good point I'll check out the async exception.

ntp-client/src/Network/NTP/Client.hs
coot: There is, but it's not explicit (by updating the `TVar`).

coot: No need to keep both api's at the same time, but if you want to keep it for the time being of working on this PR that's ok.

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: This does not work:
```
test :: IO ()
test = do
    t <- async $ startNtpClient (contramapM (return . show) stdoutTracer) settings
    threadDelay 30000000
    putStrLn "trying to kill the client"
    cancel t
```
Also an explicit function may be better.
An other option may be to use a proper with-style wrapper
`withNtpClient :: Tracer IO NtpTrace -> NtpClientSettings -> (NtpClient -> IO a) -> IO a`

coot: yes, but this will for sure succeed.  The question is if all the threads that `startNtpClient` starts will be terminated as well. In Haskell if you start a thread it can out live the parent thread. One needs to analyse the code.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -265,21 +304,6 @@ spawnNtpClient tracer settings ncStatus = do
             resolve
           else return addrs
 
--- |
--- Run NTP client in a separate thread; it returns a mutable cell which holds
--- @'NtpStatus'@.
---
--- This function should be called once, it will run an NTP client in a new
--- thread until the program terminates.
-withNtpClient :: Tracer IO NtpTrace -> NtpClientSettings -> IO (TVar NtpStatus)
```
coot: Provide a simlar top level function.  Maybe `withNtpClient :: Tracer IO NptTrace -> NtpClientSetting -> IO NtpClient`.
Including `Tracer` in `NtpClientSettings` might be a good idea too.

MarcFontaine: It is already there. It has just been renamed to `startNtpClient`

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -65,6 +56,54 @@ data NtpClientSettings = NtpClientSettings
...
+    return $ NtpClient
+        { ntpGetStatus = do
+            traceWith tracer NtpTraceClientGetStatus
+            atomically $ readTVar ncStatus
+        , ntpActNow    = do
+            traceWith tracer NtpTraceClientActNow
+            atomically $ writeTVar ncStatus NtpSyncPending
+        , ntpForceCheck = do
+            traceWith tracer NtpTraceClientForceCheck
+            error "ntpForceCheck: Todo"
```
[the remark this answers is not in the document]
MarcFontaine: OK.
We can also discuss the final API later

ntp-client/src/Network/NTP/Client.hs
coot: ok I see now; I would rename it as `forkNtpClient`.

ntp-client/src/Network/NTP/Client.hs
coot: btw, `startNtpClient` is already using `async` to fork the client, so there's no need to to fork it again as you do you the snippet above.

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: OK done

stack.yaml
```diff
@@ -10,6 +10,7 @@ packages:
   - ./ouroboros-consensus
   - ./io-sim
   - ./io-sim-classes
+  - ./ntp-client
```
mrBliss: Can you also add it to https://github.com/input-output-hk/ouroboros-network/blob/adf9378f7a4afee344ca6353a89ebfa152eb564a/release.nix#L21

I'm no Nix expert, but I see the components of this repo mentioned in some other nix files as well. Maybe they will be added when running `./nix/regenerate.sh`? We want to be sure that this component is also compiled by CI.

MarcFontaine: It not ready yet to make it a required build but CI can already start building it.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -58,10 +58,10 @@ data NtpClientSettings = NtpClientSettings
 data NtpClient = NtpClient
     { ntpGetStatus        :: IO NtpStatus -- Use STM here ?
       -- ^ Query the current status (non-blocking).
-    , ntpActNow           :: IO ()
+    , ntpForceCheck           :: IO ()
```
coot: Shouldn't `ntpForceCheck` return `NtpStatus`? Otherwise there's no way to know that it has already finished (`ntpGetStatus` only allows polling which is not performant).  Since you have the `TVar NtpStatus` at hand you can await untill `NtpStatus` is not `NtpSyncPending`:

```
await :: TVar NtpStatus -> IO NtpStatus
await v = atomically $ do
   s <- readTVar v
   when (s == NtpSyncPending) retry
   pure s
```

MarcFontaine: Yes but lets do that later and first merge the easy stuff, i.e. renaming the package etc.

coot: ok

Issue #925, opened 2019-08-15
Resolve semantic conflict between `WithEBBs` and CS intersections at EBBs


nfrisby:
The `WithEBB` protocol combinator's `applyChainState` method forwards regular blocks to the underlying protocol but suppresses EBBs. (See #704.)

https://github.com/input-output-hk/ouroboros-network/blob/4ae0997add0305102810a54097e684753782d4f7/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/WithEBBs.hs#L55-L60

However, a chain sync client may receive `MsgIntersectFound` with an EBB intersection point. This can lead the corresponding `rewindChainState` to fail, cascading to an exception such as a chain sync client's `InvalidIntersection`. I'm specifically seeing this failure in the `RealPBFT` test when the chain is just one EBB and the `MsgIntersectFound` identifies that EBB. This arises on my #773 branch that lets some nodes join later than others.

That is a very specific and seemingly unrealistic corner case, but it highlights the more general conflict between `WithEBB`'s suppression of EBBs and the general semantics of `applyChainState`/`rewindChainState`: Can `rewindChainState` assume that the target point corresponds to a prior call to `applyChainState`? Currently `applyChainState @WithEBB` violates that assumption.

`PBFT` currently makes a similar assumption. Its `rewindChainState` refuses to rewind past the earliest block in its chain state for a non-`Origin` request. Hence the `RealPBFT` failure, since the requested EBB intersection point is the first (and only) block in the chain but never made it into the chain state.

https://github.com/input-output-hk/ouroboros-network/blob/4ae0997add0305102810a54097e684753782d4f7/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs#L232-L236

So: Should `WithEBB` stop suppressing EBBs or should `rewindChainState` methods make fewer assumptions? Or something else?

cc: @nc6 @edsko @mrBliss

nfrisby:
My current repro is unfortunately tangled up with #773 details, and I'm unfamiliar with the ledger details etc involved in setting up a standalone repro forked off `master`.

FYI, simplifying `PBFT`'s `rewindChainState` to permit returning the empty sequence in the `At` alternative avoids the exception -- I'm doing that locally for now, but it seems presumptuous to assume that's The Right Fix.

mrBliss:
@nfrisby Another nice find! 

@nc6: is there a reason for ignoring EBBs in `rewindChainState`?

[...]

PR #1189, opened 2019-11-04
Support EBBs directly in PBFT/Remove ByronBlock (leaving only ByronBlockOrEBB)

edsko:
This is still part of the ongoing push towards a version of the consensus layer where we can give chain selection a `SupportedHeader` rather than `IsHeader` constraint.

Note that this removes one back but does not yet insert the proper solution; need to talk to @mrBliss about that (https://github.com/input-output-hk/ouroboros-network/issues/925#issuecomment-527184811).


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
```diff
@@ -150,373 +146,63 @@ newtype ByronHash = ByronHash { unByronHash :: CC.Block.HeaderHash }
...
-        $ hdr
-
 {-------------------------------------------------------------------------------
   Epoch Boundary Blocks
 -------------------------------------------------------------------------------}
 
 data ByronBlockOrEBB = ByronBlockOrEBB
   { bbRaw    :: !(CC.Block.ABlockOrBoundary ByteString)
   , bbSlotNo :: !SlotNo
+  , bbHash   :: ByronHash
```
edsko: Bring back comment regarding laziness.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
```diff
@@ -150,373 +146,63 @@ newtype ByronHash = ByronHash { unByronHash :: CC.Block.HeaderHash }
...
-mkByronHeaderOrEBB' slotNo header = case header of
-    -- By pattern-matching on the two cases, we satisfy the format
-    -- 'ByronHeaderOrEBB' desires.
-    Left ebb -> ByronHeaderBoundary ebb slotNo h
-      where
-        h = ByronHash $ CC.Block.boundaryHeaderHashAnnotated ebb
-    Right mb -> ByronHeaderRegular mb slotNo h
-      where
-        h = ByronHash $ CC.Block.headerHashAnnotated mb
+mkByronHeaderOrEBB' slotNo hdrHash header = case header of
```
edsko: Assert that slot number explicitly passed matches the one in the (regular) header.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
```diff
@@ -662,12 +375,83 @@ instance UpdateLedger ByronBlockOrEBB where
...
 
-deriving newtype instance UpdateLedger ByronBlock => NoUnexpectedThunks (LedgerState ByronBlockOrEBB)
+instance NoUnexpectedThunks (LedgerState ByronBlockOrEBB)
+  -- use generic instance
 
 instance ConfigContainsGenesis (LedgerConfig ByronBlockOrEBB) where
-  genesisConfig = genesisConfig . unByronEBBLedgerConfig
+  genesisConfig = unByronLedgerConfig
+
+applyByronLedgerBlock :: ValidationMode
```
edsko: Rename to `applyABlock`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
```diff
@@ -676,36 +460,38 @@ applyByronLedgerBlockOrEBB :: ValidationMode
...
+    , bbSlotNo = computeHeaderSlot epochSlots hdr
+    , bbHash   = mkByronHash hdr
     }
   where
-    computeSlot :: CC.Block.ABlockOrBoundary ByteString -> SlotNo
-    computeSlot (CC.Block.ABOBBlock blk) =
-        computeHeaderSlot epochSlots $ Right (CC.Block.blockHeader blk)
-    computeSlot (CC.Block.ABOBBoundary blk) =
-        computeHeaderSlot epochSlots $ Left (CC.Block.boundaryHeader blk)
+    hdr :: Either (CC.Block.ABoundaryHeader ByteString)
```
edsko: Abstract this out into its own function.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron.hs
edsko: Can't. Instead added comment:

```
-- NOTE: The @slotNo@ should correspond to the one that we can compute from the
-- header (using 'computeHeaderSlot') -- except we can't actually /do/ that
-- conversion here since we'd need the know @epochSlots@ for that.
```

PR #1190, opened 2019-11-04
NTP-client inline sendTo into sendPacket

MarcFontaine:
Inline the `sendTo` function into `sendPacket`. (`sendPacket` is the only user of `sendTo`).
Handle all IO Exceptions in `sendTo` and remove `SendToException`.
This changes how `sendPacket` works:
If I understand correctly, the old code does the following:
1) send IPv6 (-> suppose that works OK)
2) send IPv4 (-> suppose the throws an exception)
3) try to resend IPv6

Is that the desired behavior ?
The need to resend in the exception handler seems to be an artifact of the way exceptions are handled ?
The new code would (in the `These` case) to send to IPv6 and IPv4 exactly once,
and trace exceptions but ignore them otherwise.

### DIFF
--- a/ntp/src/Ntp/Util.hs
+++ b/ntp/src/Ntp/Util.hs
@@ -23,7 +23,7 @@ module Ntp.Util
     , pairThese
     ) where
 
-import           Control.Exception (Exception, IOException, catch, throw)
+import           Control.Exception (IOException, catch)
 import           Control.Monad (void)
 import           Control.Tracer
 import           Data.Bifunctor (Bifunctor (..))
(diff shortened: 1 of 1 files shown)


ntp/src/Ntp/Util.hs
```diff
@@ -244,27 +216,27 @@ sendPacket
...
+    fn :: ( Last (WithAddrFamily t Socket)
+          , First (WithAddrFamily t SockAddr)
+          )
+       -> IO ()
+    fn (Last sock_, First addr_) =
+        void (Socket.ByteString.sendTo (runWithAddrFamily sock_) bs (runWithAddrFamily addr_))
+            `catch` handleIOException (getAddrFamily addr_)
+
+    handleIOException :: AddrFamily -> IOException -> IO ()
+    handleIOException addressFamily e = traceWith tracer $ NtpTraceSentToIOException (show addressFamily) e
```
coot: why not use `NtpTraceSentToIOException addressFamily e` (no `show`).

ntp/src/Ntp/Util.hs
```diff
@@ -244,27 +216,27 @@ sendPacket
...
-            _                 -> traceWith tracer $ NtpTraceSentNotRetrying
+    bs :: ByteString
+    bs = LBS.toStrict $ encode $ packet
+
+    sendTo
+        :: Addresses
+        -- ^ addresses to send to
+        -> IO ()
+    sendTo addr = case pairThese sock addr of
+        Just s -> foldThese $ bimap fn fn s
```
coot: where does `foldThese` come from?  Wouldn't `foldMap` (or `bifoldMap` from `bifoldable`) just work (note `IO ()` is a monoid)

MarcFontaine:
> So there's a slight change of behaviour here right?

No there is a big change because I did not understand the original version at first place.
I hope understand some things better..
The original version would retry infinitely often while the new version never retries.
Original :
IPv6 (fail)
forever : retry IPv4.

Maybe I still have it wrong, but i think the original version first tries to
send with IPv6.. and if this succeeds it still also sends with IPv4
(which is probably not wanted)
At least that is my understanding of `Semiring (IO ())`.

ntp/src/Ntp/Util.hs
MarcFontaine: Yes..this was just left unchanged with the `foldThese` from within that module.

MarcFontaine:
IMHO there are too many things at once: The existing code, some new code, and what everybody thinks the code should do.
Maybe its an overkill but it would be nice to have a spec of the agreed-on desired behavior.
I just took the existing code as starting point but to me its unclear what it does and its also unclear
if this is what we want the code to do.

MarcFontaine:
There is is test in an other part of the code that counts the number of incoming replies.
(I don't think this is really needed)

coot:
@MarcFontaine the `IO ()` semigroup will simply do both `IO` actions synchroniously:
```
ghci> putStrLn "a" <> putStrLn "b"
a
b
```
Thus a `foldThese` will send both requests: `IPv4` and `IPv6` one in sequence. If the first `IO` throws an unhandled exception the other one will never execute.  I think `<>` is equal `>>` for `IO ()`.

Do we need to send both IPv6 and IPv4? Maybe it's enough to send one and fallback to the other if there was an error.  It would be also nice to have some retry policy: send next request after some delay.  I would suggest an exponential backoff bounded by 300s.

At the end we would like to have `3` responses from `3` different servers.  The receiver side then picks the median if I remember correctly.

MarcFontaine:
superseded by
https://github.com/input-output-hk/ouroboros-network/pull/1327

Issue #1194, opened 2019-11-06
Make sure Byron block generation is feature complete and tested


mrBliss:
In `Ouroboros.Consensus.Ledger.Byron.Forge.forgeBlock`, we forge a block that can contain transactions, delegation certificates, update votes and proposals. Since we don't have a Byron transaction generator #888, this code hasn't been properly tested yet, as we only test with empty blocks. This should be done before we let the new node forge new blocks.

mrBliss:
See more details in https://github.com/input-output-hk/ouroboros-network/issues/888#issuecomment-541607125

edsko:
WIP PR to run Byron with dual ledgers at https://github.com/input-output-hk/ouroboros-network/pull/1425 .

edsko:
Testing with delegation and update proposals will depend on https://github.com/input-output-hk/ouroboros-network/issues/1495 .

edsko:
WIP PR to generate Byron transactions at https://github.com/input-output-hk/ouroboros-network/pull/1503 .

edsko:
#1503 extends the DualPBFT tests with regular transactions. Opened a separate ticket about adding tests for update proposals in #1514, and a lower priority ticket for merging these two tests in #1515.

Issue #888, opened 2019-08-06
Generate real Byron transactions


mrBliss:
We need a `TxGen` (see #855) instance for `ByronBlockOrEBB ByronConfig`. Currently, we're not generating any transactions in the consensus tests for real Byron blocks (we are for mock blocks).

It would be great if we could reuse the existing generators from [cardano-ledger](https://github.com/input-output-hk/cardano-ledger/).

ruhatch:
@intricate do you think you could take this on? The idea was to do something similar to your UTxO tests in `cardano-ledger`, where you generate and elaborate.

mrBliss:
Notes:

* `cardano-ledger` contains [Hedgehog generators](https://github.com/input-output-hk/cardano-ledger/blob/abef065b4c4f2b7edaebed06729e6bcd7fc803a1/cardano-ledger/test/Test/Cardano/Chain/UTxO/Gen.hs) for the UTxO and transactions, but they simply generate arbitrary values for most fields. The chance that a transaction is valid w.r.t. to some UTxO is close to zero.

  Conclusion: we cannot use these generators to generate real Byron transactions.

* `cardano-ledger-specs` contains [Hedgehog generators](https://github.com/input-output-hk/cardano-ledger/blob/abef065b4c4f2b7edaebed06729e6bcd7fc803a1/cardano-ledger/test/Test/Cardano/Chain/UTxO/Gen.hs) for an *abstract* `UTxO` and *abstract* transactions. Using these generators, we can generate a valid transaction w.r.t. to some UTxO. The generated abstract transaction first has to be *elaborated* into a concrete transaction. Simply put: abstract = `cardano-ledger-specs` types, concrete = `cardano-ledger` types. The `cardano-ledger` repository contains [functions](https://github.com/input-output-hk/cardano-ledger/blob/abef065b4c4f2b7edaebed06729e6bcd7fc803a1/cardano-ledger/test/Test/Cardano/Chain/Elaboration/UTxO.hs) to perform this elaboration.

  In our situation we have a concrete UTxO (current ledger state) for which we want to generate valid and invalid concrete transactions. We have the following steps:
  ```
  concrete UTxO ---> abstract UTxO ---> abstract TX ---> concrete TX
                (1)                (2)              (3)
  ```
  1. *abstract* a concrete UTxO into an abstract UTxO: **missing**
  2. generate an abstract transaction w.r.t. an abstract UTxO: available
[...]

PR #1195, opened 2019-11-06
Fix bug in block forging: the order of the payloads was reversed

mrBliss:
When forging a block containing transactions A and B where B depends on A, we
were forging a block containing first transaction B and then A. This meant
that transaction B is invalid and, consequently, that the newly forged block
is invalid too.


mrBliss:
See #1194 and #888 for testing this bugfix.

PR #1196, opened 2019-11-06
Cleanup Byron integration

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+    updState  = CC.cvsUpdateState state
+    bodyEnv   = mkBodyEnv cfg (getProtocolParams state) (CC.blockSlot block)
+    bodyState = mkBodyState state
+
+-- | Apply a boundary block
+--
+-- We don't validate boundary blocks and so this cannot fail.
+--
+-- TODO:
+-- * Should we check that the prev hash lines up?
```
edsko: Comment: we are checking this elsewhere (in chain sync client).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+-- * Is the cvsLastSlot update compatible with what we do in consensus?
```
edsko: Yes, it is. Compatible with `abobHdrSlotNo`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+    , CC.cvsLastSlot     = CC.SlotNumber $ epochSlots * CC.boundaryEpoch hdr
+    }
+  where
+    hdr = CC.boundaryHeader blk
+    CC.EpochSlots epochSlots = Gen.configEpochSlots cfg
+
+applyScheduledDelegations :: Seq D.Sched.ScheduledDelegation
+                          -> Delegation.Map -> Delegation.Map
+applyScheduledDelegations update (Delegation.Map del) =
+    Delegation.Map $ Foldable.foldl' (flip applyOne) del update
```
edsko: Is this `foldl'` correct?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+                              (D.Sched.sdDelegate  x)
+
+{-------------------------------------------------------------------------------
+  Applying transactions
+-------------------------------------------------------------------------------}
+
+mkUtxoEnvironment :: Gen.Config
+                  -> CC.ChainValidationState
+                  -> Utxo.Environment
+mkUtxoEnvironment cfg state = Utxo.Environment {
```
edsko: Naming

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+    , U.Iface.delegationMap = delegationMap
+    }
+  where
+    k             = Gen.configK cfg
+    protocolMagic = reAnnotateMagic (Gen.configProtocolMagic cfg)
+    currentSlot   = CC.cvsLastSlot state
+    numGenKeys    = toNumGenKeys $ Set.size (allowedDelegators cfg)
+    delegationMap = getDelegationMap state
+
+    toNumGenKeys :: Int -> Word8
```
edsko: Add comment: This actually already exists in the ledger, but is not exported. When these functions are moved to `cardano-ledger`, no need to include this function.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+      | otherwise = fromIntegral n
+
+applyTxAux :: MonadError Utxo.UTxOValidationError m
+           => CC.ValidationMode
+           -> Gen.Config
+           -> Utxo.ATxAux ByteString
+           -> CC.ChainValidationState -> m CC.ChainValidationState
+applyTxAux validationMode cfg tx state =
+    (`setUTxO` state) <$>
+      runReaderT (Utxo.updateUTxO utxoEnv utxo [tx]) validationMode
```
edsko: Generalize signature to allow for list of transactions.
Use same `runReaderT` style as above.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+    utxoEnv = mkUtxoEnvironment cfg state
+    utxo    = CC.cvsUtxo state
+
+applyCertificate :: MonadError D.Sched.Error m
+                 => Gen.Config
+                 -> Delegation.ACertificate ByteString
+                 -> CC.ChainValidationState -> m CC.ChainValidationState
+applyCertificate cfg cert state =
+    (`setDelegationState` state) <$>
+      D.Iface.updateDelegation dlgEnv dlgState [cert]
```
edsko: List.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+
+reAnnotateBlock :: CC.EpochSlots -> CC.ABlock () -> CC.ABlock ByteString
+reAnnotateBlock epochSlots =
+      (\bs -> splice bs $ CBOR.deserialiseFromBytes
+                            (CC.fromCBORABlock epochSlots)
+                            bs)
+    . CBOR.toLazyByteString
+    . CC.toCBORBlock epochSlots
+  where
+    splice :: Show err
```
edsko: Generalize `splice`, then use it in tests (`Test.Consensus.Ledger.Byron`, see `annotate`).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+
+  The ledger layer defines 'ABlockOrBoundary', but no equivalent for headers.
+-------------------------------------------------------------------------------}
+
+data ABlockOrBoundaryHdr a =
+    ABOBBlockHdr    !(CC.AHeader         a)
+  | ABOBBoundaryHdr !(CC.ABoundaryHeader a)
+  deriving (Eq, Show, Functor, Generic, NoUnexpectedThunks)
+
+fromCBORABlockOrBoundaryHdr :: CC.EpochSlots
```
edsko: Comment: Verify and document what (if anything) this should be binary compatible with.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+                            -> Decoder s (ABlockOrBoundaryHdr ByteSpan)
+fromCBORABlockOrBoundaryHdr epochSlots = do
+    enforceSize "Block" 2
```
edsko: Wrong string.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Block.hs
```diff
@@ -0,0 +1,242 @@
...
+-------------------------------------------------------------------------------}
+
+newtype ByronHash = ByronHash { unByronHash :: CC.HeaderHash }
+  deriving stock   (Eq, Ord, Show, Generic)
+  deriving newtype (ToCBOR, FromCBOR, Condense)
+  deriving anyclass NoUnexpectedThunks
+
+mkByronHash :: ABlockOrBoundaryHdr ByteString -> ByronHash
+mkByronHash = \case
+    ABOBBoundaryHdr ebb -> ByronHash $ CC.boundaryHeaderHashAnnotated ebb
```
edsko: Move logic without `ByronHash` to `.Aux`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Block.hs
```diff
@@ -0,0 +1,242 @@
...
+    ABOBBlockHdr    mb  -> ByronHash $ CC.headerHashAnnotated mb
```
edsko: Rename `mb`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Block.hs
```diff
@@ -0,0 +1,242 @@
...
+
+{-------------------------------------------------------------------------------
+  Block
+-------------------------------------------------------------------------------}
+
+data ByronBlock = ByronBlock {
```
edsko: Discuss why we cache the slot number (for EBBs) and the hash (because we need it all the time and it's expensive to compute).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Block.hs
```diff
@@ -0,0 +1,242 @@
...
+-- This should be used only when forging blocks (not when receiving blocks
+-- over the wire).
+annotateByronBlock :: CC.EpochSlots -> CC.Block -> ByronBlock
+annotateByronBlock es = mkByronBlock es . CC.ABOBBlock . reAnnotateBlock es
+
+{-------------------------------------------------------------------------------
+  Header
+-------------------------------------------------------------------------------}
+
+instance GetHeader ByronBlock where
```
edsko: Comment regarding hashing (refer to `ByronBlock`).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Block.hs
```diff
@@ -0,0 +1,242 @@
...
+  blockPrevHash  = castHash . blockPrevHash . getHeader
+  blockSlot      =            blockSlot     . getHeader
+  blockNo        =            blockNo       . getHeader
+  blockInvariant = const True
+
+instance HasHeader (Header ByronBlock) where
+  blockHash      = byronHeaderHash
+  blockSlot      = byronHeaderSlotNo
+  blockPrevHash  = fromByronPrevHash' . abobHdrPrevHash        . byronHeaderRaw
+  blockNo        = fromByronBlock     . abobHdrChainDifficulty . byronHeaderRaw
```
edsko: Rename `fromByronBlock` to `fromByronBlockNo`.

edsko: (or `fromByronChainDifficulty`?)

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Ledger.hs
```diff
@@ -0,0 +1,299 @@
...
+-- /looks/ like it can fail (since its type doesn't change based on the
+-- 'ValidationMode') and we must still treat it as such.
+validationErrorImpossible :: forall err a. Except err a -> a
+validationErrorImpossible = cantBeError . runExcept
+  where
+    cantBeError :: Either err a -> a
+    cantBeError (Left  _) = error "validationErrorImpossible: unexpected error"
+    cantBeError (Right a) = a
+
+-- | Split scheduled delegations into past and future
```
edsko: Big ass comment: we deviate from the spec here! In `applyExtLedgerState`, we _first_ apply updates from the body, and _then_ do the PBFT signer check. What this means is that for us delegation updates happen _before_ the signer is checked, and so an update scheduled for slot `n` actually applies in slot `n`. In the spec however the signer is checked _before_ the body and so an update scheduled for slot `n` _really_ applies at slot `n+1`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Ledger.hs
```diff
@@ -0,0 +1,299 @@
...
+  Most of the work here is done by the ledger layer. We just need to pass
+  the right arguments, and maintain the snapshots.
+-------------------------------------------------------------------------------}
+
+applyByronBlock :: CC.ValidationMode
+                -> LedgerConfig ByronBlock
+                -> ByronBlock
+                -> LedgerState ByronBlock
+                -> Except (LedgerError ByronBlock) (LedgerState ByronBlock)
+applyByronBlock validationMode
```
edsko: Alignment.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Ledger.hs
```diff
@@ -0,0 +1,299 @@
...
+    cantBeError (Right a) = a
+
+-- | Split scheduled delegations into past and future
+splitScheduledDelegations :: WithOrigin SlotNo
+                          -> Seq D.Sched.ScheduledDelegation
+                          -> ( Seq D.Sched.ScheduledDelegation
+                             , Seq D.Sched.ScheduledDelegation
+                             )
+splitScheduledDelegations slot =
+    Seq.Lazy.spanl shouldApply
```
edsko: Remind reader of `spanl` spec.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -0,0 +1,545 @@
...
+      CC.updateBody bodyEnv bodyState block
+
+validateBlock :: MonadError CC.ChainValidationError m
+              => Gen.Config
+              -> CC.ValidationMode
+              -> CC.ABlock ByteString
+              -> CC.HeaderHash
+              -> CC.ChainValidationState -> m CC.ChainValidationState
+validateBlock cfg validationMode block blkHash state = do
+    validateHeader validationMode updState (CC.blockHeader block)
```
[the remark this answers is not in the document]
edsko: Added comment

```haskell
-- NOTE: Header validation does not produce any state changes; the only state
-- changes arising from processing headers come from 'applyEpochTransition'.
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
edsko: Modified this code to call `updateChainBoundary`, which _does_ check the hashes. Additionally, added this to `validateBLlck`:

```haskell
    -- TODO: How come this check isn't done in 'updateBlock'
    -- (but it /is/ done in 'updateChainBoundary')?
    --
    -- TODO: It could be argued that hash checking isn't part of consensus /or/
    -- the ledger. If we take that point of view serious, we should think about
    -- what that third thing is precisely and what it's responsibilities are.
    case ( CC.cvsPreviousHash state
         , unAnnotated $ CC.aHeaderPrevHash (CC.blockHeader block)
         ) of
      (Left gh, hh) ->
         throwError $ CC.ChainValidationExpectedGenesisHash gh hh
      (Right expected, actual) ->
         unless (expected == actual) $
           throwError $ CC.ChainValidationInvalidHash expected actual
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
edsko: Replaced both computations with

```
boundaryBlockSlot :: CC.EpochSlots -> Word64 -> CC.SlotNumber
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
edsko: Added comment:

```
    -- The order in which we apply the updates does not matter, because the spec
    -- says: "Any given key can issue at most one certificate in a given slot."
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
edsko: Now available as `reAnnotateUsing`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
edsko: Since this type doesn't exist in `cardano-sl`, we can basically pick whatever we want. This PR does not change the format. It must be compatible with `encodeByronHeader`, but that already has a comment (that it must be compatible with `fromCBORABlockOrBoundaryHdr`).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Ledger.hs
edsko: This comment is now part of `applyExtLedgerState`.

Issue #1188, opened 2019-11-01
Parallel tests for VolatileDB


kderme:
After https://github.com/input-output-hk/ouroboros-network/pull/1165 concurrent readers are enabled. We should also have tests taht use the api concurrently https://github.com/input-output-hk/ouroboros-network/pull/1165#pullrequestreview-309903649

PR #1212, opened 2019-11-11
Add parallel tests

kderme:
Builds on top of https://github.com/input-output-hk/ouroboros-network/pull/1180
Addresses https://github.com/input-output-hk/ouroboros-network/issues/1188
This pr adds paralle qsm tests for the VolatileDB.
It doesn't use the MockFS like sequential tests, but the real filesystem.
Only the basic api is supported for now: no simulated errors or corruptions.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -212,3 +212,17 @@ source-repository-package
   location: https://github.com/input-output-hk/cardano-crypto/
   tag: 3c707936ba0a665375acf5bd240dc4b6eaa6c0bc
   --sha256: 0g8ln8k8wx4csdv92bz09pr7v9dp4lcyv1334b09c9rgwdwhqg1b
+
+source-repository-package
+  type: git
+  location: https://github.com/advancedtelematic/quickcheck-state-machine
+  tag: 41718e9b17e3151178ce27600f84ca1d9d852c65
+
+package contra-tracer
+  tests: False
+
+constraints:
+  ip < 1.5,
+  hedgehog >= 1.0,
+  bimap >= 0.4.0,
+  primitive < 0.7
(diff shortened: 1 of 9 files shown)


mrBliss:
> It doesn't use the MockFS like sequential tests, but the real filesystem.

I have not reviewed this PR yet, but why do you use the real file system instead of the MockFS?

> Only the basic api is supported for now: no simulated errors or corruptions.

That's fine.

kderme:
> I have not reviewed this PR yet, but why do you use the real file system instead of the MockFS?

Can the MockFS detect race conditions i.e. linearizability violations?

mrBliss:
> > I have not reviewed this PR yet, but why do you use the real file system instead of the MockFS?
> 
> Can the MockFS detect race conditions i.e. linearizability violations?

I don't follow. Can the real file system detect race conditions? Does this matter? Maybe you can give an example of how things go wrong with the MockFS?

kderme:
> > > I have not reviewed this PR yet, but why do you use the real file system instead of the MockFS?
> > 
> > 
> > Can the MockFS detect race conditions i.e. linearizability violations?
> 
> I don't follow. Can the real file system detect race conditions? Does this matter? Maybe you can give an example of how things go wrong with the MockFS?

In the sequential case, we have qsm tests for the FS (these prove the equivalence of MockFS with the real filesystem) and then the volatileDB tests run on the MockFS. 
However there are currently no qsm tests for the FS in the parallel case. So having volatileDB tests on the MockFS will only say half the story.

Jasagredo:
Probably not applicable anymore, also probably will be superseeded by IOSimPOR + q-d. cc @dnadales

dnadales:
Thanks for checking @Jasagredo. @kderme I'll close this, let me know if you still want to merge it. I'm really sorry that you got no feedback :( This should not happen.

PR #1214, opened 2019-11-12
Split SupportedHeader into CanSelect/CanValidate

edsko:
I _think_ this now clears the last hurdle for the hard fork combinator (at least as far as the protocol is concerned, haven't looked at ledger integration yet).

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE ConstraintKinds            #-}
+{-# LANGUAGE DefaultSignatures          #-}
 {-# LANGUAGE DeriveGeneric              #-}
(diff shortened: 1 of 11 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -137,25 +141,21 @@ class ( Show (ChainState    p)
   --
   -- Note: we do not assume that the candidate fragments fork less than @k@
   -- blocks back.
-  preferCandidate :: HasHeader hdr
+  preferCandidate :: CanSelect p hdr
                   => NodeConfig p
                   -> AnchoredFragment hdr      -- ^ Our chain
-                  -> AnchoredFragment hdr      -- ^ Candidate
+                  -> AnchoredFragment hdr      --` ^ Candidate
```
mrBliss: Stray backtick

edsko: :roll_eyes:

Issue #398, opened 2019-03-26
Delete remaining unit test


kderme:
TagWriteInvalid from fs q-s-m tests never succeeds. So `test_hOpenWriteInvalid` is not deleted yet.

PR #1216, opened 2019-11-12
Remove unit test

kderme:
Addresses https://github.com/input-output-hk/ouroboros-network/issues/398
There was a tag `TagWriteInvalid` which never appeared. This happened because opening with ReadMode failed almost always, because the files did not exist. This is now fixed by changing the frequency: `ReadMode` happens more frequently when the file exists.

There was also a small bug in tags. The related unit test is deleted.

### DIFF
--- a/ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS.hs
+++ b/ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS.hs
@@ -11,21 +11,18 @@ import           Test.Tasty (TestTree, testGroup)
 import           Test.Tasty.HUnit
 
 import           Ouroboros.Storage.FS.API
-import           Ouroboros.Storage.FS.API.Types
 
 tests :: FilePath -> TestTree
 tests tmpDir = testGroup "HasFS"
     [ StateMachine.tests tmpDir
-      -- hOpen
-    , testCase "hOpen ReadOnly and hPut" test_hOpenWriteInvalid
     ]
 
 {------------------------------------------------------------------------------
  The tests proper
 -------------------------------------------------------------------------------}
 
-test_hOpenWriteInvalid :: Assertion
-test_hOpenWriteInvalid = apiEquivalenceFs (expectFsError FsInvalidArgument) $ \hasFS@HasFS{..} _err -> do
-    _  <- hOpen (mkFsPath ["foo.txt"]) (WriteMode MustBeNew)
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS.hs
```diff
@@ -3,6 +3,8 @@
 
 module Test.Ouroboros.Storage.FS
   ( tests
+  -- exported to silence warnings.
+  , test_example
```
mrBliss: Rename it to `_test_example` so you don't have to export it.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -475,18 +481,21 @@ generator Model{..} = oneof $ concat [
     genPath :: Gen FsPath
     genPath = do
         n <- choose (0, 3)
-        mkFsPath <$> replicateM n (elements ["x", "y", "z"])
+        mkFsPath <$> replicateM n (elements ["x", "y"])
```
mrBliss: Is it possible to tweak the frequencies in `genMode` instead of dropping `"z"`?

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -1185,11 +1194,10 @@ tag = C.classify [
           Left TagReadInvalid
         _otherwise -> Right $ tagReadInvalid openAppend
 
-    -- never succeeds, not sure why.
     tagWriteInvalid :: Set HandleMock -> EventPred
     tagWriteInvalid openRead = C.predicate $ \ev ->
       case (eventMockCmd ev, eventMockResp ev) of
-        (Open _ ReadMode, Resp (Right (WHandle _ (Handle h _)))) ->
+        (Open _ ReadMode, Resp (Right (RHandle (Handle h _)))) ->
```
mrBliss: Nice find :+1:

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
kderme: I rolled this back. I checked again and frequencies are not affected a lot by this.

PR #1220, opened 2019-11-13
Update WIP-1202-PLAN.hs

edsko:
Seems good. Left some comments inline.

### DIFF
--- a/ouroboros-consensus/test-consensus/Test/Dynamic/WIP-1202-PLAN.hs
+++ b/ouroboros-consensus/test-consensus/Test/Dynamic/WIP-1202-PLAN.hs
@@ -20,12 +20,21 @@ Assumed by the test infrastructure:
 Guaranteed by the test infrastructure:
 
   * The test infrastructure maintains the *actual* delegation map.
+    
+    EdV: This is the only part of this plan that I feel needs some clarification. 
+    What does the "actual" delegation map mean, in the context of a blockchain with 
+    potential forks? Actual according to who? How does this relate to the
+    delegation map in the chain?
 
   * The genesis chain's ledger delegation map equals the initial actual
     delegation map.
 
   * Each ledger delegation map is either synchronized with the actual
     delegation map or else out-of-date.
+    
+    EdV: Not sure I understand "out of date". This and the point I commented on 
+    above suggest that history is linear, and nodes may merely be "behind" on
+    updates. But what if they disagree?
 
   * When the test infrastucture changes the actual delegation map, it also
     promptly generates the necessary delegation certificate transactions to
@@ -68,5 +77,21 @@ Concern:
 
 It seems like the new node may need to persistently re-generate that
 transaction until it finds it in its immutable DB.
+
+EdV: So these are not test concerns, these are concerns inherent in delegating.
+Indeed, a wallet node would do something similar to this (monitor the chain
+to see if its transactions have been included, and resubmit after a while). 
+I'm not sure that we should include this ability in the node right now; let's
+perhaps open a ticket for that particular subproblem. 
+
+EdV: One thing that is useful to realize is that a node _can_ safely start
+producing blocks as soon as a certificate has been included on the chain; 
+it doesn't need to wait. After all, other nodes that synchronize that same
+chain will see the certificate before the new blocks, and so the certificate
+will come first; if they are on a different chain, they won't see either.
+
+EdV: Note however that the spec limits when delegation certificates can take
+effect, precisely to address some chain stability problems. See the Byron
+ledger specification. 
 
 -}


ouroboros-consensus/test-consensus/Test/Dynamic/WIP-1202-PLAN.hs
```diff
@@ -20,12 +20,21 @@ Assumed by the test infrastructure:
 Guaranteed by the test infrastructure:
 
   * The test infrastructure maintains the *actual* delegation map.
+    
+    EdV: This is the only part of this plan that I feel needs some clarification. 
+    What does the "actual" delegation map mean, in the context of a blockchain with 
+    potential forks? Actual according to who? How does this relate to the
+    delegation map in the chain?
```
mrBliss: I had the same question. I think the following is meant: the delegation map the honest core node operators want to be in place. The delegation map in the chain will lag behind. So this indeed needs some clarification.

edsko: So are we going to assume the 7 core node operators are going to off-chain coordinate these changes? It's not an unreasonable assumption, perhaps; though perhaps it is, I'm not entirely sure. But if they don't coordinate, two core node operators may both independently delegate, at which point the "active" map  would be the map with both changes applied. But how "active" is already a lot murkier.

nfrisby: > I think the following is meant: the delegation map the honest core node operators want to be in place. The delegation map in the chain will lag behind. 

Precisely right: It's the _actual_ map in the sense that it characterize the _actual_ delegations desired by each genesis key owner (i.e. some human being). And he's also right that the chains will lag behind that until they include the most recent delegation certificates.

I'll try to make that clearer.

> So are we going to assume the 7 core node operators are going to off-chain coordinate these changes?

Do multiple operators have to coordinate in order to decide to whom a particular genesis key has actually been delegated? I'd assumed that at most one node operator "owns" each genesis key.

Maybe "desired" instead of "actual" would be clearer?

> But if they don't coordinate, two core node operators may both independently delegate, at which point the "active" map would be the map with both changes applied. But how "active" is already a lot murkier.

I'm assuming you meant _actual_ not _active_. Related to my previous question: two node operators cannot "independently delegate" the same genesis key, can they? If not, then I think the notion of the "actual"/"desired" map seems easy to maintain.

edsko: No, they would only make decisions on their own genesis keys, but nonetheless, that means that two operators can independently make changes to the delegation map, which means that the concept of a single global "active" delegation map may not exist, right?

nfrisby: It seems that we're considering two different things. I haven't made any claims about what I think you mean by the "active" delegation map. I agree that there can be different delegation maps on different chains in the net.

I'm so far only concerned with an off-chain record of the latest intentions of all genesis key owners -- I called that the "actual" delegation map because it contains the latest delegations. It's clear now that the word "actual" is confusing. So I'm going to start writing "intended" delegation map instead. Does that make sense?

-----

Basic idea: The test infrastructure maintains an _intended delegation map_ that evolves over time, and it promptly issues delegation certificates when that map changes so that the nets' chains' ledgers' delegation maps will eventually be updated to equal the latest intended delegation map.

If we further assume that the test infrastructure immediately shuts down a node before it removes that node's relevant mapping in the intended delegation map, then we will ensure the invariant that we had previously decided was desirable:

```
  * Invariant: For all genesis keys @gk@, no two node instances simultaneously
    have a current ledger that maps @gk@ to their own operational key.
```

edsko: I still find that confusing. I guess the confusion comes from the fact that there _isn't_ a _single_ intended/actual delegation map. Each node might have its own, no? If two nodes each make changes to the delegation map (which will become active at some point in the future), what exactly is _the_ intended delegation map?

nfrisby:
I didn't do this in the simplest way, so I'll summarize here.

Merging this PR resulted in:

```
be4377a2 - (origin/nfrisby/issue-1202-gen-delegations) Merge pull request #1220 from input-output-hk/edsko-patch-1
```

But I had local implementation changes and I integrated your changes and our most recent discussions into the document without merging/rebasing to include this PR's commit. The most recent tip is now:

```
efa2c18c - (HEAD -> nfrisby/issue-1202-gen-delegations, origin/nfrisby/issue-1202-gen-delegations) WIP update PLAN
```

Issue #1088, opened 2019-10-01
Resolver test fails on Windows 10


coot:
```
  Subscription
    Resolve Subscribe (IO): FAIL (1.41s)
      *** Failed! Falsified (after 14 tests):
      LookupResultIO: ipv4: Right [2,0,3,1] ipv6: Right [3,1,2] first AF_INET valency 7
      Use --quickcheck-replay=598265 to reproduce.
```

karknu:
Fixed by #1222 .

PR #1222, opened 2019-11-13
Increase grace time when waiting for A result.

karknu:
Increase the time spent waiting on A result from 10ms to
100ms to ensure that the thread waiting for the A result has had a
chance to run. This matches the time we wait for AAAA results.


karknu:
Fixes #1088 .

PR #1227, opened 2019-11-14
Pass only headers to chainsel

edsko:


nix/.stack.nix/default.nix
```diff
@@ -50,6 +50,7 @@
         cardano-prelude = ./cardano-prelude.nix;
         cardano-prelude-test = ./cardano-prelude-test.nix;
         cardano-crypto = ./cardano-crypto.nix;
+        cardano-crypto = ./cardano-crypto.nix;
```
mrBliss: Is this change needed?

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -277,12 +278,12 @@ initBlockFetchConsensusInterface cfg chainDB getCandidates blockFetchSize
...
                             -> AnchoredFragment (Header blk)
                             -> Bool
-    plausibleCandidateChain = preferCandidate cfg
+    plausibleCandidateChain = preferAnchoredCandidate cfg
 
     compareCandidateChains :: AnchoredFragment (Header blk)
                            -> AnchoredFragment (Header blk)
                            -> Ordering
-    compareCandidateChains = compareCandidates cfg
+    compareCandidateChains = compareAnchoredCandidates cfg
```
mrBliss: (Not necessarily in this PR:) could the BlockFetch client call our `preferCandidate` and `compareCandidates` directly? Maybe that might simplify the BFC implementation? See https://github.com/input-output-hk/ouroboros-network/blob/a7b3055c1a4ccc26455939910a6283d8947c108e/ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs#L136

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -226,16 +229,15 @@ class ( Show (ChainState    p)
 
   default preferCandidate :: HasHeader hdr
                           => NodeConfig p
-                          -> AnchoredFragment hdr      -- ^ Our chain
-                          -> AnchoredFragment hdr      --` ^ Candidate
+                          -> hdr      -- ^ Our chain
+                          -> hdr      --` ^ Candidate
```
mrBliss: Stray backtick (was already there)

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -226,16 +229,15 @@ class ( Show (ChainState    p)
...
                           -> Bool
-  preferCandidate _ ours cand =
-    AF.compareHeadBlockNo cand ours == GT
+  preferCandidate _ ours cand = blockNo cand > blockNo ours
```
mrBliss: Ah, this is nice :relieved:

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -137,29 +136,33 @@ class ( Show (ChainState    p)
...
+  -- consensus layer is that if a chain can be extended, it always should (e.g.,
+  -- see the chain database spec in @ChainDB.md@). This means that any chain
+  -- is always preferred over the empty chain, and 'preferCandidate' does not
+  -- need (indeed, cannot) be called if our current chain is empty.
   preferCandidate :: CanSelect p hdr
                   => NodeConfig p
-                  -> AnchoredFragment hdr      -- ^ Our chain
-                  -> AnchoredFragment hdr      -- ^ Candidate
+                  -> hdr      -- ^ Our chain
+                  -> hdr      -- ^ Candidate
```
mrBliss: Our chain -> Header at our chain's tip? Or something like that

nix/.stack.nix/default.nix
edsko: You are reviewing nix files now..? :D It was a change produced by the stack2nix script from CI in https://github.com/input-output-hk/ouroboros-network/pull/1380 :man_shrugging:

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
edsko: I don't follow?

mrBliss: It's about the comment

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
nfrisby: That will conflict with PR 1353, which we're hoping to merge soon and currently has

```
-  preferCandidate _ ours cand =
-    AF.compareHeadBlockNo cand ours == GT
+  preferCandidate cfg ours cand =
+    compareCandidates cfg ours cand == LT
```

(Note the unflip of the arguments and the switch to `compareCandidates`.)

edsko: I see. Not a fundamental problem though is it? The equivalent of that code can still be done with this new approach also?

nfrisby: They're compatible. I was just calling out an anticipated `git` conflict, but forgot to add "`git`".

edsko: Ok, good :)

nix/.stack.nix/default.nix
edsko: Fixed in #1380.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
edsko: https://github.com/input-output-hk/ouroboros-network/issues/1382

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
edsko: Ah, gotcha. Done.

PR #1238, opened 2019-11-18
Remove `tip` argument to ChainSyncClient

edsko:
This also removes ExampleTip (which was isomorphic to Tip).


edsko:
@coot approved on slack.

bors r+

PR #1240, opened 2019-11-18
Remove type arguments from error types

edsko:
NOTE:  "Remove `tip` argument to ChainSyncClient" is #1238.

This commit paves the way for consensus error policies. It removes type parameters from exception types, so that we cannot introduce bugs where we miss certain exceptions because due to (type) matching on exceptions of the wrong type (e.g., instantiated to header instead of block). 

It also removes `onSlot` from `BlockchainTime`. This was used only in tests, where it is easily replaced.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime.hs
@@ -12,7 +12,6 @@ module Ouroboros.Consensus.BlockchainTime (
     BlockchainTime(..)
   , blockUntilSlot
   , onSlotChange
-  , onSlot
     -- * Use in testing
   , NumSlots(..)
   , TestBlockchainTime(..)
@@ -33,7 +32,6 @@ module Ouroboros.Consensus.BlockchainTime (
   , SlotNo (..)
   ) where
 
-import           Control.Exception (Exception (..))
 import           Control.Monad
 import           Data.Fixed
 import           Data.Time.Clock (NominalDiffTime, addUTCTime, diffUTCTime)
@@ -41,8 +39,6 @@ import           Data.Void
 import           Data.Word (Word64)
 import           GHC.Generics (Generic)
 import           GHC.Stack
-
-import           Control.Monad.Class.MonadThrow
 
 import           Cardano.Prelude (NoUnexpectedThunks)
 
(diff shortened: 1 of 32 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime.hs
```diff
@@ -106,38 +91,6 @@ blockUntilSlot btime slot = do
...
-                   else Nothing
-
-data OnSlotException =
-    -- | An action was scheduled via 'onSlot' for a slot in the past.
-    -- First slot is requested, second slot is current as of raising.
-    OnSlotTooLate SlotNo SlotNo
-  deriving (Eq, Show)
-
-instance Exception OnSlotException
-
```
mrBliss: @nfrisby I remember you added this. What do you think about this change?

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -841,93 +841,151 @@ continueWithState !s (Stateful f) = checkInvariant (unsafeNoThunks s) $ f s
...
+    case eqT @blk @blk' of
+      Nothing   -> False
+      Just Refl -> (a, b, c, d) == (a', b', c', d')
+  DoesntFit{} == _ = False
+
+  InvalidBlock (a :: Point blk) == InvalidBlock (a' :: Point blk') =
+    case eqT @blk @blk' of
+      Nothing   -> False
+      Just Refl -> a == a'
+  InvalidBlock{} == _ = False
```
mrBliss: :scream:

ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime.hs
edsko: (If the exception is needed, we can re-introduce it in the test.)

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
edsko: I know :(

edsko:
Yes, will do. Once 1238 is merged, will rebase and reintroduce the exception in the consensus tests.

ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime.hs
edsko: Rebased, and moved this to `Test.Util.BlockchainTime` where it is now defined as

```haskell
onSlot :: (HasCallStack, IOLike m) => BlockchainTime m -> SlotNo -> m () -> m ()
onSlot btime slot k = do
    startingSlot <- atomically $ getCurrentSlot btime
    when (startingSlot >= slot) $
      throwM $ OnSlotTooLate slot startingSlot
    onSlotChange btime $ \slot' ->
      when (slot == slot') k

data OnSlotException =
    -- | An action was scheduled via 'onSlot' for a slot in the past.
    -- First slot is requested, second slot is current as of raising.
    OnSlotTooLate SlotNo SlotNo
  deriving (Eq, Show)

instance Exception OnSlotException
```

nfrisby: Moving it is no problem. Thanks for the notice.

PR #1241, opened 2019-11-18
Consensus error policy

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/Node/ErrorPolicy.hs
```diff
@@ -0,0 +1,152 @@
...
+              InvalidRollBack{}        -> Just theyBuggyOrEvil
+              InvalidIntersection{}    -> Just theyBuggyOrEvil
+              NoMoreIntersection{}     -> Just distantPeer
+              DoesntFit{}              -> Just theyBuggyOrEvil
+              InvalidBlock{}           -> Just theyBuggyOrEvil
+
+          -- Dispatch on nested exception
+        , ErrorPolicy $ \(ExceptionInLinkedThread _ e) ->
+            evalErrorPolicies e (epAppErrorPolicies consensusErrorPolicy)
+        ]
```
coot: I am not familiar with all the exceptions in the `ouroboros-consensus`.  So I grepped, what about:
* `OnSlotException` which can be raised by `onSlot` function
* `BlockFetchServerException` defined in `Ouroboros.Consensus.BlockFetchServer`

Just a note to myself: we also need to take care about `ouroboros-network` exceptions:
* `ClientProtocolError` - thrown as an application exception by the `handshake` protocol
* `TxSubsmissionProtocolError`s (inbound and outbound)
* `BlockFetchProtocolFailure`

edsko: `OnSlotException` I moved to the tests so it's not a concern. `BlockFetchServerException` we do handle:

```haskell
        , ErrorPolicy $ \(_ :: BlockFetchServerException) -> Just distantPeer
```

PR #1252, opened 2019-11-19
network error policies

coot:


ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
```diff
@@ -386,3 +389,31 @@ ncSubscriptionWorker_V1
...
+        -- version or we refused it.  This is only for outbound connections to
+        -- a local node, thus we throw the exception.
+        ErrorPolicy
+          $ \(_ :: HandshakeClientProtocolError NodeToClientVersion)
+                -> Just ourBug
+      ]
+    , epConErrorPolicies = [
+        -- If an 'IOException' is thrown by the 'connect' call, kill the
+        -- application.
+        ErrorPolicy $ \(_ :: IOException) -> Just ourBug
```
karknu: Does this mean that a temporary connection error experienced by `ncSubscriptionWorker` will cause to node to restart?
I would expect this to be a short timeout.

coot: This will kill the node if the `connect` throws an exception. So yes, if there was a network failure, and the node tries to connect and the failure still persists it will halt the node.

Also I am not handling any `IOExceptions` in the `epAppErrorPolicies` and we should, for the reason you mentioned.

coot: I will modify the policy, and justify it in a haddock.

coot:
I also need to add `Mux` exceptions to the `networkErrorPolicies`.

coot:
Thanks @karknu 

bors r+

PR #1261, opened 2019-11-20
Tune error policies

coot:


coot:
I dropped the `dns` related patch - it seems it was a problem with my `cabal` environement.

PR #1262, opened 2019-11-20
Fix accept loop exception handling

coot:


edsko:
@coot Cancelled bors because the tests seem to loop with this PR, causing Hydra to time out and bors to stall all other PRs. Sorry.

ouroboros-network/src/Ouroboros/Network/Socket.hs
```diff
@@ -485,5 +485,4 @@ withServerNode muxTracer handshakeTracer errorPolicyTracer tbl stVar addr encode
 
       acceptException :: Socket.SockAddr -> SomeException -> IO ()
       acceptException a e = do
-        traceWith (WithAddr a `contramap` errorPolicyTracer) $ ErrorPolicyAccept e
-        throwIO e
+        traceWith (WithAddr a `contramap` errorPolicyTracer) $ ErrorPolicyAcceptException e
```
avieth: Must rethrow async exceptions, or at the very least, `AsyncCancelled` and `ThreadKilled` exceptions.

avieth: This is the exception handler around `Network.Socket.accept` right? I'd recommend only catching `IOExceptions` here.

coot: Yes, we should only catch `IOException`s here.

PR #1267, opened 2019-11-21
WIP Add ncSubscriptionWorker_V2

erikd:
This has a single parameter, a struct, with field names, which makes it much easier to fix the type errors when the struct changes.

This is not a serious PR, but an example of how using a struct would make type errors far easier to fix when the struct changes.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
@@ -21,8 +21,10 @@ module Ouroboros.Network.NodeToClient (
   , withServer_V1
   , withServer
 
+  , SubscriptionWorkerV2 (..)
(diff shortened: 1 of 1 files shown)


erikd:
At the call site, this can be called as:
```
  ncSubscriptionWorker_V2data
    SubscriptionWorkerV2
      { sw2SubscriptionTrace = ..
      , sw2MuxBearerTrace = ..
      , sw2SendRecvTrace = ...
      , sw2AddrTrace = ..
       ..
      }

coot:
@erikd thanks for the PR. I'd like to divided it into small number of related records that will keep the number of arguments small (let say <=5), but will make use of the api easier. There are things that can be shared between various functions.

I'll close this PR if you don't mind (your contribution is appreciated!) and file a new PR and tag you for review.

PR #1273, opened 2019-11-21
Simplify the anachronistic ledger view

### LINKED ISSUE #581: Chain Sync Client: reuse anachronistic ledger view?
We currently ask for the anachronistic ledger view for every block we get. We should reconsider that. If that is indeed okay, then there is no need for the window at all, and so we can simplify the code. If that is not okay (bad performance) we should change the chain sync client.

edsko:
This closes #581.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Ledger.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Ledger.hs
@@ -29,7 +29,6 @@ import           Control.Monad.Except
 import           Data.ByteString (ByteString)
 import           Data.Sequence (Seq)
 import qualified Data.Sequence as Seq.Lazy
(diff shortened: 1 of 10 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Ledger.hs
```diff
@@ -159,40 +156,18 @@ instance ProtocolLedgerView ByronBlock where
   -- i.e., that delegate is allowed to issue a block in that very same slot.
   anachronisticProtocolLedgerView cfg (ByronLedgerState ls ss) slot =
```
mrBliss: Should the docstring be updated?

edsko: No, I think it's still accurate. The same bounds still apply. It's just that we return something from within those windows.

Issue #1254, opened 2019-11-19
Record validation error in InvalidBlock error in chain sync client


edsko:

nfrisby:
I just noticed this the other day as well. On my Issue 1202 branch, I'm now recording these via a `Tracer` and the ultimate `Property` is checking that it's empty (or only as-expected for tests that we expect to create invalid blocks -- eg due to stale delegation).

What do you mean by "record"? Apply a `tracer`? Raise an exception?

mrBliss:
Addressed in #1279.

Issue #1225, opened 2019-11-14
Remove transactions that cause invalid blocks to be created


mrBliss:
If there's an inconsistency between the validation we use for the transactions in the mempool and the validation we use for the block we forged containing the transactions from the mempool, we end up making no progress and we get stuck. The mempool won't remove this transactions from the mempool, as they are valid and not in the current chain, but we cannot adopt the block as it is deemed invalid.

Note that #622 will eventually fix this.

Another idea: after adding a newly forged block to the ChainDB, check whether its point ends up in the set of invalid points. This can have false negatives: another chain was selected so our block was not validated yet. If the block point occurs in the invalid points set, we log it since it clearly is a bug and we purge all transactions in the invalid from the mempool.

njd42:
this is a good idea. It will be the case (eventually) - even when the code is "perfect" where environmental factors will damage internal state (http://dance.csc.ncsu.edu/papers/ic2e2015.pdf captures such). Ensuring that systemic progress will be made in this eventually will be important.

Issue #787, opened 2019-07-17
Trace adding a self-produced block to the ChainDB


mrBliss:
Quoting from https://github.com/input-output-hk/ouroboros-network/pull/777#pullrequestreview-262410795

> And then idependently, make sure we have a TODO or ticket to improve the tracing around this use of
>
>        ChainDB.addBlock chainDB newBlock
>
> We should use structured tracing of course, and we should distinguish before/after, and in the after case we should include whether the block was added to the tip of the current chain or whether the race condition occured.

The code in question: https://github.com/input-output-hk/ouroboros-network/blob/74b5c33fc105eca2292177dda9c6ad6058bf876c/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L352

PR #1279, opened 2019-11-22
ChainSyncClient: include the reason why a block is invalid

mrBliss:
Addresses #1254.

The ChainSyncClient will disconnect from a peer when it sees a header in the
candidate chain that it knows corresponds to an invalid block. How does it
know it is invalid? The block has been downloaded in the past, the ChainDB has
validated the corresponding block, found it to be invalid, traced this along
with the reason why it was invalid, and recorded that the block is invalid.

However, many people are wondering why the ChainSyncClient thinks a block is
invalid, as it doesn't include the reason in its exception/trace message. Note
that the ChainDB **must** have already traced the reason why that block was
invalid, but that trace was maybe not enabled or people don't bother to scroll
up.

So to make all our lives easier, we now remember the exact reason why a block
is invalid in the ChainDB, instead of merely remembering the fact that a block
was invalid. This allows us to include the exact reason in the exception/trace
message why a ChainSyncClient thinks a block is invalid.


mrBliss:
This PR also makes it possible to have better trace messages in #787 and #1225.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Orphans.hs
```diff
@@ -123,3 +130,16 @@ instance Condense (CC.AMempoolPayload a) where
...
+-------------------------------------------------------------------------------}
+
+deriving instance Generic CC.ChainValidationError
+
+-- TODO a proper instance.
+--
+-- Problem: Cardano.Chain.Delegation.Validation.Registration.TooLarge is not
+-- exported, but occurs somewhere in CC.ChainValidationError.
+deriving via (OnlyCheckIsWHNF "ChainValidationError" CC.ChainValidationError)
+  instance NoUnexpectedThunks CC.ChainValidationError
```
mrBliss: https://github.com/input-output-hk/cardano-ledger/issues/683

edsko: Use `UseIsNormalForm`.

Issue #1150, opened 2019-10-21
Handle clock skew when producing a new block


mrBliss:
When the node tries to produce a new block:
https://github.com/input-output-hk/ouroboros-network/blob/5bf6c1a3ca15f40dac50913d9168916ebeb62232/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs#L346
It throws an `error` when the current chain's tip is a block with a slot that is in the future. However, this can happen in practice, as we allow for some [clock skew](https://github.com/input-output-hk/ouroboros-network/blob/5bf6c1a3ca15f40dac50913d9168916ebeb62232/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs#L131).

We should handle this gracefully. We can ignore the block that is in the future so that *we* can produce our block in this slot, the slot of which we are the leader. Or, we could forfeit our leadership of the slot, but I think this is not a good idea, as it would open the door for attacks.

What should we do if it is still too far in the future, even while taking the clock skew into account? This should be impossible because the ChainSync client will reject headers that are too far in the future (taking the same clock skew into account). But what if we start a node with a wrong clock such that our initial chain loaded from disk is already too far in the future? See #759 and #1143.

edsko:
The problem here is not that we should handle this case: we should never adopt blocks that are from the future (chain sync client might _store_ them, but the chain DB should not _adopt_ them). This should be fixed in the chain DB, not here.

mrBliss:
Superseded by #1287.

PR #1281, opened 2019-11-22
Block forging improvements

mrBliss:
~This builds further upon #1279 and thus includes it for now.~ *Update: merged*

Addresses #787: trace whether we adopted or block or not.

Addresses #1225: if we produced an invalid block, we remove all transactions
in it from the Mempool.

~Addresses #1150: handle clock skew when producing a new block~ *Update: this will be handled by the ChainDB in a separate PR*

It also includes some minor improvements/fixes that I made along the way.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
@@ -19,12 +19,14 @@ import           Ouroboros.Network.TxSubmission.Inbound
 import           Ouroboros.Network.TxSubmission.Outbound
                      (TraceTxSubmissionOutbound)
 
-import           Ouroboros.Consensus.Block (Header, SupportedBlock)
+import           Ouroboros.Consensus.Block (Header)
 import           Ouroboros.Consensus.BlockFetchServer
                      (TraceBlockFetchServerEvent)
-import           Ouroboros.Consensus.ChainSyncClient (TraceChainSyncClientEvent)
+import           Ouroboros.Consensus.ChainSyncClient (InvalidBlockReason,
+                     TraceChainSyncClientEvent)
 import           Ouroboros.Consensus.ChainSyncServer (TraceChainSyncServerEvent)
-import           Ouroboros.Consensus.Ledger.Abstract (AnachronyFailure)
+import           Ouroboros.Consensus.Ledger.Abstract (AnachronyFailure,
+                     ProtocolLedgerView)
 import           Ouroboros.Consensus.Mempool.API (ApplyTxErr, GenTx, GenTxId,
                      TraceEventMempool)
 import           Ouroboros.Consensus.TxSubmission
@@ -73,7 +75,7 @@ showTracers :: ( Show blk
                , Show (ApplyTxErr blk)
                , Show (Header blk)
                , Show peer
-               , SupportedBlock blk
+               , ProtocolLedgerView blk
                )
             => Tracer m String -> Tracers m peer blk
 showTracers tr = Tracers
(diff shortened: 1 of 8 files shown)


mrBliss:
That's one low priority and two high priorities in one PR :sunglasses:.

I didn't split them up into separate PRs because they are all related to one piece of code, so that would just cause merge conflicts.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
```diff
@@ -105,4 +107,15 @@ data TraceForgeEvent blk
   -- 'anachronisticProtocolLedgerView', although we expect this to be
   -- 'TooFarAhead', never 'TooFarBehind'.
   | TraceCouldNotForge SlotNo AnachronyFailure
+
+  -- | We adopted the block we produced
+  | TraceAdoptedBlock SlotNo blk
+
+  -- | We did not adopt the block we produced
```
mrBliss: Say that this is very rare, i.e. warrants a warning.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
```diff
@@ -105,4 +107,15 @@ data TraceForgeEvent blk
...
+  | TraceDidntAdoptBlock SlotNo blk
+
+  -- | We forged a block that is invalid according to the ledger in the
+  -- ChainDB. This means there is an inconsistency between the mempool
+  -- validation and the ledger validation.
+  | TraceForgedInvalidBlock SlotNo blk (InvalidBlockReason blk)
```
mrBliss: Make it clear that this is an error

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -347,11 +350,32 @@ forkBlockProduction IS{..} =
...
-          trace $ TraceForgeEvent currentSlot newBlock
-          ChainDB.addBlock chainDB newBlock
         FailedToProduce err ->
           trace $ TraceCouldNotForge currentSlot err
+        ProducedBlock newBlock txs -> do
+          trace $ TraceForgeEvent currentSlot newBlock
+          -- Adding a block is synchronous
+          ChainDB.addBlock chainDB newBlock
+          -- Check whether we adopted our block
+          curTip <- atomically $ ChainDB.getTipPoint chainDB
```
mrBliss: Mention in its docstring that `addBlock` will run chain selection.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -347,11 +350,32 @@ forkBlockProduction IS{..} =
...
+          if curTip == blockPoint newBlock then
+            trace $ TraceAdoptedBlock currentSlot newBlock
+          else do
+            trace $ TraceDidntAdoptBlock currentSlot newBlock
+            isInvalid <- atomically $
+              ($ blockHash newBlock) . forgetFingerprint <$>
+              ChainDB.getIsInvalidBlock chainDB
+            whenJust isInvalid $ \reason -> do
```
mrBliss: Do this check first, if it is invalid, trace `TraceForgedInvalidBlock`, it is is valid, trace `TraceDidntAdoptBlock` (include in the documentation of this message that it is only traced when the block is valid but not adopted).

ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
edsko: We can even be a bit more specific maybe? At least indicate it's not an error either: the block we produced was _valid_ (at least after the changes to this PR), but chain selection decided not to adopt it (so it must have found a better -- longer -- chain, probably due to a concurrent call to `addBlock`?).

mrBliss:
Rebased on master and addressed the feedback.

PR #1285, opened 2019-11-24
Refactoring of Ouroboros-Network API

coot:
This series of patches refactors ouroboros-network APIs.  Instead of #1267 the refactoring goes deep into the ouroboros-network which will simplify maintnence costs for the networking team.

* provides ADTs which simplify subscription workers
* provides `ConnectionId` which removes `peerid` abstraction leak - I am quite happy that it was possible to do without deep changes in `ourboros-consensus`
* using `Void` instead of `forall void. void` - which has the effect on `Ouroboros.Consensus.Node`, where we don't need to wait on threads which return `Void`.
* in `Ouroboros.Consensus.Node` the `NetworkMutableState ~ (ConnectionTable, PeerStates)` is shared between local and remote peers.

The PR is a series of patches suitable for review patch by patch.


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -320,16 +321,20 @@ initNetwork registry nodeArgs kernel RunNetworkArgs{..} = do
...
          runDnsSubscriptionWorker networkState ipv4Address ipv6Address dnsProducer
 
-    let threads = localServer : ipSubscriptions : cleanPeerStatesThread : dnsSubscriptions ++ peerServers
-    void $ waitAnyThread threads
+    -- the only thread that terminates is 'cleanNetworkStateThread', all the
+    -- other threads return `Void`, thus they can only throw, since all of them
+    -- are linked to this thread we can just wait on the 'cleanNetworkStateThread'
+    -- Also note that 'cleanNetworkStateThread' terminates only if one of the
+    -- other threads throws an exception.
+    void $ waitThread cleanNetworkStateThread
```
coot: @mrBliss are you happy with this?

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -342,9 +344,9 @@ initNetwork registry nodeArgs kernel RunNetworkArgs{..} = do
     nodeToNodeVersionData   = NodeToNodeVersionData { networkMagic   = rnaNetworkMagic }
     nodeToClientVersionData = NodeToClientVersionData { networkMagic = rnaNetworkMagic }
 
-    runLocalServer :: IO ()
-    runLocalServer = do
-      networkState <- newNetworkMutableState
+    runLocalServer :: NetworkMutableState
+                   -> IO ()
+    runLocalServer networkState = do
```
coot: @karknu I think it's ok to share the `NetworkMutableState ~ (ConnectionTable, PeerStates)` between local and remote peers.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -345,7 +350,7 @@ initNetwork registry nodeArgs kernel RunNetworkArgs{..} = do
     nodeToClientVersionData = NodeToClientVersionData { networkMagic = rnaNetworkMagic }
 
     runLocalServer :: NetworkMutableState
-                   -> IO ()
+                   -> IO Void
```
erikd: \0/ !!!!!!

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -391,7 +392,7 @@ dnsSubscriptionWorker_V1
           ConnectionId
           NodeToNodeProtocols
           IO BL.ByteString x y)
-    -> IO void
+    -> IO Void
```
erikd: \o/ !!!!!!

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -87,21 +87,18 @@ import           Ouroboros.Storage.LedgerDB.InMemory (ledgerDbDefaultParams)
...
+  => Tracers IO ConnectionId blk          -- ^ Consensus tracers
   -> Tracer  IO (ChainDB.TraceEvent blk)  -- ^ ChainDB tracer
-  -> RunNetworkArgs peer blk              -- ^ Network args
+  -> RunNetworkArgs ConnectionId blk      -- ^ Network args
   -> FilePath                             -- ^ Database path
   -> ProtocolInfo blk
   -> (ChainDbArgs IO blk -> ChainDbArgs IO blk)
       -- ^ Customise the 'ChainDbArgs'
-  -> (NodeArgs IO peer blk -> NodeArgs IO peer blk)
+  -> (NodeArgs IO ConnectionId blk -> NodeArgs IO ConnectionId blk)
```
erikd: I like seeing type vars replaced with concrete types.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -297,24 +294,25 @@ data RunNetworkArgs peer blk = RunNetworkArgs
...
+    _ <- forkLinkedThread registry (runLocalServer networkState)
 
     -- serve downstream nodes
-    connTable  <- newConnectionTable
-    peerStatesVar <- newPeerStatesVar
-    -- clean peer states every 200s
-    cleanPeerStatesThread <- forkLinkedThread registry (NodeToNode.cleanPeerStates 200 peerStatesVar)
-    peerServers <- forM rnaMyAddrs
-        (\a -> forkLinkedThread registry $ runPeerServer connTable peerStatesVar a)
+    _ <- forM rnaMyAddrs
```
erikd: `forM_` ?

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -323,21 +321,25 @@ initNetwork registry nodeArgs kernel RunNetworkArgs{..} = do
...
                          else Nothing
 
-    ipSubscriptions <- forkLinkedThread registry $
-                         runIpSubscriptionWorker connTable peerStatesVar ipv4Address ipv6Address
+    _ <- forkLinkedThread registry $
+                         runIpSubscriptionWorker networkState ipv4Address ipv6Address
 
     -- dns subscription managers
-    dnsSubscriptions <- forM rnaDnsProducers $ \dnsProducer -> do
+    _ <- forM rnaDnsProducers $ \dnsProducer -> do
```
erikd: `forM_` ?

ouroboros-network/demo/chain-sync.hs
```diff
@@ -84,22 +84,22 @@ main = do
       "pingpong":"client-pipelined":[] -> clientPingPong True
       "pingpong":"server":[] -> do
         rmIfExists defaultLocalSocketAddrPath
-        serverPingPong
+        void serverPingPong
```
erikd: Other places use:
```
    _ <- action
```
I personally prefer `void`

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
mrBliss: That sounds ok. It would be even better if this logic would be somewhere in `Ouroboros.Network` instead of in `Ouroboros.Consensus`.

In fact, I think that `Ouroboros.Consensus.Node` still contains waaaay too much low-level networking setup (I moved it there from `cardano-node`) that should live in `Ouroboros.Network`.

Ideally, there would just be a single function that I need to call to start a server/client, instead of having to manually start multiple worker threads. They should all be internal to `Ouroboros.Network`. Changes in the network stack should not affect `Ouroboros.Consensus` at all.

I think `RunNetworkArgs` and the network-related functions in this file should move to `Ouroboros.Network`.

cabal.project
```diff
@@ -1,13 +1,25 @@
...
-          ./io-sim-classes
-          ./ntp-client
-          ./Win32-network
+	  ./network-mux
+	  ./ouroboros-network
+	  ./ouroboros-consensus
+	  ./io-sim
+	  ./io-sim-classes
+	  ./ntp-client
+	  ./Win32-network
```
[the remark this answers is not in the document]
coot: ups, I'll fix this

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
coot: This PR is already quite big, I'll address this in a follow up PR.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
karknu: I don't think we should do this. It would introduce an unnecessary synchronization point between services that should have nothing to do with each other.
There is a pre-existing problem with the connection table, we track the connection but we don't know if it was negotiated for NodeToNode or NodeToClient.

cabal.project
```diff
@@ -1,13 +1,13 @@
...
-	  ./io-sim-classes
-	  ./ntp-client
-	  ./Win32-network
+          ./network-mux
+          ./ouroboros-network
+          ./ouroboros-consensus
+          ./io-sim
+          ./io-sim-classes
+          ./ntp-client
+          ./Win32-network
```
mrBliss: I would squash this change into the right commit, so that we're not needlessly tainting the history of this file :slightly_smiling_face: But I guess you were planning on doing that before merging it.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
coot: > All local clients are expected to run on a unix socket - though we can run remote local clients.
Ok, if we want to share the `ConnectionTable` we need to make this explicit.  I wouldn't be worried about to much synchronization: local clients are not expected to die frequently and there's only a few of them running (probably at most 3 at a time).

For now I'll revert this change.

cabal.project
coot: Yes, I'll do that.

ouroboros-network/src/Ouroboros/Network/Socket.hs
```diff
@@ -327,6 +345,22 @@ fromSocket tblVar sd = Server.Socket
         Socket.close sd'
 
 
+data NetworkServerTracers ptcl vNumber peerid = NetworkServerTracers {
```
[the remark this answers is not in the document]
coot: Because of a circular dependency :/.

ouroboros-network/src/Ouroboros/Network/Subscription/Dns.hs
```diff
@@ -46,21 +47,25 @@ import           Ouroboros.Network.Subscription.Subscriber
 import           Ouroboros.Network.Subscription.Worker
 import           Ouroboros.Network.Socket
 
+
```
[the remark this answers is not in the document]
coot: Yes, that's just a stylistic change.

ouroboros-network/src/Ouroboros/Network/Tracers.hs
```diff
@@ -0,0 +1,57 @@
...
+
+import           Network.TypedProtocol.Driver (TraceSendRecv)
+import           Network.TypedProtocol.Driver.ByteLimit
+
+import           Ouroboros.Network.ErrorPolicy
+import           Ouroboros.Network.Protocol.Handshake.Type
+import           Ouroboros.Network.Subscription.Ip
+import           Ouroboros.Network.Subscription.Dns
+
+data NetworkIPSubscriptionTracers ptcl vNumber peerid = NetworkIPSubscriptionTracers {
```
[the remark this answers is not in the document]
coot: :)

Issue #1287, opened 2019-11-25
Chain DB: never adopt blocks from the future


mrBliss:
Replaces #1150. Because the ChainSync client allows for some clock skew, blocks from the future might be added to the ChainDB. The current implementation simply treats them as any other block, even though the spec tells us to ignore them: https://github.com/input-output-hk/ouroboros-network/blob/c23a5bdd212d4b550036cbf21836566bda2aa054/ouroboros-consensus/docs/ChainDB.md#L286

However, there is a problem with the spec: if the clock of our node is running behind, all blocks we receive might be from the future. Since we ignore all of them for chain selection, we will adopt no blocks at all. Forging a new block doesn't help, since it will fit onto the current tip of our chain, which does not include any of those blocks.

Note that receiving a single or a few blocks from the future is not a big deal, because when we receive a new block extending the fork they're in that is not from the future, we can still adopt that fork, including the blocks that were from the future at the time we received them.

What we should do instead is the following. Ignore the newly arrived block from the future, but run chain selection again using the blocks we already have *that are not from the future*. This makes sure that we eventually adopt blocks from the future when they no longer are from the future, *even when we only receive blocks from the future*. The spec and implementation must be updated accordingly.

edsko:
> even though the spec tells us to ignore them

:roll_eyes: 

:trollface:

mrBliss:
Addressed in #1288.

Issue #423, opened 2019-04-05
Rate limiting in the Chain Sync client


mrBliss:
The Chain Sync client can be exposed to adversarial behaviour:

* An upstream node is rapidly switching forks, e.g., producing multiple blocks
  for the same slot. A reliable upstream node never produces more than one
  block per slot.

* Multiple upstream nodes are collaborating in adversarial behaviour by each
  producing a block for the same slot, where each block has a different hot
  key (with different certificate numbers), but all hot keys are signed by the
  same (compromised) cold key.

  When this happens with two upstream nodes, there is no outright adversarial
  behaviour, as the block signed with the hot key with the higher certificate
  number "wins". This can happen when the hot key (with the lower certificate
  number) was compromised and is now used by some attacker (the node sending
  the first block). The reliable node then produced a new hot key with an
  incremented certificate number and sent the second block.

  However, when a third block is produced for the same slot with a different
  hot key signed by the same cold key, there is adversarial behaviour, as no
  reliable node would produce multiple new hot keys and sign blocks in the
  same slot.

The Chain Sync client must perform some form of rate limiting, e.g., by
delaying the interaction with an adversarial node or by disconnecting from it
(by throwing a specific exception). There will need to be some state per
upstream node to detect this, but also some state shared among all upstream
nodes.

We should not reject these blocks based on the order in which we happen to
receive them, as this could destroy consensus. It _might_ be possible to rule
out this adversarial behaviour at the consensus level (only one replacement
block allowed, all further blocks must be regarded as invalid) but this should
be properly discussed at the level of the consensus protocol with the
researchers.

mrBliss:
While item 2 is Shelley-specific, item 1 also applies to Byron. So I'm moving this to Byron.

edsko:
Some thoughts:

1. Generating new keys is relevant, because in order for an upstream node to keep convincing our node to switch to a new chain, that chain would either have to be longer and longer (and so the malicious node would have to be able to produce lots of blocks), _but_ if the malicious node is simply creating new keys all the time, he can cause lots of rollbacks even if he only has the right to produce a block in one slot.
2. Even if we rate limit individual upstream peers, we might still exceed a "global rate" if peers coordinate, though this depends on both (a) how many upstream peers we have at any given time and (b) how quickly the network layer allows _switching_ between peers.
[...]

PR #1288, opened 2019-11-25
Chain DB: trigger chain selection for blocks from the future

mrBliss:
Addresses #1287.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Background.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Background.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE PatternSynonyms     #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
@@ -8,7 +9,8 @@
 --
 -- * Copying blocks from the VolatileDB to the ImmutableDB
 -- * Performing and scheduling garbage collections on the VolatileDB
--- * Writing snapshots of the LedgerDB to disk and deleting old ones.
+-- * Writing snapshots of the LedgerDB to disk and deleting old ones
+-- * Executing scheduled chain selections
 module Ouroboros.Storage.ChainDB.Impl.Background
   ( -- * Launch background tasks
     launchBgTasks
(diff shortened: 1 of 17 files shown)


mrBliss:
Idea: when a block from the future arrives, after storing it in the VolatileDB, schedule an `addBlock` call in the future (for the given `Point`). Then we no longer need full chain selection. The model will need some changes: it will need to be aware of slot ticks, as they can affect the chain.

Bound of the queue size: `number of peers * max clock skew * ChainSync rate limit` #423

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/VolDB.hs
```diff
@@ -419,7 +419,7 @@ getKnownBlock db hash = do
       Left err -> throwM err
 
 getHeader :: (MonadCatch m, StandardHash blk, Typeable blk, GetHeader blk)
-               => VolDB m blk -> HeaderHash blk -> m (Maybe (Header blk))
+          => VolDB m blk -> HeaderHash blk -> m (Maybe (Header blk))
```
mrBliss: Not in this PR.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Background.hs
```diff
@@ -322,3 +333,84 @@ updateLedgerSnapshotsRunner cdb@CDB{..} = loop
...
+  :: (IOLike m, ProtocolLedgerView blk, HasCallStack)
+  => ChainDbEnv m blk
+  -> SlotNo  -- ^ The current slot
+  -> m ()
+scheduledChainSelection cdb@CDB{..} curSlot = do
+    mbHashes <- atomically $ updateTVar cdbFutureBlocks $
+      -- Extract and delete the value stored at @curSlot@
+      swap . Map.updateLookupWithKey (\_ _ -> Nothing) curSlot
+    whenJust mbHashes $ \hashes -> do
+    -- The list of hashes is stored in reverse order so we can easily prepend.
```
mrBliss: They're all in the same slot number, so they must all conflict; we can only adopt one.
Even for EBBs this doesn't matter, explain why.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -161,26 +171,53 @@ data Cmd blk it rdr
 -- something we are testing in 'prop_trace', see
 -- 'invalidBlockNeverValidatedAgain'.
 
+-- = Blocks from the future
```
mrBliss: Instead: add a separate "advanceTime" function to the model that is called when executing `Add(Future)Block`.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -989,7 +1119,10 @@ genBlk Model{..} = frequency
                 slotNo <- chooseSlot (blockSlot b + 1) (blockSlot b + 3)
                 body   <- genBody
                 return $ mkNextBlock b slotNo body)
-        , (1, do
+        -- An EBB is never followed directly by another EBB, otherwise they
+        -- would have the same 'BlockNo', as the EBB has the same 'BlockNo' of
+        -- the block before it.
+        , (if fromIsEBB (testBlockIsEBB b) then 0 else 1, do
```
mrBliss: Do I still need this?

PR #1289, opened 2019-11-25
Slot number in tx validation

edsko:


edsko:
Superseded by #1295 .

PR #1292, opened 2019-11-26
Change networkErrorPolicy

coot:
Be permissive in case of MuxBearerClosed && IOExceptions (during read
/ write operations).

The change is well explained in the code.  Currently, when a remote peer is not interested in our chain it will close the connection and the node shuts down connection to the peer for 20s - not long but there's no need to do that at all.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
@@ -568,16 +568,28 @@ remoteNetworkErrorPolicy = ErrorPolicies {
                         MuxControlProtocolError -> Just theyBuggyOrEvil
                         MuxTooLargeMessage      -> Just theyBuggyOrEvil
 
-                        -- in case of bearer closed / or IOException we suspend
-                        -- the peer for a short time
-                        --
-                        -- TODO: an exponential backoff would be nicer than a fixed 20s
-                        -- TODO: right now we cannot suspend just the
-                        -- 'responder'.  If a 'responder' throws 'MuxError' we
-                        -- might not want to shutdown the consumer (which is
-                        -- using different connection), as we do below:
-                        MuxBearerClosed         -> Just (SuspendPeer shortDelay shortDelay)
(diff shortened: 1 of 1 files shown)


coot:
In this case the remote node will only suspend its consumer.  The consensus error policy is [here](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node/ErrorPolicy.hs#L129)

karknu:
> In this case the remote node will only suspend its consumer. The consensus error policy is [here](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node/ErrorPolicy.hs#L129).

Consensus has consumer and producer running on different ports, so they are in effect different peers.

coot:
That's true for now, but in the future this might not be the case any more.

coot:
I think you're not right: we bind the address and the `PeerState` is per address not per connection: so both inbound and outbound connections for that remote peer will use the same key in `PeerStates` map.

karknu:
> That's true for now, but in the future this might not be the case any more.

I know, since I'm very much working on that. :)
https://github.com/input-output-hk/ouroboros-network/tree/karknu/i_and_r
https://github.com/input-output-hk/ouroboros-network/tree/karknu/mux_lazy

karknu:
> I think you're not right: we bind the address and the `PeerState` is per address not per connection: so both inbound and outbound connections for that remote peer will use the same key in `PeerStates` map.

The subscription worker is run with port 0, so will get a free ephemeral port. The server will bind to the configured port number.
However in the future they will use the same port number, see the `karknu/i_and_r` branch.

coot:
You're right then :).  Is there a reason why we're not binding to the actual port?

Still this change will protect us in the future.  I'll update the comment though.

karknu:
> You're right then :). Is there a reason why we're not binding to the actual port?
We do bind to the ephemeral port too, sorry I should have made that clear.

Issue #1291, opened 2019-11-25
Use applyChainTick before transaction validation


edsko:
The mempool must be able to validate transactions which have not yet been included in a block. At the moment, it validates these transactions against the _current ledger state_, in the assumption that they will be included in the "next" block. The set of transactions in the mempool is then updated every time a block comes in, and reevaluated against the new current ledger state.

_This is not correct_. The transactions in the mempool will be part of the _body_ of a block, but a block consists of a header and a body, and the full validation of a block insists of _first_ processing its header and only _then_ processing the body. This is important, because processing the header may change the state of the ledger: the update system might be updated, scheduled delegations might be applied, etc., _and such changes should take affect before we validate any transactions_. 

We therefore need to make sure, consensus side, that the transactions in the mempool are validated against the current ledger state _to which we have applied the "chain tick" function for the header of the block in which they will end up_. Of course, we don't have that header yet; it turns out, all we need is a slot, but we don't even have that slot number yet. We will therefore have to make some assumptions in the mempool: when we are validating transactions without producing a block (that is, when adding transactions into the mempool), we could assume the slot number is the next available one; when we are actually producing a block, then at that point we _do_ have the correct slot number and we should use it.

This will also require some changes to the Byron wrapper. At the moment, the chain tick function _only_ updates the update proposal state; it does _not_ change either `cvsLastSlot` or the delegation state; for the validation to be correct, it should change both. Technically this requires changes to the spec (see https://github.com/input-output-hk/cardano-ledger-specs/issues/1046), but since updates to the delegation state are idempotent, it's okay to (at least for the time being) do this consensus side (it will happen twice when we apply a block: once in the tick function and then again when ledger processes the block body, but that's not a big concern).

edsko:
Unfortunately `applyChainTick` was only recently removed from the API (https://github.com/input-output-hk/ouroboros-network/pull/1226); we will have to bring it back.

edsko:
https://github.com/input-output-hk/ouroboros-network/pull/1293 reintroduces the chain tick function and extends the Byron chain tick function.

PR #1293, opened 2019-11-26
Return of the slot tick

edsko:
This is required before we can implement #1291 (which will then hopefully resolve a lot of concerns about the application of the ledger rules for transaction validation).


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Aux.hs
```diff
@@ -141,6 +143,41 @@ setUpdateState :: U.Iface.State
                -> CC.ChainValidationState -> CC.ChainValidationState
 setUpdateState newUpdate state = state { CC.cvsUpdateState = newUpdate }
 
+{-------------------------------------------------------------------------------
+  Tick the delegation state
+
+  See 'applyChainTick' for discission.
```
dnadales:
```suggestion
  See 'applyChainTick' for discussion.
```

PR #1294, opened 2019-11-26
cddl-tests: Add the local transaction submission protocol

MarcFontaine:
CDDL-tests: Add the local transaction submission protocol.
* add localTxSubmission to the CDDL spec.
* test both directions (encoding and decoding).


ouroboros-network/test/CDDL.hs
```diff
@@ -205,6 +217,7 @@ decodeMsg (tag, input) = case tag of
     3 -> tryParsers ["blockFetch"]    blockFetchParsers
     4 -> tryParsers ["txSubmission"]  txSubmissionParsers
     5 -> tryParsers ["handshake"]     handshakeParsers
+    6 -> tryParsers ["localTxSubmission"]     localTxSubmissionParsers
```
coot: It's not essential, but the right hand sides are misaligned.

PR #1295, opened 2019-11-26
Use applyChainTick in mempool

### LINKED ISSUE #1291: Use applyChainTick before transaction validation

edsko:
This closes #1291.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Abstract.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Abstract.hs
@@ -7,6 +7,9 @@
 module Ouroboros.Consensus.Ledger.Abstract (
     -- * Interaction with the ledger layer
     UpdateLedger(..)
+  , TickedLedgerState(..)
+  , ledgerTipHash
+  , ledgerTipSlot
   , BlockProtocol
   , ProtocolLedgerView(..)
   , AnachronyFailure(..)
@@ -17,7 +20,8 @@ import           GHC.Stack (HasCallStack)
 
 import           Cardano.Prelude (NoUnexpectedThunks)
 
-import           Ouroboros.Network.Block (Point, SlotNo)
+import           Ouroboros.Network.Block (ChainHash, Point, SlotNo, pointHash,
+                     pointSlot)
 import           Ouroboros.Network.Point (WithOrigin)
 
 import           Ouroboros.Consensus.Block
@@ -59,10 +63,17 @@ class ( SupportedBlock blk
   -- primarily when validating transactions in the mempool (which, conceptually,
   -- live in "some block in the future") or when extracting valid transactions
   -- from the mempool to insert into a new block to be produced.
+  --
+  -- This is not allowed to throw any errors. After all, if this could fail,
+  -- it would mean a /previous/ block set up the ledger state in such a way
+  -- that as soon as a certain slot was reached, /any/ block would be invalid.
+  --
+  -- PRECONDITION: The slot number must be strictly greater than the slot at
+  -- the tip of the ledger.
   applyChainTick :: LedgerConfig blk
                  -> SlotNo
                  -> LedgerState blk
-                 -> Except (LedgerError blk) (LedgerState blk)
+                 -> TickedLedgerState blk
 
   -- | Apply a block to the ledger state.
   --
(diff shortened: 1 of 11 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -218,6 +223,30 @@ data Mempool m blk idx = Mempool {
     , zeroIdx       :: idx
     }
 
+-- | The slot of the block in which the transactions in the mempool will end up
+--
+-- The transactions in the mempool will be part of the body of a block, but a
+-- block consists of a header and a body, and the full validation of a block
+-- insists of first processing its header and only then processing the body.
```
edsko:
```suggestion
-- consists of first processing its header and only then processing the body.
```

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -218,6 +223,30 @@ data Mempool m blk idx = Mempool {
...
+-- This is important, because processing the header may change the state of the
+-- ledger: the update system might be updated, scheduled delegations might be
+-- applied, etc., and such changes should take affect before we validate any
```
edsko:
```suggestion
-- applied, etc., and such changes should take effect before we validate any
```

ouroboros-consensus/test-consensus/Test/Consensus/Mempool.hs
```diff
@@ -513,15 +513,23 @@ applyTxToLedger :: LedgerState TestBlock
...
-    updateLedgerTipPoint ledgerState = ledgerState
+    updateLedgerTipPoint (TickedLedgerState ledgerState) = ledgerState
         { tlLastApplied = BlockPoint { withHash = unSlotNo slot', atSlot = slot' } }
       where
-        slot' = case pointSlot (ledgerTipPoint ledgerState) of
+        slot' = case ledgerTipSlot ledgerState of
           Origin  -> SlotNo 0
           At slot -> succ slot
 
+
```
edsko:
```suggestion
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -59,10 +63,14 @@ class ( SupportedBlock blk
   -- primarily when validating transactions in the mempool (which, conceptually,
   -- live in "some block in the future") or when extracting valid transactions
   -- from the mempool to insert into a new block to be produced.
+  --
+  -- This is not allowed to throw any errors. After all, if this could fail,
+  -- it would mean a /previous/ block set up the ledger state in such a way
+  -- that as soon as a certain slot was reached, /any/ block would be invalid.
   applyChainTick :: LedgerConfig blk
```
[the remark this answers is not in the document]
edsko: Added

```
  -- PRECONDITION: The slot number must be strictly greater than the slot at
  -- the tip of the ledger.
```

PR #1176, opened 2019-10-29
[#622] Implement mempool transaction expiration

intricate:
Closes #622

Issue #622, opened 2019-06-07
Mempool: expire transactions


edsko:
Currently we remove transactions from the mempool only when they become invalid. However, it is possible for a transaction to be valid and yet never be included in a block. We should expire transactions after they have been sitting in the mempool for too long.

intricate:
A nice method of doing this could be expand upon the `TxSeq` `FingerTree` implementation such that we support measuring by slot number. That way, we could associate each `TxTicket` with a slot number and perform splits based on that. So to expire transactions, we'd split and discard `TxTicket`s associated with a slot number `< currentSlot - expirationThreshold`.

mrBliss:
@edsko pointed out that doing expiration on a slot-basis is not such a good idea:
* In Shelley, this would mean expiring transactions every 2 seconds. In the worst case, a full (expensive) revalidation of the Mempool is needed because a the oldest transaction is removed from the Mempool.
* In Shelley, all transactions already have a TTL, which is enforced by the ledger validation (*) so manual expiration of transactions will rarely happen.

(*) The background sync thread will deal with this and will let sit expired transactions in the Mempool to avoid revalidating every 2s.

Instead, expiration should be in terms of wall time, e.g., after 1 hour. So no `BlockchainTime`, but monotonic time.

PR #1300, opened 2019-11-26
Use the current BlockchainTime with applyChainTick in the Mempool

### LINKED ISSUE #1298: Utilize the current BlockchainTime with applyChainTick in the mempool
Related to issue #1291 and PR #1295.

Since the merge of PR #1295, when syncing the contents of the mempool with the ledger state, we first utilize `applyChainTick` before mempool transaction validation. 

If we're actually forging a block, we provide the mempool sync function with the `SlotNo` for which that block is to be forged. However, there are times where we sync the mempool contents without knowing the slot number of the block that they'll end up in. In this particular case, we just sync the mempool contents given the `ledgerTipSlot + 1`.

The problem with using the `ledgerTipSlot` is that it only represents the slot number in which the last block that was applied to the ledger was forged. i.e. this isn't necessarily the chain's current slot number as the `ledgerTipSlot` is only updated when new blocks are applied. For example, if 5 consecutive slots are missed, the `ledgerTipSlot` will be 5 slots behind the chain's current slot number.

To remedy this and ensure that we use the chain's actual current slot number, we should instead utilize the `SlotNo` from the current `BlockchainTime`.

intricate:
Closes #1298

### DIFF
--- a/ouroboros-consensus/test-consensus/Test/Consensus/Mempool.hs
+++ b/ouroboros-consensus/test-consensus/Test/Consensus/Mempool.hs
@@ -18,6 +18,7 @@ import qualified Data.Map as Map
 import           Data.Maybe (isJust, isNothing)
 import           Data.Set (Set)
 import qualified Data.Set as Set
+import           Data.Time.Clock (secondsToDiffTime)
 
 import           Test.QuickCheck
 import           Test.Tasty (TestTree, testGroup)
@@ -31,6 +32,8 @@ import           Ouroboros.Network.Block (pattern BlockPoint, SlotNo (..),
                      atSlot, withHash)
 import           Ouroboros.Network.Point (WithOrigin (..))
 
+import           Ouroboros.Consensus.BlockchainTime (NumSlots (..),
+                     TestBlockchainTime (..), newTestBlockchainTime)
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Mempool
 import           Ouroboros.Consensus.Mempool.TxSeq as TxSeq
@@ -70,7 +73,7 @@ tests = testGroup "Mempool"
 -- | Test that @snapshotTxs == snapshotTxsAfter zeroIdx@.
 prop_Mempool_snapshotTxs_snapshotTxsAfter :: TestSetup -> Property
 prop_Mempool_snapshotTxs_snapshotTxsAfter setup =
-    withTestMempool setup $ \TestMempool { mempool } -> do
+    withTestMempool setup $ \TestMempool { mempool } _ -> do
       let Mempool { zeroIdx, getSnapshot } = mempool
       MempoolSnapshot { snapshotTxs, snapshotTxsAfter} <- atomically getSnapshot
       return $ snapshotTxs === snapshotTxsAfter zeroIdx
@@ -79,7 +82,7 @@ prop_Mempool_snapshotTxs_snapshotTxsAfter setup =
 -- afterward.
 prop_Mempool_addTxs_getTxs :: TestSetupWithTxs -> Property
 prop_Mempool_addTxs_getTxs setup =
-    withTestMempool (testSetup setup) $ \TestMempool { mempool } -> do
+    withTestMempool (testSetup setup) $ \TestMempool { mempool } _ -> do
       let Mempool { addTxs, getSnapshot } = mempool
       _ <- addTxs (allTxs setup)
       MempoolSnapshot { snapshotTxs } <- atomically getSnapshot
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/test-consensus/Test/Consensus/Mempool.hs
```diff
@@ -646,7 +649,12 @@ withTestMempool setup@TestSetup { testLedgerState, testInitialTxs, testMempoolCa
     cfg = ledgerConfigView singleNodeTestConfig
 
     setUpAndRun :: forall m. IOLike m => m Property
-    setUpAndRun = do
+    setUpAndRun = withRegistry $ \registry -> do
+
+      -- Set up the BlockchainTime
+      TestBlockchainTime
+        { testBlockchainTime
+        } <- newTestBlockchainTime registry (NumSlots 1) (secondsToDiffTime 0)
```
intricate: In PR #1176, this will be replaced with a new type of `BlockchainTime`, `ManualBlockchainTime`, that can be manually ticked.

mrBliss: BTW, do you really need a separate type for it? I have written the following:
```haskell
settableBlockchainTime :: IOLike m => StrictTVar m SlotNo -> BlockchainTime m
settableBlockchainTime varCurSlot = BlockchainTime {
      getCurrentSlot = readTVar varCurSlot
    , onSlotChange_  = error "unimplemented onSlotChange_"
    }
```
If you need `onSlotChange_`, you'll need to do some more work, though.

edsko: `onSlotChange_` is also easy, just use `onEachChange`.

edsko: Indeed, the real block chain time is not much more than that; it just has an additional thread that updates the TVar when the wallclock changes. Could use this in the implementation of `realBlockchainTime`.

intricate:
Discussed this a bit with @mrBliss on Slack.

I believe that these changes may be valid after mempool transaction expiration is merged (issue #622, PR #1176).

For example, in the transaction expiry code, the expiry logic is run on each slot change according to the blockchain time. Whenever the transaction expiry logic results in some useful work being done (i.e. transactions were expired), mempool transaction validation occurs again since some of the expired transactions could have been dependencies of other unexpired transactions in the mempool. In this case, we probably wouldn't want to re-validate the mempool contents given the `ledgerTipSlot` but, rather, the current slot number according to the `BlockchainTime`.

As of right now, I agree that we should keep this PR closed as there doesn't seem to be much benefit.

_n.b. This PR doesn't change how often mempool syncing occurs. The sync thread still only performs a sync whenever the ledger is updated (not on each slot change according to the `BlockchainTime`). This PR just ensures that we use the current `BlockchainTime`'s `SlotNo` instead of the `ledgerTipSlot` when syncing given `TxsForUnknownBlock`._

mrBliss:
> _n.b. This PR doesn't change how often mempool syncing occurs. The sync thread still only performs a sync whenever the ledger is updated (not on each slot change according to the `BlockchainTime`). This PR just ensures that we use the current `BlockchainTime`'s `SlotNo` instead of the `ledgerTipSlot` when syncing given `TxsForUnknownBlock`._

Indeed, but it changes how often we'll need the expensive revalidation when syncing the Mempool.

edsko:
> n.b. This PR doesn't change how often mempool syncing occurs. The sync thread still only performs a sync whenever the ledger is updated (not on each slot change according to the BlockchainTime). This PR just ensures that we use the current BlockchainTime's SlotNo instead of the ledgerTipSlot when syncing given TxsForUnknownBlock.

This is not correct. If we have a steady flow of transactions coming in, this PR would mean that we'd re-evaluate the entire mempool every time the slot ticks (which, for Shelley, would be every 2 seconds). If we aim to reach high TPS (say >1000), this is not okay.

> For example, in the transaction expiry code, the expiry logic is run on each slot change according to the blockchain time.

Then maybe this is not correct either, and the expiry logic should not run so often?

edsko:
Discussed with @mrBliss who will leave a more detailed comment :)

mrBliss:
> > n.b. This PR doesn't change how often mempool syncing occurs. The sync thread still only performs a sync whenever the ledger is updated (not on each slot change according to the BlockchainTime). This PR just ensures that we use the current BlockchainTime's SlotNo instead of the ledgerTipSlot when syncing given TxsForUnknownBlock.
> 
> This is not correct. If we have a steady flow of transactions coming in, this PR would mean that we'd re-evaluate the entire mempool every time the slot ticks (which, for Shelley, would be every 2 seconds). If we aim to reach high TPS (say >1000), this is not okay.

This is due to a difference of understanding: the number of synchronisations is the same, but the number of expensive revalidation vs. cheap revalidations will be higher. With #1301 this would be even worse.

> 
> > For example, in the transaction expiry code, the expiry logic is run on each slot change according to the blockchain time.
> 
> Then maybe this is not correct either, and the expiry logic should not run so often?

Good point, see https://github.com/input-output-hk/ouroboros-network/issues/622#issuecomment-559110035

Conclusion: expiry logic should be time-based, not slot-based. But when expiration is triggered and some validations have actually been removed, a full revalidation is needed, at which point we /should/ use the current slot number for this full revalidation, which would incur no extra cost but give us more correctness.

PR #1305, opened 2019-11-27
Don't require Serialise instance for ChainState

edsko:
We already avoided this in the general setup, but still required it for
mock blocks. We need to avoid it altogether so that the encoding of the
PBFT chain state can depend on the configuration (required in the next
commit).

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Mock/Block/PBFT.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Mock/Block/PBFT.hs
@@ -7,6 +7,8 @@
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE UndecidableInstances       #-}
 
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
 module Ouroboros.Consensus.Ledger.Mock.Block.PBFT (
     SimplePBftBlock
   , SimplePBftHeader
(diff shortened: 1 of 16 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Mock/Block/PBFT.hs
```diff
@@ -91,11 +94,18 @@ instance ( SimpleCrypto c
...
+  mockProtocolMagicId  = const constructMockProtocolMagicId
+  mockEncodeChainState = const CS.encodePBftChainState
+  mockDecodeChainState = const CS.decodePBftChainState
+
 instance ( SimpleCrypto c
          , PBftCrypto c'
          , Signable (PBftDSIGN c') (SignedSimplePBft c c')
          , ConstructContextDSIGN ext c'
-         ) => ForgeExt (PBft ext c') c (SimplePBftExt c c') where
+         , Serialise (PBftVerKeyHash c')
```
mrBliss: Why is this constraint needed?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Mock/Run.hs
```diff
@@ -0,0 +1,46 @@
...
+           -> SimpleBlock' c ext ()
+           -> m (SimpleBlock c ext)
+
+-- | Construct protocol magic ID depending on where in the code this is called
+--
+-- The sole purpose of this is to make sure that these mock protocols have
+-- different IDs from each other and from regular protocols.
+constructMockProtocolMagicId :: HasCallStack => ProtocolMagicId
+constructMockProtocolMagicId =
+    ProtocolMagicId $ fromIntegral $ hash (prettyCallStack callStack)
```
mrBliss: Clever!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Mock/Block/PBFT.hs
edsko: It's not required by `PBftCrypto`, I suspect because we don't have an instance in scope for all kinds of keys?

PR #1313, opened 2019-12-03
Allow rewinding the PBftChainState to the genesis EBB

mrBliss:
This includes #1284 and #1307, the PR after which the tests in #1284 start to fail.

We first extend the RealPBFT to (optionally) generate an EBB in each epoch instead of just the genesis EBB.

Then we apply a quick 'n' dirty fix to `PBftChainState` so it can cope with rewinding to slot 0 at a time where only the genesis EBB exists in that slot, but not a regular block.

TODO: so far, the tests haven't triggered the other failure I would expect: rewinding to another EBB (slot > 0) at a time where there is no block in that slot yet. Maybe we can trigger this when not every slot is filled with a block? I'd like to see the test fail before I "fix" it.

### DIFF
--- a/nix/.stack.nix/ouroboros-consensus.nix
+++ b/nix/.stack.nix/ouroboros-consensus.nix
@@ -113,6 +113,7 @@
             (hsPkgs.ouroboros-consensus)
             (hsPkgs.io-sim-classes)
             (hsPkgs.io-sim)
+            (hsPkgs.binary-search)
             (hsPkgs.cborg)
             (hsPkgs.containers)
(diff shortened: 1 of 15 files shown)


mrBliss:
This is not meant to be merged as-is. After addressing the aforementioned TODO, it'll first need rebasing on master. Maybe this can be merged in #1284 or subsume #1284.

mrBliss:
Closing because this is/will be superseded by other PRs by @nfrisby: #1284 includes the "hotfix" and #1317 will include a more comprehensive solution.

PR #1317, opened 2019-12-04
consensus: track EBBs in PBFT chain state

### LINKED ISSUE #925: Resolve semantic conflict between `WithEBBs` and CS intersections at EBBs

### LINKED ISSUE #1312: Rewinding the PBFT ChainState to an EBB fails
See https://github.com/input-output-hk/ouroboros-network/pull/1284#issuecomment-560976105

nfrisby:
Closes #1312. Closes #925.

This is Option A on #1312: complicate the PBFT chain state to also track EBBs. This lets `rewind` target EBBs, which the CS client (unknowingly) needs it to do.

This is a Draft PR because:

  * This is not the option we were favoring, but I believed it would be useful to actually see it prototyped before dismissing it.
  * The diff does not extend the PBFT `ChainState` tests to test the new `ebbs` field.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DerivingStrategies         #-}
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
@@ -17,18 +18,19 @@ module Ouroboros.Consensus.Protocol.PBFT.ChainState (
     -- * Construction
   , empty
   , append
-  , appendMany
+  , appendEBB
   , rewind
     -- * Queries
   , countSignatures
   , countInWindow
   , countSignedBy
-  , lastSlot
+  , lastSignedSlot
     -- * Support for tests
-  , invariant
-  , toList
-  , fromList
+  , EbbMap (..)
   , PBftSigner(..)
+  , invariant
+  , fromList
+  , toList
     -- ** Serialization
   , encodePBftChainState
   , decodePBftChainState
@@ -39,6 +41,7 @@ import           Codec.Serialise.Decoding (Decoder)
 import qualified Codec.Serialise.Decoding as Serialise
 import           Codec.Serialise.Encoding (Encoding)
 import qualified Codec.Serialise.Encoding as Serialise
+import qualified Control.Exception as Exn
 import           Control.Monad.Except
 import qualified Data.Foldable as Foldable
 import           Data.Map.Strict (Map)
@@ -95,7 +98,7 @@ import           Ouroboros.Consensus.Util (repeatedly)
 -- not change the maximum rollback point).
 --
 -- This means that unless we are near genesis, we will at least have @n@
--- signatures in the history (after a maximum rollback of @k@), and under
+-- signatures in the history (even after a maximum rollback of @k@), and under
 -- normal circumstances (i.e., when not halfway a switch to a fork), @k+n@
 -- signatures.
 --
@@ -127,10 +130,41 @@ data PBftChainState c = PBftChainState {
       --
       -- We should have precisely @n@ signatures in the window, unless we are
       -- near genesis.
+      --
+      -- INVARIANT Empty if and only if we are exactly at genesis.
     , inWindow   :: !(StrictSeq (PBftSigner c))
 
       -- | Cached counts of the signatures in the window
     , counts     :: !(Map (PBftVerKeyHash c) Word64)
+
+      -- | Map from slots of relevant epoch boundary blocks (EBBs) to signed
+      -- slots
+      --
+      -- EBBs are not signed, so the 'preAnchor', 'postAnchor', 'preWindow',
+      -- and 'postWindow' fields are unaffected by EBBs. However, EBBs must
+      -- also be valid targets for 'rewind', so this field maps each slot that
+      -- contains an EBB to the preceding signed slot, /if/ that signed slot is
+      -- still a valid target for 'rewind'.
+      --
+      -- See INVARIANTs on 'EbbMap'.
+      --
+      -- INVARIANT For all @(ebbSlot, mSlot)@ in @'ebbs' (cs :: 'ChainState')@,
+      --
+      --  * @mSlot >= anchorSlot cs@; see 'pruneEBBsLT'
+      --
+      --  * @'At' ebbSlot <= tgt@ if @cs@ is the result of a 'rewind' to @tgt@;
+      --    see 'pruneEBBsGT'
+      --
+      --  * @and [ At s <= mSlot | s <- precedingSignedSlots ]@
+      --
+      --  * @'rewind' k n ('At' ebbSlot) cs = 'rewind' k n mSlot cs@
+      --
+      --   where
+      --
+      --  * @precedingSignedSlots = filter (< ebbSlot) signedSlots@
+      --
+      --  * @signedSlots = 'pbftSignerSlotNo' <$> ('preAnchor' <> 'postAnchor')@
+    , ebbs       :: !EbbMap
     }
   deriving (Generic)
 
@@ -175,9 +209,34 @@ invariant (SecurityParam k)
 
     unless (computeCounts inWindow == counts) $
       failure "Cached counts incorrect"
+
+    unless (allEbbs $ \_ mSlot -> mSlot >= anchorSlot st) $
+      failure "EBB mapped to slot before anchor"
+
+    unless (allEbbs $ \slot mSlot ->
+               let signedSlots          =
+                     fmap pbftSignerSlotNo $
+                     Foldable.toList $ preAnchor <> postAnchor
+                   precedingSignedSlots = filter (< slot) signedSlots
+               in all (\s -> At s <= mSlot) precedingSignedSlots
+           ) $
+      failure "EBB does not map to the preceding signature"
+
+    -- 'EbbMap''s "Key greater"
+    unless (allEbbs $ \slot mSlot -> At slot > mSlot) $
+      failure "EBB mapped to a simultaneous or future slot"
+
+    -- 'EbbMap''s "Non-descending"
+    unless (let nonDescending es = and $ zipWith (<=) es (tail es)
+            in nonDescending $ map snd $ Map.toAscList $ unEbbMap ebbs) $
+      failure $ "EBB mappings are not non-descending"
   where
     failure :: String -> Except String ()
     failure err = throwError $ err ++ ": " ++ show st
+
+    allEbbs p =
+      Map.null $
+      Map.filterWithKey (\slot mSlot -> not (p slot mSlot)) (unEbbMap ebbs)
 
 -- | The 'PBftChainState' tests don't rely on this flag but check the
 -- invariant manually. This flag is here so that the invariant checks could be
@@ -246,10 +305,24 @@ countSignedBy PBftChainState{..} gk = Map.findWithDefault 0 gk counts
 -- | The last (most recent) signed slot in the window
 --
 -- Returns 'Origin' if there are no signatures in the window (this will happen
--- near genesis only).
-lastSlot :: PBftChainState c -> WithOrigin SlotNo
-lastSlot PBftChainState{..} =
+-- exactly at genesis only).
+--
+-- Unaffected by EBBs, since they're not signed.
+lastSignedSlot :: PBftChainState c -> WithOrigin SlotNo
+lastSignedSlot PBftChainState{..} =
     case inWindow of
+      _ :|> signer -> At (pbftSignerSlotNo signer)
+      _otherwise   -> Origin
+
+-- | The anchor slot
+--
+-- Returns 'Origin' if there are no signatures in the window (this will happen
+-- exactly at genesis only).
+--
+-- Unaffected by EBBs, since they're not signed.
+anchorSlot :: PBftChainState c -> WithOrigin SlotNo
+anchorSlot PBftChainState{..} =
+    case preAnchor of
       _ :|> signer -> At (pbftSignerSlotNo signer)
       _otherwise   -> Origin
 
@@ -267,6 +340,7 @@ empty = PBftChainState {
     , preWindow  = Empty
     , inWindow   = Empty
     , counts     = Map.empty
+    , ebbs       = ebbsEmpty
     }
 
 -- | Append new signature
@@ -277,13 +351,16 @@ append :: forall c. PBftCrypto c
        -> WindowSize
        -> PBftSigner c
        -> PBftChainState c -> PBftChainState c
-append k n signer@(PBftSigner _ gk) PBftChainState{..} = assertInvariant k n $
+append k n signer@(PBftSigner _ gk) PBftChainState{..} =
+    assertInvariant k n $
+    pruneEBBsLT $
     PBftChainState {
         preAnchor  = preAnchor'
       , postAnchor = postAnchor'
       , preWindow  = preWindow'
       , inWindow   = inWindow'
       , counts     = updateCounts counts
+      , ebbs       = ebbs   -- NB this needs to be pruned
       }
   where
     (preAnchor', postAnchor') =
@@ -304,14 +381,6 @@ append k n signer@(PBftSigner _ gk) PBftChainState{..} = assertInvariant k n $
           , incrementKey gk
           )
 
--- | Append a bunch of blocks
-appendMany :: forall c. PBftCrypto c
-           => SecurityParam
-           -> WindowSize
-           -> [PBftSigner c] -- ^ Old to new
-           -> PBftChainState c -> PBftChainState c
-appendMany k n = repeatedly (append k n)
-
 -- | Rewind the state to the specified slot
 --
 -- This matches the semantics of 'rewindChainState' in 'OuroborosTag', in that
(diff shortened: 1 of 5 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -281,26 +288,69 @@ append k n signer@(PBftSigner _ gk) PBftChainState{..} = assertInvariant k n $
...
+      , preWindow  = preWindow
+      , inWindow   = inWindow
+      , counts     = counts
+      , ebbs       = Map.insert s latestNonEBBSlot ebbs
+      }
+  where
+    latestNonEBBSlot =
+      case inWindow of
+        Empty                                       -> Origin
+        _ :|> x | s >= s'                           -> err
```
mrBliss: `s` = slot of the EBB, `s'` = slot of the last signer in the window. Shouldn't this condition be: `s' > s`? I think it would be more readable to rename `s` and `s'` to `ebbSlot` and `lastSignerSlot`.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -281,26 +288,69 @@ append k n signer@(PBftSigner _ gk) PBftChainState{..} = assertInvariant k n $
...
+                | Just s'' <- mbLatestEBB, s >= s'' -> err
```
mrBliss: Same story here, right? But now it should be `s'' >= s` because there can be no two EBBs in the same slot. Rename `s''` to `prevEBBSlot` for readability.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -128,6 +130,10 @@ data PBftChainState c = PBftChainState {
 
       -- | Cached counts of the signatures in the window
     , counts     :: !(Map (PBftVerKeyHash c) Word64)
+
+      -- | Map from slots known to contain epoch boundary blocks (EBBs) to the
+      -- latest preceding slot known to contain a non-EBB block
+    , ebbs       :: !(Map SlotNo (WithOrigin SlotNo))
```
mrBliss: Maybe mention that EBBs are not signed and don't contribute to the counts.

Which invariants could we formulate for `ebbs`?

* For each `(k, v)` in `ebbs`, `v <= At k`: maps to *preceding* slots
* What is the relation to `pre/inWindow`?
* What is the relation to `pre/postAnchor`?

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
nfrisby: I renamed the variables and made the logic more direct.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
nfrisby: Same comment regarding logic. I renamed this variable too.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
nfrisby: I pushed a commit that ensures (still no tests yet) and documents the following invariants. It also mentions in a few comments that EBBs are unsigned -- that is an excellent clarifying observation!

```
      -- INVARIANT For all @(ebbSlot, mSlot)@ in @'ebbs' (cs :: 'ChainState')@,
      --
      --  * @mSlot >= anchorSlot cs@; see 'pruneEBBsLT'
      --
      --  * @'At' ebbSlot <= tgt@ if @cs@ is the result of a 'rewind' to @tgt@;
      --    see 'pruneEBBsGT'
      --
      --  * @and [ At s <= mSlot | s <- precedingSignedSlots ]@
      --
      --  * @'rewind' k n ('At' ebbSlot) cs = 'rewind' k n mSlot cs@
      --
      --   where
      --
      --  * @precedingSignedSlots = Seq.filter (< ebbSlot) signedSlots@
      --
      --  * @signedSlots = 'pbftSignerSlotNo' <$> ('preAnchor' <> 'postAnchor')@
```

and

```
-- INVARIANT Key greater: For all @(k, v)@, @At k > v@.
--
-- INVARIANT Non-descending: For all @(k1, v1)@ and @(k2, v2)@, @k1 < k2@
-- implies @v1 <= v2@.
```

nfrisby:
I just pushed a squashed/amended commit with a much more mature diff. Still no specific tests; doing that now.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -127,10 +129,41 @@ data PBftChainState c = PBftChainState {
       --
       -- We should have precisely @n@ signatures in the window, unless we are
       -- near genesis.
+      --
+      -- INVARIANT Empty if and only if we are exactly at genesis.
```
nfrisby: I think this invariant already held -- I didn't add anything to ensure it. I just added the comment to explicit list it.

nfrisby:
That last commit has a complicated diff. It updates Edsko's `PBFT.hs` tests to involve EBBs. I'm not happy with it -- his was much cleaner. Some of that is because it was previously was only counting one thing (signed slots), not two interleaved (EBBs and signed slots).

I hope to spend a couple hours tomorrow making it clearer; I think I can remove a couple auxiliary functions and/or improve the flow of the big generator.

But: all tests are passing, and they're exercising EBBs. The `runNodeNetwork` still only involves the 0-EBB, but the `PBFT.hs` tests involve all kinds of EBBs (they passed 10,000 tests locally, multiple times).

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -490,3 +548,88 @@ instance Serialise (PBftVerKeyHash c) => Serialise (PBftSigner c) where
...
+      At s   -> go gt
+        where
+          -- NB the \"Key greater\" invariant
+          gt = snd $ Map.split s m0
+  where
+    -- NB the \"Non-descending\" invariant
+    go m = case Map.minViewWithKey m of
+      Just ((_, mSlot), m')
+        | mSlot < mSlot0 -> go m'
+      _                  -> m
```
mrBliss: Can you add a property test checking that this function is equivalent to `\mSlot -> Map.filter (>= mSlot)`?

Since the `EbbMap` will in practice contain at most one entry (one EBB every `10k` blocks and the window size is `k`, I believe), I think we should actually just use `Map.filter` as it is much easier to read, evidently correct, and probably not slower than this approach.

ouroboros-consensus/test-consensus/Test/Consensus/Protocol/PBFT.hs
```diff
@@ -313,3 +378,105 @@ prop_appendOldStateRestoresPreWindow TestChainState{..} =
...
+  InputSigner   x -> CS.pbftSignerSlotNo x
+
+slotInputs :: Inputs c -> [SlotNo]
+slotInputs = map slotInput . unInputs
+
+lengthInputs :: Num a => Inputs c -> a
+lengthInputs = fromIntegral . length . unInputs
+
+headSigner :: Inputs c -> CS.PBftSigner c
+headSigner = head . signatureInputs
```
mrBliss: If you want, you can use `NonEmpty` list to avoid this partial `head`, but since this is test code, I didn't bother.

ouroboros-consensus/test-consensus/Test/Consensus/Protocol/PBFT.hs
```diff
@@ -313,3 +378,105 @@ prop_appendOldStateRestoresPreWindow TestChainState{..} =
...
+  -> CS.WindowSize
+  -> Inputs c
+  -> CS.PBftChainState c -> CS.PBftChainState c
+appendInputs k n = repeatedly (appendInput k n) . unInputs
+
+-- | May include EBBs before the first signer and after the last signer
+takeSigners :: Word64 -> Inputs c -> Inputs c
+takeSigners = \n0 -> Inputs . go n0 . unInputs
+  where
+    go !n = \case
```
mrBliss: (Curious about the bang: since we compare it against 0 before it can become a thunk (`n - 1`), it will either be in normal form (`v` for value) or `v - 1`. You can ignore this comment.)

mrBliss:
I'm thinking about what I said in https://github.com/input-output-hk/ouroboros-network/pull/1317#discussion_r354167568. I'm starting to think that the chances that -- when deployed -- we run into the problem this PR fixes are so small(*), that it would be fine to take more time to get this into a state you're completely happy with and that we don't need it done by the end of today. If you think you can get this done today, that's also fine :slightly_smiling_face:.

(*) There is at most one EBB in the window. Nearly all slots are filled (chain density is near 100%), so the chance that the EBB slot is not filled with a regular block is small. Moreover, roll backs are very rare. So the chance that we have to rewind after finding an EBB as intersection without there being a regular block in the same slot is *very* small.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -127,10 +129,41 @@ data PBftChainState c = PBftChainState {
...
+      -- | Map from slots of relevant epoch boundary blocks (EBBs) to signed
+      -- slots
+      --
+      -- EBBs are not signed, so the 'preAnchor', 'postAnchor', 'preWindow',
+      -- and 'postWindow' fields are unaffected by EBBs. However, EBBs must
+      -- also be valid targets for 'rewind', so this field maps each slot that
+      -- contains an EBB to the preceding signed slot, /if/ that signed slot is
+      -- still a valid target for 'rewind'.
+      --
+      -- See INVARIANTs on 'EbbMap'.
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -490,3 +548,88 @@ instance Serialise (PBftVerKeyHash c) => Serialise (PBftSigner c) where
...
+          -> PBftChainState c -> PBftChainState c
+appendEBB k n newEbbSlot cs@PBftChainState{..} =
+    assertInvariant k n $
+    Exn.assert valid $
+    cs{ebbs = ebbsInsert newEbbSlot latestNonEbbSlot ebbs}
+  where
+    latestEbbSlot    = ebbsMax ebbs
+    latestNonEbbSlot = lastSignedSlot cs
+
+    valid = At newEbbSlot > max latestEbbSlot latestNonEbbSlot
```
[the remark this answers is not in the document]
nfrisby: I don't think it can: e.g. if `newEbbSlot = latestEbbSlot` then this function would just override the existing value in the map. `assertInvariant` runs too late to know that the same key had already existed in the map. Similar for `latestNonEbbSlot`.

Basically: this assertion constrains the order of `append` and `appendEBB` operations, which are not themselves recorded in the 'ChainState'.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -490,3 +548,88 @@ instance Serialise (PBftVerKeyHash c) => Serialise (PBftSigner c) where
...
+pruneEBBsGT :: WithOrigin SlotNo -> PBftChainState c -> PBftChainState c
+pruneEBBsGT mSlot cs@PBftChainState{..} =
+  cs{ebbs = ebbsPruneKeyGT mSlot ebbs}
+
+-- | A map from the slots containing an EBB to the preceding signed slot
+--
+-- INVARIANT Key greater: For all @(k, v)@, @At k > v@.
+--
+-- INVARIANT Non-descending: For all @(k1, v1)@ and @(k2, v2)@, @k1 < k2@
+-- implies @v1 <= v2@.
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
nfrisby: I added the tests for both. Yes, I was worried the functions were overkill, at least for production. I suppose they might accelerate some hypothetical tests with small epochs?

ouroboros-consensus/test-consensus/Test/Consensus/Protocol/PBFT.hs
nfrisby: I added that by habit: since the `[]` case doesn't use `n`, GHC would not infer that the `go` function is strict, so the WorkerWrapper optimization wouldn't fire to unbox the `n` argument, yadda yadda loop yadda yadda.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
nfrisby: Fixed, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -464,23 +520,25 @@ fromList k n (anchor, signers) = assertInvariant k n $ PBftChainState {..}
...
                      -> WindowSize
                      -> Decoder s (PBftChainState c)
 decodePBftChainState k n = do
-      Serialise.decodeListLenOf 2
+      Serialise.decodeListLenOf 3
       anchor  <- withOriginFromMaybe <$> decode
       signers <- decode
-      return $ fromList k n (anchor, signers)
+      ebbs'   <- decode
+      return $ fromList k n (anchor, signers, ebbs')
```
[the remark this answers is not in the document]
nfrisby: Thanks for the notice, fixed with a version marker.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
nfrisby: These invariants are now included in `assertInvariant`.

(Also, previous commits added EBBs to the the existing PBFT ChainState tests.)

nfrisby:
@mrBliss I think I've addressed everything you've mentioned. Thanks for the great comments.

Would you review my responses and Resolve those Conversations?

Remaining and/or follow-up (please advise which is which) work:

  * Squash the commits.

  * The PBFT ChainState tests now include EBBs in all sorts of slots, with and without signed blocks also in those slots. But the test node net only ever creates Epoch 0's EBB and without a signed block in that slot. (Is there an "Protocol Testing" Issue for this incompleteness? It's awkward since our real nodes will only ever interact with historical EBBs, never actually forging them.)

  * Is there something we can do to emulate PR 1302 to validate this diff?

  *  Final judgement regarding whether we should go with this Option A versus something else for Issue 1312.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -543,27 +544,45 @@ fromList k n (anchor, signers, ebbs) =
...
                      -> WindowSize
                      -> Decoder s (PBftChainState c)
-decodePBftChainState k n = do
-      Serialise.decodeListLenOf 3
+decodePBftChainState k n = Serialise.decodeListLen >>= \case
+   2 -> do -- Version is 0
+      anchor  <- withOriginFromMaybe <$> decode
+      signers <- decode
+      return $ fromList k n (anchor, signers, ebbsEmpty)
+   4 -> do -- Version is >0
```
mrBliss: Hmm, what if the next version (I don't think we'll need one, but hypothetically speaking) happens to need more or fewer fields, e.g., 5? Then the `listLen` is incorrect (I know, *I* suggested this approach :slightly_smiling_face: ). I'm thinking now that it would be better to put the version number first (make sure it has the same size as the `encodeListLen` tag) and reserve version 2 (= `encodeListLen 2`) for the old format. Let's call the new format version 3 (newer version -> higher number). In the new format, the version number is then followed by another, proper, `encodeListLen 3`. Or does that go against the rules of CBOR?

Thoughts?

nfrisby:
I squashed and rebased onto `origin/master`. Removing the Draft qualifier.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
nfrisby: Thomas and I talked about it on Slack. We concluded:

  * "The list ship as sailed" insofar as we want to preserve the fact that the chain state's outermost structure as a CBOR term is that of a list.
  * As long as no future version of the format has exactly 2 fields, the fact that at least one field will always be an explicit version marker ensures future-proofness and backwards-compatibility are possible.
  * Byte size isn't a large concern, so there's significant cost to including a version marker.

Issue #1253, opened 2019-11-19
ImmutableDB: Detect all bitflips in epoch files


edsko:

mrBliss:
Addressed in #1320

Issue #290, opened 2019-02-14
Simulate random bit flip corruptions for the immutable DB


mrBliss:
#289 partially addresses #278. Unaddressed in #289:

> Randomly flipping some bits of a file is still future work because this type of corruption may result in valid files, which makes it much harder to keep the model in sync with the real database.

mrBliss:
Addressed in #1320

PR #1320, opened 2019-12-04
Immutable DB: detect bitflips in epoch files

mrBliss:
Addresses #290 and #1253.

Previously, we could only detect corruptions that cause deserialisation to fail, but not arbitrary bitflips. This PR fixes that.

* Add the ability to check whether a block got corrupted (= "integrity check") by checking its signature, body hash, etc. This works for all Byron blocks (see the property tests), except for those idiotic EBBs, which don't contain a signature or hash!

* Refactor the recovery to use streams (from http://hackage.haskell.org/package/streaming), making the code easier (IMO) and also making it easier to avoid having to keep all blocks from an epoch in memory.

* Use the CRC from the (secondary) index file to avoid the expensive the integrity check. Doing the integrity check for each block in https://github.com/input-output-hk/cardano-mainnet-mirror takes almost 4 min on my machine, using the CRC takes 1 min.

* Along the way, I fixed `db-convert` and `db-analyse` tools such that they they're less likely to break again in the future.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Util/CBOR.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Util/CBOR.hs
@@ -15,6 +15,7 @@ module Ouroboros.Consensus.Util.CBOR (
   , ReadIncrementalErr(..)
   , readIncremental
   , readIncrementalOffsets
+  , withStreamIncrementalOffsets
   ) where
 
 import qualified Codec.CBOR.Decoding as CBOR (Decoder)
@@ -29,6 +30,9 @@ import qualified Data.ByteString.Lazy as LBS
 import           Data.IORef
 import           Data.Word (Word64)
 import           GHC.Stack (HasCallStack)
+import qualified Streaming as S
+import           Streaming.Prelude (Of (..), Stream)
+import qualified Streaming.Prelude as S
 
 import           Ouroboros.Consensus.Util.IOLike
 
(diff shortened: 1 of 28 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Util/CBOR.hs
```diff
@@ -158,42 +162,57 @@ readIncrementalOffsets
...
+
+-- | Streaming variant of 'readIncrementalOffsets'
+--
+-- Continuation-passing style to ensure proper closure of the file.
+withStreamIncrementalOffsets
+  :: forall m h a r. (IOLike m, HasCallStack)
+  => HasFS m h
+  -> (forall s . CBOR.Decoder s (LBS.ByteString -> a))
+  -> FsPath
+  -> (Stream (Of (Word64, (Word64, a))) m (Maybe (ReadIncrementalErr, Word64)) -> m r)
```
edsko: Minor sad face about the use of `streaming` (the package) rather than `conduit`. Let's add a comment here that if we introduce user-facing streaming API also, the fact that we are using `streaming` here should not dictate that we should stick with it later; rather, we should revisit this code at that point.

ouroboros-consensus/src/Ouroboros/Consensus/Util/CBOR.hs
```diff
@@ -158,42 +162,57 @@ readIncrementalOffsets
...
+  -> m r
+withStreamIncrementalOffsets hasFS@HasFS{..} decoder fp = \k ->
+    withLiftST $ \liftST ->
+      withFile hasFS fp ReadMode $ \h -> k $ do
+        fileSize <- S.lift $ hGetSize h
+        if fileSize == 0
+          -- If the file is empty, we will immediately get "end of input"
+          then return Nothing
+          else S.lift (liftST (CBOR.deserialiseIncremental decoder)) >>=
+            go liftST h 0 Nothing [] fileSize
```
edsko: Alignment.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Parser.hs
```diff
@@ -76,77 +81,61 @@ epochFileParser'
...
-                   offset = Secondary.unBlockOffset $ Secondary.blockOffset entry
-               in (reverse acc, Just (err, offset))
+    stopAtCorruption
+      :: Stream (Of (Word64, (Word64, BinaryInfo (BlockInfo hash))))
+                m
+                (Maybe (Util.CBOR.ReadIncrementalErr, Word64))
+      -> Stream (Of (Word64, (Word64, BinaryInfo (BlockInfo hash))))
+                m
+                (Maybe (EpochFileError hash, Word64))
+    stopAtCorruption input = do
```
edsko: Nice example of where `streaming` is just not very nice :)

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Validation.hs
```diff
@@ -373,6 +358,38 @@ validateEpoch ValidateEnv{..} shouldBeFinalised epoch mbPrevHash = do
         UnexpectedError (e@InvalidFileError {}) -> Just e
         _ -> Nothing) err
 
+    -- | When reading the entries from the secondary index file, we need to
+    -- pass in a value of type 'IsEBB' so we know whether the first entry
+    -- corresponds to an EBB or a regular block. We need this information to
+    -- correctly interpret the deserialised 'Word64' as a 'BlockOrEBB': if
+    -- it's an EBB, it is the 'EpochNo' ('Word64'), it's a regular block, is a
```
mrBliss: Missing "if"

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Validation.hs
```diff
@@ -373,6 +358,38 @@ validateEpoch ValidateEnv{..} shouldBeFinalised epoch mbPrevHash = do
...
+    -- yet know whether the first block will be an EBB or a regular block. We
+    -- will find that out when we read the actual block from the epoch file.
+    --
+    -- Fortunately, we can make a /very/ good guess: if the 'Word64' of the
+    -- 'BlockOrEBB' matches the epoch number, it is almost certainly an EBB,
+    -- as the slot numbers increase @10k@ times faster than epoch numbers.
+    -- E.g., the first slot number in epoch 1 will be 21600 if @k = 2160@. We
+    -- could only make the wrong guess in the first very first epoch, i.e.,
+    -- epoch 0, as the first slot number is also 0. However, we know that the
+    -- real blockchain starts with an EBB, so even in that case we're fine.
```
edsko: Property: for every epoch `e`, for all slot numbers `s` in the epoch we have `s > e`.

edsko: We need a stronger consequence (follows from this property): it's an EBB _if and only if_ it matches (because you are actually depending on the opposite direction in `fixupEBB`.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Validation.hs
```diff
@@ -373,6 +358,38 @@ validateEpoch ValidateEnv{..} shouldBeFinalised epoch mbPrevHash = do
...
+    --
+    -- If the epoch size were 1, then we would make the wrong guess for each
+    -- epoch that contains an EBB, which is a rather unrealistic scenario.
+    --
+    -- Note that even making the wrong guess is not a problem. The only thing
+    -- that doesn't go according to plan is that we will needlessly overwrite
+    -- the secondary index file.
```
edsko: This is a bit subtle, and could do with some additional documentation: we extract _only_ the CRC from the index file, pass that to the parser, and then _reconstruct_ the entire index file and rewrite if there is a mismatch. This isn't totally obvious; for example, it wouldn't be crazy to also pass to the parser whether or not we expect the block to be an EBB (after all, this information is present in the index file we just read). That would break the logic here.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Parser.hs
```diff
@@ -79,74 +71,81 @@ epochFileParser'
...
+    --
+    -- If there's an expected checksum and it matches the block's checksum,
+    -- then the block is correct. Continue with the next.
+    --
+    -- If they do not match or if there's no expected checksum in the stream,
+    -- check the integrity of the block (expensive). When corrupt, stop
+    -- parsing blocks and return an error that the block is corrupt. When not
+    -- corrupt, continue with the next.
+    checkEntries
+      :: Stream (Of  CRC) m ()
```
edsko: I'd prefer this to be a list just to facilitate a potential future changeover to `conduit`.

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
```diff
@@ -187,6 +197,78 @@ prop_byronHashInfo_hashSize h =
...
+  Integrity
+-------------------------------------------------------------------------------}
+
+-- | Test that we can detect random bitflips in blocks.
+--
+-- We cannot do this for EBBs, as they are not signed nor have a hash, so we
+-- only test with regular blocks.
+prop_detectCorruption_RegularBlock :: RegularBlock -> Corruption -> Property
+prop_detectCorruption_RegularBlock (RegularBlock blk) =
+    detectCorruption
```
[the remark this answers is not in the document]
mrBliss: Good that you pointed this out! We're generating random blocks which don't even pass the check. We need a proper generator that generates valid blocks.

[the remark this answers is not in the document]
mrBliss: I have created https://github.com/input-output-hk/ouroboros-network/issues/1364 for this.

PR #1244, opened 2019-11-19
New ImmutableDB index format

mrBliss:

PR #1323, opened 2019-12-04
Immutable DB: misc improvements

mrBliss:
Now that #1244 is merged, we can take advantage of the ImmutableDB being aware of hashes.

Previously, we often had to read and parse the whole block from the ImmutableDB to know its hash. For example, when reading the block from the ImmutableDB corresponding to some `Point`, we first have to read and parse the block to check whether the block in that slot has the hash of the point. It's even worse when the slot can contain an EBB: we first read the block in that slot and when its hash doesn't match, we read the EBB of the epoch and check its hash, uggh.

This check can be done much more efficiently in the ImmutableDB itself, where can check the index file(s) to see whether we have the block with right slot *and* hash, without having to parse the block first.

This PR contains some other improvements: add missing docstrings & tweak a test generator.


mrBliss:
A few commits from this PR moved over to #1340.

mrBliss:
Included in/superseded by #1331.

PR #1325, opened 2019-12-04
remove illegal MonadAsync super constraints

avieth:
You cannot have type families in quantified constraints
https://gitlab.haskell.org/ghc/ghc/issues/17202#note_241130


avieth:
Alternatively, we could do something like

```Haskell
data Async m t = Async
  { asyncThreadId :: ThreadId m
    -- AsyncRep is the associated type in the MonadAsync class
  , asyncRep :: AsyncRep m t
  }

instance Eq (ThreadId m) => Eq (Async m t)
instance Ord (ThreadId m) => Ord (Async m t)
```

Issue #1266, opened 2019-11-20
Disallow opening the same ChainDB multiple times


mrBliss:
Use a lockfile, as suggested [here](https://github.com/input-output-hk/ouroboros-network/blob/fc164b5a3cbcd239a5803b5d3925205d29837a7c/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl.hs#L64).

mrBliss:
This looks like a good candidate to use for this: https://hackage.haskell.org/package/lukko

kderme:
Isn't filelock enough for this?
I have done some work here https://github.com/input-output-hk/ouroboros-network/pull/1326

mrBliss:
Addressed in #1326.

PR #1326, opened 2019-12-05
Use file locks for chainDB

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/1266


ouroboros-consensus/src/Ouroboros/Consensus/Node/DbMarker.hs
```diff
@@ -20,12 +22,15 @@ import           Data.ByteString.Lazy (fromStrict, toStrict)
 import qualified Data.Set as Set
 import           Data.Text (Text)
 import           Data.Word
+import           System.FileLock hiding (lockFile)
```
mrBliss:
```suggestion
import           System.FileLock
```

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbMarker.hs
```diff
@@ -141,6 +150,32 @@ instance Exception DbMarkerError where
...
+-- | If the file is locked, it throws an exception.
+-- The file is unlocked when the @registry@ is closed.
+lockDbMarkerFile :: ResourceRegistry IO -> FilePath -> IO ()
+lockDbMarkerFile registry dbPath = do
+    mlockFile <- allocateEither
+                    registry
+                    (const $ justToRight <$> tryLockFile fullPath Exclusive)
+                    unlockFile
+    case mlockFile of
+      Left () -> throwIO $ FileLocked fullPath
```
mrBliss:
```suggestion
      Left () -> throwM $ FileLocked fullPath
```
And remove the `Control.Exception` import.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbMarker.hs
```diff
@@ -141,6 +150,32 @@ instance Exception DbMarkerError where
...
+      Right _ -> return ()
+
+  where
+
```
mrBliss: Remove the blank lines around `where`. If this function were large, then they would be more welcome.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbMarker.hs
```diff
@@ -141,6 +150,32 @@ instance Exception DbMarkerError where
...
+    pFile    = fsPathFromList [dbMarkerFile]
+    -- just append dbMarkerFile instead?
```
mrBliss: You can remove this comment

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbMarker.hs
```diff
@@ -130,6 +135,10 @@ data DbMarkerError =
     -- be read. The file has been tampered with or it was corrupted somehow.
   | CorruptDbMarker
       FilePath         -- ^ The full path to the 'dbMarkerFile'
+
+    -- | The database is used by another process.
+  | FileLocked
```
mrBliss: I think `DbLocked` is a better name.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -111,6 +111,7 @@ run tracers chainDbTracer diffusionTracers diffusionArguments networkMagic dbPat
       (nodeProtocolMagicId (Proxy @blk) cfg)
     withRegistry $ \registry -> do
 
+      lockDbMarkerFile registry dbPath
```
mrBliss: I first thought that the functionality should be integrated into `checkDbMarker`, but that function is parametric in the file-system, and `HasFS` doesn't support file-locking.

The cleanest solution would be to add file-locking support to `HasFS`, but I don't think it is worth the effort, as we're not really going to use it in the tests.

So I think it is fine as is.

mrBliss:
Perfect!

PR #1327, opened 2019-12-05
Rewrite of the ntp-client

MarcFontaine:
This PR started as a refactoring of the ntp-client but evolved into a re-implementation.
(Therefore it does not make sense to review it as a diff)


ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +68,308 @@ data NtpStatus =
...
+            This v4 -> send v4
+            That v6 -> send v6
+            These v4 v6 -> send v6 >> send v4
+
+        send (sock, addrs) = forM_ addrs $ \addr -> do
+            p <- mkNtpPacket
+            void $ Socket.ByteString.sendTo sock (LBS.toStrict $ encode p) (setNtpPort $ Socket.addrAddress addr)
+            traceWith tracer NtpTracePacketSent
+            
+        loopForever = forever $ threadDelay 1_000_000
```
coot: This loop will wake up every second, it's better to use much longer delay.  An hour (or more) would be ok.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +68,308 @@ data NtpStatus =
...
+                   Report r -> return r
+
+        sendBoth s = case s of
+            This v4 -> send v4
+            That v6 -> send v6
+            These v4 v6 -> send v6 >> send v4
+
+        send (sock, addrs) = forM_ addrs $ \addr -> do
+            p <- mkNtpPacket
+            void $ Socket.ByteString.sendTo sock (LBS.toStrict $ encode p) (setNtpPort $ Socket.addrAddress addr)
```
coot: Instead of using expensive `LBS.toStrict` it's better to use vectored I/O in general (e.g. `sendManyTo (LBS.toChunks $ encode p)`.  It should be ok here as each ntp packet easily fits into a single chunk `32kb`, just a comment would do.

ntp-client/ntp-client.cabal
```diff
@@ -11,61 +11,35 @@ cabal-version:       >=1.20
...
+                      , time            >=1.6 && <1.10
 
   hs-source-dirs:       src
   default-language:     Haskell2010
-  ghc-options:          -Wall -Werror -fwarn-redundant-constraints
+  ghc-options:          -Wall 
   default-extensions:   DeriveDataTypeable
                         DeriveGeneric
                         GeneralizedNewtypeDeriving
                         OverloadedStrings
```
coot: Are `OverloadedStrings` really neded in this package.  Are we by any chance forming `ByteString`s manually?

ntp-client/ntp-client.cabal
```diff
@@ -11,61 +11,35 @@ cabal-version:       >=1.20
...
                         MonadFailDesugaring
```
coot: This is also out of scope, but here it goes :) - `MonadFailDesugaring` is turned on by default in `ghc-8.6.5`, so no need to turn in on any-more.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +68,308 @@ data NtpStatus =
...
     | NtpSyncUnavailable deriving (Eq, Show)
 
--- | Setup a NtpClient and run a computation that uses that client.
--- Todo : proper bracket-style tear-down of the NTP client.
-withNtpClient :: Tracer IO NtpTrace -> NtpClientSettings -> (NtpClient -> IO a) -> IO a
-withNtpClient tracer ntpSettings action
-   = forkNtpClient tracer ntpSettings >>= action
+type ReportPolicy = [ReceivedPacket] -> Summary
+
+data Summary = Report NtpOffset | Wait
```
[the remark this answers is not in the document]
MarcFontaine: yes

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +68,308 @@ data NtpStatus =
...
+    deriving (Eq, Show)
+
+data ReceivedPacket = ReceivedPacket
+    { receivedPacket    :: NtpPacket
+    , receivedLocalTime :: Microsecond
+    , receivedOffset    :: NtpOffset
+    } deriving (Eq, Show)
```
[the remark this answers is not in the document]
MarcFontaine: yes

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +68,308 @@ data NtpStatus =
...
+                return $ Right sock
+            Left err -> do
+                Socket.close sock
+                return $ Left err
+    where
+        trySock s = do
+            Socket.setSocketOption s ReuseAddr 1
+            Socket.bind s (addrAddress addr)
+            -- Send testPacket ?
+            void $ Socket.ByteString.sendTo s "test" (setNtpPort $ Socket.addrAddress someDest)
```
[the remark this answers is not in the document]
MarcFontaine: I want to catch the case that I open a socket, but the destination is not reachable.
This is exactly what happens when I try it with my network setup.
I can open an IPv6 socket but I cannot send on it, because there are no outgoing routes
for the destination.

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: How about 500_000_000 ?

ntp-client/src/Network/NTP/Client.hs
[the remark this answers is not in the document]
coot: This will be reviled eventually by sending ntp packet.  This is non-recoverable error, so it's fine to just propagate the error to the application that is using the `ntp-client` to decide what to do with it.

ntp-client/src/Network/NTP/Client.hs
coot: https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/Protocol/LocalTxSubmission/Client.hs#L47
and a few other places (note that the `threadDelay` there is not the same as one you're using here, hence the missing `0`s).

ntp-client/src/Network/NTP/Client.hs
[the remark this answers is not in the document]
MarcFontaine: No it is more complicated:
If there is an error in the `runQueryLoop`, I terminate `oneshotClient` pause for some delay
and start from a clean state.
If there is an error while setting up IPv4 or IPv6 I start the `runQueryLoop` with whatever protocol is available.
Before I start the `runQueryLoop` with either socket I must ensure that I can send at least one packet.
I need to trigger the error while setting up the sockets, because after the setup the error will propagate differently

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/Protocol/LocalTxSubmission/Client.hs#L47
43200s is a day (but only the 12h without the night)
I just did not want to cause some Int overrun and not sleep too long.
I can change it to 43_200_000_000 but some docs say only 2^29 is safe for Ints.
(Is Karl running cardano on 32Bit systems ? I can't remember)

ntp-client/src/Network/NTP/Client.hs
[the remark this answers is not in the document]
MarcFontaine: During setup the error will cause the protocol version to be ignored.
and after setup an error will cause a restart

ntp-client/src/Network/NTP/Client.hs
coot: We can just use `maxBound` then :)

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -1,45 +1,45 @@
...
+                     Socket, SocketOption (ReuseAddr), SocketType (Datagram),
+                     addrAddress, addrFamily, addrFlags, addrSocketType)
 import qualified Network.Socket as Socket
-import           Network.Socket.ByteString (recvFrom)
+import qualified Network.Socket.ByteString as Socket.ByteString (recvFrom, sendTo)
 
-import           Network.NTP.Packet (NtpOffset (..) , NtpPacket (..), clockOffset,
-                     mkNtpPacket, ntpPacketSize)
+import           Network.NTP.Packet (NtpPacket, mkNtpPacket, ntpPacketSize, Microsecond
+                                    , NtpOffset (..), getCurrentTime, clockOffsetPure)
```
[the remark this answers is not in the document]
MarcFontaine: I don't know what do do here.
fix the indentation of the second part of the imports ?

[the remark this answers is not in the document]
coot: I am refering to this style of imports:
```
import           Network.NTP.Packet ( NtpPacket
                                                       , mkNtpPacket
                                                       , ntpPacketSize
                                                       , Microsecond
                                                       , NtpOffset (..)
                                                       , getCurrentTime
                                                       , clockOffsetPure
                                                       )
```

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +64,221 @@ data NtpStatus =
...
+
+    release :: Socket -> IO ()
+    release s = do
+        Socket.close s
+        traceWith tracer NtpTraceSocketClosed
+
+    action :: Socket -> IO [ReceivedPacket]
+    action socket = do
+        Socket.setSocketOption socket ReuseAddr 1
+        Socket.bind socket (addrAddress localAddr)
```
coot: if we don't bind, we will get an ephemeral port, which will be different each time we run the loop. This will make it even harder to DDOS.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +64,221 @@ data NtpStatus =
...
+        inQueue <- atomically $ newTVar []
+        _err <- withAsync (send socket  >> loopForever)      $ \sender ->
+               withAsync timeout    $ \delay ->
+               withAsync (reader socket inQueue ) $ \revc ->
+                    waitAnyCancel [sender, delay, revc]
```
coot: I thought we agreed with @semanticphilosopher that we can pipeline the requests and then await for responses.  This would required only two threads (including one for timeout one).

The indentation and spacing is wired.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +64,221 @@ data NtpStatus =
...
+    send :: Socket -> IO ()
+    send sock = forM_ destAddrs $ \addr -> do
+        p <- mkNtpPacket
+        err <- tryIOError $ Socket.ByteString.sendManyTo sock
+                          (LBS.toChunks $ encode p) (setNtpPort $ Socket.addrAddress addr)
+        case err of
+            Right _ -> traceWith tracer NtpTracePacketSent
+            Left e  -> do
+                traceWith tracer $ NtpTracePacketSentError e
+                ioError e
```
coot: I would use a `onException` (or `catch` if you only want to act on `IOException`s`).  This will make transition to `MonadCatch` easier (it does not have the low level `System.IO` exception control flow functions).

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +64,221 @@ data NtpStatus =
...
+                    traceWith tracer $ NtpTraceRunProtocolSuccess version
+                    return r
+
+socketAction ::
+       Tracer IO NtpTrace
+    -> NtpClientSettings
+    -> AddrInfo
+    -> [AddrInfo]
+    -> IO (Either IOError [ReceivedPacket])
+socketAction tracer netSettings localAddr destAddrs
```
coot: Maybe there's a more descriptive name for this function. What about `requestNtpMeasurements`?

It would be also good to add haddoc documentation what it does in a descriptive manner.

ntp-client/src/Network/NTP/Packet.hs
```diff
@@ -108,7 +111,7 @@ ntpToRealMcs sec frac =
 -- @
 -- (in microseconds; this is roughly 66 years, so we're fine untill 2036).
 realMcsToNtp :: Microsecond -> (Word32, Word32)
-realMcsToNtp (toMicroseconds -> mcs) =
+realMcsToNtp (Microsecond mcs) =
```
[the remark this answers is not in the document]
MarcFontaine: Yes there are two tests:
`prop_timeConvert1` and `prop_timeConvert2`

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -1,65 +1,58 @@
...
+      -- ^ How long to wait between two rounds of requests.
+    , ntpReportPolicy    :: [ReceivedPacket] -> Maybe NtpOffset
     }
 
 data NtpClient = NtpClient
     { -- | Query the current NTP status.
       ntpGetStatus        :: STM NtpStatus
-
       -- | Bypass all internal threadDelays and trigger a new NTP query.
     , ntpTriggerUpdate    :: IO ()
```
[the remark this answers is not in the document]
MarcFontaine: Do you mean make `ntpTriggerUpdate` blocking and wait until `NtpStatus` has changed ?
Should it wait until an a measurement is available ?

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +64,221 @@ data NtpStatus =
...
+    -> IO ()
+oneshotClient tracer (ntpSettings, ntpStatus) = do
+    traceWith tracer NtpTraceClientStartQuery
+    (v4Servers,   v6Servers)   <- lookupServers $ ntpServers ntpSettings
+    (v4LocalAddr, v6LocalAddr) <- udpLocalAddresses >>= firstAddr "localhost"
+    (v4Replies, v6Replies) <- concurrently (runProtocol IPv4 v4LocalAddr v4Servers)
+                                           (runProtocol IPv6 v6LocalAddr v6Servers)
+    when (null v4Replies && null v6Replies) $ do
+        traceWith tracer NtpTraceIPv4IPv6BothFailed
+        ioError $ userError "IPv4 and IPv6 failed"
```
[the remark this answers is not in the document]
MarcFontaine: I want to make the case where there are no replies from IPv4 nor IPv6 an error that
propagates one level up and causes the client to be respawn (after 10 s) instead of just a regular query.
In the other case there was at least one reply and (maybe) its is more OK to just wait
for 30 minutes and try again.

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: Interesting. I didn't know about that. I was just following the pattern of the old ntp-client.

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: I like this better.
I think it reflects better what we actually want and it is not more complicated either.
sendManyTo is not expected to block, but it does not harm to have it covered by a timeout.
Also there is the implicit assumption that we time-stamp the incoming packets as fast as possible (or more precise there is the assumption that the transmission times for a request/response are the same in both directions)
If we wait for all requests to be send, it will skew the measurement.
(And also there is now the 100ms rate-limiting when sending.)

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: `queryServers` ?

ntp-client/src/Network/NTP/Client.hs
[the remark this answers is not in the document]
coot: Yes, it should block until the result is available or it should timeout.

ntp-client/src/Network/NTP/Client.hs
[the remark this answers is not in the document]
coot: If there are no replies that should be visible without throwing exceptions.  Just pattern matching on the list of results.

ntp-client/src/Network/NTP/Client.hs
coot: That's ok.

ntp-client/src/Network/NTP/Client.hs
coot: or `runNtpQueries`

ntp-client/src/Network/NTP/Client.hs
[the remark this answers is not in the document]
MarcFontaine: How about leaving `ntpTriggerUpdate` in the API as is
and adding `oneshotClient` as and independent API
that does not interfere with `TVar NtpStatus` at all.
With a nice self-contained  `oneshotClient`, short-lived sockets ,fast queries and short timeouts,
there are few benefits for using `withNtpClient` and `TVar NtpStatus` at all.
Extra advantage: a user of `oneshotClient`  can change the settings and use a different set of
NTP-servers for every query.
`oneshotClient` does exactly what you propose as the new semantics of `ntpTriggerUpdate`

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -1,65 +1,58 @@
...
-      -- ^ delay between making requests and response collection
+      -- ^ Timeout between sending NTP requests and response collection.
     , ntpPollDelay       :: Microsecond
-      -- ^ how long to wait between to send requests to the servers
-    , ntpSelection       :: NonEmpty NtpOffset -> NtpOffset
-      -- ^ way to sumarize results received from different servers.
-      -- this may accept list of lesser size than @length ntpServers@ in case
-      -- some servers failed to respond in time, but never an empty list
+      -- ^ How long to wait between two rounds of requests.
+    , ntpReportPolicy    :: [ReceivedPacket] -> Maybe NtpOffset
```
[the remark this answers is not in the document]
MarcFontaine: I removed `ReceivedPacket` alltogether.
(The idea was to support more features (like spoof protection) that need more info the the time offset`.
`ReceivedPacket` can added again later.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -71,248 +64,221 @@ data NtpStatus =
...
+resolveHost host = Socket.getAddrInfo (Just hints) (Just host) Nothing
+  where
+    hints = Socket.defaultHints
+            { addrSocketType = Datagram
+            , addrFlags = [AI_ADDRCONFIG]  -- since we use @AF_INET@ family
+            }
+
+firstAddr :: String -> [AddrInfo] -> IO (Maybe AddrInfo, Maybe AddrInfo)
+firstAddr name l = case (find isV4Addr l, find isV6Addr l) of
+    (Nothing, Nothing) -> ioError $ userError $ "lookup host failed :" ++ name
```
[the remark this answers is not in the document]
MarcFontaine: I rearranged it and improved tracing

ntp-client/src/Network/NTP/Client.hs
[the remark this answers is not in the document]
MarcFontaine: I renamed and reordered the functions and added `ntpQueryBlocking` to the API

ntp-client/src/Network/NTP/Client.hs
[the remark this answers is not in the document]
MarcFontaine: OK I changed it:
new behavior: No difference between no replies and errors/exceptions.
fast respawn in both cases

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: lets leave that for a new PR

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -1,318 +1,85 @@
...
-            `concurrently_` sendLoop tracer cli addrs
-            `concurrently_` traceWith tracer NtpTraceSpawnNtpClientStarted
-
+    ntpStatus <- newTVarIO NtpSyncPending
+    withAsync (ntpClientThread tracer ntpSettings ntpStatus) $ \tid -> do
+        let client = NtpClient
+              { ntpGetStatus = readTVar ntpStatus
+              , ntpTriggerUpdate = do
+                  traceWith tracer NtpTraceTriggerUpdate
+                  atomically $ writeTVar ntpStatus NtpSyncPending
```
coot: ^ I don't think the wallet will need this method. `ntpGetStatus` and `ntpQueryBlocking` is all what they need.  You could also expose the `TVar`. The wallet already has logic which is using it.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -1,318 +1,85 @@
...
-        threadDelay 5000000
-        doMkSockets
+        restartDelay :: [Int]
+        restartDelay = [0, 5, 10, 20, 60, 180, 600] ++ repeat 600
+
+        queryLoop = ntpQuery tracer ntpSettings >>= \case
+            status@(NtpDrift _ ) -> do
+                atomically $ writeTVar ntpStatus status
+                traceWith tracer NtpTraceClientSleeping
+                awaitPendingWithTimeout ntpStatus $ fromIntegral $ ntpPollDelay ntpSettings
```
coot: this `queryLoop` will send requests much more often than `ntpPollDelay`, which is not our intention.
 
At this point of execution, `ntpStatus` will be likely (e.g. in good network conditions) set to `NtpDrift`, since `ntpQuery` awaits on results, and thus `awaitPendingWithTimeout` will return immediately.  Even if `ntpQuery` did not await on results, this would await only as long as the responses are in-flight, not as long as the `ntpPollDelay` instructs.

ntp-client/src/Network/NTP/Query.hs
```diff
@@ -0,0 +1,211 @@
...
+    -> AddrInfo
+    -> [AddrInfo]
+    -> IO (Either IOError [NtpOffset])
+runNtpQueries tracer protocol netSettings localAddr destAddrs
+    = tryIOError $ bracket acquire release action
+  where
+    acquire :: IO Socket
+    acquire = do
+        s <- Socket.socket (addrFamily localAddr) Datagram Socket.defaultProtocol
+        traceWith tracer $ NtpTraceSocketOpen protocol
```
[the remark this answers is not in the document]
MarcFontaine: OK I moved the tracer to action, its more safe.
(async exceptions are masked in the bracket but who knows..)

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: OK I removed the `ntpTriggerUpdate`.
We can put the TVar in the API if the wallet really needs it, but I think it is an implementation detail, I would rather not put it in the API because it is not explicit how to use it.
I using a single TVar for reporting the result and forcing an update looks like a hack to me.
(though I admit that it gives a concise implementation)

ntp-client/src/Network/NTP/Client.hs
MarcFontaine: I think it is correct as is it.
The intention is that `awaitPendingWithTimeout` waits until `ntpStatus` gets set to `NtpSyncPending` or the timeout.
`Network.NTP.Test.testClient` also works as expected.

ntp-client/src/Network/NTP/Query.hs
```diff
@@ -0,0 +1,211 @@
...
+          , addrSocketType = Datagram
+          }
+    port = Socket.defaultPort
+
+resolveHost :: String -> IO [AddrInfo]
+resolveHost host = Socket.getAddrInfo (Just hints) (Just host) Nothing
+  where
+    hints = Socket.defaultHints
+            { addrSocketType = Datagram
+            , addrFlags = [AI_ADDRCONFIG]  -- since we use @AF_INET@ family
```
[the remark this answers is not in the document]
MarcFontaine: Good, these are the platforms we need to support,
we keep it as it is, or is it better to use`AI_ALL` ?

ntp-client/src/Network/NTP/Query.hs
```diff
@@ -0,0 +1,211 @@
...
+        ntpPort = 123
+
+lookupServers :: Tracer IO NtpTrace -> [String] -> IO ([AddrInfo], [AddrInfo])
+lookupServers tracer names = do
+    dests <- forM names $ \server -> do
+        addr <- resolveHost server
+        case (firstAddr AF_INET addr, firstAddr AF_INET6 addr) of
+            (Nothing, Nothing) -> do
+                traceWith tracer $ NtpTraceLookupServerFailed server
+                ioError $ userError $ "lookup NTP server failed " ++ server
```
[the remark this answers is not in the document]
MarcFontaine: Yes, to keep it simple the rest of the code only handles IOErrors explicitly and only IOErrors
and Async exceptions are expected to happen. All other exception are propagated.
All other exceptions are probably caused by bugs in the program.

ntp-client/src/Network/NTP/Query.hs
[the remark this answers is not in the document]
coot: > (async exceptions are masked in the bracket but who knows..)

masked async exceptions doesn't mean they will not happen at all, it only means they will not happen in non-blocking code. We don't control what happens inside the tracer, whether it's blocking or non-blocking `io`.

ntp-client/src/Network/NTP/Query.hs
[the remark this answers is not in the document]
coot: I think it's ok. I would only add a `TODO` to test that `AI_ADDRCONFIG` works fine.

ntp-client/src/Network/NTP/Client.hs
coot: You're right, it's ok.

ntp-client/src/Network/NTP/Client.hs
```diff
@@ -1,318 +1,85 @@
...
+                queryLoop
+            _ -> return ()
```
coot: If we didn't received any results, we will just terminate?

ntp-client/src/Network/NTP/Client.hs
coot: As you feel. Exposing `TVar` gives more power to the user.  My only point was that we don't need `ntpTriggerUpdate` when we expose `ntpQueryBlocking`.

ntp-client/src/Network/NTP/Client.hs
coot: Found out this by my self, the `ntpClientThread` will re-start in that case.

You could use this type signature:
```
ntpClientThread ::
       Tracer IO NtpTrace
    -> NtpSettings
    -> TVar NtpStatus
    -> IO Void
```
to indicate that `ntpClientThread` is non terminating.

coot:
@MarcFontaine could you elaborate. The client will use `5 10 20 ...` up to `600` intervals between restarts: https://github.com/input-output-hk/ouroboros-network/blob/master/ntp-client/src/Network/NTP/Client.hs#L98

coot:
arg, there's a mistake, it should be:
```
2 * delay `min` 600
```
not `max` :D

MarcFontaine:
I created an issue we can discuss it there:
 https://github.com/input-output-hk/ouroboros-network/issues/1583

ntp-client/src/Network/NTP/Query.hs
```diff
@@ -56,22 +56,32 @@ data NtpStatus =
...
 udpLocalAddresses :: IO [AddrInfo]
---                                      Hints        Host    Service
-udpLocalAddresses = Socket.getAddrInfo (Just hints) Nothing (Just $ show port)
+udpLocalAddresses = Socket.getAddrInfo (Just hints) Nothing Nothing
   where
     hints = Socket.defaultHints
           { addrFlags = [AI_PASSIVE]
           , addrSocketType = Datagram
           }
-    port = Socket.defaultPort
```
MarcFontaine: This does not work:
`getAddrInfo` needs at least one Just argument.

mark-stopka:
Hi,
is there an option to disable built-in `ntp-client` and use `system time` instead?

coot:
Hi @mark-stopka 
The reason for having this component is to check that system clock time is in sync with the rest of the world. It is not used to set or read time.

It's worth to mention that this is not to substitute a proper ntp client run as a system service.

mark-stopka:
Hi @coot, thanks for your feedback,
I was just asking, because [Linux Kernel patches introducing Time Namespaces](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?h=timers/core&id=769071ac9f20b6a447410c7eaa55d1a5233ef40c) is *maybe* going to be introduced in Linux 5.6.0. Lot of people running pools on the ITNv1 with Jormungandr were following official guides, which included steps to configure **Chrony** NTP client and they were obviously falling for all those who did not have real VMs, but rather LXC containers. I'd expect Linux 5.6.x to be LTS distributions in few years time =).

Thanks for confirming,
I'll have a look around a logic that you use your application specific time or it's delta with system time for.

PR #1331, opened 2019-12-06
Binary streaming of headers

### LINKED ISSUE #1330: ChainSyncServer: stream binary headers
Just like #1271 implemented streaming binary blocks for the BlockFetchServer, we'll need to do the same thing for the ChainSyncServer.

mrBliss:
Closes #1330.

* ImmutableDB: add a header envelope

* ChainDB.ImmDB: allow reading just the header

* ChainDB.Reader: split readers into block and header readers

* Add codecChainSyncSerialised

* ChainDB: let Readers return `Deserialisable` headers/blocks

* ChainSyncServer: stream binary headers/blocks 

* ChainDB.Reader: add `ReaderInit` state as an optimisation 

* Includes/supersedes #1323.


mrBliss:
@kderme: just a heads-up: https://github.com/input-output-hk/ouroboros-network/pull/1331/commits/0162c894f31fae558324c0e73085c1818033773b

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/API.hs
```diff
@@ -558,6 +559,18 @@ data Reader m blk a = Reader {
 instance Eq (Reader m blk a) where
   (==) = (==) `on` readerId
 
+-- | Deserialise the results of a reader that yields serialised blocks or
+-- headers.
+deserialiseReader
+  :: Monad m => Reader m blk (Deserialisable m blk b) -> Reader m blk b
+deserialiseReader rdr = Reader
+    { readerInstruction         = readerInstruction         rdr >>= traverse (traverse  deserialise)
```
mrBliss: Two spaces!

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Reader.hs
```diff
@@ -268,45 +260,64 @@ instructionHelper registry fromMaybeSTM fromPure CDB{..} varReader = do
...
-      ImmDB.IteratorExhausted      -> Nothing
-
-    rollForwardImmDB :: ImmDB.Iterator (HeaderHash blk) m blk -> Point blk
-                     -> m r
-    rollForwardImmDB immIt pt = nextBlock immIt >>= \case
-      Just blk -> do
-        let pt'          = blockPoint blk
+    toDeserialisableHeader :: Header blk -> m (Deserialisable m blk (Header blk))
+    toDeserialisableHeader hdr = return Deserialisable
+      { serialised         = Serialised $ toLazyByteString (encodeHeader hdr)
```
edsko: This needs justification: we are always constructing the serialized form, even if the user might not necessarily care about it. This is cheap for us, because that serialized form just the annotation; however, that's then something that should be documented at the API level (docstring for "encode header" should say: we assume that this is O(1) -- maybe even rename to "get encoding" rather than "encode"), with a comment that if that assumption cannot be satisfied, we should rethink think this here -- after all, as you point out, the blocks/headers near the tip of the chain are the ones that we will read most.

edsko: (If we are streaming stuff near the tip of the chain, and this assumption cannot be satisfied, perhaps we should change e.g. the representation of the tip of the chain to also includes serialized forms -- not needed now, but we should document assumptions and rationale at least.)

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Args.hs
```diff
@@ -50,12 +50,14 @@ data ChainDbArgs m blk = forall h1 h2 h3. ChainDbArgs {
...
     , cdbDecodeBlock      :: forall s. Decoder s (Lazy.ByteString -> blk)
+    , cdbDecodeHeader     :: forall s. Decoder s (Lazy.ByteString -> Header blk)
     , cdbDecodeLedger     :: forall s. Decoder s (LedgerState blk)
     , cdbDecodeChainState :: forall s. Decoder s (ChainState (BlockProtocol blk))
 
       -- Encoders
-    , cdbEncodeBlock      :: blk -> BinaryInfo Encoding
     , cdbEncodeHash       :: HeaderHash blk -> Encoding
+    , cdbEncodeBlock      :: blk -> BinaryInfo Encoding
+    , cdbEncodeHeader     :: Header blk -> Encoding
```
edsko: We should document that the encoder and decoder for headers include the envelope already. The chain DB will _only_ add the envelope if we are extracting the header from a block and it never _removes_ the envelope. This whole situation is a bit confusing, will be good to be quite explicit about it.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
```diff
@@ -226,34 +230,70 @@ mkImmDB immDB decHeader decBlock encBlock epochInfo isEBB addHdrEnv err = ImmDB
...
+
+    addHdrEnvelope
+      :: Either EpochNo SlotNo
+      -> Lazy.ByteString
+      -> (Either EpochNo SlotNo, Lazy.ByteString)
+    addHdrEnvelope epochOrSlot bs = (epochOrSlot, addHdrEnv isEBB bs)
+      where
+        isEBB = case epochOrSlot of
+          Left _epochNo  -> IsEBB
+          Right  _slotNo -> IsNotEBB
```
mrBliss: Two spaces

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
```diff
@@ -226,34 +230,70 @@ mkImmDB immDB decHeader decBlock encBlock epochInfo isEBB addHdrEnv err = ImmDB
...
+  -> m (Maybe b)
+getBlockOrHeaderWithPoint db blockOrHeader pt =
+    traverse deserialise =<<
+    getDeserialisableBlockOrHeaderWithPoint db blockOrHeader pt
+
+getDeserialisableBlockOrHeaderWithPoint
+  :: (MonadCatch m, HasCallStack)
+  => ImmDB m blk
+  -> BlockOrHeader blk b
+  -> Point blk
```
edsko: Let's create a ticket to revisit this setup to see if we can unify some of these variants (it's conceivable that `BlockComponent` could help.)

(Can get rid of some simple wrappers already now perhaps.)

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
```diff
@@ -230,6 +232,52 @@ mkImmDB immDB decHeader decBlock encBlock epochInfo isEBB addHdrEnv err = ImmDB
   Getting and parsing blocks
 -------------------------------------------------------------------------------}
 
+-- | Return 'True' when the given point is in the ImmutableDB.
+--
+-- This is much more efficient than 'getBlockOrHeaderWithPoint' as no block or
+-- header has to be read from disk.
+--
+-- If the point corresponds to some slot in the future, 'False' is returned.
```
edsko: It would be good to document why this change in behaviour compared to other functions is necessary (wrt future slots).

Because otherwise we could just use `GetWhat` here instantiated to hash or something like that, rather than introducing yet another function? (It's possible that this will need the `BlockComponent` changes first.)

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
edsko: 🤦‍♂

ouroboros-network/src/Ouroboros/Network/Protocol/ChainSync/Codec.hs
```diff
@@ -130,6 +130,20 @@ codecChainSync encodeHeader decodeHeader
...
+  -> (forall s . CBOR.Decoder s (LBS.ByteString -> header))
+  -> (Point header -> CBOR.Encoding)
+  -> (forall s . CBOR.Decoder s (Point header))
+  -> (tip -> CBOR.Encoding)
+  -> (forall s. CBOR.Decoder s tip)
+  -> Codec (ChainSync header tip)
+           CBOR.DeserialiseFailure m LBS.ByteString
+codecChainSync encodeHeader decodeHeader =
+    codecChainSyncUnwrapped encodeHeaderWrapped decodeHeaderWrapped
+  where
```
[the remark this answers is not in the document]
mrBliss: @dcoutts ^^

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Types.hs
```diff
@@ -176,12 +176,15 @@ data ChainDbEnv m blk = CDB
...
-    -- ^ The readers.
+  , cdbBlockReaders   :: !(StrictTVar m (Map ReaderId (StrictTVar m (ReaderState m blk blk))))
+    -- ^ The (block) readers.
     --
     -- INVARIANT: the 'readerPoint' of each reader is 'withinFragmentBounds'
     -- of the current chain fragment (retrieved 'cdbGetCurrentChain', not by
     -- reading 'cdbChain' directly).
+  , cdbHeaderReaders  :: !(StrictTVar m (Map ReaderId (StrictTVar m (ReaderState m blk (Header blk)))))
+    -- ^ The header readers. Same as the block readers ('cdbBlockReaders'),
+    -- but instead of returning whole blocks, they only return the headers.
```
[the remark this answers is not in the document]
mrBliss: In the ImmutableDB, we don't know about `blk`, only of `hash` and `ByteString`:

```haskell
-- ImmutableDB
data GetWhat hash res where
  GetBlock  :: GetWhat hash (hash, ByteString)
  GetHeader :: GetWhat hash (hash, ByteString)
  GetHash   :: GetWhat hash hash
```
vs.
```haskell
-- ChainDB
data BlockOrHeader blk b where
  Block  :: BlockOrHeader blk blk
  Header :: BlockOrHeader blk (Header blk)
```
So I don't think we should/can unify them. However, what if we make them more similar and translate from one to the other in the `ImmDB`?

```haskell
-- ImmutableDB
data BlockComponent hash b where
  Block  :: BlockComponent hash (hash, ByteString)
  Header :: BlockComponent hash (hash, ByteString)
  Hash   :: BlockComponent hash hash

-- ChainDB
data BlockComponent blk b where
  Block  :: BlockOrHeader blk blk
  Header :: BlockOrHeader blk (Header blk)
  Hash   :: BlockOrHeader blk (HeaderHash blk)
```

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
mrBliss: Hmm, it returns `False` instead of throwing a `ReadFuture*Error` because actually checking whether the block/EBB is not in the future is non-trivial and involves converting to/from slots/epochs. The former implementation of `getBlock(OrHeader)WithPoint` used to do it, but the current implementation no longer does it.

So I'm afraid that we can't simply use `GetHash` here, as it won't check the bounds. Note that we can't know upfront whether a `Point` is in the future, we have to access the ImmutableDB for that, in particular to know whether the `Point` corresponds to a block or slot. Alternatively, we could just catch the `ReadFuture*Error`. Would you be okay with that?

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Types.hs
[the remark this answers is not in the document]
mrBliss: I have created #1372 for this.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
mrBliss: I have created #1372 for this.

PR #1332, opened 2019-12-06
consensus: make PBFT serialisation more general

nfrisby:
I'm unaware of any specification that N >= K; this module's Haddock only says "The window size itself is pretty much arbitrary and will be fixed by a particular blockchain specification (e.g., Byron)."

> Without this patch, deserialization would not round-trip unless the window size
> N were at least as large as the security parameter K. That s true in the
> current implementation, but it is not necessary in general.


nfrisby:
Closing as WRONG. I was confused. It's only `inWindow` that has length `n`, both of the `pre <> post` concatenations have length `n + k`.

mrBliss:
> Closing as WRONG. I was confused. It's only `inWindow` that has length `n`, both of the `pre <> post` concatenations have length `n + k`.
> 
> (I've been working on this module's tests, which involve the "old" implementation, which did in fact only hold `n` signatures.)

:heavy_check_mark:

Issue #1296, opened 2019-11-26
`genesisSlotNo` is a misnomer


edsko:
Currently we have 

```haskell
genesisSlotNo :: SlotNo
genesisSlotNo = SlotNo 0
```

and friends (`genesisBlockNo` etc). This is a misnomer: this is not the slot number of the genesis block, but rather the slot number of the first _actual_ block (which, for Byron, happens to be an EBB, but for a sane chain would be the first normal block). After the introduction of `WithOrigin`, the `SlotNo` of the _genesis_ block is `Origin`.

edsko:
This got deleted in https://github.com/input-output-hk/cardano-base/pull/79 .

PR #1348, opened 2019-12-12
CAD-355: Move slots and Strict data structures into base/prelude

nc6:
`BlockNo` and `SlotNo` now live in `cardano-slotting` in `cardano-base`.

The strict fingertree and seq types, as well as semigroup actions, now live in `cardano-prelude`.

This PR removes them from `ouroboros-network` and re-exports the variants from `cardano-slotting`


ouroboros-network/src/Ouroboros/Network/Block.hs
```diff
@@ -54,7 +54,8 @@ module Ouroboros.Network.Block (
   , Serialised(..)
   ) where
 
-import           Cardano.Binary (ToCBOR (..))
+import           Cardano.Slotting.Block
+import           Cardano.Slotting.Slot (SlotNo(..), genesisSlotNo)
```
mrBliss: Move these down to L69, near `Cardano.Prelude`.

ouroboros-network/src/Ouroboros/Network/Block.hs
```diff
@@ -72,27 +72,6 @@ import           Ouroboros.Network.Point (WithOrigin (..), block, origin,
...
-
--- | The 0-based index of the block in the blockchain.
--- BlockNo is <= SlotNo and is only equal at slot N if there is a block
--- for every slot where N <= SlotNo.
-newtype BlockNo = BlockNo { unBlockNo :: Word64 }
-  deriving stock   (Show, Eq, Ord, Generic)
-  deriving newtype (Enum, Bounded, Num, Serialise, NoUnexpectedThunks)
-
-genesisSlotNo :: SlotNo
-genesisSlotNo = SlotNo 0
```
mrBliss: See #1296. That's now something that will have to be addressed in `cardano-base/prelude`.

ouroboros-network/src/Ouroboros/Network/Point.hs
```diff
@@ -19,40 +19,14 @@ module Ouroboros.Network.Point
 import           GHC.Generics (Generic)
 
 import           Cardano.Prelude (NoUnexpectedThunks)
-
+import           Cardano.Slotting.Slot
```
mrBliss: Move it one line up, so that it is grouped with `Cardano.Prelude`.

Actually, the `Codec.Serialise` import should be grouped with the `GHC.Generics` import.

I don't think we have defined this as a convention, but it is the de facto convention Edsko (I believe) and I are following.

ouroboros-network/src/Ouroboros/Network/Point.hs
```diff
@@ -19,40 +19,14 @@ module Ouroboros.Network.Point
...
   deriving (Eq, Ord, Show, Generic, NoUnexpectedThunks)
 
-at :: t -> WithOrigin t
-at = At
-
-origin :: WithOrigin t
-origin = Origin
-
 block :: slot -> hash -> WithOrigin (Block slot hash)
 block slot hash = at (Block slot hash)
```
mrBliss: Is this one even used (check other repos too)? If not, then I wouldn't mind removing it.

nc6: It is used, yes

Issue #1319, opened 2019-12-04
Use duplex sockets over NodeToNode Interface


karknu:
Mux bearers are bidirectional by default and we should take advantage of that fact in NodeToNode. This is especially important in p2p scenarios where it is possible that host A can open connections to host B, but not the other way around (due to NATs, Firewalls,...).

karknu:
This feature is needed for the shelley testnet because without it all our nodes would have to be manually configured with the ip address/domain name and port number of every stake pool in the test network.

coot:
This is done in `p2p-master`, we don't need to track it separately from `connection-manager`.

Issue #1318, opened 2019-12-04
Support starting and restarting of miniprotcols.


karknu:
Support starting and restarting of individual miniprotocols within a given mux bearer. This is needed to reduce resource consumption for p2p and it will make it possible to pause a chainsync client instead of having it bring down the entire Mux bearer incase the peer is too far behind.

karknu:
This feature is needed for the shelley testnet because without it chainsyc would trigger a disconnect for every peer that is more than `k` blocks behind once #1319 is implemented.

coot:
Done in #1536

coot:
Done #2221

PR #1349, opened 2019-12-12
Lazy and bidirectional Mux.

karknu:
This PR implements #1318 and #1319.

### DIFF
--- a/network-mux/src/Network/Mux.hs
+++ b/network-mux/src/Network/Mux.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE DataKinds           #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE RankNTypes          #-}
@@ -10,6 +10,8 @@ module Network.Mux (
       muxStart
     , muxBearerSetState
     , MuxSDU (..)
+    , MiniProtocolInitiatorControl (..)
+    , MiniProtocolResponderControl (..)
     , MiniProtocolLimits (..)
     , ProtocolEnum (..)
     , MiniProtocolId (..)
@@ -100,6 +102,8 @@ muxStart tracer peerid app bearer = do
     tbl <- setupTbl
     tq <- atomically $ newTBQueue 100
     cnt <- newTVarM 0
+    itbl <- setupInitiatorCtrlTbl
+    rtbl <- setupResponderCtrlTbl
 
     let pmss = PerMuxSS tbl tq bearer
         jobs = [ (demux pmss, "Demuxer")
@@ -107,11 +111,16 @@ muxStart tracer peerid app bearer = do
                , (muxControl pmss ModeResponder, "MuxControl Responder")
                , (muxControl pmss ModeInitiator, "MuxControl Initiator")
                ]
-    mjobs <- sequence [ mpsJob cnt pmss ptcl
+        cs = case app of
+                  MuxInitiatorApplication             i _     -> [ (i (mpsInitiatorCtrl itbl) >> muxExit cnt, "Initiator Continuation")]
+                  MuxInitiatorAndResponderApplication i _ r _ -> [ (i (mpsInitiatorCtrl itbl) >> muxExit cnt, "Initiator Continuation")
+                                                                 , (r (mpsResponderCtrl rtbl) >> muxExit cnt, "Responder Continuation")]
+                  MuxResponderApplication             r _     -> [ (r (mpsResponderCtrl rtbl) >> muxExit cnt, "Responder Continuation")]
+    mjobs <- sequence [ mpsJob cnt pmss itbl rtbl ptcl
                       | ptcl <- [minBound..maxBound] ]
 
     mask $ \unmask -> do
-      aidsAndNames <- traverse (\(io, name) -> (,name) <$> async (unmask io)) (jobs ++ concat mjobs)
+      aidsAndNames <- traverse (\(io, name) -> (,name) <$> async (unmask io)) (jobs ++ cs ++ concat mjobs)
 
       muxBearerSetState tracer bearer Mature
       unmask $ do
@@ -136,13 +145,55 @@ muxStart tracer peerid app bearer = do
                         | ptcl <- [minBound..maxBound]
                         , mode <- [ModeInitiator, ModeResponder] ]
 
+    -- Construct an array used for controlling when an initiator application should start and
+    -- a return value for the initiator applications
+    setupInitiatorCtrlTbl :: m (MiniProtocolInitiatorControlTable ptcl m a)
+    setupInitiatorCtrlTbl = MiniProtocolInitiatorControlTable
+             . array (minBound, maxBound)
+           <$> sequence [ do q <- atomically newEmptyTMVar
+                             p <- atomically newEmptyTMVar
+                             return (ptcl, (q, p))
+                        | ptcl <- [minBound..maxBound]
+                        ]
+
+    -- Construct an array used for providing the return value for responder applications
+    setupResponderCtrlTbl :: m (MiniProtocolResponderControlTable ptcl m b)
+    setupResponderCtrlTbl = MiniProtocolResponderControlTable
+             . array (minBound, maxBound)
+           <$> sequence [ do q <- atomically newEmptyTMVar
+                             return (ptcl, q)
+                        | ptcl <- [minBound..maxBound]
+                        ]
+
+    -- Provides two STM actions that is used to control Initiator miniprotocols.
+    mpsInitiatorCtrl :: MiniProtocolInitiatorControlTable ptcl m a
+                     -> ptcl
+                     -> MiniProtocolInitiatorControl m a
+    mpsInitiatorCtrl (MiniProtocolInitiatorControlTable tbl) mpdId =
+        let release = do
+                -- Signal to the miniprotocol that it is time to start.
+                putTMVar (fst $ tbl ! AppProtocolId mpdId) ()
+                return result
+            result  =
+                -- Wait for the miniprotocol to finish.
+                takeTMVar (snd $ tbl ! AppProtocolId mpdId) in
+        MiniProtocolInitiatorControl release
+
+    -- Provide a STM action that is used to fetch the result from Responder miniprotocols.
+    mpsResponderCtrl :: MiniProtocolResponderControlTable ptcl m b
+                     -> ptcl
+                     -> MiniProtocolResponderControl m b
+    mpsResponderCtrl (MiniProtocolResponderControlTable tbl) mpdId =
+        MiniProtocolResponderControl $ takeTMVar (tbl ! AppProtocolId mpdId)
 
     mpsJob
       :: StrictTVar m Int
       -> PerMuxSharedState ptcl m
+      -> MiniProtocolInitiatorControlTable ptcl m a
+      -> MiniProtocolResponderControlTable ptcl m b
       -> ptcl
       -> m [(m (), String)]
-    mpsJob cnt pmss mpdId = do
+    mpsJob cnt pmss itbl rtbl mpdId = do
 
         initiatorChannel <- muxChannel tracer
                               pmss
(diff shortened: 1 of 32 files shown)


network-mux/src/Network/Mux.hs
```diff
@@ -136,13 +145,55 @@ muxStart tracer peerid app bearer = do
...
                         , mode <- [ModeInitiator, ModeResponder] ]
 
+    -- Construct an array used for controlling when an initiator application should start and
+    -- a return value for the initiator applications
+    setupInitiatorCtrlTbl :: m (MiniProtocolInitiatorControlTable ptcl m a)
+    setupInitiatorCtrlTbl = MiniProtocolInitiatorControlTable
+             . array (minBound, maxBound)
+           <$> sequence [ do q <- atomically newEmptyTMVar
+                             p <- atomically newEmptyTMVar
+                             return (ptcl, (q, p))
```
coot: why not just `(ptcl, q, p)`?

network-mux/src/Network/Mux/Types.hs
```diff
@@ -150,6 +154,14 @@ instance (MiniProtocolLimits ptcl) => MiniProtocolLimits (MiniProtocolId ptcl) w
...
     maximumIngressQueue (AppProtocolId pid) = maximumIngressQueue pid
 
+-- | Control Interface for the Initiator side of Miniprotocols.
+-- The outer STM action is used to signal that the miniprotocol should start.
+-- The inner STM action provides the result of the miniprotocol.
+newtype MiniProtocolInitiatorControl m a = MiniProtocolInitiatorControl (STM m (STM m a))
+
+-- | STM action for reading the result of the responder side of a Miniprotocol.
+newtype MiniProtocolResponderControl m b = MiniProtocolResponderControl (STM m b)
+
```
coot: It would be easier to read this patch if these types where directly in `Mux.hs`.  They are internal so no need to expose them here.

network-mux/src/Network/Mux/Types.hs
```diff
@@ -158,6 +170,13 @@ newtype MiniProtocolDispatch ptcl m =
         MiniProtocolDispatch (Array (MiniProtocolId ptcl, MiniProtocolMode)
                                     (StrictTVar m BL.ByteString))
 
+newtype MiniProtocolInitiatorControlTable ptcl m a =
+        MiniProtocolInitiatorControlTable (Array (MiniProtocolId ptcl)
+                                                 (StrictTMVar m (), StrictTMVar m a))
+newtype MiniProtocolResponderControlTable ptcl m b =
+        MiniProtocolResponderControlTable (Array (MiniProtocolId ptcl)
+                                                 (StrictTMVar m b))
+
```
coot: The above comment applies here as well.

network-mux/src/Network/Mux.hs
coot: Never mind, I see that they land in `MiniProtocolInitiatorControlTable`

network-mux/src/Network/Mux/Types.hs
```diff
@@ -158,6 +170,13 @@ newtype MiniProtocolDispatch ptcl m =
         MiniProtocolDispatch (Array (MiniProtocolId ptcl, MiniProtocolMode)
                                     (StrictTVar m BL.ByteString))
 
+newtype MiniProtocolInitiatorControlTable ptcl m a =
+        MiniProtocolInitiatorControlTable (Array (MiniProtocolId ptcl)
+                                                 (StrictTMVar m (), StrictTMVar m a))
```
coot: add haddocks to document what both `StrictTMVar` are for

network-mux/src/Network/Mux.hs
```diff
@@ -157,26 +208,68 @@ muxStart tracer peerid app bearer = do
...
+      -> MiniProtocolResponderControlTable ptcl m b
+      -> ptcl
+      -> MuxApplication appType peerid ptcl m a b
+      -> Channel m
+      -> m ()
+    mpsResponderBrooder pmss (MiniProtocolResponderControlTable rtbl) mpdId rspApp channel = forever $ do
+        -- Wait for initiator until we startup
+        traceWith tracer $ MuxTraceBrooderResponderBrood (AppProtocolId mpdId)
+        atomically $ do
+            buf <- readTVar (ingressQueue (dispatchTable pmss) (AppProtocolId mpdId) ModeResponder)
```
coot: You might have fixed this in one of the next commits: here we read from a queue and if there was something we ignore it, which will cause a problem when later we start the miniprotocol (likely a decoding failure, or a wrong message or a timeout error).

network-mux/src/Network/Mux.hs
```diff
@@ -157,26 +208,68 @@ muxStart tracer peerid app bearer = do
...
-    mpsJobExit cnt = do
+    -- Start and restart responder side miniprotocols on demand.
+    mpsResponderBrooder
+      :: HasResponder appType ~ True
+      => PerMuxSharedState ptcl m
+      -> MiniProtocolResponderControlTable ptcl m b
+      -> ptcl
+      -> MuxApplication appType peerid ptcl m a b
+      -> Channel m
+      -> m ()
```
coot: You could return `Void` instead of `()`.  We use it in other places where `forever` or other infinite loops are used.

network-mux/src/Network/Mux.hs
```diff
@@ -157,26 +208,68 @@ muxStart tracer peerid app bearer = do
...
+            buf <- readTVar (ingressQueue (dispatchTable pmss) (AppProtocolId mpdId) ModeResponder)
+            when (buf == BL.empty)
+                retry
+        -- Initiator is active, start the responder side
+        traceWith tracer $ MuxTraceBrooderResponderHatch (AppProtocolId mpdId)
+        b <- responderApplication rspApp peerid mpdId channel
+        -- Responder side exited, wait until we should hatch another responder
+        traceWith tracer $ MuxTraceBrooderResponderDone (AppProtocolId mpdId)
+        -- Propagate the miniprotocol result.
+        atomically $ putTMVar (rtbl ! AppProtocolId mpdId) b
```
coot: We're putting values inside a `TMVar` in a `forever` loop.  It should be explained, who reads them?  They will block if not taken in a timely manner.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -538,3 +548,74 @@ localNetworkErrorPolicy = ErrorPolicies {
...
+    waitOnRestartQueue :: TBQueue m NodeToNodeProtocols -> STM m MiniProtocolClientResult
+    waitOnRestartQueue q = do
+        ptcl <- readTBQueue q
+        return $ MpcRestart ptcl
+
+    waitOnMiniProtocolClientResult :: (NodeToNodeProtocols, STM m a)
+                                    -> STM m MiniProtocolClientResult
+    waitOnMiniProtocolClientResult (ptcl, fetchResult) = do
+        void fetchResult
+        return $ MpcResult ptcl
```
coot: this could just be `MpsResult ptcl $> fetchResult`

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -538,3 +548,74 @@ localNetworkErrorPolicy = ErrorPolicies {
...
+
+    delayRestart :: TBQueue m NodeToNodeProtocols -> NodeToNodeProtocols -> m ()
+    delayRestart q ptcl = do
+        threadDelay restartDelay
+        atomically $ writeTBQueue q ptcl
+
+    waitOnRestartQueue :: TBQueue m NodeToNodeProtocols -> STM m MiniProtocolClientResult
+    waitOnRestartQueue q = do
+        ptcl <- readTBQueue q
+        return $ MpcRestart ptcl
```
coot: This could just be written as `MpcRestart <$> readTBQueue q`.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -538,3 +548,74 @@ localNetworkErrorPolicy = ErrorPolicies {
...
+        writeTBQueue startQ ChainSyncWithHeadersPtcl
+        writeTBQueue startQ BlockFetchPtcl
+        writeTBQueue startQ TxSubmissionPtcl
+    clientK startQ []
+
+  where
+    clientK :: TBQueue m NodeToNodeProtocols -> [(NodeToNodeProtocols, STM m a)] -> m ()
+    clientK restartQ resultActions = do
+        cr <- atomically $ waitOnRestartQueue restartQ
+                `orElse` foldr (orElse . waitOnMiniProtocolClientResult) retry resultActions
```
coot: I think you can turn this `foldr` into `asum . fmap waitOnMiniProcolClientResult`. `STM` has `Alternative` instance for which `<|>` is `orElse` and `empty` is `retry`.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -538,3 +548,74 @@ localNetworkErrorPolicy = ErrorPolicies {
...
+simpleResponderControl rspFn = forever $
+    -- Fetch, but ignore the result of any miniprotocol so that it can
+    -- be restarted incase the initiator decides to talk to us again.
+    void $ atomically $ fetchResponderResult ChainSyncWithHeadersPtcl
+               `orElse` fetchResponderResult BlockFetchPtcl
+               `orElse` fetchResponderResult TxSubmissionPtcl
+  where
+    fetchResponderResult :: MonadSTM m => NodeToNodeProtocols -> STM m a
+    fetchResponderResult ptcl =
+        let (MiniProtocolResponderControl action) = rspFn ptcl in
```
coot: I'd prefer `case rspFn ptcl of MiniProtocolResponderControl action -> action` - though it's just matter of taste.

It does not directly apply to Haskell, but in the core language: `case` expressions drive the evaluation and `let`s are where memory allocations happen.

ouroboros-network/demo/chain-sync.hs
```diff
@@ -212,6 +218,11 @@ serverPingPong = do
         codecPingPong
         (pingPongServerPeer pingPongServerStandard)
 
+    serverK rspFn = do
+        let (MiniProtocolResponderControl ppResult) = rspFn PingPong0
```
coot: it seems to be a common pattern to unpack the `MiniProtocolResponderControl` newtype (or `MiniProtocolInitiatorControl`).  Maybe provide a destructor:
```
newtype MiniProtocolInitiatorControl = MiniProtocolInitiatorControl { getMiniProtocolInitiatorControl :: ... }
```
```

File: network-mux/src/Network/Mux.hs [hunk 13]
```diff
@@ -136,13 +145,55 @@ muxStart tracer peerid app bearer = do
...
+                             return (ptcl, (q, p))
+                        | ptcl <- [minBound..maxBound]
+                        ]
+
+    -- Construct an array used for providing the return value for responder applications
+    setupResponderCtrlTbl :: m (MiniProtocolResponderControlTable ptcl m b)
+    setupResponderCtrlTbl = MiniProtocolResponderControlTable
+             . array (minBound, maxBound)
+           <$> sequence [ do q <- atomically newEmptyTMVar
+                             return (ptcl, q)
```
coot: why not `(ptcl, ) <$> newEmptyTMVarM`?

coot: I think one can rewrite `... <$> sequence ...` with `traverse`

network-mux/src/Network/Mux.hs
karknu: It is a `readTVar` not `readTMvar`, no data is actually consumed.

network-mux/src/Network/Mux.hs
karknu: The values are consumed by the `MiniProtocolResponderControl` stm operation specified in the Mux application. I'll add a reference to it.

network-mux/src/Network/Mux/Types.hs
karknu: They are not internal to `Mux.hs`. They are part of the mux interface, see Interface.hs.

network-mux/src/Network/Mux/Types.hs
karknu: Will address.

network-mux/src/Network/Mux/Types.hs
karknu: Will address.

network-mux/src/Network/Mux.hs
```diff
@@ -157,26 +208,68 @@ muxStart tracer peerid app bearer = do
                               cnt
 
         return $ case app of
-          MuxInitiatorApplication initiator ->
-            [ ( initiator peerid mpdId initiatorChannel >> mpsJobExit cnt
+          MuxInitiatorApplication _ _ ->
```
coot: You can use `MuxInitiatorApplication {}` to ignore all the fields.  It should be future proof as you pass `app` to the next function.

ouroboros-network/src/Ouroboros/Network/Mux.hs
```diff
@@ -35,30 +37,38 @@ import           Ouroboros.Network.Channel
 --
 data OuroborosApplication (appType :: AppType) peerid ptcl m bytes a b where
      OuroborosInitiatorApplication
-       :: (peerid -> ptcl -> Channel m bytes -> m a)
+       :: ((ptcl -> MiniProtocolInitiatorControl m a) -> m ())
```
coot: Please expand haddock documentation of this type to include the role and purpose of this field.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
karknu: I wrote this without `<|>` and `Applicative` since it meant that I didn't have to force a constraint that was purely for ascetic reasons  on to the users of this function.

coot: @dcoutts I think it's worth adding `Alternative` to `MonadSTM` constraints:
```
class (Monad m, Monad (STM m), Alternative (STM m)) => MonadSTM m
```
Possibly also `MonadPlus`.

network-mux/src/Network/Mux.hs
karknu: Its is written in the same style as `setupTbl`, and I don't see a reason to change it.

karknu: > why not `(ptcl, ) <$> newEmptyTMVarM`?

Will address.

network-mux/src/Network/Mux.hs
karknu: Will fix.

network-mux/src/Network/Mux.hs
karknu: It needs to match the type of other mux jobs, some of which can return.

ouroboros-network/src/Ouroboros/Network/Mux.hs
karknu: Will do.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
karknu: Will address.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
karknu: Will address.

network-mux/src/Network/Mux/Types.hs
coot: Still I would put them in `Mux.hs` and then we can export them from `Mux.Interface`.

At some point we'll need to refactor how things are exported, `Network.Mux` ought to be the interface module and `Network.Mux.Muxxer` or `.Internal` could have the implementation.

karknu: There is a PR for refactoring Mux, #1247 which moves declarations around.
This PR conforms to how types are declared right now.

ouroboros-network/demo/chain-sync.hs
karknu: Will address.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
karknu: `let` isn't used any more after the addition of `getMiniProtocolResponderControl`.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
karknu: I'm not going to address this. To me `foldr` and `orElse` is more readable.

PR #1353, opened 2019-12-14
Generate EBBs more realistically in the RealPBFT consensus test

### LINKED ISSUE #966: Include mock EBBs
As #872, except for `test-consensus` instead of `test-storage`. Likely useful overlap between the two.

This Issue partly represents the obligation to write a more specific test for PR #951 (Issue #925).

### LINKED ISSUE #1352: ChainSync client cannot rollback to an EBB that shares a slot with an actual block
In this code

https://github.com/input-output-hk/ouroboros-network/blob/e9453b316dd47d54ec640cac312d6bf79957b31d/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs#L332-L334

Suppose `intersection :: Point blk` is an EBB and there's also a proper block after it in that same slot.

While the `AF.rollback` can use the hash to distinguish the two blocks, the `rewindChainState` only has the slot number and not the hash and so will incorrectly retain the block after the EBB even though that block is after the intersection point.

nfrisby:
Fixes #966. Fixes #1352.

I'm opening this as a Draft PR because I would like your feedback on the basic schemes before I polish it up. Basic schemes:

  * Issue #966 - Use a new `Tracer` in `Node.Tracers` and a new `TVar` in `Test.Dynamic.Tracer` to ensure that each node has added all EBBs to its `ChainDB` _before_ forging a block when leading. Blocking the traced thread seems like an abuse of `Tracer`s, but I haven't come up with a less invasive option. We also had to update `preferCandidate` to care about length-after-intersection in addition to block number, since it was just ignoring the EBBs we were adding: except the epoch 0 EBB, they'd only be selected once their successor arrived -- but that meant they never got a successor! That in turn required some gymnastics in BlockFetch to ensure we only compare chain fragments that intersect.

  * Nodes always forge an actual block when leading; they never waste their turn forging an EBB.

  * `suchThat` to preclude node join plans that cause unpredictable final chains. The ambiguity is currently limited to at most the first `numCoreNodes` blocks, but that is enough to cause trouble because it could affect the PBft signature threshold, which could subvert the "at least `k` blocks in `2k` slots" invariant established by the RealPBFT `NodeJoinPlan` generator. (I tried to integrate this check into the recursive generator itself, but the inherent non-monotonicity spoils the inductive hypothesis that "at the very least the plan is OK if all remaining nodes ASAP".)

  * Issue #1352 - when the CS client rollsback a candidate fragment and the oldest dropped header has the same slot as the target, then the target is an EBB with an accompanying block in the same slot and so the primary chain state should essentially be rolled back to the preceding slot instead of the EBB's slot. Otherwise we'd retain the block sharing the slot, which the intersection point precedes. We address this by storing the header hash of an EBB in the chain state and taking a `Point` as the target of rewinds. We retain the EBB when rolling back to its slot if its hash matches that of the given `Point`. As an unfortunate compromise, we store the `HeaderHash` as a `ByteString`, since a sufficient argument for `HeaderHash` is not made available by the arguments to the `ChainState` family.


ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -767,6 +763,28 @@ rejectInvalidBlocks tracer registry getIsInvalidBlock getCandidate =
...
+
+    let slot = pointSlot i
+    let isAccompaniedEBB = case AF.successorBlock i ourFrag of
+          Nothing  -> False
+            -- intersection might be an EBB, but not an accompanied one
+          Just hdr -> At (blockSlot hdr) == slot
+    cs' <- rewindChainState cfg ourChainState slot isAccompaniedEBB
+
+    pure (frag', cs')
+
```
mrBliss: I'm thinking about the alternative solution you proposed when we first ran into the EBB+PBFT.ChainState problem:

* Make the ChainSyncClient EBB-aware, i.e. let it take a `Header blk -> IsEBB` function.
* In `rollback`: if `isEBB i == IsEBB`, then look at the predecessor (I'll write `AF.predecessorBlock`) and use its `SlotNo` to rewind the chain state. In case of the Genesis EBB, rewind to `Origin`. *Problem:* what if the rollback point is the anchor point of the fragment? Then we can't look at its predecessor. Maybe it's fine to throw `InvalidRollBack` (or a new exception constructor?) in that case. I'll think some more about this.
* In `PBFT.ChainState`: remove all traces from EBBs (and make Edsko happy again :slightly_smiling_face: ). (And avoid #1356! :smile:) 
* No need to touch the signature of `rewindChainState`.
* Modify the decoder of `PBFT.ChainState` to ignore the EBB field and increase the version number.

Am I forgetting something? I'm thinking now I made the wrong decision the first time :grin:. The main downside for me is that the ChainSyncClient is now EBB-aware, which feels a bit dirty, as EBBs are specific to one consensus protocol.

nfrisby: Hmm.

> The main downside for me is that the ChainSyncClient is now EBB-aware, which feels a bit dirty, as EBBs are specific to one consensus protocol.

I agree: it seems better for the PBFT ChainState to be complicated by EBBs, since PBFT actually involves EBBs.

Moreover, what's the main upside? The one listed is

> No need to touch the signature of `rewindChainState`.

but that's not particularly compelling? In this workaround, I've added an "is EBB" argument to `rewindChainState`. We could instead change its `SlotNo` argument to a `Point` and have the chain state retain hashes. That's also be viable, and wouldn't explicitly involve EBBs (it'd only explicitly allow for multiple blocks in a slot, which we (and the PBFT code) know currently only happens with EBBs.)

nfrisby: By the way, the `ebbs` field could be abstracted out of the `PBFT` chain state, if we wanted to recover the "pure PBFT" chain state.

mrBliss: > Hmm.
> 
> > The main downside for me is that the ChainSyncClient is now EBB-aware, which feels a bit dirty, as EBBs are specific to one consensus protocol.
> 
> I agree: it seems better for the PBFT ChainState to be complicated by EBBs, since PBFT actually involves EBBs.
> 
> Moreover, what's the main upside? The one listed is
> 
> > No need to touch the signature of `rewindChainState`.

True. My first draft of that comment included "Much simpler .. blah blah", but then I added the "*Problem*" part and removed it.

> but that's not particularly compelling? In this workaround, I've added an "is EBB" argument to `rewindChainState`. We could instead change its `SlotNo` argument to a `Point` and have the chain state retain hashes. That'd also be viable and wouldn't explicitly involve EBBs (it'd only explicitly allow for multiple blocks in a slot, which we (and the PBFT code) know currently only happens with EBBs.)

I like that idea :+1: Let me think about it.

mrBliss: > > [..] In this workaround, I've added an "is EBB" argument to `rewindChainState`. We could instead change its `SlotNo` argument to a `Point` and have the chain state retain hashes. That'd also be viable and wouldn't explicitly involve EBBs (it'd only explicitly allow for multiple blocks in a slot, which we (and the PBFT code) know currently only happens with EBBs.)
> 
> I like that idea +1 Let me think about it.

Concretely, this means that `PBFT.ChainState` will track the hash of each applied EBB, so we can intercept a `rewindChainState` call to an EBB. The hash is ignored for regular blocks. Hmm, it feels a bit wrong to pass in the hash if we're not using it to check whether it matches a previously applied header. Not that I'm arguing we should, as that would mean that we'd have to store the hashes too, requiring even more bytes, and breaking binary compatibility. It is less ad-hoc than the `Bool` argument you added to `rewindChainState`.

ouroboros-consensus/test-consensus/Test/Dynamic/Network.hs
```diff
@@ -228,6 +234,32 @@ runNodeNetwork registry testBtime numCoreNodes nodeJoinPlan nodeTopology
           simChaChaT varDRG id $ testGenTxs numCoreNodes cfg ledger
         void $ addTxs mempool txs
 
+    ebbProducer :: HasCallStack
+                => StrictTVar m SlotNo
+                -> NodeConfig (BlockProtocol blk)
+                -> ChainDB.ChainDB m blk
+                -> EpochInfo m
+                -> m ()
+    ebbProducer v cfg chainDB epochInfo = go 0
```
mrBliss: Can you rename `v` to `nextEbbSlotVar`?

ouroboros-consensus/test-consensus/Test/Dynamic/General.hs
```diff
@@ -125,10 +126,12 @@ runTestNetwork ::
      )
   => (CoreNodeId -> ProtocolInfo blk)
   -> TestConfig
+  -> Maybe (NodeConfig (BlockProtocol blk) -> SlotNo -> BlockNo -> ChainHash blk -> blk)
```
mrBliss: I'd like a type synonym for this, e.g., `EBBProducer blk` or `EbbProducer blk`.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -4,6 +4,7 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts      #-}
 {-# LANGUAGE NamedFieldPuns        #-}
+{-# LANGUAGE PatternGuards         #-}
```
mrBliss: Is this pragma actually needed? I have been writing pattern guards without it.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -400,7 +401,9 @@ forkBlockProduction IS{..} =
                         -> (Point blk, BlockNo)
     prevPointAndBlockNo slot c = case c of
         Empty _   -> (genesisPoint, genesisBlockNo)
-        c' :> hdr -> case blockSlot hdr `compare` slot of
+        c' :> hdr
+          | Just{} <- nodeIsEBB hdr -> (headerPoint hdr, blockNo hdr)
```
mrBliss: This can be a bit more explicit: check that the `blockSlot` is not `> slot` + add a comment saying that we allow forging a block in the same slot as an EBB.

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
nfrisby: If you're ready to commit to that route, would you update Issue 1352 (which blocks this PR)? And would you want to open that PR yourself or shall I?

ouroboros-consensus/test-consensus/Test/Dynamic/Network.hs
```diff
@@ -228,6 +234,32 @@ runNodeNetwork registry testBtime numCoreNodes nodeJoinPlan nodeTopology
...
+          void $ blockUntilSlot btime ebbSlotNo
+
+          (prevSlot, ebbBlockNo, prevHash) <- ChainDB.getTipHeader chainDB <&> \case
+            Nothing  -> (Origin            , 0,           GenesisHash)
+            Just hdr -> (At (blockSlot hdr), blockNo hdr, BlockHash (headerHash hdr))
+          when (prevSlot < At ebbSlotNo) $ case mbForgeEBB of
+            Nothing       -> pure ()
+            Just forgeEBB -> do
+              let ebb = forgeEBB cfg ebbSlotNo ebbBlockNo prevHash
+              ChainDB.addBlock chainDB ebb
```
nfrisby: I just realized that this `addBlock` achieves nothing for epochs after epoch 0.

https://github.com/input-output-hk/ouroboros-network/blob/602856e08e0ebe9d63751185d39580d3b22989c1/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ChainSel.hs#L377-L379

So:

  * The ChainDB will only select an EBB if that EBB already has a successor (with an exception for EBB0 since it's predecessor is `Origin`).

But that can never happen since

  * The node only forges blocks onto the current tip of the ChainDB.

(We have neither the chicken nor the egg!)

To test with EBBs in later epochs, we'll need to relax/override one of those bullet points.

nfrisby:
I've pushed up three new commits.

  * b783ef68 - test-consensus: fixup PR 966 no EBBs more than 2k later

Without this, late joining nodes could forge multiple EBBs on top of Origin. They're 10k slots apart, which means the ">= k blocks in 2k slots" was violated.

  * 333ea012 - test-consensus: check for all expected EBBs

I noticed that I was only seeing EBBs for Epoch 0. So I added this test to confirm. This test fails without the next commit.

  * ad8199d5 - TENTATIVE consensus: PBFT preferCandidate no longer ignores EBBs

I'm unfamiliar with the consequences here, but this override in `OuroborosTag PBFT` seems to suffice at first glance to allow us to select EBBs for later epochs.

```
  -- we override the default so that the ChainDB will select a new EBB
  -- /promptly/
  preferCandidate _ ours cand =
    case AF.compareHeadBlockNo ours cand of
      LT -> True
      EQ -> case AF.intersect ours cand of
        Nothing                   -> False
        Just (_, _, ours', cand') -> AF.length ours' < AF.length cand'
      GT -> False
```

I haven't formulated it crisply yet, but I think there's an invariant along the lines of "we only forge EBBs in the correct slots and every node forges EBBs", so the presence/absence of EBBs will never be a deciding factor in the tests.

I suspect this change should probably be isolated in a protocol combinator that the `RealPBFT` test uses.

ouroboros-consensus/test-consensus/Test/Dynamic/Network.hs
nfrisby: I may have found another option. See the third commit discussed at https://github.com/input-output-hk/ouroboros-network/pull/1353#issuecomment-566223460

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: Ah; it seems they are part of Haskell2010: http://www.haskell.org/onlinereport/haskell2010/haskellch3.html#x8-460003.13

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: Fixed, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: Fixed, thanks.

ouroboros-consensus/test-consensus/Test/Dynamic/Network.hs
nfrisby: Fixed, thanks.

ouroboros-consensus/test-consensus/Test/Dynamic/General.hs
nfrisby: Fixed, thanks. I went with a bespoke variant of `Maybe`, so the `Nothing` would be more legible.

(I still intend to repackage these arguments in a record type, so I might undo that then. But these PR seems to have enough moving parts as-is.)

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
mrBliss: It's not ideal, but yeah, let's go for it. I'll leave the PR up to you :slightly_smiling_face:

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/ModChainSel.hs
```diff
@@ -67,3 +70,26 @@ instance (Typeable p, Typeable s, ChainSelection p s) => OuroborosTag (ModChainS
...
+
+{-------------------------------------------------------------------------------
+  Individual modifiers
+-------------------------------------------------------------------------------}
+
+-- | Override 'preferCandidate' such that the ChainDB will select a new EBB
+-- /promptly/.
+--
+-- Uses the default 'compareCandidates' from the definition of 'OuroborosTag'.
+data SelectEBBsPromptly
```
mrBliss: I'm not entirely convinced we should add this as a modifier. My thoughts:
* Edsko recently removed most modifiers as they were making things more complex.
* (Similar:) while modifiers can be elegant, maybe it would be simpler if we just inlined it in `PBFT`?
* ~Only RealPBFT needs to deal with EBBs, so the combinator is not really reusable anyway.~ We could use it for MockPBFT. Moreover, EBBs will always be in the chain, so all future protocols will have to deal with it, at least in case of corruptions, such that the chain is truncated into an era that still has EBBs. I suppose this means that we should use `compareHeadBlockNoThenLength` (see comment below) as the default for all protocols.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/ModChainSel.hs
```diff
@@ -67,3 +70,26 @@ instance (Typeable p, Typeable s, ChainSelection p s) => OuroborosTag (ModChainS
 
 instance OuroborosTag p => NoUnexpectedThunks (NodeConfig (ModChainSel p s))
   -- use generic instance
+
+{-------------------------------------------------------------------------------
+  Individual modifiers
+-------------------------------------------------------------------------------}
+
+-- | Override 'preferCandidate' such that the ChainDB will select a new EBB
+-- /promptly/.
```
mrBliss: I wouldn't mention ChainDB as this is about chain selection in general.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/ModChainSel.hs
nfrisby: We chatted on Slack. Thomas indeed wants to adopt this improved `preferCandidates` (with a matching `compareCandidates`) in `PBFT` itself. Seems like the previous half-support for EBBs (ie "we'll only select an EBB once its successor arrives") was itself a compromise, and the new `intersection`-based `preferCandidates` is itself desirable.

Thus: no combinator: just bake it in `PBFT` and also update `compareCandidates`.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/ModChainSel.hs
nfrisby: Fixed, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/ModChainSel.hs
nfrisby: OK, I updated the default `compareCandidates` to compare by block number and then by length-after-intersection. All protocols use this default.

I also updated the default `preferCandidates` to delegate to `compareCandidates`. All protocols except Praos use this default, but I also updated its custom `preferCandidates` to use `compareCandidates`.

I also needed to update `BlockFetch.Decision.ChainSuffix` to carry around the original candidate so that we could compare based on that field. Before this, it only carried the blocks after the intersection with the current chain, which might not themselves intersect. This new field is guaranteed to intersect among the candidates, since this is downstream of the "filter plausible" etc. After this, it seems like the new PRECONDITION on `compareCandidates` (that they intersect, at the very least when the block number is equal) is satisfied.

mrBliss:
Edsko and I came up with a better plan:

Instead of computing the intersection and then the lengths of the suffices, we can instead simply check whether one of the headers is an EBB or not. We can use `headerPBftFields :: cfg -> hdr -> Maybe ..` for this, as it returns a `Just` for a regular header and `Nothing` for an EBB header. So more concretely:
```haskell
-- Look at the tips
case (AF.head ours, AF.head cand) of
    -- Both empty
    (Left _,       Left _)        -> EQ
    -- Our chain is empty, candidate is not
    (Left _,       Right _)       -> LT
    -- Our chain is not empty, the candidate is
    (Right _,      Left _)        -> GT
    -- Our chain and candidate not empty
    (Right ourHdr, Right candHdr) ->
      -- Prefer the highest block number, as it is a proxy for chain length
      case blockNo ourHdr `compare` blockNo candHdr of
        LT -> LT
        GT -> GT
        -- If the block numbers are the same, check if one of them is an EBB.
        -- An EBB has the same block number as the block before it, so the
        -- chain ending with an EBB is actually longer than the one ending with a
        -- regular block. Note that `headerPBftFields` returns `Nothing` for
        -- an EBB.
        EQ ->
          case (headerPBftFields cfg ourHdr, headerPBftFields cfg candHdr) of
            (Just _,  Just _)  -> EQ
            (Just _,  Nothing) -> LT
            (Nothing, Nothing) -> EQ
            (Nothing, Just _)  -> GT
```
This only needs to be done for `PBFT`.

This also means that we don't need the extra invariants and don't have to change BlockFetch.

Furthermore, this is compatible with Edsko's plan to look only at the headers at the tip (#1227) instead of the fragments.

nfrisby:
I've force-pushed some new work. Summary:

  * All tests pass (since I updated the expected result for the golden test for a serialised PBFT ChainState). These tests include exercising Issue 1352 (I removed the list comprehension guard that avoided it in the PBFT ChainState tests) and also confirm that the final chains can include EBBs for multiple epochs.

  * Commit 4ac8528c implements the new `preferCandidates` you and Edsko most recently suggested.

  * Commit 124ab0dd implements the new PBFT ChainState scheme that allows rewinding to an EBB that shares a slot with a non-EBB (i.e. Issue 1352). This commit uses `ByteString` to workaround the fact that the `ChainState` type cannot depend on the block/header type.

I'm not excited about the `ByteString`-based approach, but:

  * It proves out the basic approach of tracking EBB's hashes in the ChainState in order to fix Issue 1352.

  * I used a heavily restricted newtype to ensure the `HeaderHash` is indeed what's being serialised and compared.

  * The alternative we last discussed was to use an existentially quantified `headerHash` type with a `Typeable` constraint. I spent a while trying different approaches with this but ultimately took the `ByteString` escape hatch since there were so many obstacles. In particular, I'm not sure we can deserialize the `ChainState` if it existentially quantifies one of the types that occurs in it. In other words, we need to know the `HeaderHash` type before deserializing the `ChainState`...

  * ... And if that's the case, then I suspect we might as well just go ahead add the `block` or `hdr` type parameter to the `ChainState` family itself. Then everything would work nicely and without needing existentials. For example `RunMockProtocol` would mostly be folded into `RunMockBlock` because `ChainState p` would become `ChainState p (SimpleHeader c ext)`.

```
-- | The part of 'RunMock' that depends only on @p@
class RunMockProtocol p where
  mockProtocolMagicId  ::           NodeConfig p -> ProtocolMagicId
  mockEncodeChainState ::           NodeConfig p -> ChainState p -> Encoding
  mockDecodeChainState :: forall s. NodeConfig p -> Decoder s (ChainState p)

-- | Protocol specific functionality required to run consensus with mock blocks
class RunMockProtocol p => RunMockBlock p c ext where
  ...
```

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
```diff
@@ -224,6 +226,7 @@ test_golden_ChainState = goldenTestCBOR
     secParam = SecurityParam 2
     windowSize = CS.WindowSize 3
     signers = map (`CS.PBftSigner` CC.exampleKeyHash) [1..5]
+    hashBytes = mkHeaderHashBytesForTestingOnly (Lazy8.pack "test_golden_ChainState6")
```
mrBliss: Why `test_golden_ChainState6`?

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
```diff
@@ -215,7 +217,7 @@ test_golden_ChainState = goldenTestCBOR
     exampleChainState
     "test-consensus/golden/cbor/byron/ChainState"
   where
-    exampleChainState = CS.appendEBB secParam windowSize 6 $
+    exampleChainState = CS.appendEBB secParam windowSize 6 hashBytes $
```
mrBliss: We should have another test in the other direction: we should be able to decode the chain state. We can have one of these for each chain state version.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -236,7 +239,8 @@ invariant (SecurityParam k)
 
     allEbbs p =
       Map.null $
-      Map.filterWithKey (\slot mSlot -> not (p slot mSlot)) (unEbbMap ebbs)
+      Map.filterWithKey (\slot (EbbInfo _ mSlot) -> not (p slot mSlot)) $
+      unEbbMap ebbs
```
mrBliss: Don't forget about #1357. Run the tests with the `checktvarinvariant` flag on the `io-sim-classes` package to see whether we're still affected by the strange leak.

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
mrBliss: I have pushed a commit that checks this.

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
```diff
@@ -208,22 +215,56 @@ prop_byronHashInfo_hashSize h =
...
+      (decodeByronChainState secParam)
+      exampleChainStateWithoutEBB
+      "test-consensus/golden/cbor/byron/ChainState0"
+
+test_golden_ChainState_backwardsCompat_version1 :: TestTree
+test_golden_ChainState_backwardsCompat_version1 =
+    testCase "ChainState version 1" $ goldenTestCBORBackwardsCompat
+      (decodeByronChainState secParam)
+      exampleChainStateWithoutEBB
+      "test-consensus/golden/cbor/byron/ChainState1"
```
nfrisby: @mrBliss This tests the ability to deserialize a V1 blob that arose with an empty `EbbMap`. Should we also test a V1 blob that ha a non-empty `EbbMap`? It'll be a bit tedius to generate one. And if I understand correctly we've probably never actually generated one in practice, right? Since we have no recent EBBs.

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
nfrisby: It's so the "hash" is indeed "specific" to the EBB. It's likely overkill, but I think it conveys an important subtlety. I added signatures and a comment that try to emphasize that.

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
nfrisby: Resolution: implemented.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/ModChainSel.hs
```diff
@@ -67,3 +70,26 @@ instance (Typeable p, Typeable s, ChainSelection p s) => OuroborosTag (ModChainS
...
+
+  preferCandidate' _ _ ours cand =
+    case AF.compareHeadBlockNo ours cand of
+      LT -> True
+      EQ -> case AF.intersect ours cand of
+        Nothing                   -> False
+        Just (_, _, ours', cand') -> AF.length ours' < AF.length cand'
+      GT -> False
+
+  compareCandidates' _ _ = AF.compareHeadBlockNo
```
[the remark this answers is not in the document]
nfrisby: Resolution: Edsko and Thomas came up with a simpler scheme.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/ModChainSel.hs
nfrisby: Resolution: we backed this out and used the simpler scheme that Edsko and Thomas came up with.

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -330,8 +330,9 @@ chainSyncClient mkPipelineDecision0 tracer cfg btime
         -- guarantees that the ChainSync protocol /does/ in fact give us a
         -- switch-to-fork instead of a true rollback.
         (theirFrag, theirChainState) <-
-          case (,) <$> AF.rollback (castPoint intersection) ourFrag
-                   <*> rewindChainState cfg ourChainState (pointSlot intersection)
+          let i = castPoint intersection in
```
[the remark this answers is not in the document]
nfrisby: Done. I had to move the `of` up to the previous line.

ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
```diff
@@ -635,8 +636,9 @@ chainSyncClient mkPipelineDecision0 tracer cfg btime
                    , ourTip
                    } -> traceException $ do
       (theirFrag', theirChainState') <-
-        case (,) <$> AF.rollback (castPoint intersection) theirFrag
-                 <*> rewindChainState cfg theirChainState (pointSlot intersection)
+        let i = castPoint intersection in
```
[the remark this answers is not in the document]
nfrisby: Done. I had to move the `of` up to the previous line.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -189,10 +188,10 @@ class ( Show (ChainState    p)
   -- blocks.
   --
   -- This function should attempt to rewind the chain state to the state at some
-  -- given slot, or Origin to rewind to the state with no blocks.
+  -- given point.
   --
-  -- PRECONDITION: the slot to rewind to must correspond to the slot of a
-  -- header (or 'Origin') that was previously applied to the chain state using
+  -- PRECONDITION: the point to rewind to must correspond to a header (or
+  -- 'Origin') that was previously applied to the chain state using
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -50,8 +52,8 @@ import           Cardano.Crypto.DSIGN.Class
 import           Cardano.Prelude (NoUnexpectedThunks)
 
 import qualified Ouroboros.Network.AnchoredFragment as AF
-import           Ouroboros.Network.Block (HasHeader (..), SlotNo (..))
-import           Ouroboros.Network.Point (WithOrigin (At))
+import           Ouroboros.Network.Block (pattern BlockPoint, pattern GenesisPoint, HasHeader (..), HeaderHash, Point, SlotNo (..))
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

nfrisby:
The push `ea2bfdb3...4cb2df4f` is just squashing.

nfrisby:
Commit `948503e2` replaces the `Map` of EBB info in the chain state with just a `Maybe`.

nfrisby:
The push 9430dbc5...11a745c4 squashes `9430dbc5 - consensus: store info for at most one EBB in PBFT ChainState` into `4cb2df4f - consensus: store EBB header hash bytes in PBFT ChainState`.

nfrisby:
If I run with `checktvarinvariant` set to `True` and the `RealPBFT` consensus tests disabled, then everything passes. If I try to run the `RealPBFT` consensus tests, I see hundreds of thousands (~a million at this moment) of `closurePtrs: Cannot handle type CONSTR_NOCAF yet` messages. These seem to swamp the process; I'm not sure if it will finish, but I'm letting it run now.

mrBliss:
> If I run with `checktvarinvariant` set to `True` and the `RealPBFT` consensus tests disabled, then everything passes. If I try to run the `RealPBFT` consensus tests, I see hundreds of thousands (~a million at this moment) of `closurePtrs: Cannot handle type CONSTR_NOCAF yet` messages. These seem to swamp the process; I'm not sure if it will finish, but I'm letting it run now.
> 
> Edit: It did pass; 3million lines of output later.

Yeah, that's expected.

ouroboros-consensus/test-consensus/Test/Consensus/Ledger/Byron.hs
mrBliss: Meh, it would be nice, but I don't care much about it.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -353,15 +366,14 @@ append :: forall c. PBftCrypto c
...
     PBftChainState {
         preAnchor  = preAnchor'
       , postAnchor = postAnchor'
       , preWindow  = preWindow'
       , inWindow   = inWindow'
       , counts     = updateCounts counts
-        -- NOTE: 'pruneEBBsLT' is inlined here to avoid a strange space leak
-        -- that also goes away with @-O0@, see #1356.
-      , ebbs       = EbbMap $ Map.filter (>= anchorSlot') (unEbbMap ebbs)
+      , ebbs       = ebbs   -- TODO needs to be pruned
```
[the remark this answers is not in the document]
nfrisby: Ah, yes; that was mindless of me.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -221,16 +221,12 @@ class ( Show (ChainState    p)
...
 
-  default preferCandidate :: HasHeader hdr
-                          => NodeConfig p
-                          -> AnchoredFragment hdr      -- ^ Our chain
-                          -> AnchoredFragment hdr      --` ^ Candidate
-                          -> Bool
-  preferCandidate _ ours cand =
-    AF.compareHeadBlockNo cand ours == GT
+  preferCandidate cfg ours cand =
+    compareCandidates cfg ours cand == LT
```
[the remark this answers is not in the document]
nfrisby: > Shouldn't it be `cand > ours` instead of `ours < cand`?

I think this change was a refactor because of the usual property that `y > x` if and only if `x < y`. Am I wrong to assume that that property of `compareCandidates`? (I'm a bit confused by your comment about `EQ` -- `preferCandidate` before and after the change is `False` if `ours` and `cand` compare as `EQ`.)

My motivation for the change was because I misread the original at a glance: I wasn't expecting the arguments of the function to switch positions in the delegation.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -275,6 +276,28 @@ instance ( PBftCrypto c
...
+      (Right lHdr, Right rHdr) ->
+        -- Prefer the highest block number, as it is a proxy for chain length
+        case blockNo lHdr `compare` blockNo rHdr of
+          LT -> LT
+          GT -> GT
+          -- If the block numbers are the same, check if one of them is an EBB.
+          -- An EBB has the same block number as the block before it, so the
+          -- chain ending with an EBB is actually longer than the one ending
+          -- with a regular block. Note that `headerPBftFields` returns
+          -- `Nothing` for an EBB.
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

nfrisby:
I rebased onto `origin/master` after `fixup`ing a couple commits based on @mrBliss's most recent comments.

nfrisby:
The build failed on the `BlockchainTime.delayNextSlot` test. I've seen this fail several times locally, but this is the first time I've seen it fail on CI.

I'll rebase onto `origin/master` in order to trigger a full rebuild.

nfrisby:
`bors` finally merged something into `master`, so I rebased onto that.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
nfrisby: I ran this on this same day you commented; it passed. The `INLINE` pragma I added seems sufficient.

nfrisby:
I rebased onto `origin/master` in order to resolve conflicts with PR 1227.

nfrisby:
I rebased onto master, now that PR 1385 (slot lengths) was merged.

@mrBliss @edsko Is there any other planned work we're waiting on before merging this PR, or should I continue rebasing it as master advances? Thanks.

mrBliss:
> I rebased onto master, now that PR 1385 (slot lengths) was merged.
> 
> @mrBliss @edsko Is there any other planned work we're waiting on before merging this PR, or should I continue rebasing it as master advances? Thanks.
> 
> Edit: DO NOT MERGE YET. Ach. [8d27d41](https://github.com/input-output-hk/ouroboros-network/commit/8d27d411df7a8d825ee98f4e5feee65cdda58dfd) fails locally (but not on CI??) with `ResourceRegistryUsedFromUnknownThread`. Investigating now.
>
> 
> Edit 2: I pushed [98249d4](https://github.com/input-output-hk/ouroboros-network/commit/98249d45bf1b173fb409c3ffd9a3727c4e8e3c86). @mrBliss, does that look right? Would you like to cherry-pick it to a separate PR?

That was my fault, I fixed it in #1389. In that same PR I had to make some changes in the consensus (temporary kludge), so you'll have to rebase on that :grin:. Otherwise, I don't expect any other upcoming conflicts.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
```diff
@@ -157,7 +156,11 @@ class ( Show (ChainState    p)
                   -> hdr      -- ^ Tip of the candidate
                   -> Bool
 
-  -- | Compare two candidates
+  -- | Compare two candidates, both of which we prefer to our own chain
+  --
+  -- Note: we make no assumptions about the anchor points of the fragments,
+  -- except indirectly such that non-empty fragments are greater than empty
+  -- fragments and empty fragments are indistinguishable.
```
mrBliss: This is out of date since #1227.

nfrisby:
I rebased onto master. This picks up PR 1389, and drops commit 98249d4.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/Abstract.hs
nfrisby: Fixed with a rebase, thanks.

ouroboros-consensus/test-consensus/Test/Dynamic/Network.hs
```diff
@@ -248,6 +262,41 @@ runNodeNetwork numCoreNodes numSlots slotLengths nodeJoinPlan nodeTopology
...
+            JustForgeEBB forgeEBB -> do
+              (prevSlot, ebbBlockNo, prevHash) <- atomically $ do
+                p <- ChainDB.getTipPoint chainDB
+                let mSlot = pointSlot p
+                let k = SlotNo $ maxRollbacks $ protocolSecurityParam cfg
+                check $ case mSlot of
+                  Origin -> True
+                  At s   -> s >= (ebbSlotNo - min ebbSlotNo (2 * k))
+                bno <- ChainDB.getTipBlockNo chainDB
+                pure $ (mSlot, bno, pointHash p)
```
mrBliss: Nitpick: redundant `$`

ouroboros-consensus/test-consensus/Test/Dynamic/Network.hs
```diff
@@ -365,9 +414,21 @@ runNodeNetwork numCoreNodes numSlots slotLengths nodeJoinPlan nodeTopology
...
+
       let nodeArgs = NodeArgs
             { tracers             = nullDebugTracers
-                { forgeTracer = nodeEventsForges nodeInfoEvents
+                { forgeTracer       = Tracer $ \case
+                    TraceForgeAboutToLead s -> do
+                      atomically $ do
+                        lim <- readTVar nextEbbSlotVar
+                        check $ s < lim
+                    o -> traceWith (nodeEventsForges nodeInfoEvents) o
```
mrBliss: What if we want to make EBBs optional? Or even stop producing after some epoch?
It doesn't have to be part of this PR, but it would nice to be able to do that in the future.

nfrisby: If `runNodeNetwork` is invoked with `NothingForgeEBB` then there will be no EBBs, but this `nextEbbSlotVar` will still increment -- there's a comment saying as much at the binding site of `nextEbbSlotVar`. It's slightly off-putting, but seemed preferable to having multiple `Maybe`s, etc.

No support for stopping EBBs at a certain slot yet.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/ChainState.hs
```diff
@@ -609,53 +638,90 @@ appendEBB :: forall c. (PBftCrypto c, HasCallStack)
...
+-- rewind point.
 pruneEBBsLT :: PBftChainState c -> PBftChainState c
 pruneEBBsLT cs@PBftChainState{..} =
-  cs{ebbs = EbbMap $ Map.filter (>= anchorSlot cs) (unEbbMap ebbs)}
-
--- | Discard 'ebbs' mappings whose /key/ is after the given slot
+  cs{ ebbs = ebbsFilter ebbs $ \EbbInfo{..} ->
+        eiPrevSlot >= anchorSlot cs ||
+        At eiSlot == anchorSlot cs }
+-- NOTE: this INLINE seems redudant but we add it here to avoid a strange space
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

ouroboros-consensus/test-consensus/Test/Dynamic/Network.hs
nfrisby: Fixed, thanks.

nfrisby:
Two rebases just now: first `fixup`ed the minor suggested edits, and second rebased onto `origin/master`. We're so close! :)

nfrisby:
Another rebase: I exploded the `address PR comments` commit into `fixup`s for appropriate preceding commits. I also confirmed that `test-consensus` successfully builds after each of the commits on this PR.

```
$ git diff nfrisby/issue-966-more-EBBs origin/nfrisby/issue-966-more-EBBs 
nfrisby@frisbycomp:~/iohk$ git push --force-with-lease origin nfrisby/issue-966-more-EBBs
Counting objects: 126, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (113/113), done.
Writing objects: 100% (126/126), 27.16 KiB | 0 bytes/s, done.
Total 126 (delta 73), reused 0 (delta 0)
remote: Resolving deltas: 100% (73/73), completed with 34 local objects.
To github.com-iohk:input-output-hk/ouroboros-network.git
 + 41372f2c...36c8ab23 nfrisby/issue-966-more-EBBs -> nfrisby/issue-966-more-EBBs (forced update)
nfrisby@frisbycomp:~/iohk$ 
```

nfrisby:
🎉🎉🎉

PR #1361, opened 2019-12-16
MonadSTM - Alternative and MonadPlus constraints

coot:


avieth:
This is interesting. It's definitely good to always have `Alternative` on the STM monad, but what we *really* want is not just any old `Alternative`, but rather

```
empty = retry
(<|>) = orElse
```

and a similar story for `MonadPlus`.

With the way things are presented now, I don't think it's possible to express this. In order to define a `MonadSTM m` the programmer _must_ define an `Alternative (STM m)`, _could_ choose anything for this instance, but morally should not even be required to give one!

Here's a different approach: give a proper `STM` type with all of the instances that the real `STM` has (`MonadFix`, `MonadPlus`, `MArray TArray e STM` omitted here) and use a type family for the underlying representation. The definition for proper IO should then give no runtime overhead cost, right?

```Haskell
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE UndecidableInstances #-}

import Control.Applicative (Alternative (..))
import qualified Control.Concurrent.STM as RTS
import Data.Kind (Type)
import Data.Proxy (Proxy (..))

class HasSTM (impl :: k) where

  type STM_Rep  impl :: Type -> Type

  -- Proxies could be got rid of by making STM_Rep injective
  retry  :: Proxy impl -> STM_Rep impl t
  orElse :: Proxy impl -> STM_Rep impl t -> STM_Rep impl t -> STM_Rep impl t

newtype STM (impl :: k) t = STM
  { runSTM :: STM_Rep impl t }

-- The functor/applicative/monad comes from the underlying representation
deriving instance Functor (STM_Rep impl) => Functor (STM impl)
deriving instance Applicative (STM_Rep impl) => Applicative (STM impl)
deriving instance Monad (STM_Rep impl) => Monad (STM impl)

-- The alternative/monadplus comes from orElse/retry
instance (HasSTM impl, Applicative (STM_Rep impl)) => Alternative (STM impl) where
  empty = STM (retry (Proxy :: Proxy impl))
  left <|> right = STM (orElse (Proxy :: Proxy impl) (runSTM left) (runSTM right))

data RTS

instance HasSTM RTS where

  type STM_Rep RTS = RTS.STM

  retry  _ = RTS.retry
  orElse _ = RTS.orElse
```

PR #1365, opened 2019-12-18
Clean cabal.project file

coot:
* consolidate `constraints`
* organise `package` options together

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -9,20 +9,12 @@ packages: ./typed-protocols
           ./ntp-client
           ./Win32-network
 
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/cardano-prelude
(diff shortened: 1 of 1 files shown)


cabal.project
```diff
@@ -45,6 +37,21 @@ package ouroboros-network
...
+  type: git
+  location: https://github.com/input-output-hk/cardano-prelude
+  tag: 9352518fd9f957ab5e9611a2e27f495339c597d1
+  subdir: .
+
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/cardano-prelude
+  tag: 9352518fd9f957ab5e9611a2e27f495339c597d1
+  subdir: .
```
mrBliss: Twice the same repo + subdir

coot: Good catch, I carried this from the current `cabal.project`.

coot: It's even repeated three times :D

PR #1371, opened 2019-12-18
Make block production optional

mrBliss:
Controlled through `IsProducer`.


mrBliss:
I'll update the node and proxy after this is merged.

PR #1373, opened 2019-12-18
Update dependencies

mrBliss:


mrBliss:
@coot Any reason why #1365 hasn't been merged yet? I can rebase this PR after merging it.

intricate:
> @coot Any reason why #1365 hasn't been merged yet? I can rebase this PR after merging it.

#1365 is merged now.

mrBliss:
CI fails:
```
    Golden tests
      ChainState:                                         OK
```
It suspect changes to `cardano-ledger`, but I'll investigate it.

If we blindly update our golden test, the new version may not be able to load old ledger snapshots. It can handle that, but it'll have to rebuild the ledger from scratch on start-up, which takes ~10 min, so we prefer to avoid that. We (consensus) have been very careful to avoid this, so we shouldn't just let all our effort go to waste. If it is inevitable, then we could think about simplifying #1353, which is being careful to avoid it.

mrBliss:
> CI fails:
> 
> ```
>     Golden tests
>       ChainState:                                         OK
>       LedgerState:                                        FAIL
> ```
> 
> It suspect changes to `cardano-ledger`, but I'll investigate it.
> 
> If we blindly update our golden test, the new version may not be able to load old ledger snapshots. It can handle that, but it'll have to rebuild the ledger from scratch on start-up, which takes ~10 min, so we prefer to avoid that. We (consensus) have been very careful to avoid this, so we shouldn't just let all our effort go to waste. If it is inevitable, then we could think about simplifying #1353, which is being careful to avoid it.

The golden test of the ledger state changed because the dummy genesis changed.

PR #1374, opened 2019-12-19
Add missing ChainDB parameter

mrBliss:


mrBliss:
Nodes and proxies using the current master of `ouroboros-network` will fail to open the ChainDB because a parameter is missing.

We really need a better approach to handle default parameters so that at least get a type error if we forget a required parameter.

PR #1378, opened 2019-12-24
Optimise chain serving

mrBliss:
This PR drastically reduces the CPU usage while serving headers and blocks.

I went over the profiling report and optimised what I could.

The main improvement comes from the following:

> [In `ChainDB.Iterator`] we were only passing a start bound to an `ImmutableDB.Iterator`, never an end bound. We would close the iterator manually. This was fine with the old database format, as an iterator would only load a small index in memory. Now with the new database format, the whole secondary index of the epoch is loaded into memory, even if only a few blocks from it needed to be streamed.

This means that for each incoming BlockFetch request, regardless its size (in practice, the requested batches typically only contain 1 or 2 blocks), we read and parsed all secondary index entries starting from the first block to the end of the epoch (so on average, 21600/2 entries!), instead of just the necessary secondary index entries for the requested batch.

We now apply a proper end bound and also *cache the indices in memory*.

The end result is that when a "client" is syncing at 100% CPU, the "server" is only uses a few % CPU.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
@@ -150,7 +150,7 @@ swapMVar StrictMVar { tmvar, tvar, invariant } !a = do
 isEmptyMVar :: MonadSTM m => StrictMVar m a -> m Bool
 isEmptyMVar StrictMVar { tmvar } = atomically $ Lazy.isEmptyTMVar tmvar
 
-updateMVar :: MonadSTM m => StrictMVar m a -> (a -> (a, b)) -> m b
+updateMVar :: (MonadSTM m, HasCallStack) => StrictMVar m a -> (a -> (a, b)) -> m b
 updateMVar StrictMVar { tmvar, tvar, invariant } f = do
     (a', b) <- atomically $ do
         a <- Lazy.takeTMVar tmvar
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE BangPatterns      #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NamedFieldPuns    #-}
+{-# LANGUAGE TupleSections     #-}
 
 module Ouroboros.Consensus.Util.MonadSTM.StrictMVar (
     StrictMVar(..) -- constructors exported for benefit of tests
@@ -18,6 +19,8 @@ module Ouroboros.Consensus.Util.MonadSTM.StrictMVar (
   , swapMVar
   , isEmptyMVar
   , updateMVar
+  , updateMVar_
+  , modifyMVar
   ) where
 
 import           Control.Concurrent.STM (readTVarIO)
@@ -25,6 +28,8 @@ import           Control.Monad (when)
 import           Control.Monad.Class.MonadSTM (MonadSTM (..))
 import qualified Control.Monad.Class.MonadSTM as Lazy
 import           Control.Monad.Class.MonadSTM.Strict (checkInvariant)
+import           Control.Monad.Class.MonadThrow (ExitCase (..), MonadCatch,
+                     generalBracket)
 import           GHC.Stack
 
 import           Cardano.Prelude (NoUnexpectedThunks (..))
@@ -159,6 +164,20 @@ updateMVar StrictMVar { tmvar, tvar, invariant } f = do
         Lazy.writeTVar tvar a'
         return (a', b)
     checkInvariant (invariant a') $ return b
+
+updateMVar_ :: (MonadSTM m, HasCallStack) => StrictMVar m a -> (a -> a) -> m ()
+updateMVar_ var f = updateMVar var ((, ()) . f)
+
+modifyMVar :: (MonadSTM m, MonadCatch m, HasCallStack)
+           => StrictMVar m a -> (a -> m (a, b)) -> m b
+modifyMVar var action =
+    snd . fst <$> generalBracket (takeMVar var) putBack action
+  where
+    putBack a ec = case ec of
+      ExitCaseSuccess (a', _) -> putMVar var a'
+      ExitCaseException _ex   -> putMVar var a
+      ExitCaseAbort           -> putMVar var a
+
 
 {-------------------------------------------------------------------------------
   NoUnexpectedThunks
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
@@ -52,6 +52,7 @@ module Ouroboros.Storage.ChainDB.Impl.ImmDB (
   , ImmDB.ImmutableDBError
   , ImmDB.BinaryInfo (..)
   , ImmDB.HashInfo (..)
+  , Index.CacheConfig (..)
     -- * Exported for testing purposes
   , openDB
   , mkImmDB
@@ -101,6 +102,8 @@ import           Ouroboros.Storage.ImmutableDB (BinaryInfo (..), HashInfo (..),
                      ImmutableDB, Iterator (Iterator), IteratorResult (..),
                      WithHash (..))
 import qualified Ouroboros.Storage.ImmutableDB as ImmDB
+import qualified Ouroboros.Storage.ImmutableDB.Impl.Index as Index
+                     (CacheConfig (..))
 import qualified Ouroboros.Storage.ImmutableDB.Parser as ImmDB
 import           Ouroboros.Storage.Util.ErrorHandling (ErrorHandling)
 import qualified Ouroboros.Storage.Util.ErrorHandling as EH
@@ -160,6 +163,7 @@ data ImmDbArgs m blk = forall h. ImmDbArgs {
     , immAddHdrEnv      :: IsEBB -> Lazy.ByteString -> Lazy.ByteString
     , immHasFS          :: HasFS m h
     , immTracer         :: Tracer m (TraceEvent blk)
+    , immCacheConfig    :: Index.CacheConfig
     }
 
 -- | Default arguments when using the 'IO' monad
@@ -181,6 +185,7 @@ defaultArgs :: FilePath -> ImmDbArgs IO blk
 defaultArgs fp = ImmDbArgs{
       immErr          = EH.exceptions
     , immHasFS        = ioHasFS $ MountPoint (fp </> "immutable")
+    , immCacheConfig  = cacheConfig
       -- Fields without a default
     , immDecodeHash     = error "no default for immDecodeHash"
     , immDecodeBlock    = error "no default for immDecodeBlock"
@@ -195,6 +200,18 @@ defaultArgs fp = ImmDbArgs{
     , immAddHdrEnv      = error "no default for immAddHdrEnv"
     , immTracer         = nullTracer
     }
+  where
+    -- Cache 250 past epochs by default. This will take roughly 250 MB of RAM.
+    -- At the time of writing (1/2020), there are 166 epochs, so even one year
+    -- from now, we will be able to cache all epochs' indices in the chain.
+    --
+    -- If this number were too low, i.e., less than the number of epochs that
+    -- that clients are requesting blocks from, we would constantly evict and
+    -- reparse indices, causing a much higher CPU load.
+    cacheConfig = Index.CacheConfig
+      { pastEpochsToCache = 250
+      , expireUnusedAfter = 5 * 60 -- Expire after 1 minute
+      }
 
 withImmDB :: (IOLike m, HasHeader blk)
           => ImmDbArgs m blk -> (ImmDB m blk -> m a) -> m a
@@ -217,6 +234,7 @@ openDB registry ImmDbArgs{..} = do
       immValidation
       parser
       immTracer
+      immCacheConfig
     return ImmDB
       { immDB     = immDB
       , decHeader = immDecodeHeader
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
@@ -30,7 +30,7 @@ module Ouroboros.Storage.ChainDB.Impl.ImmDB (
     -- * Appending a block
   , appendBlock
     -- * Streaming
-  , streamFrom
+  , stream
   , streamAfter
   , deserialiseIterator
   , deserialisableIterator
@@ -68,6 +68,7 @@ import           Control.Tracer (Tracer, nullTracer)
 import           Data.Bifunctor (second)
 import qualified Data.ByteString.Lazy as Lazy
 import           Data.Functor ((<&>))
+import           Data.Functor (($>))
 import           GHC.Stack (HasCallStack)
 import           System.FilePath ((</>))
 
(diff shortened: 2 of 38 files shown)


ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Iterator.hs
```diff
@@ -295,26 +295,16 @@ newIterator itEnv@IteratorEnv{..} getItEnv registry from to = do
...
-      streamFromImmDBHelper True
-
-    streamFromImmDBHelper
-      :: HasCallStack
-      => Bool -- ^ Check the hash of the upper bound
-      -> ExceptT (UnknownRange blk) m (DeserialisableIterator m blk)
-    streamFromImmDBHelper checkUpperBound = do
-        -- First check whether the block in the ImmDB at the end bound has the
-        -- correct hash.
-        when checkUpperBound $
```
mrBliss: We were checking the upper bound!

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Iterator.hs
```diff
@@ -387,6 +387,8 @@ iteratorNextImpl dbEnv it@IteratorHandle {itHasFS = hasFS :: HasFS m h, ..}
     ImmutableDBEnv { _dbErr, _dbEpochInfo, _dbHashInfo } = dbEnv
     HasFS { hClose } = hasFS
 
+    -- | We don't rely on the position of the handle, we always use
+    -- 'hGetExactlyAt', i.e. @pread@ for reading from a given offset.
```
mrBliss: `hGetExactlyAtCRC`

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,615 @@
...
+--
+-- We only care about the ordering.
+newtype LastUsed = LastUsed Word64
+  deriving newtype (Eq, Ord, Enum, NoUnexpectedThunks)
+
+data Cached hash = Cached
+  { currentEpoch     :: !EpochNo
+  , currentEpochInfo :: !(CurrentEpochInfo hash)
+  , pastEpochsInfo   :: !(HashPSQ EpochNo LastUsed (PastEpochInfo hash))
+  , nbPastEpochs     :: Word
```
mrBliss: Missing bang

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,615 @@
...
+  }
+  deriving (Generic, NoUnexpectedThunks)
+
+-- | The last time a cached epoch was accessed.
+--
+-- We only care about the ordering.
+newtype LastUsed = LastUsed Word64
+  deriving newtype (Eq, Ord, Enum, NoUnexpectedThunks)
+
+data Cached hash = Cached
```
mrBliss: Mention LRU

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,615 @@
...
+      Nothing                -> (Nothing, Nothing)
+      Just (_lastUsed, info) -> (Just info, Just (lastUsed, info))
+
+openEpoch
+  :: EpochNo
+  -> CurrentEpochInfo hash
+  -> Cached hash
+  -> Cached hash
+openEpoch epoch newCurrentEpochInfo cached
+    | currentEpoch == epoch || currentEpoch + 1 == epoch
```
edsko: It deserves a comment somewhere that the current epoch isn't in the map.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,615 @@
...
+      , lastUsed         = succ lastUsed
+      , nbPastEpochs     = succ nbPastEpochs
+      }
+    | otherwise
+    = error $ "Going from epoch " <> show currentEpoch <> " to " <> show epoch
+  where
+    Cached
+      { currentEpoch, currentEpochInfo, pastEpochsInfo, lastUsed
+      , nbPastEpochs } = cached
+    pastEpochsInfo' = PSQ.insert currentEpoch lastUsed
```
edsko: Use of `lastUsed` needs justification here.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,615 @@
...
+        -- Cache hit for the current epoch
+        putMVar cacheVar cached $> Left currentEpochInfo
+      | Just (pastEpochInfo, cached') <- lookupPastEpochInfo epoch cached ->
+        -- Cache hit for an epoch in the past
+        putMVar cacheVar cached' $> Right pastEpochInfo
+      | otherwise -> do
+          -- Cache miss for an epoch in the past, load both entire indices for
+          -- the epoch from disk. First put back @cacheVar@ so we're not
+          -- hogging the cache.
+          putMVar cacheVar cached -- TODO bracket?
```
edsko: For sure.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,615 @@
...
+          modifyMVar cacheVar $
+            evictIfNecessary pastEpochsToCache .
+            addPastEpochInfo epoch pastEpochInfo
+          return $ Right pastEpochInfo
+
+{------------------------------------------------------------------------------
+  Operations
+------------------------------------------------------------------------------}
+
+forget
```
edsko: Confusing name; this isn't forgetting a particular epoch. Rename to `reset`.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,615 @@
...
+              currentEpochEntries currentEpochInfo Seq.|> entry
+          }
+        }
+
+{------------------------------------------------------------------------------
+  Helpers
+------------------------------------------------------------------------------}
+
+-- | Take items until the condition is true, but include the item for which
+-- the condition is true.
```
edsko: Precondition: the element is present.

edsko: (Or weaken the postcondition mentioned in the docstring.)

edsko: Maybe add an example also, this is a slightly confusing function.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,615 @@
...
+  Orphans
+------------------------------------------------------------------------------}
+
+instance ( NoUnexpectedThunks k
+         , NoUnexpectedThunks p
+         , NoUnexpectedThunks v
+         , Hashable k
+         , Ord k
+         , Ord p
+         ) => NoUnexpectedThunks (HashPSQ k p v) where
```
edsko: I think we have a module specifically for orphans somewhere?

edsko:
One thought I had: in my comments above, I worried that once the caching limit was exceeded by even just 1, performance of the entire server would suddenly be dramatically worse. Perhaps the right answer here is to simply disallow this? A bit akin to the OS complaining "not enough open file handles" or something like that. For example, if we set the limit at 100, the server could easily support a whole bunch of clients near the tip of the chain, since they will all use the last few epochs, which will all be cached, and could hen support up to , say, 97 or so clients that are syncing from scratch or near scratch, but if we are already supporting say 97 clients all syncing from scratch, we would then simply reject the next client trying to sync from scratch (unless they happen to be able to be in sync with another client); they'd get a "server busy" message or something like that.

This feels like the right thing to do to me. @mrBliss , @dcoutts , thoughts on this?

ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -160,6 +162,9 @@ updateMVar StrictMVar { tmvar, tvar, invariant } f = do
         return (a', b)
     checkInvariant (invariant a') $ return b
 
+modifyMVar :: (MonadSTM m, HasCallStack) => StrictMVar m a -> (a -> a) -> m ()
```
mrBliss: Rename to `updateMVar_`

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/State.hs
```diff
@@ -276,28 +270,21 @@ withOpenState ImmutableDBEnv { _dbHasFS = hasFS :: HasFS m h, .. } action = do
...
            -> m r
     access (DbClosed _) = throwUserError _dbErr ClosedDBError
     access (DbOpen ost) = action hasFS ost
 
-closeOpenHandles :: Monad m => HasFS m h -> InternalState m hash h -> m ()
-closeOpenHandles hasFS = \case
-    DbClosed _       -> return ()
-    DbOpen openState -> closeOpenStateHandles hasFS openState
-
 -- TODO what if this fails? Use a 'ResourceRegistry'?
```
mrBliss: Update comment: if one of the `hClose`s fails, it will bubble up and the bracketed `withRegistry` will close the registry and thus the remaining handles.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs
```diff
@@ -195,6 +200,11 @@ defaultArgs fp = ImmDbArgs{
     , immAddHdrEnv      = error "no default for immAddHdrEnv"
     , immTracer         = nullTracer
     }
+  where
+    cacheConfig = Index.CacheConfig
+      { pastEpochsToCache = 100
+      , expireUnusedAfter = 1 * 60 -- Expire after 1 minute
```
mrBliss: Change `pastEpochsToCache` to 250. Takes roughly 250 MB RAM. There are currently 166 epochs, so even one year from now, we will be able to cache all epochs in the chain. If this number were too low, i.e., less than the number of epochs that clients are requesting blocks from, we would constantly evict and reparse indices, causing a much higher CPU load.

Change `expireUnusedAfter` to 5 minutes.

mrBliss: Make sure this type is exported so that it could be changed in `cardano-node` if necessary.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,818 @@
...
+    -- Also return the 'LastUsed' of the /next least recently used/ past
+    -- epoch. When no past epoch was removed, this corresponds to the least
+    -- recently used past epoch in the /given/ 'Cached'. When a past epoch was
+    -- removed, this corresponds to the least recently used past epoch in the
+    -- /returned/ 'Cached'. If there are no more past epochs in the respective
+    -- 'Cached', return 'Nothing'.
+    --
+    -- When a 'TracePastEpochsExpired' event should traced, it is included in
+    -- the result.
+    removeLeastRecentlyUsed
```
mrBliss: Rename it to `garbageCollect`

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,818 @@
...
+        -- looking again. We don't have to expire anything before that time.
+        Nothing                               -> threadDelay expireUnusedAfter
+        -- Wait until the @nextLeastRecentlyUsed@ will be expired. Maybe it
+        -- will have been used again by the time our wait is over, in that
+        -- case, we won't expire it (and wait again).
+        Just (LastUsed nextLeastRecentlyUsed) -> do
+          now' <- getMonotonicTime
+          let expiredAt  = expireUnusedAfter `addTime` nextLeastRecentlyUsed
+              timeToWait = expiredAt `diffTime` now'
+          when (timeToWait > 0) $ threadDelay timeToWait
```
mrBliss: Remove this, just wait for `expireUnusedAfter`

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -0,0 +1,818 @@
...
+          PSQ.atMostView expiredLastUsedTime pastEpochsInfo
+
+        nbPastEpochs' = nbPastEpochs - fromIntegral (length expiredPastEpochs)
+
+        cached' = cached
+          { pastEpochsInfo = pastEpochsInfo'
+          , nbPastEpochs   = nbPastEpochs'
+          }
+
+        traceMsg = TracePastEpochsExpired
```
mrBliss: Put a bang here

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ImmutableDB/StateMachine.hs
```diff
@@ -1332,3 +1334,14 @@ internalUnused = error "semantics and internal DB API used during command genera
...
+
+instance Arbitrary Index.CacheConfig where
+  arbitrary = do
+    pastEpochsToCache <- frequency
+      -- Pick small values so that we exercise cache eviction
+      [ (1, return 1)
+      , (1, return 2)
+      , (1, choose (3, 10))
+      ]
+    expireUnusedAfter <- (fromIntegral :: Int -> DiffTime) <$> choose (1, 100)
```
mrBliss: Add a TODO/ticket about creating a Cmd that advances time

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ImmutableDB/StateMachine.hs
```diff
@@ -1279,7 +1282,22 @@ prop_sequential cacheConfig = forAllCommands smUnused Nothing $ \cmds -> QC.mona
           (hist, model, res) <- runCommands sm' cmds
           trace <- QC.run getTrace
           QC.monitor $ counterexample ("Trace: " <> unlines (map show trace))
-          fsDump <- QC.run $ Mock.pretty <$> atomically (readTVar fsVar)
+          fs <- QC.run $ atomically $ readTVar fsVar
+          let fsDump = Mock.pretty fs
+              openHandles = Mock.numOpenHandles fs
+              maxExpectedOpenHandles = 1 {- epoch file -}
+                                     + 1 {- primary index file -}
+                                     + 1 {- secondary -}
```
mrBliss: ... index file

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ImmutableDB/StateMachine.hs
```diff
@@ -1279,7 +1282,22 @@ prop_sequential cacheConfig = forAllCommands smUnused Nothing $ \cmds -> QC.mona
           (hist, model, res) <- runCommands sm' cmds
           trace <- QC.run getTrace
           QC.monitor $ counterexample ("Trace: " <> unlines (map show trace))
-          fsDump <- QC.run $ Mock.pretty <$> atomically (readTVar fsVar)
+          fs <- QC.run $ atomically $ readTVar fsVar
+          let fsDump = Mock.pretty fs
+              openHandles = Mock.numOpenHandles fs
+              maxExpectedOpenHandles = 1 {- epoch file -}
```
mrBliss: Explain these

PR #1379, opened 2019-12-26
VolatileDB: fix memory leak

mrBliss:
The VolatileDB maintains a number of in-memory indices among which:

    type SuccessorsIndex blockId = Map (WithOrigin blockId) (Set blockId)

During garbage collection, this index was not being culled correctly: when the last `blockId` was removed from a `Set`, the empty `Set` was kept in the `Map` instead of removing the entry altogether.


mrBliss:
Uggh, bors fails because of some [CI flakiness](https://hydra.iohk.io/build/1531410/nixlog/1)

PR #1389, opened 2020-01-03
ImmutableDB: use the ChainDB's registry instead of an internal one

mrBliss:
This fixes master.

See the separate commits for more details

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/Run/Mock.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/Run/Mock.hs
@@ -18,6 +18,10 @@ import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Ledger.Mock
 import           Ouroboros.Consensus.Ledger.Mock.Run
 import           Ouroboros.Consensus.Node.Run.Abstract
+import           Ouroboros.Consensus.Protocol.Abstract (SecurityParam (..),
+                     protocolSecurityParam)
+
+import           Ouroboros.Storage.Common (EpochSize (..))
 
 {-------------------------------------------------------------------------------
   RunNode instance for the mock ledger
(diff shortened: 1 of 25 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node/Run/Mock.hs
```diff
@@ -38,7 +42,8 @@ instance ( ProtocolLedgerView (SimpleBlock SimpleMockCrypto ext)
   nodeBlockMatchesHeader  = matchesSimpleHeader
   nodeBlockFetchSize      = fromIntegral . simpleBlockSize . simpleHeaderStd
   nodeIsEBB               = const Nothing
-  nodeEpochSize           = \_ _ _ -> return 100
+  nodeEpochSize           = \_ cfg _ -> return $
+    EpochSize $ 10 * maxRollbacks (protocolSecurityParam cfg)
```
nfrisby: Would it be more consistent to rely on `Cardano.Chain.ProtocolConstants.kEpochSlots` here?

edsko: The mock ledger should I think be able to make independent decisions?

nfrisby: I agree it should, but do we want it to here? My intuition is to minimize differences between tests and production where possible.

edsko: ... Or should we avoid having the code too dependent on specific values..? :)

nfrisby: Well right now they're both `10 * k` :)

edsko: Fair point, let's change that ;D

PR #1401, opened 2020-01-06
Refactor consensus tests

nfrisby:
This PR addresses various minor "we should eventually do X" comments that accrued on past PRs affecting the `test-consensus` tests.

Overview of changes:

  * The argument record changes eliminate some noise and improve some signatures.
  * The indentation changes are mostly for legibility and consistency.
  * I also improve the names of test trees, for example.
  * I renamed `*.Dynamic.*` to `*.ThreadNet.*` -- please suggest a better term if you have one.


nfrisby:
@mrBliss Would you like to see that first commit broken up? (It's the raw result of me figuring out where I wanted to end up.) If so, any broader course corrections to address before I do that? Thanks.

PR #1402, opened 2020-01-06
Set immRegistry field of ImmDbArgs in db-analyse

intricate:
At the moment, when trying to execute `analyse-db`, this is the output:

```
analyse-db: no default for immRegistry
  error, called at src/Ouroboros/Storage/ChainDB/Impl/ImmDB.hs:204:27...
```

This PR resolves that by ensuring that we set the `immRegistry` field of the `ImmDbArgs` constructed in `db-analyse`.


mrBliss:
Uggh, these `defaultArgs` again. We really should fix that.

Issue #1431, opened 2020-01-09
Add a function that can return a `MempoolSize` from a `MempoolSnapshot`


ksaric:
We need to include some information about the mempool in some of the trace events.

For example, we would like to trace events about the time it takes the block to be forged and there we have access to the `MempoolSnapshot` - https://github.com/input-output-hk/ouroboros-network/blob/966b34daed4a89e47e45f99a980bea69b463a1c8/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs#L315-L341

@dcoutts idea was to add a function that can return `MempoolSize` from a `MempoolSnapshot`.

PR #1432, opened 2020-01-09
[CAD-402] Adding changes to the trace events we need and removing time from them

ksaric:
https://jira.iohk.io/browse/CAD-402

Let me just copy the description from there:
```
The current issue (https://jira.iohk.io/browse/CAD-314) was measuring too much.

We need to start the micro benchmarking from small and precise steps, the opposite that I started with.

A good initial microbenchmark would be the time it takes for the block to be forged during the slot the leader is producing a block.
```

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -368,11 +368,7 @@ forkBlockProduction maxBlockBodySize IS{..} BlockProduction{..} =
           -- Check whether we adopted our block
           curTip <- atomically $ ChainDB.getTipPoint chainDB
           if curTip == blockPoint newBlock then do
-            -- We measure the time directly.
-            -- We could also measure it when collecting the trace events, but this
-            -- is much more precise.
-            time <- getMonotonicTime
-            trace $ TraceAdoptedBlock currentSlot newBlock txs time
+            trace $ TraceAdoptedBlock currentSlot newBlock txs
           else do
             isInvalid <- atomically $
               ($ blockHash newBlock) . forgetFingerprint <$>
@@ -392,6 +388,7 @@ forkBlockProduction maxBlockBodySize IS{..} BlockProduction{..} =
                 -- process.
                 removeTxs mempool (map txId txs)
   where
+    trace :: TraceForgeEvent blk (GenTx blk) -> m ()
     trace = traceWith (forgeTracer tracers)
 
     -- Return the point and block number of the most recent block in the
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
@@ -305,16 +305,12 @@ data TraceEventMempool blk
       -- ^ New, valid transaction were added to the Mempool.
       !MempoolSize
       -- ^ The current size of the Mempool.
-      !Time
-      -- ^ The time at which the transactions were added.
   | TraceMempoolRejectedTxs
       ![(GenTx blk, ApplyTxErr blk)]
       -- ^ New, invalid transaction were rejected and thus not added to the
       -- Mempool.
       !MempoolSize
       -- ^ The current size of the Mempool.
-      !Time
-      -- ^ The time at which the transactions were rejected.
   | TraceMempoolRemoveTxs
       ![GenTx blk]
       -- ^ Previously valid transactions that are no longer valid because of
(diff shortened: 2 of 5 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -368,11 +377,10 @@ forkBlockProduction maxBlockBodySize IS{..} BlockProduction{..} =
...
           curTip <- atomically $ ChainDB.getTipPoint chainDB
           if curTip == blockPoint newBlock then do
-            -- We measure the time directly.
-            -- We could also measure it when collecting the trace events, but this
-            -- is much more precise.
-            time <- getMonotonicTime
-            trace $ TraceAdoptedBlock currentSlot newBlock txs time
+            -- There is no need to measure the time directly, the precision of
+            -- adding the actual time measurement on the contravariant tracer is
+            -- the same.
```
mrBliss: I appreciate the comment, but outside the context of this PR, it seems kind of strange to talk about something that we *used* to do. I think a comment in the PR itself would have made more sense. So I'd remove it.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -328,6 +329,14 @@ forkBlockProduction maxBlockBodySize IS{..} BlockProduction{..} =
             case mIsLeader of
               Nothing    -> return NotLeader
               Just proof -> do
+                -- We need to measure the trace just before we actually run the
+                -- protocol because it is possible that the the __current node
+                -- is not the slot leader__ and will __not lead__ the protocol/block
+                -- production.
+                --
+                -- TODO(KS): We need to add the mempool size for measurement.
+                pure . trace $ TraceForgeAboutToLead currentSlot
```
mrBliss: Hmm, we're inside a `STM` here, so you can't trace here. This `pure` means you ignore the action to trace it :confused:

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -328,6 +329,14 @@ forkBlockProduction maxBlockBodySize IS{..} BlockProduction{..} =
             case mIsLeader of
               Nothing    -> return NotLeader
               Just proof -> do
+                -- We need to measure the trace just before we actually run the
+                -- protocol because it is possible that the the __current node
+                -- is not the slot leader__ and will __not lead__ the protocol/block
+                -- production.
+                --
+                -- TODO(KS): We need to add the mempool size for measurement.
```
mrBliss: No need for this TODO, we have a ticket open for it (#1431) that will shortly be addressed.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
mrBliss: Also, now you're tracing `TraceForgeAboutToLead` twice.

ksaric: Yep, forgot to remove that one :grin:

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -311,7 +311,14 @@ forkBlockProduction maxBlockBodySize IS{..} BlockProduction{..} =
...
           withTxs = withSyncState mempool (TxsForBlockInSlot currentSlot)
 
+      -- We need to measure the trace just before we actually run the
+      -- protocol because it is possible that the the __current node
+      -- is not the slot leader__ and will __not lead__ the protocol/block
+      -- production.
+      -- Currently, we will simply drop the unmatched trace event and take
+      -- the diff between the last event of this type and the forged block event.
       trace $ TraceForgeAboutToLead currentSlot
+
```
mrBliss: In fact, do we need this comment here? This seems like something that belongs in the monitoring/benchmarking code?

ksaric: Removed.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -312,6 +312,7 @@ forkBlockProduction maxBlockBodySize IS{..} BlockProduction{..} =
           withTxs = withSyncState mempool (TxsForBlockInSlot currentSlot)
 
       trace $ TraceForgeAboutToLead currentSlot
+
```
mrBliss: Nitpicking: can you remove this newly added blank line?

Issue #1420, opened 2020-01-08
Performance analysis and performance improvments of the networking code


coot:
@karknu performance analysis revealed:
```
COST CENTRE                           MODULE                                   SRC                                                               %time %alloc

mkCodecCborLazyBS                     Network.TypedProtocol.Codec.Cbor         src/Network/TypedProtocol/Codec/Cbor.hs:(110,1)-(126,53)            9.8   22.0

throwSocketErrorIfMinus1RetryMayBlock Network.Socket.Internal                  Network/Socket/Internal.hs:(177,1)-(178,53)                         7.4    0.3

mux                                   Network.Mux.Egress                       src/Network/Mux/Egress.hs:(112,1)-(117,75)                          7.0    3.6

driverSimple                          Network.TypedProtocol.Driver.Simple      src/Network/TypedProtocol/Driver/Simple.hs:(118,1)-(142,24)         6.0    2.6

fdPreadBuf                            System.Posix.IO.ByteString               src/System/Posix/IO/ByteString.hsc:(374,1)-(383,41)                 4.2    0.3

try                                   Ouroboros.Storage.Util.ErrorHandling     src/Ouroboros/Storage/Util/ErrorHandling.hs:73:1-72                 3.7    3.6

demux                                 Network.Mux.Ingress                      src/Network/Mux/Ingress.hs:(79,1)-(92,39)                           3.7    9.0

socketAsMuxBearer                     Network.Mux.Bearer.Socket                src/Network/Mux/Bearer/Socket.hs:(51,1)-(127,86)                    2.9    1.4
```

karknu:
```
        Thu Jan  9 12:19 2020 Time and Allocation Profiling Report  (Final)

           cardano-node +RTS -T -I0 -A128M -N2 -I0 -n64m -p -RTS --config configuration/mainnet-proxy-follower-node4.yaml --genesis-file configuration/mainnet-genesis.json --genesis-hash 5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb --port 3000 --topology configuration/ipv6-topology.json --database-path db --socket-dir socket/

        total time  =      259.37 secs   (518740 ticks @ 1000 us, 2 processors)
        total alloc = 114,349,304,880 bytes  (excludes profiling overheads)

COST CENTRE                           MODULE                                      SRC                                                                  %time %alloc

[...]

PR #1434, opened 2020-01-09
Perfermance improvements for sending.

karknu:
Implements performance improvements discussed in #1420 .

### DIFF
--- a/network-mux/src/Network/Mux.hs
+++ b/network-mux/src/Network/Mux.hs
@@ -28,6 +28,7 @@ import           Control.Monad.Class.MonadThrow
 import           Control.Tracer
 import           Data.Array
 import qualified Data.ByteString.Lazy as BL
+import           Data.Int (Int64)
 import           Data.List (lookup)
 import           Data.Maybe (fromMaybe)
 import           GHC.Stack
@@ -214,10 +215,14 @@ muxChannel
     -> StrictTVar m Int
     -> m (Channel m)
 muxChannel tracer pmss mid md cnt = do
-    w <- newEmptyTMVarM
+    w <- newTVarM BL.empty
     return $ Channel { send = send (Wanton w)
                      , recv}
   where
+    -- Limit for the message buffer between send and mux thread.
+    perMiniProtocolBufferSize :: Int64
+    perMiniProtocolBufferSize = 0x3ffff
+
     send :: Wanton m
          -> BL.ByteString
          -> m ()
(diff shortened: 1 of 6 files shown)


network-mux/src/Network/Mux/Bearer/Socket.hs
```diff
@@ -112,17 +111,4 @@ socketAsMuxBearer tracer sd = do
...
-          {- getSocketOption for MaxSegment is not supported for AF_UNIX sockets -}
-          addr <- Socket.getSocketName sd
-          case addr of
-               Socket.SockAddrUnix _ -> return 0xffff
-               _                     -> do
-                    mss <- Socket.getSocketOption sd Socket.MaxSegment
-                    -- 1260 = IPv6 min MTU minus TCP header, 8 = mux header size
-                    return $ fromIntegral $ max (1260 - 8) (min 0xffff (15 * mss - 8))
-#endif
+      sduSize = return 12288
```
coot: We don't need to wrap it in `IO` unless we plan to have dynamic sdu size in the future.

karknu: True, will change it.

karknu:
> You don't use vectored IO here yet?
Socket send use vectored IO internally, every chunk will end up in a `struct iovec` in the call to `writev`.

network-mux/src/Network/Mux.hs
```diff
@@ -232,10 +237,20 @@ muxChannel tracer pmss mid md cnt = do
...
                 callStack
-        atomically $ modifyTVar cnt (+ 1)
-        atomically $ putTMVar w encoding
+
         traceWith tracer $ MuxTraceChannelSendStart mid encoding
-        atomically $ writeTBQueue (tsrQueue pmss) (TLSRDemand mid md want)
+
+        atomically $ do
+            buf <- readTVar w
+            if BL.length buf < perMiniProtocolBufferSize
```
coot: `perMiniProtocolBufferSize` seems to be to do rate limiting of writes for a mini-protocol.  In the future we might experiment with different values per mini-protocol: chian-sync just sends headers but block-fetch sends whole blocks. I think ideally we would have it proportional with respect to header / block size. What do you think @karknu?

karknu: Previously there was a limit of one message per miniprotocol. When I moved from a TMVar to a TVar I wanted to still have some kind of limit to constrain the memory consumption. A miniprotocol can still have at least one message in flight (we can still cope with a 1Mbyte block message).
I don't see a point in making this any more complicated than it is. The buffer only has to be larger than a maximum mux segment so that there is a full message ready to go once `mux` thread is ready to service the miniprotocol again.

PR #1436, opened 2020-01-10
[CAD-402] Add mempool info.

ksaric:
https://jira.iohk.io/browse/CAD-402

Adding the mempool information into the trace forge events, as agreed.
Uses https://github.com/input-output-hk/ouroboros-network/pull/1433

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -300,6 +300,7 @@ initBlockFetchConsensusInterface cfg chainDB getCandidates blockFetchSize
                            -> Ordering
     compareCandidateChains = compareAnchoredCandidates cfg
 
+
 forkBlockProduction
     :: forall m peer blk.
        (IOLike m, RunNode blk)
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -300,7 +300,6 @@ initBlockFetchConsensusInterface cfg chainDB getCandidates blockFetchSize
                            -> Ordering
     compareCandidateChains = compareAnchoredCandidates cfg
 
-
 forkBlockProduction
     :: forall m peer blk.
        (IOLike m, RunNode blk)


ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
```diff
@@ -30,7 +30,7 @@ import           Ouroboros.Consensus.ChainSyncServer (TraceChainSyncServerEvent)
 import           Ouroboros.Consensus.Ledger.Abstract (AnachronyFailure,
                      ProtocolLedgerView)
 import           Ouroboros.Consensus.Mempool.API (ApplyTxErr, GenTx, GenTxId,
-                     TraceEventMempool)
+                     TraceEventMempool, MempoolSize)
```
intricate: Nitpick :smiley: 

```suggestion
                     MempoolSize, TraceEventMempool)
```

mrBliss:
@ksaric Can you rebase on master? This code has very recently been updated.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -345,6 +344,7 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
       case mIsLeader of
         Nothing    -> return ()
         Just proof -> do
+          trace $ TraceForgeAboutToLead currentSlot
```
intricate: Good catch.

mrBliss: We rely on this event in the protocol tests. This is why CI is failing for this PR. 

I do admit that its name is a bit confusing. Maybe we can rename the existing trace point to `TraceForgeStartOfSlot` and use `TraceForgeAboutToLead` for the new location. /cc: @nfrisby

nfrisby: The event needs to happen before the thread reads the ChainDB. Specifically, in the `ThreadNet` tests, the tracer itself (!!!) actually forges and adds an EBB before returning, which must be able to affect `(extLedger, (prevPoint, prevNo))`.

(... I absolutely should have added a comment here saying as much. :()

mrBliss: Right. Do you agree that we should change its name to be slightly less misleading? The tests should of course also be updated.

nfrisby: Sure. My thoughts on the name:

  * If we consider only its role in the tests, then something extremely concrete like `TraceForgeQueringChainDBTipBeforeForging` would be very telling.
  * If we prefer a more general constructor name, then `TraceForgeSlotOnset` (with comments at its definition and use sites about needing to precede the ChainDB query) might be more reasonable.

intricate: Oops, I guess I got confused by the name too.

ksaric:
I will add that when I actually run the commit that fails here, it passes locally using `stack test --fast -j12 --nix`.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: FYI, @edsko renamed this tracer in PR #1459.

mrBliss:
@ksaric Can you rebase again? :slightly_smiling_face:  As mentioned above, Edsko has recently changed this code.

ksaric:
Rebased. It's a beauty :expressionless:

mrBliss:
> Rebased. It's a beauty expressionless

Ah, an empty PR, I didn't realise everything you wanted to be traced is now already traced :slightly_smiling_face:. Closing.

(If the empty PR was not your intention, reopen it)

PR #1444, opened 2020-01-10
Move EpochNo and EpochInfo to cardano-slotting.

nc6:
We also chance the DB cache to use an IntPSQ rather than a HashPSQ.
The `EpochNo` type in cardano-slotting does not have a Hashable
instance, and we don't need one in this case.

Note that since we are using the EpochNo as an integer only as a key for
lookup, the Word64-Int64 conversion is fine on a 64-bit platform. On a
32 bit platform, we maybe have to worry in a few million years, unless
we use much smaller epochs.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/Common.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/Common.hs
@@ -20,34 +20,20 @@ module Ouroboros.Storage.Common (
   , BinaryInfo (..)
   ) where
 
-import           Cardano.Binary (ToCBOR (..))
 import           Codec.CBOR.Decoding (Decoder)
 import qualified Codec.CBOR.Decoding as Dec
 import           Codec.CBOR.Encoding (Encoding)
 import qualified Codec.CBOR.Encoding as Enc
 import           Codec.Serialise (Serialise (..))
-import           Data.Hashable (Hashable)
 import           Data.Word
 import           GHC.Generics
 
 import           Cardano.Prelude (NoUnexpectedThunks)
+import           Cardano.Slotting.Slot (EpochNo (..), EpochSize (..))
 
 import           Ouroboros.Network.Block (Point, genesisPoint)
 
 import           Ouroboros.Consensus.Util.Condense
-
-{-------------------------------------------------------------------------------
-  Epochs
--------------------------------------------------------------------------------}
-
--- | An epoch, i.e. the number of the epoch.
-newtype EpochNo = EpochNo { unEpochNo :: Word64 }
-  deriving stock (Eq, Ord, Show, Generic)
-  deriving newtype (Enum, Num, Serialise, ToCBOR, NoUnexpectedThunks, Hashable)
-
-newtype EpochSize = EpochSize { unEpochSize :: Word64 }
-  deriving stock (Eq, Ord, Show, Generic)
-  deriving newtype (Enum, Num, Real, Integral, NoUnexpectedThunks)
 
 {-------------------------------------------------------------------------------
   Indexing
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/Ouroboros/Storage/Common.hs
```diff
@@ -20,35 +20,21 @@ module Ouroboros.Storage.Common (
...
 import qualified Codec.CBOR.Decoding as Dec
 import           Codec.CBOR.Encoding (Encoding)
 import qualified Codec.CBOR.Encoding as Enc
 import           Codec.Serialise (Serialise (..))
-import           Data.Hashable (Hashable)
 import           Data.Word
 import           GHC.Generics
 
 import           Cardano.Prelude (NoUnexpectedThunks)
+import           Cardano.Slotting.Slot (EpochNo(..), EpochSize(..))
```
mrBliss:
```suggestion
import           Cardano.Slotting.Slot (EpochNo (..), EpochSize (..))
```
Nitpicking: next time I touch this file, `stylish-haskell` will add those spaces.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -161,7 +162,7 @@ data Cached hash = Cached
     -- from the current epoch, so it is worth optimising this case.
     -- Additionally, by appending to the current epoch through the cache, we
     -- are sure the current epoch info is never stale.
-  , pastEpochsInfo   :: !(HashPSQ EpochNo LastUsed (PastEpochInfo hash))
+  , pastEpochsInfo   :: !(IntPSQ LastUsed (PastEpochInfo hash))
```
intricate: Should we necessarily use an `IntPSQ`? i.e. is the performance of an `OrdPSQ` a problem?

I'm also a bit uncomfortable with having to convert from an `EpochNo` (`newtype` wrapper around a `Word64`) to an `Int` (only guaranteed to be at least 30 bits). This conversion seems like it could be problematic.

mrBliss: Luke has got a point. Although it won't be a problem in practice, as we'll probably never reach such an epoch, it at least deserves a comment saying why we think this conversion is fine (so the auditors won't complain).

nc6: My response to this was in the comments - our epochs are about a week long. Unless they get orders of magnitude shorter, this suggests than on 32-bit platforms we may have problems in a few million years. Given it's a cache, I'd rather that and have the extra performance?

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -161,7 +162,7 @@ data Cached hash = Cached
     -- from the current epoch, so it is worth optimising this case.
     -- Additionally, by appending to the current epoch through the cache, we
     -- are sure the current epoch info is never stale.
```
nc6:
```suggestion
    -- are sure the current epoch info is never stale.
    --
    -- We use an 'IntPSQ' here, where the keys are in fact epoch numbers. Since
    -- epoch numbers are internally represented by a 'Word64', one might be worried
    -- about a potential overflow here. While possible, it's not worth worrying about:
    -- - Whilst guaranteed to be only at least 30 bits, in practice, 64-bit GHC has 64-bit
    --   integers, so the conversion is bijective.
    -- - An epoch currently lasts around a week. Systems using a smaller representation
    --   might need to worry in a million years or so.
    -- - In the event of running for a million years, we're unlikely to have a problem anyway,
    --   since we only really cache _recent_ epochs. So the fact that they clash with the
    --   epochs from a million years ago isn't likely to be an issue.
```

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
mrBliss: Agreed. Comment looks good. If you can squash the last two commits, we can merge.

nc6:
@mrBliss @intricate Detailed comment added explaining why we shouldn't care about the int conversion

nc6:
bors retry

nc6:
bors retry

PR #1449, opened 2020-01-13
Milder exception handlers in ImmDB and VolDB

### LINKED ISSUE #1448: Reduce severity of exception handlers in ImmDB and VolDB
While implementing Issue #1202 via PR #1419, I noticed the tests failing often with `CloseDBError`s. Investigation eventually identified these two case alternatives.

https://github.com/input-output-hk/ouroboros-network/blob/e0eba52e8e7c81d3f002c8d36a3e5f6ed30bb216/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/State.hs#L212

https://github.com/input-output-hk/ouroboros-network/blob/e0eba52e8e7c81d3f002c8d36a3e5f6ed30bb216/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/State.hs#L267

Both of those `shutDown` functions close the database. @mrBliss also added this case alternative to my list.

https://github.com/input-output-hk/ouroboros-network/blob/779b0dd353ab5891e2eec997046132c7e91ff8f0/ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs#L684-L686

Closing the DB in response to any exception seems to be too severe of a handler.

This Issue is to confirm that no exception (none whatsoever?) should cause the DBs to close and to update the handlers accordingly.

-----

In the cases I observed, the exception was an `AsyncCancelled` incurred by peer B because it was scheduled (via PR 1419) to restart. The exception ultimately arrived at peer A's corresponding ChainSyncServer thread, which happened to be running ImmDB code (via the thread's `Reader`/`Iterator`), so the arrival of the exception closed peer A's ImmDB. Note that peer A was not scheduled for a restart! Even if it were, closing the ImmDB in this way would likely cascade as other local threads dependent upon the ImmDB then raised `Imm.CloseDBError` exceptions (node shutdown must release threads/resources in a careful order).

I have repros for these handlers, but I expect they're fragile (e.g. based on timing coincidences), and I'm unsure if I can divorce them from the preceding PR 1419 changes. I'll see what I can do. Notably, those repros stopped failing when I temporarily patched those case alternatives to handle `AsyncCancelled` using the same code as the juxtaposed `ExitCaseAbort` alternatives.

(Engineering note: The enclosing `generalBracket` itself re-raises the inciting exception; its `close` argument is not responsible for that.)

nfrisby:
Fixes #1448.

I've opened as a Draft PR, because I haven't added any test cases. @mrBliss would you advise on how to do that? I think the repros from the discovery on PR 1419 are too fragile, and I'm not even sure I can disentangle them from that PR.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/State.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/State.hs
@@ -206,10 +206,9 @@ modifyOpenState ImmutableDBEnv { _dbHasFS = hasFS :: HasFS m h, .. } action = do
           -> ExitCase (Either ImmutableDBError (r, OpenState m hash h))
           -> m ()
     close !st ec = case ec of
-      -- Restore the original state in case of an abort
+      -- If we were interrupted, restore the original state.
       ExitCaseAbort                               -> putMVar _dbInternalState st
-      -- In case of an exception, close the DB for safety.
-      ExitCaseException _ex                       -> shutDown st
+      ExitCaseException _ex                       -> putMVar _dbInternalState st
       -- In case of success, update to the newest state
       ExitCaseSuccess (Right (_, ost))            -> putMVar _dbInternalState (DbOpen ost)
       -- In case of an unexpected error (not an exception), close the DB for safety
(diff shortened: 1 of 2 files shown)


nfrisby:
I rebased on to master --- totally automatic.

PR #1450, opened 2020-01-13
Always produce a genesis EBB for Byron

### LINKED ISSUE #1447: Create a genesis EBB for Byron
In `cardano-sl`, an EBB is produced at the start of each epoch.

In our rewrite (but before #1353), we don't produce EBBs at all since EBBs will be a thing of the past by the time it is deployed (finally!). There is however, *one exception*: a slot leader with an empty chain produces a genesis EBB as its first block instead of a regular block containing transactions.

Since #1353, we no longer produce this genesis EBB. In the protocol tests, we automatically produce an EBB at the start of each epoch, just like in the old codebase.

The problem is that only `BoundaryHeader`s (EBBs) can have the empty chain (genesis) as their predecessor. Compare:
https://github.com/input-output-hk/cardano-ledger/blob/c26bfbd0e6726c1c36a17bf7bde815d2e84bccf5/cardano-ledger/src/Cardano/Chain/Block/Header.hs#L449
with:
https://github.com/input-output-hk/cardano-ledger/blob/c26bfbd0e6726c1c36a17bf7bde815d2e84bccf5/cardano-ledger/src/Cardano/Chain/Block/Header.hs#L138

Now the benchmarking team are performing TPS tests, in which they are producing Byron blocks, but not EBBs. The first block they produce doesn't fit onto genesis, as its previous hash is `BlockHash (ByronHash {unByronHash = AbstractHash c89ac6b8c6e68ef58ad6a4ed0f7ba29ab33f16354d4437a28a6bf1d9a3f90508})` instead of `GenesisHash`.

This is due to the following line:
https://github.com/input-output-hk/ouroboros-network/blob/e0eba52e8e7c81d3f002c8d36a3e5f6ed30bb216/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Forge.hs#L194

Proposed solution: add a hook to the node (`RunNode`) to run an action on the ChainDB on startup, before block production begins. For Byron, we can then use this hook to create a genesis EBB in case our chain is empty. Throw an `error` on the problematic line above instead of using the wrong prev hash.

mrBliss:
Closes #1447.

Test it by running the RealPBFT tests both with and without EBBs for each
epoch.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Forge.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Forge.hs
@@ -141,8 +141,7 @@ forgeRegularBlock cfg curSlot curNo prevHash txs isLeader = do
   where
     -- TODO: Might be sufficient to add 'ConfigContainsGenesis' constraint.
     ByronConfig
-      { pbftGenesisHash
-      , pbftEpochSlots
+      { pbftEpochSlots
       , pbftProtocolVersion
       , pbftSoftwareVersion
       , pbftProtocolMagic
@@ -191,7 +190,8 @@ forgeRegularBlock cfg curSlot curNo prevHash txs isLeader = do
 
     prevHeaderHash :: CC.Block.HeaderHash
     prevHeaderHash = case prevHash of
-      GenesisHash             -> CC.Block.genesisHeaderHash pbftGenesisHash
+      GenesisHash             -> error
+        "the first block on the Byron chain must be an EBB"
       BlockHash (ByronHash h) -> h
 
     epochAndSlotCount :: CC.Slot.EpochAndSlotCount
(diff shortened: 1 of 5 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Forge.hs
```diff
@@ -191,7 +190,8 @@ forgeRegularBlock cfg curSlot curNo prevHash txs isLeader = do
 
     prevHeaderHash :: CC.Block.HeaderHash
     prevHeaderHash = case prevHash of
-      GenesisHash             -> CC.Block.genesisHeaderHash pbftGenesisHash
+      GenesisHash             -> error
+        "forging a regular block instead of an EBB on top of an empty chain"
```
edsko: Slightly confusing error message. Maybe "the first block on the Byron ledger _must_ be an EBB".

nfrisby:
bors retry

PR #1453, opened 2020-01-13
Add uninterruptibleMask wrapper to a takeMVar in ImmDB and VolDB 

### LINKED ISSUE #1452: Protect a takeMVar inside `mask` in ImmDB and VolDB
This Issue blocks Issue 1202 by blocking PR 1419.

While working on PR 1419, which introduces node restarts during the `ThreadNet` test, I noticed that the ImmDB and the VolDB both call `takeMVar` inside of a `generalBracket`. Since `takeMVar` is interruptible, this means the DBs' invariants could be violated by an asynchronous exception arriving while that `takeMVar` is blocked, regardless of the surrounding `mask`s.

Specifically, during PR 1419 development, I investigated a fatal exception during the protocol tests. It turns out that the `takeMVar` in `ImmDB.State.modifyOpenState` was (re)raising an `AsyncCancelled` that a ChainSyncServer thread received _while opening_ a new `ImmDB.Iterator` via its `ChainDB.Reader`. The `mask` violation lead to that iterator never being closed, which the test infrastructure caught as a "the ImmDB is leaking file handles" error.

This Issue is to essentially either carefully add `uninterruptibleMask_` or re-architect to avoid the need.

nfrisby:
Fixes #1452.

This PR prevents wraps the `takeMVar` calls mentioned in the Issue in the sledgehammer `uninterruptibleMask_`. I've scanned for other uses of the `STM` variables to ensure that they're never empty for very long (`uninterruptibleMask_` should not be used on calls that may block for "long" durations) and also added another couple `mask` calls to ensure that every `take` is always paired with a subsequent `put`.

I'm opening this as a Draft PR (edit: we're proceeding, see Issue 1464) because:

  * I want another developer to confirm that `uninterruptibleMask_` is the desired solution here. Maybe we can re-architect instead? A note: the immediately surrounding `bracket` starts with `takeMVar`, so the `uninterruptibleMask_` isn't really necessary for that one. But in the repros (currently only on my local PR 1419, sadly) there are more outer layers of `mask` (actually `bracket`, I think) that do other things before reaching this `takeMVar`.
  * The timing involved on this problem seems delicate enough that it's not obvious to me how to add a repro to the test suite. (Beyond being delicate, I'm unsure I can disentangle my current repros from PR 1419.) Maybe there's a promising way to add this to `test-storage` (with which I am not yet familiar).

@mrBliss @edsko, can you advise?

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl.hs
@@ -107,7 +107,7 @@ import           Data.Functor (($>), (<&>))
 
 import           GHC.Stack (HasCallStack)
 
-import           Control.Monad.Class.MonadThrow (bracket, finally)
+import           Control.Monad.Class.MonadThrow (bracket, bracketOnError, finally)
 
 import           Ouroboros.Consensus.Block (IsEBB (..))
 import           Ouroboros.Consensus.Util (SomePair (..))
@@ -289,8 +289,8 @@ closeDBImpl ImmutableDBEnv {..} = releaseAll _dbRegistry `finally` do
     case internalState of
       -- Already closed
       DbClosed  _ -> do
-        traceWith _dbTracer $ DBAlreadyClosed
         putMVar _dbInternalState internalState
+        traceWith _dbTracer $ DBAlreadyClosed
       DbOpen OpenState {..} -> do
         let !closedState = closedStateFromInternalState internalState
         -- Close the database before doing the file-system operations so that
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl.hs
```diff
@@ -284,13 +284,13 @@ closeDBImpl
   :: forall m hash. (HasCallStack, IOLike m)
   => ImmutableDBEnv m hash
   -> m ()
-closeDBImpl ImmutableDBEnv {..} = releaseAll _dbRegistry `finally` do
+closeDBImpl ImmutableDBEnv {..} = (releaseAll _dbRegistry `finally`) $ mask_ $ do
```
edsko: I think this `mask_` is not needed; the finalizer from `finally` will be run with exceptions masked.

nfrisby: Fixed, thanks.

nfrisby:
I rebased to remove my overzealous `mask_`, add the TODOs, and onto `origin/master`.

nfrisby:
I opened Issue #1464 to revisit this route.

PR #1455, opened 2020-01-13
Encapsulation of connection management

avieth:
There are simple socket-backed client and server definitions in
Client/Inet.hs and Server/Inet.hs. These simply attempt to include
established socket connections into a Connections term. If they are
rejected, they close them; if they are accepted, then the Connections
term is now responsible for closing them.

An example of a Connections term is in Concurrent.hs. It accepts at most
one incoming and outgoing connection to a given address, and runs a
callback on each of them once they have been accepted.

The point is to decouple the mechanisms for

- Establishing a bearer between the local node and some peer, which must
  be done both passively (server accept loop) and actively (P2P
  selection governor, DNS subscription worker, etc.)
- Actually doing something with those connections, managing threads
  which act on them, querying and poking and prodding the states of
  these high level connections (hot, warm, cold)

The idea is to come up with a Connections term which controls version
negotiation and mux start/stop, and then a corresponding
PeerConnection type which can be used by the outside world (P2P
governor) to observe a peer's current state and to suggest changes to
it.

## Update

This now also includes a refactor of node-to-node and node-to-client servers. They now use `concurrent` `Connections` to deal with sockets, as opposed to having servers and clients work independently and be bound by a `ConnectionTable` in `NetworkMutableState`. The approach is to create a `Connections` term which knows what to do based on a type-indexed request type (indexed by `Provenance = Local | Remote`). See its use in `Ouroboros.Network.Diffusion`: there is one `Connections` term for local (node-to-client) and one for peer-to-peer (node-to-node). Once they are created, `acceptLoopOn` is used against them to create accept loops. Next steps are to refactor the IP and DNS subscription workers to connect by way of the `Connections` term.

To hook up the p2p governor, we can get an `establishPeerConnection` function by using a `Connections` term. The `peerconn` handle that it returns is yet to be defined (currently we use `()` as a handle).


ouroboros-network/src/Ouroboros/Network/Connections/Client/Inet.hs
```diff
@@ -0,0 +1,37 @@
+{-# LANGUAGE DataKinds #-}
```
karknu: Inet is a rather poor name given that we're supporting UNIX domain sockets too right now, and probably windows named pipes in the future.

ouroboros-network/src/Ouroboros/Network/Connections/Client/Inet.hs
```diff
@@ -0,0 +1,37 @@
...
+inetClient family bindaddr sockaddr k = k sockaddr openSocket closeSocket
+
+  where
+
+  -- The Connections term is expected to take care of exception handling to
+  -- ensure closeSocket is always called. But we need to do bracketing even
+  -- here within openSocket, in case bind or connect fails
+  openSocket :: IO Socket
+  openSocket = bracket createSocket closeSocket $ \sock -> do
+    Socket.bind sock bindaddr
```
karknu: Please look at for example `connectToNode` to see which options we should set before we bind to a socket, i.e `ReuseAddr`,  `ReusePort` and `IPv6Only`.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
```diff
@@ -0,0 +1,143 @@
...
+
+data Connection
+  = Remote (Async ())
+  | Local (Async ())
+  -- | First is remote, second is local
+  | Both (Async ()) (Async ())
+
+-- | Simple concurrent connection manager as a proof of concept.
+-- It maintains at most 2 connections for each SockAddr: one outgoing and
+-- one incoming. Rejects new ones only if there's already one known to it.
```
karknu: This is not what is needed. We need *one* connection, and we can have at most *one* connection since the local and remote port will be fixed.

avieth:
@karknu Sorry I should have been more clear. This is a proof of concept, it's not supposed to be a feature-complete drop-in replacement for what we have now. The `Concurrent.hs` module in particular: it's a demonstration of how it would work in general, and can easily be modified to allow for only one connection.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
```diff
@@ -6,138 +6,215 @@
...
-  | Local (Async ())
-  -- | First is remote, second is local
-  | Both (Async ()) (Async ())
+-- | An action to run for each connection, and a handle giving an interface
+-- to that action, probably using STM for synchronization.
+-- The connection for which this handler is spawned will be closed whenever
+-- the action returns, exceptionally or not.
+data Handler handle = Handler
+  { handle :: !handle
+  , action :: !(IO ())
```
karknu: Does this have to be in `IO`? Mux can run in `io-sim` and with Snockets on the way I think we should be able to test this in sim.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
```diff
@@ -6,138 +6,215 @@
...
+  -> (Connections connectionId resource (Reject reason) (Accept handle) IO -> IO t)
   -> IO t
-concurrent withSocket k = do
-  -- Cannot use an IORef for state because spawned threads will modify this
-  -- when they finish.
-  state :: MVar (Map SockAddr Connection) <- newMVar Map.empty
-  let connections = Connections { include = includeOne state }
-  k connections `finally` killThreads state
+concurrent withResource k = do
+  stateVar :: MVar (State connectionId handle) <- newMVar $ State
```
karknu: Why is this an MVar and not a StrictTVar?

karknu: How will the connection tables be shared between the governor and the server? We need to run one `concurrent` for IPv4 and one for IPv6 for the server (same port number but different socket to listen to), and at least one `concurrent` for the client side.
Technically we can have a different table for IPv4 and IPv6 since the connections can never be shared, but the client still must be able to access both tables since it needs to be able to talk to both types of peers.

ouroboros-network/demo-connections/Main.hs
```diff
@@ -0,0 +1,143 @@
+{-# LANGUAGE ExistentialQuantification #-}
```
avieth: This file is just a demonstration of how to use the `Connections` interface. Has nothing to do with running an ouroboros application.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
avieth: No, I will port it all over to io-sim stuff. I'm just a bit annoyed that there is no `MVar` equivalent in there, so I have to simulate it using `TMVar`s.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
avieth: It's an `MVar` because that's IMO the most appropriate, simplest thing to use (i.e. anyone else writing this who doesn't know about io-sim would have chosen it). I'll have to change it to be a `TMVar` in order to make this thing fit with io-sim. Is there a `StrictTMVar` too?

avieth: > How will the connection tables be shared between the governor and the server?

This `concurrent` function takes care of it. So long as the governor and the server use the same `Connections` term, it's all good. 

> We need to run one concurrent for IPv4 and one for IPv6 for the server (same port number but different socket to listen to), and at least one concurrent for the client side.

In fact, we'll have just _one_ for all node-to-node connections. We can run two different accept loops using the same manager, one for IPv6 and one for IPv4 (could even use UNIX sockets as well but they have identifier problems).

For client side, you can give any bind and connect address and use that same `Connections` term to try and bring up a connection.

> Technically we can have a different table for IPv4 and IPv6 since the connections can never be shared, but the client still must be able to access both tables since it needs to be able to talk to both types of peers.

The table will be the same.

ouroboros-network/src/Ouroboros/Network/Subscription/Worker.hs
```diff
@@ -155,7 +155,6 @@ sockAddrFamily
 sockAddrFamily (Socket.SockAddrInet  _ _    ) = Socket.AF_INET
 sockAddrFamily (Socket.SockAddrInet6 _ _ _ _) = Socket.AF_INET6
 sockAddrFamily (Socket.SockAddrUnix _       ) = Socket.AF_UNIX
-sockAddrFamily _                              = Socket.AF_UNSPEC
```
avieth: I've updated to `network-3.0.0.0`. Now `SockAddr` has only these 3 variants.

ouroboros-network/src/Ouroboros/Network/Connections/Client/Inet.hs
avieth: Changed to `Socket/Client.hs` and `Socket/Server.hs`

ouroboros-network/src/Ouroboros/Network/Connections/Client/Inet.hs
avieth: Done.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
avieth: Done

ouroboros-network/src/Ouroboros/Network/Connections/Socket/Types.hs
```diff
@@ -0,0 +1,141 @@
...
+-- `SockAddrUnix ""` term, maybe using its file descriptor.
+--
+data ConnectionId where
+  ConnectionIdIPv6 :: !(SockAddr IPv6) -> !(SockAddr IPv6) -> ConnectionId
+  ConnectionIdIPv4 :: !(SockAddr IPv4) -> !(SockAddr IPv4) -> ConnectionId
+  ConnectionIdUnix :: !(SockAddr Unix) -> !(SockAddr Unix) -> ConnectionId
+
+deriving instance Show ConnectionId
+
+instance Eq ConnectionId where
```
avieth: This and the `Ord` instance are actually wrong I believe. The order of the tuple should not matter. A connection from `hostA:portA` to `hostB:portB` is also a connection from `hostB:portB` to `hostA:portA`. Could use the `Ord SockAddr => Ord (Set SockAddr)` instance to get this:

```
Set.fromList [forgetSockType a, forgetSockType b] == Set.fromList [forgetSockType c, fogetSockType d]
Set.fromList [forgetSockType a, forgetSockType b] `compare` Set.fromList [forgetSockType c, fogetSockType d]
```

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
```diff
@@ -0,0 +1,257 @@
...
+    -> connectionId
+    -> Resource provenance IO resource
+    -> request provenance
+    -> IO (Types.Decision provenance (Reject reject) (Accept handle))
+  includeOne stateVar connId resource request = mask $ \restore -> modifyMVar stateVar $ \state ->
+    case Map.lookup connId (connectionMap state) of
+      Nothing   -> case resource of
+        Existing res closeResource -> do
+          -- User must ensure that the handler does not take very long to.
+          -- set up. It should just be creating shared state.
```
coot: Is the first sentence a stale comment?

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
coot: I think @karknu question makes sense: you don't need to await when the shared stated is created, so between `TVar` and `TMVar`, `TVar` makes more sense. 

Some bangs can be removed when using `MonadSTM.Strict`.

ouroboros-network/src/Ouroboros/Network/Subscription/Worker.hs
coot: CI might fail (at least it failed for me when I needed to upgrade to `3.1`) in another PR. I will merge the fix soon.

coot:
To get around CI issue, rebase on top of #1455 (it's about to land in master, or already there by the time you're around :)).

ouroboros-network/src/Ouroboros/Network/Connections/Socket/Types.hs
karknu: I think that the order of the tuples should have meaning, or rather the socket fields in the Connection should have meaning. One field should correspond to the local address and one field should correspond to the remote address so that one doesn't have to guess when interpreting logs in the future.

avieth: Right. Maybe the better solution is to always choose local as the first component

avieth:
> To get around CI issue, rebase on top of #1455 (it's about to land in master, or already there by the time you're around :)).

I think you linked the wrong PR. This is 1455 :)

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
avieth: Yes indeed. It's relevant to the `New` case not the `Existing`. Will update.

avieth: Oh actually on further consideration, it _is_ relevant, just confusing and not very helpful.

coot:
>  I think you linked the wrong PR. This is 1455 :)
:D,  the changes are already in master :)

ouroboros-network/src/Ouroboros/Network/Connections/Types.hs
```diff
@@ -0,0 +1,135 @@
...
+-- TODO may need a way to check / be notified when there is an incoming
+-- connection at a given identifier.
+-- Could just put that on the concurrent connections: an `STM (Map id handle)`.
+-- When you get the handle back, the underlying connection and its handler
+-- thread may of course have died, but that's always a possibility even if
+-- you `include`. 
+--
+-- TODO may be useful to bring `request` to the end and make this a
+-- contravariant functor.
+data Connections id socket request reject accept m = Connections
```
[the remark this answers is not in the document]
coot: Further along `socket` is renamed to `resource` (e.g. in `Connections.Concurrent`).  I think `resource` suits better as we will use it with sockets as well as named pipes (file handles) on Windows.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
```diff
@@ -0,0 +1,267 @@
...
+  k connections `finally` killThreads stateVar
+
+  where
+
+  -- Exception handling in here should ensure that if the handler is succesfully
+  -- created, then it ends up in the shared state.
+  --
+  -- FIXME this one definition is too complex. Try to factor it into simpler
+  -- pieces.
+  includeOne
```
coot: It's de-facto a smart constructor for `Connections`, maybe refactor it into
```
connections :: MVar (State connectionId handle) -> Connections id resource (Reject reject) (Accept handle) IO
```
 
Another question: do we need `handle` and `resource`, or can we simplify it by just using one (e.g. `handle ~ resource`).

ouroboros-network/src/Ouroboros/Network/Connections/Types.hs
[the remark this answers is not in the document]
avieth: Yeah agreed, should use `resource` rather than `socket` throughout. I'll also make it a newtype.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
avieth: It's supposed to be a handle to that resource. In general it's not a good idea to have them be the same. The handle should be a more restricted, higher-level interface to  the resource. If we just gave back the resource then, for example, if it were a socket or file, the caller could close it, but that shouldn't be allowed.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
```diff
@@ -0,0 +1,267 @@
...
+                    unmask (action handler) `finally` cleanup
+              let conn = Connection
+                    { connectionThread = thread
+                    , connectionHandle = handle handler
+                    }
+                  !state' = insertConnection connId conn state
+              pure (state', Types.Accepted (Accepted (connectionHandle conn)))
+      Just numConn -> case resource of
+        -- Do not call _close; the caller is responsible for that, and knows
+        -- it because we give `Rejected`.
```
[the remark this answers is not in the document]
avieth: When we reject it with `Domain reason` we also don't close it.

In that case, and in this case of a duplicate connection, the resource was created by some other system (e.g. an accept loop) and this connections manager has decided not to take ownership of it. It wouldn't be right to close it because that eliminates the possibility of doing something else with it like, for example, trying to include it in another connections manager.

NB: in this case (context of this github comment) the `resource` is the one created by some other system, but there is also a resource in `numConn :: NumberedConnection handle` which represents a connection between the same peers (identifier `connId`).

avieth:
> This could be done by using SimIO monad and traces.

RE testing: I think it should be feasible to get a nice test suite going using `MVar`s to represent sockets.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
coot: I see, this makes sense; what kind of operations on the resource a handle should expose? Probably just getting it's identifier or the identifier of remote peer aka `getSocketName`, `getPeerName`.

coot:
> RE testing: I think it should be feasible to get a nice test suite going using MVars to represent sockets.

With an `MVar` approach you'll be only able to capture properties about the socket, `IOSim` might be more flexible in the properties you could express, including ones that are not about socket. Whether that's useful in this case requires a bit of a thought.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
avieth: In a patch that I'm going to push hopefully later today, I have set the handle to be

```Haskell
data ConnectionHandle m = ConnectionHandle
  { status :: STM m ConnectionResult }

data ConnectionResult = Running | Finished (Maybe SomeException)

waitForConnection :: MonadSTM m => ConnectionHandle m -> m (Maybe SomeException)
```

I don't plan on exposing `getSocketName`, there's no good reason why a user should need to know it's a socket (we'll see).

Eventually we'll need to extend it to support the required P2P `PeerActions` record. Things like "promote this connection from warm to hot".

ouroboros-network/demo-connections/Main.hs
coot: ping ;)

avieth: Sorry, forgot about it. Will do.

avieth:
Note that I removed the DNS resolution "happy eyeball" stuff just for the sake of simplicity while figuring out what's going on with the subscription system. Looking at the `Test/Subscription` module I see that much of the testing is related to this very feature. I'll put it back in if we think it's worthwhile. But then, we're not making connections at anywhere near the rate of, say, a web browser, so I wonder if it's actually worth the complexity.

coot:
Whey I run the demo
```
cabal run demo-connections -- 127.0.0.1 6000 2
```
I see these errors:
```
client 127.0.0.1:6001 : 127.0.0.1:6000 : unsupported operation : Network.Socket.connect: <socket: 14>: unsupported operation (Cannot assign requested address)

Incoming : ConnectionIdIPv4 127.0.0.1:6000 127.0.0.1:6001
Outgoing : ConnectionIdIPv4 127.0.0.1:6001 127.0.0.1:6000
demo-connections: Network.Socket.connect: <socket: 14>: unsupported operation (Cannot assign requested address)
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -193,86 +282,56 @@ runDataDiffusion tracers
...
+          (contramap (WithIPList initiatorLocalAddresses (ispIps daIpProducers)) dtIpSubscriptionTracer)
+          dtErrorPolicyTracer
+          connIds
+          (ispValency daIpProducers)
+          ipRetryDelay
+          (\connId -> Socket.client connId IpSubscriptionConnection)
+          connections
+
+    -- FIXME probably not ideal that we make a resolver `mkResolverIO` for
+    -- each DNS target. Should make one (not have it take a port) and use it
```
coot: ^ a typo, it should be 'take a part'

ouroboros-network/src/Ouroboros/Network/Socket.hs
```diff
@@ -389,38 +221,215 @@ nullNetworkServerTracers = NetworkServerTracers {
...
+-- | Contains everything needed to run a mux'd versioned ouroboros application
+-- as initiator or responder. See use in `withConnections`. The idea is that
+-- the user will give a custom `request :: Provenance -> Type` type and must
+-- create a `ConnectionData ptcl vNumber provenance` with the same provenance
+-- as the request. In this way, the choice of `request` GADT can determine
+-- whether initiator or responder are even allowed (node-to-client server, for
+-- instance, does not allow initiation).
+data ConnectionData ptcl vNumber provenance where
+  -- | Locally-initiated connection data.
+  ConnectionDataLocal
```
coot: Why `ConnectionDataLoca` does not require `ErrorPolicies`?

ouroboros-network/src/Ouroboros/Network/Socket.hs
```diff
@@ -389,38 +221,215 @@ nullNetworkServerTracers = NetworkServerTracers {
...
+    -> VersionDataCodec vDataT CBOR.Term
+    -> (forall vData . vDataT vData -> vData -> vData -> Accept)
+    -> Versions vNumber vDataT (OuroborosApplication appType ConnectionId ptcl IO BL.ByteString a b)
+    -> ConnectionData ptcl vNumber Remote
+
+-- | Handle giving an interface to a connection.
+data ConnectionHandle m = ConnectionHandle
+  { status :: STM m ConnectionStatus }
+
+data ConnectionStatus = Running | Finished (Maybe SomeException)
```
coot: What about using:
```
data ConnectionStatus = Running | Finished | Errored (!SomeException)
```
to avoid lazy thunks in `ConnectionHandle`.

ouroboros-network/src/Ouroboros/Network/Socket.hs
```diff
@@ -389,38 +221,215 @@ nullNetworkServerTracers = NetworkServerTracers {
...
+  -> IO (Connection.Decision IO Local reject (ConnectionHandle IO))
+outgoingConnection versionDataCodec tracers versions connId sd =
+    -- Always accept and run initiator mode mux on the socket.
+    pure $ Connection.Accept $ \_connThread -> do
+        statusVar <- atomically (newTVar Running)
+        let connectionHandle = ConnectionHandle
+              { status = readTVar statusVar }
+            action = mask $ \restore -> do
+              restore (runInitiator versionDataCodec tracers versions connId sd) `catch`
+                (\e -> atomically (writeTVar statusVar (Finished (Just e))))
```
coot: `e` might be an un-evaluated thunk, using `Control.Monad.Class.MonadSTM.Strict` (and the suggested `ConnectionStatus`) would fix that.

ouroboros-network/src/Ouroboros/Network/Socket.hs
```diff
@@ -435,67 +444,60 @@ runServerThread
...
+    -> Socket.Socket            -- ^ Established by the remote peer.
+    -> IO (Connection.Decision IO Remote reject (ConnectionHandle IO))
+incomingConnection NetworkServerTracers { nstMuxTracer
+                                        , nstHandshakeTracer }
+                                        -- , nstErrorPolicyTracer }
+                   versionDataCodec
+                   acceptVersion
+                   versions
+                   -- TODO use these for exception handling.
+                   _errorPolicies
```
coot: ^ not used error policies - do you plan to integrate them in the follow up PR?

coot: Instead of using error policies inside `outgoingConnection` and `incomingConnection` it is better to use them at the call sites of these two functions.

ouroboros-network/src/Ouroboros/Network/Connections/Socket/Types.hs
```diff
@@ -0,0 +1,129 @@
...
+-- kind of address it is: IPv4, IPv6, or Unix domain.
+data SockAddr (sockType :: SockType) where
+  SockAddrIPv4 :: !Socket.PortNumber -> !Socket.HostAddress -> SockAddr IPv4
+  SockAddrIPv6 :: !Socket.PortNumber
+               -> !Socket.FlowInfo
+               -> !Socket.HostAddress6
+               -> !Socket.ScopeID
+               -> SockAddr IPv6
+  -- No strictness here because the Network.Socket analogue is not strict.
+  SockAddrUnix :: String -> SockAddr Unix
```
coot: hmm, so why laziness here is a good choice?

ouroboros-network/src/Ouroboros/Network/Connections/Socket/Types.hs
coot: I suggest we use record fields then which one can use with GADTs.  It will make it clear which one is the local and which one is the remote.

ouroboros-network/src/Ouroboros/Network/Connections/Types.hs
```diff
@@ -0,0 +1,178 @@
...
+-- should be re-thrown. This is consistent with bracketing of the acquire and
+-- release fields of the `New` constructor, which also re-throws the exception.
+--
+-- TODO may need a way to check / be notified when there is an incoming
+-- connection at a given identifier.
+-- Could just put that on the concurrent connections: an `STM (Map id handle)`.
+-- When you get the handle back, the underlying connection and its handler
+-- thread may of course have died, but that's always a possibility even if
+-- you `include`. 
+newtype Connections id resource request reject accept m = Connections
```
coot: The type name is misleading, what about changing it to `ConnectionDecision` or `IncludeConnection`.

ouroboros-network/src/Ouroboros/Network/Connections/Types.hs
```diff
@@ -0,0 +1,178 @@
...
+  { include = \ident res req -> fmap (mapDecision frej facc) (include connections ident res req) }
+
+-- Client and Server types take any suitable Connections include callback,
+-- specialized to Outgoing or Incoming, and use it to get a decision, for any
+-- accept or reject type.
+
+-- | A `Client` takes a continuation that is derived from a `Connections`
+-- (see `runClientWith`) and uses it to get a decision parameteric in accept
+-- and reject.
+type Client addr sock m request = forall accept reject .
```
coot: I would rename this `Client` type to `ClientSocket` or something similar. The construction of `Client` instruments how to construct a socket for a client connection - I would also put such a sentence to the haddock, and link add a reference to `Connections.Socket.Client.client`.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -181,231 +169,35 @@ nodeToNodeCodecCBORTerm = CodecCBORTerm {encodeTerm, decodeTerm}
...
-          application)
+-- | `Ouroboros.Network.Socket.withConnections` but with the protocol types
+-- specialized. It also fills in the version data codec
+-- `cborTermVersionDataCodec` and `acceptEq` to determine when to accept a
+-- version.
+withConnections
+  :: forall request t.
+     ErrorPolicies Socket.SockAddr ()
+  -> (forall provenance . request provenance -> SomeVersionedApplication NodeToNodeProtocols NodeToNodeVersion DictVersion provenance)
+  -> (Connections Connections.ConnectionId Socket.Socket request (Connection.Reject RejectConnection) (Connection.Accept (ConnectionHandle IO)) IO -> IO t)
```
coot: Could you break this line so it fits in <100 columns.

ouroboros-network/src/Ouroboros/Network/Socket.hs
```diff
@@ -389,38 +221,215 @@ nullNetworkServerTracers = NetworkServerTracers {
...
+     , Bounded ptcl
+     , Ord vNumber
+     , Enum vNumber
+     , Serialise vNumber
+     , Typeable vNumber
+     , Show vNumber
+     , Show ptcl
+     )
+  => (forall provenance. request provenance -> ConnectionData ptcl vNumber provenance)
+  -> (Connections Connections.ConnectionId Socket.Socket request (Connection.Reject reject) (Connection.Accept (ConnectionHandle IO)) IO -> IO t)
```
coot: Could you break this line so it fits `<= 100` (ideally `<= 80`) columns.

ouroboros-network/src/Ouroboros/Network/Subscription/Worker.hs
```diff
@@ -80,544 +67,91 @@ minConnectionAttemptDelay = 0.010 -- 10ms delay
...
+  -> Word
+  -- ^ Valency: how many concurrent connections to try to keep up. Will be
+  -- clamped to at least 1 and at most the length of the list of targets.
+  -> DiffTime
+  -- ^ Minimum delay between subscriptions. If a subscription attempt lasts
+  -- less than this duration, it will wait for the difference.
+  -> (ConnectionId -> Client ConnectionId socket IO request)
+  -- ^ "How" to carry out a request for a given connection identifier (pair of
+  -- addresses in the same family).
+  -> Connections ConnectionId socket request (Concurrent.Reject reject) (Concurrent.Accept (ConnectionHandle IO)) IO
```
coot: this line seems quite long, could you break it

ouroboros-network/src/Ouroboros/Network/Subscription/Worker.hs
```diff
@@ -80,544 +67,91 @@ minConnectionAttemptDelay = 0.010 -- 10ms delay
...
+-- Includes a built-in delay of `ipRetryDelay` between connections but
+-- FIXME this should not be built-in.
+--
+workerOneTarget
+  :: Tracer IO (SubscriptionTrace Socket.SockAddr)
+  -> Tracer IO (WithAddr Socket.SockAddr ErrorPolicyTrace)
+  -> DiffTime
+  -> TQueue ConnectionId
+  -> (ConnectionId -> Client ConnectionId socket IO request)
+  -> Connections ConnectionId socket request (Concurrent.Reject reject) (Concurrent.Accept (ConnectionHandle IO)) IO
```
coot: could you break this line, it's too long

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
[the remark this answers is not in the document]
coot: I see, it makes sense.

ouroboros-network/src/Ouroboros/Network/Connections/Types.hs
coot: I take my comment back.  But I would add a comment that the closure of the `Connection` contains a mutable state of all the connections, which justifies the name.

ouroboros-network/src/Ouroboros/Network/Connections/Socket/Server.hs
```diff
@@ -0,0 +1,158 @@
...
+    Just peeraddr -> do
+      -- Note the order of the addresses in the identifier: we always put our
+      -- bind address first. The client (sibling module Client.hs) also puts
+      -- the bind address first.
+      let connId = makeConnectionId bindaddr peeraddr
+      -- Including the connection could fail exceptionally, in which case we
+      -- are still responsible for closing the socket.
+      includeResult <- restore (includeConnection connId sock (Socket.close sock) (mkRequest peeraddr))
+                       `onException`
+                       Socket.close sock
```
coot: Are you sure we are not closing a socket that has already been closed by `includeConnection`? It's not clear to me.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -144,32 +168,97 @@ runDataDiffusion tracers
...
+        classifyRequest _ = PeerConnection
+
+        acceptException :: Socket.SockAddr -> IOException -> IO ()
+        acceptException a e = do
+          traceWith (WithAddr a `contramap` dtErrorPolicyTracer) $ ErrorPolicyAcceptException e
+
+        -- How to run a local server: take the `daLocalAddress` and run an
+        -- accept loop on it against the node-to-client connetions (yet to
+        -- be constructed).
+        runLocalServer n2cConnections = withSockType addr $ \addr' ->
```
coot: Alex, could you annotate this with its type. I think it is something like `Connections ... -> IO ...`

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -144,32 +168,97 @@ runDataDiffusion tracers
...
+          Server.acceptLoopOn addr' localClassifyRequest (acceptException addr)
+            n2cConnections
+          where
+          addr = Socket.addrAddress daLocalAddress
+
+        -- A node-to-node server will be run against a common connections
+        -- manager (yet to be constructed) but on potentially many different
+        -- bind addresses. This function will be mapped over the `daAddresses`.
+        runServer n2nConnections addrInfo = withSockType addr $ \addr' ->
```
coot: The same here, a type signature would make it easier to audit / review.

ouroboros-network/src/Ouroboros/Network/Connections/Socket/Server.hs
coot: it doesn't matter since `network` package protects us from that.

ouroboros-network/src/Ouroboros/Network/Connections/Concurrent.hs
avieth: I'll switch over to `MonadSTM.Strict`, but it really must be `MVar`-style semantics, because we need to do IO while holding the lock.

ouroboros-network/src/Ouroboros/Network/Connections/Types.hs
avieth: How about `RunClient` or `MkClient`?

ouroboros-network/src/Ouroboros/Network/Connections/Socket/Types.hs
avieth: Just because that's how the analogue type in `network` does it. It'd probably be fine to make it strict. But also, not really a big deal IMO since at worst we keep around 4 unevaluated thunks that are probably not very big. i.e. the computation of the host or port is probably not a very complicated closure.

avieth: Open to making it strict. I am indifferent.

ouroboros-network/src/Ouroboros/Network/Socket.hs
avieth: Yes this is something I was unsure about. What I gleaned was that the error policies are used by way of `completeApplicationTx`, which would be called whenever a connection ends. This thing has type

```Haskell
completeApplicationTx
  :: forall m addr a.
     ( MonadAsync  m
     , Ord addr
     , Ord (Async m ())
     )
  => ErrorPolicies addr a
  -> CompleteApplication m
       (PeerStates m addr)
       addr
       a
```

but what I'd like to have is

```Haskell
-- Takes the exception which was raised by the connection (or none), and gives the
-- exception which should be thrown.
runErrorPolicies :: ErrorPolicies addr a -> addr -> Maybe SomeException -> (a, Maybe SomeException)
```

Or something similar which can be used to derive

```Haskell
handleWithPolicy :: ErrorPolicies addr a -> addr -> IO () -> IO a
```

Then we can just tack on `handleWithPolicy` to the `runResponder` term.

ouroboros-network/src/Ouroboros/Network/Socket.hs
avieth: Is this a big deal? It's an exception, so almost surely not a very big thunk, right? And for how long does it live? Not very long I think: we finish that transaction to write `Finished` and then throw it.

If the caller leaks handles to these connections, that's a whole other story. Forcing this exception wouldn't make a difference.

ouroboros-network/src/Ouroboros/Network/Socket.hs
avieth: AFAICT the locally-initiated connections didn't use error policies before...

avieth: Or maybe it's just that it wasn't at all clear that they did? Maybe it was the subscription modules which attached the error policies?

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
avieth: No "port" is what I meant. Unclear language though. I meant to say, we should make it so that `mkResolverIO` does not hard-code one port for all resolutions. Rather, you should get back results which are IP addresses only. This way we could use one resolver for multiple different `dstPort`s.

avieth:
> Whey I run the demo
> 
> ```
> cabal run demo-connections -- 127.0.0.1 6000 2
> ```
> 
> I see these errors:
> 
> ```
> client 127.0.0.1:6001 : 127.0.0.1:6000 : unsupported operation : Network.Socket.connect: <socket: 14>: unsupported operation (Cannot assign requested address)
> 
> Incoming : ConnectionIdIPv4 127.0.0.1:6000 127.0.0.1:6001
> Outgoing : ConnectionIdIPv4 127.0.0.1:6001 127.0.0.1:6000
> demo-connections: Network.Socket.connect: <socket: 14>: unsupported operation (Cannot assign requested address)
> ```

The demo-connections executable spawns a configurable number of threads (the final parameter) each of which has a listening socket, and each of which makes a connection to each of the other threads, concurrently.

The problem is that there is always an interval between which `accept` returns, and the shared mutable state of a `Connections` term can be updated. In this interval, the OS has a socket for a connection identifier that the `Connections` term is not aware of. Since we bind to particular rather than ephemeral ports (to avoid crossed connections), this shows address already in use errors. This *can* happen even in a real deployment between TCP/IP stacks on different machines, but it's significantly less likely. If we add a random delay of at least 1000ms to the `connect` calls in the demo, the problem disappears (although if you use 32 threads or more you run out of file handles, as 1024 is a typical default limit; 32 * 31 + 32 = 1024).

ouroboros-network-framework/src/Ouroboros/Network/ConnectionId.hs
```diff
@@ -0,0 +1,21 @@
+{-# LANGUAGE DerivingVia #-}
```
avieth: This file extracted from a different file, so that `ConnectionId` is more widely available.

ouroboros-network-framework/src/Ouroboros/Network/ErrorPolicy.hs
```diff
@@ -47,7 +43,51 @@ import           Control.Monad.Class.MonadAsync
...
 import           Control.Monad.Class.MonadTime
 
-import           Ouroboros.Network.Subscription.PeerState
+-- | Semigroup of commands which acts on 'PeerState'.  The @t@ variable might
+-- be initiated to 'DiffTime' or @Time m@.
+--
+-- This semigroup allows to either suspend both consumer and producer or just
+-- the consumer part.
+--
+data SuspendDecision t
```
avieth: This was here before this patch. I don't know why it appears as a new addition.

ouroboros-network-framework/src/Ouroboros/Network/Subscription/Dns.hs
```diff
@@ -67,236 +53,82 @@ data Resolver m = Resolver {
...
+     , MonadCatch m
      , MonadThrow m
      )
-    => Tracer m DnsTrace
-    -> Resolver m
-    -> StrictTVar m s
-    -> BeforeConnect m s Socket.SockAddr
-    -> DnsSubscriptionTarget
-    -> m (SubscriptionTarget m Socket.SockAddr)
-dnsResolve tracer resolver peerStatesVar beforeConnect (DnsSubscriptionTarget domain _ _) = do
```
avieth: I took out this happy eyeball thing for simplicity. Maybe we want it back? AFAICT though it's not worth it: how often are we doing a resolve-to-connect? Not very. It was intended to be used by lookup-intensive interactive applications like web browsers, no?

Issue #593, opened 2019-05-30
Chain Sync Client: instantiate with header and BlockSize


mrBliss:
The Block Fetch logic wants to know the size of a block when we only have a header. If we instantiate the Chain Sync Client with the header *and* the `BlockSize`, we will have enough information.

Note that adding the block size to the header is not enough, as we still have to deal with blocks from the existing (real-world) chain.

mrBliss:
Ideally present for system-level performance testing. Needed for mainnet, as it could be a security risk.

mrBliss:
Fixing this means having to change the ChainSync protocol instantiation, breaking compatibility with already deployed nodes. So this should be done sooner rather than later (if it is not yet too late?).

If we don't fix it, the accounting for in flight bytes done by the BlockFetch client will be wrong, as we currently hardcode the size of each block to 2000.

@dcoutts Do you agree that we should do this ASAP? And is it still ok to make such breaking changes?

mrBliss:
Duncan's response:
> I think we have to do one [compatibility break] for other reasons in the network code anyway. Lets review with the network team and try to do them at the same time, or use the versioning scheme.

mrBliss:
Hmm, this is more invasive than I thought, as the ChainSyncServer will need to be able to return a header *and* the block size. The ImmutableDB will need to be modified to support this change. Even the implementation of the Reader has to be modified (other than simply forwarding the block size from the ImmutableDB): when returning the headers from the in-memory chain fragment (the last `k` blocks), we'll either need to keep their block sizes in memory too or retrieve them from the VolatileDB.

kderme:
> we'll either need to keep their block sizes in memory too or retrieve them from the VolatileDB.

A new api call getBlockSize makes sense for this?

mrBliss:
> > we'll either need to keep their block sizes in memory too or retrieve them from the VolatileDB.
> 
> A new api call getBlockSize makes sense for this?

See #1471 for a more general approach to this.

PR #1471, opened 2020-01-16
Use BlockComponent in the Storage API

### LINKED ISSUE #1372: Unify BlockComponent and BlockOrHeader
In the ImmutableDB and VolatileDB we have `BlockComponent` (previously `GetWhat`):

```haskell
-- ImmutableDB
data BlockComponent hash res where
  GetBlock  :: BlockComponent hash (hash, ByteString)
  GetHeader :: BlockComponent hash (hash, ByteString)
  GetHash   :: BlockComponent hash hash
-- VolatileDB
data BlockComponent = GetBlock | GetHeader
```

In the ChainDB we have `BlockOrHeader`:
```haskell
data BlockOrHeader blk b where
  Block  :: BlockOrHeader blk blk
  Header :: BlockOrHeader blk (Header blk)
```
We should try to unify these, or at least align these more. It will make it easier in the future to support things like "get the block *body*".

Also see https://github.com/input-output-hk/ouroboros-network/pull/1331#discussion_r359316352

We could go even further and include `SlotNo`, `Hash`, and `Point` as well. Or even `Deserialisable`, but that is already parameterised by `b` determined by `BlockOrHeader`.

mrBliss:
Closes #1372.

We have the ChainDB, ImmutableDB, and VolatileDB, that each provide operations to get a block or a header. The ImmutableDB also provides Iterators that can stream block or headers, and operations to get the hash of a block or EBB. The ChainDB provides Readers that can stream blocks or headers, and Iterators that can stream blocks (but not headers). In #1271 and #1331, we changed the ChainDB to return and stream `Deserialisable` blocks or headers. In the end, we had many variants of each operation, returning either headers, blocks, hashes, and/or `Deserialisable` variants thereof. Many variants were missing.

Moreover, to implement #593 (instantiate Chain Sync Client with header and block size), we should have the ability to let a Reader return the header *and* the block size, which would mean adding yet another variant or changing header everywhere to header *and* block size.

In this commit, we unify all such operations using the `BlockComponent` concept. All operations that could work for both block or headers work are now generalised to any `BlockComponent` *or combination thereof*.

We have the following `BlockComponent` type:

```haskell
data BlockComponent db a where
  GetBlock      :: BlockComponent db (DBBlock db)
  GetRawBlock   :: BlockComponent db ByteString
  GetHeader     :: BlockComponent db (DBHeader db)
  GetRawHeader  :: BlockComponent db ByteString
  GetHash       :: BlockComponent db (DBHeaderHash db)
  GetSlot       :: BlockComponent db SlotNo
  GetIsEBB      :: BlockComponent db IsEBB
  GetBlockSize  :: BlockComponent db Word32
  GetHeaderSize :: BlockComponent db Word16
  GetPure       :: a
                -> BlockComponent db a
  GetApply      :: BlockComponent db (a -> b)
                -> BlockComponent db a
```

This is a GADT so that the constructors determine the return type `a`. The `GetPure` and `GetApply` constructors allow this type to be a `Functor` and `Applicative`.

For example, to request a block from the ImmutableDB, one can now use `getBlockComponent GetBlock` instead of `getBlock`. Moreover, to request the header, slot, and block size, one can use `(,,) <*> GetHeader <*> GetSlot <*> GetBlockSize`. To request the `Point`, one can use `BlockPoint <$> GetSlot <*> GetHash`, and so on.

The three different databases each implement the following new type class:

```haskell
class DB db where
  type DBBlock     db
  type DBHeader    db
  type DBHeaderHash db
```

The ImmutableDB and VolatileDB don't support parsed blocks, only raw blocks, and thus have: `DBBlock (ImmutableDB hash m) = ()` and so on. The ChainDB, on the other hand, supports them and has `DBBlock (ChainDB m blk) = m blk`. Note the `m` there, indicating that parsing a block is monadic, as it can result in throwing parse failure exception.

The `Deserialisable` type was removed in favour of `SerialisedWithPoint`, which does not have an `m` parameter and thus no field to deserialise the block. This functionality is redundant anyway, now that there is `GetBlock`.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/API.hs
@@ -1,12 +1,14 @@
-{-# LANGUAGE AllowAmbiguousTypes        #-}
 {-# LANGUAGE DeriveAnyClass             #-}
-{-# LANGUAGE DeriveFunctor              #-}
 {-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DeriveTraversable          #-}
 {-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE GADTs                      #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE NamedFieldPuns             #-}
 {-# LANGUAGE PatternSynonyms            #-}
+{-# LANGUAGE RankNTypes                 #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeApplications           #-}
@@ -16,11 +18,17 @@
 module Ouroboros.Storage.ChainDB.API (
     -- * Main ChainDB API
     ChainDB(..)
-    -- * Block or header
-  , BlockOrHeader(..)
-    -- * Deserialisable block/header
-  , Deserialisable(..)
-  , deserialisablePoint
+    -- * Useful utilities
+  , getBlock
+  , streamBlocks
+  , newBlockReader
+    -- * Serialised block/header with its point
+  , SerialisedWithPoint(..)
+  , getSerialisedBlockWithPoint
+  , getSerialisedHeaderWithPoint
+  , getPoint
+    -- * BlockComponent
+  , BlockComponent(..)
     -- * Support for tests
   , toChain
   , fromChain
@@ -30,9 +38,8 @@ module Ouroboros.Storage.ChainDB.API (
   , Iterator(..)
   , IteratorId(..)
   , IteratorResult(..)
+  , traverseIterator
   , UnknownRange(..)
-  , deserialiseIterator
-  , deserialiseIteratorResult
   , validBounds
   , streamAll
     -- * Invalid block reason
@@ -40,9 +47,11 @@ module Ouroboros.Storage.ChainDB.API (
     -- * Readers
   , Reader(..)
   , ReaderId
-  , deserialiseReader
+  , traverseReader
     -- * Recovery
   , ChainDbFailure(..)
+  , BlockRef(..)
+  , IsEBB(..)
     -- * Exceptions
   , ChainDbError(..)
   ) where
@@ -60,10 +69,10 @@ import           Cardano.Prelude (NoUnexpectedThunks)
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment)
 import           Ouroboros.Network.Block (BlockNo, pattern BlockPoint,
                      ChainUpdate, pattern GenesisPoint, HasHeader (..),
-                     HeaderHash, MaxSlotNo, Point, Serialised, SlotNo,
+                     HeaderHash, MaxSlotNo, Point, Serialised (..), SlotNo,
                      StandardHash, atSlot, genesisPoint)
 
-import           Ouroboros.Consensus.Block (GetHeader (..))
+import           Ouroboros.Consensus.Block (GetHeader (..), IsEBB (..))
 import           Ouroboros.Consensus.Ledger.Abstract (ProtocolLedgerView)
 import           Ouroboros.Consensus.Ledger.Extended
 import           Ouroboros.Consensus.Util.IOLike
@@ -71,7 +80,7 @@ import           Ouroboros.Consensus.Util.ResourceRegistry
 import           Ouroboros.Consensus.Util.STM (WithFingerprint)
 
 import           Ouroboros.Storage.Common
-import           Ouroboros.Storage.FS.API.Types (FsError)
+import           Ouroboros.Storage.FS.API.Types (FsError, sameFsError)
 import qualified Ouroboros.Storage.ImmutableDB as ImmDB
 import qualified Ouroboros.Storage.VolatileDB as VolDB
 
@@ -185,8 +194,10 @@ data ChainDB m blk = ChainDB {
       -- Will return 'genesisBlockNo' if the database is empty.
     , getTipBlockNo      :: STM m BlockNo
 
-      -- | Get block at the specified point (if it exists)
-    , getBlock           :: Point blk -> m (Maybe blk)
+      -- | Get the given component(s) of the block at the specified point. If
+      -- there is no block at the given point, 'Nothing' is returned.
+    , getBlockComponent  :: forall b. BlockComponent (ChainDB m blk) b
+                         -> Point blk -> m (Maybe b)
 
       -- | Return membership check function for recent blocks
       --
@@ -238,10 +249,12 @@ data ChainDB m blk = ChainDB {
       --
       -- To stream all blocks from the current chain, use 'streamAll', as it
       -- correctly handles an empty ChainDB.
-    , streamBlocks
-        :: ResourceRegistry m
+    , stream
+        :: forall b.
+           ResourceRegistry m
+        -> BlockComponent (ChainDB m blk) b
         -> StreamFrom blk -> StreamTo blk
-        -> m (Either (UnknownRange blk) (Iterator m (Deserialisable m blk blk)))
+        -> m (Either (UnknownRange blk) (Iterator m blk b))
 
       -- | Chain reader
       --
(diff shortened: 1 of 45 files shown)


mrBliss:
Closes #1372.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Reader.hs
```diff
@@ -274,48 +266,69 @@ instructionHelper registry encodeHeader blockOrHeader fromMaybeSTM CDB{..} varRe
...
-      cdbImmDB cdbVolDB Block (headerPoint hdr)
+    headerUpdateToBlockComponentUpdate
+      :: f (ChainUpdate blk (Header blk)) -> m (f (ChainUpdate blk b))
+    headerUpdateToBlockComponentUpdate =
+      traverse (traverse (`headerToBlockComponent` blockComponent))
+
+    -- | We only got the header for the in-memory chain fragment, so depending
+    -- on the 'BlockComponent' that's requested, we might have to read the
+    -- whole block.
+    headerToBlockComponent
```
mrBliss: Rename to `getBlockComponentFromHeader`

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Reader.hs
```diff
@@ -274,48 +266,69 @@ instructionHelper registry encodeHeader blockOrHeader fromMaybeSTM CDB{..} varRe
...
+      :: forall b'. Header blk -> BlockComponent (ChainDB m blk) b' -> m b'
+    headerToBlockComponent hdr = \case
+        GetBlock      -> getBlockComponent GetBlock
+        GetRawBlock   -> getBlockComponent GetRawBlock
+        GetHeader     -> return $ return hdr
+        GetRawHeader  -> return $ toLazyByteString $ encodeHeader hdr
+        GetHash       -> return $ headerHash hdr
+        GetSlot       -> return $ blockSlot hdr
+        GetIsEBB      -> getBlockComponent GetIsEBB
```
mrBliss: Change `cdbIsEBB` to `Header blk -> Maybe EpochNo` and use that here.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Reader.hs
```diff
@@ -274,48 +266,69 @@ instructionHelper registry encodeHeader blockOrHeader fromMaybeSTM CDB{..} varRe
...
+        GetBlockSize  -> getBlockComponent GetBlockSize
+        -- We could look up the header size in the index of the VolatileDB,
+        -- but getting the serialisation is cheap, the following way is less
```
mrBliss: Say why it is cheap.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
```diff
@@ -504,6 +522,24 @@ readerClose rdrId m
...
+--
+-- The real ChainDB takes these as function arguments. For convenience (we
+-- don't want to pass around an environment throughout the model and the state
+-- machine tests), we bundle them in a testing-only type class.
+class ( HasHeader blk
+      , GetHeader blk
+      , HasHeader (Header blk)
+      , Serialise blk
+      , Serialise (Header blk)
+      ) => SupportedBlock blk where
```
mrBliss: Rename to `ModelSupportsBlock`

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -197,25 +201,51 @@ deriving instance SOP.HasDatatypeInfo (Cmd blk it rdr)
...
+  | MbPoint             (Maybe (Point blk))
+  | MaxSlot             MaxSlotNo
   deriving (Functor, Foldable, Traversable)
 
+-- | Product of all 'BlockComponent's. As this is a GADT, generating random
+-- values of it (and combinations!) is not so simple. Therefore, we just
+-- always request all block components.
+allComponents :: BlockComponent (ChainDB m blk) (AllComponents blk)
+allComponents = (,,,,,,)
+       -- TODO GetBlock and GetHeader
```
mrBliss: Try to get rid of the `m` in `run` by using `traverseIterator` and `traverseReader`, etc.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/API.hs
```diff
@@ -396,34 +445,28 @@ deserialiseIterator it = Iterator
 -- NOTE: Iterators created by /different instances of the DB/ may end up with
 -- the same ID. This should not matter in practice since there should not /be/
 -- more than one DB, but it should nonetheless be noted.
-instance Eq (Iterator m blk) where
+instance Eq (Iterator m blk b) where
   (==) = (==) `on` iteratorId
 
+-- TODO use IteratorID from ImmutableDB.Types (move that to Common)
```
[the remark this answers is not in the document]
mrBliss: I'll use the existing `IteratorID` for now and leave the rest for #1477.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/API.hs
```diff
@@ -438,18 +436,8 @@ traverseIterator f it = it {
...
---
--- NOTE: Iterators created by /different instances of the DB/ may end up with
--- the same ID. This should not matter in practice since there should not /be/
--- more than one DB, but it should nonetheless be noted.
-instance Eq (Iterator m blk b) where
-  (==) = (==) `on` iteratorId
-
-newtype IteratorId = IteratorId Int
-  deriving stock (Show)
+newtype IteratorId = IteratorId Word
```
mrBliss: Rename it to `IteratorKey` + add a comment (why do we need it?)

Same for `ReaderId` + say why it's not a newtype.

Export them for testing purposes only.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/API.hs
```diff
@@ -424,7 +423,6 @@ data StreamTo blk =
 data Iterator m blk b = Iterator {
       iteratorNext  :: m (IteratorResult blk b)
     , iteratorClose :: m ()
```
mrBliss: Add a comment saying that an `fmap`-ed or traversed iterator still refers to the original one, so closing either will close both.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -332,6 +334,10 @@ run ChainDB{..} internal registry varCurSlot = \case
           intScheduledChainSelection internal slot
       addBlock blk
 
+    giveWithEq :: a -> m (WithEq a)
+    giveWithEq a =
+      fmap (`WithEq` a) $ atomically $ updateTVar varNextId $ \i -> (succ i, i)
```
mrBliss: Move this to `WithEq` and use a global counter (`StrictTVar`).

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -287,9 +288,10 @@ run :: forall m blk. (IOLike m, HasHeader blk)
     -> ChainDB.Internal m blk
     -> ResourceRegistry m
     -> StrictTVar m SlotNo
-    ->    Cmd     blk (Iterator m blk (AllComponents blk)) (Reader m blk (AllComponents blk))
-    -> m (Success blk (Iterator m blk (AllComponents blk)) (Reader m blk (AllComponents blk)))
-run ChainDB{..} internal registry varCurSlot = \case
+    -> StrictTVar m Id
+    ->    Cmd     blk (WithEq (Iterator m blk (AllComponents blk))) (WithEq (Reader m blk (AllComponents blk)))
```
mrBliss: Create a type alias for `WithEq (Iterator m blk (AllComponents blk))` (same for `Reader`).

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ImmutableDB/StateMachine.hs
```diff
@@ -1095,8 +1106,8 @@ instance CommandNames (At CmdErr m) where
   cmdNames (_ :: Proxy (At CmdErr m r)) =
     constrNames (Proxy @(Cmd (IterRef m r)))
 
-instance Show (Iterator hash m a) where
-  show it = "<iterator " <> show (iteratorID it) <> ">"
+-- instance Show (Iterator hash m a) where
+--   show it = "<iterator " <> show (iteratorID it) <> ">"
```
mrBliss: Remove

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
mrBliss: Unfortunately, we don't have an `unsafePerformIO` (needed to created the global counter) that we can use in the simulator and the real IO monad.

So we're sticking with the current approach.

PR #1474, opened 2020-01-17
Instantiate node-to-node ChainSync with WithBlockSize (Header blk)

### LINKED ISSUE #593: Chain Sync Client: instantiate with header and BlockSize

mrBliss:
Closes #593.

* Modify the ChainSync codec.

* The current chain fragment of the ChainDB now stores `WithBlockSize (Header
  blk)` instead of `Header`.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Block.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Block.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE DeriveAnyClass          #-}
 {-# LANGUAGE DeriveGeneric           #-}
-{-# LANGUAGE DeriveTraversable       #-}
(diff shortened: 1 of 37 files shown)


mrBliss:
This depends on #1471.

Note that this PR **changes the network protocol**, breaking compatibility with deployed nodes. The version number in the network code should thus be bumped. I was told the network team were also planning to make a breaking change. We should plan so that both breaking changes make it into the same release.

mrBliss:
Closing because we're going to take a different approach. In Shelley, the header will already contain the block size. We can just modify the Byron header that is sent across the network (but not the on-disk one!) to include the block size.

PR #1476, opened 2020-01-20
More intelligent transaction requests

### LINKED ISSUE #1161: More intelligent transaction requests
**UPDATE:** new plan described in https://github.com/input-output-hk/ouroboros-network/issues/1161#issuecomment-594423279

Old plan:

---

Currently the `txSubmissionInbound` protocol handler does not look at the Mempool to see if txs are already there before requesting them. So in practice, it will needlessly download txs it already has. More confusingly, we can observe that node A sends a tx to B, which then sends it right back to A.

There are multiple options to improve this situation:
1. Look at the Mempool and don't request txs already in there.
2. Maintain a shared (across all `txSubmissionInbound` protocol handlers) set of `TxId`s that were recently in the Mempool. They should also expire from this set, so perhaps a PSQ would be a good fit. This can subsume (1).
3. Keep track of which `TxId`s we're asking from which peers and don't ask for the same one from multiple peers unless that would otherwise cause things to stall, in which case we should make another request for the same `TxId`.

Because of the more complex concurrency of (3), we'll keep that for later. Getting (1) working will be easy. Our current aim is to achieve (2).

intricate:
Closes #1161

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/General.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/General.hs
@@ -24,6 +24,7 @@ module Test.ThreadNet.General (
   ) where
 
 import           Control.Monad (guard)
+import           Data.Map (Map)
 import qualified Data.Map as Map
 import           Data.Set (Set)
 import qualified Data.Set as Set
@@ -38,6 +39,8 @@ import           Ouroboros.Network.Block (BlockNo (..), pattern BlockPoint,
                      SlotNo (..), blockPoint)
 import qualified Ouroboros.Network.MockChain.Chain as MockChain
 import           Ouroboros.Network.Point (WithOrigin (..))
+import qualified Ouroboros.Network.RecentTxIds as RecentTxIds
+import           Ouroboros.Network.RecentTxIds (RecentTxIds (..))
 
 import           Ouroboros.Consensus.Block (BlockProtocol)
 import           Ouroboros.Consensus.BlockchainTime
@@ -267,6 +270,7 @@ prop_general ::
      , HasHeader blk
      , RunNode blk
      , Show (LedgerView (BlockProtocol blk))
+     , Show (TxId (GenTx blk))
      )
   => (blk -> Word64) -- ^ Count transactions
   -> SecurityParam
@@ -299,6 +303,7 @@ prop_general countTxs k TestConfig{numSlots, nodeJoinPlan, nodeRestarts, nodeTop
         (Map.elems nodeChains) .&&.
     prop_all_growth .&&.
     prop_no_unexpected_message_delays .&&.
+    prop_no_txs_repeatedly_requested .&&.
     conjoin
       [ fileHandleLeakCheck nid nodeDBs
       | (nid, nodeDBs) <- Map.toList nodeOutputDBs ]
@@ -392,8 +397,9 @@ prop_general countTxs k TestConfig{numSlots, nodeJoinPlan, nodeRestarts, nodeTop
           where
             NodeJoinPlan mjoin = nodeJoinPlan
 
-    nodeChains    = nodeOutputFinalChain <$> testOutputNodes
-    nodeOutputDBs = nodeOutputNodeDBs    <$> testOutputNodes
+    nodeChains     = nodeOutputFinalChain <$> testOutputNodes
+    nodeOutputDBs  = nodeOutputNodeDBs    <$> testOutputNodes
+    nodeTxRequests = nodeOutputTxRequests <$> testOutputNodes
 
     isConsensusExpected :: Bool
     isConsensusExpected = consensusExpected k nodeJoinPlan schedule
@@ -586,6 +592,32 @@ prop_general countTxs k TestConfig{numSlots, nodeJoinPlan, nodeRestarts, nodeTop
                 _            -> False
               where
                 LeaderSchedule sched = actualLeaderSchedule
+
+    -- Ensure that transactions that have been recently requested and added to
+    -- the mempool are not requested again.
+    --
+    -- i.e. Transactions that are in a node's 'RecentTxIds' should not be
+    -- requested from a peer.
+    prop_no_txs_repeatedly_requested :: Property
+    prop_no_txs_repeatedly_requested =
+        counterexample ("nodeTxRequests: " <> show counterexampleTxReqs) $
+        counterexample counterexampleMsg $
+        property (all noIntersecting txRequests)
+      where
+        counterexampleTxReqs :: Map NodeId [([GenTxId blk], [(GenTxId blk, Time)])]
+        counterexampleTxReqs = Map.map
+          (map (\(txids, recentTxIds) -> (txids, RecentTxIds.toList recentTxIds)))
+          nodeTxRequests
+
+        counterexampleMsg :: String
+        counterexampleMsg =
+          "Transactions that are in the RecentTxIds have been requested."
+
+        txRequests :: [([GenTxId blk], RecentTxIds (GenTxId blk))]
+        txRequests = concat (Map.elems nodeTxRequests)
+
+        noIntersecting :: ([GenTxId blk], RecentTxIds (GenTxId blk)) -> Bool
+        noIntersecting = null . (uncurry RecentTxIds.intersection)
 
     hasNodeRekey :: Bool
     hasNodeRekey =
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -52,11 +52,14 @@ import           GHC.Stack
 import           Network.TypedProtocol.Channel
 import           Network.TypedProtocol.Codec (AnyMessage (..), CodecFailure,
                      mapFailureCodec)
+import           Network.TypedProtocol.Driver (TraceSendRecv (..))
 
 import           Ouroboros.Network.Block
+import           Ouroboros.Network.BlockFetch.ClientState (TraceLabelPeer (..))
 import           Ouroboros.Network.MockChain.Chain (Chain (Genesis))
 import           Ouroboros.Network.Point (WithOrigin (..), fromWithOrigin)
 import qualified Ouroboros.Network.RecentTxIds as RecentTxIds
+import           Ouroboros.Network.RecentTxIds (RecentTxIds (..))
 
 import qualified Ouroboros.Network.BlockFetch.Client as BFClient
 import           Ouroboros.Network.Protocol.ChainSync.PipelineDecision
@@ -654,11 +657,22 @@ runThreadNetwork ThreadNetworkArgs
             }
 
       nodeKernel <- initNodeKernel nodeArgs
-      let app = consensusNetworkApps
+
+      let instrumentedProtocolTracers = nullProtocolTracers
+            { ptTxSubmissionTracer = Tracer $ \(TraceLabelPeer _peerid traceSendRecv) ->
+                case traceSendRecv of
+                  -- We only care about sent requests for transactions.
+                  TraceSendMsg (AnyMessage (MsgRequestTxs txids)) -> do
+                    recentTxIds <- atomically $ readTVar (getRecentTxIds nodeKernel)
+                    traceWith (nodeEventsTxRequests nodeInfoEvents) (txids, recentTxIds)
+                  _ -> pure ()
+            }
+
+          app = consensusNetworkApps
                   nodeKernel
                   -- these tracers report every message sent/received by this
                   -- node
-                  nullDebugProtocolTracers
+                  (instrumentedProtocolTracers <> nullDebugProtocolTracers)
                   (customProtocolCodecs pInfoConfig)
                   (protocolHandlers nodeArgs nodeKernel)
 
@@ -944,6 +958,10 @@ data NodeEvents blk ev = NodeEvents
     -- ^ the point of every 'ChainDB.InvalidBlock' event
   , nodeEventsTipBlockNos :: ev (SlotNo, WithOrigin BlockNo)
     -- ^ 'ChainDB.getTipBlockNo' for each node at the onset of each slot
+  , nodeEventsTxRequests  :: ev ([GenTxId blk], RecentTxIds (GenTxId blk))
+    -- ^ every request for transactions made by the node-to-node
+    -- 'TxSubmission' mini-protocol along with the 'RecentTxIds' at the time
+    -- each request was made.
   }
 
 -- | A vector with an element for each database of a node
@@ -968,9 +986,10 @@ newNodeInfo = do
       (t2, m2) <- recordingTracerTVar
       (t3, m3) <- recordingTracerTVar
       (t4, m4) <- recordingTracerTVar
+      (t5, m5) <- recordingTracerTVar
       pure
-          ( NodeEvents     t1     t2     t3     t4
-          , NodeEvents <$> m1 <*> m2 <*> m3 <*> m4
+          ( NodeEvents     t1     t2     t3     t4     t5
+          , NodeEvents <$> m1 <*> m2 <*> m3 <*> m4 <*> m5
           )
 
   (nodeInfoDBs, readDBs) <- do
@@ -1001,6 +1020,7 @@ data NodeOutput blk = NodeOutput
   , nodeOutputNodeDBs    :: NodeDBs MockFS
   , nodeOutputForges     :: Map SlotNo blk
   , nodeOutputInvalids   :: Map (Point blk) [ExtValidationError blk]
+  , nodeOutputTxRequests :: [([GenTxId blk], RecentTxIds (GenTxId blk))]
   }
 
 data TestOutput blk = TestOutput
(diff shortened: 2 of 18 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -277,6 +278,7 @@ mkNodeArgs registry cfg initState tracers btime chainDB isProducer = NodeArgs
     , blockProduction
     , blockFetchSize      = nodeBlockFetchSize
     , blockMatchesHeader  = nodeBlockMatchesHeader
+    , recentTxIdsExpiryThresh = Time (secondsToDiffTime 3600)
```
intricate: I'll fix the alignment in another commit :smiley:. Just didn't want to add much `diff` noise to this draft PR.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -98,6 +104,36 @@ data ServerState txid tx = ServerState {
...
+       --                    made some changes since. Don't forget to update
+       --                    it.
+       --
+       --   For example, if we request some transaction IDs and otice that
+       --   some subset of them already exist within our mempool, we wouldn't
+       --   want to bother asking for those specific transactions since we
+       --   already have them. Therefore, we would just add those transaction
+       --   IDs mapped to 'Nothing' to the 'bufferedTxs' such that those
+       --   transactions are eventually acknowledged, but never actually
+       --   requested.
```
intricate: As I note in the `TODO` comment, this paragraph is slightly outdated. Ignore it for now.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -263,36 +322,73 @@ txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
...
-        SendMsgRequestTxsPipelined
+                 -> m (ServerStIdle n txid tx m ())
+    serverReqTxs n st = do
+        -- We can pick txs out of order if we want to, and we can decide not
+        -- to pick some txs at all (if for example we have seen them already).
+        -- We don't have to pick them all in one go, we can stick to a limit
+        -- like maxTxToRequest.
+
+        -- TODO @intricate: Should this whole thing be an STM tx?
+        recentTxIds <- atomically $ readTVar recentTxIdsVar
```
mrBliss: How so? This `readTVar` is the only thing that needs `STM`, the remainder of this function is pure.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -98,6 +104,36 @@ data ServerState txid tx = ServerState {
...
+       --
+       --   For example, if we request some transaction IDs and otice that
+       --   some subset of them already exist within our mempool, we wouldn't
+       --   want to bother asking for those specific transactions since we
+       --   already have them. Therefore, we would just add those transaction
+       --   IDs mapped to 'Nothing' to the 'bufferedTxs' such that those
+       --   transactions are eventually acknowledged, but never actually
+       --   requested.
+       --
        bufferedTxs         :: Map txid (Maybe tx),
```
mrBliss: This record is the state of this server, so we should check whether this contains thunks to avoid space leaks. This record is not stored in a `StrictTVar` (or variant), but passed around. See https://github.com/input-output-hk/ouroboros-network/blob/5c90d6e6e57f46f7fe22aa5fd657554493a71948/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs#L786 for an example on how to check for thunks in such a state. Slightly annoying: this code lives in `ouroboros-network`, while the thunks checking code lives in consensus. So something should be moved at some point.

This is orthogonal to this PR, so let's create a separate issue for this.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -250,9 +301,17 @@ txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
...
                                  bufferedTxs' acknowledgedTxIds
 
-      _ <- mempoolAddTxs txsReady
+      addedTxIds <- mempoolAddTxs txsReady
+
+      -- Insert the transactions that were added to the mempool into the
+      -- 'RecentTxIds'.
+      let Time expThreshDiffTime = txIdExpThresh
+      currTime <- getMonotonicTime
+      _ <- atomically $ modifyTVar
```
mrBliss:
```suggestion
      atomically $ modifyTVar
```

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -263,36 +322,73 @@ txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
...
+        let (potentialTxsToRequest, availableTxids') =
+              Map.splitAt (fromIntegral maxTxToRequest) (availableTxids st)
+
+            -- Transactions to not request at all because we've already
+            -- recently added them to the mempool.
+            txsToIgnore = Set.fromList $
+              RecentTxIds.intersection (Map.keys potentialTxsToRequest) recentTxIds
+
+            -- Transactions to request now.
+            txsToRequest = Map.withoutKeys potentialTxsToRequest txsToIgnore
```
mrBliss: So `potentialTxsToRequest` are the first `maxTxToRequest` in `availableTxids` (ordered by `txid`, which has no real meaning). Then you filter out the ones already in `recentTxIds`. What if all of the `potentialTxsToRequest` are in `recentTxIds`? Then you wont request any txs, even though `availableTxids` might contain txs not in `recentTxIds`. You need something like `take maxTxToRequest $ filter <not in recentTxIds> (availableTxids st)`.

(I'll continue with the rest of this PR tomorrow.)

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
intricate: Oh, I hadn't realized

ouroboros-network/src/Ouroboros/Network/RecentTxIds.hs
```diff
@@ -0,0 +1,146 @@
...
+
+-- | \( O(1) \). Find the earliest expiry time of all the elements in the
+-- collection.
+--
+-- If the collection is empty, this will return 'Nothing'.
+earliestExpiryTime :: Ord txid => RecentTxIds txid -> Maybe Time
+earliestExpiryTime (RecentTxIds psq) = maybe
+  Nothing
+  (\(_, time, _) -> Just time)
+  (OrdPSQ.findMin psq)
```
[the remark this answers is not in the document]
intricate: That _is_ nicer :smiley:

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
intricate: Ah, right.

intricate: Looks like you resolved this

ouroboros-network/src/Ouroboros/Network/RecentTxIds.hs
```diff
@@ -0,0 +1,152 @@
...
+  (\k p _ acc -> (k, p):acc)
+  []
+  psq
+
+{-----------------------------------------------------------------------------
+  Orphan instances
+-----------------------------------------------------------------------------}
+
+deriving via OnlyCheckIsWHNF "OrdPSQ" (OrdPSQ k p v)
+  instance NoUnexpectedThunks (OrdPSQ k p v)
```
coot: Do we need this orphaned instance? Why not define`OnlyCheckIsWHNF` instance directly for `RecentTxIds`?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
intricate: Yup: https://github.com/input-output-hk/ouroboros-network/issues/1645

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -114,12 +144,16 @@ initialServerState = ServerState 0 Seq.empty Map.empty Map.empty 0
...
+  -> StrictTVar m (RecentTxIds txid)
+  -- ^ A collection of transaction IDs that we've most recently added to the
+  -- mempool from instances of the transaction submission server.
   -> TxSubmissionMempoolWriter txid tx idx m
   -> TxSubmissionServerPipelined txid tx m ()
-txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
+txSubmissionInbound _tracer maxUnacked recentTxIdsVar mpWriter =
     TxSubmissionServerPipelined (serverIdle Zero initialServerState)
   where
     --TODO: replace these fixed limits by policies based on TxSizeInBytes
```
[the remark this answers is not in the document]
intricate: @coot: Sure.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -128,15 +162,24 @@ txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
...
     serverIdle Zero st
         -- There are no replies in flight, but we do know some more txs we can
         -- ask for, so lets ask for them and more txids.
       | canRequestMoreTxs st
-      = serverReqTxs Zero st
+      = do
+        serverReqTxsRes <-serverReqTxs Zero st
+        case serverReqTxsRes of
+          Left st'  -> serverIdle Zero st'
+          Right res -> pure res
```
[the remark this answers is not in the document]
intricate: I've removed this altogether on my local branch, so marking this as resolved.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -250,49 +302,136 @@ txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
...
+          then
+            -- There are no transactions that can be requested.
+            -- Return the updated 'ServerState' and allow the caller to decide
+            -- how to proceed.
+            Left st'
+
+          else
+            Right $ SendMsgRequestTxsPipelined
+              (Map.keys txsToRequest)
+              (serverReqTxIds (Succ n) st')
```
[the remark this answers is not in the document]
intricate: Thanks! I've actually removed this altogether on my local branch, so marking this as resolved.

ouroboros-network/src/Ouroboros/Network/RecentTxIds.hs
mrBliss: Actually, let's put a proper instance next to the one for `IntPSQ` here: https://github.com/input-output-hk/ouroboros-network/blob/64006b7331b3f70580ee70fad7808e7a0fe242d2/ouroboros-consensus/src/Ouroboros/Consensus/Util/Orphans.hs#L101

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -128,15 +162,24 @@ txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
     maxTxIdsToRequest = 3 :: Word16
     maxTxToRequest    = 2 :: Word16
 
+    TxSubmissionMempoolWriter
+      { txId
+      , mempoolAddTxs
+      } = mpWriter
```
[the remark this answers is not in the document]
mrBliss: That would make the line too long.

mrBliss:
To test this, I have the following approach in mind:

Modify the protocol tracers of the ThreadNet tests to trace all requests for transactions. Also trace transactions added and expired to/from the `RecentTxIds`. In `prop_general`, check for each node (by looking at its traces) that it never requested a transaction that is still in its `RecentTxIds`.

nix/.stack.nix/ouroboros-network.nix
```diff
@@ -0,0 +1,169 @@
+{ system, compiler, flags, pkgs, hsPkgs, pkgconfPkgs, ... }:
```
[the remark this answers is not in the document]
intricate: Hadn't even noticed that. Thanks.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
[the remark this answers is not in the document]
intricate: Right. I find it looks cleaner this way unless I break up the function args line a bit.

ouroboros-network/src/Ouroboros/Network/RecentTxIds.hs
```diff
@@ -0,0 +1,152 @@
...
+
+-- | \( O(\log(n)) \). Determine whether the provided transaction ID is
+-- present in the 'RecentTxIds'.
+member :: Ord txid => txid -> RecentTxIds txid -> Bool
+member txId (RecentTxIds psq) = OrdPSQ.member txId psq
+
+-- | \( O(m*\log(n)) \). Determine which of the provided transaction IDs
+-- already exist within the provided 'RecentTxIds'.
+intersection :: Ord txid => [txid] -> RecentTxIds txid -> [txid]
+intersection txIds (RecentTxIds psq) = List.filter (`OrdPSQ.member` psq) txIds
```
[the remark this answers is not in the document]
intricate: I'm utilizing this in my implementation of the test you described (https://github.com/input-output-hk/ouroboros-network/pull/1476#issuecomment-588993248).

ouroboros-network/src/Ouroboros/Network/RecentTxIds.hs
intricate: @mrBliss: That instance for `IntPSQ` is in `ouroboros-consensus`. I could add a `Ouroboros.Network.Util.Orphans` module or, as Marcin suggested, define the instance for `RecentTxIds`.

ouroboros-consensus/test-consensus/Test/ThreadNet/Network.hs
```diff
@@ -642,7 +643,7 @@ runThreadNetwork ThreadNetworkArgs
             , blockProduction         = Just blockProduction
             , blockFetchSize          = nodeBlockFetchSize
             , blockMatchesHeader      = nodeBlockMatchesHeader
-            , recentTxIdsExpiryThresh = Time (secondsToDiffTime (60 * 60)) -- TODO
+            , recentTxIdsExpiryThresh = RecentTxIds.ExpiryThreshold (secondsToDiffTime (60 * 60))
```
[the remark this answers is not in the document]
intricate: Would it be obvious to someone reading the code that the value is interpreted in seconds? If so, I can make the change.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -941,6 +955,10 @@ data NodeEvents blk ev = NodeEvents
     -- ^ the point of every 'ChainDB.InvalidBlock' event
   , nodeEventsTipBlockNos :: ev (SlotNo, WithOrigin BlockNo)
     -- ^ 'ChainDB.getTipBlockNo' for each node at the onset of each slot
+  , nodeEventsTxRequests  :: ev ([GenTxId blk], RecentTxIds (GenTxId blk))
+    -- ^ every request for transactions made by the node-to-node
+    -- 'TxSubmission' mini-protocol along with the 'RecentTxIds' at the time
+    -- each request was made.
```
mrBliss: So we store the `RecentTxIds` each time a request for txs is sent. I suppose it will be fine, but can you check that the memory usage while running the tests doesn't increase by too much? And can you check what the impact on the run time is? (Use the same seed to run the tests a few times with and without the new checks.) The tests are already slow enough :slightly_smiling_face:

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/General.hs
```diff
@@ -583,6 +589,32 @@ prop_general countTxs k TestConfig{numSlots, nodeJoinPlan, nodeRestarts, nodeTop
               where
                 LeaderSchedule sched = actualLeaderSchedule
 
+    -- Ensure that transactions that have been recently requested and added to
+    -- the mempool are not requested again.
+    --
+    -- i.e. Transactions that are in a node's 'RecentTxIds' should not be
+    -- requested from a peer.
+    prop_no_txs_repeatedly_requested :: Property
+    prop_no_txs_repeatedly_requested =
```
mrBliss: Very nice! This is exactly what I meant (even simpler :+1:).

ouroboros-network/src/Ouroboros/Network/Util/Orphans.hs
```diff
@@ -0,0 +1,17 @@
...
+
+import           Cardano.Prelude (NoUnexpectedThunks)
+
+import           Control.Monad.Class.MonadTime (Time (..))
+
+{-------------------------------------------------------------------------------
+  NoUnexpectedThunks
+-------------------------------------------------------------------------------}
+
+deriving newtype instance NoUnexpectedThunks Time
```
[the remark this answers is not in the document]
intricate: Good idea, that'd work.

intricate:
Closed in favour of #1749 which implements the [new plan](https://github.com/input-output-hk/ouroboros-network/issues/1161#issuecomment-594423279).

PR #1478, opened 2020-01-20
test-consenus: update README

nfrisby:
@i-o-m requested an update to this document "so that [others] can see the good work here". ❤️ 

> We can set out what testing we have covered. What has been completed and what remains. Also some words that support the types of issues we've uncovered would be great.

@edsko, @mrBliss: do you see an glaring omissions? I'm looking forward to the subsequent update when #888 is in place.

@i-o-m: Is this current presentation as an amendment to the existing document acceptable for your use cases?


mrBliss:
I deleted the file in question in #1657 because it was totally out of date (sorry!)

PR #1480, opened 2020-01-21
Strengthen PBFT window check

edsko:
Don't require minimum window size; we are not computing a percentage of
slots _in the window_ but rather the maximum number of signature based on
the _maximum_ window size; hence, no need for a separate minimum window
check. This brings the spec in line with the implementation.


nc6:
Yes, this makes sense. Capping the minimum size made sense when we were checking the proportion, but not when we check the absolute number of blocks.

edsko:
> Capping the minimum size made sense when we were checking the proportion, but not when we check the absolute number of blocks.

Yeah, exactly.

PR #1483, opened 2020-01-21
Add blockSize to ABlockOrBoundaryHdr

### LINKED ISSUE #593: Chain Sync Client: instantiate with header and BlockSize

mrBliss:
Closes #593.

We can extend the `ABlockOrBoundaryHdr` type with the block size, as this type does not have to binary-compatible (when sent across the network) with old nodes, only with new nodes.

When decoding a header extracted from the bytestring of an on-disk block, we use the header envelope to prepend the block size.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Auxiliary.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Auxiliary.hs
@@ -12,6 +12,7 @@
 {-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE TypeFamilies        #-}
 
+{-# OPTIONS_GHC -Wno-orphans #-}
 -- | Auxiliary definitions to make working with the Byron ledger easier
 --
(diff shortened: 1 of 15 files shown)


mrBliss:
Note that this PR changes the header type the network protocol, *breaking binary compatibility with deployed nodes*. The version number in the network code should thus be bumped. I was told the network team were also planning to make a breaking change. We should plan so that both breaking changes make it into the same release.

edsko:
> The Auxiliary module should go, so parts of this need to move to cardano-ledger

Actually, probably this should stay consensus side, we just need to wrap.

edsko:
Will open a new PR updating this for the latest `master`.

edsko:
Superseded by #1632 .

edsko:
@mrBliss Didn't want to delete the branch without your ok, so maybe you can do that?

PR #1485, opened 2020-01-21
Also strengthen the PBFT check in Test.ThreadNet.Ref.RealPBFT

### LINKED ISSUE #1484: Re-establish monotonicity of RealPBFT join plan viability
The Quickcheck `Gen`erator for the RealPBFT node join plans uses a reference implementation to test if a prefix of a node join plan is _viable_, meaning it's possible for the remaining nodes to join (not before any already joined nodes) and still ensure that the chain always has at least `k` blocks in every span of `2k` slots. The current `Gen`erator uses a binary search as an optimization. This is sound if the the viability of a node's join slot is monotonic: slots up to X are viable, and slots after X are not.

However, on PR #1425 for Issue #1194 (dual ledger towards Byron STS txs), Edsko discovered a (relatively simple) test case where a particular viability check is apparently non-monotonic! It seems that `k = 4` reveals a richness I did not initially expect (I had only considered/explored `k = 5` and `k = 10`).

This Issue is to resolve this unsoundness (e.g. stop assuming monotonicity).

I set the `priority high` label because this is at least interfering if not blocking PR #1425.

Edit: I think the `Test.ThreadNet.Ref.RealPBFT.viable` function is indeed monotonic if `k <= n`, which is the only regime we have been using. Edsko's branch involved `k > n`. In that regime, the "weak" PBFT check is non-monotonic. However, PR 1480 removed that check from the actual PBFT implementation but not from the ref simulator. If we update the ref simulator to match that change, then it is monotonic even in the `k > n` regime. (I think the timing is somewhat coincidental probably because other consequences of `k > n` lead Edsko to also open PR 1480 today.)

nfrisby:
Fixes #1484.

This PR makes the same change to the abstract `RealPBFT` implementation used to generate sufficient join plans as were made to the actual PBFT implementation in PR #1480.

I'm not sure of the exact details, but my suspicion is that we were never generating a test case in which `n > k`. We had a `SecurityParam` consensus yielding either 5 or 10 in `RealPBFT` cases, and we had `arbitrary :: NumCoreNodes` of 2-5. But on Edsko's dual ledger branch, we had e.g. `k = 4` and `n = 5`. In that regime and with the "weaker" PBFT check, the viability of a partial node join plan's next slot is indeed non-monotonic.

But with PR 1480 and this PR, the viability is again monotonic, even in the `k < n` regime.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
@@ -333,7 +333,6 @@ pbftWindowSize (SecurityParam k) = CS.WindowSize k
 -- | Does the number of blocks signed by this key exceed the threshold?
 --
 -- Returns @Just@ the number of blocks signed if exceeded.
--- Returns 'Nothing' if not 'shouldCheckThreshold'
 exceedsThreshold :: PBftCrypto c
                  => PBftWindowParams
                  -> PBftChainState c -> PBftVerKeyHash c -> Maybe Word64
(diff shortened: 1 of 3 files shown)


edsko:
Superseded by #1425.

PR #1487, opened 2020-01-22
Minor decoupling in the consensus abstractions

edsko:
In order to run the Byron dual ledger, we need to decouple consensus from ledger concerns. This was mostly already done, but there were two things stopping the integration:

* `UpdateLedger` unnecessarily required `SupportedBlock`. It really shouldn't refer to `NodeConfig` at all (since that is a consensus concern)
* `ApplyTx` required transactions to have an identifier, but those two concepts are orthogonal. This is now `HasTxId`.


edsko:
Superseded by #1425.

PR #1488, opened 2020-01-22
Properly shut down a node, its DBs, and its threads

### LINKED ISSUE #1470: ImmDB.closeDB should not call releaseAll on the passed-in registry
@mrBliss and I discussed this on a call today.

There are at least two problems with this `releaseAll` call.

https://github.com/input-output-hk/ouroboros-network/blob/db34b1f6b9a6eff320c1661d9be72ba553fc6fd3/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl.hs#L287

  1. The registry may contain contents not owned by the `ImmDB`.
  1. If a call to `closeRegistry` ends up running a releaser that calls `ImmDB.closeDB`, then its internal call to `releaseAll` will throw `RegistryClosedException` since the registry's state has already been set to `Closed`.

Right now `ImmDB.closeDB` always relies on `releaseAll` to release any remaining open file handles. It should instead only rely on the registry for that in exceptional circumstances.

mrBliss:
Fixes #1470.

* Don't use `releaseAll` or `unsafeReleaseAll` in the ImmutableDB, as that   will terminate all background threads, not just the resources using the   ImmutableDB. Instead, use `Index.close` to shut down the background thread.

* Start and close the ChainDB using the `ResourceRegistry` instead of using   `withChainDB`, this makes sure that resources are released in the right   order. Previously, the ChainDB was being closed before the background   threads using it were terminated, resulting in `ClosedDBError`s.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
@@ -22,7 +22,8 @@ module Ouroboros.Storage.ImmutableDB.Impl.Index.Cache
     -- * Background thread
   , expireUnusedEpochs
     -- * Operations
-  , reset
+  , close
+  , restart
     -- ** On the primary index
   , readOffsets
   , readFirstFilledSlot
@@ -52,6 +53,8 @@ import           Data.Word (Word32, Word64)
 import           GHC.Generics (Generic)
 import           GHC.Stack (HasCallStack, callStack)
 
+import           Control.Monad.Class.MonadSTM.Strict (StrictTMVar, isEmptyTMVar,
+                     newEmptyTMVarM, putTMVar, tryTakeTMVar)
 import           Control.Monad.Class.MonadThrow (bracketOnError)
 import           Control.Monad.Class.MonadTime (Time (..))
 
@@ -353,6 +356,7 @@ data CacheEnv m hash h = CacheEnv
   , tracer      :: Tracer m TraceCacheEvent
   , cacheVar    :: StrictMVar m (Cached hash)
   , cacheConfig :: CacheConfig
+  , bgThreadVar :: StrictTMVar m (Thread m Void)
   }
 
 -- | Creates a new 'CacheEnv' and launches a background thread that expires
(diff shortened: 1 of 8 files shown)


ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -581,19 +587,34 @@ getEpochInfo cacheEnv epoch = do
...
 -- PRECONDITION: the background thread expiring unused past epochs must have
 -- been terminated.
-reset
+restart
   :: IOLike m
   => CacheEnv m hash h
   -> EpochNo  -- ^ The new current epoch
   -> m ()
-reset cacheEnv@CacheEnv { hasFS, err, hashInfo, registry, cacheVar } epoch = do
+restart cacheEnv epoch = do
```
mrBliss: Add a comment that this is used only in the tests (it doesn't leak past the ImmutableDB, because the only code path that triggers it is part of the `Internal`r ecord) and that we don't have to worry about race conditions involving `bgThreadVar`.

ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/State.hs
```diff
@@ -206,11 +207,7 @@ modifyOpenState ImmutableDBEnv { _dbHasFS = hasFS :: HasFS m h, .. } action = do
           -- In case of an unexpected error (not an exception), close the DB
           -- for safety
           ExitCaseSuccess (Left (UnexpectedError {})) ->
-              -- This function ('modifyOpenState') can be called from all
-              -- threads, not necessarily the one that opened the ImmutableDB
-              -- and the registry, so we must use 'unsafeReleaseAll' instead of
-              -- 'releaseAll'.
-              (DbClosed, unsafeReleaseAll _dbRegistry)
+              (DbClosed, cleanUp hasFS st)
```
mrBliss: Let `open` return an `OpenState` and `close` take one, so that we can make sure that we only clean up an `OpenState`. I.e., we cannot have accidentally closed the state already before trying to clean it up.

Same for `withOpenState`.

edsko:
:roll_eyes:

PR #1499, opened 2020-01-24
Snockets

coot:
This PR introduce snockets which abstract the interface for Berkeley sockets and named pipes (windows).

The PR consist of smaller reviewable and mostly buildable patches.

Highlight of changes:

Win32-network
--------------------

A bug fixed in `System.Win32.Async.connect`

network-mux
-----------------

* `runMuxWithPipes` handles named pipes as well
* new tests which include posix pipes, named pipes and queues
* mux-demo - windows named pipe demo using mux: an echo server and client

ouroboros-network
-------------------------

* `IOManager` - system indepdendent IO manager.  On Windows using `System.Win32.Async.withIOManager`, on posix it's a no-op
* `Snockets` - abstraction which handles both Berkeley sockets and Windows named pipes, together with smart constructors for both.
* `Snocket` shim layer with type alliases for different paltforms (posix / win32)
* demo-chain-sync - can now run on Windows with named pipes

future work
---------------
* `NodeToClient` and `NodeToNode` which are using named pipes for local clients on windows.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Diffusion.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Diffusion.hs
@@ -130,10 +130,8 @@ data DiffusionApplications = DiffusionApplications {
                                           ())
       -- ^ NodeToClient responder applicaton (server role)
 
-    , daErrorPolicies :: ErrorPolicies SockAddr ()
+    , daErrorPolicies :: ErrorPolicies
       -- ^ error policies
-      --
-      -- TODO: one cannot use `forall a. ErrorPolicies SockAddr a`
     }
 
 runDataDiffusion
@@ -218,7 +216,7 @@ runDataDiffusion tracers
       , laUnix = Nothing
       }
 
-    remoteErrorPolicy, localErrorPolicy :: ErrorPolicies SockAddr ()
+    remoteErrorPolicy, localErrorPolicy :: ErrorPolicies
     remoteErrorPolicy = NodeToNode.remoteNetworkErrorPolicy <> daErrorPolicies
     localErrorPolicy  = NodeToNode.localNetworkErrorPolicy <> daErrorPolicies
 
--- a/ouroboros-network/src/Ouroboros/Network/Diffusion.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Diffusion.hs
@@ -28,7 +28,7 @@ import           Network.Mux (MuxTrace (..), WithMuxBearer (..))
 import           Network.Socket (SockAddr, AddrInfo)
 import qualified Network.Socket as Socket
 
-import           Ouroboros.Network.Snocket (SocketSnocket)
+import           Ouroboros.Network.Snocket (SocketSnocket, LocalSnocket)
 import qualified Ouroboros.Network.Snocket as Snocket
 
 import           Ouroboros.Network.Protocol.Handshake.Type (Handshake)
@@ -82,7 +82,7 @@ data DiffusionTracers = DiffusionTracers {
 data DiffusionArguments = DiffusionArguments {
       daAddresses    :: [AddrInfo]
       -- ^ diffusion addresses
-    , daLocalAddress :: AddrInfo
+    , daLocalAddress :: FilePath
       -- ^ address for local clients
     , daIpProducers  :: IPSubscriptionTarget
       -- ^ ip subscription addresses
@@ -150,8 +150,15 @@ runDataDiffusion tracers
                  applications@DiffusionApplications { daErrorPolicies } =
     withIOManager $ \iocp -> do
 
-    let snocket :: SocketSnocket
+    let -- snocket for remote communication.
+        snocket :: SocketSnocket
         snocket = Snocket.socketSnocket iocp
+
+        -- snocket for local clients connected using Unix socket or named pipe.
+        -- we currently don't support remotely connected local clients.  If we
+        -- need to we can add another adress for local clients.
+        localSnocket :: LocalSnocket
+        localSnocket = Snocket.localSnocket iocp daLocalAddress
 
     -- networking mutable state
     networkState <- newNetworkMutableState
@@ -165,7 +172,7 @@ runDataDiffusion tracers
         Async.withAsync (cleanNetworkMutableState networkLocalState) $ \cleanLocalNetworkStateThread ->
 
           -- fork server for local clients
-          Async.withAsync (runLocalServer snocket networkLocalState) $ \_ ->
+          Async.withAsync (runLocalServer localSnocket networkLocalState) $ \_ ->
 
             -- fork servers for remote peers
             withAsyncs (runServer snocket networkState . Socket.addrAddress <$> daAddresses) $ \_ ->
@@ -181,8 +188,6 @@ runDataDiffusion tracers
                   Async.waitEither_ cleanNetworkStateThread cleanLocalNetworkStateThread
 
   where
-    -- TODO: this is POSIX only, Windows support will be built later
-
     DiffusionTracers { dtIpSubscriptionTracer
                      , dtDnsSubscriptionTracer
                      , dtDnsResolverTracer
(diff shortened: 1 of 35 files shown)


ouroboros-network/src/Ouroboros/Network/Snocket.hs
```diff
@@ -129,3 +134,115 @@ data Snocket m fd addr = Snocket {
...
+          Socket.setSocketOption sd Socket.ReusePort 1
+#endif
+        when (fml == Socket.AF_INET6)
+          -- An AF_INET6 socket can be used to talk to both IPv4 and IPv6 end points, and
+          -- it is enabled by default on some systems. Disabled here since we run a separate
+          -- IPv4 server instance if configured to use IPv4.
+          $ Socket.setSocketOption sd Socket.IPv6Only 1
+
+        Socket.bind sd addr
+    , listen   = \s -> Socket.listen s 1
```
karknu: A leftover from the old code, but we should increase the listening queue. Also in `rawSocketSnocket`.

ouroboros-network/src/Ouroboros/Network/Snocket.hs
```diff
@@ -129,3 +134,115 @@ data Snocket m fd addr = Snocket {
...
+        `catch` \(SomeAsyncException _) -> do
+          Socket.close sd
+          throwIO e
+      return sd
+
+
+
+-- | Create a snocket for the given 'Socket.Family'.
+--
+rawSocketSnocket
```
karknu: Add a comment on what makes this socket "raw" to make it clear why one would use this instead of `socketSnocket`.

ouroboros-network/src/Ouroboros/Network/Snocket.hs
```diff
@@ -129,3 +134,115 @@ data Snocket m fd addr = Snocket {
...
+      getLocalAddr  = Socket.getSocketName
+    , getRemoteAddr = Socket.getPeerName
+    , addrFamily    = socketAddrFamily
+    , connect       = \s a -> do
+#if !defined(mingw32_HOST_OS)
+        Socket.connect s a
+#else
+        Win32.Async.connect s a
+#endif
+    , bind          = \fd addr -> Socket.bind fd addr
```
karknu: I'm not sure about the missing `setSocketOption` here. Are these "raw" snockets only to be used with ephemeral port numbers on with non-InitatorAndResponder mux applications?

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -156,22 +165,23 @@ runDataDiffusion tracers
...
                 -- fork dns subscriptions
-                withAsyncs (runDnsSubscriptionWorker networkState <$> daDnsProducers) $ \_ ->
+                withAsyncs (runDnsSubscriptionWorker snocket networkState <$> daDnsProducers) $ \_ ->
 
                   -- If any other threads throws 'cleanNetowrkStateThread' and
                   -- 'cleanLocalNetworkStateThread' threads will will finish.
                   Async.waitEither_ cleanNetworkStateThread cleanLocalNetworkStateThread
 
   where
+    -- TODO: this is POSIX only, Windows support will be build later
```
karknu: s/build/built

ouroboros-network/src/Ouroboros/Network/Snocket.hs
```diff
@@ -129,3 +134,115 @@ data Snocket m fd addr = Snocket {
...
+socketAddrFamily (Socket.SockAddrInet6 _ _ _ _) = SocketFamily Socket.AF_INET6
+socketAddrFamily (Socket.SockAddrUnix _       ) = SocketFamily Socket.AF_UNIX
+
+type SocketSnocket = Snocket IO Socket SockAddr
+
+-- | Create a 'Snocket' for the given 'Socket.Family'. In the 'bind' method set
+-- 'Socket.ReuseAddr` and 'Socket.ReusePort'.
+--
+socketSnocket
+  :: AssociateWithIOCP
```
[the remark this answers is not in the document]
karknu: Arg, it is everywhere!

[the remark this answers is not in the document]
coot: It depends on the platform.  I think it's cleaner than using `#if` everywhere where we need to pass the io completion port.

ouroboros-network/src/Ouroboros/Network/Snocket.hs
coot: I actually wanted to ask you, what the value should be? Great that you picked it yourself :)

ouroboros-network/src/Ouroboros/Network/Snocket.hs
coot: we use it in:
* local clients - which are using unix sockets - so no need for binding;
* and in one of the tests
I think it would be cleaner to move this inside `#if` where `clientSnocket = rawSocketSnocket` is set.  This way it will be cleaner that it is only for local clients over unix sockets.

ouroboros-network/src/Ouroboros/Network/Snocket.hs
karknu: 8 sounds good to me, we can revisit it once we have an overall limit on the number of open file descriptors.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -223,16 +230,23 @@ connectTo_V1 tracers versionData application localAddr remoteAddr =
 -- * when an async exception is thrown to kill the main thread the server thread
 --   will be cancelled as well (by 'withAsync')
 --
+-- TODO: specialise to use `Socket`, node-to-node protocol is only used over
+-- TCP sockets.
```
karknu: I don't think we should limit node-to-node to only TCP.

ouroboros-network/src/Ouroboros/Network/Snocket.hs
```diff
@@ -10,6 +10,10 @@ module Ouroboros.Network.Snocket
   , SocketSnocket
   , socketSnocket
   , rawSocketSnocket
+  , ClientSnocket
```
karknu: I see that you got client from Client in Node-to-client, but I think it is rather confusing name. You introduce the term RemoteConnectionId in a later patch, how about using `Local` as a prefix instead of `Client`?

coot: Yes that will be less confusing.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
coot: I agree; I'll update this comment.

ouroboros-network/src/Ouroboros/Network/Snocket.hs
coot: Done

ouroboros-network/src/Ouroboros/Network/Snocket.hs
coot: Done; I also defined `LocalConnectionId` type alias.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
coot: Fixed; I finished the un-finished todos.  `NodeToNode` module is using `SocketSnocket` everywhere.  It's a public interface, so we don't want to be more general (polymorphic) than it is needed.

ouroboros-network/src/Ouroboros/Network/Snocket.hs
```diff
@@ -0,0 +1,405 @@
...
+  , SocketSnocket
+  , socketSnocket
+  , rawSocketSnocket
+    -- ** Client Snockets
+    -- Using unix sockets (posix) or named pipes (windows)
+    --
+  , LocalSnocket
+  , localSnocket
+  , LocalAddress
+  , clientAddressFromPath
```
karknu: Shouldn't this be `localAddressFromPath`?

coot: yes, it should :)

coot: done

coot:
bors +cancel

coot:
bors try

coot:
bors ping

PR #1504, opened 2020-01-24
WIP: limit messages per segment

karknu:
Limit messages to at one per mux segment in case of a single threaded
rts.


karknu:
Nope.

PR #1512, opened 2020-01-28
Lower default connection delay

karknu:
By lowering the default connection attempt delay there is a good chance
that a client will avoid peers not on the same continent.


njd42:
This is not quite the simple idea it may at first appear. 
Clients could easily be more that 50ms away from any peer (on certain networks).
Can I ask - what is the issue that is driving this change?

karknu:
> Can I ask - what is the issue that is driving this change?
The default testing of `cardano-node` should switch from using four connections  two ip addresses to 1 connection for a round robin dns name. This matches how cardano-sl is configured

Issue #1501, opened 2020-01-24
Windows API for ouroboros-network


coot:
This is a followup on #1499.  With Snockets expose NodeToNode and NodeToClient apis that will work on Windows with named pipes and on posix with Unix sockets.

coot:
Done in #1516

PR #1516, opened 2020-01-28
Windows compatiblity

coot:
This PR includes three things:
* simplifies `ErrorPolicies` by removing callback which classifies return values of error policies
* adds `clientSubscriptionWorker` which subscribes to a unix socket or named pipe on windows.  It is a simplified version of `ipSubscriptionWorker`.
* adds `./scripts/test.sh` - a script to run tests on widnows - all are green :tada: :sunglasses:

It depends on #1499 - which needs to be merge first, it addresses #1501.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
@@ -225,7 +225,7 @@ withServer
   -> LocalAddress
   -> Versions NodeToClientVersion DictVersion
               (OuroborosApplication appType (ConnectionId LocalAddress) NodeToClientProtocols IO BL.ByteString a b)
-  -> ErrorPolicies LocalAddress ()
+  -> ErrorPolicies
   -> IO Void
 withServer sn tracers networkState addr versions errPolicies =
   withServerNode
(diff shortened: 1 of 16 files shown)


coot:
I changed target to `coot/snockets` branch (#1499) - just to make it easier for the review.  After #1499 is merged I'll re-target it to `master` back again.

ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
```diff
@@ -290,44 +290,41 @@ ncSubscriptionWorker
...
-  NetworkIPSubscriptionTracers
-    { nistSubscriptionTracer
-    , nistMuxTracer
-    , nistHandshakeTracer
-    , nistErrorPolicyTracer
+  NetworkSubscriptionTracers
+    { nsSubscriptionTracer
+    , nsMuxTracer
+    , nsHandshakeTracer
+    , nsErrorPolicyTracer
```
karknu: In `cardano-sl`'s style the prefix would be `nst` not ns. But I'm not sure we have any rules for this in `ourboros-network`.

coot:
> How will this PR's subscription work interact with #1455? Is there a point to adding it?
Yes, once snockets are merged they can be used in `#1455` in the same way as they are used in `Ouroboros.Network.Server`. But we can add the snocket support later on, `byron` release does not require from us to be worried about window's users.

ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
coot: I know, but we already have `nst` prefix in `NetworkServerTracers` :/

coot:
> I have the same reservation on the use of client in this pr as on #1499.

Fixed :)

coot:
I merged this into #1499 - so I have less PRs to re-base.

PR #1518, opened 2020-01-28
Better ledger DB policy

### LINKED ISSUE #1264: Revise ledger DB disk snapshot taking policy
We should probably

~1. Take a snapshot when the node shuts down~ *Moved to #1387.*
2. Adjust it so that when doing bulk sync, we take snapshots faster than once every 12 hours.

It seems that currently we also take a snapshot after replying? Not sure where/how.

### LINKED ISSUE #1456: Do not write ledger snapshot without copying blocks to imm DB
At the moment there are two processes happening in the node that are unaware of each other:

1. There is a thread moving blocks from the vol DB to the imm DB.
2. There is a thread taking period on-disk snapshots of the ledger DB.

The problem is that if the latter thread takes a snapshot, it might consider a block to be immutable (and hence use it for the snapshot) which hasn't yet been copied into the imm DB by the former thread. When this happens, on start up the node will be unable to use this ledger snapshot and will have to default to an older one (#1454). 

The problem is exacerbated by the fact that the node now takes a ledger snapshot on shutdown, _without also doing a final volatile to immutable DB copy_. It's therefore not all that unlikely to happen. What I think needs to happen is

1. Only take a ledger DB snapshot after copying blocks, making sure that there aren't any race conditions (in other words, decide which ledger snapshot to take, _then_ copy to imm DB, _then_ write the snapshot).
2. Don't take the final snapshot on node shutdown, and instead adjust the snapshot policy (#1264).

edsko:
Closes #1456
Closes #1264

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Background.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Background.hs
@@ -16,7 +16,8 @@ module Ouroboros.Storage.ChainDB.Impl.Background
     launchBgTasks
     -- * Copying blocks from the VolatileDB to the ImmutableDB
   , copyToImmDB
-  , copyToImmDBRunner
+  , copyAndSnapshotRunner
+  , updateLedgerSnapshots
      -- * Executing garbage collection
   , garbageCollect
     -- * Scheduling garbage collections
@@ -25,8 +26,6 @@ module Ouroboros.Storage.ChainDB.Impl.Background
   , scheduleGC
   , gcScheduleRunner
     -- * Taking and trimming ledger snapshots
-  , updateLedgerSnapshots
-  , updateLedgerSnapshotsRunner
     -- * Executing scheduled chain selections
   , scheduledChainSelection
   , scheduledChainSelectionRunner
@@ -38,9 +37,11 @@ import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (fromMaybe)
 import           Data.Void (Void)
+import           Data.Word
 import           GHC.Stack (HasCallStack)
 
 import           Control.Monad.Class.MonadThrow
+import           Control.Monad.Class.MonadTime
 import           Control.Tracer
 
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment (..))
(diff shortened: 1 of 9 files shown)


ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Background.hs
```diff
@@ -186,12 +186,29 @@ copyToImmDB CDB{..} = withCopyLock $ do
     mustBeUnlocked = fromMaybe
                    $ error "copyToImmDB running concurrently with itself"
 
--- | Watches the current chain for changes. Whenever the chain is longer than
--- @k@, then the headers older than @k@ are copied from the VolatileDB to the
--- ImmutableDB (with 'copyToImmDB'). Afterwards, a garbage collection of the
--- VolatileDB is scheduled using ('scheduleGC') for the 'SlotNo' of the most
--- recent block that was copied.
-copyToImmDBRunner
+-- | Copy blocks from the vol DB to imm DB and take snapshots of the ledger DB
```
mrBliss: This docstring contains a mix of "vol DB", "VolatileDB" and "volatile DB" (same for the other two). Can you make it consistent and use either "VolatileDB" or "VolDB"?

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Background.hs
```diff
@@ -201,18 +218,54 @@ copyToImmDBRunner
...
-    atomically $ do
-      curChain <- readTVar cdbChain
-      check $ fromIntegral (AF.length curChain) > k
-
-    mSlotNo <- copyToImmDB cdb
-    case mSlotNo of
-      Origin    -> pure ()
-      At slotNo -> scheduleGC (contramap TraceGCEvent cdbTracer) slotNo cdbGcDelay gcSchedule
+copyAndSnapshotRunner cdb@CDB{..} gcSchedule = \replayed ->
+    loop Nothing replayed
```
mrBliss: I would replace `\replayed -> loop Nothing replayed` with `loop Nothing`.

edsko:
Trying this out is blocked on https://github.com/input-output-hk/cardano-node/pull/496.

mrBliss:
I have tested out this change (also played with the policy) using https://github.com/input-output-hk/cardano-node/pull/496 and it seems to be working fine. I believe we can merge this.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/Reopen.hs
```diff
@@ -144,7 +144,7 @@ reopen (CDBHandle varState) launchBgTasks = do
           cdbEpochInfo
           immDbTipPoint
           (contramap TraceLedgerReplayEvent cdbTracer)
-        LgrDB.reopen cdbLgrDB cdbImmDB lgrReplayTracer
+        replayed <- LgrDB.reopen cdbLgrDB cdbImmDB lgrReplayTracer
```
mrBliss: Line 80 of this file needs to be removed: it takes a snapshot on shut down, which is exactly a bug this PR is solving.

Issue #1411, opened 2020-01-07
Add test coverage


coot:
This requires redesigning `*.cabal` files.

PR #1519, opened 2020-01-28
Reorganise ouroboros-network modules

coot:
* add `ouroboros-protocol-tests` library. It includes all the Example
  and Test modules which were under `Ouroboros.Network.Protocols`, with
  one exception: `Ouroboros.Network.Protocols.ChainSync.Example` which is
  required for `test-conesnsus`.  We might be able to fix this in future
  by makeing ouroboros-protocol-test library public.
* `test-network` library depends on the internal
  `ouroboros-protocol-tests` library.


coot:
One step towards #1411

PR #1522, opened 2020-01-29
Enable the LocalStateQuery protocol, breaking network compatibility

mrBliss:


mrBliss:
This includes #1507 until that is merged. This PR is only about the final commit.

The `NodeToClientVersion` will also need to be bumped.

This should only be merged when we're ready to break "network compatibility".

ouroboros-network/src/Ouroboros/Network/Protocol/LocalStateQuery/Client.hs
```diff
@@ -0,0 +1,138 @@
...
+    , localStateQueryClientPeer
+    ) where
+
+import           Network.TypedProtocol.Core
+
+import           Ouroboros.Network.Block (Point)
+import           Ouroboros.Network.Protocol.LocalStateQuery.Type
+
+
+newtype LocalStateQueryClient block query result m a = LocalStateQueryClient {
```
coot: There's no need for the new type wrapper, but let it be this way.  We plan to remove all the top level newtype wrappers from all of the mini-protocols.

ouroboros-network/src/Ouroboros/Network/Protocol/LocalStateQuery/Client.hs
```diff
@@ -0,0 +1,138 @@
...
+-- | In the 'StAcquiring' protocol state, the client does not have agency.
+-- Instead it is waiting for:
+--
+--  * acquired
+--  * failure to acquire
+--
+-- It must be prepared to handle either.
+--
+data ClientStAcquiring block query result m a = ClientStAcquiring {
+      recvMsgAcquired ::    ClientStAcquired block query result m a,
```
coot: could you align the type

mrBliss:
> LGTM - but I only looked under `Ouroboros.Network.Protocol.LocalStateQuery`.

Thanks for giving this a look! BTW, #1507 is the PR that I want to merge now, not this one, which is #1507 + the breaking change that enables it.

> Somebody needs to review the rest.

That's what Edsko did in #1507 :slightly_smiling_face:

> You'll have a conflict with current master. To resolve it you will need to put `Examples,Tests,Direct` modules in the private library `ouroboros-protocol-tests`. If you need to re-use `Examples` in `ouroboros-consensus` you can leave it in the public library.

No, I rebased this morning already. I have moved `Tests` and `Direct` to `ouroboros-protocol-tests` and left `Examples` in the public library, as I indeed use it in the consensus tests.

ouroboros-network/src/Ouroboros/Network/Protocol/LocalStateQuery/Client.hs
mrBliss: Initially, I aligned all state types on their type parameters, even when some had an `m (..)` around them. Now that I have added so many `m`s, there's no point in alignment on this line, I'll remove it.

ouroboros-network/src/Ouroboros/Network/Protocol/LocalStateQuery/Client.hs
mrBliss: Ok. I'll leave it there for consistency.

edsko:
@karknu , this re-introduces the config parameter for the node-to-client protocols, _but_ it is only a `BlockConfig`, not a `TopLevelConfig`. For Byron, that is the Byron genesis config, which the wallet should have access to. See the comment for motivation:

```haskell
-- We pass the 'BlockConfig' here, even though it is currently unused. If at any
-- point we want to introduce local protocols that for example send Byron blocks
-- or headers across, we will need to have the epoch size, which comes from the
-- Byron config. Unlike the full 'TopLevelConfig', it should not be difficult
-- for a wallet to construct the 'BlockConfig'.
--
-- NOTE: Somewhat confusingly, 'pcChainSyncCodec' currently /does/ send Byron
-- blocks across, but it does not deserialize them (the user of the codec is
-- itself responsible for doing that), which is why it currently does not need
-- the config.
```

coot:
bors merge

Issue #1533, opened 2020-01-30
Chain DB q-s-m: relax precondition on reopening


mrBliss:
In the ChainDB q-s-m tests, we have a precondition on the `Close` command (which will be followed with a `Reopen` command): there cannot be an equally preferable or an even more preferable fork in the ChainDB, as the real implementation might switch to that chain when reopening, as it will perform chain selection. In the model we cannot predict which chain will be selected, as the model itself doesn't perform chain selection on startup.

The goal of this ticket is to relax that precondition: either allow more preferable forks (see the text below) or allow any fork.

> We're on a chain with 3 blocks, an EBB (blockNo = 0) and two regular blocks. The blockNo of our tip is 2
>
> Other blocks have been added such that the ChainDB contains a fork, containing 3 blocks, starting at genesis with a regular block. The blockNo of that chain's tip is 3. We can't switch to that chain since it forks off 3 blocks back and we have k = 2.
>
> Now we close our ChainDB without having run the background tasks. This means that our ImmutableDB is empty and that all blocks are in the VolatileDB. In practice, the EBB at the start of our chain would most likely, but not necessarily have been copied in the background to the ImmutableDB. When we now restart, chain selection starts with Genesis, which means that we select the fork, as its tip has blockNo = 3, which is the highest block number.
>
> In the model, closing and reopening just means toggling a Bool, no chain selection is run.
>
> NOTE: we currently have a precondition stating that we cannot close when there are multiple equally preferable chains, as the real implementation might pick another chain on reopening, going out of sync with the model. The precondition does not apply to the problem described above, as the fork is /more preferable/ than the current chain.
>
> Possible solution: let the model also perform chain selection when reopening. Problem: what if we had run the background tasks before closing? Then the EBB would have been copied to the ImmutableDB and we would have reselected the original chain when reopening, not the more preferable fork, as it now forks off too far in the past. The model, however, is not aware of blocks being part of the ImmutableDB, so chain selection would start from genesis again. Ah, but the model is aware of garbage collection, so it would have garbage collected the blocks of the fork that are older than the "immutable" block. Unfortunately, that block is the EBB with slot number 0, so the first block of the fork would still be part of the model's VolatileDB, as it has slot number 2.
>
> The model will have to be aware of which blocks have been copied to the ImmutableDB, so it can make the same decisions.

mrBliss:
[...]

PR #1537, opened 2020-01-30
ChainDB.StateMachine: strengthen the precondition on Close

mrBliss:
The ChainDB.StateMachine would very occasionally fail. Fix these failures by
strengthening the precondition on the `Close` command. See the comment for
more details.


mrBliss:
> Let's open a ticket to revisit this later.

You mean #1533? :grin:

edsko:
> You mean #1533? grin

That's exactly what I mean, how did you know? :D

Issue #1511, opened 2020-01-27
Test with varying `k` values in test-consensus


nfrisby:
This Issue is a follow-up to Issue #1489.

While working on #1489 et al, it became clear that `PBFT` and `RealPBFT` should use more `k` values than just `5` or `10`. In particular, the recently added `DualPBFT` tests use other values, and fixing #1489 unmasked related bugs there.

My imminent PR  for #1489 will resolve this for `k > 1`. But the `k = 1` case seems to be confounded in at least three ways. I don't want these to block a PR for #1489 -- hence this issue.

  1. `prevPointAndBlockNo` ignores the anchor point of the current chain and falls all the way back to `genesisPoint` instead. I don't know why only `k = 1` seems to trigger this. The tests fail with `the first block on the Byron chain must be an EBB...`.
  1. `Test.ThreadNet.Ref.PBFT.viable` and Byron's `anachronisticLedgerView` disagree on when `TooFarAhead` should be raised. In the `k = 1` case, "not viable" is "zeroes block in `2k` slots", which conceptually aligns with `TooFarAhead`. However, `TooFarAhead` uses `now + 2k` as an _exclusive_ upper bound. If I naively change it to _inclusive_ then I see no test failures of this kind. I doubt that's the right fix, however. The tests fail with `Missing or unexpected EBBs...`. (The net is stuck in a way the `Ref` simulator did not anticipate, so it's unable to produce anything, including the expected EBBs.)
  1. In the `ProduceEBB` case, a `RealPBFT` node may produce two blocks during a single slot: an EBB and a subsequent proper block. With `k = 1`, this might introduce a wedge. The test fail with `after dropping X blocks from...` (i.e. there's too deep of a fork in the final chains). (Another node might be able to fit a proper block before that EBB --- I think this requires a node to restart and also lead in the new EBB's slot -- it'll forge the new EBB and its subsequent block before syncing the pre-EBB slot from the others? Something like that, I think.)

PR #1544, opened 2020-01-31
consensus: do not assume the anchor point is genesis in prevPointAndBlockNo

nfrisby:
Fixes the first item in #1511.

Unfortunately I don't have a test case here, because the only repro I have only fails on my WIP branch for Issue 1489. On that branch, this commit does fix the repro.

I'm preparing an omnibus PR for 1489, but I thought I'd open this one too, since the fix seems clear in hindsight (unless I'm overlooking some invariant about the current chain here?) and it's in the library not just the test infrastructure.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
@@ -179,8 +179,7 @@ data TraceForgeEvent blk tx
   -- See also <https://github.com/input-output-hk/ouroboros-network/issues/1462>
   | TraceBlockFromFuture SlotNo SlotNo
 
-  -- | Leadership check failed: the tip of the ImmDB inhabits a slot /after/
-  -- the current slot
+  -- | Leadership check failed: the tip of the ImmDB inhabits the current slot
   --
   -- This might happen in two cases.
   --
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
@@ -11,7 +11,7 @@ module Ouroboros.Consensus.Node.Tracers
 
 import           Control.Tracer (Tracer, nullTracer, showTracing)
 
-import           Ouroboros.Network.Block (Point, SlotNo)
+import           Ouroboros.Network.Block (BlockNo, Point, SlotNo)
 import           Ouroboros.Network.BlockFetch (FetchDecision,
                      TraceFetchClientState, TraceLabelPeer)
 import           Ouroboros.Network.TxSubmission.Inbound
@@ -110,6 +110,8 @@ showTracers tr = Tracers
 -- >          |
 -- >          +--- TraceBlockFromFuture (leadership check failed)
 -- >          |
+-- >          +--- TraceSlotIsImmutable (leadership check failed)
+-- >          |
 -- >          +--- TraceNoLedgerState (leadership check failed)
 -- >          |
 -- >          +--- TraceNoLedgerView (leadership check failed)-- >
@@ -176,6 +178,24 @@ data TraceForgeEvent blk tx
   --
   -- See also <https://github.com/input-output-hk/ouroboros-network/issues/1462>
   | TraceBlockFromFuture SlotNo SlotNo
+
+  -- | Leadership check failed: the tip of the ImmDB inhabits a slot /after/
+  -- the current slot
+  --
+  -- This might happen in two cases.
+  --
+  --  1. the clock moved backwards, on restart we ignored everything from the
+  --     VolatileDB since it's all in the future, and now the tip of the
+  --     ImmutableDB points to a block produced in the same slot we're trying
+  --     to produce a block in
+  --
+  --  2. k = 0 and we already adopted a block from another leader of the same
+  --     slot.
+  --
+  -- We record both the current slot number as well as the tip of the ImmDB.
+  --
+  -- See also <https://github.com/input-output-hk/ouroboros-network/issues/1462>
+  | TraceSlotIsImmutable SlotNo (Point blk) BlockNo
 
   -- | We did the leadership check and concluded we /are/ the leader
   --
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -504,10 +504,10 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
 newtype PRNG = PRNG ChaChaDRG
   deriving NoUnexpectedThunks via UseIsNormalForm PRNG
 
--- | Data about a block's position in the chain
+-- | Context required to forge a block
 data BlockContext blk = BlockContext
   { bcBlockNo   :: !BlockNo
-    -- ^ the block number of the block
+    -- ^ the block number of the block to be forged
   , bcPrevPoint :: !(Point blk)
     -- ^ the point of /the predecessor of/ the block
     --
@@ -546,7 +546,7 @@ mkCurrentBlockContext
      -- If the fragment is 'Empty', then this is the block number of the
      -- tip of the ImmDB.
   -> Either (TraceForgeEvent blk (GenTx blk)) (BlockContext blk)
-     -- ^ the event is records the cause of the failure
+     -- ^ the event records the cause of the failure
 mkCurrentBlockContext currentSlot c bno = case c of
     Empty p   -- thus: bno and p both refer to the tip of the ImmDB
       | pointSlot p < At currentSlot
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -25,17 +25,18 @@ module Ouroboros.Consensus.NodeKernel (
   , ProtocolM
   ) where
 
+import           Control.Exception (assert)
 import           Control.Monad
 import           Crypto.Random (ChaChaDRG)
 import           Data.Map.Strict (Map)
-import           Data.Maybe (isNothing)
+import           Data.Maybe (isJust, isNothing)
 import           Data.Word (Word16, Word32)
 
 import           Cardano.Prelude (UseIsNormalForm (..))
 import           Control.Tracer
 
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment (..),
-                     headSlot)
+                     headPoint, headSlot)
 import           Ouroboros.Network.Block
 import           Ouroboros.Network.BlockFetch
 import           Ouroboros.Network.BlockFetch.State (FetchMode (..))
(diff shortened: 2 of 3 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -359,8 +359,10 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
         -- Normally this will be the current block at the tip, but it may
         -- be the /previous/ block, if there were multiple slot leaders
         (prevPoint, prevNo) <- do
-          mPrev <- lift $ atomically $ prevPointAndBlockNo currentSlot <$>
-                     ChainDB.getCurrentChain chainDB
+          (cc, cBno) <- lift $ atomically $
+            (,) <$> ChainDB.getCurrentChain chainDB
+                <*> ChainDB.getTipBlockNo chainDB
+          let mPrev = prevPointAndBlockNo currentSlot cBno cc
```
mrBliss: What about the following:
```haskell
          mPrev <- lift $ atomically $
            prevPointAndBlockNo currentSlot
                <$> ChainDB.getTipBlockNo chainDB
                <*> ChainDB.getCurrentChain chainDB
```
Because of laziness, we won't be doing the extra work in the transaction.

edsko:
> There must have been a maximal rollback (so that we have an empty fragment not anchored at genesis) before we try to forge a block. Nice find! +1

Hold on a second. Does this mean switch-to-fork as rollback-followed-by-foll-forward is observable, and we might produce a block after the rollback but before the roll-forward? That can't happen, can it? We just add those blocks to the chain DB, and it picks a different chain. Why would block forging be able to observe a rollback (as opposed to a switch-to-fork)?

mrBliss:
> > There must have been a maximal rollback (so that we have an empty fragment not anchored at genesis) before we try to forge a block. Nice find! +1
> 
> Hold on a second. Does this mean switch-to-fork as rollback-followed-by-foll-forward is observable, and we might produce a block after the rollback but before the roll-forward? That can't happen, can it? We just add those blocks to the chain DB, and it picks a different chain. Why would block forging be able to observe a rollback (as opposed to a switch-to-fork)?

Hmm, yeah, that would be very strange :thinking:. @nfrisby Can you give some more details about the scenario in which this happens?

nfrisby:
> Hmm, yeah, that would be very strange 🤔. @nfrisby Can you give some more details about the scenario in which this happens?

I'll dig in a bit and report back. As of yet, all I retained was `k = 1`.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -538,11 +541,22 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
...
+      :: Point (Header blk)
+         -- ^ point of predecessor of block @b@
+      -> BlockNo
+         -- ^ block number of @b@ (NOTE not of the predecessor)
+      -> (Point blk, BlockNo)
+         -- ^ point and block number of predecessor of @b@
+    mkPredecessor prevP bno
+        -- There is no block before it, so use genesis.
+      | 0 == bno  = (genesisPoint, genesisBlockNo)
+      | otherwise = (castPoint prevP, pred bno)
```
nfrisby: Gah EBBs!

If @b@ (as defined by the Haddock on `mkPredecessor`) is an EBB, then `pred bno` is wrong; it should be `bno` in that case.

nfrisby:
> So I think the only adjustment would be to change the outer `Empty` alternative to assert that the anchor is genesis. Agreed? (I suppose that's the one Edsko had in mind when he raised his objection.)

Hmm. Could there be a scenario where the ChainDB was truncated by exactly `k` blocks? Or would that cause a restart/reopen/etc such that the `getCurrentChain` `AnchorFragment` would still be non-`Empty`?

mrBliss:
> > So I think the only adjustment would be to change the outer `Empty` alternative to assert that the anchor is genesis. Agreed? (I suppose that's the one Edsko had in mind when he raised his objection.)
> 
> Hmm. Could there be a scenario where the ChainDB was truncated by exactly `k` blocks? Or would that cause a restart/reopen/etc such that the `getCurrentChain` `AnchorFragment` would still be non-`Empty`?

The chain of the ChainDB will never shrink while running, only when starting up. For example, when the whole VolatileDB folder is missing, the ChainDB can start with an empty fragment that is anchored at a point that is not genesis.

mrBliss:
I think this can be merged as-is or with [my suggestion](https://github.com/input-output-hk/ouroboros-network/pull/1544/files#r373366607).

edsko:
Ok, so, this can happen without EBBs or node restarts. Let `k = 1`. This means that under normal circumstances our fragment will be 1 block long. Now if another node manages to produce a block and send it to us before we manage to produce our own, we strip off the one block from our fragment and are left with an empty fragment. We should make sure that this scenario can occur in tests (#1569).

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -498,10 +500,11 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
     -- another node was also elected leader and managed to produce a block
     -- before us, the header right before the one at the tip of the chain.
     prevPointAndBlockNo :: SlotNo
+                        -> BlockNo
                         -> AnchoredFragment (Header blk)
                         -> Either SlotNo (Point blk, BlockNo)
-    prevPointAndBlockNo slot c = case c of
-        Empty _   -> Right (genesisPoint, genesisBlockNo)
+    prevPointAndBlockNo slot bno c = case c of
+        Empty p   -> Right $ mkPredecessor p bno
```
mrBliss:
```suggestion
        Empty p   -> Right (castPoint p, succ bno)
```

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -538,11 +541,22 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
...
                -- the same slot.
              -> Right (headerPoint hdr, blockNo hdr)
-             | _ :> hdr' <- c'
-             -> Right (headerPoint hdr', blockNo hdr')
              | otherwise
-               -- If there is no block before it, so use genesis.
-             -> Right (genesisPoint, genesisBlockNo)
+             -> Right $ case c' of
+               _ :> hdr' -> (headerPoint hdr', blockNo hdr')
+               Empty p   -> mkPredecessor p (blockNo hdr)
```
mrBliss:
```suggestion
               Empty p   -> (castPoint p, blockNo hdr)
```
And we won't even need `mkPredecessor`

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -498,10 +500,11 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
     -- another node was also elected leader and managed to produce a block
     -- before us, the header right before the one at the tip of the chain.
     prevPointAndBlockNo :: SlotNo
+                        -> BlockNo
                         -> AnchoredFragment (Header blk)
                         -> Either SlotNo (Point blk, BlockNo)
```
mrBliss: Let's change this `BlockNo` into the block number of the block *we're going to produce*, not the block number of the predecessor, because *there might not be a predecessor*. The docstring needs to be updated and the name of this function too.

edsko:
So discussed this in detail with @mrBliss . We left concrete suggestions, but here's what could go wrong with the PR as it stands:

1. If the fragment is empty (say, due to data corruption and an empty volatile DB), then `bno` will be the tip of the immutable DB. That `bno` should then _be_ the block number of the predecessor; we should try to take _its_ predecessor (which is what currently happens).

2. If the fragment is empty after stripping off a block, and we reach genesis, `mkPredecessor` must produce a block number for the predecessor, _but there is none_, and so it has little choice but to return `genesisBlockNo`. However, that is wrong: when we then call `succ` later, we will have _skipped_ a block number.

I think the suggestions we made above will address both of these issues.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
edsko: Yes, let's do this.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
mrBliss: Instead of returning the slot from the future as `Left`, return the actual error, since we can fail not only because of a slot from the future, but also when there's already a block in the slot we're trying to produce a block for, and, moreover, that block is already immutable. This might happen in two cases: 1) the clock moved backwards, on restart we ignored everything from the VolatileDB since it's all in the future, and now the tip of the ImmutableDB (= anchor point of our empty chain fragment) points to a block produced in the same slot we're trying to produce a block in; 2) k = 0 and we already adopted a block from another leader of the same slot.

So introduce an actual error type for these two cases (use the above as part of its docstring) and add a new trace message for the extra case.

mrBliss:
I have created #1570 for tracking the bugs.

nfrisby:
I rewrote it and rebased onto origin/master. The diff is big, but I think it's clearer this way.

I am seeing one failure in RealPBFT when I cherry-pick this onto my omnibus PR for Issue 1489. I'm not sure if it's due to this, but it might be. On the other hand, BFT, LeaderSchedule, PBFT, and DualPBFT all passed 40000 tests. The error involves `ChainValidationDelegationSchedulingError`, so there's plenty of other moving parts besides this PR that could be responsible for it.

I unfortunately need to prepare for the meeting in the morning, so I'm stopping on this for now.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Tracers.hs
```diff
@@ -177,6 +179,24 @@ data TraceForgeEvent blk tx
   -- See also <https://github.com/input-output-hk/ouroboros-network/issues/1462>
   | TraceBlockFromFuture SlotNo SlotNo
 
+  -- | Leadership check failed: the tip of the ImmDB inhabits a slot /after/
```
mrBliss: Not "after the current slot" but "the current slot". The ImmutableDB truncates blocks from the future, but not from the present (= the current slot), see #1563.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -553,6 +504,95 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
 newtype PRNG = PRNG ChaChaDRG
   deriving NoUnexpectedThunks via UseIsNormalForm PRNG
 
+-- | Data about a block's position in the chain
```
mrBliss: Mention that it's *about a block to be forged*, not an existing block. Otherwise the `succ` on L525 is a bit confusing.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -553,6 +504,95 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
...
+     -- ^ the current chain fragment
+     --
+     -- Recall that the anchor point is the tip of the ImmDB.
+  -> BlockNo
+     -- ^ the block number of the tip of the ChainDB
+     --
+     -- If the fragment is 'Empty', then this is the block number of the
+     -- tip of the ImmDB.
+  -> Either (TraceForgeEvent blk (GenTx blk)) (BlockContext blk)
+     -- ^ the event is records the cause of the failure
```
mrBliss:
```suggestion
     -- ^ the event records the cause of the failure
```

edsko:
I'll take care of these minor changes.

Issue #1390, opened 2020-01-03
Consensus test: run each node in a separate thread


mrBliss:
While working on #1389, I noticed that all the nodes were sharing a single `ResourceRegistry`. Also, the ChainDBs were not opened in a bracketed fashion. This causes some issues with cleanly shutting down a node and accessing a registry from a different thread.

What we should do is:
* Spawn a thread for each node. In that thread:
* Create a `ResourceRegistry` (using `withRegistry`) that is used to allocate all the resources/threads that node needs.
* Create `BlockchainTime`. This means each node will have its own `BlockchainTime`, so to terminate the network simulation: call `testBlockchainTimeDone` on all `BlockchainTime`s.
* Open a ChainDB using `withDB`.
* How mini-protocol threads are started will have to change a bit: it will be more unidirectional than the current bidirectional approach.
* At the end of the simulation, each node is shut down by killing each node's main thread.

**Important:** this will more closely resemble how a real node is set up and torn down.

Maybe this will bring up some issues with proper termination, possibly the issue we're seeing in https://github.com/input-output-hk/ouroboros-network/issues/1350.

nfrisby:
I'd like to merge the PR I have for Issue 1202 (re-delegating) before working on this one. My commits there already address some of this. They retain the "bidirectional" edge logic.

I will spend some time trying to summarize the pros and cons of the unidirectional vs bidirectional edges. For simplicity of the `test-consensus` code, I currently think the bidirectional wins out (especially once nodes can restart during the simulation, as required by Issue 1202). But favoring the unidirectional approach's resemblance of the real node could very well outweigh that.

nfrisby:
The following `git grep` is an inventory of sorts of things that use the `ResourceRegistry` in `ouroboros-consensus:lib`. To create it:

  * I added an `_RR_USE` suffix to the `ResourceRegistry` functions that add things (allocate and fork).
  * I added `_RR_USE1` to functions that are light wrappers around `_RR_USE` functions.
  * I added `_RR_THREAD` to functions that use `_RR_USE` or `_RR_USE1` functions that fork a thread and to functions that use any `_RR_THREAD` function (so this one is a transitive closure).

The following query just focuses on the first two, not the transitive closure. These are all of the low-level `ResourceRegistry` uses in `ouroboros-consensus:lib`.

```
$ git lg -3
[...]

PR #1547, opened 2020-01-31
ImmutableDB.Iterator: allocate handles in the ResourceRegistry

### LINKED ISSUE #1543: ImmutableDB is leaking file handles
On master (commit 62d8e669), this test case fails with `ImmutableDB is leaking file handles`.

```
          let ncn5 = NumCoreNodes 5 in
          prop_simple_real_pbft_convergence NoEBBs (SecurityParam 2) TestConfig
            { numCoreNodes = ncn5
            -- Still fails if I increase numSlots.
            , numSlots     = NumSlots 54
            , nodeJoinPlan = NodeJoinPlan $ Map.fromList
              [(CoreNodeId 0,SlotNo {unSlotNo = 0})
              ,(CoreNodeId 1,SlotNo {unSlotNo = 0})
              ,(CoreNodeId 2,SlotNo {unSlotNo = 0})
              ,(CoreNodeId 3,SlotNo {unSlotNo = 53}),(CoreNodeId 4,SlotNo {unSlotNo = 53})
              ]
              -- Passes if I drop either of these restarts.
            , nodeRestarts = NodeRestarts (Map.fromList [(SlotNo {unSlotNo = 50},Map.fromList [(CoreNodeId 0,NodeRestart)]),(SlotNo {unSlotNo = 53},Map.fromList [(CoreNodeId 3,NodeRestart)])])
            , nodeTopology = meshNodeTopology ncn5
              -- Slot length of 19s passes, and 21s also fails; I haven't seen this matter before.
            , slotLengths  = singletonSlotLengths (slotLengthFromSec 20)
            , initSeed     = Seed {getSeed = (15062108706768000853,6202101653126031470,15211681930891010376,1718914402782239589,12639712845887620121)}
            }
```

With some instrumentation, I determined that node c0 leaks one ImmDB file handle, for path @00000.epoch@, read only, offset at 0. It was opened by a `ChainSyncServer` thread.

The test case seems somewhat fragile, since the 'slotLengths' value seems to matter.

mrBliss:
Fixes #1543.

The ChainDB.Iterators and ChainDB.Readers use ImmutableDB.Iterators internally. Both take a ResourceRegistry to allocate the ImmutableDB.Iterator in so that the ImmutableDB.Iterator and the open handle it contains is closed when an exception occurs.

If an exception occurs at the wrong time while opening an ImmutableDB.Iterator, after opening the handle, we might leak the handle. Fix this by directly allocating the handle in the ResourceRegistry instead of the ImmutableDB.Iterator that refers to the handle.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Iterator.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Iterator.hs
@@ -35,13 +35,14 @@ import           GHC.Stack (HasCallStack)
 
 import           Ouroboros.Consensus.Block (IsEBB (..))
 import           Ouroboros.Consensus.Util.IOLike
+import           Ouroboros.Consensus.Util.ResourceRegistry (ResourceKey,
+                     ResourceRegistry, allocate, release, unsafeRelease)
 
 import           Ouroboros.Storage.Common
 import           Ouroboros.Storage.EpochInfo
 import           Ouroboros.Storage.FS.API
 import           Ouroboros.Storage.FS.API.Types
 import           Ouroboros.Storage.FS.CRC
-import           Ouroboros.Storage.Util.ErrorHandling (ErrorHandling)
 
 import           Ouroboros.Storage.ImmutableDB.API
 import           Ouroboros.Storage.ImmutableDB.Impl.Index (Index)
@@ -65,7 +66,7 @@ data IteratorHandle hash m = forall h. IteratorHandle
     -- ^ Bundled HasFS instance allows to hide type parameters
   , itIndex   :: !(Index m hash h)
     -- ^ Bundled Index instance allows to hide type parameters
-  , itState   :: !(StrictTVar m (IteratorStateOrExhausted hash h))
+  , itState   :: !(StrictTVar m (IteratorStateOrExhausted hash m h))
     -- ^ The state of the iterator. If it is 'Nothing', the iterator is
     -- exhausted and/or closed.
   , itEnd     :: !EpochSlot
@@ -74,8 +75,8 @@ data IteratorHandle hash m = forall h. IteratorHandle
     -- ^ The @hash@ of the last block the iterator should return.
   }
 
-data IteratorStateOrExhausted hash h =
-    IteratorStateOpen !(IteratorState hash h)
+data IteratorStateOrExhausted hash m h =
+    IteratorStateOpen !(IteratorState hash m h)
   | IteratorStateExhausted
   deriving (Generic, NoUnexpectedThunks)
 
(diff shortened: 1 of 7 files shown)


ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Iterator.hs
```diff
@@ -503,24 +517,27 @@ iteratorCloseImpl
...
         atomically $ writeTVar itState IteratorStateExhausted
-        hClose itEpochHandle
-  where
-    HasFS { hClose } = itHasFS
+        -- NOTE: we must use 'unsafeRelease', as we might be closing the
+        -- iterator from a different thread than the one in which it was
+        -- opened. For example, when closing the ImmutableDB/ChainDB from the
+        -- main thread, we'll close all open iterators, each of them opened in
+        -- a different thread.
+        unsafeRelease itEpochKey
```
mrBliss: Comment is wrong (same for other `unsafeRelease`): the reason this was needed is because in the consensus tests, the protocol threads are spawned in a different `ResourceRegistry` than the ChainDB, see #1390. We're not going to fix that in this PR, but add a comment: "this is dangerous, because this allows the iterator to be used by a thread not spawned by this registry, so the iterator can disappear before the thread is terminated".

mrBliss: Add a comment on #1390 about this.

nfrisby: I'm currently assuming that `iteratorClose` is sometimes called from a different thread than the one that created that registry. E.G. at a glance, it looks like `ChainDB.Impl.implIteratorNext` could be called from a spawned thread, and that calls `ImmDB.iteratorClose`.

So, even if iterators were managed by the ChainDB registry, wouldn't we see `ResourceRegistryUsedFromUnknownThread` without `unsafeRelease`? Or is my assumption wrong?

mrBliss: My comment (in the code) was wrong.

A resource can in fact be released from another thread, *as long as that thread has been spawned by the same registry*.

nfrisby: Ah ha! I was conflating the two error conditions in just the same way as the comment.

The name `ResourceRegistryUsedFromUnknownThread` is not wrong, but it does admit this misinterpretation. `ResourceRegistryUsedFromUntrackedThread` (or just the full `ResourceRegistryUsedFromThreadNotInRegistry`) would be clearer in this aspect, in hindsight.

mrBliss: > The name `ResourceRegistryUsedFromUnknownThread` is not wrong, but it does admit this misinterpretation. `ResourceRegistryUsedFromUntrackedThread` (or just the full `ResourceRegistryUsedFromThreadNotInRegistry`) would be clearer in this aspect, in hindsight.

Agreed. Good idea!

mrBliss: I have renamed `ResourceRegistryUsedFromUnknownThread` to `ResourceRegistryUsedFromUntrackedThread` (and updated the comments) in a separate commit.

PR #1552, opened 2020-01-31
windows vectored io for sockets

coot:
Currently this PR is opened to merge into `coot/windows-api`, but after #1516 is merged it will be re-targeted to `master` - it's just to make the diff adequate.

This PR provides vectored IO for sockets, vectored IO for file handles will be done at a later stage.
It also includes documentation overview, which will be useful for other teams (`cardano-wallet` or `cardano-launcher`).

### DIFF
--- a/Win32-network/src/System/Win32/Async/Socket/ByteString/Lazy.hs
+++ b/Win32-network/src/System/Win32/Async/Socket/ByteString/Lazy.hs
@@ -34,6 +34,10 @@ import qualified System.Win32.Async.Socket.ByteString as Socket.ByteString
 -- transmit at most 1024 chunks, it is safe to transmit less than 4194304
 -- bytes.
 --
+-- The socket must be in a connected state, and must be associated with an IO
+-- completion port via
+-- 'System.Win32.Async.IOManager.associateWithIOCompletionProt'.
+--
 send :: Socket
      -> ByteString
      -> IO Int64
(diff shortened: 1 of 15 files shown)


Win32-network/src/System/Win32/Async/Socket/ByteString/Lazy.hs
```diff
@@ -0,0 +1,77 @@
...
+              -> Ptr WSABuf  -- ^ lpBuffers
+              -> DWORD       -- ^ dwBufferCount
+              -> StablePtr b
+              -> IO ()
+
+
+sendAll :: Socket
+        -> ByteString
+        -> IO ()
+sendAll sock bs = do
```
karknu: Network.Socket.Lazy.sendAll handles "sending" an empty bytestring? It is not clear to me that your version does.

Win32-network/src/System/Win32/Async/WSABuf.hsc
```diff
@@ -0,0 +1,33 @@
...
+import Data.Word (Word8)
+import Foreign.Ptr (Ptr)
+import Foreign.Storable (Storable (..))
+import System.Win32.Types (ULONG)
+
+
+-- | 'WSABuf' is Haskell representation of 'WSABUF' struct.
+--
+data WSABuf = WSABuf {
+    len :: ULONG,
```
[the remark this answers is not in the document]
coot: It doesn't matter, we only use it to put in a momory address and send to an FFI, so it will be promptly evaluated. We never read thm from memory.

Win32-network/src/System/Win32/Async/Socket/ByteString/Lazy.hs
coot: good catch, I'll fix this.

coot: Done.

PR #1564, opened 2020-02-03
Add optional expiry slot to Mock transactions

### LINKED ISSUE #1297: Test that we use the right slot number for transaction validation throughout
The edge case we want to test: a transaction is valid when added to the mempool, but has become invalid by the time that the node gets around to producing the block.

(If the mempool would use the wrong slot number to validate it, it wouldn't notice this, and actually produce an invalid block.)

The easiest way to test this is to add a TTL to a transaction.

Note that this is *not* a Mempool test, but a protocol test, as it requires  the Mempool, block forging, and a ticking blockchain time.

This would exercise the chain ticking stuff introduced in #1295.

nfrisby:
Fixes #1297.

Follow-up: `prop_general` is accumulating several block/protocol specific arguments; they probably deserve their own record type/class soon.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block.hs
+++ b/ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block.hs
@@ -223,14 +223,14 @@ instance (SimpleCrypto c, Typeable ext) => HasHeader (SimpleBlock c ext) where
 instance (SimpleCrypto c, Typeable ext) => StandardHash (SimpleBlock c ext)
 
 {-------------------------------------------------------------------------------
-  HasTxs instance
+  HasMockTxs instance
 -------------------------------------------------------------------------------}
 
-instance Mock.HasTxs (SimpleBlock' c ext ext') where
-  getTxs = Mock.getTxs . simpleBody
+instance Mock.HasMockTxs (SimpleBlock' c ext ext') where
+  getMockTxs = Mock.getMockTxs . simpleBody
 
-instance Mock.HasTxs SimpleBody where
-  getTxs = simpleTxs
+instance Mock.HasMockTxs SimpleBody where
+  getMockTxs = simpleTxs
 
 {-------------------------------------------------------------------------------
   Envelope validation
@@ -320,7 +320,7 @@ updateSimpleLedgerState :: (SimpleCrypto c, Typeable ext)
 updateSimpleLedgerState b (SimpleLedgerState st) =
     SimpleLedgerState <$> updateMockState b st
 
-updateSimpleUTxO :: Mock.HasTxs a
+updateSimpleUTxO :: Mock.HasMockTxs a
                  => a
                  -> TickedLedgerState (SimpleBlock c ext)
                  -> Except (MockError (SimpleBlock c ext))
@@ -364,8 +364,8 @@ instance (Typeable p, Typeable c) => NoUnexpectedThunks (GenTx (SimpleBlock p c)
 instance HasTxs (SimpleBlock c ext) where
   extractTxs = map mkSimpleGenTx . simpleTxs . simpleBody
 
-instance Mock.HasTxs (GenTx (SimpleBlock p c)) where
-  getTxs = Mock.getTxs . simpleGenTx
+instance Mock.HasMockTxs (GenTx (SimpleBlock p c)) where
+  getMockTxs = Mock.getMockTxs . simpleGenTx
 
 instance Condense (GenTx (SimpleBlock p c)) where
     condense = condense . simpleGenTx
--- a/ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block.hs
+++ b/ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block.hs
@@ -223,20 +223,14 @@ instance (SimpleCrypto c, Typeable ext) => HasHeader (SimpleBlock c ext) where
 instance (SimpleCrypto c, Typeable ext) => StandardHash (SimpleBlock c ext)
 
 {-------------------------------------------------------------------------------
-  HasUTxO instance
+  HasTxs instance
 -------------------------------------------------------------------------------}
 
-instance Mock.HasUtxo (SimpleBlock' c ext ext') where
-  txIns      = Mock.txIns      . simpleBody
-  txOuts     = Mock.txOuts     . simpleBody
-  confirmed  = Mock.confirmed  . simpleBody
-  updateUtxo = Mock.updateUtxo . simpleBody
+instance Mock.HasTxs (SimpleBlock' c ext ext') where
+  getTxs = Mock.getTxs . simpleBody
 
-instance Mock.HasUtxo SimpleBody where
-  txIns      = Mock.txIns      . simpleTxs
-  txOuts     = Mock.txOuts     . simpleTxs
-  confirmed  = Mock.confirmed  . simpleTxs
-  updateUtxo = Mock.updateUtxo . simpleTxs
+instance Mock.HasTxs SimpleBody where
+  getTxs = simpleTxs
 
 {-------------------------------------------------------------------------------
   Envelope validation
@@ -326,13 +320,14 @@ updateSimpleLedgerState :: (SimpleCrypto c, Typeable ext)
 updateSimpleLedgerState b (SimpleLedgerState st) =
     SimpleLedgerState <$> updateMockState b st
 
-updateSimpleUTxO :: Mock.HasUtxo a
+updateSimpleUTxO :: Mock.HasTxs a
                  => a
                  -> TickedLedgerState (SimpleBlock c ext)
                  -> Except (MockError (SimpleBlock c ext))
                            (TickedLedgerState (SimpleBlock c ext))
-updateSimpleUTxO b (TickedLedgerState slot (SimpleLedgerState st)) =
-    TickedLedgerState slot . SimpleLedgerState <$> updateMockUTxO b st
+updateSimpleUTxO x (TickedLedgerState slot (SimpleLedgerState st)) =
+    TickedLedgerState slot . SimpleLedgerState <$>
+      updateMockUTxO slot x st
 
 genesisSimpleLedgerState :: AddrDist -> LedgerState (SimpleBlock c ext)
 genesisSimpleLedgerState = SimpleLedgerState . genesisMockState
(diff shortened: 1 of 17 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Mock/UTxO.hs
```diff
@@ -50,22 +52,31 @@ import           Ouroboros.Consensus.Ledger.Mock.Address
   Basic definitions
 -------------------------------------------------------------------------------}
 
-data Tx = UnsafeTx (Set TxIn) [TxOut]
+data Expiry
+  = DoNotExpire
+  | ExpireAtOnsetOf !SlotNo
+  deriving stock    (Show, Eq, Ord, Generic)
+  deriving anyclass (Serialise, NFData, NoUnexpectedThunks)
```
mrBliss: Why is `NFData` (and that orphan) needed?

ouroboros-consensus/test-consensus/Test/ThreadNet/General.hs
```diff
@@ -594,3 +598,18 @@ prop_general countTxs k TestConfig{numSlots, nodeJoinPlan, nodeRestarts, nodeTop
         average :: [Double] -> Double
         average [] = 0
         average xs = sum xs / fromIntegral (length xs)
+
+    -- The 'prop_valid_block' argument could, for example, check for no expired
+    -- transactions.
+    --
+    -- TODO apply to all 'TraceForgeAdopted' blocks, not just those in the
+    -- final chain?
```
mrBliss: Is that possible? Maybe by (ab?)using the tracer you have for `TraceForgeAdopted`?

nfrisby: Yeah, I was thinking via tracer. It doesn't seem like an abuse, I don't think: the tracer could just accumulate the adopted blocks in a set and check them at the end.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Mock/UTxO.hs
nfrisby: `Tx` derives it, and `Expiry` occurs in `Tx`. I didn't question why `Tx` derives it.

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/TxGen/Mock.hs
```diff
@@ -40,8 +42,12 @@ genSimpleTx addrs u = do
         outs         = if amount == fortune
             then [outRecipient]
             else [outRecipient, (sender, fortune - amount)]
-    return $ Tx ins outs
+    expiry <- (mkExpiry . SlotNo . fromIntegral) <$> generateBetween 2 10
```
nfrisby: This expiry doesn't seem to affect anything else, so 2-10 seems reasonable.

Should we also include no expiry here? If I recall correctly, I only introduced the `DoNotExpire` ctor because I wasn't sure how to update the mempool tests, so I side-stepped it that way: expiry in `ThreadNet`, no expiry in those other tests.

@mrBliss Please advise.

mrBliss: I don't think it matters that much. I think I would include `DoNotExpire` here too, just to cover everything.

Should the expiry range be dependent on `k`? Although they are unrelated, it might make a difference when we're testing with network partitions and such? I suppose we can keep it [2, 10] now and update it when/if needed.

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block.hs
```diff
@@ -330,8 +331,22 @@ instance MockProtocolSpecific c ext => ApplyTx (SimpleBlock c ext) where
 
   type ApplyTxErr (SimpleBlock c ext) = MockError (SimpleBlock c ext)
 
-  applyTx   = const updateSimpleUTxO
-  reapplyTx = const updateSimpleUTxO
+  applyTx           = reapplyTx
+  reapplyTx _ tx tl =
+      case hasExpired of
```
mrBliss: I see CI is failing with a test related to tx/block validation. I believe you also have to take transaction expiration into account when validating blocks (`applyLedgerBlock`). You can move the expiration logic to some common place.

edsko: It shouldn't live in `applyTx` _or_ `applyLedgerBlock`, but rather in the ledger rules.

mrBliss: See Lemma 3 in #1765 :slightly_smiling_face:

nfrisby:
The latest CI failure was due to the unrelated #1779.

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block.hs
nfrisby: I pushed it into the ledger rules. I introduced the `Ticked` type used in a few modules; should it be renamed and/or located somewhere more general? I think of it (as commented) as a generalization of `TickedLedgerState`. @edsko @mrBliss

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/TxGen/Mock.hs
nfrisby: Done, thanks.

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/BFT.hs
```diff
@@ -4,19 +4,22 @@ module Test.ThreadNet.BFT (
...
+import qualified Data.Map.Strict as Map
+
 import           Test.QuickCheck
 import           Test.Tasty
 import           Test.Tasty.QuickCheck
 
 import           Cardano.Slotting.Slot
 
+import           Ouroboros.Network.Block (SlotNo (..))
+
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

nfrisby:
I think this is ready for merge, unless you'd like to see changes to the following.

  * the way I pushed the logic further into the ledger rule code
  * the `Ticked` type I used to do so

If not, I think this is ready for `bors`.

nfrisby:
@edsko and I had a call. He pointed out a slightly higher-level location in which to check for expiry. Moving it there let me remove `Ticked`.

In order to relocate the check, I also simplified `HasUtxo` to `HasTxs` -- I need to access the actual `Tx`s in order to check them for expiry. Is this too much of a concretization? I suppose I could add something about expiry to `HasUtxo`, but this simplification seems fine in the mock ledger code.

nfrisby:
I just now rebased onto `origin/master`.

nfrisby:
I just now rebased onto `origin/bors/staging`, to address the conflicts from the new `PropGeneralArgs`.

nfrisby:
I rebased on `origin/master`.

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/BFT.hs
```diff
@@ -44,11 +46,30 @@ tests = testGroup "BFT"
...
+          , initSeed     = Seed (6358650144370660550,17563794202468751585,17692838336641672274,12649320068211251815,18441126729279419067)
+          }
+    , testProperty "simple convergence" $ \tc ->
+        -- TODO k > 1 as a workaround for Issue #1511.
+        --
         forAll (SecurityParam <$> elements [2 .. 10]) $ \k ->
         prop_simple_bft_convergence k tc
     ]
+  where
+    defaultSlotLengths = singletonSlotLengths $ slotLengthFromSec 1
```
mrBliss: Does the repro depend on the exact slot length?

I'm inclined to use 20s here, because that's the default slot length for BFT.

Also, you could use `defaultSlotLengths` in the first test case too.

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/UTxO.hs
```diff
@@ -89,61 +104,63 @@ data UtxoError
...
-  txOuts     = foldr (Map.union . txOuts)    Map.empty
-  confirmed  = foldr (Set.union . confirmed) Set.empty
-  updateUtxo = repeatedlyM updateUtxo
-
-instance HasUtxo a => HasUtxo (Chain a) where
-  txIns      = txIns      . toOldestFirst
-  txOuts     = txOuts     . toOldestFirst
-  updateUtxo = updateUtxo . toOldestFirst
-  confirmed  = confirmed  . toOldestFirst
+class HasTxs a where
```
mrBliss: There already exists such a class: https://github.com/input-output-hk/ouroboros-network/blob/cca25adb2cc393177286feaba4c503a033343c36/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs#L93

Either we reuse that class (maybe generalise it to `a` instead of `blk`), and if that's not possible (`Tx` vs `GenTx`), rename this new class to avoid confusion.

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/State.hs
```diff
@@ -75,13 +82,32 @@ updateMockTip hdr (MockState u c t)
...
+                -> Tx
+                -> MockState blk
+                -> Except (MockError blk) (MockState blk)
+updateMockUTxO1 now tx (MockState u c t) = case hasExpired of
+    Just e  -> throwError e
+    Nothing -> do
+      u' <- withExcept MockUtxoError $ updateUtxo tx u
+      return $ MockState u' (c `Set.union` confirmed tx) t
+  where
+      Tx expiry _ins _outs       = tx
```
mrBliss:
```suggestion
      Tx expiry _ins _outs = tx
```

nfrisby: Done, thanks.

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/UTxO.hs
nfrisby: I renamed the new one; the two have slightly different semantics (going by the existing class's comment). Moreover, this new one is specific to the mock Tx -- I don't think generalizing the existing one would be trivial and/or as clear.

Done, thanks.

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/BFT.hs
nfrisby: I switched it to 20s. Done, thanks.

nfrisby:
Update: I addressed the comments and rebased onto master.

Issue #1304, opened 2019-11-27
Check block validity when copying from VolatileDB to ImmutableDB


mrBliss:
When we compute the CRC of a block written to the ImmutableDB, we should first check whether the block has not already been corrupted while in the VolatileDB, i.e., we should whether the body matches its hash in the header. If we don't do this check, we'd compute the CRC of the corrupted block and wouldn't notice the corruption of the block when reading it again from the ImmutableDB.

mrBliss:
What if we detect that a block in the VolatileDB has gotten corrupted? We can't simply delete or overwrite that block, that's not supported by the VolatileDB (and that's a deliberate decision). We can only try to garbage collect all the block by triggering a garbage collection for its slot number, but that's not guaranteed to delete the block from disk, as newer blocks can reside in the same file.

Even if we manage to delete the invalid block, what should we do next? How do we let the BlockFetch client download it again? I think this will quickly become very complex and will break many of the assumptions upon which we based our design.

I think for now the simplest solution is to **wipe the entire VolatileDB and shut down the node**. When restarting the node, we'll restart from a valid immutable chain, and redownload the missing blocks, including the corrupted block. Let's call this plan A.

We can improve this by *adding validation support to the VolatileDB*: check the integrity of each block. This will not be a checksum check like in the ImmutableDB (as we'd need to store the checksums on disk too), but the more expensive "block matches header" check + "check the signature of the header". This validation can detect incorrect blocks in files, but what should be done when encountering such an invalid block? The simplest thing would be **delete the file in question** (plan B). This is still better than wiping the entire VolatileDB, as we leave the other files in place. One step better: **truncate the file in question to the offset before the first invalid block** (plan C). If that means truncating to 0, delete the file.

If we want to save even more blocks from truncation, we could think about the following: if a file contains at least one invalid block that is followed by at least one valid block, create a new file (with a special filename), copy all valid blocks from the corrupted file to the new file. At the end, delete the corrupted file and rename the new file to the file that was corrupted (plan D). This approach avoids in-place mutation and gives us atomicity.

kderme:
Should I consider adding an api call `deleteBlock`, which implements some of the plans above? Maybe with an additional argument to choose a specific plan B,C,D?

mrBliss:
> Should I consider adding an api call `deleteBlock`, which implements some of the plans above? Maybe with an additional argument to choose a specific plan B,C,D?

[...]

PR #1568, opened 2020-02-04
Validation of the VolatileDB

kderme:
Addresses part 2 of https://github.com/input-output-hk/ouroboros-network/issues/1304#issuecomment-581289381 and more:

- there is an option to check if each parsed block is valid, while parsing.
- the db now recovers when there is a duplicated slot and ignores invalid filenames.
- added tests for this.
- added traces.

`readIncrementalOffsets` is no longer used anywhere. Worth keeping?


ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
```diff
@@ -418,10 +429,11 @@ tryCollectFile :: forall m h blockId
                -> InternalState blockId h
                -> (FileId, FileInfo blockId)
                -> m (InternalState blockId h)
-tryCollectFile hasFS@HasFS{..} env slot st@InternalState{..} (fileId, fileInfo) =
+tryCollectFile hasFS env@VolatileDBEnv{..} slot st@InternalState{..}
+  (fileId, fileInfo) =
```
mrBliss: I would revert this change (so it's one line again) and use the following instead:

```haskell
tryCollectFile ...
    ....
  where
    VolatileDBEnv { _dbErr, _tracer } = env
```

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/VolDB.hs
```diff
@@ -494,22 +514,36 @@ blockFileParser' :: forall m blk h. (IOLike m, HasHeader blk)
...
                  -> (forall s. Decoder s (Lazy.ByteString -> blk))
+                 -> (blk -> Bool)
                  -> VolDB.Parser
-                     Util.CBOR.ReadIncrementalErr
+                     (BlockFileParserError (HeaderHash blk))
                      m
                      (HeaderHash blk)
-blockFileParser' hasFS isEBB encodeBlock decodeBlock = VolDB.Parser $
-       fmap (fmap (fmap fst)) -- Drop the offset of the error
-     . Util.CBOR.readIncrementalOffsets hasFS decoder'
```
mrBliss: Yeah, if `readIncrementalOffsets` is no longer used, I would remove it. We can always resurrect it if necessary.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/VolDB.hs
```diff
@@ -494,22 +514,36 @@ blockFileParser' :: forall m blk h. (IOLike m, HasHeader blk)
...
+                    m
+                    (Maybe (Util.CBOR.ReadIncrementalErr, Word64))
+                 -> m (VolDB.ParsedInfo (HeaderHash blk),
+                    Maybe (BlockFileParserError (HeaderHash blk)))
+    checkEntries = go []
+      where
+        go parsed stream = S.next stream >>= \case
+          Left mbErr -> return (reverse parsed, BlockReadErr . fst <$> mbErr)
+          Right ((offset, (size, blk)), stream') | checkIntegrity blk ->
+            let newParsed = (offset, (VolDB.BlockSize size, extractInfo' blk))
```
mrBliss: We have to be careful here to not accidentally hold on to the block. As we're returning a tuple, we can't rely on the data type to do the forcing for us. You can do the following:
```haskell
let !blockInfo = extractInfo' blk
    newParsed  = (offset, (VolDB.BlockSize size, blockInfo))
...
```

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/VolDB.hs
```diff
@@ -494,22 +514,36 @@ blockFileParser' :: forall m blk h. (IOLike m, HasHeader blk)
...
+            in go (newParsed : parsed) stream'
+          Right ((_, (_, blk)), _) -> return (reverse parsed,
+            Just $ BlockCorruptedErr $ VolDB.bbid $ extractInfo' blk)
```
mrBliss: Same here:
```haskell
Right ((_, (_, blk)), _) ->
  let !bid = VolDB.bbid $ extractInfo' blk
  in return (reverse parsed, Just (BlockCorruptedErr bid))
```

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/VolDB.hs
```diff
@@ -475,17 +490,22 @@ getBlockComponent db blockComponent hash = withDB db $ \vol ->
...
 
 {-------------------------------------------------------------------------------
-  Auxiliary: parsing
+  Parsing
 -------------------------------------------------------------------------------}
 
+data BlockFileParserError hash =
+    BlockReadErr Util.CBOR.ReadIncrementalErr
+  | BlockCorruptedErr hash
+  deriving (Eq, Show)
```
mrBliss: Export this type too.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/VolatileDB/StateMachine.hs
```diff
@@ -381,6 +387,10 @@ preconditionImpl Model{..} (At (CmdErr cmd err)) =
         case fmap fst . snd <$> M.lookup file (index dbModel) of
           Nothing   -> Bot
           Just bids -> isOpen .&& (thHash $ testHeader tb) `elem` bids
+      PutCorrupted tb        -> isOpen .&& Boolean (and
+        [ (not $ checkBlockIntegrity tb)
+        , (not $ M.member (thHash $ testHeader tb) (mp dbModel))
+        , (slotAfterGC $ Just $ thSlotNo $ testHeader tb) ])
```
mrBliss: I would try to use [`Logic`](https://hackage.haskell.org/package/quickcheck-state-machine-0.6.0/docs/Test-StateMachine-Logic.html) as much as possible instead of `Bool`, so you get a more precise message in case it no longer holds. (Also: use `M.notMember` to avoid the extra `Not`)

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/VolatileDB/StateMachine.hs
```diff
@@ -609,6 +638,11 @@ mockImpl model cmdErr = At <$> return mockResp
   where
     (mockResp, _dbModel') = step model cmdErr
 
+checkBlockIntegrity :: TestBlock -> Bool
+checkBlockIntegrity testBlock =
+  hashHeader (testHeader testBlock) == thHash (testHeader testBlock)
+  || not (tbIsValid (testBody testBlock))
+
```
mrBliss: Isn't this the same as `Test.Ouroboros.Storage.TestBlock.testBlockIsValid`?

ouroboros-consensus/src/Ouroboros/Storage/VolatileDB/Impl.hs
kderme: This is not so easy because `VolatileDBEnv` is Existential type, but I did the same for `InternalState` to keep it a single like.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/VolDB.hs
```diff
@@ -494,22 +516,38 @@ blockFileParser' :: forall m blk h. (IOLike m, HasHeader blk)
...
+    checkEntries :: Stream (Of (Word64, (Word64, blk)))
+                    m
+                    (Maybe (Util.CBOR.ReadIncrementalErr, Word64))
+                 -> m (VolDB.ParsedInfo (HeaderHash blk),
+                    Maybe (BlockFileParserError (HeaderHash blk)))
+    checkEntries = go []
+      where
+        go parsed stream = S.next stream >>= \case
+          Left mbErr -> return (reverse parsed, BlockReadErr . fst <$> mbErr)
+          Right ((offset, (size, blk)), stream') | checkIntegrity blk ->
```
mrBliss: This check is fairly expensive and will slow down startup. Can you make the `checkIntegrity` check optional? Just like we have the two validation policies for the ImmutableDB. By default, we won't do the check, we'll only do it when started in recovery mode. The validation policy should be an argument to the VolatileDB.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/VolatileDB/StateMachine.hs
```diff
@@ -157,6 +158,7 @@ data Cmd
     | CreateFile
     | CreateInvalidFile
     | DuplicateBlock FsPath TestBlock
+    | PutCorrupted TestBlock
```
mrBliss: How is this different from `Corrupt`? To me `Corrupt` sounds more general and thus preferable.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/VolDB.hs
kderme: I missed this comment, don't review yet...

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/VolatileDB/StateMachine.hs
kderme: Almost yes, this also checkes `tbIsValid`

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -141,9 +143,12 @@ run tracers protocolTracers chainDbTracer diffusionTracers diffusionArguments
               -- When the last shutdown was not clean, validate the complete
               -- ChainDB to detect and recover from any corruptions. This will
               -- override the default value /and/ the user-customised value of
-              -- the 'ChainDB.cdbValidation' field.
+              -- the 'ChainDB.cdbValidation' and the
+              -- 'ChainDB.cdbBlockValidation' fields.
             = (customiseChainDbArgs args)
-              { ChainDB.cdbValidation = ValidateAllEpochs }
+              { ChainDB.cdbValidation      = ValidateAllEpochs
```
mrBliss:
```suggestion
              { ChainDB.cdbImmValidation      = ValidateAllEpochs
```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -141,9 +143,12 @@ run tracers protocolTracers chainDbTracer diffusionTracers diffusionArguments
...
+              , ChainDB.cdbBlockValidation = ValidateAll
```
mrBliss:
```suggestion
              , ChainDB.cdbVolValidation = ValidateAll
```

mrBliss:
This ready for merging. Thanks @kderme for the validation code. I cleaned up some bits and implemented part 1 of https://github.com/input-output-hk/ouroboros-network/issues/1304#issuecomment-581289381.

PR #1582, opened 2020-02-05
ntp-client: fixed calculations of delay interval

coot:
It should use `min` rather than `max` function on each recursive step:
```
2 * delay `min` 600
```


coot:
Already fixed in another PR

PR #1585, opened 2020-02-05
ChainSyncClient: No `BlockNo` when tip is genesis

edsko:
The `Tip` data-type was

```haskell
data Tip b = Tip
  { tipPoint   :: !(Point b)
  , tipBlockNo :: !BlockNo
  }
```

where `Point` uses `WithOrigin`. This is not correct. When `tipPoint` is `Origin`, then there _is_ no `tipBlockNo` (`genesisBlockNo` is the block number of the first block on the chain). In this commit we change this to

```haskell
data Tip b =
    -- | The tip is genesis
    TipGenesis

    -- | The tip is not genesis
  | Tip !SlotNo !(HeaderHash b) !BlockNo
```

It doesn't, however, make any real changes, providing instead some "legacy" API that pretends that we _do_ always have a `BlockNo`. This primarily affects consensus only, in networking this only appears in examples and tests.

We also use this legacy format to avoid changing the binary presentation of `Tip`, so that this PR does _not_ break backwards compatibility.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs
@@ -498,7 +498,7 @@ chainSyncClient mkPipelineDecision0 tracer cfg btime
               Nothing
               (handleNext kis mkPipelineDecision' n')
       where
-        theirTipBlockNo = tipBlockNo (unTheir theirTip)
+        theirTipBlockNo = getLegacyTipBlockNo (unTheir theirTip)
         decision = runPipelineDecision
           mkPipelineDecision
           n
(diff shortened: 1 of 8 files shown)


edsko:
Will create the ticket.

edsko:
https://github.com/input-output-hk/ouroboros-network/issues/1587 is the ticket.

avieth:
LGTM.

Used to be we had to have a block number for some reason, and we chose 0 for the genesis, and each EBB would have the same block number as the 0th block number of its epoch. If this isn't needed anymore then good riddance!

edsko:
> Used to be we had to have a block number for some reason, and we chose 0 for the genesis, and each EBB would have the same block number as the 0th block number of its epoch. If this isn't needed anymore then good riddance!

This might still be needed, but if so, it should be Byron specific, and not live in the general infrastructure. The addition to of `WithOrigin` to `Point` was already a step towards that direction. My hope is `genesisSlotNo` and `genesisBLockNo` can be banished altogether https://github.com/input-output-hk/cardano-base/issues/78.

PR #1590, opened 2020-02-06
SRE-34 haskell overhaul

craigem:
This is the overhaul of Haskell for `ouroboros-network` along the lines of `cardano-node` [PR #507](https://github.com/input-output-hk/cardano-node/pull/507/).


cabal.project
```diff
@@ -55,95 +55,109 @@ package contra-tracer
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/iohk-monitoring-framework
-  subdir:   contra-tracer
   tag: dd30455144e11efb435619383ba84ce02aee720d
+  --sha256: 1g08bg99fvss99kg27l7pmxm7lh60573xln8l8x2rzvwfvfgk2i5
```
coot: just out of curiosity, what are these commented out `sha256` fields?

default.nix
```diff
@@ -1,48 +1,42 @@
...
+# eg. to test build against local checkout of nixpkgs and iohk-nix:
+# nix build -f default.nix cardano-ledger --arg sourcesOverride '{
+#   iohk-nix = ../iohk-nix;
+# }'
+, sourcesOverride ? {}
+# pinned version of nixpkgs augmented with overlays (iohk-nix and our packages).
+, pkgs ? import ./nix { inherit system crossSystem config sourcesOverride; }
+, gitrev ? pkgs.iohkNix.commitIdFromGitRepoOrZero ./.git
+}:
+with pkgs; with commonLib;
```
coot: where can we find instructions how to build various parts with nix?
Did the build instruction changed (if that the case we'll need to update some READMEs).

ouroboros-network/ouroboros-network.cabal
```diff
@@ -27,8 +27,8 @@ Flag ipv6
   -- Default to False since travis lacks IPv6 support
   Default: False
 
-flag cddl
-  Description: Enable CDDL based tests of the CBOR encoding
+flag nocddl
+  Description: Disable CDDL based tests of the CBOR encoding
```
coot: that's a good catch :)

.buildkite/pipeline.yml
```diff
@@ -1,13 +1,27 @@
...
+    command:
+      - "rm -rf /build/ouroroboros-network"
+      - "cp -R . /build/ouroroboros-network"
+      - "cd /build/ouroroboros-network"
+      - "nix-build scripts/buildkite -o stack-rebuild"
+      - "./stack-rebuild"
     agents:
       system: x86_64-linux
-  - label: 'dependencies-in-master'
-    command: 'ci/check-dependencies-merged-in-master.sh'
```
mrBliss: Why is this gone?

cabal.project
```diff
@@ -55,95 +55,109 @@ package contra-tracer
...
+  subdir:   contra-tracer
 
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-prelude
   tag: 00487726c4bc21b4744e59d913334ebfeac7d68e
-  subdir: .
+  --sha256: 0v4fcq5kdd2r5dgwys8kv46ff33qp756n26ycxrca10wq14zkwm5
```
mrBliss: Isn't there a script to generate these checksums?

scripts/buildkite/stack-cabal_config_check.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+set -euo pipefail
+
+# This script checks that the `stack.yaml` and `cabal.project` files have
+# consistent git hashes for the packages they depend on. We use
+# `cardano-repo-tool`'s `update-cabal-project` command which modifies
+# `cabal.project` to be consistent with `stack.yaml`s versions. If the
+# diff is non-empty, we know they're out of sync.
```
coot: This description is the same as in the `scripts/buildkite/stack-cabal-sync.sh` script which suggests that one is not needed.

disassembler: we're upstreaming that to iohk-nix currently. Will push as soon as that merges.

.buildkite/pipeline.yml
disassembler: There is no .stack.nix directory anymore. We may need to see if there's a better way to check this.

cabal.project
disassembler: > Previously, we had to run `./nix/regenerate.sh` whenever our dependencies changed. What should we do now?

I've restored that script to run cabalProjectRegenerate now. It should be much faster (and less error prone). The only thing to be mindful of is order does matter. That might be improved int he future.

cabal.project
craigem: IIRC, `haskell.nix` requires them but cabal can't yet use them.

cabal.project
mrBliss: The order of the dependencies in `cabal.project`, I assume?

scripts/buildkite/rebuild.hs
```diff
@@ -13,7 +13,7 @@ import Turtle
 -- | Run build and upload coverage information when successful
 main :: IO ()
 main = do
-  buildResult <- buildStep (Just ["--scenario=ContinuousIntegration"])
```
mrBliss: Yeah, that's something from `cardano-ledger`.

cabal.project
jbgi: No, it's actually the order of sub-nodes of each `source-repository-package`, in particular, `--sha256` must currently appear before `subdir`.

mrBliss:
What's the difference between `scripts/buildkite/stack-cabal-sync.sh` and `scripts/buildkite/stack-cabal_config_check.sh`? Does the former generate the patch whereas the latter does the check? They seem to overlap quite a bit.

PR #1596, opened 2020-02-06
improved retry policy

MarcFontaine:
When an NTP query fails the client retries after 5 seconds.
On every failed retry the client doubles the restart delay,
but does not wait longer than 600s.


MarcFontaine:
https://github.com/input-output-hk/ouroboros-network/pull/1596

PR #1599, opened 2020-02-06
Mempool: better revalidation and caching of the ledger state

### LINKED ISSUE #1565: Mempool.removeTxs should not use the fast path
See https://github.com/input-output-hk/ouroboros-network/issues/1505#issuecomment-581699557 for a detailed bug report.

mrBliss:
Fixes #1565 and #1301.

* Replace `Mempool.addTxs` with `Mempool.tryAddTxs`, which doesn't block. This   function will be much easier to test. We provide an implementation of `Mempool.addTxs` in terms of `Mempool.tryAddTxs`. We simplify the implementations by relying on the background thread for synchronising the Mempool with an updated ledger state.

* Cache the `TickedLedgerState` after applying all transactions in the Mempool   to it (#1301). Previously, we had to reapply all transactions whenever we added a new transaction. The new approach is much faster.

* As we're keeping the `TickedLedgerState` in memory, make sure it is thunk-free.

* Correct revalidation in (#1564) after explicitly removing transactions from the Mempool, see `revalidateTxsFor`.

* Cleanup of conversion functions in `TxSeq`.

* O(1) `MempoolSize` calculation for a `TxSeq` using its fingertree measure.

Also includes a rewrite of the Mempool tests using the `Mock.Utxo`. Using the `Mock.Utxo` instead of the overly simplistic `TestTx` we were using allows for more complicated dependencies between transactions. This better resembles the real transactions, at the cost of some complexity. With the former transaction type, we weren't even able to trigger #1565.


ouroboros-consensus/test-consensus/Test/Consensus/Mempool.hs
```diff
@@ -58,8 +65,6 @@ tests = testGroup "Mempool"
   , testProperty "addTxs txs == mapM (addTxs . pure) txs"  prop_Mempool_addTxs_one_vs_multiple
   , testProperty "result of addTxs"                        prop_Mempool_addTxs_result
   , testProperty "Invalid transactions are never added"    prop_Mempool_InvalidTxsNeverAdded
-  , testProperty "result of getCapacity"                   prop_Mempool_getCapacity
-  , testProperty "Mempool capacity implementation"         prop_Mempool_Capacity
```
mrBliss: Draft because I temporarily removed these two tests. `prop_Mempool_Capacity` is a major headache to rewrite (I have a WIP lying around) because transactions now have dependencies. For example, removing a transaction can cause other transaction to be removed as well.

Idea: what if we add another transaction type (part of `GenTx`) that is totally independent (of the UTxO)? I bet that with such a transaction we can just restore the previous version of the test.

mrBliss: Alternative approach: expose an internal variant of `addTxs` that instead of blocking, returns the added txs and the remaining txs. This function would be much easier to test. `addTxs` should then be implemented using this function. Downside: we're not testing the actual `addTxs`.

mrBliss: I have taken the alternative approach: add a non-blocking variant of `addTxs`: `tryAddTxs` which is easily testable. This non-blocking variant is then used to implement `addTxs`.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/TxSeq.hs
```diff
@@ -218,34 +224,26 @@ splitAfterTxSizeSpec txseq n =
...
--- | Convert a 'TxSeq' to a list of 'TxTicket's.
-txTickets :: TxSeq tx -> [TxTicket tx]
-txTickets (TxSeq ftree) = Foldable.toList ftree
+toTuples :: TxSeq tx -> [(tx, TicketNo)]
+toTuples (TxSeq ftree) = fmap
+    (\ticket -> (txTicketTx ticket, txTicketNo ticket))
+    (Foldable.toList ftree)
+
+-- | \( O(1) \). Return the 'MempoolSize' of the given 'TxSeq'.
+toMempoolSize :: TxSeq tx -> MempoolSize
```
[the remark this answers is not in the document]
intricate: Ah, I see why you placed it here though. Specifically to utilize `FingerTree.measure`.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -315,15 +351,19 @@ instance Monoid MempoolSize where
 
 -- | Events traced by the Mempool.
 data TraceEventMempool blk
-  = TraceMempoolAddTxs
-      ![GenTx blk]
+  = TraceMempoolAddedTx
+      !(GenTx blk)
       -- ^ New, valid transaction were added to the Mempool.
```
intricate:
```suggestion
      -- ^ New valid transaction that was added to the Mempool.
```

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -315,15 +351,19 @@ instance Monoid MempoolSize where
...
       !MempoolSize
-      -- ^ The current size of the Mempool.
-  | TraceMempoolRejectedTxs
-      ![(GenTx blk, ApplyTxErr blk)]
+      -- ^ The size of the Mempool before adding the transaction.
+      !MempoolSize
+      -- ^ The size of the Mempool after adding the transaction.
+  | TraceMempoolRejectedTx
+      !(GenTx blk)
       -- ^ New, invalid transaction were rejected and thus not added to the
```
intricate:
```suggestion
      -- ^ New invalid transaction that was rejected and thus not added to the
```

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -180,233 +203,113 @@ forkSyncStateOnTipPointChange registry menv =
...
---   transactions.
---
---   If the sync resulted in no work being done, we don't have to worry about
---   losing any changes in the event of a 'retry'. So we continue by calling
---   'validateNew', providing the new transactions as an argument.
---
--- * In 'validateNew', we first attempt to individually validate the first
---   transaction of the provided list.
+-- This function returns two list: the transaction that were added or rejected
+-- and the transaction that could yet be added, because the Mempool capacity
```
intricate: Did you mean:

```suggestion
-- and the transaction that could not yet be added, because the Mempool capacity
```

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -180,233 +203,113 @@ forkSyncStateOnTipPointChange registry menv =
...
+          unless (null (vrNewValid vr)) $
+            -- Each time we have found a valid transaction, we update the
+            -- Mempool. This keeps our STM transactions short, avoiding
+            -- repeated work.
+            writeTVar mpEnvStateVar is'
+
+          -- We only extended the ValidationResult with a single transaction
+          -- ('firstTx'). So if it's not in 'vrInvalid', it must be in
+          -- 'vrNewValid'.
+          return $ case listToMaybe (vrInvalid vr) of
```
intricate: Do you think it'd be unnecessary to also `assert` that we somehow didn't get to this point with a `vrInvalid vr` whose length is greater than 1? 

Maybe I'm being overly cautious without good reason because this should be impossible given that you're only validating the first transaction at this point.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -180,233 +203,113 @@ forkSyncStateOnTipPointChange registry menv =
...
+      join $ atomically $ do
+        is <- readTVar mpEnvStateVar
+        let curSize = msNumBytes $ isMempoolSize is
+        if curSize + firstTxSize > capacity then
+          -- No space in the Mempool
+          return $ done acc toAdd
+        else do
+          let vr  = extendVRNew cfg firstTx $ validationResultFromIS is
+              is' = internalStateFromVR vr
+          unless (null (vrNewValid vr)) $
```
mrBliss: Even when the tx is invalid, we could still write it, because we would have that `is == is'`. Here we just avoid an extra write.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -554,41 +447,41 @@ data ValidationResult blk = ValidationResult {
...
+    , vrAfter        = isLedgerState
     , vrInvalid      = []
-    , vrLastTicketNo = lastTicketNo
+    , vrLastTicketNo = isLastTicketNo
     }
   where
-    afterKnownValid :: [GenTx blk]
-                    -> TickedLedgerState blk -> TickedLedgerState blk
-    afterKnownValid []       = id
-    afterKnownValid (tx:txs) = afterKnownValid txs . reapplyTxSameState cfg tx
```
mrBliss: We no longer need `reapplyTxSameState`, remove it.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
intricate: With this new implementation, it looks like `vrNewValid` doesn't need to be a list since we're recursively validating the head transaction against a newly constructed `ValidationResult` that _always_ has a `vrNewValid` value of `[]`. Perhaps we should just define the `vrNewValid` field as a `Maybe`?

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -232,6 +227,47 @@ data Mempool m blk idx = Mempool {
...
+-- | Wrapper around 'implTryAddTxs' that blocks until all transaction have
+-- either been added to the Mempool or rejected.
+--
+-- This function does not sync the Mempool contents with the ledger state in
+-- case the latter changes, it relies on the background thread to do that.
+--
+-- POSTCONDITON:
+-- > processed <- addTxs mpEnv txs
+-- > map fst processed == txs
+addTxs
```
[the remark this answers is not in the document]
mrBliss: I think it is nice that `addTxs` is general enough that it can work for any Mempool. Now I can't accidentally rely on implementation details.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/TxSeq.hs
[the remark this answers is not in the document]
mrBliss: Yeah, I agree that importing `MempoolSize` is not the cleanest solution, but I think the alternative would be worse in different ways.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
mrBliss: Sure, I'll add it.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
mrBliss: Ah, good point! I thought that `revalidateTxsFor` still needed it, but I was wrong, it uses `vrValid` (through `extendVRPrevApplied`).

Although, thinking about it now, what should `extendVRNew` do when `vrNewValid` is a `Just`? Return an `error`, I suppose?

intricate: Yeah, I think I would probably return an `error` or `assert` that it is `Nothing`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Mempool.hs
```diff
@@ -106,12 +106,6 @@ instance ApplyTx ByronBlock where
     where
       validationMode = CC.ValidationMode CC.NoBlockValidation Utxo.TxValidationNoCrypto
 
-  reapplyTxSameState cfg tx st =
-      validationErrorImpossible $
-        applyByronGenTx validationMode cfg tx st
-    where
-      validationMode = CC.ValidationMode CC.NoBlockValidation Utxo.NoTxValidation
```
intricate: Does this mean that we no longer need the `NoTxValidation` mode in `cardano-ledger`?

ouroboros-consensus/test-util/Test/Util/QuickCheck.hs
```diff
@@ -21,3 +26,15 @@ x `ge` y = counterexample (show x ++ " < " ++ show y) $ x >= y
...
+{-------------------------------------------------------------------------------
+  Improved variants
+-------------------------------------------------------------------------------}
+
+-- | Generates one of the given values. The input list must be non-empty.
+--
+-- NOTE unlike the standard @elements@, this variant has a 'HasCallStack'
+-- constraint, which makes debugging the 'error' much easier.
+elements :: HasCallStack => [a] -> Gen a
+elements [] = error "QuickCheck.elements used with empty list"
```
intricate: Perhaps an irrelevant nitpick but, because this could be confused with the actual module from the `quickcheck` package, I'd be inclined to be more explicit here:

```suggestion
elements [] = error "Test.Util.QuickCheck.elements used with empty list"
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Byron/Mempool.hs
mrBliss: Indeed not. @edsko and I have been looking for a ticket about this (that we could close), but we couldn't find it. Do you know which one that would be?

intricate: I'm not sure that there has been a ticket in ledger that specifically speaks to removing `NoTxValidation`.

PR #1600, opened 2020-02-07
Add ProtocolTracers parameter to Ouroboros.Consensus.Node.run

intricate:
Requirement for https://github.com/input-output-hk/cardano-node/issues/527.

Specifically:

>We also need to trace at the default log level for the outbound side of tx submission:
>* when a remote peer has requested a tx id, and we do send the tx to them, then we should log that the txid was requested and sent

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -32,6 +32,7 @@ module Ouroboros.Consensus.Node
   , mkNodeArgs
   ) where
 
+import           Codec.Serialise (DeserialiseFailure)
 import           Control.Tracer (Tracer)
 import           Crypto.Random
 import           Data.ByteString.Lazy (ByteString)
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -146,7 +148,7 @@ run tracers chainDbTracer diffusionTracers diffusionArguments networkMagic
                            ()
           networkApps = consensusNetworkApps
             nodeKernel
-            nullProtocolTracers
+            protocolTracers
```
mrBliss: Huh, I'm surprised to see this. I remember seeing flags in the node to trace the different protocols, apparently they didn't work?

intricate: Yeah so the protocol tracers were never being passed through.

mrBliss:
What's left to do?

intricate:
> What's left to do?

Nothing, this is good to go. Just marked the PR as "ready for review".

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -89,12 +90,13 @@ data IsProducer
 run
   :: forall blk.
      RunNode blk
-  => Tracers IO ConnectionId blk          -- ^ Consensus tracers
-  -> Tracer  IO (ChainDB.TraceEvent blk)  -- ^ ChainDB tracer
-  -> DiffusionTracers                     -- ^ Diffusion tracers
-  -> DiffusionArguments                   -- ^ Diffusion arguments
+  => Tracers IO ConnectionId blk                             -- ^ Consensus tracers
+  -> ProtocolTracers IO ConnectionId blk DeserialiseFailure  -- ^ Protocol tracers
```
mrBliss: Nitpick: this line is now longer 80 chars. Suggestion:
```haskell
  => Tracers IO ConnectionId blk          -- ^ Consensus tracers
  -> ProtocolTracers IO ConnectionId blk DeserialiseFailure
     -- ^ Protocol tracers
  -> Tracer  IO (ChainDB.TraceEvent blk)  -- ^ ChainDB tracer
```
You can revert the changes that realign the Haddock comments.

Issue #1047, opened 2019-09-20
Enable `checktvarinvariant` on CI


edsko:
In https://github.com/input-output-hk/ouroboros-network/pull/1040 I added an invariant to enable assertions on `StrictTVars`. We should enable this on CI.

mrBliss:
Maybe somebody with Nix experience can help us out? We need the `test-storage` and `test-consensus` testsuites from `ouroboros-consensus` to run with the `checktvarinvariant` flag (for the `io-sim-classes` package) enabled.

Running the tests with that flag enabled takes quite long, so it **should not** be part of the CI checks for PRs. That would slow down the merge process even more. We want a nightly CI job for this.

@nc6 You wrote https://github.com/input-output-hk/ouroboros-network/blob/master/nix/validate-mainnet.nix, maybe you know what to do?

mrBliss:
I have received some pointers:

* Add
  ```
  package.ouroboros-consensus.component.test.test-storage.flags = { checktvarinvariant = true; };
  ```
  to https://github.com/input-output-hk/ouroboros-network/blob/da1bd0117399d12e485c112f8413e865fed5bb0f/nix/pkgs.nix#L44
  But we want to build that component with the flag enabled for `io-sim-classes`? We actually want to define an additional variant of an existing component with different flags.
* To define a nightly job, create `.buildkite/nightly.yml` (look at the other `.yml` file for inspiration). Create a new pipeline in Buildkite: https://buildkite.com/input-output-hk/cardano-sl-nightly/settings/schedules. You can use the `new build` button to force a build instead of waiting a day.

PR #1602, opened 2020-02-07
Fix thunks in RealPBFT

mrBliss:
This fixes the thunks found by running the RealPBFT consensus tests with the `checktvarinvariant` flag enabled for the `io-sim-classes` package.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -116,25 +116,25 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-ledger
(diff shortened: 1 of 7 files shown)


mrBliss:
If we had #1047, we would have discovered this much sooner.

mrBliss:
CI failure (dependency not in master) is because bors still hasn't merged https://github.com/input-output-hk/cardano-ledger/pull/717 :weary:

edsko:
Oh, right. Come back in a week.

PR #1603, opened 2020-02-07
Avoid use of TMVar

edsko:
As this does not support an invariant (thunks check). This also avoids
some race conditions in the code.

I didn't end up using `modifyMVar_` despite introducing it, but since it
might be useful at another point, left it in.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
@@ -53,9 +53,7 @@ import           Data.Word (Word32, Word64)
 import           GHC.Generics (Generic)
 import           GHC.Stack (HasCallStack, callStack)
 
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/Ouroboros/Storage/ImmutableDB/Impl/Index/Cache.hs
```diff
@@ -356,7 +354,8 @@ data CacheEnv m hash h = CacheEnv
   , tracer      :: Tracer m TraceCacheEvent
   , cacheVar    :: StrictMVar m (Cached hash)
   , cacheConfig :: CacheConfig
-  , bgThreadVar :: StrictTMVar m (Thread m Void)
+  , bgThreadVar :: StrictMVar m (Maybe (Thread m Void))
+                -- ^ Nothing if no thread running
```
mrBliss: Align on `bgThreadVar` :pray: :smile:

edsko: Well okay then, but only because you helped debug this.

PR #1605, opened 2020-02-07
Fix the remaining thunks in the consensus tests

mrBliss:


ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/NormalForm.hs
```diff
@@ -60,8 +60,19 @@ unsafeNoThunks a = unsafePerformIO $ errorMessage =<< noUnexpectedThunks [] a
...
+        --
+        -- TODO rendering the tree has been disabled for now, as this loops
+        -- indefinitely, consuming gigabytes of memory, and prevents us from
+        -- printing a message about the thunk. Moreover, the thunk info is in
+        -- most cases a clearer indication of where the thunk is than the
+        -- /huge/ tree. Use the two commented-out lines below to include the
+        -- tree in the message.
+        --
+        -- tree <- Text.unpack <$> buildAndRenderClosureTree opts a
+        -- return $ Just $ show info ++ "\nTree:\n" ++ tree
```
intricate: btw just a note that you can also set a max closure tree depth to render via the `ClosureTreeOptions`. However, I'd agree that the thunk info is more useful :smile:

Issue #1355, opened 2019-12-16
New protocol version / migration for lazy mux and block sizes


karknu:
We need to define a new protocol version to accommodate a couple changes in the network and consensus.

The new version will support one flag `InitiatorOnly`, which indicates that the peer that started the version negotiation is not running the responder side of the miniprotocols. This change may not be able to use the protocol versioning mechanism, needs further review.

The consensus layer also needs to instantiate the chain sync to include the block size (needed to provide that info to block fetch). This might be done by a breaking change, or by using the protocol versioning.

edsko:
Consensus is ready after https://github.com/input-output-hk/ouroboros-network/pull/1632, we just need to enable it.

PR #1611, opened 2020-02-10
Version negotiation refactoring

karknu:
Version negotiation refactoring which will make #1355 easier to implement.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Protocol/Handshake/Version.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Protocol/Handshake/Version.hs
@@ -11,7 +11,8 @@ module Ouroboros.Network.Protocol.Handshake.Version
   , Application (..)
   , Version (..)
   , Sigma (..)
-
+  , Accept (..)
+  , Acceptable (..)
   , Dict (..)
   , DictVersion (..)
   , pickVersions
@@ -74,6 +75,18 @@ simpleSingletonVersions vNum vData extra r =
 data Sigma f where
   Sigma :: !t -> !(f t) -> Sigma f
 
+
+-- |
+-- A @'Maybe'@ like type which better explains its purpose.
+--
+data Accept
+  = Accept
+  | Refuse !Text
+  deriving (Eq, Show)
+
+class Acceptable v where
+  acceptableVersion :: v -> v -> Accept
+
 -- | Takes a pair of version data: local then remote.
 newtype Application r vData = Application
   { runApplication :: vData -> vData -> r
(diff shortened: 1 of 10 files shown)


ouroboros-network/src/Ouroboros/Network/Protocol/Handshake/Version.hs
```diff
@@ -60,6 +61,18 @@ simpleSingletonVersions vNum vData extra r =
...
   Sigma :: t -> f t -> Sigma f
 
+
+-- |
+-- A @'Maybe'@ like type which better explains its purpose.
+--
+data Accept
+  = Accept
+  | Refuse !Text
+  deriving (Eq, Show)
```
coot: That's a good move, they did not belong to the `Types` module.

ouroboros-network/src/Ouroboros/Network/Protocol/Handshake/Version.hs
```diff
@@ -60,6 +61,18 @@ simpleSingletonVersions vNum vData extra r =
...
+
+class Acceptable v where
+  acceptV :: v -> v -> Accept
```
coot: let's use 
```
acceptVersion :: versionParameters -> versionParameters -> Accept
```
Also this class, as part of public api, needs to be well documented: what is the intention how it is used, what are the two parameters, etc.

coot: Just an idea:
If instead of a type class we'd embed a `v -> v -> Accept` function, this would give us more flexibility how versions are accepted.  With a type class every time we use `v` we end-up with the same `Acceptable` instance.

coot: And it would also be easier for testing - no need to write instances for custom types.  

The expense is that `simpleSingletonVersion` would get additional argument.

karknu: There already is such a function for the responder side, see https://github.com/input-output-hk/ouroboros-network/blob/0841d21fa05cd02b2f606b62b0241edbca6138b7/ouroboros-network/src/Ouroboros/Network/Protocol/Handshake/Type.hs#L246 .
This pr only changes `Eq` instance which is used by the handshake test cases client to verify that the server managed to pick a correct version.
I'm not sure what to do with `simpleSingletonVersion`  yet, adding a function for two comparing two fields it currently ignores seems premature.

coot: Yes, we can easily change it later, when / if needed.

PR #1577, opened 2020-02-05
Introduce header validation as concept

edsko:
This then paves the way for introducing additional checks
(which this doesn't do yet).

PR #1621, opened 2020-02-11
Allow EBBs only when we expect them

edsko:


edsko:
This is only one commit, the first is https://github.com/input-output-hk/ouroboros-network/pull/1577 .

edsko:
I verified that this can still verify the (mirrored prefix of) the real chain, and that it fails to do so when we change the definition of `canBeEBB`.

PR #1623, opened 2020-02-11
Validate the ChainDB when the clean shutdown marker is missing

### LINKED ISSUE #1551: Handle restarting with validation automatically
Related to #1200 and #1201. When the node is being run by the wallet, it uses error codes to communicate information about the fatal error to the wallet (see #1541). For example, in case a file from the database is missing, the wallet should restart the node with the validation flag enabled.

In a server setting, where the node process is being run by systemd or runit, the node process will automatically be restarted when it terminated. There will be no human reading the error message or checking the error code to enable the validation flag when necessary.

We should be able to automatically validate when necessary.

Plan: add a marker file to the database on a clean shutdown (TODO define "clean"). On startup, if the marker is there, then we don't have to do any additional validation (just the last epoch, as we always do by default). If the marker is missing, we do a full validation.

mrBliss:
Fixes #1551.


ouroboros-consensus/src/Ouroboros/Consensus/Node/ExitFailure.hs
```diff
@@ -54,23 +53,18 @@ configurationError = 2
 noNetwork :: ExitFailure
 noNetwork = 3
 
--- | Something went wrong with the database, restart the node with recovery
--- enabled.
-restartWithRecovery :: ExitFailure
-restartWithRecovery = 4
```
edsko: Good thing this isn't in a release yet :)

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
```diff
@@ -0,0 +1,141 @@
...
+      _                         -> recoveryRequired
+
+{-------------------------------------------------------------------------------
+  Auxiliary
+-------------------------------------------------------------------------------}
+
+onExceptionIf
+  :: (IOLike m, Exception e)
+  => (e -> Bool)  -- ^ Predicate to selection exceptions
+  -> m b          -- ^ Exception handler
```
edsko: I never really understand why this customary. Wouldn't `m ()` make a lot more sense? (No need to change it.)

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
```diff
@@ -0,0 +1,141 @@
...
+hasCleanShutdownMarker
+  :: IOLike m
+  => HasFS m h
+  -> m Bool
+hasCleanShutdownMarker hasFS =
+    doesFileExist hasFS cleanShutdownMarkerFile
+
+-- | Create the 'cleanShutdownMarkerFile'.
+--
+-- Will throw an 'FsResourceAlreadyExist' error when it already exists.
```
edsko: Stale comment?

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
```diff
@@ -0,0 +1,141 @@
...
+removeCleanShutdownMarker hasFS =
+    removeFile hasFS cleanShutdownMarkerFile
+
+-- | Return 'True' if the given exception indicates that recovery of the
+-- database is required on the next startup.
+exceptionRequiresRecovery :: SomeException -> Bool
+exceptionRequiresRecovery e
+    | Just (ExceptionInLinkedThread _ e') <- fromException e
+    = exceptionRequiresRecovery e'
+    | Just (e' :: ChainDbFailure) <- fromException e
```
edsko: Quite a bit of duplicated logic now between here and `toExitFailure`?

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
```diff
@@ -0,0 +1,141 @@
...
+        _                        -> noRecoveryRequired
+    | Just (e' :: VolatileDBError) <- fromException e
+    = case e' of
+        VolDB.UnexpectedError ue -> volDbUnexpectedError ue
+        _                        -> noRecoveryRequired
+    | Just (e' :: FsError) <- fromException e
+    = fsError e'
+
+    | otherwise
+    = noRecoveryRequired
```
edsko: It makes me slightly uneasy that the _default_ is "no recovery required". But I guess it makes sense.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -122,59 +123,58 @@ run tracers chainDbTracer diffusionTracers diffusionArguments networkMagic
...
-                 (initiatorNetworkApplication networkApps)
-           , daLocalResponderApplication =
-               simpleSingletonVersions
-                 NodeToClientV_1
-                 nodeToClientVersionData
-                 (DictVersion nodeToClientCodecCBORTerm)
-                 (localResponderNetworkApplication networkApps)
-           , daErrorPolicies = consensusErrorPolicy
-           }
+        let nodeArgs = customiseNodeArgs $ mkNodeArgs
```
edsko: All these unrelated changes :roll_eyes: :grin:

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
mrBliss: Yeah, convenience (no `void`)? Or to make clear from the types which `m X` is the exception handler?

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
mrBliss: Ah, yes.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
edsko: I don't like it though. No `void` for the caller, sure, _because we'll implicitly through your result away_. Even `ghc` warns when you do that (in a `do` block). I prefer if it's the caller explicit choice to throw results away -- do you really want an exception handler with a result?

Issue #1571, opened 2020-02-04
Check that block numbers are consecutive


mrBliss:
Apparently, nowhere in the Byron or Shelley ledger is it verified that blocks have consecutive block numbers.

We do check it in [an assertion `ChainFragment`](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/ChainFragment.hs#L221), but in production, assertions will be disabled.

We should verify this at the earliest opportunity we have, which is `applyChainState` (called for synced headers but also when we apply blocks we produced ourselves, as part of `applyExtLedgerState`). In addition to check whether block numbers are consecutive, we should also check that the first block has the right block number, i.e., `genesisBlockNo`.

It is important that we check this, because we're assuming that block numbers are a reliable proxy for chain length. So if this is not the case, we might prefer a chain that is actually shorter. So both consecutiveness and the right initial block number are important. For example, an attacker might forge a single block with a very large block number, and we would incorrectly prefer it over all other chains.

edsko:
@nc6 , change of plan, no need to do this for each ledger (although the spec should still verify it, I think). I will do this once and for all for _all_ ledgers. You actually discussed doing that when discussing the Byron integration, albeit for hashes, not block numbers; I'm coming round to your point of view now :) PR to follow, unassigning you from this ticket.

edsko:
WIP PR at #1577 .

edsko:
So #1578 cleans up a lot of our `BlockNo` hygiene, but I'd like to fix https://github.com/input-output-hk/cardano-base/issues/78 before fixing this ticket, in order to ensure that we don't have other places with bad `BlockNo` hygiene.

PR #1625, opened 2020-02-11
ChainDB: allow EBB with same block number as the immutable block

### LINKED ISSUE #1624: Chain DB can't add the next EBB when the VolatileDB is empty
Say the VolatileDB has been wiped/corrupted. Now the tip of our current chain is the tip of the ImmutableDB. Say that tip is the last block of some epoch. When we now try to add the next EBB, that EBB has the same block number as the immutable block's tip, and is thus ignored. Which means we *can't make progress anymore*. Restarting and rewiping the VolatileDB doesn't help.

The bug is here: https://github.com/input-output-hk/ouroboros-network/blob/c0f1e1cc2d61e850c2f422272effea4047c58ba9/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ChainSel.hs#L189

The `=` in the `<=` excludes EBBs. The second condition is already sign there's something wrong, it just special cases the very first EBB.

mrBliss:
Fixes #1624.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ChainSel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ChainSel.hs
@@ -19,6 +19,8 @@ module Ouroboros.Storage.ChainDB.Impl.ChainSel
   , ChainAndLedger -- Opaque
   , clChain
   , clLedger
+    -- * Exported for testing purposes
+  , olderThanK
   ) where
 
 import           Control.Exception (assert)
@@ -49,7 +51,7 @@ import           Ouroboros.Network.Block (BlockNo, pattern BlockPoint,
 import           Ouroboros.Network.Point (WithOrigin (..))
 
 import           Ouroboros.Consensus.Block (BlockProtocol, GetHeader (..),
-                     headerHash, headerPoint)
+                     IsEBB (..), headerHash, headerPoint)
 import           Ouroboros.Consensus.BlockchainTime (BlockchainTime (..))
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Ledger.Extended
@@ -186,7 +188,7 @@ addBlock cdb@CDB{..} b = do
 
     -- ### Ignore
     if
-      | At (blockNo b) <= immBlockNo && blockNo b /= 0 ->
+      | olderThanK hdr (cdbIsEBB hdr) immBlockNo ->
         trace $ IgnoreBlockOlderThanK (blockPoint b)
       | isMember (blockHash b) ->
         trace $ IgnoreBlockAlreadyInVolDB (blockPoint b)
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
```diff
@@ -364,12 +361,7 @@ addBlock cfg blk m
   where
     secParam = protocolSecurityParam cfg
 
-    -- If we switch to PBFT for these tests, this case is not required anymore
-    -- TODO: We should not make assumptions about the underlying
-    -- ledger. We will fix this in
-    -- <https://github.com/input-output-hk/ouroboros-network/issues/1571>
```
mrBliss: This special case is gone without #1571. Unverified: maybe we only needed it to not ignore the EBB, but we still don't adopt EBBs immediately when adding them, only when their successor is added? If that's true, then PBFT would change that, but then it is not that important to make that change.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/Model/Test.hs
```diff
@@ -93,3 +94,10 @@ prop_between_currentChain bt =
     from     = StreamFromExclusive genesisPoint
     to       = StreamToInclusive $ M.tipPoint model
     secParam = protocolSecurityParam singleNodeTestConfig
+
+{-------------------------------------------------------------------------------
+  Orphan instances
+-------------------------------------------------------------------------------}
+
+instance ModelSupportsBlock TestBlock where
+  isEBB = const IsNotEBB
```
mrBliss: A bit annoying: this TestBlock lives in `test-util`, so it can implements `ModelSupportsBlock` (`test-consensus`) itself, hence the orphans :disappointed:.

ouroboros-consensus/src/Ouroboros/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -306,7 +310,11 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr = do
     if
       -- The chain might have grown since we added the block such that the
       -- block is older than @k@.
-      | At (blockNo hdr) <= immBlockNo && blockNo hdr /= 0 ->
+      | At (blockNo hdr) <= immBlockNo &&
+        -- Exception: the block we're adding is the EBB with the same block
+        -- number as the immutable tip. This can only happen when the
+        -- VolatileDB was corrupted.
+        not (At (blockNo hdr) == immBlockNo && cdbIsEBB hdr == IsEBB) ->
```
edsko: Should we introduce an `olderThanK` predicate, document it with this comment, and then use that here and above rather than duplicate this logic?

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
edsko: Hang on, didn't we verify this? And it still failed? What am I getting confused with?

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
```diff
@@ -326,18 +326,15 @@ advanceCurSlot cfg curSlot m =
...
     advance slot m' = m' { currentSlot = slot `max` currentSlot m' }
 
-addBlock :: forall blk. ProtocolLedgerView blk
+addBlock :: forall blk. (ProtocolLedgerView blk, ModelSupportsBlock blk)
          => NodeConfig (BlockProtocol blk)
          -> blk
          -> Model blk -> Model blk
 addBlock cfg blk m
     -- If the block is as old as the tip of the ImmutableDB, i.e. older than
-    -- @k@, we ignore it, as we can never switch to it. TODO what about EBBs?
```
edsko: What about EBBs indeed :angry:

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
edsko: (This will have a minor conflict with my PR I guess.)

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
```diff
@@ -326,18 +326,15 @@ advanceCurSlot cfg curSlot m =
...
-  | At (Block.blockNo blk) <= immutableBlockNo secParam m
-    -- Unless we're adding the first block to the empty chain: the empty chain
-    -- has the same block number as the genesis EBB, i.e. 0, so we don't want
-    -- to ignore it in this case.
-  , not addingGenesisEBBToEmptyDB
+    -- @k@, we ignore it, as we can never switch to it.
+  | At (Block.blockNo blk) <= immBlockNo
+  , not (At (Block.blockNo blk) == immBlockNo && isEBB (getHeader blk) == IsEBB)
```
edsko: So this is basically the same change from the real DB to the model.

mrBliss: Indeed.

mrBliss: In my last version, I reuse the new function from `.ChainSel` in the model.

mrBliss:
I have created #1628 for testing this properly. Note that I have already verified it manually using `cardano-node` (I discovered this bug when I wiped the VolatileDB of a `cardano-node`).

PR #1634, opened 2020-02-12
versions semigroup

coot:
I think the `Semigroup` instance is enough, `mempty` shouldn't be really used for `Versions`.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Protocol/Handshake/Version.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Protocol/Handshake/Version.hs
@@ -72,7 +72,7 @@ simpleSingletonVersions vNum vData extra r =
         (Sigma vData (Version (Application $ \_ _ -> r) extra))
 
 data Sigma f where
-  Sigma :: t -> f t -> Sigma f
+  Sigma :: !t -> !(f t) -> Sigma f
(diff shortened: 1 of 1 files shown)


ouroboros-network/src/Ouroboros/Network/Protocol/Handshake/Version.hs
```diff
@@ -35,9 +36,22 @@ import Codec.SerialiseTerm
...
+-- >       :| [ (simpleSingletonVersions ...)
+-- >          , (simpleSingletonVersions ...)
+-- >          , ...
+-- >          ]
+-- >
+--
 newtype Versions vNum extra r = Versions
   { getVersions :: Map vNum (Sigma (Version extra r))
   }
+  deriving Semigroup via (Map vNum (Sigma (Version extra r)))
```
mrBliss: Wouldn't `deriving Semigroup` also have worked? Or `deriving newtype Semigroup`?

coot: I am not sure.  But this is very explicit, so I like it over `deriving Semigroup`, `deriving newtype` - is equally good to me.

manveru:
bors retry

PR #1643, opened 2020-02-13
Handshake protocol number belongs to 'ouroboros-network'

coot:
It should be defined in 'ouroboros-network' rather than in
'network-mux'.  There is no control protocol in 'network-mux' anymore.
This patch adds haddoc docstrings, renames 'muxBearerAsControlChannel'
to 'muxBearerAsChannel'.

### DIFF
--- a/network-mux/src/Network/Mux/Types.hs
+++ b/network-mux/src/Network/Mux/Types.hs
@@ -18,7 +18,7 @@ module Network.Mux.Types (
 
     , MiniProtocolIx
     , MuxBearer (..)
-    , muxBearerAsControlChannel
+    , muxBearerAsChannel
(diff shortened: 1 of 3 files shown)


network-mux/src/Network/Mux/Types.hs
```diff
@@ -185,14 +182,15 @@ data MuxBearer m = MuxBearer {
     }
 
 
--- | A channel which wraps each message as an 'MuxSDU', each sdu is send as
--- 'Mx.Muxcontrol'.
+-- | A channel which wraps each message as an 'MuxSDU' using giving
```
karknu: Shouldn't it be `given`?

Issue #1652, opened 2020-02-17
Separate out Byron and Shelley integration from consensus


edsko:
At the moment, the Byron and Shelley integrations in the consensus layer are part of the main library. This made sense as an initial design where the API was still very much in flux. However, as the API is stabilizing and the consensus layer is starting to be used for other integrations as well, this should be separated out. Doing so will moreover clarify the existing code and also clarify responsibilities between teams.

PR #1657, opened 2020-02-18
Reorganise ouroboros-consensus

### LINKED ISSUE #1651: Extract consensus test infrastructure into separate lib
In particular, this involves

* `ThreadNet.Network`, `.General` and `.Util`
* Everything in `test-util`

This will make it possible for other applications of the consensus layer to reuse our tests as well.

mrBliss:
Fixes #1651 and #1652.

* Split off Byron related code and tests to `./ouroboros-consensus-byron`.

* Split off Byron Spec related code and tests to
  `./ouroboros-consensus-byronspec`.

* Split off mock block related code and tests to
  `./ouroboros-consensus/ouroboros-consensus-mock`.

* Split off test infrastructure code and tests to
  `./ouroboros-consensus/ouroboros-consensus-test-infra`.

These last two are both internal libraries of `ouroboros-consensus` and
separate packages, because the test suites of `ouroboros-consensus` depend on
them, while other packages, i.e., `ouroboros-consensus-byron`, `cardano-node`,
..., will also depend on them. If only Cabal supported multiple public
libraries in one project.

* Remove the `Exception` instance of `PBftLeaderCredentialsError`, as it was
  never actually thrown.

* Add `nodeToExitReason`, as it can be `blk`-dependent.

* Rename `Ouroboros.Storage` to `Ouroboros.Consensus.Storage`.


ouroboros-consensus/test-consensus/Test/Consensus/Protocol/PBFT.hs
```diff
@@ -545,6 +546,18 @@ prop_appendOldStateRestoresPreWindow TestChainState{..} =
        .&&.
        size (CS.preWindow state') === maxRollbacks testChainStateK
 
+{-------------------------------------------------------------------------------
+  Serialisation roundtrip
+-------------------------------------------------------------------------------}
+
+-- | For now we test the roundtrip using mock crypto instead of real crypto
```
mrBliss: Update comment

ouroboros-consensus/test-consensus/Test/Consensus/BlockchainTime/SlotLengths.hs
```diff
@@ -171,67 +171,3 @@ prop_delayEffective (SystemStart start) ls (SmallDiffTime offset) =
 
     (_this'', (slotAfterDelay, timeInSlot)) =
         slotFromUTCTime (SystemStart start) timeAfterDelay focused
-
-
-{-------------------------------------------------------------------------------
-  Arbitrary instances
--------------------------------------------------------------------------------}
-
-instance Arbitrary SlotLengths where
```
mrBliss: Moved to `Test.Util.Orphans.Arbitrary`

edsko:
While we're at it (can be separate PR)): `ChainSync.Client` and `ChainSync.Server`?

edsko:
The `-byron` package should probably have a different module hiearchy: `Ouroboros.Consensus.Byron.*`? Otherwise people like @mrBliss might start complaining it's hard to locate modules :)

edsko:
Moving `Ouroboros.Storage` to `Ouroboros.Consensus.Storage`, but we should probably flatten this a bit:

* Merge `Ouroboros.Consensus.Util` and `Ouroboros.Consensus.Storage.Util`
* Move up `Ouroboros.Consensus.Storage.Common`
* Move up some other modules (`EpochInfo` for instance)?

PR #1672, opened 2020-02-19
PraosValidationError:  export

deepfire:
Export `PraosValidationError`.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
+++ b/ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
@@ -29,6 +29,7 @@ module Ouroboros.Consensus.Mock.Protocol.Praos (
   , PraosMockCrypto
   , PraosValidateView(..)
   , praosValidateView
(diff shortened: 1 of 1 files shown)


deepfire:
> Just curious: what do you need it for?

@mrBliss, @edsko, I would appreciate a hint to avoid this requirement indeed.

Basically the situation is:

1. `mkTracers` in https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Tracing/Tracers.hs#L154 is polymorphic re `blk`, and in it I need `ToObject (LedgerError blk)` satisfied, so `toLogObject` can be called -- thereby I need `ToObject (LedgerError blk)` added to the constraint of `mkTracers`
2. but then, this abstract constraint needs to be satisfied at the call site of `mkTracers`, in `runNode`: https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/Node/Run.hs#L120 : `(Cardano.BM.Data.Tracer.ToObject                           (Ouroboros.Consensus.Ledger.Abstract.LedgerError blk))`
3. ..at which point it must come from `TraceConstraints` -- which was my solution -- perhaps a misguided one?

edsko:
I see, yes, that makes sense!

PR #1676, opened 2020-02-20
Cardano client library

MarcFontaine:


cardano-api/src/Cardano/Client/API.hs
```diff
@@ -0,0 +1,318 @@
+-- consensus uses byron types
+-- type-instantiations for codecs
+
+module Cardano.Client.API
+(
+-- Ouroboros.Network.Point
```
coot: You can turn those comment in to haddock section headers
```
-- * Ouroboros.Network.Point re-exports
```

It's also good to align them vertically with the rest of the list

cardano-api/src/Cardano/Client/API.hs
```diff
@@ -0,0 +1,318 @@
...
+    , ByronBlock (ByronBlock, byronBlockRaw, byronBlockSlotNo, byronBlockHash)
+    , ByronHash (..)
+    , GenTx
+-- Ouroboros.Consensus.Ledger.Abstract
+    , BlockProtocol
+-- Ouroboros.Consensus.Node.ProtocolInfo
+    , pInfoConfig
+    , protocolInfo
+-- Ouroboros.Consensus.Node.Run.Abstract
+    , RunNode
```
coot: ^ This class is for running a node.  If it's really needed then we can suggest a refactoring of this class into smaller parts.
Note that we are not aiming to support `cardano-node` cabal package in `cardano-node` repository (e.g. the node); only clients that are in that repo (`cardano-cli`, `chairman`, ...).

cardano-api/src/Cardano/Client/API.hs
```diff
@@ -0,0 +1,318 @@
...
+    , chainSyncClientPeer
+-- import Ouroboros.Network.Protocol.Handshake.Version
+    , DictVersion (..)
+    , simpleSingletonVersions
+-- import Network.Mux.Interface
+-- import Network.Mux.Types
+--    , MuxError
+    )
+
+where
```
coot: I think it's just nicer to put the `where` just after the closing bracket.  That what we've been using in `ouroboros-network`.

mrBliss:
Please rebase against master, consensus recently did a large reorganisation of the packages (#1657).

Also, just re-exporting the imports from one place is not enough. You need some more abstractions and encapsulation for the network code. We have tried that for consensus in the form of `Node.run`:
https://github.com/input-output-hk/ouroboros-network/blob/64006b7331b3f70580ee70fad7808e7a0fe242d2/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L120

In `cardano-node`, they can just call that function and it hides most of the consensus internals:
https://github.com/input-output-hk/cardano-node/blob/1ef04c946bfce4b6545d3c138e6eb2881c4a48cf/cardano-node/src/Cardano/Node/Run.hs#L200

Note that most of the arguments that function takes are related to tracing and _networking_.

Various places in `cardano-node` need to set up some network connections. These should just be one or two function calls with some arguments. Currently, way too many implementation details leak through to these places. For example: https://github.com/input-output-hk/cardano-node/blob/29f02eb9741516586c68ee8345e5e3d59dbc0581/cardano-node/src/Cardano/CLI/Benchmarking/Tx/NodeToNode.hs
Changes to the network code almost always require fixes on these locations at the end of the sprint.

A start could be to create a facade for lines 198-252 in https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs

MarcFontaine:
The latest approved version of the API is in `cardano-client`.
This is superseded by: https://github.com/input-output-hk/ouroboros-network/pull/2159

Issue #1545, opened 2020-01-31
Re-enable Praos tests and update for KES


edsko:

edsko:
Alternatively, if the TPraos implementation is parameterized over crypto and can be run with our mock blocks, we could consider removing BFT and Praos entirely, leaving only PBFT and TPraos (PBFT _is_ supported by the mock blocks).

edsko:
Depending on what we decide, this may or may not be obsoleted by #1653 .

edsko:
This will be done as part of https://github.com/input-output-hk/ouroboros-network/pull/1677 , although it's currently a bit of a hack; see https://github.com/input-output-hk/cardano-base/issues/81.

PR #1677, opened 2020-02-20
Use the IO instance of MonadRandom

### LINKED ISSUE #1616: Don't run block production in STM
This is a left-over from the days when we did lots of things in a single transaction in block producton. However, as of https://github.com/input-output-hk/ouroboros-network/pull/1445 that is not the case anymore, and the call to actually forge a block is now a singleton `atomically`. There should be no need to run this in STM, and _not_ running in STM has benefits: we can avoid the PRNG indirection, just removing some potential doubts about crypto entropy sources, and it will also benefit @nfrisby 's testing infrastructure.

mrBliss:
Fixes #1616.

Previously, block production had to live in STM, but this is no longer the
case as of #1445. This means that we no longer need to run `MonadRandom`
computations in STM. To do that, we stored a DRG in a TVar.

Now, we can simply use IO's instance of `MonadRandom`. In the tests we still
use the DRG-in-a-TVar trick, but we split it whenever we get a DRG.

We use the `RunMonadRandom` record for this purpose.

---

Replace NodeState with MonadState

Previously, we need a separate `HasNodeState` class and `NodeStateT` monad
transformer because there was already a `StateT` in our stack, i.e., the one
containing the DRG. As that is gone, we can switch back to a regular
`MonadState` + `StateT`.

The cost is an orphan instance: `MonadRandom (StateT s m)`

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
+++ b/ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
@@ -1,12 +1,11 @@
 {-# LANGUAGE BangPatterns              #-}
-{-# LANGUAGE DataKinds                 #-}
 {-# LANGUAGE DeriveAnyClass            #-}
 {-# LANGUAGE DeriveGeneric             #-}
-{-# LANGUAGE DerivingVia               #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleContexts          #-}
 {-# LANGUAGE FlexibleInstances         #-}
-{-# LANGUAGE MultiParamTypeClasses     #-}
+{-# LANGUAGE KindSignatures            #-}
+{-# LANGUAGE LambdaCase                #-}
 {-# LANGUAGE RecordWildCards           #-}
 {-# LANGUAGE ScopedTypeVariables       #-}
 {-# LANGUAGE StandaloneDeriving        #-}
@@ -47,6 +46,7 @@ import           Data.Proxy (Proxy (..))
 import           Data.Typeable
 import           Data.Word (Word64)
 import           GHC.Generics (Generic)
+import           GHC.Stack
 import           Numeric.Natural
 
 import           Cardano.Crypto.DSIGN.Ed448 (Ed448DSIGN)
@@ -66,6 +66,7 @@ import           Ouroboros.Network.Block (HasHeader (..), SlotNo (..),
 import           Ouroboros.Network.Point (WithOrigin (At))
 
 import           Ouroboros.Consensus.Mock.Ledger.Stake
+import           Ouroboros.Consensus.Node.State
 import           Ouroboros.Consensus.NodeId (CoreNodeId (..), NodeId (..))
 import           Ouroboros.Consensus.Protocol.Abstract
 import           Ouroboros.Consensus.Protocol.Signed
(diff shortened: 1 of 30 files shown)


ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -125,7 +126,7 @@ forgePraosFields :: ( HasNodeState (Praos c) m
                  -> (PraosExtraFields c -> toSign)
                  -> m (PraosFields c toSign)
 forgePraosFields PraosNodeConfig{..} PraosProof{..} mkToSign = do
```
mrBliss: Re-enable the Praos tests (#1545).

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -497,17 +492,22 @@ forkBlockProduction maxBlockSizeOverride IS{..} BlockProduction{..} =
...
+      -> m a
+    runProtocol n = do
+      nodeState <- atomically $ readTVar varState
+
+      (a, nodeState') <-
+        runMonadRandom runMonadRandomDict $
+        runNodeStateT n nodeState
+
+      atomically $ writeTVar varState nodeState'
+      return a
```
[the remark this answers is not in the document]
edsko: Done. Also removed the orphan `MonadRandom ` instance for `StateT`, which is now not necessary anymore.

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
edsko: Done.

ouroboros-consensus/src/Ouroboros/Consensus/Node/State.hs
```diff
@@ -0,0 +1,48 @@
...
+           -> Update m small
+liftUpdate get set (Update update) = Update $ \f ->
+    update $ \large ->
+      first (flip set large) <$> (f (get large))
+
+{-------------------------------------------------------------------------------
+  State
+-------------------------------------------------------------------------------}
+
+-- | (Chain-independent) node required to run the protocol
```
edsko: node state

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -115,39 +116,85 @@ praosValidateView :: ( HasHeader    hdr
...
                  => NodeConfig (Praos c)
+                 -> Update m (PraosNodeState c)
                  -> PraosProof c
                  -> (PraosExtraFields c -> toSign)
                  -> m (PraosFields c toSign)
-forgePraosFields PraosNodeConfig{..} PraosProof{..} mkToSign = do
-    keyKES <- unPraosNodeState <$> getNodeState
+forgePraosFields PraosNodeConfig{..} updateState PraosProof{..} mkToSign = do
+    -- For the mock implementation, we consider the KES period to be the slot.
+    -- In reality, there will be some kind of constants slotsPerPeriod factor.
```
edsko: constant

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -115,39 +116,85 @@ praosValidateView :: ( HasHeader    hdr
...
-                    , MonadRandom m
+{-------------------------------------------------------------------------------
+  Forging
+-------------------------------------------------------------------------------}
+
+data PraosNodeState c =
+    -- | The KES key is available
+    PraosKeyAvailable (SignKeyKES (PraosKES c))
+
+    -- | The KES key is being evolved
```
edsko: Comment: any thread that modifies the state and sees this value should back off and wait.

PR #1679, opened 2020-02-20
Reduce the use of `TopLevelConfig`

edsko:
Instead, attempt to pass only _parts_ of the config. This also does some additional refactoring.


ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Conversions.hs
```diff
@@ -32,9 +40,32 @@ fromByronSlotNo = coerce
...
+  Extract info from genesis
+-------------------------------------------------------------------------------}
+
+genesisSecurityParam :: Genesis.Config -> SecurityParam
+genesisSecurityParam =
+      fromByronBlockCount
+    . Genesis.gdK
+    . Genesis.configGenesisData
+
+
```
edsko: Newline

Issue #1638, opened 2020-02-13
Duplicate block in vol DB?


edsko:
One of the Byron proxies was failing with

```
cardano-byron-proxy: UnexpectedError (ParserError (DuplicatedSlot (ByronHash{unByronHash = AbstractHash 194922fbdfd6060503a0cdce7b1d979a1802fa7b2c023fce59a83e54b982b254}) blocks-170.dat blocks-170.dat))
```

After https://github.com/input-output-hk/ouroboros-network/pull/1568 we will handle these errors and truncate the volatile DB when this happens, but it would still be good to know why it happened in the first place. Unfortunately, the fact that we will now just automatically truncate the volatile DB when we see a parse failure like this means that we're not very likely to see it in the wild. Reproducing it may be difficult.

kderme:
The volatile db had a single file, which surprisingly had 18665 blocks. Two consecutive blocks (ordered 18627 and 18628) has same hash `194922fbdfd6060503a0cdce7b1d979a1802fa7b2c023fce59a83e54b982b254`.

edsko:
> which surprisingly had 18665 blocks.

Why do you say this is surprising?

kderme:
I think we have 1000 max blocks per file.

kderme:
`ChainDB.cdbBlocksPerFile    = 1000`

edsko:
But this is the proxy, which has

```haskell
  customiseArgs :: ChainDbArgs IO ByronBlock -> ChainDbArgs IO ByronBlock
  customiseArgs args = args
    { cdbBlocksPerFile = 21600 -- ?
    , cdbGcDelay = secondsToDiffTime 20
    }
```

(Don't ask me why :man_shrugging: :) )

kderme:
I think this is related to how we don't close the db when there is an exception while modifying its state: `ExitCaseException _ex                 -> (mst, return ())`
Also possibly related, I count at least two threads which write blocks concurrently and it was mentioned that the process was going out of memory. According to these, a possible scenario is the following:

A thread tries to insert a block, manages to write the blocks to disk, but fails to update the in memory state due to an async exception. Before the exception kills the process, another thread tries to insert the same block, writes it again to disk and updates the in memory state. It manages to write a few other blocks before the process dies. Upon restarting the db is found corrupted.

Sounds possible?

I think, in any case, the internal state should be tagged as corrupted when we get an exception and we shouldn't just restore the previous state.

[...]

PR #1680, opened 2020-02-21
VolatileDB: rewrite the model and q-s-m tests

mrBliss:
Fixes #1639.

* Follow the ChainDB and ImmutableDB: write the model as pure functions
  instead of using `MonadState`.

* Simplify the model and the q-s-m tests by imitating the VolatileDB's
  file-to-block mapping. While a map of `BlockId`s would be nicer, following
  the implementation more closely makes things so much simpler: simulating
  corruption is now a piece of cake. Keeping in sync during GC is also simple,
  avoiding complex precondition (e.g., don't ask for a block that might have
  been GC'ed).

* Use `Test.Storage.TestBlock` and its `Corruptions` directly.

* Adapt the Mock accordingly.

* Hide/remove internals from the implementation that are no longer needed for
  tests.

PR #1681, opened 2020-02-21
Handle async exceptions better

kderme:
This pr improves the way the Volatile DB handles exceptions. See also https://github.com/input-output-hk/ouroboros-network/issues/1638#issuecomment-586964762

So far only filesystem errors were handled and tested. In these cases the db closes, reopens, parses all blocks and, recovers to a consistent state. The db has the property that if any command that failed is executed again, the db reaches a state as if it never failed in the first place. For example if during gc some files fail to be removed, calling the same gc will remove them successfully. Or if half the bytes of a block are written, they will be truncated during reopening and the block can be successfully inserted after that.

This pr tries to get the idea to the next level: no reopening in between should actually be needed. We can use this property to make sure the db stays consistent even after async exceptions. The main idea is the following: if some bytes were written, but the internal state was not updated because of an exception, the db will overwrite the existing bytes with the next block. There are some edge cases that were also taken care of:

- `pwrite` is introduced, so that we never use the file offset that the os keeps.
- It's possible that a `putBlock` call creates a new file before it receives an exception. The next `putBlock` will overwrite the bytes and find that the next file exists, so it will fail. To fix this, we now open with `AllowExisting`.
- It's possible that a `putBlock` call opens a new handle before it receives an exception, so this handle may leak. To fix we use `ResourceRegistry` for these handles. This `ResourceRegistry` differs from regular ones, in that it only releases its resources if it meets an exception (`bracketOnError`). So it doesn't have a global scope, but we open a new ResourceRegistry when we call `putBlock`.
- It's possible that a `putBlock` call closes the current handle before it receives an exception, so the next call should not use this handle. To fix this, we add a new flag `_exceptionEncountered` which tags the internal state when an exception is encountered. This is an indication that the current write handle may be closed and should not be used. A new handle is opened in this case. Again `ResourceRegistry` is used, so that no handles leak on next exceptions.
- we never gc the most recent file, as this may change the file offset.

This pr also adds tests both in the fs, for the new `pwrite` syscall, but also in the VolatileDB. In the VolatileDB we introduce a new command, where a thread receives an async exception while it calls putBlock. The property that we test is that this call is actually a no-op. There is some uncertainty though, in exactly which point this exception was thrown, so we take some additional steps to make sure we simulate a case where the exception was thrown in the middle of a `putBlock` call.

Things TODO:
- documentation should be added
- gc can't handle async exceptions. This is because, on a next call, if we try to remove a file which it is not there, that's an error. If we handle these errors in a lenient way, this is ready to change though.
- Add windows support for `pwrite`

### DIFF
--- a/ouroboros-consensus/src-unix/Ouroboros/Consensus/Storage/IO.hs
+++ b/ouroboros-consensus/src-unix/Ouroboros/Consensus/Storage/IO.hs
@@ -8,6 +8,7 @@ module Ouroboros.Consensus.Storage.IO (
     , read
     , pread
     , write
+    , pwrite
     , close
     , getSize
     , sameError
@@ -25,7 +26,8 @@ import           System.Posix (Fd)
 import qualified System.Posix as Posix
 
 -- Package 'unix' exports the same module.
-import           "unix-bytestring" System.Posix.IO.ByteString (fdPreadBuf)
+import           "unix-bytestring" System.Posix.IO.ByteString (fdPreadBuf,
+                     fdPwriteBuf)
 
 import           Ouroboros.Consensus.Storage.FS.API.Types (AllowExisting (..),
                      FsError, OpenMode (..), SeekMode (..), sameFsError)
@@ -78,9 +80,14 @@ open fp openMode = Posix.openFd fp posixOpenMode fileMode fileFlags
 
 
 -- | Writes the data pointed by the input 'Ptr Word8' into the input 'FHandle'.
-write :: FHandle -> Ptr Word8 -> Int64 -> IO Word32
+write :: FHandle -> Ptr Word8 -> Word32 -> IO Word32
 write h data' bytes = withOpenHandle "write" h $ \fd ->
     fromIntegral <$> Posix.fdWriteBuf fd data' (fromIntegral bytes)
+
+pwrite :: FHandle -> Ptr Word8 -> Word32 -> Word64 -> IO Word32
+pwrite h data' bytes offset = withOpenHandle "pwrite" h $ \fd ->
+    fromIntegral <$>
+      fdPwriteBuf fd data' (fromIntegral bytes) (fromIntegral offset)
 
 -- | Seek within the file.
 --
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/FS/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/FS/API.hs
@@ -14,8 +14,11 @@ module Ouroboros.Consensus.Storage.FS.API (
     , hGetAll
     , hGetAllAt
     , hPut
+    , hPutAt
     , hPutAll
+    , hPutAllAt
     , hPutAllStrict
+    , hPutAllStrictAt
     , withFile
     ) where
 
@@ -96,6 +99,13 @@ data HasFS m h = HasFS {
     -- bytestring is empty, in which case the return value can be 0.
   , hPutSome                 :: HasCallStack => Handle h -> BS.ByteString -> m Word64
 
+    -- | Same as 'hPutSome', but does not affect the file offset. An additional
+    -- argument is used to specify the offset.
+  , hPutSomeAt               :: HasCallStack
+                             => Handle h
+                             -> BS.ByteString
+                             -> AbsOffset
+                             -> m Word64
     -- | Truncate the file to the specified size
     --
     -- NOTE: Only supported in append mode.
@@ -243,6 +253,26 @@ hPutAllStrict hasFS h = go 0
         then return written'
         else go written' bs'
 
+-- | Like 'hPutAllStrict' but it does not change or depend on the file offset.
+-- @pwrite@ syscall is used internally.
+hPutAllStrictAt :: forall m h
+                .  (HasCallStack, Monad m)
+                => HasFS m h
+                -> Handle h
+                -> BS.ByteString
+                -> AbsOffset
+                -> m Word64
+hPutAllStrictAt hasFS h = go 0
+  where
+    go :: Word64 -> BS.ByteString -> AbsOffset ->  m Word64
+    go !written bs offset = do
+      n <- hPutSomeAt hasFS h bs offset
+      let bs'      = BS.drop (fromIntegral n) bs
+          written' = written + n
+      if BS.null bs'
+        then return written'
+        else go written' bs' (offset + fromIntegral n)
+
 -- | This function makes sure that the whole 'BL.ByteString' is written.
 hPutAll :: forall m h
         .  (HasCallStack, Monad m)
@@ -257,6 +287,24 @@ hPutAll hasFS h = foldM putChunk 0 . BL.toChunks
       written' <- hPutAllStrict hasFS h chunk
       return $! written + written'
 
+-- | Like 'hPutAll' but it does not change or depend on the file offset.
+-- @pwrite@ syscall is used internally.
+hPutAllAt :: forall m h
+          .  (HasCallStack, Monad m)
+          => HasFS m h
+          -> Handle h
+          -> BL.ByteString
+          -> AbsOffset
+          -> m Word64
+hPutAllAt hasFS h bs offset = fst <$> foldM putChunk (0, offset) (BL.toChunks bs)
+  where
+    putChunk :: (Word64, AbsOffset) -> BS.ByteString -> m (Word64, AbsOffset)
+    putChunk (written, offset') chunk = do
+      written' <- hPutAllStrictAt hasFS h chunk offset'
+      let !writtenTotal = written +              written'
+          !newOffset    = offset' + fromIntegral written'
+      return (writtenTotal, newOffset)
+
 -- | This function makes sure that the whole 'Builder' is written.
 --
 -- The chunk size of the resulting 'BL.ByteString' determines how much memory
@@ -268,3 +316,14 @@ hPut :: forall m h
      -> Builder
      -> m Word64
 hPut hasFS g = hPutAll hasFS g . BS.toLazyByteString
+
+-- | Like 'hPut' but it does not change or depend on the file offset.
+-- @pwrite@ syscall is used internally.
+hPutAt :: forall m h
+       .  (HasCallStack, Monad m)
+       => HasFS m h
+       -> Handle h
+       -> Builder
+       -> AbsOffset
+       -> m Word64
+hPutAt hasFS g builder = hPutAllAt hasFS g (BS.toLazyByteString builder)
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/FileInfo.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/FileInfo.hs
@@ -25,10 +25,11 @@ import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           GHC.Generics (Generic)
 
-import           Cardano.Prelude (NoUnexpectedThunks)
+import           Cardano.Prelude (NoUnexpectedThunks, Word64)
 
 import           Ouroboros.Network.Block (MaxSlotNo (..), SlotNo)
 
+import           Ouroboros.Consensus.Storage.FS.API.Types
 import           Ouroboros.Consensus.Storage.VolatileDB.Types
 
 {-------------------------------------------------------------------------------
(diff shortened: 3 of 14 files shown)


kderme:
@mrBliss pr https://github.com/input-output-hk/ouroboros-network/pull/1680 removed many exports of the VolatileDB.Impl, for test purposes, which are used by this pr. I'm not sure if it's ok to restore them.

ouroboros-consensus/src-unix/Ouroboros/Consensus/Storage/IO.hs
```diff
@@ -82,6 +84,11 @@ write :: FHandle -> Ptr Word8 -> Int64 -> IO Word32
 write h data' bytes = withOpenHandle "write" h $ \fd ->
     fromIntegral <$> Posix.fdWriteBuf fd data' (fromIntegral bytes)
 
+pwrite :: FHandle -> Ptr Word8 -> Int64 -> Word64 -> IO Word32
```
mrBliss: So the `Int64` is the number of bytes? Shouldn't that be a `Word64`?

I think you meant for the offset to be an `Int64`, because `FileOffset` is basically an `Int64`. Even then, what would a negative offset in a file mean? Like we did for `pread`, i would go for `Word64` (or `Word32` for the offset?) in both cases.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/FS/API.hs
```diff
@@ -262,6 +292,24 @@ hPutAll hasFS h = foldM putChunk 0 . BL.toChunks
...
+          -> BL.ByteString
+          -> AbsOffset
+          -> m Word64
+hPutAllAt hasFS h bs offset = fst <$> foldM putChunk (0, offset) (BL.toChunks bs)
+  where
+    putChunk :: (Word64, AbsOffset) -> BS.ByteString -> m (Word64, AbsOffset)
+    putChunk (written, offset') chunk = do
+      written' <- hPutAllStrictAt hasFS h chunk offset'
+      let !writtenTotal = written + (fromIntegral written')
+      let !newOffset = offset' + (fromIntegral written')
```
mrBliss: Not parentheses need here.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/FS/API.hs
```diff
@@ -273,3 +321,14 @@ hPut :: forall m h
...
+
+-- | Like 'hPut' but it does not change or depend on the file offset.
+-- @pwrite@ syscall is used internally.
+hPutAt :: forall m h
+     .  (HasCallStack, Monad m)
+     => HasFS m h
+     -> Handle h
+     -> Builder
+     -> AbsOffset
+     -> m Word64
```
mrBliss: There are misaligned w.r.t. the `::`.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -998,6 +1022,12 @@ tag = C.classify [
               f (Just n) = Just $ (BS.length bs) + n
               put' = Map.alter f (h, fp) put
           in Right $ tagPutTruncateGet put' truncated
+        PutAt (Handle h fp) bs _ | BS.length bs /= 0 ->
+          let
```
mrBliss: No need for a newline here.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -998,6 +1022,12 @@ tag = C.classify [
               f (Just n) = Just $ (BS.length bs) + n
               put' = Map.alter f (h, fp) put
           in Right $ tagPutTruncateGet put' truncated
+        PutAt (Handle h fp) bs _ | BS.length bs /= 0 ->
```
mrBliss: I prefer `not (BS.null bs)`

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/FS/StateMachine.hs
```diff
@@ -1091,6 +1121,10 @@ tag = C.classify [
           let f Nothing  = Just 0
               f (Just x) = Just $ x + 1
           in Right $ tagWriteWriteRead $ Map.alter f (h, fp) wr
+        PutAt (Handle h fp) bs _ | BS.length bs > 0 ->
```
mrBliss: I prefer `not (BS.null bs)`

ouroboros-consensus/src-unix/Ouroboros/Consensus/Storage/IO.hs
```diff
@@ -8,6 +8,7 @@ module Ouroboros.Consensus.Storage.IO (
     , read
     , pread
     , write
+    , pwrite
```
mrBliss: What about `src-win32`?

ouroboros-consensus/src-unix/Ouroboros/Consensus/Storage/IO.hs
kderme: I use Word64 for othe offset because AbsOffset is a Word64 wrapper. The number of bytes is Int64 because it's the same for write. But maybe this is a historical accident. The unix wrapper for write/pwrite gets a Word64 wrapper for the size, while the windows gets Word32. So I guess Word32 is the right choice for the size.

ouroboros-consensus/src-unix/Ouroboros/Consensus/Storage/IO.hs
kderme: Fixed now for win32 and tests pass

kderme:
> This is review of your first commit. I'll try to review the rest tomorrow.
> 
> I'm not inclined to re-expose the internals, we were happy to finally to hide them.

The third commit extends the tests and needs the internal structures. In this case, we can skip this commit

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -567,6 +575,17 @@ releaseResources rr keys releaser = do
...
 
+-- | Like 'withRegistry', but it only releases the resources of the registry
+-- if there is some exception.
+withRegistryOnError :: (IOLike m, HasCallStack)
+                    => (ResourceRegistry m -> m a)
+                    -> m a
+withRegistryOnError action =
+    bracketOnError unsafeNewRegistry closeRegistry $ \rr -> do
+      a <- action rr
+      void $ closeRegistryState rr
```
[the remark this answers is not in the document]
kderme: The registry is local, so we can close it when the api action is finished. `closeRegistryState` doesn't deallocate any resources though.

kderme:
> The local ResourceRegistry doesn't feel right, but I see why you went for it. I think we can simplify this away by using bracketOnError.

Yes definitely I can use bracketOnError for this, I just thought using ResourceRegistry makes the code cleaner.

> Tracking `_exceptionEncountered` in the state feels wrong. I prefer either making things more idempotent (`pwrite` + `AllowExisting`) and/or restoring the file system to be in sync with the state again in case of an exception. For example, in `modifyOpenState`, we could delete the newly created file in case it's not the new current file and an exception was thrown.

The `_exceptionEncountered` is mainly used because the write handle may be closed and we have to open a new one. We could reopen it when the exception is thrown, but this doesn't seem right: when the exception is thrown we should do as little as possible as cleanup and only release resources that may leak, not opening new files. Next calls that find `_exceptionEncountered = True` can reopen the file. I can't think of any other solution for this..

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
```diff
@@ -317,22 +349,41 @@ putBlockImpl :: forall m blockId. (IOLike m, Ord blockId)
...
              -> m ()
 putBlockImpl env@VolatileDBEnv{..} blockInfo@BlockInfo { bbid, bslot, bpreBid } builder =
-    modifyState env $ \hasFS@HasFS{..} st@InternalState{..} ->
-      if Map.member bbid _currentRevMap then do
-        traceWith _tracer $ BlockAlreadyHere bbid
-        return (st, ()) -- putting an existing block is a no-op.
-      else do
-        bytesWritten <- hPut hasFS _currentWriteHandle builder
-        updateStateAfterWrite hasFS st bytesWritten
+    withRegistryOnError $ \ rr ->
```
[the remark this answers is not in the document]
kderme: I'm only concerned that if the bracketOnError is nested in the modifyState action, there will always be some time where the bracketOnError has ended, but the internal state has not been updated yet. If we receive an async exception at this time, the handle we just opened will leak.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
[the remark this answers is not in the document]
kderme: I removed `void $ closeRegistryState rr`, there's no reason for it.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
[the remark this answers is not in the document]
mrBliss: For an alternative approach, see https://github.com/input-output-hk/ouroboros-network/commit/52d9738ec0cdd4e0dd96fff1eeec902335b91f16. No `ResourceRegistry` needed and updating the state is pure and disentangled from file-system operations.

kderme:
> For an alternative approach, see 52d9738. No ResourceRegistry needed and updating the state is pure and disentangled from file-system operations.

I understand we don't need this anymore, but is this the same to what I did? I see that the bracketOnError body, simply returns the internalState, without updating the MVar. If we get an async exception in this small time window (after the bracketOnError and before updating the MVar) the handle will leak. That's why I also nested the modifyState bracket inside the registry bracket and not the other way around.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
```diff
@@ -324,17 +324,18 @@ putBlockImpl env@VolatileDBEnv{ maxBlocksPerFile, tracer }
...
             (Index.lookup currentWriteId currentMap)
         fileBlockInfo = FileInfo.mkFileBlockInfo (BlockSize bytesWritten) bbid
-        fileInfo' = FileInfo.addBlock bslot currentWriteOffset fileBlockInfo fileInfo
+        AbsOffset offset = currentWriteOffset
+        fileInfo' = FileInfo.addBlock bslot offset fileBlockInfo fileInfo
         currentMap' = Index.insert currentWriteId fileInfo' currentMap
         internalBlockInfo' = InternalBlockInfo {
             ibFile         = currentWritePath
-          , ibBlockOffset  = currentWriteOffset
+          , ibBlockOffset  = offset
```
mrBliss: What if you change `BlockOffset` (in VolatileDB.Types) to `AbsOffset` too? I'm just wondering if that's a good change.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/FileInfo.hs
```diff
@@ -38,7 +39,7 @@ import           Ouroboros.Consensus.Storage.VolatileDB.Types
 -- | The internal information the db keeps for each file
 data FileInfo blockId = MkFileInfo {
       fLatestSlot :: !MaxSlotNo
-    , fContents   :: !(Map BlockOffset (FileBlockInfo blockId))
+    , fContents   :: !(Map Word64 (FileBlockInfo blockId))
```
kderme: I keep this `Word64` because this will be a big Map (1000-10000 entries) and all the wrapping and unwrapping could be costly. And also `FileInfo` is opaque.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Types.hs
```diff
@@ -129,8 +129,9 @@ sameUnexpectedError e1 e2 = case (e1, e2) of
 newtype BlockSize = BlockSize {unBlockSize :: Word64}
     deriving (Show, Generic, NoUnexpectedThunks)
 
--- | The offset at which a block is stored in a file.
-type BlockOffset = Word64
+-- | In case of a parser error, this is the offset at which the file should be
+-- truncated.
+type TruncateOffset = Word64
```
kderme: CBOR uses `Word64` for this offset and it seemed weird to go and wrap them with `AbsOffset`. I also think using a different name for this is a bit more expressive.

mrBliss:
Not needed anymore thanks to #1709 (blocks are added in a separate thread that won't get cancelled) and #1787 (expands on the idea of a temporary `ResourceRegistry` introduced in this PR).

PR #1689, opened 2020-02-24
Allow arbitrary actions during block production

edsko:


nfrisby:
My amendments to the commit were to satisfy Haddock: it was choking on the uses of `-- ^` within higher order functions. I made them normal comments instead.

PR #1506, opened 2020-01-27
Fix cloneBlockchainTime

nfrisby:
Fixes #1489. Fixes #1524.

Fixing Issue #1489 (let nodes lead when they join) and letting k vary in the
range [2 .. 10] since the dual-ledger tests do that now revealed several
Issues.

Issues in the library, not just the test infrastructure:

  * Issue #1505 -- `removeTxs` cannot use the fast path when validating after
    removing or else we might have dangling tx inputs references. Was fixed
    by #1565.

  * Issue #1511, bullet 1 (closed) -- The `Empty` cases in
    `prevPointAndBlockNo` were wrong. Recent PRs have addressed this: #1544 and
    #1589.

  * Issue #1543 (closed) -- A bracket in `registeredStream` was spoiled by an
    interruptible operation. Thomas' PR re-designed the vulnerability away. I
    think this is unrelated to the other changes; it was lurking and happened
    to pop up here just because I've been running hundreds of thousands of
    tests.

Issues only in the test infrastructure:

  * Issue #1489: let nodes lead when they join. This bug slipped in recently,
    when I added cloning of `BlockchainTime`s as part of the
    restarting/rekeying loop in the test infrastructure.

  * _PBFT reference simulator_. (Was masked by 1489.) Model competing 1-block
    chains in Ref.PBFT and use its results where applicable instead of the
    .Expectations module. Check that the PBFT threadnet and Ref.PBFT.simulate
    results agree on `Ref.Nominal` slots.

    The Ref.PBFT module had been making assumptions that were accurate given
    the guards in RealPBFT generators, given the `k >= n` regime. But outside
    of that regime, when the security parameter exceeds the node count, it
    wasn't enough. Also, it couldn't be compared against the PBFT threadnet
    because of those assumptions.

  * _PBFT reference simulator_. (Cascade of above.) Add
    `definitelyEnoughBlocks` to confirm the "at least k blocks in 2k slots"
    invariant in `genRealPBFTNodeJoinPlan`. The existing guards in the RealPBFT
    generators are intentionally insufficient by themselves; this way we can
    optimize them to avoid O(n^2) complexity without risking divergence from
    the `suchThat`s.

  * _Origin corner-case_. (Was masked by 1489.) Discard DualPBFT tests that
    forge in slot 0. The current `cardano-ledger-specs` doesn't allow for that.
    My hypothesis is that `cvsLastSlot` would need to be able to represent
    origin.

[...]

Issue #1489, opened 2020-01-23
Fix cloning of `TestBlockChaintime`


nfrisby:
PR #1202 introduced a bug in the test infrastructure that prevents nodes from forging during ~slot 0~ the slot in which they join.

The new `BlockhainTime` cloning introduced there does not allow the cloned `btime`'s `onSlotChange` from ticking with `s` as its argument if `s` is the slot in which the node joins the net.

This Issue is to fix that.

PR #1691, opened 2020-02-24
consensus: handle EBBs in rewindHeaderState

### LINKED ISSUE #1690: bug in `rewindHeaderState` triggered by EBBs
https://github.com/input-output-hk/ouroboros-network/blob/599afc730af5f8d64accc6ab1c8d05bdb27ee7d8/ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs#L200-L209

If `p` is an EBB with a proper block in the same slot, then the `annTipPoint t > p` is correct only half of the time: it will either correctly rewind the tips to the EBB or it will incorrectly also retain the proper block, depending on how the hashes happen to compare.

nfrisby:
Fixes #1690.

I believe this bug was masked by Issue #1489; I discovered it and confirmed that this patch fixes it on my related WIP branch. This PR does not include a repro for this bug, but that branch does, and PR #1506 will as soon as I'm able to update it with that WIP branch -- that's my current focus.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
@@ -192,6 +192,11 @@ castHeaderState HeaderState{..} = HeaderState{
     castSeq :: (a -> b) -> StrictSeq a -> StrictSeq b
     castSeq f = Seq.fromList . map f . toList
 
+-- | Rewind the header state
+--
+-- This involves 'rewindChainState', and so inherits its PRECONDITION that the
+-- target point must have been previously applied.
+--
 rewindHeaderState :: forall blk.
                      ( BlockSupportsProtocol blk
                      , Serialise (HeaderHash blk)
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
```diff
@@ -201,12 +201,13 @@ rewindHeaderState cfg p HeaderState{..} = do
...
     return $ HeaderState {
         headerStateChain  = chainState'
-      , headerStateTips   = Seq.dropWhileR rolledBack headerStateTips
+      , headerStateTips   = Seq.dropWhileR rollBack headerStateTips
       , headerStateAnchor = headerStateAnchor
       }
   where
-    rolledBack :: AnnTip blk -> Bool
-    rolledBack t = annTipPoint t > p
+    rollBack :: AnnTip blk -> Bool
```
nfrisby: The name change was accidental, but doesn't seem worth another force push :)

edsko: Did you see https://github.com/input-output-hk/ouroboros-network/issues/1690#issuecomment-590699588?

nfrisby: I did now. I've adopted it.

edsko:
@nfrisby feel free to `bors` this once you feel confident it solves the problem :)

nfrisby:
My big update to PR #1506 is still a work-in-progress, but I do think it is not "feature complete" with respect to PBFT and RealPBFT. In particular the commit on this PR passed hundreds of thousands of tests (though I can't be sure how many exercised it), including a couple repros derived from this PR's Issue.

So I'm going to merge. Thanks.

Issue #1659, opened 2020-02-18
Make ouroboros-consensus extra warning free


edsko:
Currently we enforce `-Wall` cleanliness, but we should enable

```
    ghc-options: -Wall
                 -Wcompat
                 -Wincomplete-uni-patterns
                 -Wincomplete-record-updates
                 -Wpartial-fields
                 -Widentities
                 -Wredundant-constraints
                 -Wmissing-export-lists
```

Issue #1637, opened 2020-02-13
Make `protocolSlotLengths` stateful


edsko:
At the moment, we have

```haskell
  protocolSlotLengths :: NodeConfig p -> SlotLengths
```

which assumes that the slot lengths are known _statically_. That is not the case, since the timing of the hard fork is not known statically. This therefore becomes ledger state dependent, which is a bit of a headache.

edsko:
Not required for the Shelley testnet, which will have a single slot length.

edsko:
Related: https://github.com/input-output-hk/ouroboros-network/issues/1205

edsko:
Rename `protocolSlotLengths` to `knownSlotLength`.

PR #1692, opened 2020-02-25
Remove redundant constraints

edsko:
This goes some way towards #1659, though the main reason for doing this
change now is that these redundant constraints were making it hard to
see the consequences of some of the changes I am making as part of
the work on the hard fork combinator (in particular, #1637).


edsko:
> Ah, I was looking forward to doing that slightly_smiling_face

Plenty of other warnings to remove still :)

Issue #1699, opened 2020-02-25
Make Immutable DB independent from EpochSize


edsko:
At the moment, the immutable DB stores blocks per epoch. Unfortunately, this introduces a dependency between the immutable DB and the ledger state, because the ledger state tells us when epoch sizes change. Instead we should give the immutable DB a fixed "chunk" parameter, with a precondition that _if_ there are EBBs present, they _must_ line up with the first slot in each chunk. Unfortunately, this means that the size of the files stored on disk will go down a factor of 10 or even 20 (depending on the choice of the Praos `f` parameter), but after discussing this with @dcoutts , we considered this an acceptable compromise.

This is blocking the work on the hard fork combinator; indeed, it is blocking the work at #1698 on #1637 / #1205 .

edsko:
WIP PR at https://github.com/input-output-hk/ouroboros-network/pull/1717.

edsko:
Closed in https://github.com/input-output-hk/ouroboros-network/pull/1750.

PR #1700, opened 2020-02-25
Avoid EpochInfo in ChainDB

edsko:
WIP. Addresses #1699.


edsko:
Now that I understand better what needs to happen, I will redo this from scratch as a series of incremental changes.

Issue #1155, opened 2019-10-22
Avoid use of `Tip Point`


edsko:
In various places in the consensus layer we use a type from `consensus`

```haskell
data Tip r = Tip !r | TipGen
  deriving (Show, Eq, Generic, NoUnexpectedThunks)
```

instantiated at `Point`, from `network`

```haskell
newtype Point block = Point
    { getPoint :: WithOrigin (Point.Block SlotNo (HeaderHash block))
    }
```

Now that `Point` uses `WithOrigin`, however, that is no longer reasonable, and we should just use `Point` directly.

mrBliss:
While we're at it, we should introduce a shorter name for `Point.Block SlotNo (HeaderHash block)`, e.g., `BlockPoint`, and use this to enforce that something cannot be genesis. For example, for some of the iterator bounds (`StreamToInclusive`) and for `getBlock` (we'll no longer need the `NoGenesisBlock` exception).

PR #1704, opened 2020-02-26
Replace `Tip` by `WithOrigin`

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/Storage/Common.hs
```diff
@@ -32,84 +31,53 @@ module Ouroboros.Consensus.Storage.Common (
...
-      0 -> return TipGen
-      1 -> Tip <$> decodeR
-      _ -> fail "decodeTip: invalid tag"
+encodeTip :: (r            -> Encoding)
+          -> (WithOrigin r -> Encoding)
+encodeTip = encodeWithOrigin
+
+decodeTip :: (forall s. Decoder s             r)
+          -> (forall s. Decoder s (WithOrigin r))
+decodeTip = decodeWithOrigin
```
mrBliss: Those should be removed

edsko: Will do so in subsequent PR, which will close #1155.

PR #1708, opened 2020-02-27
Check for unexpected thunks in TxSubmission.Inbound state

### LINKED ISSUE #1645: Avoid space leaks in TxSubmission.Inbound
Copying from https://github.com/input-output-hk/ouroboros-network/pull/1476#discussion_r369689613:

https://github.com/input-output-hk/ouroboros-network/blob/f5741efd37c5558d8a29d0d2b142e01f650a3d45/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs#L76

This record is the state of this server, so we should check whether this contains thunks to avoid space leaks. This record is not stored in a `StrictTVar` (or variant), but passed around. See
https://github.com/input-output-hk/ouroboros-network/blob/5c90d6e6e57f46f7fe22aa5fd657554493a71948/ouroboros-consensus/src/Ouroboros/Consensus/ChainSyncClient.hs#L786 
for an example on how to check for thunks in such a state. Slightly annoying: this code lives in `ouroboros-network`, while the thunks checking code lives in `ouroboros-consensus`. So something should be moved at some point.

intricate:
Closes #1645

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -42,6 +42,8 @@ import           Data.Void (Void)
 import           Data.Word (Word64)
 import           GHC.Generics (Generic)
 
+import           Cardano.Prelude (unsafeNoUnexpectedThunks)
+
 import           Network.TypedProtocol.Pipelined
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment (..))
 import qualified Ouroboros.Network.AnchoredFragment as AF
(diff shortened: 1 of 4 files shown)


stack.yaml
```diff
@@ -1,3 +1,4 @@
+# Something is wrong with the new cardano-prelude resolver @ e0c6e29
 resolver: https://raw.githubusercontent.com/input-output-hk/cardano-prelude/00487726c4bc21b4744e59d913334ebfeac7d68e/snapshot.yaml
```
intricate: For some reason this doesn't `stack build` when using `cardano-prelude@e0c6e29`. Looking into it.

mrBliss:
bors merge

PR #1717, opened 2020-02-27
Avoid use of EpochInfo within ChainDB (and actually within consensus at all)

edsko:
This is a second attempt at #1699 (the first was #1700), but got stuck on #1155. We've now solved that in #1706, but as a consequence this branch is once again horribly out of date and I will need to go again.


edsko:
I should remember to close the branch used for #1700 once we're sure all the changes there have an analogue in this PR.

edsko:
Superseded by #1750.

PR #1722, opened 2020-02-28
Avoid underscores in field names

### LINKED ISSUE #1675: Remove underscores from record fields
In various records we prefix fields with underscores (I must have gotten the habit from using `lens`), which is unpractical, because you don't get warnings when you no longer use them. We should get rid of these underscores and unify the prefixes. For example, `ImmutableDB`, `VolatileDB`, ...

mrBliss:
Fixes #1675.


intricate:
It's also great that you've removed a bunch of partial field accessors as part of these changes :+1:

mrBliss:
C'mon bors, stop slowing us down

PR #1728, opened 2020-03-02
Added ghc-tags-plugin

coot:
ghc-tags-plugin is a GHC compiler plugin which inspects parsed tree and
writes tags file (currently only in vim-format, emacs format will come
soon).  It leaves the parsed tree unchanged.  The plugin is much better
in finding tags than let say `hasktags` (inspecting syntax try vs
pattern matching), especially for type level heavy code
(typed-protocols, ourobors-network protocol implementations or
ouroboros-consensus).

This patch adds `tags` cabal flag to all cabal files which enabled the
plugin.  Then one can configure each local package in
`cabal.package.local` to use it, e.g.

```
package typed-protocols
  flags: +tags
  ghc-options: -fplugin=Plugin.GhcTags
```

Disclaimer: I am the author of the plugin :)

https://hackage.haskell.org/package/ghc-tags-plugin-0.1.0.1

A better option would be to write a nix expression which does that
without modifying all the cabal files, but that's more complex.

### DIFF
--- a/Win32-network/Win32-network.cabal
+++ b/Win32-network/Win32-network.cabal
@@ -19,6 +19,11 @@ flag demo
   description: Build the named pipe demos
   default:     False
 
+flag tags
+  Description: Enable ghc-tags-plugin to generate tags file
+  Manual: True
+  Default: False
+
 library
   hs-source-dirs:      src
   if os(windows)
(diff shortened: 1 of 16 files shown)


coot:
Closing as there is a way to configure it in `cabal.project.local` file.

Issue #1729, opened 2020-03-02
mux demuxing (IO) test case failure


karknu:
https://hydra.iohk.io/build/1824338/nixlog/3

karknu:
Fixed

PR #1730, opened 2020-03-02
Fix a race condition in demuxing (IO)

karknu:
For 0 length SDUs there was a risk that the responder application
finished before the ingress thread parsed the header and threw an exception.

### DIFF
--- a/network-mux/test/Test/Mux.hs
+++ b/network-mux/test/Test/Mux.hs
@@ -101,7 +101,7 @@ smallMiniProtocolLimit = 16*1024
 
 activeTracer :: forall m a. (MonadSay m, Show a) => Tracer m a
 activeTracer = nullTracer
---activeTracer = showTracing sayTracer
+--activeTracer = showTracing _sayTracer
 
 _sayTracer :: MonadSay m => Tracer m String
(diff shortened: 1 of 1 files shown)


karknu:
Fixes #1729.

network-mux/test/Test/Mux.hs
```diff
@@ -101,7 +101,7 @@ smallMiniProtocolLimit = 16*1024
 
 activeTracer :: forall m a. (MonadSay m, Show a) => Tracer m a
 activeTracer = nullTracer
---activeTracer = showTracing sayTracer
+--activeTracer = showTracing _sayTracer
```
coot: no need to change this line ;)

karknu: Agreed, but it is changed to match the current name of the debug function.

coot: hmm, I don't like this upstream change: `_names` are reserved for not used identifiers.

karknu: The function is called `_sayTracer` and it is unused since active tracer is commented out.
But ifdefs is another alternative...

PR #1734, opened 2020-03-03
Give the ThreadNet tests their own EpochInfo

edsko:
They are currently using `nodeEpochSize` from `RunNode`, but that function will soon disappear (consensus will not really need to be aware of epoch boundaries at all anymore).

This will need to be revised again once we have the hard fork combinator, in which case `EpochInfo` should be derived from the ledger state instead.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/General.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/General.hs
@@ -57,7 +57,8 @@ import           Ouroboros.Consensus.Util.Orphans ()
 import           Ouroboros.Consensus.Util.Random
 import           Ouroboros.Consensus.Util.RedundantConstraints
 
-import           Ouroboros.Consensus.Storage.Common (EpochNo)
+import           Ouroboros.Consensus.Storage.Common
+import           Ouroboros.Consensus.Storage.EpochInfo
 
 import           Test.ThreadNet.Network
 import           Test.ThreadNet.TxGen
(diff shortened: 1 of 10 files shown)


ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/General.hs
```diff
@@ -200,6 +201,9 @@ runTestNetwork ::
      , Show (LedgerView (BlockProtocol blk))
      )
   => TestConfig
+  -> EpochSize
+  -- ^ Temporary: until we start testing the hard fork combinator, we just
+  -- support a single 'EpochSize'. See also comments for 'tnaEpochInfo'.
```
mrBliss: This should be part of `TestConfig` or `TestConfigBlock`.

edsko: `TestConfigBlock`  is block specific. `TestConfig`  would make sense in _principle_, but it gets generated independent from `k`, and so I can't produce a value for `EpochSize` there.

mrBliss: Solution: `TestConfig` should include `SecurityParam`. But okay, let's not do that in this PR.

edsko: https://github.com/input-output-hk/ouroboros-network/issues/1735

edsko: But see https://github.com/input-output-hk/ouroboros-network/issues/1735#issuecomment-593822774.

edsko:
As I was rebasing my imm DB changes on top of this, realized that this is still awkward. The threadnet tests intialize a chain DB, but they want a chunk info -- but all we have is an epoch info. I think I need topass the epoch size down instead (until such time we have the hard fork combinator).

PR #1739, opened 2020-03-03
Update q-s-m to allow bracketing

### LINKED ISSUE #1715: Avoid `PropertyM` in tests
So that we can use `bracket` and `bracket`-like functions. This will require an upstream patch; I've already submitted one at https://github.com/advancedtelematic/quickcheck-state-machine/pull/358 but it needs some adjustment.

mrBliss:
Fixes #1715

* Update dependency on quickcheck-state-machine

* Use `runCommands'` in `*.StateMachine`

  The new `runCommands'` command no longer lives in `PropertyM`, which means we can finally do proper bracketing.

  We don't use each bracketed function where possible, as in some cases we do some manual closing and reopening.


ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -1297,33 +1298,33 @@ smUnused =
...
 prop_sequential :: Property
 prop_sequential = forAllCommands smUnused Nothing $ \cmds -> QC.monadicIO $ do
-    (hist, prop) <- test cmds
+    (hist, prop) <- QC.run $ test cmds
     -- TODO label tags
     prettyCommands smUnused hist prop
   where
     test :: QSM.Commands (At Cmd Blk IO) (At Resp Blk IO)
-         -> QC.PropertyM IO
+         -> IO
```
edsko: Yay :)

PR #1740, opened 2020-03-03
Mini Protocol Mux Limts

coot:
This PR consists of two major parts: estimations of inbound queue size for each mini-protocol, plus a refactoring to gather parameters that are needed to set them in one central place.

* calculations of limits for each protocol (every one in each commit)
* when we have those we can add 'MiniProtocolParameters' which take all the required data to compute the limits: chain sync pipelineing parameters, block fetch pipelineing & tx submission unacknowledged `TxIds` (from which one can compute upper bound for pipelining).

`MiniProtocolParameters` are set by `mkNodeArgs` function in `ouroboros-consensus` and passed down the stack.  `NodeArgs` can be change by the `cardano-node`, this makes us safe from changing the parameters at some level without making required changes in `ouroboros-network` - which potentially could end-up badly.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
@@ -14,6 +14,7 @@ module Ouroboros.Network.NodeToNode (
     nodeToNodeProtocols
   , NodeToNodeProtocols (..)
   , MiniProtocolParameters (..)
+  , defaultMiniProtocolParameters
   , NodeToNodeVersion (..)
   , NodeToNodeVersionData (..)
   , DictVersion (..)
@@ -165,6 +166,14 @@ data MiniProtocolParameters = MiniProtocolParameters {
       -- ^ maximal number of unacked tx (pipelineing is bounded by twice this
       -- number)
     }
+
+defaultMiniProtocolParameters :: MiniProtocolParameters
+defaultMiniProtocolParameters = MiniProtocolParameters {
+      chainSyncPipelineingLowMark  = 200
+    , chainSyncPipelineingHighMark = 300
+    , blockFetchPipelineingMax     = 100
+    , txSubmissionMaxUnacked       = 10
+  }
 
 -- | Make an 'OuroborosApplication' for the bundle of mini-protocols that
 -- make up the overall node-to-node protocol.
--- a/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
@@ -13,6 +13,7 @@
 module Ouroboros.Network.NodeToNode (
     nodeToNodeProtocols
   , NodeToNodeProtocols (..)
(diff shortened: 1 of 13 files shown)


coot:
This PR addresses #575.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -160,6 +161,75 @@ nodeToNodeProtocols chainSync
...
       }
     ]
+  where
+    -- TODO: these limits depends on various parameters instantiated by
+    -- 'Ouroboros.Consensus.Node'.  We should parametrize over them, but these
+    -- limits must only increase.
+    chainSyncProtocolLimits :: MiniProtocolLimits
+    chainSyncProtocolLimits =
+      MiniProtocolLimits {
+          -- chain sync has two potencially large messages:
```
njd42: potencially  -> potentially

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -165,7 +165,8 @@ nodeToNodeProtocols chainSync
     -- TODO: these limits depends on various parameters instantiated by
     -- 'Ouroboros.Consensus.Node'.  We should parametrize over them, but these
```
njd42: parametrize -> parameterize (or -....se - I'm UK,...)

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -252,19 +292,77 @@ nodeToNodeProtocols chainSync
...
+          -- the 'TxId' without the size of 'Tx' in bytes).  So the ingress
+          -- queue of 'txSubmissionOutbound' is bounded by the ingress side of
+          -- the 'txSubmissionInbound'
+          --
+          -- Currently the value of 'txSubmissionMaxUnacked' is '100', for
+          -- which the upper bound is `100 * (44 + 65_540) = 6_558_400`, we add
+          -- 10% as a safety margin.
+          --
+          maximumIngressQueue = addSafetyMargin $
+              fromIntegral txSubmissionMaxUnacked * (44 + 65540)
```
njd42: for consistency (given you've turned on the feature) you could make it `(44 + 65_540)`

coot:
>  I did get my calculator out to check no schoolboy arithmetical errors - I found none.
Thanks for doing that!

I managed to sync the node from mainnet with this patch without problems.

ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
coot: o really, in UK its spelled this way, I wouldn't think.

coot: This comment needs to be removed.  I fixed it in a later patch.

coot:
@karknu I created an issue for that: #1802

PR #1741, opened 2020-03-03
Attempt to document the hard fork transition

edsko:


ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+`chain-and-ledger/executable-spec/src/PParams.hs`).
+
+The hard fork from Shelley to its successor (Goguen?) will be initiated
+once the hard fork component of this version gets incremented.
+
+### The update mechanism for the protocol version
+
+The update mechanism in Shelley is simpler than it is in Byron. There is no
+distinction between votes and proposals: to "vote" for a proposal one merely
+submits the exact same proposal. There is also no separate endorsement step.
```
nfrisby: > There is also no separate endorsement step.

Well -- as covered in the next section -- there is something similar to endorsement, but it's only required for hard-forks and it happens before voting instead of after.

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+The procedure is as follows:
+
+1. Proposals (change parameter `x` to value `y` are submitted to the chain)
+2. "Voting" (submitting of proposals) ends `2k` slots before the end of the
+   epoch (that `2k` might need to be adjusted pending more detailed analysis
+   from the researchers -- the spec records this as a single specification
+   parameter (not protocol parameter!) called `slotsPrior`).
+3. By the end of the epoch, if a majority (for some definition of majority) is
+   reached for _all_ proposed parameters, then _all_ parameters take their new
+   value at the start of the next epoch. If no majority is reached on _any one_
```
nfrisby: > If no majority is reached on _any one_ of the parameters

This is an aside. But I suspect the wording here is slightly inaccurate or else that we have an additional validation rule that prevents contemporary proposals from altering the same parameter. Specifically: if two proposals adjust the same parameter and one of them is confirmed but the other is not, then that unconfirmed proposal would still spoil all the proposals, right?

nfrisby:
LGTM, but I'm leaving Approval to someone from a different team, for the sake of broad awareness and agreement.

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+a convention _could_ be established that the protocol version must change if
+any of the parameters do; but the specification itself does not mandate this.
+
+### Initiating the hard fork
+
+The timing of the hard fork in Shelley is different to the one in Byron;
+in Byron, we _first_ vote and then wait for people to get ready, in Shelley
+it is the other way around.
+
+(Core?) node operators will want to know that a (significant?) majority of the
```
JaredCorduan: yes, I think `Core` is appropriate there, since only they can vote during Shelley. I also think "significant" is appropriate, but if you want a specific number we would have to ask around.

ouroboros-consensus/docs/HardFork.md
JaredCorduan: I think it is correct as written. "proposals" and "votes" are really the same thing in Shelley. Each core node is allowed to include a update proposal during the voting period (any time before "2k" slots from the end of the epoch). either there are at least five proposals that agreed exactly, or there is not.

JaredCorduan: in particular, if four nodes want `{foo = 1, bar = 2}` and three nodes `{foo = 1}`, there is _no consensus_ and nothing is changed.

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
+# Details of the hard fork transition
+
+This document attempts to describe the details of the hard fork transition
+from Byron to Shelley, and from Shelley to future versions of the ledger.
+
+## Byron
+
+The Byron specification can be found at
+https://hydra.iohk.io/job/Cardano/cardano-ledger-specs/byronLedgerSpec/latest/download-by-type/doc-pdf/ledger-spec .
```
dnadales: Maybe use a markdown link here?

dnadales:
```suggestion
The Byron specification can be found [here](https://hydra.iohk.io/job/Cardano/cardano-ledger-specs/byronLedgerSpec/latest/download-by-type/doc-pdf/ledger-spec).
```

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+adoptedProtocolVersion :: ProtocolVersion
+```
+
+in the `State` type from `Cardano.Chain.Update.Validation.Interface`.
+
+This protocol version is a three-tuple `major`, `minor`, `alt`. The Byron
+specification does not provide any semantic interpretation of these components.
+By convention (outside of the purview of the Byron specification), the hard fork
+is initiated the moment that the `major` component reaches a predefined,
+hardcoded, value.
```
dnadales:
```suggestion
is initiated the moment that the `major` component of `adoptedProtocolVersion`
reaches a predefined, hardcoded, value.
```

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+block size), except that in the case of a hard fork, we care only about changing
+the `ProtocolVersion`, and not any of the parameters themselves.
+
+The general mechanism for updating protocol parameters in Byron is as follows:
+
+1. A protocol update _proposal_ transaction is created. It proposes new values
+   for some protocol parameters and a greater _protocol version_ number as an
+   identifier. There cannot be two proposals with the same version number.
+
+2. Genesis key delegates can add _vote_ transactions that refer to such a
```
dnadales: Maybe stress that **only** genesis keys delegates can vote.

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+"endorsement transaction", but instead restarts the node (probably after a
+software update that makes the node ready to support the hard fork) with a new
+protocol version (as part of a config file or command line parameter), which
+then gets included in the blocks that the node produces (this value is part of
+the static `ByronConfig`: `byronProtocolVersion`).
+
+(Note that a node restart is necessary for _any_ change to a protocol parameter,
+even though most parameters do not require any change to the software at all.)
+
+### Software version
```
dnadales: Maybe rename this to `Header software version` lest people will think we ignore the software version in the spec, whereas we only ignore the software version in the block header.

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+Similar to the Byron ledger, the Shelley ledger provides a "current protocol
+version", but it is a two-tuple (not a three-tuple), containing only a
+`hard fork` component and `soft fork` component:
+
+```haskell
+_protocolVersion :: (Natural, Natural)
+```
+
+in `PParams` (currently, module `PParams` in
+`chain-and-ledger/executable-spec/src/PParams.hs`).
```
dnadales:
```suggestion
[`chain-and-ledger/executable-spec/src/PParams.hs`](https://github.com/input-output-hk/cardano-ledger-specs/blob/93ff3480b9585bf1b7a9bd4013d7baed48135521/shelley/chain-and-ledger/executable-spec/src/PParams.hs#L92)).
```

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+(Core?) node operators will want to know that a (significant?) majority of the
+core nodes is ready (supports the hard fork) before initiating it. To make this
+visible, Shelley blocks contain a protocol version. This is not related to the
+current protocol version as reported by the ledger state (`_protocolVersion` as
+discussed in the previous section), but it is the _maximum_ protocol version
+that the node which produced that block can support.
+
+Once we see blocks from all or nearly all core nodes with the `hard fork`
+component of their protocol version equal to the post-hard-fork value, nodes
+will submit their votes to initiate the hard fork.
```
dnadales: This is not entirely clear to me. What does submitting a vote means here? Submitting a proposal with the version change?

ouroboros-consensus/docs/HardFork.md
nfrisby: @JaredCorduan It sounds like you're saying that, in Shelley:

  * Each node is allowed per epoch to propose zero or one udpates to the parameter values, not two or more proposals.

  * At an epoch transition, an update is only adopted if every node had proposed that same exact update "early enough" before the transition.

Is that right? Disclaimer: I haven't yet looked at the Shelley spec; I've only been testing Byron as of yet, my brain's bandwith, and so on.

edsko: > Each node is allowed per epoch to propose zero or one udpates to the parameter values, not two or more proposals.

Not sure which distinction you're drawing here?

nfrisby: Disclaimer: This is a pretty low-level concern, but I think it's a reasonable interpretation of the current text.

Consider this scenario. There are 3 nodes.

  * Node A proposes `foo = 1`.
  * Node B proposes `foo = 1`.
  * Node C proposes `foo = 2`.

If 2/3 is a sufficient majority, then the text as written suggests that `foo = 1` will be adopted. The text currently says "If no majority is reached on any one of the parameters". Even though there is a failed proposal for `foo`, there is also a confirmed proposal for `foo`; so "every one of the parameters" that has reached a majority, but not "every one of the proposals" has.

If I correctly interpreted Jared's comment and unanimity is required, then the same sort of scenario would have to look like this.

  * Node A proposes `foo = 1`.
  * Node A also proposes `foo = 2`.
  * Node B proposes `foo = 1`.
  * Node C proposes `foo = 1`.

But if I also correctly interpreted his comment to mean both unanimity is required and also that the same node can't put out more than one proposal in a given epoch, then this kind of scenario is impossible and the current wording in the text is ultimately fine.

JaredCorduan: @nfrisby - a majority is required, not unanimity. each core node can provide multiple updates during each epoch, but the last one overrides any previous ones. So in your last example, `foo = 1` is in fact adopted, since it gets 2/3 votes.

nfrisby: Aha! I was wrong on both accounts :). I think we should revise this paragraph. Edsko and I didn't have a correct understanding of Shelley protocol parameter updates when we wrote it.

I've now skimmed the Shelley spec for this purpose. But the pdf notation puns on the exact types pretty heavily for the sake of brevity, so I've turned to the code to find the details. As of commit e09a3efa9f428de474384856c0311088b1503819, I see the following behavior and code snippets.

[There is a tl;dr at the bottom. This comment got long.]

The `CHAIN` rule starts with an empty `PPUpdate Map.empty` as a field in the `UTxOState`.

```
newtype PPUpdate crypto = PPUpdate (Map (GenKeyHash crypto) PParamsUpdate)

newtype PParamsUpdate = PParamsUpdate { ppmSet :: Set Ppm }
  deriving (Show, Eq, NoUnexpectedThunks)

data Ppm = MinFeeA Integer
  | MinFeeB Natural
  | MaxBBSize Natural
  | ...
  deriving (Show, Ord, Eq, Generic)
```

So, at all times, we have a function isomorphic to `GenKeyHash crypto -> Maybe
(Set Ppm)`.

I don't see anywhere in `shelley/chain-and-ledger/executable-spec/src` that indicates how a node would update its `Set Ppm`, how it would "emit" a `Ppm`. Is there a separate repository that has code to do that? Or is that not yet implemented? I only see the `PPUP` rule, but that takes a whole chain-level `PPUpdate` as its signal, whereas I'm trying to find what it looks like from the perspective of a lone node (operator).

In the `EPOCH` rule, we count up to see how many genesis keys have the same set of `Ppm`s.

```
votedValuePParams
  :: PPUpdate crypto
  -> PParams
  -> Int
  -> Maybe PParams
votedValuePParams (PPUpdate ppup) pps quorumN =
  let
    incrTally vote tally = 1 + Map.findWithDefault 0 vote tally
    votes = Map.foldr
              (\vote tally -> Map.insert vote (incrTally vote tally) tally)
              (Map.empty :: Map PParamsUpdate Int)
              ppup
    consensus = Map.filter (>= quorumN) votes
  ...
```

As an example: if a node "emitted" (however that happens) both `MaxBBSize 100`
and `MaxBBSize 200` in the same epoch it looks like that would count as a
different proposal than either just `MaxBBSize 100` or `MaxBSize 200` without
the other.

And if enough nodes emitted both `MaxBBSize 100` and `MaxBBSize 200`, then it
looks like that pair of seemingly conflicting `Ppm`s would -- as a whole
`PParamsUpdate` -- meet quorum, and when applied to the parameters, it would be
equiavlent to `MaxBBSize 200` (due to semantics of `Set.foldr` and the derived `Ord Ppm`).

```
updatePParams :: PParams -> PParamsUpdate -> PParams
updatePParams ppms (PParamsUpdate up) = Set.foldr updatePParams' ppms up
 where
  updatePParams' (MinFeeA               p) pps = pps { _minfeeA = p }
  updatePParams' (MinFeeB               p) pps = pps { _minfeeB = p }
  updatePParams' (MaxBBSize             p) pps = pps { _maxBBSize = p }
  ...
```

*tl; dr*

In summary, that all seems like many details and corner-cases to try to explain in this document's paragraph. Also, this paragraph focuses on a node as the primary actor, and I couldn't yet determine what this part of the Shelley interface looks like from a lone node's perspective. E.G. Will there be a transaction type that emits a single `Ppm`? Or is it less granular than that?

P.S. - Maybe it's too late or maybe it's already been considered, but perhaps replace `Set Ppm` by a more precise "dependent map"-ish type? It wouldn't be too bad given that the parameters and their types are all statically known.

nfrisby: @edsko, FYI: @JaredCorduan and I had a call, and he clarified things for me. As a result, I wrote commit 2f9b7bbe42ccdd71577a03f1bc7a2a47601aff49 which you can cherry-pick/squash onto this PR if you'd like.

It rewrites this bullet list in the `HardFork.md`.

```
* 2f9b7bbe - (HEAD -> nfrisby/amend-hardforkdocs, origin/nfrisby/amend-hardforkdocs) HardFork.md: amend summary of Shelley protocol parameter update proposals (4 minutes ago) <Nicolas Frisby>
```

JaredCorduan: looks great @nfrisby , thanks!

it would indeed be great to replace the `Set Ppm` by a "dependent map"-ish type.

ouroboros-consensus/docs/HardFork.md
```diff
@@ -0,0 +1,211 @@
...
+The update mechanism in Shelley is simpler than it is in Byron. There is no
+distinction between votes and proposals: to "vote" for a proposal one merely
+submits the exact same proposal. There is also no separate endorsement step.
+The procedure is as follows:
+
+1. Proposals (change parameter `x` to value `y` are submitted to the chain)
+2. "Voting" (submitting of proposals) ends `2k` slots before the end of the
+   epoch (that `2k` might need to be adjusted pending more detailed analysis
+   from the researchers -- the spec records this as a single specification
+   parameter (not protocol parameter!) called `slotsPrior`).
```
mrBliss, line 157: This is out of date, it's now `3k/f` instead of `2k` and `slotsPrior` is now called `stabilityWindow`, see https://github.com/input-output-hk/cardano-ledger-specs/pull/1424.

ouroboros-consensus/docs/HardFork.md
edsko: Added a forward reference.

ouroboros-consensus/docs/HardFork.md
edsko: Ok, have cherry-picked that commit. Thanks :)

ouroboros-consensus/docs/HardFork.md
edsko: Will fix that once I incorporate this into the consensus TR.

edsko: Tssh. Tim Berners-Lee would wince if he saw it :roll_eyes: :grimacing:

ouroboros-consensus/docs/HardFork.md
edsko: It's actually _twice_ that :).

ouroboros-consensus/docs/HardFork.md
edsko: Yup, indeed. Will reword.

edsko:
bors merge

edsko:
bors merge

PR #1748, opened 2020-03-04
Introduce HasTxs

### LINKED ISSUE #1745: Provide a way to extract the transactions from a block
Provide an abstract way to extract the transactions out of a block.

Add `extractTxs :: blk -> [GenTx blk]` to `ApplyTx`, or, alternatively, add a new class for this, like `HasTxId`.

This class can also replace the following function:
https://github.com/input-output-hk/ouroboros-network/blob/327a4e3f4c7bc13d29b849307fd887df2bb0ed13/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/General.hs#L282

mrBliss:
Fixes #1745.


mrBliss:
@deepfire Can you check whether this type class satisfies your need?

PR #1749, opened 2020-03-05
More intelligent transaction requests

### LINKED ISSUE #1161: More intelligent transaction requests

intricate:
Closes #1161.

Implements the [new plan](https://github.com/input-output-hk/ouroboros-network/issues/1161#issuecomment-594423279).

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts        #-}
+{-# LANGUAGE OverloadedStrings       #-}
 {-# LANGUAGE RankNTypes              #-}
 {-# LANGUAGE ScopedTypeVariables     #-}
 {-# LANGUAGE StandaloneDeriving      #-}
@@ -15,6 +16,9 @@ module Ouroboros.Consensus.Mempool.API (
   , ApplyTx(..)
   , HasTxId(..)
   , GenTxId
+  , MempoolAddTxResult (..)
+  , isTxAddedOrAlreadyInMempool
+  , isTxRejected
   , MempoolSize (..)
   , TraceEventMempool(..)
   , HasTxs(..)
@@ -134,8 +138,9 @@ data Mempool m blk idx = Mempool {
       -- 1. A list containing the following transactions:
       --
       --    * Those transactions provided which were found to be valid, along
-      --      with 'Nothing' for their accompanying @Maybe (ApplyTxErr blk)@
-      --      values. These transactions are now in the Mempool.
+      --      with 'MempoolTxAdded' for their accompanying
+      --      'MempoolAddTxResult' values. These transactions are now in the
+      --      Mempool.
       --    * Those transactions provided which were found to be invalid,
       --      along with their accompanying validation errors. These
       --      transactions are not in the Mempool.
@@ -175,7 +180,7 @@ data Mempool m blk idx = Mempool {
       -- blockchain.
       --
       tryAddTxs      :: [GenTx blk]
-                     -> m ( [(GenTx blk, Maybe (ApplyTxErr blk))]
+                     -> m ( [(GenTx blk, MempoolAddTxResult blk)]
                           , [GenTx blk]
                           )
 
@@ -221,6 +226,28 @@ data Mempool m blk idx = Mempool {
     , zeroIdx        :: idx
     }
 
+-- | The result of attempting to add a transaction to the mempool.
+data MempoolAddTxResult blk
+  = MempoolTxAdded
+    -- ^ The transaction was added to the mempool.
+  | MempoolTxAlreadyInMempool
+    -- ^ The transaction already exists within the mempool and therefore could
+    -- not be added.
+    --
+    -- Note that we don't treat this like an error/rejection case.
+  | MempoolTxRejected !(ApplyTxErr blk)
+    -- ^ The transaction was rejected and could not be added to the mempool
+    -- for the specified reason.
+
+isTxAddedOrAlreadyInMempool :: MempoolAddTxResult blk -> Bool
+isTxAddedOrAlreadyInMempool MempoolTxAdded            = True
+isTxAddedOrAlreadyInMempool MempoolTxAlreadyInMempool = True
+isTxAddedOrAlreadyInMempool _                         = False
+
+isTxRejected :: MempoolAddTxResult blk -> Bool
+isTxRejected (MempoolTxRejected _) = True
+isTxRejected _                     = False
+
 -- | Wrapper around 'implTryAddTxs' that blocks until all transaction have
 -- either been added to the Mempool or rejected.
 --
@@ -234,7 +261,7 @@ addTxs
   :: forall m blk idx. (MonadSTM m, ApplyTx blk)
   => Mempool m blk idx
   -> [GenTx blk]
-  -> m [(GenTx blk, Maybe (ApplyTxErr blk))]
+  -> m [(GenTx blk, MempoolAddTxResult blk)]
 addTxs mempool = \txs -> do
     (processed, toAdd) <- tryAddTxs mempool txs
     case toAdd of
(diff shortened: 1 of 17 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -216,6 +232,53 @@ data Mempool m blk idx = Mempool {
...
+isTxAdded TxAdded = True
+isTxAdded _       = False
+
+isTxRejected :: MempoolAddTxResult blk -> Bool
+isTxRejected (TxRejected _) = True
+isTxRejected _              = False
+
+-- | An error encountered when attempting to add a transaction to the mempool.
+data MempoolAddTxError blk
+  = TxApplicationError !(ApplyTxErr blk)
```
intricate: Is the word "application" misleading?

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -216,6 +232,53 @@ data Mempool m blk idx = Mempool {
     , zeroIdx        :: idx
     }
 
+-- | The result of attempting to add a transaction to the mempool.
+data MempoolAddTxResult blk
+  = TxAdded
+    -- ^ The transaction was added to the mempool.
+  | TxRejected !(MempoolAddTxError blk)
```
intricate: Should these constructors be prefixed with `Mempool`?

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -472,6 +522,11 @@ internalStateFromVR vr = IS {
       , vrLastTicketNo
       } = vr
 
+    txIdSet = foldl'
+      (\acc tx -> Set.insert (txId tx) acc)
+      Set.empty
+      vrValid
```
intricate: I could implement this more efficiently by adding a new field to `ValidationResult` and appending to it whenever `vrValid` is appended to.

nfrisby:
I rebased my wip branch onto commit 8cd528bb, and I no longer see the huge number of transactions. Great!

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
mrBliss: Yeah, that's a good idea, it brings `ValidationResult` and `InternalState` more in sync again.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -252,13 +292,61 @@ txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
...
+    -- are already in the mempool and return the updated 'ServerState'.
+    --
+    -- This enables us to effectively filter out transactions that we don't
+    -- need to bother requesting from the client since they're already in the
+    -- mempool.
+    --
+    acknowledgeTxIdsInMempool :: ServerState txid tx
+                              -> MempoolSnapshot txid tx idx
+                              -> ServerState txid tx
+    acknowledgeTxIdsInMempool st MempoolSnapshot{mempoolHasTx} = do
```
mrBliss: This function is not-monadic, so I wouldn't use `do` here.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -192,16 +226,22 @@ txSubmissionInbound _tracer maxUnacked TxSubmissionMempoolWriter{..} =
...
+      -- transactions again in the future.
+      let st' = st {
         requestedTxIdsInFlight = requestedTxIdsInFlight st - reqNo,
         unacknowledgedTxIds    = unacknowledgedTxIds st
                               <> Seq.fromList (map fst txids),
         availableTxids         = availableTxids st
                               <> Map.fromList txids
       }
+      mpSnapshot <- atomically mempoolGetSnapshot
+      serverIdle n (acknowledgeTxIdsInMempool st' mpSnapshot)
```
mrBliss: I'm sure you and @dcoutts have discussed this already, but just checking again: is this the right and *only* place to look at the Mempool?

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
mrBliss: Hmm, for consistency, yeah.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -216,6 +232,53 @@ data Mempool m blk idx = Mempool {
...
+isTxAdded :: MempoolAddTxResult blk -> Bool
+isTxAdded TxAdded = True
+isTxAdded _       = False
+
+isTxRejected :: MempoolAddTxResult blk -> Bool
+isTxRejected (TxRejected _) = True
+isTxRejected _              = False
+
+-- | An error encountered when attempting to add a transaction to the mempool.
+data MempoolAddTxError blk
```
mrBliss: The type Edsko and I suggested in https://github.com/input-output-hk/ouroboros-network/issues/1161#issuecomment-594423279 intentionally did not classify "already in Mempool" as an error, as it's not really an error. For [example](https://github.com/input-output-hk/ouroboros-network/issues/1161#issuecomment-595093203), the wallet will resubmit transactions that didn't make it into the last block via the `LocalTxSubmissionServer`. The wallet doesn't know that the transaction is still in the Mempool, so we shouldn't treat that case as an error (think of it as a trace message severity). So go for something like:

```haskell
data MempoolAddTxResult blk
  = MempoolTxAdded
  | MempoolTxRejected (ApplyTxErr blk)
  | MempoolTxAlreadyInMempool
```

Also important: this new `MempoolAddTxError` type replaces the existing `ApplyTxError` type in the `LocalTxSubmission` protocol codec, so this breaks network compatibility, which we have to plan carefully. The 3-constructor approach doesn't break network compatibility.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/LocalTxSubmission/Server.hs
```diff
@@ -32,9 +32,11 @@ localTxSubmissionServer tracer mempool =
         traceWith tracer $ TraceReceivedTx tx
         res <- addTxs mempool [tx]
         case res of
-          [(_tx, mbErr)] -> return (mbErr, server)
+          [(_tx, addTxRes)] -> case addTxRes of
+            TxAdded             -> return (Nothing, server)
+            TxRejected addTxErr -> return (Just addTxErr, server)
```
mrBliss: So `MempoolTxAlreadyInMempool` would also result in `(Nothing, server)`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
intricate: In my discussions with Duncan, he noted that this seemed like the right place to do it, yeah. I also believe this is where it makes the most sense as well.

PR #1750, opened 2020-03-05
Make imm DB epoch independent version 3.0

edsko:
This is my third attempt at doing this, but it's now almost done. Only things left to do:

- [x] Update the immutable DB QSM tests to also test with a `ChunkSize` that doesn't support EBBs
- [x] Check for mentions of `epoch` that should be renamed to `chunk`.


edsko:
I'm not adding a migration to rename the chunk files; see https://github.com/input-output-hk/ouroboros-network/issues/1755.

PR #1751, opened 2020-03-05
Don't bind to unix sockets

karknu:
As a client we should not bind to unix sockets.


coot:
Related to https://github.com/input-output-hk/cardano-db-sync/issues/9

PR #1758, opened 2020-03-06
Remove constantSubscriptionTarget

karknu:
constantSubscriptionTarget creates an infinite stream of identical
subscription targets which causes subscriptionLoop.innerLoop to
livelock. Replace usage with a listSubscriptionTarget with a single
element instead.


karknu:
Fixes #1756

coot:
Very good @karknu!

PR #1759, opened 2020-03-07
Make mux-demo compile

coot:


erikd:
IMO, code should either get built in CI or removed from the repo.

coot:
We don't yet have Windows `CI` and this is windows only demo.  Hopefully sooner than later we will have windows ci.

PR #1777, opened 2020-03-10
Fix thunks in ChunkSize

mrBliss:
Introduced in #1750.


edsko:
Who wrote that code :angry: :flushed: :grin:

PR #1780, opened 2020-03-10
Use MD5 instead of ShortHash in Mempool tests

### LINKED ISSUE #1779: Replace ShortHash with MD5 in order to avoid hash collisions in Mempool tests
We have observed a `ShortHash` collision in the `ouroboros-consensus:test:test-consensus:Test.Consensus.Mempool` module. It uses `ShortHash`. That implementation appears to have the same compute time as MD5, but a quarter of the bits (it just takes a prefix of the MD5).

Instead of writing the tests to anticipate the occasional collision and thereby save the debugging time whenever it happens, we're just going to switch to MD5, under the assumption that it won't collide in these tests during the project's lifetime.

Thomas summarized the change as follows.

```
sed -i s/ShortHash/MD5/g ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/UTxO.hs
```

nfrisby:
Closes #1779.

I also included some of the additions that we needed in order to debug the original hash collision.

### DIFF
--- a/ouroboros-consensus/test-consensus/Test/Consensus/Mempool.hs
+++ b/ouroboros-consensus/test-consensus/Test/Consensus/Mempool.hs
@@ -318,10 +318,13 @@ ppTestSetup :: TestSetup -> String
 ppTestSetup TestSetup { testInitialTxs
                       , testMempoolCap = MempoolCapacityBytes mpCap
                       } = unlines $
-    ["Initial contents of the Mempool:"] <>
-    (map condense testInitialTxs)        <>
-    ["Mempool capacity:"]                <>
+    ["Initial contents of the Mempool:"]  <>
+    (map ppTestTxWithHash testInitialTxs) <>
+    ["Mempool capacity:"]                 <>
     [condense mpCap]
+
+ppTestTxWithHash :: TestTx -> String
+ppTestTxWithHash x = condense (hash (simpleGenTx x) :: Hash MD5 Tx, x)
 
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/test-consensus/Test/Consensus/Mempool.hs
```diff
@@ -318,11 +318,14 @@ ppTestSetup :: TestSetup -> String
...
-    ["Initial contents of the Mempool:"] <>
-    (map condense testInitialTxs)        <>
-    ["Mempool capacity:"]                <>
+    ["Initial contents of the Mempool:"]  <>
+    (map ppTestTxWithHash testInitialTxs) <>
+    ["Mempool capacity:"]                 <>
     [condense mpCap]
 
+ppTestTxWithHash :: TestTx -> String
+ppTestTxWithHash x = condense (hash (simpleGenTx x) :: Hash ShortHash Tx, x)
```
mrBliss: I'm assuming you're using `ShortHash` here instead of `MD5` for readability, but then it deserves a comment mentioning that `ShortHash` is just the first four bytes of the `MD5` hash. If that weren't the case, printing the hashes wouldn't be useful.

nfrisby: No, I meant to use MD5 for precision. Just neglected to update the snippet when copying from my debugging worktree.

PR #1787, opened 2020-03-11
ImmutableDB: use TempResourceRegistry in modifyOpenState

mrBliss:
Fixes https://github.com/input-output-hk/cardano-node/issues/651

Consider the following scenario:

* We open the ChainDB
* Which opens the ImmutableDB
* The background thread in the ChainDB that copies blocks from the VolatileDB
  to the ImmutableDB is started
* ...
* We have copied enough blocks such that we have to start a new epoch in the
  ImmutableDB, calling:
  ```haskell
  mkOpenState registry HasFS{..} index epoch tip existing = do
      eHnd <- allocateHandle $ hOpen (renderFile "epoch" epoch)     appendMode
      pHnd <- allocateHandle $ Index.openPrimaryIndex index epoch existing
      sHnd <- allocateHandle $ hOpen (renderFile "secondary" epoch) appendMode
      ...
  ```
* ...
* Now we CTRL-C the node
* The `ResourceRegistry` starts to release resources in LIFO order: It
  releases the three handles of the current epoch. Note that it does this
  before it releases the background thread that copies blocks to the
  ImmutableDB. So there's a short time frame in which the handles are closed
  and the background thread copying blocks hasn't been killed yet. In this
  time frame, we get a `FsIllegalOperation` because we're trying to write
  using a closed handle.

The solution is to not use the global registry for the per-epoch handles,
because they will be released before a thread using them is terminated. The
main reason these handles are stored in a `ResourceRegistry` is to avoid
leaking them before they have made it into the new ImmutableDB state. After
that point, they won't leak, as closing the ImmutableDB (a release action
registered in the global registry) will close them.

Instead, we use `modifyWithTempRegistry` to release them in case we encounter
an exception before we store them in the state.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE GADTs               #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
@@ -97,8 +96,7 @@ import           Prelude hiding (truncate)
 
 import           Control.Monad (replicateM_, when)
 import           Control.Monad.Except (runExceptT)
-import           Control.Monad.State.Strict (StateT (..), get, lift, modify,
-                     put)
+import           Control.Monad.State.Strict (get, lift, modify, put)
 import           Control.Tracer (Tracer, traceWith)
 import           Data.ByteString.Builder (Builder)
 import           Data.Functor (($>))
@@ -112,7 +110,8 @@ import           Ouroboros.Consensus.BlockchainTime (BlockchainTime,
                      getCurrentSlot)
 import           Ouroboros.Consensus.Util (SomePair (..))
 import           Ouroboros.Consensus.Util.IOLike
-import           Ouroboros.Consensus.Util.ResourceRegistry (ResourceRegistry)
+import           Ouroboros.Consensus.Util.ResourceRegistry (ResourceRegistry,
+                     runWithTempRegistry)
 
 import           Ouroboros.Consensus.Storage.Common
 import           Ouroboros.Consensus.Storage.FS.API
@@ -159,7 +158,7 @@ import           Ouroboros.Consensus.Storage.ImmutableDB.Parser
 -- __Note__: To be used in conjunction with 'withDB'.
 withDB
   :: forall m h hash e a.
-     (HasCallStack, IOLike m, Eq hash, NoUnexpectedThunks hash)
+     (HasCallStack, IOLike m, Eq hash, NoUnexpectedThunks hash, Eq h)
   => ResourceRegistry m
   -> HasFS m h
   -> ChunkInfo
@@ -227,9 +226,8 @@ mkDBRecord dbEnv = ImmutableDB
 -- * Non-bracketed, as @quickcheck-state-machine@ doesn't support that.
 openDBInternal
   :: forall m h hash e.
-     (HasCallStack, IOLike m, Eq hash, NoUnexpectedThunks hash)
-  => ResourceRegistry m  -- ^ The ImmutableDB will be in total control of
-                         -- this, not to be used for other resources.
+     (HasCallStack, IOLike m, Eq hash, NoUnexpectedThunks hash, Eq h)
+  => ResourceRegistry m
   -> HasFS m h
   -> ChunkInfo
   -> HashInfo hash
@@ -240,7 +238,7 @@ openDBInternal
   -> BlockchainTime m
   -> m (ImmutableDB hash m, Internal hash m)
 openDBInternal registry hasFS chunkInfo hashInfo valPol parser tracer
-               cacheConfig btime = do
+               cacheConfig btime = runWithTempRegistry $ do
     currentSlot <- atomically $ getCurrentSlot btime
     let validateEnv = ValidateEnv
           { hasFS
@@ -248,13 +246,12 @@ openDBInternal registry hasFS chunkInfo hashInfo valPol parser tracer
           , hashInfo
           , parser
           , tracer
-          , registry
           , cacheConfig
           , currentSlot
           }
-    !ost  <- validateAndReopen validateEnv valPol
+    ost <- validateAndReopen validateEnv registry valPol
 
-    stVar <- newMVar (DbOpen ost)
+    stVar <- lift $ newMVar (DbOpen ost)
 
     let dbEnv = ImmutableDBEnv
           { hasFS            = hasFS
@@ -271,7 +268,12 @@ openDBInternal registry hasFS chunkInfo hashInfo valPol parser tracer
         internal = Internal
           { deleteAfter_ = deleteAfterImpl dbEnv
           }
-    return (db, internal)
+    -- TODO move 'withTempResourceRegistry' outside of this function?
+    --
+    -- Note that we can still leak resources if the caller of
+    -- 'openDBInternal' doesn't bracket his call with 'closeDB' or doesn't
+    -- use a 'ResourceRegistry'.
+    return ((db, internal), ost)
 
 closeDBImpl
   :: forall m hash. (HasCallStack, IOLike m)
@@ -296,7 +298,8 @@ isOpenImpl ImmutableDBEnv { varInternalState } =
     dbIsOpen <$> readMVar varInternalState
 
 reopenImpl
-  :: forall m hash. (HasCallStack, IOLike m, Eq hash, NoUnexpectedThunks hash)
+  :: forall m hash.
+     (HasCallStack, IOLike m, Eq hash, NoUnexpectedThunks hash)
   => ImmutableDBEnv m hash
   -> ValidationPolicy
   -> m ()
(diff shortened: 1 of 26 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -567,6 +571,46 @@ releaseResources rr keys releaser = do
...
+-- exception. See 'TempResourceRegistry' for more details.
+withTempResourceRegistry
+  :: (IOLike m, HasCallStack)
+  => (TempResourceRegistry m -> m a)
+  -> m a
+withTempResourceRegistry =
+    bracketOnError
+      (UnsafeTempResourceRegistry <$> unsafeNewRegistry)
+      (\(TempResourceRegistry registry) -> closeRegistry registry)
+
```
mrBliss: @kderme After thinking some more about the `withRegistryOnError` you introduced in #1681 (https://github.com/input-output-hk/ouroboros-network/commit/0d935923c3eba0736d22134cae9a0f21c1e8814f), @edsko and I realised it was actually a good idea.

In the VolatileDB, we no longer need to be AsyncException-resistant (which is something much more complex) thanks to #1709.

mrBliss:
> LGTM. Is there any specific motive for `untrackResource` and the additional checks? Or is it more of **a sanity checklist that all resources have made it to the final destination**?

Exactly!

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/State.hs
```diff
@@ -177,15 +187,15 @@ modifyOpenState
   -> (  forall h.
         TempResourceRegistry m
      -> HasFS m h
-     -> StateT (OpenState m hash h) m r
+     -> StateT (OpenState m hash h) m ([ResourceKey m], r)
```
mrBliss: Use https://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-RWS-CPS.html#t:RWST with `TempResourceRegistry m` as R. Call it `ModifyOpenState m hash h`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs
```diff
@@ -800,7 +815,8 @@ startNewChunk tempRegistry hasFS index chunkInfo = do
       Index.appendOffsets index currentPrimaryHandle backfillOffsets
       `finally` cleanUp hasFS st
```
mrBliss: Unrelated to this PR: we always call `cleanUp` to close the open handles of the current epoch. However, `cleanUp` also calls `Index.close`, which stops the background thread that expires past epochs from the cache! So we are leaking memory!

mrBliss: Make sure that we have a test that catches this bug. Create a ticket for this if necessary.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/State.hs
```diff
@@ -214,7 +225,7 @@ modifyOpenState ImmutableDBEnv { hasFS = hasFS :: HasFS m h, .. } action = do
           ExitCaseAbort                               -> (DbOpen ost, return ())
           ExitCaseException _ex                       -> (DbOpen ost, return ())
           -- In case of success, update to the newest state
-          ExitCaseSuccess (Right (_, ost'))           -> (DbOpen ost', return ())
+          ExitCaseSuccess (Right ((rs, _), ost'))     -> (DbOpen ost', mapM_ untrackResource rs)
```
mrBliss: This is not quite right yet, it's too easy to make mistakes.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs
mrBliss: #1794.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -574,6 +588,182 @@ releaseResources rr keys releaser = do
...
+-- e.g., stored in a 'TVar', after which they are guaranteed to be released
+-- correctly, it is possible that an exception is thrown after allocating such
+-- a resource, but before it was stored in the final state. In that case, the
+-- resource would be leaked. 'runWithTempRegistry' solves that problem.
+--
+-- When no exception is thrown before the end of 'runWithTempRegistry', the
+-- user must have transferred all the resources it allocated to their final
+-- state. This means that these resources don't have to be released by the
+-- temporary registry anymore, the final state is now in charge of releasing
+-- them (typically by another 'ResourceRegistry').
```
mrBliss:
```suggestion
-- them.
```

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -574,6 +588,182 @@ releaseResources rr keys releaser = do
...
+--
+-- In case an exception is thrown before the end of 'runWithTempRegistry',
+-- /all/ resources allocated in the temporary registry will be released.
+--
+-- Resources must be allocated using 'allocateTemp'.
+--
+-- To make sure that the user doesn't forget to transfer a resource to the
+-- final state @st, the user must pass a function to 'allocateTemp' that
```
mrBliss:
```suggestion
-- final state @st@, the user must pass a function to 'allocateTemp' that
```

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -574,6 +588,182 @@ releaseResources rr keys releaser = do
...
+--
+-- When no exception is thrown before the end of 'runWithTempRegistry', we
+-- check whether all allocated resources have been transferred to the final
+-- state @st@. If there's a resource that hasn't been transferred to the final
+-- state /and/ that hasn't be released or closed before (see the release
+-- function passed to 'allocateTemp'), a 'TempRegistryRemainingResource'
+-- exception will be thrown.
+--
+-- For that reason, 'WithTempRegistry' is parameterised over the final state
+-- type @st@ and the given 'WithTempRegistry' must return the final state.
```
mrBliss:
```suggestion
-- type @st@ and the given 'WithTempRegistry' action must return the final state.
```

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -574,6 +588,182 @@ releaseResources rr keys releaser = do
...
+--
+-- NOTE: we explicitly don't let 'runWithTempRegistry' the final state,
```
mrBliss:
```suggestion
-- NOTE: we explicitly don't let 'runWithTempRegistry' return the final state,
```

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -574,6 +588,182 @@ releaseResources rr keys releaser = do
...
+
+-- | Allocate a resource in a temporary registry until it has been transferred
+-- to the final state @st@. See 'runWithTempRegistry' for more details.
+allocateTemp
+  :: (IOLike m, HasCallStack)
+  => m a
+     -- ^ Allocate the resource
+  -> (a -> m Bool)
+     -- ^ Release the resource, return 'True' when the resource was actually
+     -- released, return 'False' when the resource was already released.
```
mrBliss: Mention again (or refer to `allocateEither`) that the user is allowed to always return `True`.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -574,6 +588,182 @@ releaseResources rr keys releaser = do
...
+    (key, a) <- lift $ fmap mustBeRight $
+      allocateEither rr (fmap Right . const alloc) free
+    atomically $ modifyTVar varTransferredTo $ mappend $
+      TransferredTo $ \st ->
+        if isTransferred st a
+        then Set.singleton (resourceKeyId key)
+        else Set.empty
+    return a
+
+-- | Example use case of 'runWithTempRegistry': modify the given @st@,
```
mrBliss:
```suggestion
-- | Higher level API on top of `runWithTempRegistry`: modify the given @st@,
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/State.hs
```diff
@@ -152,64 +159,70 @@ getOpenState ImmutableDBEnv {..} = do
...
+  StateT (OpenState m hash h) (WithTempRegistry (OpenState m hash h) m)
+
 -- | Modify the internal state of an open database.
 --
 -- In case the database is closed, a 'ClosedDBError' is thrown.
 --
 -- In case an 'UnexpectedError' is thrown, the database is closed to prevent
 -- further appending to a database in a potentially inconsistent state.
 --
+-- The action is run in the 'ModifyOpenState' monad, which is 'StateT'
```
mrBliss:
```suggestion
-- The action is run in the 'ModifyOpenState' monad, which is a 'StateT'
```

mrBliss:
I should rebase on #1792 to be sure.

PR #1774, opened 2020-03-10
Drop openssl dep

kderme:
With this pr, ouroboros-consensus no longer needs openssl to build. This significantly simplifies building on windows. Related https://github.com/input-output-hk/ouroboros-network/issues/1082

WIP: TODO nix

PR #1789, opened 2020-03-11
Fix stack for windows

kderme:
Related https://github.com/input-output-hk/ouroboros-network/issues/1082
Continuation of https://github.com/input-output-hk/ouroboros-network/pull/1774, which makes use of https://github.com/input-output-hk/cardano-prelude/pull/100

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: https://raw.githubusercontent.com/input-output-hk/cardano-prelude/3893f21db1665e7bab9e8c6504c547fc7c28f98c/snapshot.yaml
+resolver: https://raw.githubusercontent.com/input-output-hk/cardano-prelude/15329855d4eadbee5f5f486104c7b04b50e05f7b/snapshot.yaml
 
 packages:
   - ./typed-protocols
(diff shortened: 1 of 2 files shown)


stack.yaml
```diff
@@ -91,6 +94,7 @@ extra-deps:
 
   # Windows only
   - Win32-2.6.2.0
+  - hedgehog-1.0.1@sha256:b65b1ebde7b7fe8e693070916e45aa2ed63eea5e3f84ac3a8f9a0b7f8fcf0337
```
mrBliss: Why is this needed? The `snapshot.yaml` in `cardano-prelude` mentions `hedgehog-1.0.2`, so this is overriding it. Also, the `cabal.project` does not mention this constraint.

kderme:
@mrBliss with this, `ouroboros-consensus` still depends on `3893f21db1665e7bab9e8c6504c547fc7c28f98c` of prelude, but uses the resolver of `fe76ec64f6b45259cc407a6d840dad79ee6063b6`. Do you think this is an inconsistency we should fix? I just though bumping also the dependency could mean more cascading changes.

stack.yaml
kderme: My bad. I probably first tried stack on a pretty old commit which mentioned `hedgehog-1.0`

mrBliss:
@kderme Yes, it's an inconsistency that should be fixed.

stack.yaml
mrBliss: Related: #1796

kderme:
stack and cabal work on windows

mrBliss:
FYI, the test failure on Hydra is transient and the sign of an overloaded CI machine (the test involves time).

PR #1790, opened 2020-03-12
Check that Byron can update the software version

### LINKED ISSUE #1733: Extend PBFT tests with software version update proposals/votes
Like Issue #1514 but for software updates instead of protocol parameter updates.

A proposal can include a protocol parameter update and/or a software update. The flow of a software update is a prefix of the flow of a protocol parameter update. Once a proposal is put forth, nodes can vote on it. Enough votes confirm it. Once that confirming vote is at least 2k slots old (ie "stable", based on the Byron chain density invariant), any software version update within the proposal has been adopted; nodes are expected to immediately restart in order to continue using new identified software version. This is how the nets' nodes will agree to hard-fork (which can  ultimately change anything, including in particular any protocol/node parameters not managed by parameter update proposals). Any protocol parameter update in the proposal is not yet adopted; see Issue #1514 for how that part separately continues.

This test need not actually update any software, but it should at least track that a net can extend a chain so that the tip of the starting chain and the tip extended chain induce two chain validation states that specify different adopted software version numbers.

Related: Issue #1514, Issue #1732.

### LINKED ISSUE #1773: Immutable DB: Do "blocks in future" check _after_ ChainDB almost initialized
At the moment, the imm DB verifies that there are no blocks from the future as it walks over the chunk files. However, the translation from `SlotNo` to `UTCTime` is ledger dependent :roll_eyes: We should instead initialize the imm DB and ledger DB _first_, not checking time stamps, and only then check the tip of the imm DB. If that is ahead of the current time, we should _trim_ the imm DB and kill the node.

nfrisby:
Fixes #1733.

I needed to revise the expectations to handle voting more appropriately; there were some slight inaccuracies which I think had been mostly masked by the fact that adopting protocol version requires endorsements and further waiting -- so when _exactly_ the proposal was confirmed hadn't mattered in a test case yet. But for software versions, confirmation is adoption, so accurately predicting when that happen became directly necessary.

There is only one proposal, but it proposes both a protocol version and a software version update. The PV update implies the SV update, but not vice versa. For example, the test output includes the following.

```
    	      proposed protocol version was adopted (20 in total):
     	      70% False
    	      30% True
    	      
    	      proposed software version was adopted (20 in total):
    	      95% True
    	       5% False
```

### DIFF
--- a/ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
+++ b/ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
@@ -535,6 +535,49 @@ tests = testGroup "RealPBFT" $
             , slotLength   = defaultSlotLength
             , initSeed     = Seed {getSeed = (5875984841520223242,5307155813931649482,9880810077012492572,1841667196263253753,11730891841989901381)}
             }
+    , testProperty "only check updates for mesh topologies" $
+          -- This repro exercises
+          -- 'Test.ThreadNet.RealPBFT.TrackUpdates.checkTopo'.
+          --
+          -- The predicted slot outcomes are
+          --
+          -- > leader 01234
+          -- >    s0  NAAAA
+          -- >    s5  NAAAA
+          -- >    s10 NWN
+          --
+          -- The votes of c1, c3, and c4 arrive to c2 during s11 via TxSub
+          -- /before/ the block containing the proposal does, so c2's mempool
+          -- rejects them as invalid. When it then forges in s12, it only
+          -- includes its own vote, which doesn't meet quota (3 = 5 * 0.6) and
+          -- so the proposal then expires (TTL 10 slots, but only after an
+          -- endorsement; see Issue 749 in cardano-ledger).
+          --
+          -- "Test.ThreadNet.RealPBFT.TrackUpdates" does not otherwise
+          -- correctly anticipate such races, so it makes no requirement for
+          -- non-mesh topologies.
+          prop_simple_real_pbft_convergence NoEBBs SecurityParam {maxRollbacks = 10} TestConfig
+            { numCoreNodes = NumCoreNodes 5
+            , numSlots     = NumSlots 13
+            , nodeJoinPlan = NodeJoinPlan $ Map.fromList
+              [ (CoreNodeId 0, SlotNo 0)
+              , (CoreNodeId 1, SlotNo 11)
+              , (CoreNodeId 2, SlotNo 11)
+              , (CoreNodeId 3, SlotNo 11)
+              , (CoreNodeId 4, SlotNo 11)
+              ]
+            , nodeRestarts = noRestarts
+            , nodeTopology = NodeTopology $ Map.fromList
+                -- mesh except for 0 <-> 2
+              [ (CoreNodeId 0, Set.fromList [])
+              , (CoreNodeId 1, Set.fromList [CoreNodeId 0])
+              , (CoreNodeId 2, Set.fromList [CoreNodeId 1])
+              , (CoreNodeId 3, Set.fromList [CoreNodeId 0, CoreNodeId 1, CoreNodeId 2])
+              , (CoreNodeId 4, Set.fromList [CoreNodeId 0, CoreNodeId 1, CoreNodeId 2, CoreNodeId 3])
+              ]
+            , slotLength   = defaultSlotLength
+            , initSeed     = Seed {getSeed = (8051309618816278461,2819388114162022931,16483461939305597384,11191453672390756304,8021847551866528244)}
+            }
     , testProperty "simple convergence" $
           \produceEBBs ->
           -- TODO k > 1 as a workaround for Issue #1511.
@@ -655,6 +698,7 @@ prop_simple_real_pbft_convergence produceEBBs k
     { numCoreNodes
     , numSlots
     , nodeJoinPlan
+    , nodeTopology
     , nodeRestarts
     , initSeed
     } =
@@ -768,6 +812,7 @@ prop_simple_real_pbft_convergence produceEBBs k
             numSlots
             genesisConfig
             nodeJoinPlan
+            nodeTopology
             refResult
             ldgr
         | (cid, ldgr) <- finalLedgers
--- a/ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
+++ b/ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
@@ -518,6 +518,23 @@ tests = testGroup "RealPBFT" $
             , slotLength   = defaultSlotLength
             , initSeed     = Seed {getSeed = (306806859316465898,5351335255935493133,6240542044036351784,5824248410373935607,16492982022780410836)}
             }
+    , testProperty "only expect EBBs if the reference simulator does" $
+          -- In this repro, block in the 20th slot is wasted since c2 just
+          -- joined. As a result, the final chains won't include that EBB.
+          let ncn = NumCoreNodes 3 in
+          prop_simple_real_pbft_convergence ProduceEBBs SecurityParam {maxRollbacks = 2} TestConfig
+            { numCoreNodes = ncn
+            , numSlots     = NumSlots 21
+            , nodeJoinPlan = NodeJoinPlan $ Map.fromList
+              [ (CoreNodeId 0,SlotNo {unSlotNo = 0})
+              , (CoreNodeId 1,SlotNo {unSlotNo = 0})
+              , (CoreNodeId 2,SlotNo {unSlotNo = 20})
+              ]
+            , nodeRestarts = noRestarts
+            , nodeTopology = meshNodeTopology ncn
+            , slotLength   = defaultSlotLength
+            , initSeed     = Seed {getSeed = (5875984841520223242,5307155813931649482,9880810077012492572,1841667196263253753,11730891841989901381)}
+            }
     , testProperty "simple convergence" $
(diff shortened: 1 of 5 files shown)


nfrisby:
@edsko @mrBliss I intend to back-off or at least relocate the `MempoolFingerprint`. This is a seemingly necessary complication due to wanting the voting to be prompt enough to be simple to predict. This fingerprint is used to emulate a "perfect insistent wallet" that knows exactly when to try adding a transaction again in order to maximize that transactions promptness. It feels too low-level, but I don't see how to avoid it yet.

I'd like to talk tomorrow (Thursday) afternoon, if possible.

nfrisby:
I just force-pushed commit 28b394e1f8bf999d28f7ff7858dcb7a00f9ccb4f. I think that commit is correctly testing the proposals. It has three main parts.

  1. It fixed a bug in its expectations for proposals that I think would have eventually manifested as a failed protocol version update test.
  1. It added software version update tests.
  1. It adds `MempoolFingerprint` -- I'm working now to relocate that interface/logic to emphasize that it's only for tests.

nfrisby:
Remarkably, the testing framework discovered and now correctly handles the case were a proposal expires, is re-proposed, and confirmed the second time around.

`cardano-ledger` seems correct. One oddity I've observed (and the test suite currently expects) is that a proposal can be confirmed **more than TTL slots later**, as long as the first endorsement (i.e. block) since its expiry contains the confirming vote(s). This is because the `cardano-ledger` logic (and the spec?) register the votes in a block before they registers its endorsements, and it is endorsements that trigger proposals' expiry.

@edsko @nc6 ^^^ Worth checking if that's what the spec does say and should say?

nfrisby:
>     3. It adds `MempoolFingerprint` -- I'm working now to relocate that interface/logic to emphasize that it's only for tests.

I haven't determined how to do this without duplicating a lot of code. `Test.ThreadNet.Network` access the mempool in the following way.

```
      nodeKernel <- initNodeKernel nodeArgs
      let mempool = getMempool nodeKernel
```

So in order to move the `getFingerprint` method out of the standard `Mempool` interface and into a `TestingMempool` interface, I'd have to define an `initTestNodeKernel`, a `TestingNodeKernel` interface, and so on. Is that worth it to keep `getFingerprint` out of the standard `Mempool` interface? @edsko @mrBliss

nfrisby:
I have marked the PR Ready for Review.

Thomas and I had a call. We talked through the motivation for adding `Mempool.getFingerprint` and were ultimately able to avoid it. His key insight was that a particular race could be avoided: the race exists between the emulated wallet thread and the Mempool's background thread. Both wait on the ledger to change. We don't want the wallet to trigger before the bg thread. His insight was that the bg thread is trivial, so we can just have the wallet do the same thing that the bg thread does. So we avoid the race by having both racers (begin) by doing the same exact thing.

nfrisby:
Bah; I've observed a new failure. It's related to non-mesh topologies. I'm investigating/assessing now.

The quick/easy option on the table if this takes too long: only test this property for mesh topologies and adjust the generator to generate mesh topologies a little more often.

nc6:
@nfrisby @mrBliss Yes, also saw a variation on this issue last week. I've created https://github.com/input-output-hk/cardano-ledger/issues/749 to address it.

nfrisby:
> Bah; I've observed a new failure. It's related to non-mesh topologies. I'm investigating/assessing now.

Here's my summary. Context first.

  * Node `c0` has a "wallet thread" that adds the proposal and its vote for it to its own mempool at every possible occasion (~"whenever the ledgerstate or mempool changes").
  * The others nodes have similar "wallet threads" that add their vote (but not the proposal) to their own mempool at every possible occasion. (It is rejected as invalid until the node either receives the proposal transaction via TxSub or receives and selects a block containing it.)
  * Notably, receiving a block via BlockFetch and selecting it triggers the wallet thread to add the vote again, since it might now be a valid tx for the first time.
  * As a one-time and nearly-immediate consequence of adding a transaction to its mempool, a node's TxSub client (begins the process that ultimately) pushes that transaction, once, to each of its peers.

In this failure, we have a 5-node mesh network with one missing connection: c0 is not directly connected to c2. Also, c0 joins in slot 0 and the rest of the nodes all join at the same time in slot 11. Slot 12 is the last slot in the repro.

Altogether, that means that the other nodes' votes arrive at c2 before the proposal does, and so its mempool rejects them. Since TxSub doesn't send (this instance of) the votes more than once, c2 doesn't receive them before it leads slot 12. The test `Property` doesn't anticipate that outcome of the CS+BF vs TxSub race and so expects the proposal to have succeeded, whereas in this threadnet run it actually expires.

I'm now weighing options. Some examples.

  * Try to anticipate the topology's effect on this race in the `Property`.
  * Avoid the topology's effect on this race in the test case generator.
  * Have all nodes behave like `c0`: make their wallet thread include the proposal tx in addition to their own vote. This would ensure the votes are always valid txs, since the proposal's existence is their only prerequisite.

mrBliss:
> [..]

I see.

> I'm now weighing options. Some examples.
> 
>     * Try to anticipate the topology's effect on this race in the `Property`.
> 
>     * Avoid the topology's effect on this race in the test case generator.
> 
>     * Have all nodes behave like `c0`: make their wallet thread include the proposal tx in addition to their own vote. This would ensure the votes are always valid txs, since the proposal's existence is their only prerequisite.

I suppose the first option is the "best" but hardest. Since the lower importance of this test (see https://github.com/input-output-hk/ouroboros-network/issues/1733#issuecomment-599966737), I think the second option is fine, granted that it only affects test cases in which the software version is updated, not Byron test cases in general.

nfrisby:
@mrBliss I pushed a commit that disables this particular test if the topology is not mesh.

ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
```diff
@@ -426,6 +426,158 @@ tests = testGroup "RealPBFT" $
...
+              , (CoreNodeId 2, SlotNo 4)
+              , (CoreNodeId 3, SlotNo 4)
+              ]
+            , nodeRestarts = noRestarts
+            , nodeTopology = meshNodeTopology ncn
+            , slotLength   = defaultSlotLength
+            , initSeed     = Seed {getSeed = (7536539674426109099,5947274896735415773,14396421290275890646,8359457880945605675,13921484090802881569)}
+            }
+    , testProperty "cardano-ledger checks for proposal confirmation before it checks for expiry" $
+          -- much check for quorum before checking for expiration
```
mrBliss:
```suggestion
          -- must check for quorum before checking for expiration
```

ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT/TrackUpdates.hs
```diff
@@ -50,46 +54,52 @@ import qualified Ouroboros.Consensus.Byron.Ledger as Byron
 import qualified Test.ThreadNet.Ref.PBFT as Ref
 import           Test.ThreadNet.Network (TestNodeInitialization (..))
 import           Test.ThreadNet.Util.NodeJoinPlan
+import           Test.ThreadNet.Util.NodeTopology
 
 import           Test.ThreadNet.RealPBFT.ProtocolInfo
 
+-- | The expectation and observation regarding whether the hard-fork proposal
+-- succesfully updated the protocol version
```
mrBliss:
```suggestion
-- successfully updated the protocol version
```

ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT/TrackUpdates.hs
```diff
@@ -50,46 +54,52 @@ import qualified Ouroboros.Consensus.Byron.Ledger as Byron
...
+--
 data ProtocolVersionUpdateLabel = ProtocolVersionUpdateLabel
-  { pvuRequired :: !(Maybe Bool)
+  { pvuObserved :: !Bool
+    -- ^ whether the proposed protocol version is adopted or not adopted by the
+    -- end of the test
+  , pvuRequired :: !(Maybe Bool)
```
mrBliss: Just wondering now why there are bangs here? That bang on the `Maybe` is unlikely to help much.

mrBliss:
@nfrisby You can bors this when you think it is done.

ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT/TrackUpdates.hs
nfrisby: Fixed in latest push, thanks.

ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
nfrisby: Fixed in latest push, thanks.

ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT/TrackUpdates.hs
nfrisby: Thomas and I chatted. We decided to leave strictness annotations off of test code fields unless empirically warranted. Basic idea: it should be as obvious as possible that little-to-no-effort was made to avoid space-leaks here. Lazily (har) putting a bang there without working to make it deep strictness might just give others (including the future self) a false sense of security.

nfrisby:
bors retry

nfrisby:
When I run `v2-test all` on this commit locally, it take 10 minutes. So I'm going to try again.

bors retry

PR #1806, opened 2020-03-13
Remove duplicate transaction check from tryAddTxs in mempool

### LINKED ISSUE #1801: Remove dup tx check from mempool
This would first require to fix the tx request client though.]

Related but orthogonal: #1743

intricate:
Closes #1801

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
@@ -273,13 +273,6 @@ implTryAddTxs mpEnv = go []
         join $ atomically $ readTVar mpEnvStateVar >>= tryAdd
       where
         tryAdd is
-          -- This transaction already exists within the mempool.
-          -- Note that we don't treat this as an error/rejection case.
-          | implSnapshotHasTx is (txId firstTx)
-          = return $ go
-              ((firstTx, MempoolTxAdded):acc)
-              toAdd'
-
           -- No space in the Mempool.
           | let firstTxSize = txSize firstTx
                 curSize = msNumBytes $ isMempoolSize is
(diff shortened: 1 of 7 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -267,13 +267,6 @@ implTryAddTxs mpEnv = go []
         join $ atomically $ readTVar mpEnvStateVar >>= tryAdd
       where
         tryAdd is
-          -- This transaction already exists within the mempool.
-          -- Note that we don't treat this as an error/rejection case.
-          | implSnapshotHasTx is (txId firstTx)
-          = return $ go
-              ((firstTx, MempoolTxAdded):acc)
```
intricate: It seems that I originally made a mistake here actually. This should have been `MempoolTxAlreadyInMempool` rather than `MempoolTxAdded`.

I also need to remove the `MempoolTxAlreadyInMempool` constructor of `MempoolAddTxResult` as part of this PR since we don't use it.

intricate:
Just noting that this PR is waiting on https://github.com/input-output-hk/cardano-ledger/issues/759 to be resolved.

This is because the `prop_simple_real_pbft_convergence`-related tests expect that duplicate transactions are not accepted and, after this PR removes the duplicate tx check from the mempool, we now rely on the ledger rules to ensure this.  However, as was discovered in https://github.com/input-output-hk/cardano-ledger/issues/759, it is possible for certain duplicate update proposals to be applied to the ledger state without being rejected.

intricate:
Using https://github.com/input-output-hk/cardano-ledger/pull/766, the tests on this PR run and pass as expected. 

Once that PR is merged, this should be good to go.

intricate:
https://github.com/input-output-hk/cardano-ledger/pull/766 has now been merged.

PR #1824, opened 2020-03-18
Preparation for Shelley

mrBliss:
These are all the preparatory changes I made in #1695 that don't have to be part of that PR.

This PR can be reviewed commit per commit.

### DIFF
--- a/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/DelegationHistory.hs
+++ b/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/DelegationHistory.hs
@@ -1,9 +1,9 @@
-{-# LANGUAGE DataKinds          #-}
-{-# LANGUAGE DeriveAnyClass     #-}
-{-# LANGUAGE DeriveGeneric      #-}
-{-# LANGUAGE DerivingStrategies #-}
-{-# LANGUAGE OverloadedStrings  #-}
-{-# LANGUAGE RecordWildCards    #-}
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE RecordWildCards            #-}
+{-# LANGUAGE TypeApplications           #-}
 
 -- | Delegation history
 --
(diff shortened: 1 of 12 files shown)


mrBliss:
@nfrisby I'd like to get your opinion on 46dda816a3fa7799f6a7f6cbcfedaa159bd048e0: in Shelley, slot and block number 0 are reserved for genesis/the empty chain (as opposed to `WithOrigin(Origin)`), which means that any block forged in slot number 0 is invalid, as the slot numbers must strictly increase in Shelley (no EBBs, yay!)

On mainnet, Shelley will not start at slot 0, so it won't be an issue there, but in the testnet and the ThreadNet tests, it matters. My first thought (the commit in question) was to simply start our blockchain time at slot 1, sidestepping this issue entirely. An alternative is to let the TPraos leadership check always fail in slot 0 so that no node produces a block in that slot, but that's quite ad-hoc.

nfrisby:
I like the idea to merely have the nodes join late, since that means the `NumSlots` argument retains its "obvious" semantics on some level.

For the same reason, I'm weary of the adjustment to `newTestBlockchainTime`. But it's not clear to me how we could avoid it. I think it's best to prioritize the `lastSlot = numSlots - 1` interpretation; I know we rely on that in places, whereas I can't think of any places where we rely on the test clocks _actually ticking_ `numSlots` times.

> My first thought (the commit in question) was to simply start our blockchain time at slot 1, sidestepping this issue entirely. An alternative is to let the TPraos leadership check always fail in slot 0 so that no node produces a block in that slot, but that's quite ad-hoc.

There's a similar complication with `DualPBft`, if I understand correctly. See this function call, and the nearby definition. It seems like we only need one of these workarounds (and I'm leaning toward yours so that `trivialNodeJoinPlan` doesn't cause `discard` --- it likely hasn't caused trouble in `DualPBft` yet only since we run so few of those).

https://github.com/input-output-hk/ouroboros-network/blob/85934cc7e19e4a94547c79073d28ca77bb29897f/ouroboros-consensus-byron/test/Test/ThreadNet/DualPBFT.hs#L80

Hmmm... maybe you won't be able to, at least not without further changes. DualPBFT relies on RealBFT, and RealPBFT and PBFT both rely on the reference PBFT simulator. You might need to parameterize it over the first slot number as well... that's the lone `0` on this line.

https://github.com/input-output-hk/ouroboros-network/blob/85934cc7e19e4a94547c79073d28ca77bb29897f/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Ref/PBFT.hs#L119

Hmm, this "Is it `Origin` or `SlotNo 0` question?" has always been lurking and this seems like yet another clump of complications due to it :(

I'm sending this message as-is to prompt your thoughts, but I'll think about it a bit more.

mrBliss:
@JaredCorduan is going to switch the Shelley ChainState (our LedgerState) to use `WithOrigin SlotNo` instead of `SlotNo` (and `WithOrigin BlockNo`). This means we can start producing blocks at slot 0 again, so we're not going to need this workaround anymore :tada: :relieved:. (It was also causing inconsistencies in `Ouroboros.Consensus.Shelley.Protocol.State`).

nfrisby:
bors retry

PR #1825, opened 2020-03-18
CAD-742: make the EBB serialisation match cardano-sl

deepfire:
There were two problems in the way `forgeEBB` was serialising the EBB, leading to a mismatch with `cardano-sl`:

1. the body proof was off, now fixed by @dcoutts in https://github.com/input-output-hk/cardano-ledger/pull/751/commits/9c53870095a8a6f72c3e84a3ea8cd297a8495de6
2. the ABOB tag wrapper was missing, leading to two missing leading bytes -- see `0x82`, `0x00` in the following serialisation from `cardano-sl`:
```
[nix-shell:~/cardano-node]$ cardano-cli pretty-print-cbor --filepath ebb.rewrite
    82  # list(2)
       00  # int(0)
       83  # list(3)
          85  # list(5)
             1a 00 98 96 80  # int(10000000)
             58 20 ec 4f af ee 43 d6 4b 00 9e aa e5 35 32 90
             16 72 50 b7 06 94 52 8e bb d4 79 61 96 73 12 8e
             9c c2  # bytes(32)
             58 20 af c0 da 64 18 3b f2 66 4f 3d 4e ec 72 38
             d5 24 ba 60 7f ae ea b2 4f c1 00 eb 86 1d ba 69
             97 1b  # bytes(32)
             82  # list(2)
                00  # int(0)
                81  # list(1)
                   00  # int(0)
             81  # list(1)
                a0  # map(0)
          9f  # list(*)
          ff  # break
          81  # list(1)
             a0  # map(0)
```
This PR makes the Rewrite's serialisation for the EBB match that of `cardano-sl`.

### DIFF
--- a/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Forge.hs
+++ b/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Forge.hs
@@ -67,7 +67,7 @@ forgeEBB
 forgeEBB cfg curSlot curNo prevHash =
         mkByronBlock (byronEpochSlots byronConfig)
       . CC.Block.ABOBBoundary
-      . CC.reAnnotateBoundary (byronProtocolMagicId byronConfig)
+      . CC.reAnnotateABOBBoundary (byronProtocolMagicId byronConfig)
       $ boundaryBlock
   where
     byronConfig :: BlockConfig ByronBlock
--- a/cabal.project
+++ b/cabal.project
@@ -181,29 +181,29 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-ledger
-  tag: e2240ccda93e52835cd25b04e1963c8929bbfa64
-  --sha256: 08c1dz9plnqb6i3ysnfbw6b6y2ij2f8crkij5aqrgsdjar51p91n
+  tag: d00fdb3407c5d13aa57f2ead79490a81854879a4
+  --sha256: 0hkvv14mnp9bgpii062lvid2a1k6ynj8zjdaws7d5yd040gsm1gz
   subdir: cardano-ledger
 
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-ledger
-  tag: e2240ccda93e52835cd25b04e1963c8929bbfa64
-  --sha256: 08c1dz9plnqb6i3ysnfbw6b6y2ij2f8crkij5aqrgsdjar51p91n
+  tag: d00fdb3407c5d13aa57f2ead79490a81854879a4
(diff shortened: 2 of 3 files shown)


mrBliss:
We used to have a CI check to protect us against depending on non-master dependencies like this one, but that was removed as part of the Nix overhaul :disappointed:.

The changes look good, ping me when the PR in `cardano-ledger` has been merged and you want me to merge this.

ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Forge.hs
```diff
@@ -67,7 +67,7 @@ forgeEBB
 forgeEBB cfg curSlot curNo prevHash =
         mkByronBlock (byronEpochSlots byronConfig)
       . CC.Block.ABOBBoundary
-      . CC.reAnnotateBoundary (byronProtocolMagicId byronConfig)
+      . CC.reAnnotateABOBBoundary (byronProtocolMagicId byronConfig)
```
mrBliss: Can you show what difference this makes? I'm not sure I think it does. Previously, we were reannotating the EBB (`ebb :: ABoundaryBlock`): so we serialise the EBB and use that bytestring as the annotation. Now, you're reannotating the EBB as part of the `ABOBBoundary` sum (`ABOBBoundary (ebb :: ABoundaryBlock)`), so you serialise the `ABlockOrBoundary`, which means you'll have an extra `encodeListLen 2 <> encodeWord 0` at the start of the bytestring, but then when we annotate during decoding, we slice those two extra bytes back off the annotation. The end result is that you encoded and dropped two extra bytes, without changing the end result. Or am I missing something?

deepfire: In practical terms, this removes the EBB hash disagreement between all three Chairman's cluster participant types:  legacy, proxy, rewrite.

In terms of code, the fix repairs just the `forgeEBB` code path, which is serialisation only (although it does a roundtrip to check deserialisation, but that's just a local validation).

On the decoding path, we're using a function that properly dispatches on the ABOB tag:

1. `decodeByronBlock`: https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Serialisation.hs#L120
2. `decodeUnsizedHeader`: https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Serialisation.hs#L306

deepfire:
Aww, @mrBliss, I've accidentally deleted your comment somehow.. I swear I was deleting mine.. : -/

deepfire:
Anyway, this PR is obsolete, in light of what @mrBliss said in the comment (that I've accidentally deleted :sob:).

PR #1831, opened 2020-03-19
Policy for lmitting number of accepted connections

coot:
This pr implements a policy which allows us to limit the number of accepted connections, addressing the issue #1391.

There are two thresholds:
- `acceptedConnectionsHardLimit`
- `acceptedConnectionsSoftLimit`
And a delay
- `acceptedConnectionsDelay`

If we are below the soft limit, we accept connections without any delay;
Above the *soft limit*, we start rate limiting using a linear regression of the `acceptedConnectionsDelay`.
If we hit the hard limit, then we wait until the number of connections drops
(but not shorted than `acceptedConnectionsDelay`).   (We could add a third
threshold to drop below that, but we should be good for now with this policy).

This is a standalone component, so it will be easy to bring it to
`connection-manager`.

Todo:
- [x] add a trace

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
@@ -1,0 +1,123 @@
+{-# LANGUAGE BangPatterns   #-}
+{-# LANGUAGE NamedFieldPuns #-}
+
+-- | Rage limiting of accepted connections
+--
+module Ouroboros.Network.Server.RateLimiting
+  ( AcceptedConnectionsLimit (..)
+  , runConnectionRateLimits
+  ) where
+
+import           Control.Monad.Class.MonadSTM
+import           Control.Monad.Class.MonadTime
+import           Control.Monad.Class.MonadTimer
+import           Control.Monad (when)
+
+import           Data.Word
+
+
+-- | Policy which governs how to limit the number of accepted connections.
+--
+data AcceptedConnectionsLimit = AcceptedConnectionsLimit {
+
+    -- | Hard limit of accepted connections.
+    --
+    acceptedConnectionsHardLimit :: !Word32,
+
+    -- | Soft limit of accepted connections.  If we are above this threshold,
+    -- we will start rate limiting.
+    --
+    acceptedConnectionsSoftLimit :: !Word32,
+
+    -- | Max delay for limiting accepted connections.  We use linear
+    -- regression starting from 0 at the soft limit up to
+    -- `acceptedConnectionDelay` at the hard limit.
+    --
+    acceptedConnectionsDelay     :: !DiffTime
+  }
+  deriving (Eq, Ord, Show)
+
+
+-- | Rate limiting instruction.
+--
+data RateLimitDelay =
+      -- | no rate limiting
+      --
+      NoRateLimiting
+
+      -- | We are above the soft limit, we delay accepting the next connection>
+    | SoftDelay DiffTime
+
+      -- | We are above the hard limit, wait until the number of connections
+      -- drops below the given threshold (currently this is the hard limit,
+      -- which means we keep `acceptedConnectionsHardLimit` number of
+      -- connections, later we c could be configured to something between
+      -- `acceptedConnesiontSoftLimit` and `acceptedConnectionsHardLimit`).
+      --
+    | HardLimit Word32
+
+
+-- | Interpretation of the 'AcceptedConnectionsLimit' policy.
+--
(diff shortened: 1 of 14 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
```diff
@@ -0,0 +1,156 @@
...
+                     -> AcceptedConnectionsLimit
+                     -- ^ limits
+                     -> RateLimitDelay
+getRateLimitDecision numberOfConnections
+                     AcceptedConnectionsLimit { acceptedConnectionsHardLimit
+                                              , acceptedConnectionsSoftLimit
+                                              , acceptedConnectionsDelay
+                                              }
+    -- below the soft limit we accept connections without any delay
+    | numberOfConnections  < hardLimit = NoRateLimiting
```
karknu: Shouldn't this be `numberOfConnetions < softLimit´?

ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
```diff
@@ -0,0 +1,156 @@
...
+
+    -- above the hard limit will will wait until the number of connections drops
+    -- below the soft limit
+    | numberOfConnections >= softLimit = HardLimit acceptedConnectionsHardLimit
```
karknu: Shouldn't this be `numberOfConnections >= HardLimit?

ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
```diff
@@ -0,0 +1,156 @@
...
+
+    -- in between we scale the delay using linear regression.
+    | otherwise =
+        SoftDelay $
+            fromIntegral (numberOfConnections - softLimit)
+          * acceptedConnectionsDelay
+          / fromIntegral (hardLimit - softLimit)
```
karknu: This looks like a potential division by zero problem. Where will you enforce that hardLimit /= softLimit?

ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
coot: Yes, that's why I should have added tests!

ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
coot: Yes

ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
coot: Good point; We don't enforce it, but we can take `max (hardLimit - softLimit) 1` here.

ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
```diff
@@ -0,0 +1,123 @@
...
+    --
+    acceptedConnectionsHardLimit :: !Word32,
+
+    -- | Soft limit of accepted connections.  If we are above this threshold,
+    -- we will start rate limiting.
+    --
+    acceptedConnectionsSoftLimit :: !Word32,
+
+    -- | Max delay for limiting accepted connections.  We use linear
+    -- regression starting from 0 at the soft limit up to
```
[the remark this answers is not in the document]
coot: Yes :D, I'll fix the docs.  I also think that we should give a way to configure a minimum delay (e.g. not start from `0`).

coot:
> Where do we set the actual limits for the node-to-node case? Will that have to be done in the cardano-node top level? Do we have any sane defaults?

Unfortunately so. I set current defaults here: https://github.com/input-output-hk/cardano-node/pull/705/files#diff-6f98f460a187cbfae2f479b840dc7ae4R355 It's 512 since we use one socket per connection and consensus is using at least one file descriptor too.

> What about that idea of using the relationship between the number of clients we have and the number of FDs we need, and doing the check that the configured hard limit would not put us over our OS limit on FDs? The relationship between the number of clients we have and the number of FDs we need is knowledge that comes from the consensus layer, so that should live there, along with an explanation of why it is the case. The FD limit check could be done in network subsystem initialisation.

This remains to be done.  I added an issue for this: https://github.com/input-output-hk/ouroboros-network/issues/1836  I think we can add a parameter to one of the consensus records, and use it to set the limits in `cardano-node`.

PR #1835, opened 2020-03-23
SRI-61 Add x86_64-w64-mingw32 tests checks to required job

jbgi:


release.nix
```diff
@@ -80,8 +80,10 @@ let
     #musl64 = mapTestOnCross musl64 (packagePlatformsCross project);
   } // (mkRequiredJob (concatLists [
     (collectTests jobs.x86_64-w64-mingw32.checks.tests)
-    (collectTests jobs.native.checks)
+    (collectTests jobs.native.checks.tests)
     (collectTests jobs.native.benchmarks)
+  ] ++ [
+    jobs.native.checks.validate-mainnet.x86_64-linux
```
mrBliss: Please see #1830. I expect this is slowing CI down.

mrBliss:
Can we try running `test-storage` on mingw32 with `--quickcheck-tests=10000` to be sure that the tests "always" pass? In the past, they *sometimes* failed in unrelated PRs (https://github.com/input-output-hk/ouroboros-network/pull/885). We don't want that again.

mrBliss:
(If it wasn't clear from my last message, running 10k tests is just a temporary thing. When we see that they all pass, we can revert it back to the default before we merge this PR.)

mrBliss:
It seems like all 10k tests [pass on Hydra](https://hydra.iohk.io/eval/878476), so you can drop the last commit and merge it :+1:

PR #1851, opened 2020-03-25
DONTMERGE: Add consensus to Windows CI

mrBliss:
Fixes #1847.

PR #1852, opened 2020-03-25
Hard-code expected results for golden files instead of using files

### LINKED ISSUE #1351: Golden tests can fail due to relative paths
I'm assuming this failure is due to relative paths and the process's working directory. Ideally the test wouldn't be sensitive to that.

```
$ cd ouroboros-network
...
    Golden tests
      ChainState:                                         FAIL
        Exception: test-consensus/golden/cbor/byron/ChainState: openBinaryFile: does not exist (No such file or directory)
        Exception: test-consensus/golden/cbor/byron/LedgerState: openBinaryFile: does not exist (No such file or directory)
      GenTxId:                                            FAIL
        Exception: test-consensus/golden/cbor/byron/GenTxId: openBinaryFile: does not exist (No such file or directory)
...
$ ls -1 ouroboros-consensus/test-consensus/golden/cbor/byron/
ChainState
GenTxId
LedgerState
```

After running `ln -s ouroboros-consensus/test-consensus`, the tests pass.

cc: @mrBliss

mrBliss:
Fixes #1351.

Using files to store the expected results for golden files, and thus relative paths, was a PITA: see #1351 + it was causing problems in #1851.

This PR introduces `Test.Util.Golden`, which takes the expected result as base58 bytestring. No more files, hurray!


ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/Golden.hs
```diff
@@ -0,0 +1,96 @@
...
+import           Data.Tuple (swap)
+
+import           Ouroboros.Consensus.Util (chunks)
+
+import           Test.Tasty.HUnit
+
+-- | Test whether the encoding of the given example value matches the expected
+-- binary encoding, given as a hexadecimal bytestring.
+--
+-- Hint: when writing the test, just pass an empty expected string.
```
nfrisby: Be still my beating heart! Thank you for anticipating and accomodating this need.

mrBliss:
> Tremendously well-organized commits! +1 That made it very easy to confirm nothing was (obviously) lost in the reorganization.

Thanks!

> The only improvement that comes to mind is to use a less verbose encoding, such as https://en.wikipedia.org/wiki/Base58 -- but that's minor and I'm no expert here.

I like that idea, the hex bytestrings were rather long, this helps a bit.

I was also thinking of using [`FlatTerm`](https://hackage.haskell.org/package/cborg-0.2.2.1/docs/Codec-CBOR-FlatTerm.html) instead of actual bytestrings, I think that would help with figuring out what changed.

mrBliss:
See #1855 for an alternative approach using CBOR's `FlatTerm` instead of a base58 bytestring for the expected result.

I'm going to sleep on it and make a decision tomorrow.

mrBliss:
Closing in favour of #1855.

PR #1855, opened 2020-03-25
Hard-code expected results as CBOR for golden files instead of using files

### LINKED ISSUE #1351: Golden tests can fail due to relative paths

mrBliss:
Fixes #1351.

Using files to store the expected results for golden files, and thus relative paths, was a PITA: see #1351 + it was causing problems in #1851.

This PR introduces `Test.Util.Golden`, which takes the expected result as a CBOR `FlatTerm`. No more files, hurray! Using a `FlatTerm` will make diffing much easier.


mrBliss:
This is an alternative approach to #1852. The `FlatTerm` is more verbose, making the files much longer, but when it changes, the diff will be much clearer (I'm using http://hackage.haskell.org/package/tree-diff for diffing).

nfrisby:
The more transparent diff seem like a significant advantage to me.

To mitigate the file length -- at the cost of the transparency of the git diff -- you could only force newlines after `TkBytes`, letting the other token ctors accumulate on a single ~79 character line. Or even just let all token ctors accumulate, in case there's a short `TkBytes`. I don't have much experience inspecting flatterms; maybe that presentation would be far less helpful.

FWIW, the long files don't seem too problematic. If it becomes an issue, we can separate egregiuos offenders into their own modules.

mrBliss:
> The more transparent diff seem like a significant advantage to me.
> 
> To mitigate the file length -- at the cost of the transparency of the git diff -- you could only force newlines after `TkBytes`, letting the other token ctors accumulate on a single ~79 character line. Or even just let all token ctors accumulate, in case there's a short `TkBytes`. I don't have much experience inspecting flatterms; maybe that presentation would be far less helpful.
> 
> FWIW, the long files don't seem too problematic. If it becomes an issue, we can separate egregiuos offenders into their own modules.

Yeah, I think the benefits outweigh the costs, indeed. I'm not keen on letting them accumulate on a line, because that either means manually wrapping lines or making the formatter more complex so that the copy-pastable output already does the wrapping.

mrBliss:
Uggh, a timing test failed because CI is overloaded.

mrBliss:
bors retry

PR #1863, opened 2020-03-27
Trace transaction replies in txSubmissionOutbound after transactions are sent

intricate:
I noticed that in #1688 I was mistakenly tracing `TraceTxSubmissionOutboundSendMsgReplyTxs` events _before_ the transactions had been successfully sent.


ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
```diff
@@ -189,7 +189,7 @@ txSubmissionOutbound tracer maxUnacked TxSubmissionMempoolReader{..} =
               !unackedMap' = foldl' (flip Map.delete) unackedMap txids
               client'      = client unackedSeq unackedMap' lastIdx
 
-          -- Trace the transactions to be sent in the response.
-          traceWith tracer (TraceTxSubmissionOutboundSendMsgReplyTxs txs)
+              -- Trace the transactions to be sent in the response.
+              traceTxsSent = traceWith tracer (TraceTxSubmissionOutboundSendMsgReplyTxs txs)
```
mrBliss: This requires a comment (here or near L195) saying that tracing itself will be executed after sending the reply.

intricate:
> I see. Tracing afterwards makes sense, but I'm not sure it's worth making the code that much more complex...

@mrBliss: Do you feel the same way about the approach in the first commit?

```
data ClientStTxs txid tx m a where
  SendMsgReplyTxs   :: [tx]
                    -> ([tx] -> m ())
                    -> ClientStIdle txid tx m a
                    -> ClientStTxs  txid tx m a
```

PR #1864, opened 2020-03-27
Bump dependency on cardano-ledger-specs

mrBliss:
This incorporates:
* cardano-ledger-specs#1313 which simplifies our implementation of
  `shelleyAddHeaderEnvelope`.
* cardano-ledger-specs#1342 which fixes a generator that sometimes caused a
  serialisation roundtrip property test to fail (see
  https://hydra.iohk.io/build/2157662/nixlog/28).


nc6:
bors merge

PR #1865, opened 2020-03-27
Ntp-client changes

### LINKED ISSUE #1548: Test NTP client under Windows

### LINKED ISSUE #1785: ntp-client does not work on Windows
While integrating ntp-client with cardano-wallet https://github.com/input-output-hk/cardano-wallet/issues/1394 we have learnt that ntp-client currently does not work on Windows platform (tested on Windows 10 VM).

The error reported is:
```
ntp client experienced error Network.Socket.recvBufFrom: failed (Invalid argument (WSAEINVAL)) when using IPv4 protocol
```

Please also note that, while working on the integration we also had observation that the `ouroboros-network/ntp-client` library is missing tests on error paths -> https://github.com/input-output-hk/cardano-wallet/issues/1394#issuecomment-595084104

coot:
This PR does a few things:

* make ntp-client use Win32-network library
* fixes #1785 - the crux of the error is that one needs to bind before using `recvFrom` on Windows
* fixes #1548 - test the ntp-client on Windows
* fixing concurrency error handling:
   when running the ntp client over ipv4 and ipv6 `waitCatch` on both threads
   instead of using `concurrently` which would mean if one thread fails (e.g.
   ipv6 sending fails) both thread will fail and wouldn't get any results.
* make logging information more useful for debugging purposes    
* a bit of cleaning
* documentation improvments


MarcFontaine:
> fixing concurrency error handling:
> when running the ntp client over ipv4 and ipv6 waitCatch on both threads
> instead of using concurrently which would mean if one thread fails (e.g.
> ipv6 sending fails) both thread will fail and wouldn't get any results.

I think the old Version was also OK.
In the old version, both treads that were started `concurrently` did there own timeout and error handling.
For sure the old version could handle the case that either IPv4 or IPv6 failed.
I had tested for that.
Or more precisely I mostly tested without an IPv6 network setup
an IPv6 failed and only IPv4 was used.

ntp-client/src/Network/NTP/Client/Query.hs
```diff
@@ -155,20 +155,15 @@ ntpQuery ioManager tracer ntpSettings = do
...
-            traceWith tracer NtpTraceIPv4IPv6NoReplies
-            return NtpSyncUnavailable
-        l -> case minimumOfThree l of
-            Nothing -> do
-                traceWith tracer NtpTraceReportPolicyQueryFailed
-                return NtpSyncUnavailable
-            Just offset -> do
-                traceWith tracer $ NtpTraceQueryResult $ getNtpOffset offset
-                return $ NtpDrift offset
+    withAsync (runProtocol IPv4 v4LocalAddr v4Servers) $ \ipv4Async ->
```
MarcFontaine: These are all very non-trivial changes.
`waitCatch` catches any exception.
I don't think that is wanted here.
The original version would propagate the exceptions
and these exception were handled in an other place.
As mentioned before, I think using `concurrently` is fine because the threads that are started
`concurrently` use timeouts and catch IOErrors.

coot:
> In the old version, both treads that were started concurrently did there own timeout and error handling.
You're right there was `tryIOError`; Still, I think it's cleaner to handle errors as at the top of the stack, it could be something else why I didn't see any progress and only exceptions indicating that one cannot send `ipv6` from my box.

ntp-client/src/Network/NTP/Client/Query.hs
coot: > `waitCatch` catches any exception.
> I don't think that is wanted here.

This can be changed to avoid caching async exceptions, only other exception that can happen are `IOException`s. Still we log the result, previously we hand at hock error handing at call sites, now its at the top level. I think that's an improvement.

The debatable difference is handling exceptions when we send request.  Now we will retry for all addresses instead of throwing  exception and not sending any more requests on first error.  This has pros and cons, but since we send just a handful of request (e.g. 3), then there's not much cost in doing so.

> concurrently use timeouts and catch IOErrors.

`concurrently` per se has no timeout, but I think you mean that the `runNtpQueries` have timeout (using async exceptions). This will still work the same way, I haven't changed that. We still do `waitAny` on the timeout thread and the receiving thread.

coot:
@MarcFontaine I pushed a fix. Thanks that was a good catch. The problem could be catching `AsyncException`s (in particular `UserInterrupt`).

ntp-client/src/Network/NTP/Client/Query.hs
```diff
@@ -205,17 +205,10 @@ runNtpQueries ioManager tracer protocol netSettings localAddr destAddrs
...
-            -- callbak either returns or errors.
-            mask $ \unmask ->
-              async (unmask $ send socket) >>= \senderT -> unmask $
-                waitCatch senderT >>= \case
-                  Left e  -> throwIO e
-                  Right _ -> void $ waitAny [timeoutT, receiverT]
+        withAsync timeout $ \timeoutAsync ->
+          withAsync (receiver socket inQueue) $ \receiverAsync -> do
+            send socket
+            void $ waitAny [timeoutAsync, receiverAsync]
```
MarcFontaine: The idea was to also guard against a send that is blocked indefinitely with the timeout.
I don't know if send can block forever.

ntp-client/src/Network/NTP/Client/Query.hs
MarcFontaine: Right `concurrently` starts `runProtocol` and `runProtocol` does timeout and exception handling.

coot:
You could use `(Either IOException a, Either IOException a)` but `ResultOrFailure` is tighter.  The type constructor are very descriptive:
```
data ResultOrFailure e a
    = BothSucceeded !a
    -- ^ both threads suceeded
    | SuccessAndFailure !a !IPVersion !e
    -- ^ one of the threads errors. 'IPVersion' indicates which one.
    | BothFailed !e !e
```

This captures all the information we need for two purposes:

* logging, just logging a `term :: ResultOrFailure` will make it clear what happened.
* getting the results with the standard `fold :: ResultOrFailure e a -> a`

You're right, that it's  better to move  `mkResultOrFailure` to the function definition, it's quite specific to how the threads are combined.

coot:
Since I removed the `Trace` module, all of the `ResultOrFailure` can be local now in the `Query` module :).

ntp-client/src/Network/NTP/Client/Query.hs
coot: That would be odd. We send small number of small packets. This was protecting against async exceptions being thrown from outside, not the timeouts - timeout thread woudld need to throw an async exception but it just terminates.

coot:
bors: merge

PR #1879, opened 2020-03-31

nc6:
Slightly mixed PR:

- Adds a comment explaining the confusion I had in a meeting this morning
- Updates to the latest cardano-ledger-specs, which means we can drop the requirement for an initial "previous key hash" in nonce calculations.


nc6:
Looks like after #1864 this will just be a comment!

nc6:
bors merge

PR #1886, opened 2020-04-01
Haddocks on github-pages pushed by github-actions

coot:


scripts/haddocs.sh
```diff
@@ -30,16 +30,25 @@ else
 fi
 
 
+# we don't include `--use-index` option, because then quickjump data is not
+# generated.  This is not ideal, but there is no way to
+# * generate only top level `doc-index.json` file
+# * generate only top level `doc-index.html` file
```
MarcFontaine, line 36: Doubled line

scripts/haddocs.sh
```diff
@@ -0,0 +1,68 @@
...
+
+GHC_VERSION=$(ghc --numeric-version)
+
+if [ ${OUTPUT_DIR} == "/" ]; then
+  echo "Oh no, you don't want to run 'rm -rf /*'"
+  exit 1
+else
+  mkdir -p ${OUTPUT_DIR}
+  rm -rf ${OUTPUT_DIR}/*
+fi
```
MarcFontaine, line 30: This looks so scary.
I only found a single place where `scripts/haddocs.sh` is called and at the call-site `${OUTPUT_DIR}` is generated from scratch anyway.
Maybe just add a test that `${OUTPUT_DIR}` is empty here and exit the script if that precondition is not met.

scripts/prolog
```diff
@@ -0,0 +1,65 @@
...
+
+* __Ouroboros-Network__  - networking component of the Haskell implementation of Ouroboros protocols
+* __Ouroboros-Consensus__ - Ouroboros Consensus family of protocol which integrate with `Ouroboros-Network`
+
+== Module Structure And Network Components
+
+* __IOSim & Monad Classes__
+
+    * __[io-sim](io-sim/Control-Monad-IOSim.html)__ - `IOSim` moand which allows us to build all the
+      networking in simulation
```
MarcFontaine, line 15: `IOSim monad` ... allows us to _run_ the  network in simulation

scripts/prolog
```diff
@@ -0,0 +1,65 @@
...
+
+              * __[ChainSync](ouroboros-network/Ouroboros-Network-Protocol-ChainSync-Type.html)__
+              * __[BlockFetch](ouroboros-network/Ouroboros-Network-Protocol-BlockFetch-Type.html)__
+              * __[TxSubmission](ouroboros-network/Ouroboros-Network-Protocol-TxSubmission-Type.html)__
+              * __[LocalTxSubmission](ouroboros-network/Ouroboros-Network-Protocol-LocalTxSubmission-Type.html)__
+              * __[LocalStateQuery](ouroboros-network/Ouroboros-Network-Protocol-LocalStateQuery-Type.html)__
+              * __[LocalTxMonitor](ouroboros-network/Ouroboros-Network-Protocol-LocalTxMonitor-Type.html)__
+              * __[Handshake](ouroboros-network-framework/Ouroboros-Network-Protocol-Handshake-Type.html)__
+          
+         * /Block chain related data structures that the networking code needs to be aware:/
```
MarcFontaine, line 46: needs to be aware of

scripts/prolog
```diff
@@ -0,0 +1,65 @@
...
+    * __Win32-network__
+    * __[typed-protocols](typed-protocols/Network-TypedProtocol.html)__
+    * __[network-mux](network-mux/Network-Mux.html)__
+
+* __Ouroboros.Network__
+
+    * __ouroboros-network-framework__ - our instantiations and implementation
+      of servers, clients, abstractions for working with named pipes and unix
+      sockets.  These are higher level components, which are independent of
+      ouroboros protocols, but build a solid foundation for the networking
```
MarcFontaine, line 30: which don't depend on the ouroboros protocols

scripts/haddocs.sh
coot: I'll fix this.

scripts/haddocs.sh
coot: Right, this was actually meant to work slightly differently. I'll fix this - though this still will be a bit scary.

coot: > Maybe just add a test that ${OUTPUT_DIR} is empty here and exit the script if that precondition is not met.

We don't need to, we create a directory with `mkdir` - it will fail if it already exists.

I wanted to remove the directory to avoid stale files, but we can just avoid doing this; The worse that can happen is that there will have some `.html` pages that are not linked from anywhere.

coot:
bors merge

PR #1891, opened 2020-04-02
Reinit the dns lib before each call

karknu:
The dns library only reads /etc/resolv.conf once which means that inorder to
cope with changing networks conditions we need to reinit the library
before each use.

Workaround for the bug reported in https://github.com/input-output-hk/cardano-node/issues/731

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Subscription/Dns.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Subscription/Dns.hs
@@ -28,7 +28,6 @@ module Ouroboros.Network.Subscription.Dns
     ) where
 
 import           Control.Monad.Class.MonadAsync
-import           Control.Monad.Class.MonadSTM (TVar)
 import qualified Control.Monad.Class.MonadSTM as Lazy
 import           Control.Monad.Class.MonadSTM.Strict
 import           Control.Monad.Class.MonadSay
@@ -108,7 +107,7 @@ dnsResolve tracer getSeed withResolver peerStatesVar beforeConnect (DnsSubscript
     withResolver rs $ \resolver -> do
         ipv6Rsps <- newEmptyTMVarM
         ipv4Rsps <- newEmptyTMVarM
-        gotIpv6Rsp <- Lazy.newTVarM False
+        gotIpv6Rsp <- newTVarM False
 
         -- Though the DNS lib does have its own timeouts, these do not work
         -- on Windows reliably so as a workaround we add an extra layer
@@ -215,7 +214,7 @@ dnsResolve tracer getSeed withResolver peerStatesVar beforeConnect (DnsSubscript
         case r_e of
              Left e  -> do
                  atomically $ putTMVar rspsVar []
-                 atomically $ Lazy.writeTVar gotIpv6RspVar True
+                 atomically $ writeTVar gotIpv6RspVar True
                  traceWith tracer $ DnsTraceLookupAAAAError e
                  return $ Just e
              Right r -> do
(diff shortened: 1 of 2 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Subscription/Dns.hs
```diff
@@ -77,32 +97,34 @@ dnsResolve :: forall m s.
...
-        return (SubscriptionTarget $ pure Nothing)
-      Just st ->
-        return st
-
+dnsResolve tracer getSeed withResolver peerStatesVar beforeConnect (DnsSubscriptionTarget domain _ _) = do
+    rs <- getSeed
+    withResolver rs $ \resolver -> do
+        ipv6Rsps <- newEmptyTMVarM
+        ipv4Rsps <- newEmptyTMVarM
+        gotIpv6Rsp <- Lazy.newTVarM False
```
coot, line 111: If we are modifying these lines, why not also change to `StrictTVar`.

ouroboros-network-framework/src/Ouroboros/Network/Subscription/Dns.hs
```diff
@@ -77,32 +97,34 @@ dnsResolve :: forall m s.
...
+
+        -- Though the DNS lib does have its own timeouts, these do not work
+        -- on Windows reliably so as a workaround we add an extra layer
+        -- of timeout on the outside.
+        -- TODO: fix upstream dns lib
+        res <- timeout 20 $ do
+                 aid_ipv6 <- async $ resolveAAAA resolver gotIpv6Rsp ipv6Rsps
+                 aid_ipv4 <- async $ resolveA resolver gotIpv6Rsp ipv4Rsps
```
coot, line 119: I noticed a missing comment here: if an `async` exception will hit the main thread here, the two threads will not be killed. Though we should not do this kind of change lightly, since the dns library on Windows is not using our interruptible `Win32-network`.  @karknu could you add a note here (or I can do that if you wish, in between of waiting for things to compile./ run).

ouroboros-network-framework/src/Ouroboros/Network/Subscription/Dns.hs
```diff
@@ -237,14 +259,16 @@ dnsSubscriptionWorker'
...
+    -> (b -> (Resolver IO -> IO (SubscriptionTarget IO Socket.SockAddr))
+          -> IO (SubscriptionTarget IO Socket.SockAddr))
     -> DnsSubscriptionParams a
     -> Main IO (PeerStates IO Socket.SockAddr) x
     -> (Socket.Socket -> IO a)
     -> IO x
 dnsSubscriptionWorker' snocket subTracer dnsTracer errorPolicyTracer
                        networkState@NetworkMutableState { nmsPeerStates }
-                       resolver
+                       setupResolver resolver
```
coot, line 271: where does `setupResolver` comes from? I just couldn't find it in this patch and the `dns` library.

ouroboros-network-framework/src/Ouroboros/Network/Subscription/Dns.hs
karknu: Will do

ouroboros-network-framework/src/Ouroboros/Network/Subscription/Dns.hs
karknu: `setupResolver` is the name of an argument, (DNS.makeResolvSeed DNS.defaultResolvConf) in `dnsSubscriptionWorker'`.

PR #1894, opened 2020-04-02
Export static codecs

MarcFontaine:
Export a record with all NodeToClient codecs,
These codecs (currently) do not depend on ToplevelConfig.
This PR makes these codecs usable in the wallet.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeNetwork.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeNetwork.hs
@@ -18,6 +18,8 @@ module Ouroboros.Consensus.NodeNetwork (
   , ProtocolCodecs (..)
   , protocolCodecs
   , protocolCodecsId
+  , NodeToClientCodecs (..)
+  , nodeToClientCodecs
   , ProtocolTracers
   , ProtocolTracers' (..)
   , nullProtocolTracers
@@ -273,21 +275,45 @@ protocolCodecs cfg version = ProtocolCodecs {
           nodeEncodeGenTx
           nodeDecodeGenTx
 
-    , pcLocalChainSyncCodec =
+    , pcLocalChainSyncCodec    = nodeToClientChainSyncCodec ntcCodecs
+
+    , pcLocalTxSubmissionCodec = nodeToClientTxSubmissionCodec ntcCodecs
+
+    , pcLocalStateQueryCodec   = nodeToClientStateQueryCodec ntcCodecs
+    }
+    where
+       ntcCodecs = nodeToClientCodecs NodeToClientV_1
+
+data NodeToClientCodecs blk failure m bytesLCS bytesLTX bytesLSQ = NodeToClientCodecs {
+    nodeToClientChainSyncCodec    :: Codec (ChainSync (Serialised blk) (Tip blk))
+                                           failure m bytesLCS
+  , nodeToClientTxSubmissionCodec :: Codec (LocalTxSubmission (GenTx blk) (ApplyTxErr blk))
+                                           failure m bytesLTX
+  , nodeToClientStateQueryCodec   :: Codec (LocalStateQuery blk (Query blk))
+                                           failure m bytesLSQ
+  }
+
+nodeToClientCodecs
+    :: forall m blk. (IOLike m, RunNode blk)
+    => NodeToClientVersion
+    -> NodeToClientCodecs blk DeserialiseFailure m ByteString ByteString ByteString
+nodeToClientCodecs NodeToClientV_1 = NodeToClientCodecs {..}
+  where
+    nodeToClientChainSyncCodec =
         codecChainSyncSerialised
           (encodePoint (nodeEncodeHeaderHash (Proxy @blk)))
           (decodePoint (nodeDecodeHeaderHash (Proxy @blk)))
           (encodeTip   (nodeEncodeHeaderHash (Proxy @blk)))
           (decodeTip   (nodeDecodeHeaderHash (Proxy @blk)))
 
-    , pcLocalTxSubmissionCodec =
+    nodeToClientTxSubmissionCodec =
         codecLocalTxSubmission
           nodeEncodeGenTx
           nodeDecodeGenTx
           (nodeEncodeApplyTxError (Proxy @blk))
           (nodeDecodeApplyTxError (Proxy @blk))
 
-    , pcLocalStateQueryCodec =
+    nodeToClientStateQueryCodec =
         codecLocalStateQuery
           (encodePoint (nodeEncodeHeaderHash (Proxy @blk)))
           (decodePoint (nodeDecodeHeaderHash (Proxy @blk)))
@@ -295,7 +321,6 @@ protocolCodecs cfg version = ProtocolCodecs {
           nodeDecodeQuery
           nodeEncodeResult
           nodeDecodeResult
-    }
 
 -- | Id codecs used in tests.
 --


ouroboros-consensus/src/Ouroboros/Consensus/NodeNetwork.hs
```diff
@@ -259,42 +264,72 @@ protocolCodecs cfg version = ProtocolCodecs {
...
-    , pcLocalStateQueryCodec =
-        codecLocalStateQuery
-          (encodePoint (nodeEncodeHeaderHash (Proxy @blk)))
-          (decodePoint (nodeDecodeHeaderHash (Proxy @blk)))
-          nodeEncodeQuery
-          nodeDecodeQuery
-          nodeEncodeResult
-          nodeDecodeResult
+    , pcLocalStateQueryCodec      = staticLocalStateQueryCodec
     }
```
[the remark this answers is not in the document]
MarcFontaine: Yes, the idea is to replace the locally defined codecs in the Wallet
and cardano-rest and other places with these static-Versions of the codecs.
If in the future one of codecs becomes configurable,
the codec has to be removed from the exports.
This will break the libraries that uses this codec and the problem will be noticed.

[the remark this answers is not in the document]
coot: @mrBliss let's keep this part as simple as possible.  The requirement here is to make it as easy as possible to run and maintain client applications (good example is `cardano-cli get-tip`, it just needs to know the block type).  Since `node-to-client` protocols don't depend on `TopLevelConfig` let's not include it.  Do you see some dependency in forcible future? Will the `LocalStateQuery` or `LocalTxMonitor` protocols depend on parts of `TopLevelConfig`?  If so I think it's good to start with no dependency, and then add only the necessary dependencies. This will make us think how to properly do it when we will need it.

@MarcFontaine these codes should be versioned by `NodeToClientVersion`.  There's only one version now, but still it's better to put that already in place. I guess, they no longer become static.

ouroboros-consensus/src/Ouroboros/Consensus/NodeNetwork.hs
```diff
@@ -259,42 +264,67 @@ protocolCodecs cfg version = ProtocolCodecs {
...
 
+staticBlockFetchCodecSerialised
+    :: forall m blk. (IOLike m, RunNode blk)
+    => Codec (BlockFetch (Serialised blk)) DeserialiseFailure m ByteString
+staticBlockFetchCodecSerialised
+    = codecBlockFetchSerialised
+        (nodeEncodeHeaderHash (Proxy @blk))
+        (nodeDecodeHeaderHash (Proxy @blk))
```
mrBliss, line 284: BTW, the style is:
```suggestion
staticBlockFetchCodecSerialised =
    codecBlockFetchSerialised
      (nodeEncodeHeaderHash (Proxy @blk))
      (nodeDecodeHeaderHash (Proxy @blk))
```

ouroboros-consensus/src/Ouroboros/Consensus/NodeNetwork.hs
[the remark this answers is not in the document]
mrBliss: I guess it's fine for now. But as soon as they *do* start depending on the `TopLevelConfig`, I don't want to be blocked by these `static*` definitions.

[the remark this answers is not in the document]
coot: I actually want us to be blocked by them - this will give us moment to think how to make it as simple as possible for the clients as well.

coot:
@MarcFontaine I think it's worth putting all the client codecs in a recrod.  Check this out how a client is instantiated: https://github.com/input-output-hk/cardano-node/blob/master/cardano-node/src/Cardano/CLI/Ops.hs#L406 it needs all the codecs, even if one of the mini-protocols is not sending anything.

Could you go around various clients and check that we are right that there are clients that do not need `TopLevelConfig`?
* `cardano-wallet`
* `cardano-rest`
*  clients in `cardano-node`

MarcFontaine:
The wallet uses:
``` 
codec :: Codec protocol DeserialiseFailure m ByteString
codec = codecChainSync
        encodeByronBlock
        (unwrapCBORinCBOR $ decodeByronBlock (toEpochSlots getEpochLength))
        (encodePoint encodeByronHeaderHash)
        (decodePoint decodeByronHeaderHash)
        (encodeTip encodeByronHeaderHash)
        (decodeTip decodeByronHeaderHash)
```
TopLevelConfig is not available in the wallet

MarcFontaine:
cardano-rest uses:
```
localTxSubmissionCodec
￼  :: forall m blk . (RunNode blk, MonadST m)
￼  => Codec (LocalTxSubmission (GenTx blk) (ApplyTxErr blk)) DeserialiseFailure m LBS.ByteString
￼localTxSubmissionCodec =
￼  codecLocalTxSubmission
￼    nodeEncodeGenTx
￼    nodeDecodeGenTx
￼    (nodeEncodeApplyTxError (Proxy @blk))
￼    (nodeDecodeApplyTxError (Proxy @blk))
￼```
But `TopLevelConfig` is available.

**Comment by coot** [comment]
ok, good enough evindence for me we should avoid `TopLevelConfig`

File: ouroboros-consensus/src/Ouroboros/Consensus/NodeNetwork.hs [hunk 3]
```diff
@@ -261,42 +263,86 @@ protocolCodecs cfg version = ProtocolCodecs {
...
+    , pcLocalStateQueryCodec      = staticLocalStateQueryCodec
     }
 
+data NodeToClientCodecs blk failure m bytesLCS bytesLTX bytesLSQ = NodeToClientCodecs {
+    nodeToClientChainSyncCodec       :: Codec (ChainSync (Serialised blk) (Tip blk))
+                                         failure m bytesLCS
+  , nodeToClientTxSubmissionCodec    :: Codec (LocalTxSubmission (GenTx blk) (ApplyTxErr blk))
+                                         failure m bytesLTX
+  , nodeToClientStateQueryCodec      :: Codec (LocalStateQuery blk (Query blk))
+                                         failure m bytesLSQ
```
mrBliss, line 283: Are these `::` aligned with those in `ProtocolCodecs`? That's not necessary, just align these three with each other, like `stylish-haskell` does.

ouroboros-consensus/src/Ouroboros/Consensus/NodeNetwork.hs
```diff
@@ -277,55 +274,74 @@ protocolCodecs cfg version = ProtocolCodecs {
...
+    :: forall m blk. (IOLike m, RunNode blk)
+    => NodeToClientVersion
+    -> NodeToClientCodecs blk DeserialiseFailure m ByteString ByteString ByteString
+nodeToClientCodecs NodeToClientV_1 = NodeToClientCodecs {
+    nodeToClientChainSyncCodec    = staticLocalChainSyncCodec
+  , nodeToClientTxSubmissionCodec = staticLocalTxSubmissionCodec
+  , nodeToClientStateQueryCodec   = staticLocalStateQueryCodec
+  }
+
 staticBlockFetchCodecSerialised
```
coot, line 296: Please hide them in a `where` clause`, then you can rename them as `nodeToClientCHainSyncCodec`, ... and take advantage of `NamedFieldPuns` extension:
```
nodeToClientCodecs NodeToClientV_1 = NodeToClientCodecs {
      nodeToClientChainSyncCodec
    , nodeToClientTxSubmissionCodec
    , nodeToClientStateQueryCodec
    }
  where
    nodeToClientChainSyncCodec = ...
    nodeToClientTxSubmissionCodec = ...
    nodeToClientStateQueryCodec = ...
```

MarcFontaine: At the moment the same codecs are exported via two different ways via `ProtocolCodecs` and via `NodeToClientCodecs`.
Maybe the nodeToClient codecs should be removed from ProtocolCodecs altogther ?
`staticLocalChainSyncCodec` is the single definition that is uses for both exports.

coot: I see; you should have access to `NodeToClientV_1` there too, no? Then you can unpack the record and add the codes.  This way in the future one will only need to modify `nodeToClientCodecs` to update codecs in consensus.

MarcFontaine: OK! Done

PR #1898, opened 2020-04-03
Strengthen the hard fork history tests, fix bug

edsko:
We now actually test that we can accurately predict future events. This
showed up an off-by-one error in the implementation. Yay for tests :)


edsko:
> Reviewed over Meet. Fried my brain this noon grin

Yeah, I know how you feel!

PR #1906, opened 2020-04-06
Change filelock behaviour

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/1903

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
@@ -1,30 +1,65 @@
-{-# LANGUAGE LambdaCase        #-}
-{-# LANGUAGE OverloadedStrings #-}
-
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 module Ouroboros.Consensus.Node.DbLock (
     DbLocked (..)
   , withLockDB
-  , dbLockFile
+    -- * Defaults
+  , dbLockFsPath
+  , dbLockTimeout
+    -- * For testing purposes
+  , withLockDB_
   ) where
 
 import           Control.Monad.Class.MonadTimer
-import           Data.Text (Text)
 import qualified Data.Time.Clock as Time
-import           System.FileLock
 
 import           Ouroboros.Consensus.Storage.FS.API
 import           Ouroboros.Consensus.Storage.FS.API.Types
+import           Ouroboros.Consensus.Util.FileLock
 import           Ouroboros.Consensus.Util.IOLike
 
--- | We use an empty file as a lock of the db so that the database cannot be
--- opened by more than one process. We wait up to 2 seconds to take the lock,
--- before timing out and throwing a 'DbLocked' exception. Some systems may
--- delete the empty file when all its handles are closed. This is not an issue,
--- since the file is created if it doesn't exist.
-withLockDB :: HasFS IO h ->  FilePath -> IO a -> IO a
-withLockDB hasFS dbPath action = do
+-- | We use an empty file ('dbLockFsPath') as a lock of the database so that
+-- the database cannot be opened by more than one process. We wait up to
+-- 'dbLockTimeout' to take the lock, before timing out and throwing a
+-- 'DbLocked' exception.
+withLockDB :: HasFS IO h -> MountPoint -> IO a -> IO a
+withLockDB hasFS mountPoint =
+    withLockDB_
+      ioFileLock
+      hasFS
+      mountPoint
+      dbLockFsPath
+      dbLockTimeout
+
+-- | The default lock file
+dbLockFsPath :: FsPath
+dbLockFsPath = fsPathFromList ["lock"]
+
+-- | Default time to wait on the lock
+dbLockTimeout :: DiffTime
+dbLockTimeout = Time.secondsToDiffTime 2
+
+-- | We use the given 'FsPath' in the 'MountPoint' as a lock of the database
+-- so that the database cannot be opened by more than one process. We wait the
+-- given 'DiffTime' on the thread taking the lock. In case of a timeout, we
+-- throw a 'DbLocked' exception.
+--
+-- Some systems may delete the empty file when all its handles are closed.
+-- This is not an issue, since the file is created if it doesn't exist.
+withLockDB_
+  :: forall m h a. (IOLike m, MonadTimer m)
+  => FileLock m
+  -> HasFS m h
+  -> MountPoint  -- ^ Root of the path
+  -> FsPath      -- ^ File to lock
+  -> DiffTime    -- ^ Timeout
+  -> m a
+  -> m a
+withLockDB_ fileLock hasFS mountPoint lockFsPath lockTimeout action = do
     createDirectoryIfMissing hasFS True root
-    bracket acquireLock unlockFile (const action)
+    bracket acquireLock id (const action)
   where
     -- We want to avoid blocking the main thread at an uninterruptible ffi, to
     -- avoid unresponsiveness to timeouts and ^C. So we use 'async' and let a
@@ -33,25 +68,21 @@ withLockDB hasFS dbPath action = do
     -- We shouldn't be tempted to use 'withAsync', which is usually mentioned
     -- as a better alternative, or try to synchronously cancel the forked
     -- thread during cleanup, since this would block the main thread and negate
-    -- the whole point of using 'async'. We try our best to clean up resources,
-    -- but this may not be entirely possible, because of the uninterruptible
-    -- system call. This is not a big issue though, since if we fail to take the
-    -- lock succesfully, the whole process will soon die.
-    acquireLock :: IO FileLock
+    -- the whole point of using 'async'.
+    --
+    -- This means that we leave the thread taking the lock running in case of
+    -- a timeout. This is not a problem, though, since if we fail to take the
+    -- lock, the whole process will soon die.
+    acquireLock :: m (m ())
     acquireLock = do
-      lockFileAsync <- async (lockFile lockFilePath Exclusive)
-      mbLock <- timeout (Time.secondsToDiffTime 2) $ wait lockFileAsync
-      case mbLock of
+      lockFileAsync <- async (lockFile fileLock lockFilePath)
+      timeout lockTimeout (wait lockFileAsync) >>= \case
         -- We timed out while waiting on the lock. The db is still locked.
-        Nothing   -> throwM $ DbLocked lockFilePath
-        Just lock -> return lock
+        Nothing     -> throwM $ DbLocked lockFilePath
+        Just unlock -> return unlock
 
     root         = mkFsPath []
-    pFile        = fsPathFromList [dbLockFile]
-    lockFilePath = fsToFilePath (MountPoint dbPath) pFile
-
-dbLockFile :: Text
-dbLockFile = "lock"
+    lockFilePath = fsToFilePath mountPoint lockFsPath
 
 newtype DbLocked = DbLocked FilePath
     deriving (Eq, Show)
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
@@ -1,0 +1,61 @@
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Ouroboros.Consensus.Node.DbLock (
+    DbLocked (..)
+  , withLockDB
+  , dbLockFile
+  ) where
+
+import           Control.Monad.Class.MonadTimer
+import           Data.Text (Text)
+import qualified Data.Time.Clock as Time
+import           System.FileLock
+
+import           Ouroboros.Consensus.Storage.FS.API
+import           Ouroboros.Consensus.Storage.FS.API.Types
+import           Ouroboros.Consensus.Util.IOLike
+
+-- | We use an empty file as a lock of the db so that the database cannot be
+-- opened by more than one process. We wait up to 2 seconds to take the lock,
+-- before timing out and throwing a 'DbLocked' exception. Some systems may
+-- delete the empty file when all its handles are closed. This is not an issue,
+-- since the file is created if it doesn't exist.
+withLockDB :: HasFS IO h ->  FilePath -> IO a -> IO a
+withLockDB hasFS dbPath action = do
+    createDirectoryIfMissing hasFS True root
+    bracket acquireLock unlockFile (const action)
+  where
+    -- We want to avoid blocking the main thread at an uninterruptible ffi, to
+    -- avoid unresponsiveness to timeouts and ^C. So we use 'async' and let a
+    -- new thread do the actual ffi call.
+    --
+    -- We shouldn't be tempted to use 'withAsync', which is usually mentioned
+    -- as a better alternative, or try to synchronously cancel the forked
+    -- thread during cleanup, since this would block the main thread and negate
+    -- the whole point of using 'async'. We try our best to clean up resources,
+    -- but this may not be entirely possible, because of the uninterruptible
+    -- system call. This is not a big issue though, since if we fail to take the
+    -- lock succesfully, the whole process will soon die.
+    acquireLock :: IO FileLock
+    acquireLock = do
+      lockFileAsync <- async (lockFile lockFilePath Exclusive)
+      mbLock <- timeout (Time.secondsToDiffTime 2) $ wait lockFileAsync
+      case mbLock of
+        -- We timed out while waiting on the lock. The db is still locked.
+        Nothing   -> throwM $ DbLocked lockFilePath
+        Just lock -> return lock
+
+    root         = mkFsPath []
+    pFile        = fsPathFromList [dbLockFile]
+    lockFilePath = fsToFilePath (MountPoint dbPath) pFile
+
+dbLockFile :: Text
+dbLockFile = "lock"
+
+newtype DbLocked = DbLocked FilePath
+    deriving (Eq, Show)
+
+instance Exception DbLocked where
+    displayException (DbLocked f) =
+      "The db is used by another process. File \"" <> f <> "\" is locked"
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs
@@ -52,9 +52,7 @@
 -- = Concurrency
 --
 -- The same database should not be opened multiple times concurrently.
---
--- TODO Should we ensure this with a lock file?
--- https://hackage.haskell.org/package/filelock-0.1.1.2/docs/System-FileLock.html
+-- This is ensured by the file lock of the ChainDB.
 --
 -- The database can have multiple readers, but should only have one writer.
 --
--- a/ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
+++ b/ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
@@ -1,21 +1,20 @@
+{-# LANGUAGE DeriveAnyClass    #-}
 {-# LANGUAGE NamedFieldPuns    #-}
 {-# LANGUAGE OverloadedStrings #-}
 module Test.Consensus.Node (tests) where
 
 import           Data.Bifunctor (second)
 import qualified Data.Map.Strict as Map
-import qualified Data.Text as T
 import           System.Directory (getTemporaryDirectory)
-import           System.FilePath ((</>))
 import           System.IO.Temp (withTempDirectory)
 
-import           Control.Monad.Class.MonadThrow (try)
+import           Control.Monad.Class.MonadThrow (Exception, throwM, try)
 import           Control.Monad.IOSim (runSimOrThrow)
 
 import           Cardano.Crypto (ProtocolMagicId (..))
 
+import           Ouroboros.Consensus.Node.DbLock
 import           Ouroboros.Consensus.Node.DbMarker
-import           Ouroboros.Consensus.Util.ResourceRegistry (withRegistry)
 
 import           Ouroboros.Consensus.Storage.FS.API.Types
 import           Ouroboros.Consensus.Storage.FS.IO (ioHasFS)
@@ -38,7 +37,7 @@ tests = testGroup "Node"
       , testCase "corrupt"      test_checkProtocolMagicId_corrupt
       , testCase "empty"        test_checkProtocolMagicId_empty
       ]
-    , testCase "lockDbMarkerFile" test_lockDbMarkerFile
+    , testCase "lockDb"         test_lockDb
     ]
 
 {-------------------------------------------------------------------------------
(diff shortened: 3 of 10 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+import           Control.Exception
+import           System.FileLock
+
+-- | This can run from multiple processes concurrently, without any locking.
+-- It creates the db directory and its lock file, if any of them doesn't exist
+-- already.
+createDBFolder :: Monad m => HasFS m h -> m ()
+createDBFolder hasFS = do
+    createDirectoryIfMissing hasFS True root
+    void $ hOpen hasFS pFile (WriteMode AllowExisting)
```
mrBliss, line 25: Quoting what you quoted in https://github.com/input-output-hk/ouroboros-network/issues/1903#issuecomment-609634186:
> [..], on Windows 'System.IO.openFile' for a lock file will fail when the lock is held

So this is wrong as it tries to open the lock file even if it's locked.

Question: does `tryLockFile` work when the file doesn't exist yet? If so, then that would be easier. Ah, see https://github.com/takano-akio/filelock/blob/9681ff960d2695cd67fe749edab0e2af06c6f829/System/FileLock/Internal/LockFileEx.hsc#L40

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+  where
+    root  = mkFsPath []
+    pFile = fsPathFromList ["dblock"]
+
+-- | We try to lock the db multiple times, before we give up and throw an
```
mrBliss, line 30: As I said in https://github.com/input-output-hk/ouroboros-network/issues/1903#issuecomment-609640922, I wouldn't retry this until we know we need it.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+-- | This can run from multiple processes concurrently, without any locking.
+-- It creates the db directory and its lock file, if any of them doesn't exist
+-- already.
+createDBFolder :: Monad m => HasFS m h -> m ()
+createDBFolder hasFS = do
+    createDirectoryIfMissing hasFS True root
+    void $ hOpen hasFS pFile (WriteMode AllowExisting)
+  where
+    root  = mkFsPath []
+    pFile = fsPathFromList ["dblock"]
```
mrBliss, line 28: This will go in light of my comment above it, but otherwise should have used `dbLockFile` here.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+
+    pFile        = fsPathFromList [dbLockFile]
+    lockFilePath = fsToFilePath (MountPoint dbPath) pFile
+
+    justToRight  = maybe (Left ()) Right
+
+dbLockFile :: Text
+dbLockFile = "dblock"
+
+newtype DbLocked = DbLocked FilePath
```
mrBliss, line 62: Include this exception below `DbMarkerError` in the [`ErrorPolicy`](https://github.com/input-output-hk/ouroboros-network/blob/954dcc08be8838a2e3a8d1b4231e6320c7839931/ouroboros-consensus/src/Ouroboros/Consensus/Node/ErrorPolicy.hs#L60).

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+          lockAttempt $ n + 1
+        (Right _, _) -> return ()
+
+    pFile        = fsPathFromList [dbLockFile]
+    lockFilePath = fsToFilePath (MountPoint dbPath) pFile
+
+    justToRight  = maybe (Left ()) Right
+
+dbLockFile :: Text
+dbLockFile = "dblock"
```
mrBliss, line 60: Use `"lock"`.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+module Ouroboros.Consensus.Node.DbLock (
+    DbLocked (..)
+  , dbLockFile
+  , lockDB
+  ) where
+
+import           Control.Concurrent
+import           Control.Monad (void)
+import           Data.Text (Text)
+import           Ouroboros.Consensus.Storage.FS.API
```
mrBliss, line 12: Add a blank space above this line to separate imports from this package from the others.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+import           Ouroboros.Consensus.Storage.FS.API.Types
+import           Ouroboros.Consensus.Util.ResourceRegistry
+
+import           Control.Exception
+import           System.FileLock
```
mrBliss, line 17: Put these two together with the other `Control.*` and `Data.*` imports.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Ouroboros.Consensus.Node.DbLock (
+    DbLocked (..)
+  , dbLockFile
+  , lockDB
+  ) where
+
+import           Control.Concurrent
```
mrBliss, line 9: Why is this one needed?

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+import           Control.Monad (void)
+import           Data.Text (Text)
+import           Ouroboros.Consensus.Storage.FS.API
+import           Ouroboros.Consensus.Storage.FS.API.Types
+import           Ouroboros.Consensus.Util.ResourceRegistry
+
+import           Control.Exception
```
mrBliss, line 16: Use `Ouroboros.Consensus.Util.IOLike` and `throwM` instead of `throwIO`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs
```diff
@@ -52,9 +52,7 @@
 -- = Concurrency
 --
 -- The same database should not be opened multiple times concurrently.
---
--- TODO Should we ensure this with a lock file?
--- https://hackage.haskell.org/package/filelock-0.1.1.2/docs/System-FileLock.html
+-- This is ensured by the file lock of the chainDB.
```
mrBliss, line 55: :+1:

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
mrBliss: Let's wait until we're sure: https://github.com/input-output-hk/ouroboros-network/issues/1903#issuecomment-609660069

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,67 @@
...
+                    registry
+                    (const $ justToRight <$> tryLockFile lockFilePath Exclusive)
+                    (\f -> unlockFile f >> return True)
+      case (mlockFile, n) of
+        (Left (), 5) -> throwIO $ DbLocked lockFilePath
+        (Left (), _) -> do
+          -- TODO: We should log that we failed to take the lock, sleeping and
+          -- retrying.
+          threadDelay $ (n + 1) * 100000
+          lockAttempt $ n + 1
```
mrBliss, line 51: *If* we go for retries, let's use exponential backoff, not linear. Also we should document the behaviour: first wait x millisec, then ..., upto ..., etc. And we should indeed trace something, but this requires using a `Tracer`, and it's a bit annoying to have to add an extra tracer (`cardano-node` will have to be aware of it) just to trace one message.

dcoutts: It looks like the `base` package implementation of file locking on both posix, Linux and Windows will support the blocking lock acquite being interrupted by a timeout (async exception).

This would allow for a simple scheme:
```
res <- timeout 2000000 $ hLock lockfile ExclusiveLock
```
http://hackage.haskell.org/package/base-4.12.0.0/docs/GHC-IO-Handle-Lock.html

And if we're only blocking for up to one or two seconds, we don't need any new tracing I think.
https://github.com/input-output-hk/ouroboros-network/issues/1903#issuecomment-610038489

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
dcoutts: No, it's the content of the file that is locked, not the file as a whole. That's why the failure in the logs ocured on reading the file, not opening the file.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
dcoutts: If we use a timeout on a blocking lock acquire, we don't need retries.

dcoutts:
IMHO, there's nothing wrong with using the same file for the magic id and the file lock. But if you do, certainly you must open the lock file, lock it and hold it open for the duration. Only when it's opened and locked can you try to read the content.

But it's also ok to separate these into different files.

coot:
What I recently learned (while working on `ghc-tags-plugin`, which is also using `flock` locking mechanism available in `base`) that one also needs to `fsync` (or `fdatasync`) before closing the file handle; In a highly concurrent setting, I've seen another thread (system thread / process) not reading the written but not `fsync`-ed content - I am not sure if this related in any way to `ChainDB` though.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,47 @@
...
+import           Control.Monad.Class.MonadTimer
+import           Data.Text
+import qualified Data.Time.Clock as Time
+import           System.FileLock
+
+import           Ouroboros.Consensus.Storage.FS.API
+import           Ouroboros.Consensus.Storage.FS.API.Types
+import           Ouroboros.Consensus.Util.IOLike
+
+-- We use an empty file as a lock of the db. Some systems may delete the empty
```
mrBliss, line 19: Make it a Haddock docstring and explain why we do this, e.g., by replacing the first sentence with:
> We use an empty file as a lock of the db so that the database cannot be opened by more than one process.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,47 @@
...
+-- file when all its handles are closed. This is not an issue, since the file is
```
mrBliss, line 20: > Some systems may delete the empty file when all its handles are closed.

Does Windows do this?

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,47 @@
...
+-- created if it doesn't exist.
+withLockDB :: HasFS IO h ->  FilePath -> IO a -> IO a
+withLockDB hasFS dbPath action = do
+    createDirectoryIfMissing hasFS True root
+    -- We want to avoid blocking the main thread at an ffi, to avoid
+    -- unresponsiveness. We shouldn't be tempted to use `withAsync` or anything
+    -- else which links the blocked thread to our main thread, since this would
+    -- negate the whole point of using async.
```
mrBliss, line 28: I don't understand this. Do you mean unresponsiveness w.r.t. GC by the capability being blocked in an unsafe FFI call? Is `lockFile` an unsafe FFI call?

I also don't understand what you're saying about `withAsync`.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,47 @@
...
+    a     <- async (lockFile lockFilePath Exclusive)
+    mlock <- timeout (Time.secondsToDiffTime 2) $ wait a
+    case mlock of
+      Nothing   -> throwM $ DbLocked lockFilePath
+      Just lock -> finally action (unlockFile lock)
```
mrBliss, line 33: There's some `mask`ing missing. The simplest approach is to write a `acquireLock` and a `releaseLock` function and use those in `bracket`, then `bracket` will take care of the masking for us.

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
```diff
@@ -132,35 +131,38 @@ test_checkProtocolMagicId_empty = res @?= Left e
...
+    -- Raise an exception. The lock should get cleaned.
+    _ <- tryT (withLock $ throwM TestException)
+    tryL touchLock >>=
         (@?= (Right ()))
-    -- And finally unlock it.
+
+    -- Try to lock it twice. This should fail after 2sec.
+    -- This test succeeds, but may leave a forgotten lock to the file
+    -- tryL (withLock touchLock) >>=
+    --     (@?= (Left (DbLocked (dbPath </> T.unpack dbLockFile))))
```
mrBliss, line 155: This test is commented out

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl.hs
mrBliss, line 55:
```suggestion
-- This is ensured by the file lock of the ChainDB.
```

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,47 @@
...
+-- created if it doesn't exist.
+withLockDB :: HasFS IO h ->  FilePath -> IO a -> IO a
+withLockDB hasFS dbPath action = do
+    createDirectoryIfMissing hasFS True root
+    -- We want to avoid blocking the main thread at an ffi, to avoid
+    -- unresponsiveness. We shouldn't be tempted to use `withAsync` or anything
+    -- else which links the blocked thread to our main thread, since this would
+    -- negate the whole point of using async.
+    a     <- async (lockFile lockFilePath Exclusive)
+    mlock <- timeout (Time.secondsToDiffTime 2) $ wait a
```
mrBliss, line 30: Let's also document in the docstring that we wait 2 seconds.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
kderme: No, I haven't seen it on windows or linux, but it's possible to happen.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
kderme: unresponsiveness to ^C and timeouts. lockFile wraps an ffi call which is not interruptible and we don't want to have the main thread trapped in such a call. That's why I used async and let another thread do the actual call.

`withAsync` is usually mentioned as a better alternative to simple async (and this is what I firstly used) but it sends an async exception to the async thread and wait for it to terminate. This effectively blocks the main thread again and is what we tried to avoid in the first place.

mrBliss: Ah, I see what you mean :+1:. Incorporate your response into the comment.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
kderme: @mrBliss what do you think should be masked?

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
kderme: Yes, as I explain above the test `This test succeeds, but may leave a forgotten lock to the file` and this lock is only cleaned when all tests finish and the process dies. So probably we shouldn't include it?

kderme:
I tested flock from base, but I couldn't make it work properly. It just wouldn't lock the files effectively and I din't manage to get down to why exactly this happened. So I used the `filelock` package. Also I preferred to use a different file from the DBMarket, since `filelock` suggests against accessing the locked file for other purposes.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
mrBliss: What if an async exception is thrown after obtaining the lock and before `finally`? That's the thing with async exceptions, they can happen at any point.

kderme: I'm not sure if we should mask while waiting on a timeout. Let me check it..

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
mrBliss: > Yes, as I explain above the test `This test succeeds, but may leave a forgotten lock to the file` and this lock is only cleaned when all tests finish and the process dies. So probably we shouldn't include it?

But now we are not even testing whether the locking actually works. 

> this lock is only cleaned when all tests finish and the process dies

That's fine, or what's wrong with that?

mrBliss:
> Also I preferred to use a different file from the DBMarker, since `filelock` suggests against accessing the locked file for other purposes.

Yeah, I agree, that's a good reason.

ouroboros-consensus/src/Ouroboros/Consensus/Node/DbLock.hs
```diff
@@ -0,0 +1,58 @@
...
+    -- call. This is not a big issue though, since if we fail to take the lock
+    -- succesfully, the whole process will soon die.
+    a <- async (lockFile lockFilePath Exclusive)
+    bracket (timeout (Time.secondsToDiffTime 2) $ wait a)
+            cleanup
+            (\_ -> action)
+  where
+    cleanup = \case
+      Nothing   -> throwM $ DbLocked lockFilePath
+      Just lock -> unlockFile lock
```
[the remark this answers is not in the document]
kderme: @mrBliss I'm not sure I understand. This pattern matches if we got a timeout or async returned succesfully. So it can't be performed by the async thread.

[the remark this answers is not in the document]
mrBliss: When waiting on the `async` that calls `lockFile` times out, `Nothing` is returned. This is `Nothing` is then passed to `(\_ -> action)` which ignores it. So even when we don't have the lock do we execute the action. After the `action` is done or when `action` throws an exception, we go to `cleanup`. `cleanup` then throws a `DbLocked` exception in case it got `Nothing`, which is too late, as we already executed the `action`.

What we should do is (untype-checked):
```haskell
    createDirectoryIfMissing hasFS True root
    bracket acquireLock unLockFile (const action)
  where
    -- We want to avoid blocking ...
    acquireLock :: IO FileLock
    acquireLock = do
      lockFileAsync <- async (lockFile lockFilePath Exclusive)
      mbLock <- timeout (Time.secondsToDiffTime 2) $ wait lockFileAsync
      case mbLock of
        -- We timed out while waiting on the lock, the db is still locked
        Nothing   -> throwM $ DbLocked lockFilePath
        Just lock -> return lock
```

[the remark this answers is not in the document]
kderme: oh right, fixed

mrBliss:
(I have temporarily added the commits from #1851 to this branch to do a final check on Windows. I won't merge these commits, because building/testing everything on Windows is too slow because of the lack of caching.)

mrBliss:
bors merge

mrBliss:
bors merge

mrBliss:
bors merge

ouroboros-consensus/src/Ouroboros/Consensus/Util/FileLocking.hs
```diff
@@ -0,0 +1,38 @@
...
+
+-- | Abstraction for file locks
+data FileLocking m = FileLocking {
+      -- | Obtain an exclusive lock on the given file.
+      --
+      -- Blocks until the lock is available.
+      --
+      -- We don't guarantee the ability to read/write to a locked file, not
+      -- even when holding the lock.
+      lockFile   :: FilePath -> m (FileLock m)
```
mrBliss, line 18: Turn it into `FileLock` for a single file lock, no more type family.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/FileLocking.hs
```diff
@@ -0,0 +1,88 @@
...
+          error $ "unlocking an unlocked file: " <> show path
+        Just LazyRelease ->
+          error $ "unlocking a file that is still being unlocked: " <> show path
+        Just Held -> case releaseDelay of
+          Nothing    -> do
+            writeTVar varLocks $ Map.delete path locks
+            return $ return ()
+          Just delay -> do
+            writeTVar varLocks $ Map.insert path LazyRelease locks
+            return $ void $ async $ do
```
mrBliss, line 86: `forkThread`

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
```diff
@@ -1,33 +1,40 @@
-{-# LANGUAGE DeriveAnyClass    #-}
-{-# LANGUAGE NamedFieldPuns    #-}
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE CPP                 #-}
```
mrBliss, line 1: No longer necessary

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
```diff
@@ -136,38 +147,147 @@ test_checkProtocolMagicId_empty = res @?= Left e
...
+        dbLockFsPath
+        (dbLockTimeoutFn timeout)
+        (return ())
 
-    -- Raise an exception. The lock should get released.
-    _ <- tryT (withLock $ throwM TestException)
-    -- Test that the lock was released by acquiring it again
-    tryL touchLock >>=
-        (@?= (Right ()))
+prop_acquire_held_lock :: ReleaseDelay -> Property
```
mrBliss, line 176: Use the real implementation for this test.

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
```diff
@@ -136,38 +147,147 @@ test_checkProtocolMagicId_empty = res @?= Left e
   lockDb
 -------------------------------------------------------------------------------}
 
-test_lockDb :: Assertion
```
mrBliss: State that we use the mock lock for testing that we can release and reacquire the lock, and the real lock for testing that we can't acquire the same lock twice at the same time.

ouroboros-consensus/test-consensus/Test/Consensus/Node.hs
```diff
@@ -136,38 +147,147 @@ test_checkProtocolMagicId_empty = res @?= Left e
...
+      let -- Hold the lock for 'actualDelay' and then signal we have released it
+          firstThread =
+            -- The lock will not be held, so just use the default parameters
+            -- to acquire it
+            withLock hasFS fileLocking dbLockTimeout $
+              -- Hold the lock for ACTUAL
+              threadDelay actualDelay
 
+          -- Try to obtain the held lock
+          secondThread =
```
mrBliss, line 227: Do this in the main thread instead. Link the first thread to the main thread.

mrBliss:
bors merge

PR #1908, opened 2020-04-06
Revert "ImmutableDB: truncate blocks from the future"

mrBliss:
This reverts commit 81f2f1e17f4bf331525d4ae915d9c0ae68d4b8b4.

See #1773, we first have to open the ImmutableDB /with/ all blocks because we
don't know yet in which slot we are. We need the ImmutableDB to obtain a
ledger state which will tell us which slot we're in. Afterwards, we can
truncate the future blocks from the ImmutableDB. This last step will be done
by the ChainDB and will follow in a future PR.


mrBliss:
bors merge

PR #1916, opened 2020-04-06
Label all threads spawned by consensus

mrBliss:
To make profiling with `ghc-events-analyze` more informative, threads
should be labelled.

The consensus layer consistently uses the `ResourceRegistry` to spawn
background threads, so we let `ResourceRegistry.fork(Linked)Thread` take the
thread label as argument. We are now pretty sure all threads spawned by
consensus are labelled.


mrBliss:
bors merge

PR #1919, opened 2020-04-07
VolatileDB: only obtain the write lock if we can GC a file

mrBliss:
Previously, `garbageCollect` was obtaining the exclusive write lock every time
it was called, starving any concurrent readers and appender, even if it
couldn't garbage collect any file.

Now, we read the state first to determine whether we can garbage collect at
least one file, if so, we obtain the exclusive write lock and do the actual
garbage collection.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
@@ -104,7 +104,7 @@ import           Control.Monad
 import           Control.Monad.State.Strict
 import           Control.Tracer (Tracer, traceWith)
 import qualified Data.ByteString.Builder as BS
-import           Data.List (foldl', sortOn)
+import           Data.List (foldl')
 import qualified Data.Map.Strict as Map
 import           Data.Maybe
 import           Data.Set (Set)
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
```diff
@@ -285,37 +285,59 @@ putBlockImpl env@VolatileDBEnv{ maxBlocksPerFile, tracer }
...
-      -- everything), because a GC can only delete blocks < a slot.
-      modify $ \st -> st {
-          currentMaxSlotNo = FileInfo.maxSlotInFiles
-            (Index.elems (currentMap st))
-        }
+garbageCollectImpl env slot = do
+    -- Check if we can actually GC something using a cheaper read (allowing
+    -- for more concurrency) before obtaining the more expensive exclusive
+    -- write lock.
+    usefulGC <- atomically $ getterSTM gcPossible env
```
mrBliss, line 309: Say that this is monotonic: if this is True, then `filesToGC` must be non-empty. That's because we're the only one GC'ing and because we never GC the current file.

No matter what happens in the middle, we only obtain the write lock when there is something to do.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
```diff
@@ -285,37 +285,59 @@ putBlockImpl env@VolatileDBEnv{ maxBlocksPerFile, tracer }
...
+            currentMaxSlotNo = FileInfo.maxSlotInFiles
+              (Index.elems (currentMap st))
+          }
+  where
+    -- | Return 'True' if a garbage collection would actually garbage collect
+    -- at least one file.
+    gcPossible :: OpenState blockId h -> Bool
+    gcPossible = not . null . getFilesToGC
+
+    -- | Return the list of files that can be garbage-collected.
```
mrBliss, line 329:
```suggestion
    -- | Return the list of files that can be garbage collected.
```

mrBliss:
bors merge

PR #1932, opened 2020-04-10
ChainDB: batch garbage collections

mrBliss:
Previously, we scheduled a garbage collection for each block for 10 seconds in
the future. This meant that our scheduled GCs queue was blocks/s * 10 long.
When tracing the queue length on my machine, it hovered between 5000 and 6000
entries. Moreover, a VolatileDB garbage collection is triggered at blocks/s,
which should result in a lot of contention for the VolatileDB state.

Even worse is that a 10 second delay is too short to reliably ensure the block
will have been flushed to disk (in the ImmutableDB) before it is garbage
collected. However, increasing this delay would make the queue significantly
longer.

To fix these issues, we introduce a GC interval (in seconds). We batch all GCs
in the same interval together. This means that the queue length is now at most
⌈delay / interval⌉ + 1, e.g., 60s / 10s = 7, which is much shorter than
5000-6000. Moreover, there will be at most one GC every `interval` seconds,
e.g., 10s.

The cost of switching to a longer GC delay is that the in-memory index of the
VolatileDB will be larger, making operations on it, such as lookups, more
expensive (most operations are `O(n*log(n))`). See the docstring of
'defaultSpecificArgs' for what the new default values of `gcDelay` and
`gcInterval` mean in practice.


mrBliss:
bors merge

PR #1938, opened 2020-04-10
ChainDB: update a docstring and simplify a function

mrBliss:
* ChainDB.Iterator: update costs in the docstring 

  Opening an iterator no longer requires reading the blocks corresponding to the bounds from disk, their hashes can be looked up in the cached primary and secondary indices of the ImmutableDB. Update the out-of-date docstring.

* ImmutableDB: remove unused `iteratorPeek`

* ChainDB.ImmDB: simpler implementation of exclusive end bound

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ImmDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ImmDB.hs
@@ -33,7 +33,6 @@ module Ouroboros.Consensus.Storage.ChainDB.Impl.ImmDB (
   , closeDB
   , iteratorNext
   , iteratorHasNext
-  , iteratorPeek
   , iteratorClose
     -- * Tracing
   , TraceEvent
@@ -640,12 +639,6 @@ iteratorHasNext :: (HasCallStack, MonadCatch m)
                 -> ImmDB.Iterator (HeaderHash blk) m a
                 -> m (Maybe (Either EpochNo SlotNo, HeaderHash blk))
 iteratorHasNext db it = withDB db $ const $ ImmDB.iteratorHasNext it
-
-iteratorPeek :: (HasCallStack, MonadCatch m)
-             => ImmDB m blk
-             -> ImmDB.Iterator (HeaderHash blk) m a
-             -> m (ImmDB.IteratorResult a)
-iteratorPeek db it = withDB db $ const $ ImmDB.iteratorPeek it
 
 iteratorClose :: (HasCallStack, MonadCatch m)
               => ImmDB m blk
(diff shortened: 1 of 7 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ImmDB.hs
```diff
@@ -496,31 +495,34 @@ stream db registry blockComponent from to = runExceptT $ do
           ImmDB.WrongHashError slot _ _ -> slot
 
     -- | The ImmutableDB doesn't support an exclusive end bound, so we stop
-    -- the iterator when it reaches its exclusive end bound.
-    stopAt :: forall a.
-              StreamTo blk
-           -> ImmDB.Iterator (HeaderHash blk) m (HeaderHash blk, a)
-           -> ImmDB.Iterator (HeaderHash blk) m (HeaderHash blk, a)
+    -- the iterator when the next block is the exclusive end bound.
+    stopAt :: StreamTo blk
```
mrBliss, line 499: Rename to `makeExclusive` because it is already stopping at the inclusive upper bound.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ImmDB.hs
```diff
@@ -496,31 +495,34 @@ stream db registry blockComponent from to = runExceptT $ do
...
+      StreamToExclusive endPt -> \it -> do
+          -- Whenever the iterator is moved to the next block, we check
+          -- whether the next block (cheap with 'ImmDB.iteratorHasNext') is
+          -- the end bound, in which case we close the iterator. All other
+          -- operations on the iterator can remain the same.
+          --
+          -- We must do the check immediately, because the very first block
+          -- might be the exclusive bound.
+          closeWhenEndIsNext it
+          return it
```
mrBliss, line 513: Problem: when moving to the block right before the exclusive bound, we close the iterator, which means that `iteratorHasPeek` will return `IteratorExhausted` while it should return the current block.

Plan: remove `iteratorPeek` because it is no longer used!

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ImmDB.hs
mrBliss: Actually, this function is called in both cases: when the upper bound is exclusive *or* inclusive. It will just do nothing in the second case. So I think the name is fine after all.

mrBliss:
bors merge

mrBliss:
bors merge

PR #1946, opened 2020-04-13
Introduce Query abstraction

edsko:
This cleans up the hard fork code a bit, and makes it possible to provide a clean interface for caching, which we can take advantage of in the `BlockchainTime` API.


edsko:
Thanks @erikd .

bors merge

Issue #1949, opened 2020-04-14
Add timeout for sending Mux segments.


karknu:
We need to implement a timeout for sending Mux segments. Without such a timeout a miniprotocol can block on sending a message which will prevent it from transitioning to a receive state with a timeout.

karknu:
Fixed.

PR #1953, opened 2020-04-15
Send timeout

karknu:


karknu:
Implements #1949

network-mux/src/Network/Mux/Egress.hs
```diff
@@ -138,27 +145,30 @@ newtype Wanton m = Wanton { want :: StrictTVar m BL.ByteString }
 -- shared FIFO that contains the items of work. This is processed so
 -- that each active demand gets a `maxSDU`s work of data processed
 -- each time it gets to the front of the queue
-mux :: MonadSTM m
+mux :: (MonadAsync m, MonadFork m, MonadMask m, MonadThrow (STM m),
+        MonadTimer m, MonadTime m)
```
coot, line 149: Can we put one per line? This will make nicer diffs in the future.

ouroboros-network-framework/src/Ouroboros/Network/Socket.hs
```diff
@@ -241,13 +247,17 @@ connectToNode' sn versionDataCodec NetworkConnectTracers {nctMuxTracer, nctHands
...
+            throwIO ExceededTimeLimit
+         Just (Left (errDrv :: ProtocolLimitFailure))-> do
             traceWith muxTracer $ Mx.MuxTraceHandshakeClientError errDrv (diffTime ts_end ts_start)
             throwIO errDrv
-         Right (Left err) -> do
+         Just (Right (Left err)) -> do
             traceWith muxTracer $ Mx.MuxTraceHandshakeClientError err (diffTime ts_end ts_start)
             throwIO err
-         Right (Right app) -> do
+         Just (Right (Right app)) -> do
```
coot, line 260: Can we simplify the `Maybe (Either _ (Either _ r))` type?

coot: what about this idea:

```
data HandshakeException a = 
    HandshakeProtocolLimit ProtocolLimitfailure
  | HandshakeProtocolError a
  | HanshakeTimeout
 
tryHandshake ::  m (Either a r) -> m (Either (HanshakeException a) r)
tryHandshaket = ... withTimeoutSerial ... try ... 
```

The `a` in `tryHandshake` would be different for `client` and `server`.

karknu: A great idea!

coot: Really? Even though `HandshakeException a` is polymorphic?

karknu: I tried and failed to get around the need for a `Maybe (Either _ (Either _ r))` case statement in `tryHandshake` which made things far worse.
The updated code works and is more like your initial suggestion.

karknu:
bors  r+

PR #1956, opened 2020-04-15
ChainDB: store a ledger snapshot on startup after a long replay

mrBliss:
The LgrDB decides to make snapshots based on the time passed and the number of
blocks advanced since the last snapshot. When it had to replay blocks on
startup because the on-disk snapshots were out of date or missing, that number
of replayed blocks is taken into account for the decision to make the next
snapshot (as the starting point for the number of blocks advanced).

However, the decision making logic is only triggered *after* copying a block
to the ImmutableDB. So, for example, when starting a node with lots of blocks,
but without on-disk ledger snapshots, it will replay all blocks to reconstruct
an up-to-date ledger state, but it won't write it to disk until a block has
been copied to the ImmutableDB. When experimenting, one often has a node that
no longer downloads blocks acting as a server node, so that node will never
write its snapshot to disk. This means a full ledger replay has to be done on
every startup.

Fix this by triggering the snapshot making logic before starting the loop that
waits for blocks being copied to the ImmutableDB.


mrBliss:
bors merge

Issue #1950, opened 2020-04-14
Enable local query node-to-client mini-protocol


coot:

mrBliss:
Related, but probably out of date: https://github.com/input-output-hk/ouroboros-network/pull/1522

coot:
@mrBliss thanks for mentioning this; it needs to be updated, since we change the API in the mean time.

PR #1928, opened 2020-04-09
Fix thunks in Shelley

mrBliss:
Fixes #1558.

PR #1962, opened 2020-04-16
node to client api changes

coot:
Small changes to NodeToClient api and LocalStateQuery protocol type signatures.

Related to #1950

- Added null peers in NodeToNode module
- Add kinds to type signatures of LocalQueryProtocol

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
@@ -1,6 +1,8 @@
+{-# LANGUAGE GADTs               #-}
 {-# LANGUAGE DataKinds           #-}
 {-# LANGUAGE KindSignatures      #-}
 {-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE PolyKinds           #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies        #-}
@@ -40,10 +42,10 @@ module Ouroboros.Network.NodeToClient (
   , ncSubscriptionWorker_V1
   , ncSubscriptionWorker_V2
 
-  -- * Re-exported clients
-  , chainSyncClientNull
-  , localTxSubmissionClientNull
-  , localStateQueryClientNull
+  -- * Null Protocol Peers
+  , chainSyncPeerNull
+  , localStateQueryPeerNull
+  , localTxSubmissionPeerNull
 
   -- * Re-exported network interface
   , IOManager (..)
@@ -51,7 +53,7 @@ module Ouroboros.Network.NodeToClient (
   , withIOManager
   , LocalSnocket
   , localSnocket
-  , LocalAddress
+  , LocalAddress (..)
 
     -- * Versions
   , Versions (..)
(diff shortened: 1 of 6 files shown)


ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
```diff
@@ -673,3 +683,35 @@ networkErrorPolicies = ErrorPolicies
...
+    :: forall (tx :: Type) (reject :: Type) m a. MonadTimer m
+    => Peer (LocalTxSubmission.LocalTxSubmission tx reject)
+            AsClient LocalTxSubmission.StIdle m a
+localTxSubmissionPeerNull =
+    LocalTxSubmission.localTxSubmissionClientPeer
+      (LocalTxSubmission.LocalTxSubmissionClient untilTheCowsComeHome)
+
+-- ;)
+untilTheCowsComeHome :: MonadTimer m => m a
+untilTheCowsComeHome = forever $ threadDelay 43200 {- day in seconds -}
```
MarcFontaine, line 717: 43200 is only 12 hours in seconds.
It may make sense to add a new function `threadDelayForever` to `MonadDelay`
and maybe use a different Implementation then `forever $ threadDelay 43200` (maybe `atomically retry`). 
For debugging etc it may be useful to detect that something is blocked forever (`threadDelayForever`).

coot: Isn't waking up once every 12 hours enough? The cost is negligible considering `GHz` CPU frequency.

> It may make sense to add a new function threadDelayForever to MonadDelay

I think we want to keep `io-sim-classes` as small as possible, preferably only exposing api's that are in `base` / `async` / `stm`.

> For debugging etc it may be useful to detect that something is blocked forever (threadDelayForever).

This would be a programmer error. For this reason I prefer to use `Void`, e.g. `loop  = forever someIO :: IO Void` to make it clear at the type level that `loop` never returns.

coot:
@MarcFontaine any ideas? Note that we need to trick `GHC` no to throw `BlockIndefinitely...` errors.

coot:
bors merge

MarcFontaine:
Right. I did not think about `BlockIndefinitely`.
One could use a global MVar to block everything.
In any case if we add a abstraction for `waitForever` we can change the implementation in a single point.

coot:
> One could use a global MVar to block everything.

But that's ugly :D; we even avoided such global state in `Win32-network` in favor of `withIOManager :: (IOManager -> IO a) -> IO a`.

> In any case if we add a abstraction for waitForever we can change the implementation in a single point.

`untilTheCowsComeHome` serves this purpose :), we only have one more such call in `ntp-client` and that's about it.  AFAIK, other repos are also only using this trick because they don't use the `null` peers yet.

mrBliss:
bors cancel

mrBliss:
> This PR was included in a batch that was canceled, it will be automatically retried

Seriously? This PR was closed and the job cancelled, and still it retries? Useless :weary:

coot:
@MarcFontaine why did you closed it?

coot:
bors merge

MarcFontaine:
Sorry. I just click on the wrong button I guess.
I did not notice that I closed it

mrBliss:
I have found out why the tests are taking so long: https://github.com/input-output-hk/ouroboros-network/pull/1928/commits/0ae688de9e44cd6d644f788f78aa37168132a410. I'm going to cancel this bors job because it will time out anyway so that #1928, which will fix the problem, can be merged. After that, we should be able to merge this PR without problems. Apologies in advance.

mrBliss:
bors cancel

coot:
thanks @mrBliss for looking into this.

mrBliss:
@coot Make sure to rebase on master before trying to bors it.

coot:
I did :):
```
git log
...
807f86ace Merge #1928                               (origin/master, origin/bors/staging, origin/HEAD)
```

coot:
borse merge

coot:
bors merge

PR #1963, opened 2020-04-16
Shelley: do the chainChecks in validateEnvelope

mrBliss:
In the Shelley ledger, the `CHAIN` rule is used to apply a whole block. In consensus, we split up the application of a block to the ledger into separate steps that are performed together by `applyExtLedgerState`:

* `applyChainTick`: executes the `TICK` transition
* `validateHeader`:
  - `validateEnvelope`: default checks
  - `updateConsensusState`: executes the `PRTCL` transition
* `applyLedgerBlock`: executes the `BBODY` transition and the remaining checks that are part of the `CHAIN` transition

The unfortunate thing was that we had to copy-paste some checks part of the `CHAIN` transition that were not part of any subtransition.

The Shelley has recently factored out these checks into the `chainChecks` function. This commit takes advantage of this change. The new situation:

* `applyChainTick`: executes the `TICK` transition
* `validateHeader`:
  - `validateEnvelope`: default checks + executes the `chainChecks`
  - `updateConsensusState`: executes the `PRTCL` transition
* `applyLedgerBlock`: executes the `BBODY` transition

Now all checks and transitions of the `CHAIN` rule nicely map onto separate steps in consensus.

Note that `chainChecks` can be done on the header, so this change means that we do these checks earlier, in the ChainSyncClient, instead of after downloading the block and applying it to the ledger with `applyExtLedgerState`.

---

To make this possible, some refactorings were needed:

* HeaderValidation: export defaultValidateEnvelope

* HeaderValidation: replace OtherEnvelopeError with a type family

  Instead of `Text`, we now allow a block-specific `OtherHeaderEnvelopeError`. We'll use this ability to do more header checks for Shelley without having to convert the errors to `Text`.

* HeaderValidation: pass more info to validateEnvelope

  In particular, replace `BlockConfig` with `TopLevelConfig`, and add `LedgerView`. We need this extra data to do some more checks for Shelley.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
@@ -50,14 +50,14 @@ import           Data.Type.Equality ((:~:) (Refl))
 import           GHC.Generics (Generic)
 
 import           Cardano.Binary (enforceSize, fromCBOR, toCBOR)
-import           Cardano.Prelude (Natural, NoUnexpectedThunks (..),
-                     OnlyCheckIsWHNF (..))
+import           Cardano.Prelude (NoUnexpectedThunks (..), OnlyCheckIsWHNF (..))
 import           Cardano.Slotting.Slot (EpochNo, WithOrigin (..))
 
 import           Ouroboros.Network.Block
 import           Ouroboros.Network.Protocol.LocalStateQuery.Codec (Some (..))
 
 import           Ouroboros.Consensus.BlockchainTime (singletonSlotLengths)
+import           Ouroboros.Consensus.Config
 import           Ouroboros.Consensus.Forecast
 import qualified Ouroboros.Consensus.HardFork.History as HardFork
 import           Ouroboros.Consensus.HeaderValidation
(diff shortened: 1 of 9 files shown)


mrBliss:
This includes #1928 until it is merged.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -320,6 +302,32 @@ instance Crypto c => ShowQuery (Query (ShelleyBlock c)) where
...
+      ShelleyLedgerConfig globals = configLedger cfg
+      pparams = SL.lvProtParams ledgerView
+
+newtype ShelleyOtherHeaderEnvelopeError c = ShelleyOtherHeaderEnvelopeError {
+      unShelleyOtherHeaderEnvelopeError :: STS.PredicateFailure (STS.CHAIN c)
+    }
+  deriving (Eq, Show)
+  -- TODO fix thunks
+  deriving NoUnexpectedThunks
+    via OnlyCheckIsWHNF "ShelleyOtherHeaderEnvelopeError" (ShelleyOtherHeaderEnvelopeError c)
```
mrBliss, line 329: Let's make this an orphan instance of `NoUnexpectedThunks` instead, so that I'm forced to remove it when I add the right one in `cardano-ledger-specs`.

mrBliss:
bors merge

PR #1972, opened 2020-04-20
Update dependencies

erikd:
Also remove some instance declarations that are now exported from
'cardano-ledger-specs'.


mrBliss:
bors merge

PR #1973, opened 2020-04-20
Introduce ValidatedFragment

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/Fragment/Validated.hs
```diff
@@ -0,0 +1,65 @@
...
+
+    -- | Ledger after after validation
+  , validatedLedger   :: !l
+  }
+
+invariant :: forall l blk. ApplyBlock l blk
+          => ValidatedFragment l blk -> Either String ()
+invariant ValidatedFragment{..}
+    | ledgerTip /= headPoint
+    = Left $ show ledgerTip ++ " /= " ++ show headPoint
```
mrBliss, line 44: Would it be useful to include which is which in the error message?

edsko: Will do.

PR #1976, opened 2020-04-20
Bump dependency on cardano-ledger-specs

mrBliss:
This brings in https://github.com/input-output-hk/cardano-ledger-specs/pull/1378.


mrBliss:
bors merge

PR #1978, opened 2020-04-20
ChainDB.Reader: avoid disk reads when moving forward to the same pt

### LINKED ISSUE #1977: Chain DB: avoid disk reads when moving a reader forward to the same point
Daedalus polls `GET /v2/network/information` every 2 seconds, causing `cardano-wallet` to send a `MsgFindIntersect [Origin]` request (see https://github.com/input-output-hk/cardano-wallet/blob/745aaad67004855a84c51e400c6fa1d10aedb910/lib/byron/src/Cardano/Wallet/Byron/Network.hs#L245) to the `ChainSyncServer` run by `cardano-node`. The `ChainSyncServer` uses a `ChainDB.Reader`, on which it calls `readerForward` with `[Origin]`.

Even though the reader is already at origin (a reader always starts at origin), this call to `readForward` with `[Origin]` causes it to open an ImmutableDB iterator starting from origin, costing us one (or two?) cached disk reads. These disk reads showed up here (see the small peaks):

image

We can avoid these disk reads by not opening the ImmutableDB iterator when the reader is already at the given point.

mrBliss:
Fixes #1977.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Reader.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Reader.hs
@@ -254,26 +254,22 @@ instructionHelper registry varReader blockComponent encodeHeader fromMaybeSTM CD
       -- convert to the right block component.
       Right fupdate -> headerUpdateToBlockComponentUpdate fupdate
       -- We were in the 'ReaderInImmDB' state or we need to switch to it.
-      Left (rollState, mbImmIt) -> case rollState of
-        RollForwardFrom pt -> do
-          immIt <- case mbImmIt of
-            Just immIt -> return immIt
-            -- We were in the 'ReaderInMem' state but have to switch to the
-            -- 'ReaderInImmDB' state.
-            Nothing    -> do
-              trace $ ReaderNoLongerInMem rollState
-              ImmDB.streamAfterKnownBlock cdbImmDB registry
-                ((,) <$> getPoint <*> blockComponent) pt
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Reader.hs
```diff
@@ -254,26 +254,22 @@ instructionHelper registry varReader blockComponent encodeHeader fromMaybeSTM CD
...
-        RollBackTo      pt -> do
-          case mbImmIt of
-            Just immIt -> ImmDB.iteratorClose cdbImmDB immIt
-            Nothing    -> trace $ ReaderNoLongerInMem rollState
-          immIt' <- ImmDB.streamAfterKnownBlock cdbImmDB registry
-            ((,) <$> getPoint <*> blockComponent) pt
-          let readerState' = ReaderInImmDB (RollForwardFrom pt) immIt'
-          atomically $ writeTVar varReader readerState'
-          return $ pure $ RollBack pt
+      Left (rollState, mbImmIt) -> do
```
mrBliss, line 257: State the invariant for https://github.com/input-output-hk/ouroboros-network/blob/a5ca95890f3d4a0c37ea017b5a96f309e6f3b4da/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs#L333

> for all `(ReaderInImmDB rollState immIt)`: the predecessor of the next block streamed by `immIt` must be the block identified by `readerRollStatePoint rollState`.

mrBliss:
bors merge

PR #1981, opened 2020-04-21
Version codec

### LINKED ISSUE #1971: Deserialise failure when connecting to FC4 node
I downloaded `FC4` release of `Deadalus` and run:
```
cabal run exe:cardano-cli -- get-tip --config ~/.local/share/Daedalus/mainnet_flight/config.yaml --socket-path ~/.local/share/Daedalus/mainnet_flight/cardano-node.socket
```
from https://github.com/input-output-hk/cardano-node/commit/0172a96e16f46f1688da6dd8294481bb995fd72d

The client application returned:
```
MuxError {errorType = MuxBearerClosed, errorMsg = "<socket: 11> closed when reading data, waiting on next header True", errorStack = []}
```

The logs of the node reveal more detail:
```
[fozzie:cardano.node.LocalErrorPolicy:Error:44] [2020-04-18 07:35:57.80 UTC] {"event":"ErrorPolicyUnhandledApplicationException (DeserialiseFailure 12 \"decode NodeToClientVersion: unknown tag\")","kind":"ErrorPolicyTrace","address":"LocalAddress {getFilePath = \"\"}"}
```

It's likely that the introduction of `LocalStateQuery` in #1507 wasn't actually backward compatible.

coot:
This PR changes how Handshake codec works.  Previously it would fail if 
an unknown version number was submitted, now it will not.

This change should not change wire format of the handshake messages. 

In addition it allows to confiugure which protocols versions are used by
consensus, there will be an accompanying PR in `cardano-node` repository which
adds an option to the confituration file.

I hope this fixes #1971, though I havent tested it (for that we need a new protocol
version).

- Make it possible to configure NodeToNodeVersion / NodeToClientVersion
- Make test-cddl run with cabal by default
- Updated 'test-cddl' component
- Fix a typo, added TODO comment
- Updated chain-sync demo
- Updatd test-network
- Updated ouroboros-protocol-tests library
- Updated ouroboros-network library
- CodecCBORTerm deserves its own module
- Move Unversioned to its own module
- Updated ping-pong demo
- Updated ouroboros-network-framework-tests
- Added unversionedHandshakeCodec
- RefuseReason changes
- Change Handshake protocol codec
- Change Socket API

### DIFF
--- a/ouroboros-network/ouroboros-network.cabal
+++ b/ouroboros-network/ouroboros-network.cabal
@@ -287,6 +287,8 @@ test-suite test-cddl
                        cborg,
                        containers,
                        contra-tracer,
+                       directory,
+                       filepath,
                        fingertree,
                        hashable,
                        io-sim,
--- a/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/Handshake/Test.hs
+++ b/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/Handshake/Test.hs
@@ -550,20 +550,36 @@ prop_pipe_asymmetric_IO (ArbitraryVersions clientVersions _serverVersions) =
 --
 
 instance Eq (AnyMessage (Handshake VersionNumber CBOR.Term)) where
-  AnyMessage (MsgProposeVersions vs)          == AnyMessage (MsgProposeVersions vs')  = vs == vs'
+  AnyMessage (MsgProposeVersions vs) == AnyMessage (MsgProposeVersions vs')
+    = vs == vs'
+
   AnyMessage (MsgAcceptVersion vNumber vParams) == AnyMessage (MsgAcceptVersion vNumber' vParams')
-                                                                                      = vNumber == vNumber' && vParams == vParams'
-  AnyMessage (MsgRefuse vReason)              == AnyMessage (MsgRefuse vReason')      = vReason == vReason'
-  _                                           == _                                    = False
+    = vNumber == vNumber' && vParams == vParams'
+
+  AnyMessage (MsgRefuse vReason) == AnyMessage (MsgRefuse vReason')
+    = vReason == vReason'
+
+  _ == _ = False
 
 instance Show (AnyMessageAndAgency (Handshake VersionNumber CBOR.Term)) where
   show (AnyMessageAndAgency _ msg) = show msg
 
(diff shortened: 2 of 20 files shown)


ouroboros-network/ouroboros-network.cabal
```diff
@@ -306,6 +312,8 @@ test-suite test-cddl
   ghc-options:         -Wall
                        -Wno-unticked-promoted-constructors
                        -fno-ignore-asserts
+  if flag(nix)
+    cpp-options:       -DNIX
```
karknu, line 316: Isn't this define a bit too generic? I think that this define is passed on when compiling all dependencies too.

ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/Handshake/Test.hs
```diff
@@ -447,6 +452,99 @@ prop_pipe_IO :: ArbitraryVersions -> Property
...
 
+--
+-- Asymmetric tests
+--
+
+
+-- | Run a simple handshake client and server using connected channels.
+-- The server can only decode a subset of versions send by client.
+-- This test is using a fixed serverver 'Versions' which can only accept
+-- a single version 'Version_2' (it cannot decode any other version).
```
karknu, line 463: Looking at `versionNumberCodec'` it looks like the single version should be `Version_1`.

ouroboros-network/ouroboros-network.cabal
coot: Then it's better to change it. I'll use `OUROBOROS_NETWORK_TEST_WITH_NIX`.

ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/Handshake/Test.hs
coot: You're right.

ouroboros-network/ouroboros-network.cabal
coot: Or even better, I'll drop `filepath` and `directory` dependencies to this test and just check paths, this way no `CPP` is needed.

coot:
borse merge

coot:
bors merge

PR #1988, opened 2020-04-22
Pass TestBlockchainTime around in consensus tests

edsko:
This paves the way for having the BlockchainTime API be more restrictive
(may return `PastHorizonException`).


edsko:
bors merge

PR #1989, opened 2020-04-22
Make ChainDB and chain sync independent from BlockchainTime

edsko:

PR #1990, opened 2020-04-22
Further increase use of TestBlockchainTime in the tests

edsko:
Preparing the way for reducing the scope of the blockchain API itself.


edsko:
This touches a lot of files but the changes are minimal:

* moves `TestblockchainTime` to `Test.Util.Time`, away from the main lib
* change `blockUntilSlot` to use a `TestblockchainTime` instead of a `BlockchainTime` 
* replaces the use of `SlotLengths` by `SlotLength` in the `TestBlockchainTime` (because `SlotLengths` will shortly be gone)

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/Time.hs
```diff
@@ -24,40 +19,15 @@ import           GHC.Generics (Generic)
...
-  Settable time
--------------------------------------------------------------------------------}
-
--- | The current slot can be changed by modifying the given 'StrictTVar'.
---
--- 'onSlotChange_' is not implemented and will return an 'error'.
-settableBlockchainTime :: MonadSTM m => StrictTVar m SlotNo -> BlockchainTime m
-settableBlockchainTime varCurSlot = BlockchainTime {
-      getCurrentSlot = readTVar varCurSlot
-    }
```
mrBliss, line 57: I was expecting these to move along to `Test.Util.Time` instead of moving to `BlockchainTime.API`, but since they're going away soon anyway, it doesn't really matter.

edsko: Yeah, I considered that, but they return a `BlockchainTime` not a `TestBockchainTime` and so that didn't feel quite right to me. Add yes, they are gone in https://github.com/input-output-hk/ouroboros-network/pull/1989 .

edsko:
bors merge

Issue #1828, opened 2020-03-19
Allow to use systemd socket activation


coot:
We need to expose API which takes a socket which is already in the listening
state, ready to accept connections.  The socket will be prepared by the systemd
socket activation layer.

coot:
Related to https://github.com/input-output-hk/cardano-node/issues/498

PR #1995, opened 2020-04-23
Change Diffusion interface to permit existing sockets.

karknu:
Needed for systemd adaption.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Diffusion.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Diffusion.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP            #-}
 {-# LANGUAGE DataKinds      #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes     #-}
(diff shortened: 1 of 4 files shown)


karknu:
Implements #1828 .

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -160,16 +168,16 @@ runDataDiffusion tracers
         Async.withAsync (cleanNetworkMutableState networkLocalState) $ \cleanLocalNetworkStateThread ->
 
           -- fork server for local clients
-          Async.withAsync (runLocalServer localSnocket networkLocalState) $ \_ ->
+          Async.withAsync (runLocalServer iocp networkLocalState) $ \_ ->
```
coot, line 171: That's probably not related to this PR, so you can ignore it: elsewhere we use `ioManager` rather than `iocp`.

karknu: Yes, it is called `iocp` across the file.

PR #1999, opened 2020-04-24
Stronger type-level guarantee for `Summary`

edsko:
We knew the summary must always be non-empty, but this wasn't clear from
the types. It wasn't hard to fix that, and this will be helpful in the
hardfork `BlockchainTime` instance.


ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History.hs
```diff
@@ -328,14 +332,46 @@ data EraSummary = EraSummary {
 --
 -- The summary zips 'Shape' with 'Forks', and provides detailed information
 -- about the start and end of each era.
-newtype Summary xs = Summary (AtMost xs EraSummary)
+data Summary xs where
+  -- | We have at most one summary for ech era, and at least one
```
edsko, line 336:
```suggestion
  -- | We have at most one summary for each era, and at least one
```

edsko:
bors merge

PR #2011, opened 2020-04-26
Make it possible to run effects in additional places in mini-protocols.

KtorZ:
A typical client would here want to be able to run extra computation in order to carry on and do whatever it needs to do to collect or constuct the next request to send. Being restricted to only pure computation here makes it fairly hard to pipelined another request unless the request can be constructed out of thin air or, has already been collected and buffered upfront.
    
Similarly, when acquiring a point on the chain, being able to run an extra effect while acquiring can be pretty useful.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -683,7 +683,7 @@ chainSyncClient mkPipelineDecision0 tracer cfg
               (requestNext kis mkPipelineDecision' (Succ n) theirTip candTipBlockNo)
           (Succ n', (CollectOrPipeline, mkPipelineDecision')) ->
             CollectResponse
-              (Just $ SendMsgRequestNextPipelined $
+              (Just $ pure $ SendMsgRequestNextPipelined $
                 requestNext kis mkPipelineDecision' (Succ n) theirTip candTipBlockNo)
               (handleNext kis mkPipelineDecision' n')
           (Succ n', (Collect, mkPipelineDecision')) ->
--- a/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/ChainSync/ExamplesPipelined.hs
+++ b/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/ChainSync/ExamplesPipelined.hs
@@ -102,7 +102,7 @@ chainSyncClientPipelined mkPipelineDecision0 chainvar =
             -- do not directly loop here, but send something; otherwise we
             -- would just build a busy loop polling the driver's receiving
             -- queue.
-            (Just $ SendMsgRequestNextPipelined $ go mkPipelineDecision' (Succ n) cliTipBlockNo srvTip client)
+            (Just $ pure $ SendMsgRequestNextPipelined $ go mkPipelineDecision' (Succ n) cliTipBlockNo srvTip client)
             ClientStNext {
                 recvMsgRollForward = \srvHeader srvTip' -> do
                   addBlock srvHeader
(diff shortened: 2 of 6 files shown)


coot:
Do you need it for pipelining? Have you seen [Ouroboros.Network.Protocol.ChainSync.PipelineDecision](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/Protocol/ChainSync/ClientPipelined.hs).

KtorZ:
@coot Yup, I've seen it! I am actually using a variation of the `pipelineDecisionMax`. The problem with the current pipelined client is that it "forbids" the driver to run additional effects in order to pipeline another query and this is quite limiting if the driver itself is driven by something else, and therefore, needs to run an effect to know whether it should pipeline another request or start collecting some. 

To give a concrete example: 

I have a websocket which can receive requests. These requests are then forwarded to the node through a chain sync client. Now, websockets messages can already be pipelined easily, so it'd be nice if my proxy would also pipeline them instead of forcing them one by one. In this context, my chain sync client is pretty "simple", its input can be modeled as a Queue. 

a) The first request is always pipelined (N = Z). 
b) From there, if the queue is empty, it waits and collects the next response. If the queue isn't empty, it prioritizes pipelining and push the next request (unless there are already MAX requests in flight). Then goes back to step b). 

It means that, at each step, in order to pipeline the next request, I need to lookup whether an existing message has arrived and pop it from the queue. Without the ability to run an intermediate effect in the first callback of `CollectResponse`, it means that necessary I would have to collect the messages first, and then, pipeline them all in one go -- which is basically batching.

coot:
Thanks for the explanation. This is indeed different from how we or the consensus is using it (always ask for more unless maximum pipelining is reached).

I'd expect you will also need to adjust the `chain-sync` client in [Ouroboros.Consensus.MiniProtocol.ChainSync.Client](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs#L345)

I think @mrBliss fixed a CI issue recently, so if you rebase on top of current master  branch it might not timeout.

coot:
It seems that `cardano-shell` failed to build, due to compilation error in `readline` library
https://hydra.iohk.io/build/2477014

KtorZ:
@coot I've rebased the PR on top of a recent revision and also added a similar change in the LocalStateQuery protocol so that one could also run an effect as a response to `MsgAcquired` (it was only possible as a response of `MsgFailure` quite arbitrarily I think :sweat_smile:). 

If you're still happy with these change, please go ahead :pray: ... Thank you! 

---

Should you be interested about use-cases here, it's quite crucial for me to be able to run an extra effect before sending the next message to pipeline: 

- https://github.com/KtorZ/cardano-ogmios/blob/05cfb46a26dd03b7454ecdf2fa0687e8fbe093d2/ogmios-server/src/Ogmios/Bridge.hs#L145-L178

And, similarly, being able to reply after a successful acquire is quite crucial :p 

- https://github.com/KtorZ/cardano-ogmios/blob/05cfb46a26dd03b7454ecdf2fa0687e8fbe093d2/ogmios-server/src/Ogmios/Bridge.hs#L306-L317

coot:
I am fine with the changes; There's a CI failure which seems to be unrelated to this PR: https://hydra.iohk.io/jobset/Cardano/ouroboros-network-pr-2011#tabs-errors

KtorZ:
@coot indeed, as far as I can see, this is also the case on the last 10 builds of the `master` branch :sweat_smile: 

https://hydra.iohk.io/jobset/Cardano/ouroboros-network/evals

The "impossible" seems to be quite easy to achieve :upside_down_face:

coot:
Closing, this changes where merge in #2896

PR #2003, opened 2020-04-24
Hard-fork compatible `BlockchainTime`

edsko:
Closes https://github.com/input-output-hk/ouroboros-network/issues/1637
Closes https://github.com/input-output-hk/ouroboros-network/issues/1205 (opened https://github.com/input-output-hk/ouroboros-network/issues/2002 for the remaining work).

Depends on https://github.com/input-output-hk/ouroboros-network/pull/1989 

This finally unblocks the further development of the hard fork combinator at https://github.com/input-output-hk/ouroboros-network/pull/1175 .

PR #2013, opened 2020-04-27
Remove `Init` phase from `TestBlockchainTime`

edsko:
I was hoping that this simplification would be possible. @nfrisby  explained that the _need_ for it arose from the fact that handlers to `onSlotChange` etc. we not called on the initial slot. That isn't the
case anymore, since we don't provide an initial value in `onSlotChange` (passing `Nothing`) so the action will in fact be called immediately. Unfortunately, the tests are failing, I'm not entirely sure why.

Fixing this will then make it easier to take steps towards a notion of time in the tests that is not slot based.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/Time.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/Time.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE DeriveAnyClass      #-}
-{-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
@@ -24,10 +22,8 @@ module Test.Util.Time (
 
 import           Control.Monad
 import           Data.Word
-import           GHC.Generics (Generic)
 import           GHC.Stack
 
-import           Cardano.Prelude (NoUnexpectedThunks)
 import           Cardano.Slotting.Slot
 
 import           Ouroboros.Consensus.BlockchainTime
@@ -64,13 +60,6 @@ data TestBlockchainTime m = TestBlockchainTime
 newtype NumSlots = NumSlots Word64
   deriving (Show)
 
--- | The current time during a test run.
-data TestClock =
-    Initializing
-    -- ^ This phase has a non-zero but negligible duration.
-  | Running !SlotNo
-  deriving (Eq, Generic, NoUnexpectedThunks)
-
 -- | Construct new blockchain time that ticks at the specified slot duration
 --
 -- NOTE: This is just one way to construct time. We can of course also connect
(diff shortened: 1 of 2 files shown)


nfrisby:
As of commit fa03f1156bac49a3d48ae11cf6fdd317d7c33bbf, there are a couple occurrences of `onSlotChange` that look like this change will also break.

(The first hit below is just a stale comment, I think.)

```
nfrisby@frisbycomp:~/iohk$ git grep -e onSlotChange -- '*hs' | grep -v -e test-infra
ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs:      -- @onSlotChange@ and and before we queried the chain DB for the block
ouroboros-consensus/test-consensus/Test/Consensus/MiniProtocol/ChainSync/Client.hs:    void $ onSlotChange registry testBtime "scheduled updates" $ \slot -> do
ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/Mock.hs:    void $ onSlotChange registry btime "ChainDB.Mock.advanceCurSlot" $
```

E.G. The ChainSync client test will never receive `SlotNo 0` for that `slot` binding.

edsko:
Hmm I thought about this, and I tried a similar fix, but in a much simpler way (I just ticked `numSlots - 1` times) -- and that failed. Now I don't understand why the two aren't equivalent... Maybe I just made a mistake. Let me try agian.

edsko:
so

```haskell
    loop :: StrictTVar m SlotNo -> StrictMVar m () -> m ()
    loop slotVar doneVar = go (numSlots - 1)
      where
        -- count off each requested slot
        go :: Word64 -> m ()
        go 0 = putMVar doneVar () -- signal the end of the final slot
        go n = do
            threadDelay (nominalDelay (getSlotLength slotLen))
            atomically $ modifyTVar slotVar $ succ
            go (n - 1)
```

indeed leads to test failures :thinking:

edsko:
Ok, I think the difference is rather subtle, and points to maybe a brittleness in the tests that we might want to think about addressing. I wrote this test:

```haskell
main :: IO ()
main = withRegistry $ \registry -> do
    logMsg "Initializating clock"
    clock <- newTestBlockchainTime registry (NumSlots 5) (SlotLength 0.1)
    logMsg "Starting"
    _cancel <- onSlotChange registry clock "watch" (logMsg . show)
    testBlockchainTimeDone clock
    logMsg "Aaaaaand done"

logMsg :: String -> IO ()
logMsg str = do
    time <- getCurrentTime
    putStrLn $ show time ++ ": " ++ str
```

With my version, this outputs something like

```
07:42:38: Initializating clock
07:42:38: Starting
07:42:38: SlotNo {unSlotNo = 0}
07:42:40: SlotNo {unSlotNo = 1}
07:42:42: SlotNo {unSlotNo = 2}
07:42:44: SlotNo {unSlotNo = 3}
07:42:46: SlotNo {unSlotNo = 4}
07:42:46: Aaaaaand done
```

and sometimes

```
Initializating clock
Starting
SlotNo {unSlotNo = 0}
SlotNo {unSlotNo = 1}
SlotNo {unSlotNo = 2}
SlotNo {unSlotNo = 3}
Aaaaaand done
SlotNo {unSlotNo = 4}
```

The order of the last two is not guaranteed, and depends on the precise order in which the various threads are woken up.

Your version provides a stronger guarantee: the`done` variable will be set strictly _after_ the last slot was written; in fact, a full slot length after:

```
07:47:05: Initializating clock
07:47:05: Starting
07:47:05: SlotNo {unSlotNo = 0}
07:47:07: SlotNo {unSlotNo = 1}
07:47:09: SlotNo {unSlotNo = 2}
07:47:11: SlotNo {unSlotNo = 3}
07:47:13: SlotNo {unSlotNo = 4}
07:47:15: Aaaaaand done
```

Note the full additional two second delay.

edsko:
Ok, this was useful. I will see how this plays out in my new test time abstraction.

edsko:
Ok, I have incorporated the lesson from this branch in https://github.com/input-output-hk/ouroboros-network/pull/2003 .

Issue #2004, opened 2020-04-24
Shelley: initial UTxO


mrBliss:
`protocolInfoShelley` still mentions a dummy `genesisUtxO`:
https://github.com/input-output-hk/ouroboros-network/blob/eec4a136bb51d882ab4cf4de552f1ce978459f56/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs#L251

We should define a proper initial UTxO.

nc6:
I'm not sure where that TODO comes from - to be honest I think that initial UTxO is fine. The ledger rules don't really care about what that value is - it simply allows anything in the initial UTxO.

mrBliss:
> I'm not sure where that TODO comes from - to be honest I think that initial UTxO is fine. The ledger rules don't really care about what that value is - it simply allows anything in the initial UTxO.

I suppose it's more about the string `"In the beginning"`. Do we want that in the release?

nc6:
What random string would you rather have? :-D

edsko:
I don't care very much about the string, but _is_ this the right initial UTxO? We want people to actually be able to interact with the testnet no? Doesn't that mean we need a carefully crafted initial UTxO so that we can hand out keys etc?

nc6:
The string in here concerns the _input_ to the initial UTxO. It's completely arbitrary, and is never checked anywhere in the rules. The outputs to the initial UTxO are determined by `ShelleyGenesis` - in testing, we simply allocate all funds to the core node keys.

edsko:
Ok, yes, I misunderstood. So, the string doesn't matter (the hash of the "virtual" genesis transaction). However, the _index_ does certainly matter: we cannot have a single `(txId, index)` map to multiple outputs. Indeed, this map will always be a singleton map. 

How come the tests work with this?

edsko:
Moreover, that number _matters_. It should be part of `sgInitialFunds` (otherwise people would have to search the UTxO to find which one is theirs? I guess that's also fine.)

nc6:
> However, the index does certainly matter: we cannot have a single (txId, index) map to multiple outputs. Indeed, this map will always be a singleton map.

I mean, the answer is basically that we _can_. Our rules are inductive. Anything in at the beginning is valid by definition. If we wanted to use the index to identify specific initial fundholders, we could, but it would just be for convenience.

[...]

PR #2017, opened 2020-04-28
Initialise the initial UTxO correctly.

nc6:
UTxO is implemented as a map. Previously the keys were duplicates and,
as such, there was only one entry.

Addresses #2004


edsko:
bors merge

edsko:
Right, I asked about that number in https://github.com/input-output-hk/ouroboros-network/issues/2004#issuecomment-619782861 , but then figured wallets need to do "recovery" anyway, and so they'd simply discover which one was theirs.

nc6:
Superseded by https://github.com/input-output-hk/ouroboros-network/pull/2038

PR #2018, opened 2020-04-29
Rearange PeerSelection

coot:
This PR rearanges modules under `Ouroboros.Netowk.PeerSelection` (for breviety,
the module prefix is skipped in what follows):

* `Governor.Types` contains common types, it's not possible to mergeg it with the `Types` module due to cirular dependency with `KnonwPeers`
* `Governor.ActivePeers`
* `Governor.EstablishedPeers`
* `Governor.KnownPeers`
* `Governor.RootPeers` - includes only public root peers decision
* `Governor.Monitor`   -which contains monitoring tasks for local roots, jobs and connections.

- Documented Guarded data type
- Added MkGuardedDecision type alias
- Added job-pool package
- Rename governor decisions
- Split Governor into sub-modules


job-pool/src/Control/Concurrent/JobPool.hs
```diff
@@ -3,7 +3,7 @@
 {-# LANGUAGE NamedFieldPuns #-}
 
 
-module Network.Mux.JobPool (
+module Control.Concurrent.JobPool (
```
mrBliss, line 6: Splitting things up in separate packages is generally a good thing, but does this 80 line module really deserve a separate package? Will it grow much in the future?

The list of packages is already quite long and there are so many places (CI stuff, `cabal.project` in downstream repos, ...) mentioning the packages that can be forgotten to update.

coot: > The list of packages is already quite long and there are so many places (CI stuff, cabal.project in downstream repos, ...) mentioning the packages that can be forgotten to update.

Updating the hashes in `cabal.project` is a valid concern, though we have `cardano-repo-tool` for that, so I am not worried about it.  And probably sooner than later we will move to `cabal-3.2` where speciyfing hashes will be easier (just one per repo, rather than one per _cabal project_).  

The problem with this module is that exporting it from  `network-mux` does not seem right.  It would be a nice case for a public library in `network-mux`, but this is not available.

Another option is to add it to `io-sim-classes`, though this is even worse ides.

From all the available options, probably moving it back (but still keeping it under `Control.Concurrent` scope) is the least wrong option.

coot:
bors merge

coot:
bors merge

coot:
bors merge

A strange error by `check-stylish`:
```
error: while evaluating anonymous function at /var/lib/buildkite-agent/builds/packet-ipxe-3-ci3-2/input-output-hk/ouroboros-network/shell.nix:3:1, called from undefined position:
...
getting status of '/var/lib/buildkite-agent/builds/packet-ipxe-3-ci3-2/input-output-hk/ouroboros-network/.git/objects/pack/pack-377d877718958c9fa46b056ed79134c1ac488888.pack': No such file or directory
```

PR #2020, opened 2020-04-29
Use correct slot length

edsko:
The mock PBFT, real PBFT and dual PBFT were all setting the slot length independently from the slot length in the `TestConfig`, so that we were working with two inconsistent values. This didn't matter when we were working with a fake clock that was `SlotNo` based, but it is causing test failures in the hard fork blockchain time branch. (It is also just incorrect, of course.)

### DIFF
--- a/ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
+++ b/ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
@@ -592,7 +592,7 @@ tests = testGroup "RealPBFT" $
     ]
   where
     defaultSlotLength :: SlotLength
-    defaultSlotLength = SlotLength 1
+    defaultSlotLength = slotLengthFromSec 1
 
 prop_deterministicPlan :: PBftParams -> NumSlots -> NumCoreNodes -> Property
 prop_deterministicPlan params numSlots numCoreNodes =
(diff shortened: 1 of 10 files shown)


ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
```diff
@@ -620,11 +621,14 @@ prop_setup_coreNodeId numCoreNodes coreNodeId =
 
     params :: PBftParams
     params = realPBftParams dummyK numCoreNodes
-    dummyK = SecurityParam 10   -- not really used
+
+    -- not really used
+    dummyK       = SecurityParam 10
+    dummySlotLen = slotLengthFromSec 20
```
mrBliss, line 627: Can you explain this to me? I suppose they will be overridden by some other `k` and `SlotLength`, but where/how?

edsko: Not a clue. I think this test is just a sanity check on some config. You'd have to ask @nfrisby . But I think these values just don't matter.

edsko:
bors merge

ouroboros-consensus-byron/test/Test/ThreadNet/RealPBFT.hs
nfrisby: In the scope of these definitions (10 lines up) you see that the property doesn't depend on these. But it does depend on the `ProtocolInfo`, which needs the params, which needs these values. If fields were sufficiently lazy (or similar), we could use `error` here.

edsko: :+1:

nfrisby:
Just a generally nerdy comment appreciating that PR 2020 in the year 2020 is about getting time right, and moreover it's in hindsight.

edsko:
bors merge

edsko:
bors merge

PR #2022, opened 2020-04-30
Haddocks for internal modules

coot:
Added --haddock-internl flag `cabal haddock` command.
 Also create the OUTPUT_DIR if it does not exist.

### DIFF
--- a/scripts/haddocs.sh
+++ b/scripts/haddocs.sh
@@ -34,6 +34,7 @@ HADDOCK_OPTS=(
     --builddir "${BUILD_DIR}"
     --disable-optimization
     --haddock-all
+    --haddock-internal
     --haddock-html
     --haddock-quickjump
     --haddock-hyperlink-source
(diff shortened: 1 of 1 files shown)


scripts/haddocs.sh
```diff
@@ -49,6 +50,10 @@ elif [ ${REGENERATE} != "false" ]; then
   cabal haddock "${HADDOCK_OPTS[@]}" ${REGENERATE}
 fi
 
+if [[ !( -d ${OUTPUT_DIR} ) ]]; then
+  mkdir ${OUTPUT_DIR}
+fi
+
```
karknu, line 56: Couldn't this be replaced with `mkdir -p ${OUTPUT_DIR}`?

coot:
bors merge

scripts/haddocs.sh
coot: Yes, that's better.

coot:
borse merge

coot:
bors merge

scripts/haddocs.sh
mrBliss: The point of `-p` is that you no longer need to check whether it exists or not.

PR #2024, opened 2020-04-30
Implement HasNetworkProtocolVersion instance for ShelleyBlock

intricate:
For a particular issue I'm working on in `cardano-node` (more specifically, `cardano-api` and `cardano-cli`), I'll need to be able to determine and utilize the node-to-client protocol versions supported by a Shelley node. This PR adds support for just that.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
@@ -1,0 +1,41 @@
+{-# LANGUAGE TypeFamilies #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Ouroboros.Consensus.Shelley.Ledger.NetworkProtocolVersion (
+    ShelleyNodeToNodeVersion(..)
+  , ShelleyNodeToClientVersion(..)
+  ) where
+
+import           Data.List.NonEmpty (NonEmpty (..))
+
+import qualified Ouroboros.Network.NodeToClient as N
+import qualified Ouroboros.Network.NodeToNode as N
+
+import           Ouroboros.Consensus.Node.NetworkProtocolVersion
+
+import           Ouroboros.Consensus.Shelley.Ledger.Block
+
+data ShelleyNodeToNodeVersion = ShelleyNodeToNodeVersion1
+  deriving (Show, Eq, Ord, Enum, Bounded)
+
+data ShelleyNodeToClientVersion = ShelleyNodeToClientVersion1
+  deriving (Show, Eq, Ord, Enum, Bounded)
+
+instance HasNetworkProtocolVersion (ShelleyBlock c) where
+  type NodeToNodeVersion   (ShelleyBlock c) = ShelleyNodeToNodeVersion
+  type NodeToClientVersion (ShelleyBlock c) = ShelleyNodeToClientVersion
+
+  supportedNodeToNodeVersions   _ = ShelleyNodeToNodeVersion1
+                                  :| []
+  supportedNodeToClientVersions _ = ShelleyNodeToClientVersion1
+                                  :| []
+
+  mostRecentNodeToNodeVersion   _ = ShelleyNodeToNodeVersion1
+  mostRecentNodeToClientVersion _ = ShelleyNodeToClientVersion1
+
+  nodeToNodeProtocolVersion _ ShelleyNodeToNodeVersion1 = N.NodeToNodeV_1
+
+  -- From the beginning, Shelley supports version 2 of the node-to-client
+  -- protocol (i.e. the local state query protocol is enabled from the start).
+  nodeToClientProtocolVersion _ ShelleyNodeToClientVersion1 = N.NodeToClientV_2
--- a/ouroboros-consensus-shelley/ouroboros-consensus-shelley.cabal
+++ b/ouroboros-consensus-shelley/ouroboros-consensus-shelley.cabal
@@ -34,6 +34,7 @@ library
                        Ouroboros.Consensus.Shelley.Ledger.Integrity
                        Ouroboros.Consensus.Shelley.Ledger.Ledger
                        Ouroboros.Consensus.Shelley.Ledger.Mempool
+                       Ouroboros.Consensus.Shelley.Ledger.NetworkProtocolVersion
                        Ouroboros.Consensus.Shelley.Ledger.TPraos
                        Ouroboros.Consensus.Shelley.Node
                        Ouroboros.Consensus.Shelley.Protocol
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger.hs
@@ -8,3 +8,4 @@ import           Ouroboros.Consensus.Shelley.Ledger.Forge as X
 import           Ouroboros.Consensus.Shelley.Ledger.Integrity as X
 import           Ouroboros.Consensus.Shelley.Ledger.Ledger as X
 import           Ouroboros.Consensus.Shelley.Ledger.Mempool as X
+import           Ouroboros.Consensus.Shelley.Ledger.NetworkProtocolVersion as X
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
@@ -45,7 +45,6 @@ import qualified Ouroboros.Consensus.HardFork.History as HardFork
 import           Ouroboros.Consensus.HeaderValidation
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Ledger.Extended
-import           Ouroboros.Consensus.Node.NetworkProtocolVersion
 import           Ouroboros.Consensus.Node.ProtocolInfo
 import           Ouroboros.Consensus.Node.Run
 import           Ouroboros.Consensus.Node.State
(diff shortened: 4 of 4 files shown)


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE TypeFamilies #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Ouroboros.Consensus.Shelley.Ledger.NetworkProtocolVersion (
```
intricate, line 5: @mrBliss: For this, I was simply following the example of `Ouroboros.Consensus.Byron.Ledger.NetworkProtocolVersion`. If this isn't a suitable location, would you prefer I move this module somewhere else?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -0,0 +1,34 @@
...
+import           Ouroboros.Consensus.Node.NetworkProtocolVersion
+
+import           Ouroboros.Consensus.Shelley.Ledger.Block
+
+data ShelleyNodeToClientVersion =
+    -- | No local state query protocol
+    ShelleyNodeToClientVersion1
+
+    -- | Use local state query protocol
+  | ShelleyNodeToClientVersion2
```
mrBliss, line 22: No need for both, we don't need to support "No local state query protocol". So just keep `ShelleyNodeToClientVersion1`, remove the comment and the other constructor.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -0,0 +1,34 @@
...
+instance HasNetworkProtocolVersion (ShelleyBlock c) where
+  type NodeToClientVersion (ShelleyBlock c) = ShelleyNodeToClientVersion
+
+  supportedNodeToClientVersions _ = ShelleyNodeToClientVersion1
+                                  :| [ ShelleyNodeToClientVersion2 ]
+
+  mostRecentNodeToClientVersion _ = ShelleyNodeToClientVersion2
+
+  nodeToClientProtocolVersion _ ShelleyNodeToClientVersion1 = N.NodeToClientV_1
+  nodeToClientProtocolVersion _ ShelleyNodeToClientVersion2 = N.NodeToClientV_2
```
mrBliss, line 34:
```suggestion
  nodeToClientProtocolVersion _ ShelleyNodeToClientVersion1 = N.NodeToClientV_2
```

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -0,0 +1,34 @@
...
+data ShelleyNodeToClientVersion =
+    -- | No local state query protocol
+    ShelleyNodeToClientVersion1
+
+    -- | Use local state query protocol
+  | ShelleyNodeToClientVersion2
+  deriving (Show, Eq, Ord, Enum, Bounded)
+
+instance HasNetworkProtocolVersion (ShelleyBlock c) where
+  type NodeToClientVersion (ShelleyBlock c) = ShelleyNodeToClientVersion
```
mrBliss, line 26: While you're at it, you might as well define `ShelleyNodeToNodeVersion`

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
mrBliss: The location is fine, I think.

intricate:
> I don't really understand why you need this change, can you explain?

@mrBliss: So I need to utilize the local state query protocol in order to query the current protocol parameters as well as the UTxO (filtered by specific addresses). 

In order to do this, I'm using [`connectTo`](https://github.com/input-output-hk/ouroboros-network/blob/7a6ae3a2e3d6ac4494a55d1166ae80f9f0dc5cfa/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs#L302) (now that I'm looking through the code, maybe I should instead use `ncSubscriptionWorker`) which requires a `Versions` parameter. 

[From the examples I've seen](https://github.com/input-output-hk/cardano-node/blob/15d26a05a5302efb0184cdaa9146a49f240a7243/cardano-cli/src/Cardano/CLI/Ops.hs#L459-L478), the `Versions` value is typically constructed based on the `supportedNodeToClientVersions`:

```
foldMapVersions
  (\v ->
    versionedNodeToClientProtocols
      (nodeToClientProtocolVersion proxy v)
      versionData
      (protocols v))
  (supportedNodeToClientVersions proxy)
```

Perhaps I shouldn't have said that I _needed_ it because I could probably also do it like this :smile: :

```
versionedNodeToClientProtocols
  NodeToClientV_2
  versionData
  NodeToClientProtocols { ... }
```

However, it does seem nicer to use the supported node-to-client versions.

mrBliss:
You should already be able to use
```haskell
foldMapVersions
  (\v ->
    versionedNodeToClientProtocols
      (nodeToClientProtocolVersion proxy v)
      versionData
      (protocols v))
  (supportedNodeToClientVersions proxy)
```
with the current default instance (so without this PR). It will use `()` instead of `data ShelleyNodeToClientVersion = ShelleyNodeToClientVersion1`.

mrBliss:
Eventually, we should turn it in a proper data type, certainly when we *do* need different versions, but for now the default should work perfectly fine.

EDIT: hmm, maybe the default `nodeToClientProtocolVersion _ () = N.NodeToClientV_1` is not right for Shelley. It should return `NodeToClientV_2` for Shelley. So I still suggest going forward with this PR.

intricate:
> ... with the current default instance (so without this PR). It will use () instead of data ShelleyNodeToClientVersion = ShelleyNodeToClientVersion1.

Ah, right but I need the `nodeToClientProtocolVersion` to be `NodeToClientV_2` in order to use the local state query protocol.

mrBliss:
> > ... with the current default instance (so without this PR). It will use () instead of data ShelleyNodeToClientVersion = ShelleyNodeToClientVersion1.
> 
> Ah, right but I need the `nodeToClientProtocolVersion` to be `NodeToClientV_2` in order to use the local state query protocol.

Right, see my EDIT in https://github.com/input-output-hk/ouroboros-network/pull/2024#issuecomment-621681299 :slightly_smiling_face:

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -0,0 +1,38 @@
...
+import           Data.List.NonEmpty (NonEmpty (..))
+
+import qualified Ouroboros.Network.NodeToClient as N
+import qualified Ouroboros.Network.NodeToNode as N
+
+import           Ouroboros.Consensus.Node.NetworkProtocolVersion
+
+import           Ouroboros.Consensus.Shelley.Ledger.Block
+
+data ShelleyNodeToNodeVersion = ShelleyNodeToNodeVersion1
```
mrBliss, line 18: Export this one too

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -0,0 +1,38 @@
...
+                                  :| []
+  supportedNodeToClientVersions _ = ShelleyNodeToClientVersion1
+                                  :| []
+
+  mostRecentNodeToNodeVersion   _ = ShelleyNodeToNodeVersion1
+  mostRecentNodeToClientVersion _ = ShelleyNodeToClientVersion1
+
+  nodeToNodeProtocolVersion _ ShelleyNodeToNodeVersion1 = N.NodeToNodeV_1
+
+  nodeToClientProtocolVersion _ ShelleyNodeToClientVersion1 = N.NodeToClientV_2
```
mrBliss, line 38: Document that Shelley supports V_2, i.e., the enabled local state query protocol, from the start.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -0,0 +1,38 @@
+{-# LANGUAGE TypeFamilies #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Ouroboros.Consensus.Shelley.Ledger.NetworkProtocolVersion (
```
mrBliss, line 5: Re-export this module in `Ouroboros.Consensus.Shelley.Ledger`, just like for Byron.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
intricate: Oops.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/NetworkProtocolVersion.hs
```diff
@@ -0,0 +1,41 @@
...
+                                  :| []
+
+  mostRecentNodeToNodeVersion   _ = ShelleyNodeToNodeVersion1
+  mostRecentNodeToClientVersion _ = ShelleyNodeToClientVersion1
+
+  nodeToNodeProtocolVersion _ ShelleyNodeToNodeVersion1 = N.NodeToNodeV_1
+
+  -- From the beginning, Shelley supports version 2 of the node-to-client
+  -- protocol (i.e. the local state query protocol is enabled from the start).
+  nodeToClientProtocolVersion _ ShelleyNodeToClientVersion1 = N.NodeToClientV_2
```
edsko, line 41: :+1:

PR #2037, opened 2020-05-03
Use `TickedLedger` for `updateConsensusState`

edsko:
The documentation for `updateConsensusState` said that it would be given the _validated_ ledger view. That is not correct: it is passed the _ticked_ ledger view. We didn't make that clear in the types though; now we do. Similarly, the leader check was done using a ticked ledger view, but this was also not clear from the types.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
@@ -103,7 +103,7 @@ instance TPraosCrypto c => IsLedger (LedgerState (ShelleyBlock c)) where
     globals
     slotNo
     (ShelleyLedgerState pt history bhState) =
-      TickedLedgerState slotNo
+      Ticked slotNo
         . ShelleyLedgerState pt history
         $ SL.applyTickTransition globals bhState slotNo
 
@@ -121,7 +121,7 @@ instance TPraosCrypto c
   --
   applyLedgerBlock globals
                    blk
-                   TickedLedgerState {
+                   Ticked {
                        tickedLedgerState = ShelleyLedgerState {
                            history
                          , shelleyState = oldShelleyState
@@ -316,7 +316,7 @@ instance Crypto c => ValidateEnvelope (ShelleyBlock c) where
   type OtherHeaderEnvelopeError (ShelleyBlock c) =
     STS.PredicateFailure (STS.CHAIN c)
 
-  validateEnvelope cfg (TickedLedgerState _ ledgerView) oldTip hdr = do
+  validateEnvelope cfg (Ticked _ ledgerView) oldTip hdr = do
       -- In addition to the default 'validateEnvelope' ...
       defaultValidateEnvelope oldTip hdr
       -- ... perform the @chainChecks@ that are part of the @CHAIN@ rule.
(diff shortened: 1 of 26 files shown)


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -316,7 +316,7 @@ instance Crypto c => ValidateEnvelope (ShelleyBlock c) where
   type OtherHeaderEnvelopeError (ShelleyBlock c) =
     STS.PredicateFailure (STS.CHAIN c)
 
-  validateEnvelope cfg ledgerView oldTip hdr = do
+  validateEnvelope cfg (TickedLedgerState _ ledgerView) oldTip hdr = do
```
nc6, line 319: This is somewhat confusing. At least from the names here, this suggests we're taking the full ledger state, which, ticked or not, would allow invalid things to happen here. Surely the ledger view is the correct concept to pass?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
```diff
@@ -324,7 +324,7 @@ instance TPraosCrypto c => ConsensusProtocol (TPraos c) where
       TPraosIsACoreNode{}  -> True
       TPraosIsNotACoreNode -> False
 
-  checkIsLeader cfg@TPraosConfig{..} slot lv cs =
+  checkIsLeader cfg@TPraosConfig{..} (TickedLedgerState slot lv) cs =
```
nc6, line 327: Likewise here

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
edsko: Yeah, fair enough. WIll rename.

edsko:
> I'm also not really sure what the semantics really are meant to be of a "ticked" ledger view? E.g there is a comment saying "forecasting is equivalent to ticking", which is true.. but it's not clear what slot it should be ticked to here - is this a view from a ticked ledger state, or a ticked view (I guess, a view for a specific slot?)

Not sure I follow this. It's the ledger view protected forward to a particular slot, which by that lemma is equivalent to what the chain tick is doing? The slot is clear no? It's the slot we protected to... should I rephrase this somehow?

edsko:
@nc6 Pushed a commit that renames the type to `Ticked` as we discussed, and also attempts to improve the docstring.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Abstract.hs
```diff
@@ -95,13 +96,24 @@ class ( -- Requirements on the ledger state itself
...
+  applyChainTick :: LedgerCfg l -> SlotNo -> l -> Ticked l
 
--- | Ledger state with the chain tick function already applied
+-- | Mark a ledger state or ledger view as " ticked "
 --
--- 'applyChainTick' is intended to mark the passage of time, without changing
--- the tip of the underlying ledger (i.e., no blocks have been applied).
-data TickedLedger l = TickedLedgerState {
+-- Ticking refers to the passage of time (the ticking of the clock). When a
+-- legder state of ledger view is marked as ticked, it means that changes to
```
nc6, line 104:
```suggestion
-- legder state or ledger view is marked as ticked, it means that changes to
```

nc6, line 104:
```suggestion
-- legder state of ledger view is marked as ticked, it means
```

edsko:
bors merge

nc6:
bors retry

PR #2041, opened 2020-05-06

nc6:
This pulls in two things:

- Crypto changes in base
- Key changes in cardano-ledger-specs


nc6:
I think it is now only used in the tests

edsko:
Ah, so we can't remove it from `Protocol.Abstract` yet?

nc6:
Ah, sorry, it's also still used in `evalCertified` for VRF.

PR #2042, opened 2020-05-06
Move LedgerState out of UpdateLedger

edsko:
This makes it possible to define the `LedgerState` independent from the `UpdateLedger` instance, which is useful sometimes.

Note that this is just one commit; the rest is #2037, which should be merged soon.


edsko:
> I'll take your word on it being useful!

It is :) In particular, it means I can define the state of the hard fork ledger, and some operations on it, before I am ready to give the full `UpdateLedger` instance.

nc6:
bors merge

PR #1942, opened 2020-04-10
Replace `anachronisticLedgerView` by `ledgerViewForecastAt`

edsko:
See #1933 for detailed description.

PR #2048, opened 2020-05-08
JIT EBBs in the ThreadNet tests require ExtLedgerState in block production

### LINKED ISSUE #2047: JIT EBB require ExtLedgerState in block production
PR #1942 changed block production so that the node only provides `LedgerState` instead of the fuller `ExtLedgerState`. This caused some of the logic regarding the Just-In-Time EBB generation in the Byron `ThreadNet` tests to no longer hold.

We have seen this manifest as a reproducible test failure on master. This Issue is to resolve such failures.

nfrisby:
Fixes #2047.

PR #1942 changed blockProduction to take a LedgerState instead of
ExtLedgerState, which suffices for all forging. However, the ThreadNet tests'
current approach to EBBs requires anticipating during block production whether
the produced block will be invalid in anyway. (The node does not emit the EBB
unless the EBB unless it is also emitting a valid block.) And that anticipatory
validation must include protocol-level errors, which requires the full
ExtLedgerState.

It may be possible to instead change the tests to allow for EBBs without
successors, but I'm don't see how, so I currently propose leaving that for future work if we
prioritize it.

Edit: PR 2176 essentially fixed this failure mode _en passant_. This PR now just simplifies the JIT EBB logic accordingly.

### DIFF
--- a/ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
+++ b/ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
@@ -580,6 +580,28 @@ tests = testGroup "RealPBFT" $
             , slotLength   = defaultSlotLength
             , initSeed     = Seed {getSeed = (8051309618816278461,2819388114162022931,16483461939305597384,11191453672390756304,8021847551866528244)}
             }
+    , testProperty "HeaderProtocolError prevents JIT EBB emission" $
+          -- "extra" EBB generated in anticipation of a block that ends up
+          -- being PBftExceededSignThreshold
+          --
+          -- PR 1942 reduced the input of blockProduction from ExtLedgerState
+          -- to just LedgerState, but that does not provide enough information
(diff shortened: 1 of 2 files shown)


nfrisby:
@edsko and I reviewed on a call. Thanks!

Plan: re-expanding the `produceBlock` API as this PR currently does is undesirable. Edsko saw alignment with other concerns and so pressed for the new `CannotLead` mechanism (see merged PR #2169). I'm going to rebase this PR to rely on that instead; the goal is that the tests should never generate an invalid block because the `CannotLead` check will abort the thread upstream. Until Issue #2173, this will also involve dialing back the `Gen NodeJoinPlan` so that it avoids all `PBftExceededSignThreshold` errors.

nfrisby:
Once PR #2176 is merged, this PR will change dramatically; it will merely simplify the JIT EBB generator, under the assumption that (at least in the test suite for now) nodes now never forge invalid blocks!

edsko:
bors merge

Issue #2029, opened 2020-04-30
Make consensus tests work without `UnsafeUnbounded`


edsko:
Right now the tests fail when this is not used, which means that they assume we can look ahead as far as we want.

mrBliss:
@nfrisby What do you think is left to do before we can close this ticket?

nfrisby:
PR #2066 will make this fundamentally possible. But what should be done for
eg `HasHardForkHistory ByronBlock`, which currently uses
`neverForksHardForkSummary`, which uses `EraUnbounded`?

edsko:
That's fine, that instance is only relevant when using Byron as a standalone ledger. This remark in the long comment for `hardForkSummary` is relevant:

> Critically, the HFC implements 'hardForkSummary' directly and does not call 'hardForkSummary' in the underlying ledgers.

PR #2053, opened 2020-05-08
ThreadNet.General: document primary assumptions

nfrisby:
This PR adds a comment that documents the primary assumptions currently made in the `ThreadNet` tests.

These must change to properly test Praos (specifically semi-synchrony allows for non-trivial chain diffusion delay). They also guide #2029.


edsko:
bors merge

nfrisby:
> 
> 
> Sounds good. At some point (after Shelley) we probably need to allocate a chunk of time to redesign these tests with the improved understanding we (mostly, you) now have, and try to see if we can come up with something that is a little less brittle.
> 
> It would also be good if we could separate out general concerns from protocol specific concerns a bit more cleanly; a large part of the the docs in this PR talk about PBFT, but they are part of `prop_general`. Doing a more clean separation however may well require that redesign.

Absolutely looking forward to it, once I can put down this fire extinguisher.

edsko:
> Absolutely looking forward to it, once we can put down these fire extinguishers! :)

Aye :)

PR #2054, opened 2020-05-08
Removed Enum and Bounded constraints from version numbers

coot:
Except from `NodeToNodeVersion` and `NodeToClientVersion`. We use those
instances in tests.


coot:
@edsko unfortunately we use both `Enum` and `Bounded` instances in tests, but at least they disappear from the api.

coot:
bors merge

PR #2057, opened 2020-05-09
RunNode improvements

edsko:
This PR makes two small improvements I need for the hard fork combinator:

* We pass a restricted `InitChainDB` API to `nodeInitChainDB`. Having the full power of the chain DB available there was unnecessary, and making it difficult to define a `RunNode` instance for a block in terms of a `RunNode` instance for another block.
* We pass the `LedgerConfig` to the encoder/decoder for `LedgerState`. This this goes somewhat counter in spirit to the work by @nc6 in #2050 (and @dcoutts in #2056), but it should not affect clients because we never send full ledger state across.

This PR also includes a cleanup commit that removes some unnecessary proxies.


edsko:
bors merge

PR #2034, opened 2020-05-02
Prelude in C♯ minor: The Hard Fork Combinator

edsko:
Much like [Rachmaninoff's famous prelude](https://en.wikipedia.org/wiki/Prelude_in_C-sharp_minor_(Rachmaninoff)), the work on the hard fork combinator consists of three phases:

* Various months of working through all the grim consequences of having transition points in the chain.
* A period of highly agitated hacking and writing the actual hard fork combinator (this PR)
* A quiet ending where we introduce the hard fork combinator on mainnet and everything Just Works (TM)

The new hard fork combinator is `n`-ary rather than binary. Not only does this mean we will be able to smoothly handle the transition from Shelley to post-Shelley (Shelley+MSIG, Goguen, and beyond), it also means that the code is very uniform and consistent; this also has various advantages for testing. For example, we can instantiate the combinator with a _single_ block, and then test that "hard forking between just one block" (degenerate case) is equivalent to just running with that single block without the hard fork combinator meditating.

# Supported interfaces

The consensus layer is a large piece of code that offers a lot of interfaces, and the hard fork combinator must deal with virtually all of those interfaces.

This PR is still WIP, but the vast majority of the interfaces have been completed:

## Consensus protocol integration

- [x] `ConsensusProtocol` instance
      This includes chain selection, the leader check, consensus-specific header validation and consensus-specific rollback.

- [x] `BlockSupportsProtocol` instance      

## Block integration

- [x] `BlockHasCodecConfig` instance
- [x] `GetHeader` instance
- [x] `StandardHash` instance
- [x] `HasHeader` instances (block and header)
- [x] `HasAnnTip` instance (depends on #2059)
- [x] `ValidateEnvelope` instance (#2071)
     The trickiest part here is to verify that hashes line up across the transition point.

## Ledger integration

- [x] `IsLedger` instance
- [x] `ApplyBlock` instance
- [x] `UpdateLedger` instance
- [x] `HasHardForkHistory` instance
- [x] `LedgerSupportsProtocol` instance

## Mempool integration

- [x] `ApplyTx` instance
- [x] `HasTxId` instance

## LocalStateQuery integration

- [x] `ShowQuery` instance
- [x] `QueryLedger` instance

# Integration with Byron / Shelley

- [ ] `CanHardFork` instance (#2072)
      Although the hard fork combinator takes care of most of the details, some ledger specific info must be required; in particular, the translation from the final Byron ledger state to the initial Shelley ledger state.

[...]

PR #2059, opened 2020-05-09
Generalize `AnnTip`

edsko:
Previously `AnnTip` looked like

```haskell
 data AnnTip blk = AnnTip {
       annTipSlotNo  :: !SlotNo
     , annTipHash    :: !(HeaderHash blk)
     , annTipBlockNo :: !BlockNo
     , annTipInfo    :: !(TipInfo blk)
     }
```

That double use of `blk` is problematic. Consider what happens when we instantiate this to `HardForkBlock`: the hash of a hard fork block is a hash from one of the eras (Byron hash, Shelley hash, ..), and similarly the `TipInfo` of a `HardForkBlock` is the tip info from a particular era. But if we instantiate this twice, independently, then we lose the information that these two must be _related_; it doesns't make sense to have a block with a Byron header but with Shelley info.

Therefore in this PR we delete `annTipHash` altogether, and instead insist that we must be able to _derive_ the `HeaderHash `from the `TipInfo`.

The first 4 commits in this PR are all designed specifically to ensure that introducing this change does not break backwards compatibility. The encoders/decoders for `AnnTip` still encode the information in precisely the same order; only the in-memory representation is now different.


ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
```diff
@@ -442,53 +450,82 @@ validateHeader cfg ledgerView hdr st = do
...
+                    => (HeaderHash blk -> Encoding)
+                    -> (AnnTip     blk -> Encoding)
+defaultEncodeAnnTip encodeHash AnnTip{..} = mconcat [
       encodeListLen 4
     , encode     annTipSlotNo
-    , encodeHash annTipHash
+    , encodeHash annTipInfo
     , encode     annTipBlockNo
-    , encodeInfo annTipInfo
+      -- TODO: Useless field. Remove when OK to break binary compatibility.
```
mrBliss, line 461: How will we think of this when the opportunity is there to break binary compatibility?

Do we actually expect this to happen?

Alternatively, we do the simple naive thing in the encoder and make the decoder backwards compatible by branching on the size (4 -> old, 3 -> new).

edsko: Once we move the hard fork combinator we break binary compatibility, I think that's the natural point to introduce this. When we create the tickets for the various TODOs in #2034 and create one to switch the main code base over to `HardForkBlock [Byron, Shelley]`, maybe let's change it then?

edsko: https://github.com/input-output-hk/ouroboros-network/issues/2062

edsko:
bors merge

PR #2060, opened 2020-05-11
Add a SubmitResult type to hold a transaction submission result

erikd:
This type is isomorphic with the Maybe it replaces, but the name
better describes its purpose and usage.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Protocol/LocalTxSubmission/Type.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Protocol/LocalTxSubmission/Type.hs
@@ -45,6 +45,13 @@ data LocalTxSubmission tx reject where
   StDone   :: LocalTxSubmission tx reject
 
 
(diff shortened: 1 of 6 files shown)


ouroboros-network/src/Ouroboros/Network/Protocol/LocalTxSubmission/Type.hs
```diff
@@ -45,6 +45,13 @@ data LocalTxSubmission tx reject where
   StDone   :: LocalTxSubmission tx reject
 
 
+-- | Isomorphic with Maybe but with a name that better describes its purpose and
+-- usage.
+data SubmitResult reason
```
coot, line 50: @erikd I think it makes more sense if we define this type in `Client`.  This is client related only type, rather than a part of the protocol.

erikd: `Server` also needs this type.

PR #2065, opened 2020-05-11
Move txSize from ApplyTx to RunNode and Mempool

mrBliss:
The `txSize` method returns the size of a serialised transaction in bytes.
None of the other consensus classes deal with serialisation-related details,
all such details are collected in the `RunNode` class.

Remove `txSize` from `ApplyTx` and add it as `nodeTxSize` to `RunNode`. Let
the `Mempool` implementation take a function of the given type as argument.
Additionally, expose `getTxSize` in the Mempool API. Otherwise, the function
in question would, rather inconveniently, have to be passed in a lot more
places.

This change helps with the hard-fork combinator: it is likely that we'll have
to add a container around transactions of the hard-fork block type when
encoding them so that we can distinguish Byron from Shelley transactions when
decoding them. Such a container will add at least one extra byte to the size
of a transaction. Accounting for this extra byte should not be done in
`ApplyTx`, but in `RunNode`.


mrBliss:
bors merge

PR #2066, opened 2020-05-12
begin integrating HardFork time into ThreadNet tests

### LINKED ISSUE #2027: Update consensus tests to avoid assumption that we know the current slot
`WrappedClock` should go.

nfrisby:
Fixes #2027. At least makes progress on #2029.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -578,7 +578,7 @@ runThreadNetwork systemTime ThreadNetworkArgs
                    -> Mempool m blk TicketNo
                    -> m ()
     forkTxProducer registry clock cfg varRNG getExtLedger mempool =
-      void $ OracularClock.onSlotChange registry clock "txProducer" $ \curSlotNo -> do
+      void $ OracularClock.forkEachSlot registry clock "txProducer" $ \curSlotNo -> do
         ledger <- atomically $ ledgerState <$> getExtLedger
         txs    <- simMonadRandom varRNG $
                     testGenTxs numCoreNodes curSlotNo cfg txGenExtra ledger
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -117,9 +117,9 @@ import qualified Test.Util.HardFork.Future as HFF
 import           Test.Util.HardFork.OracularClock (OracularClock (..))
 import qualified Test.Util.HardFork.OracularClock as OracularClock
 import           Test.Util.Random
+import           Test.Util.Slots (NumSlots (..))
 import           Test.Util.Time
 import           Test.Util.Tracer
-import           Test.Util.WrappedClock (NumSlots (..))
 
 -- | How to forge an EBB
 --
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -753,6 +753,25 @@ runThreadNetwork systemTime ThreadNetworkArgs
           -- traces the node's local events other than those from the -- ChainDB
           tracers = instrumentationTracers <> nullDebugTracers
 
+      let -- use a backoff delay of exactly one slot length (which the
+          -- 'OracularClock' always knows) for the following reasons
+          --
+          -- o It gives the node a chance to sync some blocks so that it will
+          --   eventually not need to backoff
+          --
+          -- o It maintains the invariant that the node's activities all happen "
+          --   during " a slot onset
+          --
+          -- o It avoids causing the node to miss a slot it could have
+          --   nominally lead. EG If we used a backoff of two slot durations,
+          --   then it might have synced during the first slot and then been
+          --   able to productively lead the second slot had it not still been
+          --   asleep.
+          --
+          -- o We assume a node will only backoff when it joins late and only
+          --   until it syncs enough of the net's existing common prefix.
+          backoffDelay =
+              BackoffDelay <$> OracularClock.delayUntilNextSlot clock
       btime <- hardForkBlockchainTime
                  registry
                  (contramap
@@ -764,6 +783,7 @@ runThreadNetwork systemTime ThreadNetworkArgs
                     (blockchainTimeTracer tracers))
                  (OracularClock.finiteSystemTime clock)
                  (configLedger pInfoConfig)
+                 backoffDelay
                  (ledgerState <$>
                     ChainDB.getCurrentLedger chainDB)
 
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -336,6 +336,10 @@ runThreadNetwork systemTime ThreadNetworkArgs
       -- fork the per-vertex state variables, including the mock filesystem
       (nodeInfo, readNodeInfo) <- newNodeInfo
 
+      -- a tvar containing the next slot to be recorded in
+      -- nodeEventsTipBlockNos
+      nextInstrSlotVar <- uncheckedNewTVarM joinSlot
+
       let myEdgeStatusVars =
             [ v
             | ((n1, n2), v) <- Map.toList edgeStatusVars
@@ -351,26 +355,14 @@ runThreadNetwork systemTime ThreadNetworkArgs
         vertexStatusVar
         myEdgeStatusVars
         nodeInfo
+        nextInstrSlotVar
 
-      -- Instrumentation: record the tip's block number at the onset of the
-      -- slot.
-      --
-      -- With such a short transaction (read a few TVars) we assume this runs
-      -- 1) before anything else in the slot and 2) once per slot.
-      --
-      -- Note: onSlotChange does not suffice because we need the current slot
-      -- and the 'VUp' to be read as part of the same " transaction ".
-      void $ forkLinkedThread sharedRegistry "instrumentation" $ do
-        let NodeInfo{nodeInfoEvents} = nodeInfo
-            get s = atomically $ do
-                readTVar vertexStatusVar >>= \case
-                  VUp kernel _ -> do
-                    bno <- ChainDB.getTipBlockNo (getChainDB kernel)
-                    pure (s, bno)
-                  _ -> retry
-            put (s, bno) =
-                traceWith (nodeEventsTipBlockNos nodeInfoEvents) (s, bno)
-        OracularClock.withinEachSlot clock joinSlot get put
+      forkInstrumentation
+        clock
+        sharedRegistry
+        vertexStatusVar
+        nodeInfo
+        nextInstrSlotVar
 
       return (coreNodeId, vertexStatusVar, readNodeInfo)
 
@@ -409,6 +401,7 @@ runThreadNetwork systemTime ThreadNetworkArgs
       -> VertexStatusVar m blk
       -> [EdgeStatusVar m]
       -> NodeInfo blk (StrictTVar m MockFS) (Tracer m)
+      -> StrictTVar m SlotNo
       -> m ()
     forkVertex
       mbRekeyM
@@ -419,7 +412,8 @@ runThreadNetwork systemTime ThreadNetworkArgs
       coreNodeId
       vertexStatusVar
       edgeStatusVars
-      nodeInfo =
+      nodeInfo
+      nextInstrSlotVar =
         void $ forkLinkedThread sharedRegistry label $ do
           loop 0 tniProtocolInfo NodeRestart restarts0
       where
@@ -482,11 +476,13 @@ runThreadNetwork systemTime ThreadNetworkArgs
                   error $ "unsatisfiable nodeRestarts: "
                     ++ show (coreNodeId, s')
 
-                -- TODO this " yield " is necessary to ensure the
-                -- instrumentation thread wins the race: we want to record the
-                -- current block number at the start of the slot, before the
-                -- node restarts
-                atomically $ pure ()
+                -- this synchronization prevents a race with the
+                -- instrumentation thread: we it want to record the current
+                -- block number at the start of the slot, before this vertex
+                -- restarts the node
+                atomically $ do
+                  nextSlot <- readTVar nextInstrSlotVar
+                  check $ nextSlot > s'
 
                 pure $ Just (s', pInfo', nr', rs')
 
@@ -510,6 +506,35 @@ runThreadNetwork systemTime ThreadNetworkArgs
           case again of
             Nothing                     -> pure ()
             Just (s', pInfo', nr', rs') -> loop s' pInfo' nr' rs'
+
+    -- | Instrumentation: record the tip's block number at the onset of the
+    -- slot.
+    --
+    -- With such a short transaction (read a few TVars) we assume this runs (1)
+    -- before anything else in the slot and (2) once per slot.
+    forkInstrumentation
+      :: OracularClock m
+      -> ResourceRegistry m
+      -> VertexStatusVar m blk
+      -> NodeInfo blk (StrictTVar m MockFS) (Tracer m)
+      -> StrictTVar m SlotNo
+      -> m ()
+    forkInstrumentation
+      clock
+      registry
+      vertexStatusVar
+      nodeInfo
+      nextInstrSlotVar =
+        void $ OracularClock.forkEachSlot registry clock lbl $ \s -> do
+          bno <- atomically $ readTVar vertexStatusVar >>= \case
+            VUp kernel _ -> ChainDB.getTipBlockNo (getChainDB kernel)
+            _            -> retry
+          traceWith nodeEventsTipBlockNos (s, bno)
+          atomically $ modifyTVar nextInstrSlotVar $ max (succ s)
+      where
+        NodeInfo{nodeInfoEvents}          = nodeInfo
+        NodeEvents{nodeEventsTipBlockNos} = nodeInfoEvents
+        lbl                               = "instrumentation"
 
     -- | Persistently attempt to add the given transactions to the mempool
     -- every time the ledger slot changes, even if successful!
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -52,7 +52,6 @@ import qualified Data.Typeable as Typeable
 import           Data.Void (Void)
 import           GHC.Stack
 
-import           Cardano.Slotting.EpochInfo
 import           Cardano.Slotting.Slot
 
 import           Ouroboros.Network.Block
@@ -70,7 +69,10 @@ import qualified Ouroboros.Network.TxSubmission.Outbound as TxOutbound
 
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.BlockchainTime
+import           Ouroboros.Consensus.BlockchainTime.WallClock.HardFork
+                     (hardForkBlockchainTime)
 import           Ouroboros.Consensus.Config
+import qualified Ouroboros.Consensus.Fragment.InFuture as InFuture
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Ledger.Extended
 import           Ouroboros.Consensus.Ledger.SupportsMempool
@@ -110,12 +112,14 @@ import           Test.ThreadNet.Util.NodeTopology
 import           Test.Util.FS.Sim.MockFS (MockFS)
 import qualified Test.Util.FS.Sim.MockFS as Mock
 import           Test.Util.FS.Sim.STM (simHasFS)
-import qualified Test.Util.LogicalClock as LogicalClock
+import           Test.Util.HardFork.Future (Future)
+import qualified Test.Util.HardFork.Future as HFF
+import           Test.Util.HardFork.OracularClock (OracularClock (..))
+import qualified Test.Util.HardFork.OracularClock as OracularClock
 import           Test.Util.Random
 import           Test.Util.Time
 import           Test.Util.Tracer
-import           Test.Util.WrappedClock (NumSlots (..), WrappedClock (..))
-import qualified Test.Util.WrappedClock as WrappedClock
+import           Test.Util.WrappedClock (NumSlots (..))
 
 -- | How to forge an EBB
 --
@@ -168,6 +172,7 @@ plainTestNodeInitialization pInfo = TestNodeInitialization
 --
 data ThreadNetworkArgs m blk = ThreadNetworkArgs
   { tnaForgeEbbEnv  :: Maybe (ForgeEbbEnv blk)
+  , tnaFuture       :: Future
   , tnaJoinPlan     :: NodeJoinPlan
   , tnaNodeInfo     :: CoreNodeId -> TestNodeInitialization m blk
   , tnaNumCoreNodes :: NumCoreNodes
@@ -175,17 +180,7 @@ data ThreadNetworkArgs m blk = ThreadNetworkArgs
   , tnaRNG          :: ChaChaDRG
   , tnaMkRekeyM     :: Maybe (m (RekeyM m blk))
   , tnaRestarts     :: NodeRestarts
-  , tnaSlotLength   :: SlotLength
   , tnaTopology     :: NodeTopology
-  , tnaEpochSize    :: EpochSize
-    -- ^ Epoch size
-    --
-    -- The ThreadNet tests need to know epoch boundaries in order to know when
-    -- to insert EBBs, when delegation certificates become active, etc. (This
-    -- is therefore not related to the /chunking/ of the immutable DB.)
-    --
-    -- This is temporary: once we have proper support for the hard fork
-    -- combinator, 'EpochInfo' must be /derived/ from the current ledger state.
   , tnaTxGenExtra   :: TxGenExtra blk
   }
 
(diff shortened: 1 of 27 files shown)


nfrisby:
This is a Draft PR because of the last commit only. We actually want the Byron ledger to rely on the known lower boundary, right? Do we need a protocol combinator to let us override it with `NoLowerBound` for testing, since the tests don't come close to approaching 180 epochs?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -747,11 +748,33 @@ runThreadNetwork ThreadNetworkArgs
...
+          -- o It avoids causing the node to miss a slot it could have
+          --   nominally lead. EG If we used a backoff of two slot durations,
+          --   then it might have synced during the first slot and then been
+          --   able to productively lead the second slot had it not still been
+          --   asleep.
+          --
+          -- o We assume a node will only backoff when it joins late and only
+          --   until it syncs enough of the net's existing common prefix.
+          --
+          -- TODO The current slot duration will depend on the era of the join
```
nfrisby, line 769: This TODO is the remaining work for Issue #2029 -- making the backoffdelay oracular in that it knows when the slot duration changes.

nfrisby: The TODO is resolved in the latest commits on this PR (`OracularClock` et al).

nfrisby:
Now that's it rebased onto the MockA/MockB HFC test, I'll update that test to also change the slot time using this PR's new infrastructure. That is both a desirable test, and will also help Edsko review this PR.

nfrisby:
I rebased.

  * New base commit: my variant discussed here https://github.com/input-output-hk/ouroboros-network/pull/2176#issuecomment-636129906

  * Fixup: adding missing `other-module`.

  * Fixup: Updated HFC test in `test-consensus` to compile again.

  * New commit: Extended HFC test in `test-consensus` to vary slot length across the fork. (No progress on epoch size yet -- I had overlooked that sofar.)

edsko:
I will get you a way to avoid the hack (although we cannot change the epoch size, because that would require changing the Byron ledger to take an `EpochInfo`).

nfrisby:
I've rebased again. Next:

  * The `RelativeTime` changes lead to several conflicts. They seem OK, but I'm going to double-check how I resolved them.
  * I'm going to add support for varying the `EpochSize` with the era; should be pretty straight-forward, now that Edsko advised me how to manage the CDB's chunk size.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -473,15 +472,22 @@ runThreadNetwork ThreadNetworkArgs
...
               -- onset of schedule restart slot
               Just ((s', nr'), rs') -> do
                 -- wait until the node should stop
-                tooLate <- WrappedClock.blockUntilSlot clock s'
+                tooLate <- OracularClock.blockUntilSlot clock s'
                 when tooLate $ do
                   error $ "unsatisfiable nodeRestarts: "
                     ++ show (coreNodeId, s')
+
+                -- TODO this " yield " is necessary to ensure the
```
nfrisby, line 485: This race is currently a disappointing part of the test-intra

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -672,28 +689,26 @@ runThreadNetwork ThreadNetworkArgs
...
+                    EpochSize y = epochSize0
+            let p = ledgerTipPoint $ tickedLedgerState tickedLdgSt
+
+            let needEBB = inFirstEra && At ebbSlot > pointSlot p
+            case mbForgeEbbEnv <* guard needEBB of
+              Nothing ->
+                 -- no EBB needed, forge without making one
                   forgeBlock pInfoConfig upd
                     currentBno tickedLdgSt txs prf
-
```
nfrisby, line 681: This logic was moved to the add-chunk above. Slightly reorganized, but the only difference is the new `inFirstEra` check.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -800,14 +835,13 @@ runThreadNetwork ThreadNetworkArgs
...
+        joinSlot
         registry
         -- a fingerprint for the ledger
-        ( (Origin, GenesisPoint)
-        , do
-            -- time matters, because some transaction expire
-            now <- WrappedClock.getCurrentSlot clock
-            p <- (ledgerTipPoint' (Proxy @blk) . ledgerState) <$> ChainDB.getCurrentLedger chainDB
-            pure (At now, p)
+        ( GenesisPoint
```
nfrisby, line 842: The new argument `joinSlot` supplants the `WithOrigin Slot` component of the STM transaction's output.

nfrisby:
I have tested this PR with a hacked Byron leader that involves a Byron-Byron hard fork that changes the slot length (can't change the epoch size because Byron assumes `10k`). But that's not included in this PR, because there's no simple, non-redundant, non-invasive way to do it yet.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
```diff
@@ -0,0 +1,151 @@
...
+  A test's whole timeline
+-------------------------------------------------------------------------------}
+
+-- | Every era in the test
+--
+-- INVARIANT: every number is @> 0@
+data Future =
+      FinalEra SlotLength EpochSize
+    | NextEra  SlotLength EpochSize EraSize Future
+  deriving (Eq, Show)
```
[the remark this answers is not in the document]
edsko: I'd maybe rename the constructors to `EraFinal` and `EraCons`. I find `Next` a bit confusing. `Next` compared to what?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
```diff
@@ -0,0 +1,151 @@
...
+    FinalEra _slotLength epochSize                  -> epochSize
+
+-- | Future as of a slot within an epoch
+data SlotFuture =
+      InFinalEra SlotLength
+    | InNextEra  SlotLength NumSlots Future
+      -- ^ INVARIANT: @'NumSlots' > 0@
+
+-- | Length of each slot in the whole 'Future'
+futureSlotLengths :: Future -> Stream SlotLength
```
edsko, line 76: Nice :)

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
```diff
@@ -0,0 +1,151 @@
...
+futureTimeToSlot :: Future
+                 -> NominalDiffTime
+                 -> (SlotNo, NominalDiffTime, SlotLength)
+futureTimeToSlot = \future d -> go2 0 d future
+  where
+    go acc d (InFinalEra slotLength) =
+        (SlotNo $ acc + n, timeInSlot, slotLength)
+      where
+        n = divide d slotLength
+        timeInSlot = d - multiply n slotLength
```
edsko, line 96: Suppose `d = 3.8`, `slotLength = 1.0`, then `n` will be `3`, and hence `timeInSlot = 3.8 - 3 = 0.8`. This is the time _spent_ in the lost, right, not the time _left_ in the slot. Is the comment above this function incorrect?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
```diff
@@ -0,0 +1,151 @@
...
+-- | 'EpochSize' of the first era
+futureFirstEpochSize :: Future -> EpochSize
+futureFirstEpochSize future = case future of
+    NextEra  _slotLength epochSize _eraSize _future -> epochSize
+    FinalEra _slotLength epochSize                  -> epochSize
+
+-- | Future as of a slot within an epoch
+data SlotFuture =
+      InFinalEra SlotLength
+    | InNextEra  SlotLength NumSlots Future
```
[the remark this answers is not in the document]
edsko: Might be useful to state the relation between `Future` and `SlotFuture` (invariant?) and an explanation of why we need both types.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
```diff
@@ -0,0 +1,151 @@
...
+    NextEra _slotLength _epochSize (EraSize n) _future ->
+        \(EpochNo e) -> e < n
+    FinalEra{} -> const True
+
+{-------------------------------------------------------------------------------
+  Miscellany
+-------------------------------------------------------------------------------}
+
+divide :: NominalDiffTime -> SlotLength -> Word64
+divide d slotLength = floor $ d / getSlotLength slotLength
```
edsko, line 140: Could use `Data.Fixed.div'`.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
```diff
@@ -0,0 +1,151 @@
...
+
+multiply :: Word64 -> SlotLength -> NominalDiffTime
+multiply m sl = fromIntegral m * getSlotLength sl
+
+-- | Find the non-negative difference
+safeSub :: (Num a, Ord a) => a -> a -> Maybe a
+safeSub x y = if y > x then Nothing else Just (x - y)
```
edsko, line 147: Is that `>` rather than `>=` intentional?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
```diff
@@ -0,0 +1,242 @@
...
+-- 'NumSlots'. As the system time advances, the clock maintains a notion of the
+-- current 'SlotNo'. Once all 'NumSlots' have passed, the clock is /exhausted/
+-- and most of its methods block indefinitely from then on.
+--
+-- Most notably, 'waitUntilDone' instead blocks until the the clock is
+-- exhausted; so the continuation of that call should expect other threads
+-- using the rest of this clock's methods to be stuck (and eg reap them).
+--
+-- Note: though the wallclock-slot correspondence depends on the ledger state,
+-- we have designed our ledgers so that all nodes necessarily use the same
```
edsko, line 48: Not so much the _ledgers_ right? It's more than we (you) have set up all _tests_ so that no matter which chain is chosen, the future in terms of slot lengths and epoch sizes is always the same?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
```diff
@@ -0,0 +1,242 @@
...
+                     futureTimeToSlot)
+import           Test.Util.Slots (NumSlots (..))
+import           Test.Util.Stream
+
+-- | A clock that knows the future
+--
+-- This clock's closure contains a 'BTime.SystemTime', a 'Future', and a
+-- 'NumSlots'. As the system time advances, the clock maintains a notion of the
+-- current 'SlotNo'. Once all 'NumSlots' have passed, the clock is /exhausted/
+-- and most of its methods block indefinitely from then on.
```
edsko, line 41: This is indefinite blocking behaviour important? Could we make it throw an exception instead? "Blocked indefinitely"  bugs are notoriously difficult to debug.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
```diff
@@ -0,0 +1,242 @@
...
+      --
+      -- Note: this 'BTime.systemTimeCurrent' blocks indefinitely if clock is
+      -- already exhausted.
+    , finiteSystemTime :: BTime.SystemTime m
+
+      -- | Convert 'Control.Monad.Class.MonadTime.getCurrentTime' to 'SlotNo'
+      --
+      -- INVARIANT: In @io-sim@, the @getCurrentTime@ use (by default) the
+      -- monotonic clock, and it advances /before/ threads correspondingly
+      -- wake-up from a @threadDelay@.
```
edsko, line 73: If you want to use the monotonic clock, perhaps `getMonotonicTime` would be more appropriate?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
```diff
@@ -0,0 +1,242 @@
...
+      --
+      -- So we rely on the (monotonic) clock instead of using a ticker thread
+      -- and internal state, since a calling thread and the ticker thread might
+      -- wake up simultaneously from a
+      -- 'Control.Monad.Class.MonadTimer.threadDelay' call: it's ambiguous
+      -- which thread would win the race to respectively read or increment the
+      -- internal state.
```
edsko, line 80: Not sure I understand why this matters. If the race is lost, time simply hasn't advanced yet. I guess there is an additional assumption here that that other thread that wakes up really expects to be in the new thread now?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
```diff
@@ -0,0 +1,242 @@
...
+onSlotChange :: HasCallStack
+             => ResourceRegistry m
+             -> OracularClock m
+             -> String
+             -> (SlotNo -> m ())
+             -> m (m ())
+onSlotChange reg clk = onSlotChange_ clk reg
+    -- jumping the hoop so HasCallStack is useful
+
+-- | Perform a " transaction " within the current slot
```
edsko, line 116: Id maybe avoid the word transaction here, it's (presumably) got nothing to do with UTxO transactions :) (And the signature doesn't ask for an STM transaction either).

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
```diff
@@ -0,0 +1,242 @@
...
+--
+-- The callback should be idempotent.
+--
+-- We don't export a 'getCurrentSlotSTM' function because that leads to races
+-- with 'threadDelay'. This function repeatedly attempts to execute the
+-- callback until it is able to do so with 'getCurrentSlot' returning the same
+-- provided value before and after.
```
edsko, line 123: Is this comment up to date? There is nothing executed repeatedly here is there?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
```diff
@@ -0,0 +1,242 @@
...
+          now <- readCurrentSlotSTM
+          case compare now slot of
+            LT -> retry
+            EQ -> pure False
+            GT -> pure True   -- ie " too late "
+      , delayUntilNextSlot = do
+          (_slot, inSlot, slotLength) <- getPresent
+          pure $ BTime.getSlotLength slotLength - inSlot
+      , finiteSystemTime = systemTime
+          { BTime.systemTimeCurrent = finiteSystemTimeCurrent
```
edsko, line 227: I am concerned about this. This is the `SystemTime` that in the end gets passed to the main consensus infrastructure (for the leadership check and for the in-future check in the chain DB), but unlike a "real" system clock, this one has infinite look-ahead. It would be better to limit it by a safe zone to increase the strength of the tests. That doesn't necessarily have to be done as part of this PR, but if not, we should create a ticket and do it relatively soon.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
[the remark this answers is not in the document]
nfrisby: 100% happy to switch to `Cons`.

FWIW, as in "the next" as opposed to "the final". So, "next" within the definition of the future. But, now `Cons` :thumbs-up:.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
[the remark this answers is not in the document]
nfrisby: I notice now that `SlotFuture` is only used in one function, so I've inlined it away.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
nfrisby: Great catch; the comment was wrong. But I updated the code (use call sites) to match the comment; it's tidier.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
nfrisby: Thanks, I've never internalized that module.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
nfrisby: Yes, because it's ok to return 0.

I switched the guard to `x < y`; does that help?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: The comment is about the fact that there is only one future, not multiple.

The ledgers are designed so that decisions about time are only determined by the content of the immutable prefix. The sentence also says "in the absence of a CP violation", and so all nodes share the same immutable prefix. Thus there is exactly one future. And that's _the_ future that the oracular clock correctly anticipates.

Does that clarify? Does it make sense? Let me know, and I'll try to refine the comment.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: I will try the exception and see if that's as workable.

Note that there's crucially one method that does not block indefinitely (`waitUntilDone` instead blocks **until** all other methods will start blocking) and `Network.hs` uses that one to kill the entire test. So at least `Network` is pretty robust here.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: I don't know what you mean by "look-ahead" in this context.

https://github.com/input-output-hk/ouroboros-network/blob/6507ec6eb08e75925e9c3072a11db84e226bf706/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Types.hs#L77-L90

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: Ah, yes, they're stale. I'll refine them.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: > If the race is lost, time simply hasn't advanced yet.

But time has advanced according to the `SystemTime`.

```
do
  threadDelay d
  t1 <- systemTimeCurrent finiteSystemTime
  s1 <- getCurrentSlot
  atomically (pure ()) -- "yield"
  t2 <- systemTimeCurrent finiteSystemTime     
  s2 <- getCurrentSlot
```

If we allow the race, then we necessarily have (b/c `io-sim`) that `t1 == t2` but it's possible that `s1 /= s2`, if the above thread had won the race against the ticker thread before yielding.

Since `hardforkBlockchainTime` (et al?) uses the `SystemTime` directly while instead other functions (eg the test infra) use `getCurrentSlot`, I was seeing confusing manifestations of the above incoherency. Thus I rewrote `getCurrentSlot` to also use `SystemTime`.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/Future.hs
edsko: Ah, right, your guard says when it's _not_ okay, I tend to say when it _is_ okay. Sorry :)

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
edsko: But it's not in the design of the _ledger_. It's in the design of the transactions we _submit_ to the ledger. We make sure that no matter which chain is picked, no matter how forks are resolved, the _Future_ always looks the same.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
edsko: Oh! Apologies. Yes, no, you are right. I was thinking of the `BlockchainTime`. As long as the `BlockchainTime` that we pass to `InFuture` and the ladershipcheck can't look ahead indefinitely, we're good.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: > But it's not in the design of the ledger.

I'm specifically thinking of the stability windows being defined on a per-ledger basis. That seems particularly relevant. Am I missing a distinction somewhere?

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: It was difficult to rely only on an exception and to "expect" that exception in Just The Right places.

Instead, I replaced the indefinite blocking with a small delay (arbitrarily: two slots), which is followed by an exception that is indicative of a bug: our clean-up (especially in the `SimM`) should be more prompt than that. (Tests currently pass because it currently is prompt enough. In fact, I anticipate that any non-zero delay suffices, so I'll try tightening it.)

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: I added one more variable to `Network.hs` for explicit synchronization, so that this function is not unnecessary. I removed it.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: Also removed (see nearby comment thread).

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: I moved it down to 1 picosecond, the minimal delay. Looks good.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: Somewhat severely, I replaced `OracularClock.new` with `mkOracularClock`. It now uses the given `SystemTime+threadDelay` for everything -- there is no ticker thread, it's even a pure function now.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: My priority is to use what the implementation uses, which is `defaultSystemTime`, IIUC. And that uses `getCurrentTime`, which is just the wallclock. (Which happens to be monotonic in `io-sim:SimM`.)

I'll add a TODO to update `OracularClock` et al in the test infra to handle clock jumps, so we can eventually test with those. (There's a ticket for this.)

nfrisby:
I've rebased onto master

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
edsko: I see, fair enough.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
edsko: Ah, I see, you're referring to the safe zones. But that's still not enough right. Outside of those safe zones the `Future` might look different, and I guess the tests _still_ need to make sure that, say, an A-B transition happens in the same place on every work, or else the oracular clock won't be able to perform it's miracles.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
nfrisby: I've eliminated the race -- introduced an explicit sync var.

nfrisby:
bors merge

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/HardFork/OracularClock.hs
nfrisby: We decided to revisit this later. Merging now.

Issue #1473, opened 2020-01-17
Connection Manager


coot:
Implementation of connection management component.
* [x] - connection manager spec: #2756
* [x] - connection manager
* [x] - `KeepAlive` protocol: #2175 
* [x] - `WarmPeer` protocol: #2210 
* [x] - server: #2213
* [x] - integration with peer-to-peer governor: #2214 
* [x] - on demand bidirectional mux: #1731, #2166 
* [x] - mux's trailing data: #2205
* [x] - backport dns fixes #2311
* [x] - integration with diffusion layer #2313
* [x] - integration with `cardano-node`.

avieth:
Implementation is here https://github.com/input-output-hk/ouroboros-network/pull/1455

Currently working on integration with lazy mux.

PR #2090, opened 2020-05-14
MsgKTnxBye in TxSubmission protocol

coot:
This PR adds a new message which changes.  Usage of it is will cause a Warning,
as we need to add a new `NodeToNode` version once it will be used.  
This PR changes tx-submission codec, but since we don't expect any `MsgKTnxBye`
it should be ok (making the codec depend on `NodeToNodeVersion` is not
difficult though but I think it's not worth the effort).

I only changed `NodeToNode` parts in `ouroboros-consensus`, similar changes
should be done in `NodeToClient` protocols, but I decided to leave it for
later, as we don't need it at the moment.

- Consensus's Handlers in should depend on NodeToNodeVersion
- Make BlockFetch client depend on negotiated version
- Make TxSubmission inbound and outbound side depend on negotiated version
- Added a GHC warning to MsgKTnxBye
- Moved NodeToNodeVersion and NodeToClientVersion to their own modules
- Updated cddl specification
- Updated protocol-tests
- Added MsgTnxBy tx-submission

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -882,7 +882,7 @@ data RestartCause
 -- | Fork two directed edges, one in each direction between the two vertices
 --
 forkBothEdges
-  :: (IOLike m, HasCallStack)
+  :: (IOLike m, HasNetworkProtocolVersion blk, HasCallStack)
   => ResourceRegistry m
   -> WrappedClock m
   -> Tracer m (SlotNo, MiniProtocolState, MiniProtocolExpectedException)
@@ -929,7 +929,7 @@ forkBothEdges sharedRegistry clock tr vertexStatusVars (node1, node2) = do
 -- edge via the @async@ interface rather than relying on some sort of mock
 -- socket semantics to convey the cancellation.
 directedEdge ::
-  forall m blk. IOLike m
+  forall m blk. (IOLike m, HasNetworkProtocolVersion blk)
   => Tracer m (SlotNo, MiniProtocolState, MiniProtocolExpectedException)
   -> WrappedClock m
   -> EdgeStatusVar m
@@ -973,7 +973,7 @@ directedEdge tr clock edgeStatusVar client server =
 --
 -- See 'directedEdge'.
 directedEdgeInner ::
-  forall m blk. IOLike m
+  forall m blk. (IOLike m, HasNetworkProtocolVersion blk)
   => EdgeStatusVar m
   -> (CoreNodeId, VertexStatusVar m blk)
      -- ^ client threads on this node
@@ -990,12 +990,14 @@ directedEdgeInner edgeStatusVar
 
     let miniProtocol ::
              (  LimitedApp' m NodeId blk
+             -> NodeToNodeVersion blk
              -> NodeId
              -> Channel m msg
              -> m ()
              )
             -- ^ client action to run on node1
           -> (  LimitedApp' m NodeId blk
+             -> NodeToNodeVersion blk
              -> NodeId
              -> Channel m msg
              -> m ()
@@ -1005,8 +1007,8 @@ directedEdgeInner edgeStatusVar
         miniProtocol client server = do
            (chan, dualChan) <- createConnectedChannels
            pure
-             ( client app1 (fromCoreNodeId node2) chan
-             , server app2 (fromCoreNodeId node1) dualChan
+             ( client app1 (mostRecentNodeToNodeVersion (Proxy @blk)) (fromCoreNodeId node2) chan
+             , server app2 (mostRecentNodeToNodeVersion (Proxy @blk)) (fromCoreNodeId node1) dualChan
              )
 
     -- NB only 'watcher' ever returns in these tests
@@ -1039,10 +1041,10 @@ directedEdgeInner edgeStatusVar
     wrapMPEE ::
          Exception e
       => (e -> MiniProtocolExpectedException)
-      -> (app -> peer -> chan -> m a)
-      -> (app -> peer -> chan -> m a)
-    wrapMPEE f m = \app them chan ->
-        catch (m app them chan) $ throwM . f
+      -> (app -> version -> peer -> chan -> m a)
+      -> (app -> version -> peer -> chan -> m a)
+    wrapMPEE f m = \app version them chan ->
+        catch (m app version them chan) $ throwM . f
 
     -- terminates when the vertex starts 'VFalling'
     --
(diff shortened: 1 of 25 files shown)


coot:
Related to #1473

coot:
The new version will be there when:
1. each miniprotocol will be have an api to track if it is supposed to shut down.
2. `ouroboros-consensus` will implement it
3. `ouroboros-network` will start using it (in p2p).
Until then it's an unused feature, and I prefer to merge smaller independent parts as separate PR.  Managing and reviewing large PRs is hard.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -115,6 +115,7 @@ import           Test.Util.Tracer
 import           Test.Util.WrappedClock (NumSlots (..), WrappedClock (..))
 import qualified Test.Util.WrappedClock as WrappedClock
 
+
```
mrBliss, line 118: Blank line

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
```diff
@@ -66,16 +66,17 @@ instance Exception BlockFetchServerException
...
 blockFetchServer
-    :: forall m blk.
+    :: forall m blk version.
        ( IOLike m
        , StandardHash blk
        , Typeable     blk
        )
     => Tracer m (TraceBlockFetchServerEvent blk)
     -> ChainDB m blk
+    -> version
```
mrBliss, line 76: I would just make it the real type (`NodeToNodeVersion blk`) that you're going to be passing to it. I know it now expresses that it doesn't care about the version, but the real type documents better what is to come. Same for the other handlers.

coot: I was just worried about testing code, but I think this can be done without any problem.

coot: Done

coot:
@mrBliss I addressed your comments.  Adding `HasNetworkProtocolVersion TestBlock` was enough to make the tests compile.

mrBliss:
Note to myself:

* All deployed blocks (Byron, Shelley, we don't care about test blocks) will need to support both the new and old `N.NodeToNodeVersion`.
* This new network version change is orthogonal to past changes (enable LocalStateQuery protocol, include the block size in Byron headers), even though it uses a single enumeration type. We have to version each block type separately and map it into a single network version.
* We should use a compound type for versioning (on the consensus side). This means we should add something like a `Bool` (`MsgKThxBye` supported or not) to the most recent version of each of our deployed blocks. I.e., to `ByronNodeToNodeVersion2` and `ShelleyNodeToNodeVersion1`.
* Not directly related: we should default `nodeToNodeProtocolVersion` and `nodeToClientProtocolVersion` to the newest network versions so that the test blocks use them.

@coot Is the plan to add `MsgKThxBye` to the ChainSync protocol too? If so, can we make that part of the same network version change?

@coot Can you also `s/MsgKTnxBye/MsgKThxBye/` in this PR? The way it is now is confusing.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/TestBlock.hs
```diff
@@ -267,6 +271,19 @@ data instance CodecConfig TestBlock = TestCodecConfig
...
+instance HasNetworkProtocolVersion TestBlock where
+    type NodeToNodeVersion TestBlock   = Unversioned
+    type NodeToClientVersion TestBlock = Unversioned
+    supportedNodeToNodeVersions   _ = Unversioned :| []
+    supportedNodeToClientVersions _ = Unversioned :| []
+    mostRecentNodeToNodeVersion   _ = Unversioned
+    mostRecentNodeToClientVersion _ = Unversioned
+    nodeToNodeProtocolVersion     _ Unversioned = NtN.NodeToNodeV_1
+    nodeToClientProtocolVersion   _ Unversioned = NtC.NodeToClientV_2
+
```
mrBliss, line 286: We have a default instance that you can use, so replace this with:
```haskell
instance HasNetworkProtocolVersion TestBlock where
```

ouroboros-network/src/Ouroboros/Network/Protocol/TxSubmission/Type.hs
```diff
@@ -222,6 +228,8 @@ instance Protocol (TxSubmission txid tx) where
 
   exclusionLemma_NobodyAndServerHaveAgency TokDone tok = case tok of {}
 
+{-# WARNING MsgKThxBye "MsgKTnxBye: should only be used with not yet introduced NodeToNodeV_2" #-}
```
mrBliss, line 231: Will V2 also include the other outstanding change: include the block size in the Byron headers? We should make sure it does.

coot:
> Is the plan to add MsgKThxBye to the ChainSync protocol too? If so, can we make that part of the same network version change?

No, there's no need. The `chain-sync` client has `MsgDone` which is enough. We needed to add `MsgKThxBye` to the server side of `tx-submission` since it runs along `chain-sync` client and `block-fetch` client.

> Can you also s/MsgKTnxBye/MsgKThxBye/ in this PR? The way it is now is confusing.

Sure

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/TestBlock.hs
coot: ok

ouroboros-network/src/Ouroboros/Network/Protocol/TxSubmission/Type.hs
coot: Yes, we should make both changes at once.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/TestBlock.hs
```diff
@@ -267,6 +268,8 @@ data instance CodecConfig TestBlock = TestCodecConfig
 instance BlockHasCodecConfig TestBlock where
   getCodecConfig = const TestCodecConfig
 
+instance HasNetworkProtocolVersion TestBlock where
```
mrBliss, line 271:
```suggestion
instance HasNetworkProtocolVersion TestBlock where
  -- Use defaults
```
This is a convention we have.

coot:
bors merge

PR #2094, opened 2020-05-15
Fix logic error in Shelley checkIsLeader.hasValidOCert

intricate:
Operational certificates should also be valid during the KES start period and not just
afterward.


nc6:
bors merge

Issue #2026, opened 2020-04-30
Audit use of threadDelay


edsko:
`threadDelay` on 32-bit machines can wait a maximum of roughly 35 minutes. That is not sufficient, in particular in `waitUntilSystemStart`.

edsko:
Can fix this in the `IOLike` wrapper (wait in 10 minute increments, say).

mrBliss:
Closed by #2095.

PR #2095, opened 2020-05-15
Make threadDelay duration unbounded

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/2026

Timeout duration is still bounded by Int

A ci failure was detected https://hydra.iohk.io/build/2668383/nixlog/1. It's related to threadDelay, but I'm not sure if it's related to this pr.

### DIFF
--- a/network-mux/src/Network/Mux/Timeout.hs
+++ b/network-mux/src/Network/Mux/Timeout.hs
@@ -94,7 +94,7 @@ withTimeoutSerialNative body = body MonadTimer.timeout
 --
 -- The implementation here only relies upon 'threadDelay' which has an
 -- efficient implementation on all platforms. It uses a separate monitoring
--- thread which will through an exception to terminate the action if the
+-- thread which will throw an exception to terminate the action if the
 -- timeout expires. This is why it uses the \"with\" style: because it keeps
 -- a monitoring thread over a region of execution that may use many timeouts.
 -- The cost of allocating this thread is amortised over all the timeouts used.
--- a/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Util.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Util.hs
@@ -17,10 +17,9 @@ import           Data.Time (NominalDiffTime, UTCTime, diffUTCTime)
 
 import           Cardano.Slotting.Slot
 
-import           Control.Monad.Class.MonadTimer (MonadDelay (..))
-
 import           Ouroboros.Consensus.BlockchainTime.WallClock.Types
 import qualified Ouroboros.Consensus.HardFork.History as HardFork
+import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.Time
 
 {-------------------------------------------------------------------------------
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
@@ -33,7 +33,9 @@ module Ouroboros.Consensus.Util.IOLike (
   , addTime
   , diffTime
     -- *** MonadDelay
-  , MonadDelay(..)
+  , MonadDelay -- Opaque to favor the usage of 'threadDelay'
+    -- *** MonadDelay Wrappers
+  , threadDelay
     -- *** MonadEventlog
   , MonadEventlog(..)
     -- *** Cardano prelude
@@ -50,7 +52,8 @@ import           Control.Monad.Class.MonadFork
 import           Control.Monad.Class.MonadST
 import           Control.Monad.Class.MonadThrow
 import           Control.Monad.Class.MonadTime hiding (MonadTime (..))
-import           Control.Monad.Class.MonadTimer
+import           Control.Monad.Class.MonadTimer hiding (threadDelay)
+import qualified Control.Monad.Class.MonadTimer as MonadTimer
 
 import           Ouroboros.Consensus.Util.MonadSTM.NormalForm
 import           Ouroboros.Consensus.Util.Orphans ()
@@ -91,3 +94,21 @@ class ( MonadAsync              m
       ) => IOLike m where
 
 instance IOLike IO
+
+{-------------------------------------------------------------------------------
+  MonadDelay Wrappers
+-------------------------------------------------------------------------------}
+
+-- | In some cases, 'MonadTimer.threadDelay' can't be used for delays over
+-- ~35mins, because it uses an 'Int', which can overflow (see
+-- 'diffTimeToMicrosecondsAsInt' for more info). This can be safely used to
+-- sleep for longer periods of time, so we favor its usage in all cases.
+threadDelay :: MonadDelay m => DiffTime -> m ()
+threadDelay time
+    | time > maxDelay
+    = MonadTimer.threadDelay maxDelay *> threadDelay (time - maxDelay)
+    | otherwise
+    = MonadTimer.threadDelay time
+  where
+    maxDelay :: DiffTime
+    maxDelay = fromIntegral (maxBound :: Int)
--- a/ouroboros-consensus/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
+++ b/ouroboros-consensus/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
@@ -29,6 +29,7 @@ import           Cardano.Prelude (AllowThunk (..), NoUnexpectedThunks)
 import           Cardano.Slotting.Slot (SlotNo (..))
 
 import           Control.Monad.Class.MonadTime
+import qualified Control.Monad.Class.MonadTimer as Unused
 import           Control.Monad.IOSim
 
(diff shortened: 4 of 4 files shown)


io-sim-classes/src/Control/Monad/Class/MonadTimer.hs
```diff
@@ -178,16 +178,26 @@ instance MonadTimer IO where
...
 
 diffTimeToMicrosecondsAsInt :: DiffTime -> Int
 diffTimeToMicrosecondsAsInt d =
     let usec :: Integer
-        usec = diffTimeToPicoseconds d `div` 1000000 in
+        usec = diffTimeToMicrosecondsAsInteger d in
     -- Can only represent usec times that fit within an Int, which on 32bit
     -- systems means 2^31 usec, which is only ~35 minutes.
     assert (usec <= fromIntegral (maxBound :: Int)) $
     fromIntegral usec
```
mrBliss, line 195: This function is the cause of the problem: the assert can fail.

You address it for `threadDelay`, but the problematic function is used by other functions as well.

I don't think @dcoutts (who wrote `io-sim(-classes)`) would be satisfied that we just fix one function as opposed to fixing it properly once and for all.

However, I think we, consensus, can work around the issue entirely on our side.

Plan:
* Undo the changes to `MonadTimer`
* Add a function to consensus:
  ```haskell
  -- | Document that this is for sleeping longer periods, mention the overflow and ~35 minute limit
  threadSleep :: MonadTimer m => DiffTime -> m ()
  threadSleep = <do the loop>
  ```
  (I first had a new `MonadSleep` class in mind, but a function is actually enough).
* Look for all uses of `threadDelay` in consensus and replace them with `threadSleep` where the chance exists that we might have to wait longer than 35 minutes. In particular, the place where we wait for the blockchain time to start.

kderme: Wouldn't it better to leave the changes in MonadTImer? This way all users of the class will have the fix, without needing any wrapper.

This function is only used on Timeout, other than threadDelay. I could take a look if this is also fixable, if we want to entirely remove it.

mrBliss: > Wouldn't it better to leave the changes in MonadTImer? This way all users of the class will have the fix, without needing any wrapper.

Sure, that would be ideal. But I'm not sure how easy that is.

> This function is only used on Timeout, other than threadDelay. I could take a look if this is also fixable, if we want to entirely remove it.

In that case you'll have to ask a review from @dcoutts.


First do the `threadSleep` fix, which should be merged ASAP, as people launching testnets that start in more than 35 minutes will run into the bug.

kderme: I replaced threadDelay with threadSleep in some places, hopefully I didn't miss something important.

Also I assumed your quotes and replies are reversed at your previous response?

ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
```diff
@@ -91,3 +93,17 @@ class ( MonadAsync              m
...
+
+-- | This can be used to sleep for longer periods of time. 'threadDelay' can't
+-- be used for delays over ~35mins, because it uses an 'Int', which can
+-- overflow. See 'diffTimeToMicrosecondsAsInt' for more info.
+threadSleep :: MonadDelay m => DiffTime -> m ()
+threadSleep time = do
+    threadDelay maxDelay
+    when (maxDelay /= time) $ threadSleep (time - maxDelay)
+  where
+    maxDelay = min time $ fromIntegral (maxBound :: Int)
```
mrBliss, line 109: I find it confusing that `maxDelay` is not really the maximum delay that we can wait for, both the minium of the maximum delay we can wait for and the requested wait.

What about:
```suggestion
threadSleep time
    | time > maxDelay
    = threadDelay maxDelay *> threadSleep (time - maxDelay)
    | otherwise
    = threadDelay time
  where
    maxDelay :: DiffTime
    maxDelay = fromIntegral (maxBound :: Int)
```

network-mux/src/Network/Mux/Timeout.hs
```diff
@@ -94,7 +94,7 @@ withTimeoutSerialNative body = body MonadTimer.timeout
 --
 -- The implementation here only relies upon 'threadDelay' which has an
 -- efficient implementation on all platforms. It uses a separate monitoring
--- thread which will through an exception to terminate the action if the
+-- thread which will throw an exception to terminate the action if the
```
mrBliss, line 97: This is unrelated to the PR, create a separate commit (a separate PR is not needed) for it.

ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
```diff
@@ -34,6 +34,8 @@ module Ouroboros.Consensus.Util.IOLike (
   , diffTime
     -- *** MonadDelay
   , MonadDelay(..)
```
mrBliss, line 36:
```suggestion
  , MonadDelay
```
Idea: just export `MonadDelay` but not its sole method `threadDelay`). This will prevent us from accidentally using `threadDelay` in the future.

It's still not perfect, as in some places we might import the following directly:
```haskell
import           Control.Monad.Class.MonadTimer (MonadDelay (..))
```

ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Util.hs
```diff
@@ -21,6 +21,7 @@ import           Control.Monad.Class.MonadTimer (MonadDelay (..))
```
mrBliss, line 21: Remove this import (and import `MonadDelay` from `IOLIke`, because `threadDelay` is now still in scope).

ouroboros-consensus/src/Ouroboros/Consensus/Util/IOLike.hs
```diff
@@ -91,3 +93,20 @@ class ( MonadAsync              m
...
+
+{-------------------------------------------------------------------------------
+  MonadDelay Wrappers
+-------------------------------------------------------------------------------}
+
+-- | This can be used to sleep for longer periods of time. In some cases,
+-- 'threadDelay' can't be used for delays over ~35mins, because it uses an
+-- 'Int', which can overflow. See 'diffTimeToMicrosecondsAsInt' for more info.
+threadSleep :: MonadDelay m => DiffTime -> m ()
+threadSleep time
```
mrBliss, line 105: An even better idea from @edsko: call this one `threadDelay` so that all other changes in this PR are no longer needed and that we reduce the chances of accidentally importing/using the broken `threadDelay` (typically you import `IOLike` too, so you'd have a conflict).

ouroboros-consensus/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
```diff
@@ -13,6 +13,7 @@
 module Test.Consensus.BlockchainTime.Simple (tests) where
 
 import           Control.Exception (SomeException, fromException)
+import           Control.Monad.Class.MonadTimer (MonadDelay (..))
```
mrBliss, line 16: Remove this import

kderme: I think this is needed for the instance declaration.

mrBliss: What error do you get when you remove it?

kderme:
```
test-consensus/Test/Consensus/BlockchainTime/Simple.hs:324:3: error:
    ‘threadDelay’ is not a (visible) method of class ‘MonadDelay’
    |
324 |   threadDelay d = OverrideDelay $ ReaderT $ \schedule -> do
    |   ^^^^^^^^^^^

test-consensus/Test/Consensus/BlockchainTime/Simple.hs:346:3: error:
    ‘threadDelay’ is not a (visible) method of class ‘MonadDelay’
    |
346 |   threadDelay d = OverrideDelay $ ReaderT $ \_schedule -> threadDelay d
    |   ^^^^^^^^^^^

```

If I import it qualified I get scope to the method, even though I never actually use the module qualified (qualified names are now allowed in bindings positions).

ouroboros-consensus/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
```diff
@@ -13,6 +13,7 @@
 module Test.Consensus.BlockchainTime.Simple (tests) where
 
 import           Control.Exception (SomeException, fromException)
+import qualified Control.Monad.Class.MonadTimer
```
kderme, line 16:
```suggestion
import qualified Control.Monad.Class.MonadTimer as Unused
```
Is this better?

mrBliss: I see. Ok.

But move it down so it's together with
```haskell
import           Control.Monad.IOSim
```

mrBliss:
bors merge

PR #2096, opened 2020-05-15
Remove CallStack from MuxError

karknu:


coot:
@karknu you need to restart the CI, the only thing that failed is `withTimeoutSerialNative`.

PR #2100, opened 2020-05-15
Don't require Eq for LedgerView

edsko:
This was used only for the lemmas, but it was getting in the way of the
HardFork combinator (which wants an EpochInfo in the ledger view).


edsko:
bors merge

PR #2101, opened 2020-05-15
Update ledger-specs and Shelley refactorings

mrBliss:
See the individual commit messages.


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
```diff
@@ -178,10 +178,10 @@ protocolInfoShelley genesis protVer mbCredentials =
       }
 
     initConsensusState :: State.TPraosState c
-    initConsensusState = State.empty $
+    initConsensusState = State.empty Origin $
       SL.PrtclState
          (SL.chainOCertIssue     initShelleyState)
-         Origin
+         SL.NeutralNonce
```
edsko, line 184: Let's mark this as having to be updated (hash of genesis or, if we must, final Byron hash)

mrBliss:
bors merge

PR #2102, opened 2020-05-16
Minor API improvements

edsko:
This makes two changes:

* `LedgerCfg` is now an independent type family. This means that we can
  now refer to the config of a particular ledger without necessarily
  having to give all the instances. We previously moved `LedgerState`
  out of `UpdateLedger` for the same reason, but that wasn't quite
  enough.
* Split `BasicEnvelopeValidation` off from `ValidateEnvelope`; unlike
  `ValidateEnvelope` itself, `BasicEnvelopeValidation` is not ledger
  dependent.


edsko:
bors merge

PR #2103, opened 2020-05-17
Reduce config scope creep

edsko:
As part of the hard fork work, we need to reconstruct the right `EpochInfo` at various points. We can do this in most places, but we don't quite always have sufficient context. This PR eliminates the need to do it in those places by making two changes:

* Pass `SecurityParam` instead of full `ConsensusConfig` to `rewindConsensusState`
* Make `ChainSelection` independent concept with its own config, and pass only the `ChainSelConfig` to `compareCandidates` and `preferCandidate`.

Details in the commits.


edsko:
bors merge

PR #2109, opened 2020-05-18
Add cardano-client package

MarcFontaine:
At the moment this only consists of the `subscribe` function.


cardano-client/src/Cardano/Client/API.hs
```diff
@@ -0,0 +1,78 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Cardano.Client.API
```
coot, line 5: I think it's worth to control what we export from this module. Otherwise it will be too easy to make changes to this module and not realize that they need to be propagated.

Then make sure that this module exports everything that the functions exported here needs..

coot: Also a better name for this module would be `Cardano.Client.Subscription`

cardano-client/src/Cardano/Client/API2.hs
```diff
@@ -0,0 +1,318 @@
+-- consensus uses byron types
+-- type-instantiations for codecs
+
+module Cardano.Client.API
```
coot, line 4: The file name and module name are different.

Do you think we should export both modules from `Cardano.Client` module?  It could be convenient to load everything with just a single (qualified) import.

cardano-client/src/Cardano/Client/API2.hs
```diff
@@ -0,0 +1,318 @@
...
+    , ClientStIdle (..)
+    , ClientStIntersect (..)
+    , chainSyncClientPeer
+-- import Ouroboros.Network.Protocol.Handshake.Version
+    , DictVersion (..)
+    , simpleSingletonVersions
+-- import Network.Mux.Interface
+-- import Network.Mux.Types
+--    , MuxError
+    )
```
coot, line 142: To not repeat your self twice you could import with qualification (can be shared across many imports) and then use module export.  I think it's something like this:
```
module Y where (module X)

import qualified Some.Qualified.Import as X
import qualified Some.Other.Qualified.Import as X
```

cardano-client/src/Cardano/Client/API2.hs
```diff
@@ -0,0 +1,318 @@
...
+    , Codec
+-- Network.TypedProtocol.Codec.Cbor
+--    , DeserialiseFailure
+-- Network.TypedProtocol.Driver
+    , runPeer
+    , runPipelinedPeer
+-- Network.TypedProtocol.Pipelined
+    , Nat(Zero, Succ)
+-- Ouroboros.Consensus.Ledger.Byron
+    , ByronBlock (ByronBlock, byronBlockRaw, byronBlockSlotNo, byronBlockHash)
```
coot, line 80: If we want to export `Byron`, `Shelley, `Gougen`  I think it would be better to export each from its own module, and then re-export them at the top level. It would give some structure to the client library.

coot: Maybe this is also a good idea for protocols: export all the `client` protocols from `Cardano.Client.Protocols` first and then from the top level module (`Cardano.Client`).

cardano-client/src/Cardano/Client/API2.hs
```diff
@@ -0,0 +1,318 @@
...
+    , decodeByronBlock
+    , decodeByronGenTx
+    , decodeByronHeaderHash
+    , encodeByronBlock
+    , encodeByronGenTx
+    , encodeByronHeaderHash
+-- import Ouroboros.Network.NodeToClient
+    , ConnectionId (..)
+    , NetworkConnectTracers (..)
+    , NodeToClientVersion (..)
```
coot, line 128: We need to think how to export it. It will clash with `NodeToClientVersion` in `ouroboros-consensus` which I think is more useful (but it would be good to check in various clients).

cardano-client/src/Cardano/Client/API2.hs
MarcFontaine: Ups API2.hs was to be deleted actually

cardano-client/src/Cardano/Client/API2.hs
MarcFontaine: This file is just a leftover, I had forgotten to delete.

cardano-client/src/Cardano/Client/API.hs
MarcFontaine: For now I only export the `subscribe` function
I want to figure out the exact exports later after this is tested in the applications.
This is just the first iteration.

mrBliss:
Maybe it would be useful to show how the new API improves the (lack of an) existing one? For example, show a small piece of code using it. Even better, translate one of the existing clients in `cardano-node` to use the new API (this is actually the approach I would follow to design the API).

MarcFontaine:
> Maybe it would be useful to show how the new API improves the (lack of an) existing one? For example, show a small piece of code using it. Even better, translate one of the existing clients in `cardano-node` to use the new API (this is actually the approach I would follow to design the API).

This has been done two weeks ago for `cardano-db-sync`
Just look here:
https://github.com/input-output-hk/cardano-db-sync/pull/90/files#diff-6f865c57ae290973d1569e858871fb2d

cardano-client/src/Cardano/Client/API2.hs
coot: Ok, we can add it later.

cardano-client/src/Cardano/Client/Subscription.hs
```diff
@@ -0,0 +1,80 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Cardano.Client.Subscription (
+  subscribe
```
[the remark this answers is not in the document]
MarcFontaine: Lets just do that in the next iteration, after one or more clients have been adapted to this
API it will be much clearer how it works.
I find it much easier to improve the API incrementally.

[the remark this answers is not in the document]
coot: ok

cardano-client/LICENSE
```diff
@@ -0,0 +1,177 @@
+
```
coot, line 1: You also need to copy the `NOTICE` file (`ouroboros-network/NOTICE`).

cardano-client/cardano-client.cabal
```diff
@@ -0,0 +1,29 @@
+name:                cardano-client
+version:             0.1.0.0
+synopsis:            An API for ouroboros-network
+-- description:
+license:             Apache-2.0
+license-files:       LICENSE
```
coot, line 6: Also add `NOTICE` here.

cardano-client/cardano-client.cabal
```diff
@@ -0,0 +1,29 @@
...
+copyright:           2019 Input Output (Hong Kong) Ltd.
+maintainer:
```
coot, line 8: I think it's good to put:
```
maintainer:          operations@iohk.io
```
For authors  it might be good to put `IOHK Engineering Team`.

cardano-client/src/Cardano/Client/API.hs
coot: If you'd use `cardano-haskell` you could play with other repos before merging it.

coot:
Related to #1013

PR #2110, opened 2020-05-18
Add GetUTxO to the shelley QueryLedger instance

Jimbo4350:


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -468,6 +477,8 @@ encodeShelleyQuery query = case query of
       CBOR.encodeListLen 1 <> CBOR.encodeWord8 5
     GetFilteredUTxO addrs ->
       CBOR.encodeListLen 2 <> CBOR.encodeWord8 6 <> toCBOR addrs
+    GetUTxO ->
+      CBOR.encodeListLen 1 <> CBOR.encodeWord8 7
```
mrBliss, line 481: The corresponding case is missing in `decodeShelleyQuery`

Jimbo4350:
> What's the reason for not using `GetFilteredUTxO Set.empty` on the node side?

`getFilteredUTxO` will return an empty `UTxO` if an empty set is passed.

mrBliss:
> > What's the reason for not using `GetFilteredUTxO Set.empty` on the node side?
> 
> `getFilteredUTxO` will return an empty `UTxO` if an empty set is passed.

Ah :slightly_smiling_face:

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -363,6 +366,11 @@ instance TPraosCrypto c => QueryLedger (ShelleyBlock c) where
     = Nothing
   eqQuery (GetFilteredUTxO _) _
     = Nothing
+  eqQuery GetUTxO GetUTxO
+    = Nothing
```
mrBliss, line 370: This needs to be `Just Refl`

Jimbo4350:
bors merge

PR #2113, opened 2020-05-19
Shelley: Implement GetCurrentLedgerState query

erikd:


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -523,3 +548,51 @@ decodeShelleyResult query = case query of
     GetStakeDistribution         -> fromCBOR
     GetFilteredUTxO {}           -> fromCBOR
     GetUTxO                      -> fromCBOR
+    GetCurrentLedgerState        -> unwrapCBORinCBOR (const <$> decode)
+
+
+-- Following instances required for the Serialise instances for LedgerState.
+-- At some stage they should be pushed down into the ledger-specs repo.
+deriving instance Generic SL.Coin
```
erikd, line 556: I would normally put orphan instances in a separate `Orphans` module, but that does not seem to the case in this repo.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -412,6 +422,17 @@ getPParams = SL.esPp . SL.nesEs
 getProposedPPUpdates :: SL.ShelleyState c -> SL.ProposedPPUpdates c
 getProposedPPUpdates = SL._ppups . SL._utxoState . SL.esLState . SL.nesEs
 
+-- Get the current LedgerState munis the UTxO and account portions. This
```
mrBliss, line 425:
```suggestion
-- | Get the current LedgerState minus the UTxO and account portions. This
```

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -318,6 +321,8 @@ instance TPraosCrypto c => QueryLedger (ShelleyBlock c) where
       -> Query (ShelleyBlock c) (SL.UTxO c)
     GetUTxO
       :: Query (ShelleyBlock c) (SL.UTxO c)
+    GetCurrentLedgerState
```
mrBliss, line 324: We should add a comment above this one:
```haskell
-- Only for debugging purposes, we don't guarantee binary compatibility
```

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -509,6 +533,7 @@ encodeShelleyResult query = case query of
     GetStakeDistribution         -> toCBOR
     GetFilteredUTxO {}           -> toCBOR
     GetUTxO                      -> toCBOR
+    GetCurrentLedgerState        -> wrapCBORinCBOR encode
```
mrBliss, line 536: Can't you just use `toCBOR` instead of `encode` (and `fromCBOR` instead of `decode`)? Then you won't need all these orphans below.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
mrBliss: See my comment above

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
mrBliss: Let also add a comment saying why we use CBOR-in-CBOR for this one.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
erikd: Fixed.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
erikd: Fixed.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
erikd: Have defined:
```
fromCborInCbor :: FromCBOR a => Decoder s a

toCborInCbor :: ToCBOR a => a -> Encoding
```

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
mrBliss: Why not `wrapCBORinCBOR toCBOR`? (Or does that not work?)

ouroboros-consensus-shelley/ouroboros-consensus-shelley.cabal
```diff
@@ -54,6 +54,7 @@ library
                      , cborg             >=0.2.2 && <0.3
                      , containers        >=0.5   && <0.7
                      , cryptonite        >=0.25  && <0.26
+                     , iproute
```
mrBliss, line 57: Should not be needed anymore

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -523,3 +548,23 @@ decodeShelleyResult query = case query of
...
+  tag <- CBOR.decodeWord8
+  when (tag /= 24) $ fail "expected tag 24 (CBOR-in-CBOR)"
+  fromCBOR
+
+toCborInCbor :: ToCBOR a => a -> Encoding
+toCborInCbor x =
+  mconcat
+    [ toCBOR (24 :: Word8)
+    , toCBOR x
+    ]
```
mrBliss, line 570: As commented elsewhere: I prefer we use `wrapCBORinCBOR toCBOR` instead of copy-pasting its implementation here.

edsko:
I don't understand the need and usefulness of CBOR-in-CBOR here. If it's CBOR-in-CBOR, the client becomes responsible for deserialization, rather than us providing the serializer; this means that if the client writes their own deserializer, we are even _more_ restricted in how we change stuff, it doens't really seem to help with compatiblity. And if not for compatibility, then why would clients want the raw CBOR?

edsko:
So @mrBliss explained to me that the idea here is that if we don't use CBOR-in-CBOR, the client running the local state query protocol will disconnect if the serialization format has changed and it gets a CBOR error, whereas the CBOR-in-CBOR forces the client to do the decoding and therefore take alternative action if that decoder fails. Ok, seems reasonable, but then I do wonder, if the client is asking for a UTxO, but it can't decode it, what kind of alternative action will it take? Is there something useful it can do? Because if not, then just getting the disconnected error doesn't seem a whole lot worse.

It also feels a bit ad-hoc by the way to have CBOR-in-CBOR for this one specific query. If this really is seful, I'd prefer something like

```haskell
GetSerialised :: Query a -> Query (Serialised a)
```

erikd:
There may be other queries that require CBOR-in-CBOR.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
erikd: The types of `wrapCBORinCBOR` and `toCborInCbor` are different as are their implementations!

erikd:
```
wrapCBORinCBOR :: (a -> Encoding) -> a -> Encoding
```
and
```
toCborInCbor :: ToCBOR a => a -> Encoding
```

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
erikd: It does not work.

mrBliss:
bors merge

PR #2114, opened 2020-05-19
Shelley: use annotation to compute transaction size

### LINKED ISSUE #2083: Shelley: use the annotation of the transaction to compute the size
The current implementation of `nodeTxSize` uses the semi-abstract `SL.txsize` function to return the post-serialisation size of a Shelley transaction: https://github.com/input-output-hk/ouroboros-network/blob/02a808741ef3fc8b244c4be221616e4fc92a8310/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs#L425

This semi-abstract function returns an overestimation of the size, which is fine, as we only need the size of a transaction to make sure we don't put too many transactions in a single block, such that it exceeds the max block size.

However, since Shelley transactions have recently been annotated with their serialisation, it's easy to do the right thing:
```haskell
nodeTxSize (ShelleyTx _ tx) = fromIntegral $ Lazy.length $ SL.txFullBytes tx
```

mrBliss:
Fixes #2083.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
@@ -20,6 +20,7 @@ module Ouroboros.Consensus.Shelley.Node (
 
 import           Codec.Serialise (decode, encode)
 import           Control.Monad.Reader (runReader)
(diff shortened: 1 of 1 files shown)


mrBliss:
bors merge

PR #2119, opened 2020-05-20
Update dependencies

erikd:
Specifically, use cardano-ledger-specs version after cardano-ledger
has been merged into it.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -74,10 +74,6 @@ package ouroboros-consensus-shelley
 package ouroboros-consensus-cardano
   flags: +asserts
 
-
-package cardano-ledger
-  tests: False
-
 package cardano-binary
   tests: False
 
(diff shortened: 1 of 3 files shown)


mrBliss:
CI fails because the Shelley golden tests fail. It looks like something related to hashes changed. Is that intentional?

To update a golden test: replace the list of CBOR terms with an empty one and run it again, it will print the expected output.

erikd:
Need to wait for Windoze :/

bors cancel

mrBliss:
> Need to wait for Windoze :/
> 
> bors cancel

You can bors before the Windows job finished.

mrBliss:
Documenting this for myself: the golden tests changed because the `networkMagicId` of `ConcreteCrypto` (which we use for the golden tests) changed from `Offline` to `Testnet` (https://github.com/input-output-hk/cardano-ledger-specs/pull/1465/commits/338a82114d09dac223242e6060fdaa56a0d8f84b#diff-9831fea0a1f20a6d53bdec4b976d7a30R50) and this value is included in the hash of a header.

PR #2121, opened 2020-05-20
Refactoring of OuroborosApplication

coot:
Instead of using `ConnectionId addr -> OuorborosApplication ...` it redefines the `OuroborosApplication`:

```
newtype OuroborosApplication (appType :: AppType) addr bytes m a b =
      OuroborosApplication (ConnectionId addr -> [MiniProtocol appType bytes m a b])
```

At a later stage we will hide one more argument inside `OuroborosApplicatoin`
newtype wrapper:  the signal `STM` transaction to gracefully terminate client
protocols.

This PR requires #2090 to be merged first (it targets #2090 branch at the moment, just to show adequate diff). 

- Updated ouroboros-consensus
- Updated demo-chain-sync
- Updated ouroboros-network (lib & tests)
- Updated demo-ping-pong
- OuroborosApplicaiton
- Put ConnectionId in its own module.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -468,20 +468,25 @@ mkApps kernel Tracers {..} Codecs {..} chainSyncTimeout Handlers {..} =
 initiator
   :: MiniProtocolParameters
   -> NodeToNodeVersion blk
-  -> Apps m peer blk b b b a
-  -> peer
-  -> OuroborosApplication 'InitiatorApp b m a Void
-initiator miniProtocolParameters version Apps {..} them =
+  -> Apps m (ConnectionId peer) blk b b b a
+  -> OuroborosApplication 'InitiatorApp peer b m a Void
+initiator miniProtocolParameters version Apps {..} =
     nodeToNodeProtocols
       miniProtocolParameters
-      NodeToNodeProtocols {
+      -- TODO: currently consensus is using 'ConnectionId' for its 'peer' type.
+      -- This is currently ok, as we might accept multiple connections from the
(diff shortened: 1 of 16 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -468,20 +468,25 @@ mkApps kernel Tracers {..} Codecs {..} chainSyncTimeout Handlers {..} =
...
+  -> Apps m (ConnectionId peer) blk b b b a
+  -> OuroborosApplication 'InitiatorApp peer b m a Void
+initiator miniProtocolParameters version Apps {..} =
     nodeToNodeProtocols
       miniProtocolParameters
-      NodeToNodeProtocols {
+      -- TODO: currently consensus is using 'ConnectionId' for its 'peer' type.
+      -- This is currently ok, as we might accept multiple connections from the
+      -- same ip address, however this will change when we will switch to
+      -- p2p-governor & connection-manager.  Then consenus can use peer's ip
```
karknu, line 479: We should not limit the number of connections per IP addresses. We need to be able to cope with scenarios where multiple nodes are behind the same IP address, for example in carrier-grade NAT deployments.
I assume that this is just a mistake in the comment and not a restriction added in this PR.

coot: Indeed this is not a limitation of this PR; and p2p-governor state is using IP addresses / ports to identify peers so that should be fine too.  The comment is indeed imprecise - I should say ip & port number.

coot: Fixed.

coot:
bors merge

PR #2122, opened 2020-05-20
Introduce CanForge

mrBliss:
The `NodeState` type was disconnected from most other things and `forgeBlock`
was part of `RunNode`. Combine them together in the `CanForge` type class.
This is part of an effort to move things out of `RunNode` and to split it up
in the right abstractions.

Moreover, we want to be able to define `forgeBlock` generically for the hard
fork combinator. This change will make that easier.

* Add the new `CanForge` type class.

* Make the `NodeState` top-level type family an associated type family of
  `CanForge` and rename it to `ForgeState`.

* Move `nodeForgeBlock` from `RunNode` and it as `forgeBlock` to `CanForge`.

* Rename all mentions of `*NodeState` and `*State` to `*ForgeState`.

* Remove `forgeExt` from `RunMockBlock` and turn it into a data type that
  `forgeSimple` takes as an argument.


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -31,19 +47,152 @@ import           Ouroboros.Consensus.Shelley.Ledger.Config
...
+                     , TPraosCrypto c
+                     , KES.Signable (KES c) toSign
+                     )
+                  => Update m (TPraosForgeState c)
+                  -> IsLeader (TPraos c)
+                  -> SL.KESPeriod
+                  -> (TPraosToSign c -> toSign)
+                  -> m (TPraosFields c toSign)
+forgeTPraosFields updateForgeState TPraosProof{..} kesPeriod mkToSign = do
+    hotKESKey <- evolveKESKeyIfNecessary updateForgeState (SL.KESPeriod kesEvolution)
```
mrBliss, line 155: Do the key evolution in `forgeShelleyBlock` and move `forgeTPraosFields` back to the `Protocol` module.

mrBliss:
bors merge

PR #2129, opened 2020-05-21
Shelley: move tests to a separate lib

mrBliss:
To enable testing the hard fork with Shelley as one of the eras, we want to be
able to reuse the test infrastructure and generators for Shelley from another
package. We don't want to put the hard fork tests in the Shelley test suite,
as that would unnecessary couple them (what about Byron then?).

To avoid circular dependencies, we put things in the following locations:

* `ouroboros-consensus-shelley/src`: the Shelley implementation (untouched).
* `ouroboros-consensus-shelley-test/src`: the generators and infrastructure
  for Shelley that we want to reuse from other packages.
* `ouroboros-consensus-shelley-test/test`: the actual Shelley tests, that use
  both the above packages.

This approach allows us to avoid having to share folders between different
packages.

Besides moving the code, as much test infrastructure was split off from the
actual tests. For example, some things got moved from
`Test.ThreadNet.RealTPraos` (now `-test/test`) to
`Test.ThreadNet.TxGen.Shelley` (now `-test/src`).

The generators and example values used by the golden tests (both `-test/src`)
also split off from the roundtrip and golden tests (both `-test/test`).


mrBliss:
@nfrisby @nc6 Sorry for the potential merge conflicts :flushed:

mrBliss:
bors merge

PR #2132, opened 2020-05-21
Update windows.yml (allow ghc downgrade)

angerman:


.github/workflows/windows.yml
```diff
@@ -31,14 +31,15 @@ jobs:
       - name: Configure Chocolatey
         run: choco source add -n mistuke -s https://www.myget.org/F/mistuke/api/v2
       - name: Install msys2
-        run: choco install -y msys2
+        run: choco install -y msys2
```
mrBliss, line 34: Can you undo this change?

angerman: 👍

coot:
I think I found a fix in #2144. Thanks @angerman for your help.

PR #2133, opened 2020-05-21
Add cardano-client to the required CI builds

MarcFontaine:
By default CI picks up all test-suites as required build.
There are no reasonable tests for the cardano-client library at the moment.
This add a dummy test-suite to still make it a required build.

### DIFF
--- a/cardano-client/cardano-client.cabal
+++ b/cardano-client/cardano-client.cabal
@@ -30,3 +30,13 @@ library
 
   ghc-options:         -Wall
                        -Wno-unticked-promoted-constructors
+
+test-suite test-cardano-client
+  type:                exitcode-stdio-1.0
+  hs-source-dirs:      test
+  main-is:             Main.hs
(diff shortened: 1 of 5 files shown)


coot:
@MarcFontaine instead of adding a dummy test suite (I don't think we will have one soon), I am quite sure there's a way modify `release.nix`.  Maybe by adding a job to `jobs`. You can ask deveops for help.

MarcFontaine:
That is also an option. I looked at `release.nix` there is a line where I could add it.
I wanted to avoid to make `cardano-client` a special case in CI and
make nix more complicated.
Next time the nix-files need to be changed the extra requirement may get lost.

MarcFontaine:
https://github.com/input-output-hk/ouroboros-network/pull/2136

MarcFontaine:
fixed in https://github.com/input-output-hk/ouroboros-network/pull/2155

PR #2135, opened 2020-05-22
threadDelay & registerDelay - unbounded delays

coot:
This PR updates both `threadDelay` and `registerDelay`.  Both accept
`DiffTime`, and they should be safe.

`threadDelay` is easy, since it's a synchronous operation;
`registerDelay` is more tricky as the returned `TVar` must be updated
asynchronously. For delays smaller than `maxBound :: Int` we reuse the original
`STM` implenetation, for larger delays we use the default 'MonadTimer`
implementation (which starts a thread to update the `TVar`).

- registerDelay - accept unbounded delays
- threadDaley - accept unbounded delays

### DIFF
--- a/io-sim-classes/src/Control/Monad/Class/MonadTimer.hs
+++ b/io-sim-classes/src/Control/Monad/Class/MonadTimer.hs
@@ -95,14 +95,21 @@ class (MonadSTM m, MonadDelay m) => MonadTimer m where
   registerDelay :: DiffTime -> m (TVar m Bool)
 
   default registerDelay :: MonadFork m => DiffTime -> m (TVar m Bool)
-  registerDelay d = do
+  registerDelay = defaultRegisterDelay
+
+  timeout :: DiffTime -> m a -> m (Maybe a)
+
+
+defaultRegisterDelay :: ( MonadTimer m
+                        , MonadFork  m
+                        )
+                     => DiffTime
+                     -> m (TVar m Bool)
+defaultRegisterDelay d = do
     v <- atomically $ newTVar False
     t <- newTimeout d
     _ <- fork $ atomically (awaitTimeout t >>= writeTVar v)
     return v
-
-  timeout :: DiffTime -> m a -> m (Maybe a)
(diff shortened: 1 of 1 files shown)


coot:
@mrBliss @kderme this is related to #2095

io-sim-classes/src/Control/Monad/Class/MonadTimer.hs
```diff
@@ -174,7 +192,16 @@ instance MonadTimer IO where
       when (not fired) $ STM.writeTVar cancelvar True
 #endif
 
-  registerDelay = STM.registerDelay . diffTimeToMicrosecondsAsInt
+  -- | For delays less (or equal) than @maxBound :: Int@ this is exactly the same as
+  -- 'STM.registerDaley'; for larger delays it will start a monitoring thread
```
mrBliss, line 196:
```suggestion
  -- 'STM.registerDelay'; for larger delays it will start a monitoring thread
```

mrBliss: You used "daley" instead of "delay" in commit messages too.

coot:
@mrBliss there's a CI failure in Byron test suite: https://hydra.iohk.io/build/2717445/nixlog/3
Could this be caused by this PR?

mrBliss:
> @mrBliss there's a CI failure in Byron test suite: https://hydra.iohk.io/build/2717445/nixlog/3
> Could this be caused by this PR?

If `threadDelay` no longer works correctly, then yes. But the implementations look fine and no other tests fail :thinking:. Can you try to reproduce it without your changes?

coot:
Locally, it run fine:
```
    topology prevents timely dlg cert tx propagation:                                      OK (16.20s)
```

mrBliss:
> We should also remove the duplicate implementation in the consensus code.

Indeed. @kderme Can you take care of this? It'll boil down to reverting #2095.

coot:
bors merge

PR #2146, opened 2020-05-25
Byron: move tests to a separate lib

mrBliss:
Do the same for Byron as for Shelley in #2129.

The ThreadNet infrastructure is still in the testsuite of
`ouroboros-consensus-byron-test`. This should be split up into reusable
infrastructure and the actual property tests in a future PR with some help
from @nfrisby.


mrBliss:
bors merge

mrBliss:
bors merge

PR #2150, opened 2020-05-26
Reorganize HFC a bit: make type family newtype wrappers HFC independent

edsko:
at the request of @mrBliss who was nonetheless too lazy to do it himself :roll_eyes:


ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator.hs
```diff
@@ -6,6 +6,7 @@ module Ouroboros.Consensus.HardFork.Combinator (
   ) where
 
 -- Defines 'SingleEraBlock' and 'CanHardFork'
+-- Defines 'SingleEraInfo' and 'LedgerEraInfo'
```
mrBliss, line 9: Join the two enumerations? Not that important

edsko: Yeah, I thought about it, but they're kind of of a different nature, so I left it.

edsko:
bors merge

PR #2151, opened 2020-05-26
Pass only BlockConfig to nodeStartTime/NetworkMagic/ProtocolMagicId

mrBliss:
None of the implementations ever needed more than the `BlockConfig`. So
replace `TopLevelConfig` with `BlockConfig` in those methods.

This makes it easier to define those methods for a non-degenerate hard-fork,
where we can't always project out an era-specific `TopLevelConfig` from the
hard-fork one, because of the partial configs. We can trivially project out
the era-specific `BlockConfig`s, though.


mrBliss:
bors merge

PR #2159, opened 2020-05-27
Exports more Types

MarcFontaine:
Export (almost all) types and helper functions that are needed to construct the argument
of the `subscribe` function.

### DIFF
--- a/cardano-client/src/Cardano/Client/Subscription.hs
+++ b/cardano-client/src/Cardano/Client/Subscription.hs
@@ -1,38 +1,54 @@
-{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DataKinds           #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeFamilies        #-}
 
 module Cardano.Client.Subscription (
-  subscribe
+    subscribe
+  , MuxMode (..)
+  , ClientCodecs
+  , ConnectionId
+  , LocalAddress
+  , NodeToClientProtocols (..)
+  , NodeToClientVersion
+  , MuxPeer (..)
+  , MuxTrace
+  , RunMiniProtocol (..)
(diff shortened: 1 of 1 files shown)


cardano-client/src/Cardano/Client/Subscription.hs
```diff
@@ -3,18 +3,32 @@
 {-# LANGUAGE TypeFamilies #-}
 
 module Cardano.Client.Subscription (
-  subscribe
+     subscribe
```
mrBliss, line 6:
```suggestion
    subscribe
```

cardano-client/src/Cardano/Client/Subscription.hs
```diff
@@ -30,8 +44,9 @@ import qualified Ouroboros.Network.Snocket as Snocket
 
 import qualified Ouroboros.Network.NodeToClient (NodeToClientVersion)
 import           Control.Monad.Class.MonadST (MonadST)
-import           Prelude
 import           Ouroboros.Consensus.Config.SupportsNode (getCodecConfig, getNetworkMagic)
+import           Ouroboros.Network.Mux (MuxPeer (..), RunMiniProtocol (..))
+import           Network.Mux.Trace (MuxTrace, WithMuxBearer)
```
mrBliss, line 49: Can you please sort the entire import list in this file? This is currently a horrible mix.

Suggestion:

* First all imports from base and standard libs (`Data.Proxy`, etc.)
* Imports from `io-sim-classes` (`Control.Monad.Class.MonadST`)
* Imports from `network-mux` (`Network.Mux.Trace`)
* Imports from `ouroboros-network`
* Imports from `ouroboros-consensus` 

Each of the above sections should be sorted alphabetically and separated by a blank line. Using `stylish-haskell` would also be nice.

Just my 2c :slightly_smiling_face:

MarcFontaine: Thx for the review. I improved the imports. Could you please look at it again.

PR #2167, opened 2020-05-29
Add custom labels to `prop_general`

edsko:
For the HFC tests, we take advantage of this to show how many A blocks and how
many B blocks we have, though I imagine it can come in handy in lots of other
places also.

I had accidentally left in a change to `prop_general` where we show the full chains always; this is not very useful for most tests.


edsko:
> No separate commits?

Some people get confused by them :roll_eyes:

edsko:
bors merge

PR #2169, opened 2020-05-29
Introduce `CannotLead`

edsko:
Before when a node had an outdated (or conversely not-yet-active) it had no way to indicate that it was a leader but that it couldn't lead before the key was wrong. We now distinguish these two cases, although we don't yet take advantage of this. We should: both PBFT and TPraos should _check_ if the key matches the delegation state, and if not, say that they cannot lead, rather than produce an invalid block.


ouroboros-consensus/src/Ouroboros/Consensus/Protocol/LeaderSchedule.hs
```diff
@@ -87,23 +87,28 @@ data instance ConsensusConfig (WithLeaderSchedule p) = WLSConfig
...
   type CanBeLeader    (WithLeaderSchedule p) = ()
+  type CannotLead     (WithLeaderSchedule p) = ScheduleMissingSlot
 
   protocolSecurityParam = protocolSecurityParam . wlsConfigP
   chainSelConfig        = chainSelConfig        . wlsConfigP
 
   checkIsLeader WLSConfig{..} () (Ticked slot _) _ = return $
     case Map.lookup slot $ getLeaderSchedule wlsConfigSchedule of
-        Nothing                           -> Nothing
+        Nothing -> CannotLead $ ScheduleMissingSlot slot
```
edsko, line 108: Maybe use `error` instead and set `CannotLead` to `Void`.

edsko:
bors merge

PR #2175, opened 2020-05-29
KeepAlive protocol

coot:
Keep alive protocol, designed for

* sending keep alive messages
* measuring RTTs

### DIFF
--- a/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Direct.hs
+++ b/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Direct.hs
@@ -1,0 +1,19 @@
+{-# LANGUAGE NamedFieldPuns #-}
+
+module Ouroboros.Network.Protocol.KeepAlive.Direct where
+
+import Ouroboros.Network.Protocol.KeepAlive.Client
+import Ouroboros.Network.Protocol.KeepAlive.Server
+
+direct :: Monad m
+       => KeepAliveServer m a
+       -> KeepAliveClient m b
+       -> m (a, b)
+direct KeepAliveServer { recvMsgDone }
+       (SendMsgDone mdone) =
+    (,) <$> recvMsgDone <*> mdone
+direct KeepAliveServer { recvMsgKeepAlive }
+       (SendMsgKeepAlive mclient) = do
+    server <- recvMsgKeepAlive
+    client <- mclient
+    direct server client
--- a/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Examples.hs
+++ b/ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Examples.hs
@@ -1,0 +1,41 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Ouroboros.Network.Protocol.KeepAlive.Examples where
+
+import           Ouroboros.Network.Protocol.KeepAlive.Server
+import           Ouroboros.Network.Protocol.KeepAlive.Client
+
+
+-- | A client which applies a function whenever it receives
+-- 'MsgKeepAliveResponse' and returns the result.
+--
+keepAliveClientApply :: forall acc m. Monad m
+                    => (acc -> acc)
+                    -> acc
+                    -> Int
+                    -- ^ count of number of requests
+                    -> KeepAliveClient m acc
+keepAliveClientApply f = go
+  where
+    go :: acc -> Int -> KeepAliveClient m acc
+    go acc n
+      | n <= 0
+      = SendMsgDone (pure acc)
+
+      | otherwise
+      = SendMsgKeepAlive $
+          pure $ go (f acc) (pred n)
+
+
+-- | A server which counts number received of 'MsgKeepAlive'.
+--
+keepAliveServerCount :: forall m. Applicative m
+                     => KeepAliveServer m Int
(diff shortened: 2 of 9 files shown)


ouroboros-network/src/Ouroboros/Network/Protocol/KeepAlive/Type.hs
```diff
@@ -0,0 +1,90 @@
...
+
+-- | The type of the keep alive protocol.
+--
+-- The keep alive protocol is used for
+--
+-- * sending keep alive messages
+-- * making round trip measurements
+--
+-- Each side will run its own version of the keep alive protocol.  It should be
+-- configured so that each side keep away its carrier grade nat from shuting
```
njd42, line 16: It is not just carrier grade nat (CGN) that has this issue - perhaps a better comment might be:

It should be configured so that any intermediate state (such as in customer premise equipment or in a carrier grade NAT) is kept alive.  This has to be a per-node configuration element as this is about the properties of that nodes network connectivity,

ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Direct.hs
```diff
@@ -0,0 +1,19 @@
+{-# LANGUAGE NamedFieldPuns #-}
+
+module Ouroboros.Network.Protocol.KeepAlive.Direct where
+
+import Ouroboros.Network.Protocol.KeepAlive.Client
+import Ouroboros.Network.Protocol.KeepAlive.Server
+
+direct :: Monad m
```
njd42, line 8: More for my edification, why did you choose to call this function `direct`?

ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Examples.hs
```diff
@@ -0,0 +1,43 @@
...
+
+
+-- | A client which applies a function whenever it receives
+-- 'MsgKeepAliveResponse' and returns the result.
+--
+keepAliveClientApply :: forall acc m. Monad m
+                    => (acc -> acc)
+                    -> acc
+                    -> Int
+                    -- ^ number of rounds
```
njd42, line 18: 'number of rounds" may be missinterpreted, given the crypto context of this project. 
 Its this really "count of number of requests"?

ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Examples.hs
```diff
@@ -0,0 +1,43 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Ouroboros.Network.Protocol.KeepAlive.Examples where
+
+import           Data.List (cycle)
```
njd42, line 5: Is this an unused input?

coot: It is.  Hmm, I think I forgot to add `-Wall` (it should fail in CI).

ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Direct.hs
coot: Because it directly connects the client ADT and server ADT without passing through Peer (where a similliar function is called `connect`, so the name is taken), or using any kind of channel.  This is a convention shared between all mini-protocol tests.

ouroboros-network/src/Ouroboros/Network/Protocol/KeepAlive/Type.hs
coot: I'll update the comment, thanks.

coot:
> Is it you intention to merge this sooner rather than later?

I prefer sooner, but it will take us some time before it is actually used by the node. Changing a protocol or adding a new one requires a protocol version, we had only one protocol change so far and this time it should be even easier than last time. I am not too worried about it.

ouroboros-network/protocol-tests/Ouroboros/Network/Protocol/KeepAlive/Examples.hs
coot: Thanks for spotting this, it turns out there are quite a few warnings with `-Wall` in `ouroboros-protocols-tests`.  It deserves its own PR.

ouroboros-network/src/Ouroboros/Network/Protocol/KeepAlive/Codec.hs
```diff
@@ -0,0 +1,55 @@
...
+import qualified Codec.CBOR.Encoding as CBOR (Encoding, encodeWord)
+import qualified Codec.CBOR.Read     as CBOR
+import qualified Codec.CBOR.Decoding as CBOR (Decoder, decodeWord)
+
+import           Network.TypedProtocol.Core
+
+import           Ouroboros.Network.Codec
+import           Ouroboros.Network.Protocol.KeepAlive.Type
+
+
```
karknu, line 25: You should define `ProtocolTimeLimits` and `ProtocolSizeLimits` for the protocol.

coot: I left this, together with an implementation of a client / server for another RP.

coot:
bors merge

Issue #2172, opened 2020-05-29
Byron: check if key is okay in checkIsLeader


edsko:
See #2169.

Marking this as high priority because logging information about invalid blocks being produced is terrifying, whereas messages about keys being out of date just means PEBKAC. (A report about invalid blocks came in only today, so it's important we address this soon.)

In addition, the construction of invalid blocks is causing trouble for @nfrisby 's tests; doing this will hopefully avoid the hack in https://github.com/input-output-hk/ouroboros-network/pull/2048 -- indeed, we should be able to avoid the whole "should we do the production of this EBB?" thing.

edsko:
Closed in https://github.com/input-output-hk/ouroboros-network/pull/2176 .

Issue #2173, opened 2020-05-29
Threshold check in PBFT


edsko:
After https://github.com/input-output-hk/ouroboros-network/pull/2169 , nodes can now indicate that they are a leader for a slot, but cannot lead due to some problem. For PBFT, this problem could be that the key does not match the delegation state (https://github.com/input-output-hk/ouroboros-network/issues/2172), but it could also mean that the block would exceed the PBFT window threshold.

Right now we don't check for this in PBFT. Initially I thought that doing so would be wrong, because it means that I could force you to give up your slot simply by producing few blocks myself. However, if a node produces a block ignoring this, that block would anyway promptly be discarded (by its own node!) as invalid, so there is no reason _not_ to do the check.

The absence of the check means that the consensus tests need to make sure that the threshold does not get exceeded, or else they end up with invalid blocks. Introducing this check would allow us to lift that restriction. It would also give a better log message should this ever happen in real life, although in reality we don't expect this to ever happen, which is why I've marked this purely as a `testing` ticket.

nfrisby:
To clarify: this ticket is a follow-on to Issue 2172 that expands the task therein from just checking the key is OK (up-to-date delegation map, etc) to also checking that the PBFT window threshold is satisfied, right?

edsko:
Correct.

edsko:
> because it means that I could force you to give up your slot simply by producing few blocks myself.

actually, this is doubly incorrect. Not only would producing a block be pointless in this case anyway (as I mention in the last paragraph in the ticket description), but also, the threshold is computed over the window size, _independent from the slots filled by other nodes_.

edsko:
Closed in https://github.com/input-output-hk/ouroboros-network/pull/2176 .

PR #2176, opened 2020-05-29
Add PBFT CannotLead checks

nc6:
- Check that we are the current delegate of the genesis key
- Check that more blocks than the singing threshold have not been issued.

Addresses #2172

### DIFF
--- a/ouroboros-consensus-byron-test/test/Test/ThreadNet/DualPBFT.hs
+++ b/ouroboros-consensus-byron-test/test/Test/ThreadNet/DualPBFT.hs
@@ -48,6 +48,7 @@ import           Ouroboros.Consensus.Node.ProtocolInfo
 import           Ouroboros.Consensus.NodeId
 import           Ouroboros.Consensus.Protocol.LeaderSchedule
 import           Ouroboros.Consensus.Protocol.PBFT
+import           Ouroboros.Consensus.TypeFamilyWrappers
 
 import           Ouroboros.Consensus.Byron.Ledger
 import           Ouroboros.Consensus.Byron.Ledger.Conversions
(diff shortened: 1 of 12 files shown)


edsko:
Ah! Also addresses https://github.com/input-output-hk/ouroboros-network/issues/2173 .

edsko:
@nfrisby could you confirm that this means that your tests will work without the invalid blocks check? That would be a great indication that this is doing the right thing.

nfrisby:
> 
> 
> @nfrisby could you confirm that this means that your tests will work without the invalid blocks check? That would be a great indication that this is doing the right thing.

OK. I'll develop a PR on top of this one to simplify the test suite accordingly.

Might not want to block this on that -- not sure how the schedule will proceed.

nfrisby:
Bah. CI failure demonstrates that the answer is "No". I'll increase my priority on it. The particular test that failed is a corner case.

nfrisby:
FYI https://github.com/input-output-hk/ouroboros-network/tree/nfrisby/issue-2172

I've pushed a commit to that branch that extends this PR. It updates the test infrastructure as necessary.

Please cherry-pick it onto this one, if you like it.

The test infrastructure has so far been expecting invalid blocks to be caught by the ChainDB. Since some tests knowingly induce invalid blocks, those tests had been validating such block rejections.

As of this PR, those invalid blocks are never even forged. But we still want to validate the failure to lead. (The fact that we no longer could do so is why the CI failed.) My commit therefore adjusts the test infrastructure to validate `CannotLead`s instead of `BlockRejection`s.

edsko:
@nfrisby that sounds great. I will take a look at this PR now, as well as the commit you refer to, and then hopefully merge.

edsko:
bors merge

edsko:
bors merge

edsko:
bors merge

PR #2181, opened 2020-06-01
Right inverse of diffTimeToMicroseconds.

### LINKED ISSUE #2179: `maxDelay` is incorrectly scaled  - should be 10^6 times smaller
I think that `maxDelay`  (https://github.com/input-output-hk/ouroboros-network/blob/be867b42cb75e3c1a837bdc3647f27021f7c8169/io-sim-classes/src/Control/Monad/Class/MonadTimer.hs#L130) is dimensioned incorrectly.

Isn't `fromIntegral $ maxBound :: Int` as a `DiffTime` that number of seconds - but the time units in IO are microseconds - this is a 10^6 too large!?

This would cause issues on 32bit machines - things would see negative delays, wake up early etc....

coot:
Provide right inverse of `diffTimeToMicroseconds` and check its
properties.

Fixes #2179

### DIFF
--- a/io-sim-classes/src/Control/Monad/Class/MonadTimer.hs
+++ b/io-sim-classes/src/Control/Monad/Class/MonadTimer.hs
@@ -1,13 +1,17 @@
-{-# LANGUAGE CPP               #-}
-{-# LANGUAGE DefaultSignatures #-}
-{-# LANGUAGE FlexibleContexts  #-}
-{-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE TypeFamilies      #-}
+{-# LANGUAGE CPP                #-}
+{-# LANGUAGE DefaultSignatures  #-}
+{-# LANGUAGE FlexibleContexts   #-}
+{-# LANGUAGE FlexibleInstances  #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE TypeFamilies       #-}
 
 module Control.Monad.Class.MonadTimer (
     MonadDelay(..)
   , MonadTimer(..)
   , TimeoutState(..)
+
+  , diffTimeToMicrosecondsAsInt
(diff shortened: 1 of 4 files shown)


io-sim-classes/src/Control/Monad/Class/MonadTimer.hs
```diff
@@ -201,20 +207,24 @@ instance MonadTimer IO where
       | otherwise =
         defaultRegisterDelay d
```
njd42, line 208: I accept that this PR fixes the immediate problem, but the underlying problem still exists. 
On a 32 bit machine with a timeout more than ~35min in the future `defaultRegisterDelay` would be invoked which, in turn, makes use of `newTimeout`. This calls `GHC.registerTimeout` where the intrinsic 32 bit constraint rears its ugly head again.

The underlying issue is still there!

coot: I didn't aim to fix it here; it is requires a bit more work.

coot:
bors merge

io-sim-classes/src/Control/Monad/Class/MonadTimer.hs
```diff
@@ -121,13 +125,15 @@ defaultRegisterDelay d = do
...
+      go :: DiffTime -> IO ()
       go d | d > maxDelay = do
         IO.threadDelay (diffTimeToMicrosecondsAsInt d)
         go (d - maxDelay)
       go d = do
         IO.threadDelay (diffTimeToMicrosecondsAsInt d)
 
-      maxDelay = fromIntegral (maxBound :: Int)
+      maxDelay :: DiffTime
+      maxDelay = microsecondsAsIntToDiffTime maxBound
```
edsko, line 136: :+1:

PR #2182, opened 2020-06-01
Pass CodecConfig to all en/decoders in RunNode

mrBliss:
Previously, it was a mix of `TopLevelConfig`, `Proxy`, `CodecConfig`, and
nothing. Now, we use `CodecConfig` for all of them.


mrBliss:
bors merge

PR #2185, opened 2020-06-01
Split off CodecConfig from ConfigSupportsNode

mrBliss:
In the future, we'll need `CodecConfig` and `getCodecConfig` in some contexts
where we don't care about the other methods of `ConfigSupportsNode`, e.g., in
the storage layer.


mrBliss:
bors merge

mrBliss:
bors merge

PR #2187, opened 2020-06-01
Decouple BinaryInfo from the block encoder

### LINKED ISSUE #2145: BinaryInfo should not be a functor
At the moment we have

```haskell
data BinaryInfo blob = BinaryInfo
  { binaryBlob   :: !blob
  , headerOffset :: !Word16
    -- ^ The offset within the 'binaryBlob' at which the header starts.
  , headerSize   :: !Word16
    -- ^ How many bytes the header is long. Extracting the 'headerSize' bytes
    -- from 'binaryBlob' starting from 'headerOffset' should yield the header.

    -- In the future, i.e. Shelley, we might want to extend this to include a
    -- field to tell where the transaction body ends and where the transaction
    -- witnesses begin so we can only extract the transaction body.
  }
```

but that feels wrong: there is no way that a `Functor` instance could preserve any kind of internal invariants of this type, and even if this isn't true in general, I suspect even for _specific_ use cases of `fmap` those invariants are very unlikely to be preserved.

mrBliss:
Closes #2145.

Split

    nodeEncodeBlockWithInfo  :: CodecConfig blk -> blk -> BinaryInfo Encoding

Up into

    nodeGetBinaryBlockInfo :: blk -> BinaryBlockInfo
    nodeEncodeBlock        :: CodecConfig blk -> blk -> Encoding

where `BinaryBlockInfo` is the former `BinaryInfo` without its `blob`
argument. This also means `Binary(Block)Info` is no longer a functor, closing
issue #2145.

The reason both were coupled is that we thought that the information would be
computed during encoding, making it more efficient to do both at the same
time. However, for all real ledgers, we use annotations, making it cheap to
compute the `BinaryBlockInfo` without any real encoding work to be done.

Splitting them up will make later refactorings of the encoders/decoders
easier.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/HasBlockBody.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/HasBlockBody.hs
@@ -16,7 +16,8 @@ module Ouroboros.Consensus.HardFork.Combinator.HasBlockBody (
     -- * HFC support
   , SerialiseHeaderBody(..)
   , Body(..)
-  , encodeHardForkBlockWithInfo
+  , encodeHardForkBlock
+  , hardForkBlockBinaryBlockInfo
   , toSerialiseOneHeader
   , toSerialiseOneBody
   , fromSerialiseOneHeader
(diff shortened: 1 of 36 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/HasBlockBody.hs
```diff
@@ -164,19 +165,31 @@ instance ( CanHardFork xs
 -- > deriving via SerialiseOne Header SomeEras
 -- >          instance Serialise (Header SomeBlock)
 --
--- TODO: The way this is currently defined is not very efficient; this is not
--- a big deal right now as this is currently only used for tests.
-encodeHardForkBlockWithInfo
+encodeHardForkBlock
```
mrBliss, line 168: No longer needed

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Dual.hs
```diff
@@ -587,23 +588,15 @@ agreeOnError f (ma, mb) =
 -- | The binary info just refers to the main block
```
mrBliss, line 588: TODO move to `dualBinaryBlockInfo`

edsko:
bors merge

PR #2189, opened 2020-06-01
Define class for projection/injection in Unary

edsko:
Together with @mrBliss 's refactoring of the serialization, with a bit of luck this will make the `Degenerate` instance a lot cleaner.

The one spanner in the works is `projLedgerView`, which doesn't have an inverse. I want to come back to that once more before merging.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Unary.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Unary.hs
@@ -80,6 +80,7 @@ import           Ouroboros.Consensus.Ledger.SupportsMempool
 import           Ouroboros.Consensus.Node.ProtocolInfo
 import           Ouroboros.Consensus.Protocol.Abstract
 import           Ouroboros.Consensus.TypeFamilyWrappers
+import           Ouroboros.Consensus.Util.SOP
 
 import           Ouroboros.Consensus.Storage.ChainDB.Init (InitChainDB)
 import qualified Ouroboros.Consensus.Storage.ChainDB.Init as InitChainDB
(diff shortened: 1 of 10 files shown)


edsko:
Will do ledger view separately.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Unary.hs
```diff
@@ -96,316 +65,442 @@ import           Ouroboros.Consensus.HardFork.Combinator.Mempool
...
-      HardForkLedgerState
-    . HardForkState
-    . Telescope.TZ
-    . State.Current History.initBound
+{-------------------------------------------------------------------------------
+  Hash
+-------------------------------------------------------------------------------}
+
+instance Isomorphic WrapHeaderHash where
+  project = aux
```
edsko, line 204: Use `InstanceSigs` instead.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Unary.hs
```diff
@@ -96,316 +65,442 @@ import           Ouroboros.Consensus.HardFork.Combinator.Mempool
...
+      aux GenesisHash   = GenesisHash
+      aux (BlockHash h) = BlockHash (inject' (Proxy @(WrapHeaderHash blk)) h)
+
+{-------------------------------------------------------------------------------
+  Config
+
+  This is a bit trickier due to the partial configs.
+-------------------------------------------------------------------------------}
+
+instance Isomorphic TopLevelConfig where
```
edsko, line 241: Separate instances for ledger config / consensus config?

edsko: Can't be done. Added a comment:

```haskell
-- | Projection/injection for 'TopLevelConfig'
--
-- NOTE: We do not define one for 'LedgerConfig' or 'ConsensusConfig', since
-- we need the 'EraParams' for their injections, which we can only derive if
-- we have the top-level config.
```

edsko:
bors merge

PR #2192, opened 2020-06-02
Generalize hard fork ledger view

edsko:
The goal of this PR is to define

```haskell
hardforkForecast :: SListI xs
                 => HardForkState_ g g xs
                 -> Exactly xs EraParams
                 -> InPairs (Translate f) xs
                 -> NP (g -.-> Maybe :.: (Forecast :.: HardForkEraLedgerView_ f)) xs
                 -> WithOrigin SlotNo
                 -> Maybe (Forecast (HardForkLedgerView_ f xs))
```

which can then be used in the definition of `ledgerViewForecastAt` for `HardForkBlock`, but, crucially, is entirely independent from `CanHardFork`, and is entirely polymorphic in the types of the various ledgers. This will then make it easier to test this as part of the HF history tests.


edsko:
bors merge

edsko:
bors merge

PR #2194, opened 2020-06-02
Use new quickcheck-state-machine release

kderme:


mrBliss:
Any particular reason for updating other than switching from a revision to a released version?

kderme:
> Any particular reason for updating other than switching from a revision to a released version?

Not really. I just remember it had caused some issues with some dependent packages, which ignore the cabal.project.

mrBliss:
> > Any particular reason for updating other than switching from a revision to a released version?
> 
> Not really. I just remember it had caused some issues with some dependent packages, which ignore the cabal.project.

"which ignore the cabal.project", what exactly do you mean? The Nix build ignoring it?

I don't think there are any remaining issues now, correct me if I'm wrong.

As this is a test-only dependency, and no more downstream packages accidentally leave the consensus tests enabled, I expect nothing to change for downstream packages with this PR. Which is exactly what we want, but also defeats the purpose of it solving the issues.

However, I think it is better to use versions released on Hackage than specific revisions. Especially since the dependency can now be stored once and for all in the global Cabal store instead of in the project-local `dist-newstyle` folder, requiring a rebuild after each `cabal clean`.

mrBliss:
bors merge

PR #2197, opened 2020-06-03
test-consensus: off-by-one in the HFC A.hs ledger

nfrisby:
Small diff. 

FWIW, my draft PR for the more general test with varying epoch sizes, slot lengths, tx slot (and hence A era size), etc and epoch size and slot length changing at the fork, fails without this change. Even without that, I think we can demonstrate that @>=@ is wrong.

For example, `countSlots 6 5` is 1. But if we're asking "how many slots (which are 1-to-1 with blocks in this test) _after_ the block forged in slot 5 will there be _before_ the onset of slot 6", then the number we want to is actually 0. Note that we should count neither slot 5 (that's the block of interest) nor slot 6 (that's the first slot of the next epoch).

`master` currently includes the transition tx in slot 2, and the A-B fork happens at the end of the first epoch. So the A era is slots 0-5, and B starts with slot 6. If we instead include the tx in slot 4, then the transition still happens after the first fork. That is wrong, since `k = 2` and the tx's block is only buried under 1 block as of the onset of slot 6. After this PR, adding the tx in slot 4 correctly delays the A-B fork to after the second epoch.

My PR #2066 is hopefully imminent; as I said above, it exercises/needs this.

### DIFF
--- a/ouroboros-consensus/test-consensus/Test/Consensus/HardFork/Combinator.hs
+++ b/ouroboros-consensus/test-consensus/Test/Consensus/HardFork/Combinator.hs
@@ -20,6 +20,7 @@ import qualified Data.Map as Map
 import           Data.SOP.Strict hiding (shape)
 import           Data.Type.Equality
 import           Data.Void
+import           Data.Word
 
 import           Cardano.Slotting.Slot
 
@@ -66,6 +67,7 @@ import qualified Ouroboros.Consensus.HardFork.History as History
 import           Test.ThreadNet.General
 import           Test.ThreadNet.Network
 import           Test.ThreadNet.TxGen
+import           Test.ThreadNet.Util
 import           Test.ThreadNet.Util.NodeJoinPlan
 import           Test.ThreadNet.Util.NodeRestarts
 import           Test.ThreadNet.Util.NodeTopology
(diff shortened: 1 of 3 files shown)


nfrisby:
I've re-requested review, for two reason:

  * I changed the computation of the scheduled slot of the transition in the A ledger. It now supports arbitrary epoch sizes (>0). **Please** see the `asTypeOf` there and related comment.

  * I expanded the `TestSetup` type and its `Arbitrary` instance to generate many more test setup parameters. In particular, it randomizes when the transition transaction is included. The test therefore computes how many slots it needs to run for in order to see some B blocks. It fails if it doesn't see as many A and B blocks as it expects.

P.S. - I think we can remove `pgaCustomLabelling`. Do you agree? Please see diff.

ouroboros-consensus/test-consensus/Test/Consensus/HardFork/Combinator.hs
```diff
@@ -82,59 +84,81 @@ tests = testGroup "HardForkCombinator" [
...
+    , testSetupTxSlot     :: SlotNo
     }
   deriving (Show)
 
 instance Arbitrary TestSetup where
   arbitrary = do
-      testSetupSeed <- arbitrary
+      testSetupEpochSize <- EpochSize     <$> choose (1, 10)
+      testSetupK         <- SecurityParam <$> choose (2, 10)
+      -- TODO why does k=1 cause the nodes to only forge in the first epoch?
```
nfrisby, line 102: I'll check if this seems related to Issue #1511, which is overdue for a revisit since so much consensus logic has changed since I last tried.

ouroboros-consensus/test-consensus/Test/Consensus/HardFork/Combinator/A.hs
```diff
@@ -248,17 +261,66 @@ instance LedgerSupportsMempool BlockA where
...
+              -- @safeFromTipA k@ blocks /after/ reporting the transition and
+              -- /before/ the start of the next era.
+              --
+              -- Thus, we schedule the next era to begin with the first
+              -- upcoming epoch that starts /after/ we're guaranteed to see
+              -- both the aforementioned @k@ additional blocks and also a
+              -- further @safeFromTipA k@ slots after the last of those.
+
+              -- the last slot that must be in the current era
+              firstPossibleLastSlotThisEra =
```
edsko, line 292: Decided to stick with "first possible last slot" after all eh?

ouroboros-consensus/test-consensus/Test/Consensus/HardFork/Combinator/A.hs
```diff
@@ -248,17 +261,66 @@ instance LedgerSupportsMempool BlockA where
...
+                     --
+                     -- The following would be equivalent if it couldn't fail
+                     -- with 'PastHorizonException', which it may since we may
+                     -- be inspecting a slot beyond the ledger's safe zone. In
+                     -- particular, the @ei@ the HFC provided to us is overly
+                     -- conservative for our specific purpose here. We're using
+                     -- it to decide here when this era should end. This era
+                     -- could only end sooner than that if this decision itself
+                     -- gets discarded! (TODO: double-check this claim)
+                     runEI epochInfoEpoch firstPossibleLastSlotThisEra
```
edsko, line 311: I think the point is that we _should_ be able to do this calculation, whilst observing safe zones. I will try to change this.

edsko: Ah, no, I am wrong. Now that we've spelled it out like this, it's almost _by definition_ that we are looking more than the safe zone ahead: indeed, we are looking _twice_ the safe zone ahead. Subtle. This means that the calculation for Byron must be careful here also.

edsko: (Except of course Byron doesn't actually _use_ `EpochInfo` and just assumes epoch sizes are forever the same. That might actually be correct, in this instance -- but for Shelley to Goguen we'll need to be careful.)

edsko:
(Also rebased this on latest `master`)

edsko:
bors merge

Issue #2201, opened 2020-06-04
Shelley: fix incorrect stabilityWindow


mrBliss:
We currently do the  `3*k/f` computation for the `stabilityWindow` in 2 places:

https://github.com/input-output-hk/ouroboros-network/blob/7826943cc9ead0d5936ec88d8ff429c496e91007/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L151

https://github.com/input-output-hk/ouroboros-network/blob/0d96f4822365b22d8998875c2c90278c0c1ab98a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs#L350

Note that the computation is slightly different (`fromRational` vs `fromIntegral`, type applications, etc.). Moreover in one place we do **3*f/k** instead of **3*k/f**!

The same applies for `randomnessStabilisationWindow`.

I'm afraid it's not straightforward to unify the two places where we create the `SL.Globals` (one uses `TPraosParams`, the other `ShelleGenesis`). But what we *can* do, is introduce two functions `computeStabilityWindow k f = ..` and `computeRandomnessStabilisationWindow k f = ..`, and use them in both places.

PR #2202, opened 2020-06-04
Rationalise stability window calculations.

nc6:
Addresses #2201

We still compute these in a couple of places, but now using a shared
function.


mrBliss:
bors merge

PR #773, opened 2019-07-15
Have some nodes join later in test-consensus

nfrisby:
~This PR is only for discussion.~ I have a first draft take on #231 and would like feedback, of all kinds.

Edit: the PR has matured and looks nearly ready to merge, but I suspect I should open a separate PR because of the 100+ comments on this one.

Edit: Fixes #231.

PR #2222, opened 2020-06-05
reorganize TestConfig and TestConfigBlock

nfrisby:
This yak shaving pays down some tech debt and paves the way for PR #2066. Now that timey wimey things like `slotLength` and `epochSize` will be varying during the execution for some, their semantics and validity is more block-dependent than ever. So we're separating them out. This enables some reuse and will simplify the story told by the commits for the imminent PR #2066.

I did some weeding while I was in there, since I was already touching so many lines.

This diff is intended as a functionality-preserving refactoring with two major exceptions.

  * I've changed generators, so QC replay seeds are boinked (I don't think we rely on these anywhere.)

  * I expanded the parameters varied during the mock Praos tests (`LeaderSchedule.hs` and `Praos.hs`). I don't expect that to reveal any bugs we had been accidentally avoiding, but it could. Or maybe might cause timeouts. But again, I don't think so.

### DIFF
--- a/ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
+++ b/ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
@@ -9,13 +9,14 @@
 module Test.ThreadNet.RealPBFT (
     tests
     -- * To support the DualPBFT tests
+  , TestSetup (..)
   , noEBBs
   , realPBftParams
-  , genRealPBFTNodeJoinPlan
-  , shrinkTestConfigSlotsOnly
+  , genTestSetup
   , expectedCannotLead
   ) where
 
+import           Control.Monad (join)
 import           Control.Monad.Except (runExceptT)
 import           Data.Coerce (coerce)
 import           Data.Functor.Identity
@@ -71,6 +72,7 @@ import           Test.ThreadNet.General
 import           Test.ThreadNet.Network (NodeOutput (..),
                      TestNodeInitialization (..))
 import qualified Test.ThreadNet.Ref.PBFT as Ref
+import           Test.ThreadNet.Rekeying
 import           Test.ThreadNet.TxGen.Byron ()
 import           Test.ThreadNet.Util
 import           Test.ThreadNet.Util.NodeJoinPlan
@@ -78,29 +80,64 @@ import           Test.ThreadNet.Util.NodeRestarts
 import           Test.ThreadNet.Util.NodeTopology
 
 import           Test.Util.Orphans.Arbitrary ()
-import           Test.Util.Shrink (andId, dropId)
 import qualified Test.Util.Stream as Stream
 import           Test.Util.WrappedClock (NumSlots (..))
 
 import           Test.ThreadNet.RealPBFT.ProtocolInfo
 import           Test.ThreadNet.RealPBFT.TrackUpdates
 
--- | Generate k values as small as this module is known to handle.
---
--- TODO Issue #1566 will bring this to k>=0, at which point we may be able to
--- relocate this to a more general module.
---
-newtype K1 = K1 SecurityParam
+data TestSetup = TestSetup
+  { setupEBBs         :: ProduceEBBs
+  , setupK            :: SecurityParam
+  , setupTestConfig   :: TestConfig
+  , setupNodeJoinPlan :: NodeJoinPlan
+  , setupNodeRestarts :: NodeRestarts
+  , setupSlotLength   :: SlotLength
+  }
   deriving (Show)
 
-instance Arbitrary K1 where
-  arbitrary                     = (K1 . SecurityParam) <$> elements [1 .. 10]
-  shrink (K1 (SecurityParam k)) = (K1 . SecurityParam) <$> [ 1 .. k - 1 ]
+instance Arbitrary TestSetup where
+  arbitrary = do
+     -- TODO Issue #1566 will bring this to k>=0
+      k <- SecurityParam <$> choose (1, 10)
+
+      join $ genTestSetup k <$> arbitrary <*> arbitrary <*> arbitrary
+
+  -- TODO shrink
+
+-- | An entrypoint used by "Test.ThreadNet.DualPBFT"
+--
+-- See the @'Arbitrary' 'Test.ThreadNet.DualPBFT.SetupDualPBft'@ instance.
+genTestSetup :: SecurityParam -> NumCoreNodes -> NumSlots -> SlotLength -> Gen TestSetup
+genTestSetup k numCoreNodes numSlots setupSlotLength = do
+    setupEBBs    <- arbitrary
+    initSeed     <- arbitrary
+    nodeTopology <- genNodeTopology numCoreNodes
+
+    let testConfig = TestConfig
+          { initSeed
+          , nodeTopology
+          , numCoreNodes
+          , numSlots
+          }
+    let params = realPBftParams k numCoreNodes
+
+    nodeJoinPlan <- genRealPBFTNodeJoinPlan params numSlots
+    nodeRestarts <- genNodeRestarts nodeJoinPlan numSlots >>=
+                    genNodeRekeys params nodeJoinPlan nodeTopology numSlots
+
+    pure $ TestSetup
+      setupEBBs
+      k
+      testConfig
+      nodeJoinPlan
+      nodeRestarts
+      setupSlotLength
 
 tests :: TestTree
 tests = testGroup "RealPBFT" $
     [ testProperty "trivial join plan is considered deterministic"
-        $ \(K1 k) numCoreNodes ->
+        $ \TestSetup{setupK = k, setupTestConfig = TestConfig{numCoreNodes}} ->
           prop_deterministicPlan $ realPBftParams k numCoreNodes
     , localOption (QuickCheckTests 10) $   -- each takes about 0.5 seconds!
       testProperty "check setup"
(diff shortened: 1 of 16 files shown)


nfrisby:
Quoting the commit message:

```
TODO I've removed all shrinking for now. I will revisit that shortly. It was a
    mess, slow, and rarely helpful except when the problem was trivial, like a bug
    preventing nodes from producing any blocks whatsoever.
```

I mean to follow-up it up with a later PR. My though process was that I'm most likely to debug an unshrink case anyway, so I'll pay that down earlier than planned if needed.

But @mrBliss and @nc6 may be relying on shrinking for their current dev work (I'd be surprised, since it's been so poor for me). Yell if you feel like these shrinkers gone missing will block you.

(Also, I expect this diff might conflict with your hot branches.)

nfrisby:
@nc6 -- I increased the KES periods for the mock praos ledgers from 100 to 1000 -- I think we exhausted them.

I have no idea the related expense. Should we try harder to find a smaller number (eg 200)?

nfrisby:
@mrBliss @nc6 I introduced slightly more duplication in `Cardano.hs` and `RealTPraos.hs` instead of expanding `Infra/Shelly.hs`. I think `Cardano.hs` is soon to change sharply anyway due to HFC, so the currently opportunity for code sharing seemed accidental, and so I didn't take it.

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/LeaderSchedule.hs
```diff
@@ -99,32 +95,44 @@ prop_simple_leader_schedule_convergence
...
+                  , praosSlotsPerEpoch = unEpochSize epochSize
+                  , praosLeaderF       = dummyF
+                  , praosLifetimeKES   = 1000000
+                  }
+                  (HardFork.defaultEraParams k slotLength)
+                  schedule
+            , mkRekeyM = Nothing
             }
 
-    -- I don't think this value actually matters if we override the leader
```
edsko, line 123: Comment intentionally deleted? It still applies, right?

edsko:
> @mrBliss @nc6 I introduced slightly more duplication in Cardano.hs and RealTPraos.hs instead of expanding Infra/Shelly.hs. I think Cardano.hs is soon to change sharply anyway due to HFC, so the currently opportunity for code sharing seemed accidental, and so I didn't take it.

In particular, `genRealTPraosTestConfig` and `shrinkRealTPraosTestConfig` have been deleted, not that they did very much. The former merely picked `numCoreNodes`, `numSlots` and `initSeed`, and the latter literally did nothing. I guess the original plan was that the generator for Cardano could _depend_ on the one for Shelley, but we'll see how things pan out.

edsko:
> But @mrBliss and @nc6 may be relying on shrinking for their current dev work (I'd be surprised, since it's been so poor for me). Yell if you feel like these shrinkers gone missing will block you.

Pretty sure this is fine. I think when you/we redesign these tests, good shrinking should be one of the priorities. It's hard to retrofit that to an existing design. (You may or may not find [my blog post about QSM](http://www.well-typed.com/blog/2019/01/qsm-in-depth/) helpful -- even if we aren't using QSM in these tests, that blog post talks a lot about shrinking.)

ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
```diff
@@ -116,14 +153,18 @@ tests = testGroup "RealPBFT" $
...
-            , nodeJoinPlan = NodeJoinPlan $ Map.fromList [(CoreNodeId 0,SlotNo 0), (CoreNodeId 1,SlotNo 20), (CoreNodeId 2,SlotNo 22)]
-            , nodeRestarts = noRestarts
-            , nodeTopology = meshNodeTopology ncn
-            , slotLength   = defaultSlotLength
-            , initSeed     = Seed (15069526818753326002, 9758937467355895013, 16548925776947010688, 13173070736975126721, 13719483751339084974)
+          prop_simple_real_pbft_convergence TestSetup
+            { setupEBBs       = ProduceEBBs
+            , setupK          = SecurityParam 10
+            , setupTestConfig = TestConfig
+              { initSeed     = Seed (15069526818753326002, 9758937467355895013, 16548925776947010688, 13173070736975126721, 13719483751339084974)
```
edsko, line 160: Just a thought for when we do redesign: we should also try hard to get rid of this `seed` parameter. It's inherently shrinkable, and very brittle.

edsko:
> I've changed generators, so QC replay seeds are boinked (I don't think we rely on these anywhere.)

But you _do_, right? There quite a few test cases that depend on specific seeds (I'm thinking of tests like `addressed by InvalidRollForward exception (PR #773)"` -- at least, those tests _specify_ a seed. Don't they rely on it?

nfrisby:
> > I've changed generators, so QC replay seeds are boinked (I don't think we rely on these anywhere.)
> 
> But you _do_, right? There quite a few test cases that depend on specific seeds (I'm thinking of tests like `addressed by InvalidRollForward exception (PR #773)"` -- at least, those tests _specify_ a seed. Don't they rely on it?

There's no place I'm aware of where we depend on the QuickCheck replay seed. EG `--quickcheck-replay`. What we do have is the `Seed` argument, which the simulation uses to generator eg transactions. That is fixed for some of our regression tests. But that is part of our own test setup data, "downstream" of the "internal" QuickCheck seed.

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/Praos.hs
```diff
@@ -103,27 +111,39 @@ prop_simple_praos_convergence
...
-                                        praosSecurityParam
-                                        praosSlotLength)
-            , rekeying    = Nothing
-            , txGenExtra  = ()
+                                        k
+                                        slotLength)
+            , mkRekeyM = Nothing
             }
-
-    -- TODO: Right now mock Praos has an explicit epoch size in its parameters
```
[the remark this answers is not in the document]
nfrisby: These `PraosParams` only need to match the values used during the relevant era, the one running praos, right? So whatever instance allows this mock `Praos` ledger to be used with the HFC can include an `assert` that compares to the `EpochInfo` that the HFC provides.

I though that seems OK for a mock protocol. Am I missing something?

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/LeaderSchedule.hs
nfrisby: I've put it back now. I think I replaced it because it was previously "justifying" using a fixed value, but now I've included it in the generator -- so we're sweeping its values. But, it is still good to comment saying that we expect its value should not matter.

ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
nfrisby: Hmm. I need to think more what it would look like to replace it. My hunch is that we can't (without a lot of work), since the nodes have to generate eg transactions that depend on behaviors of the system under test that we currently can't always anticipate (eg Praos leader schedules, which leaders non-leaders follow, etc).

For the Praos tests I'm developing with message delays, the "statistical test" aspects from the QuviQ presentation use an "internal" `forAll` in order to generate enough samples to satisfy the statistical test's desired confidence level. That `forAll` seems like the right place for generating this kind of seed. Not sure how to generalize that observation to the rest of the protocols, especially the ones that I haven't been anticipating using statistical testing for.

nfrisby:
I just pushed an amended the commit with the following commit folded into it.

```
$ git show
commit 1837273b8eb58e2366088b5b494c543632927e46
Author: Nicolas Frisby <nick.frisby@iohk.io>
Date:   Sat Jun 6 07:58:35 2020 -0700

    TOSQUASH: address PR comment

diff --git a/ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/LeaderSchedule.hs b/ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/LeaderSchedule.hs
index e5baa0dd9..34ccd464d 100644
--- a/ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/LeaderSchedule.hs
+++ b/ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/LeaderSchedule.hs
@@ -43,6 +43,8 @@ data TestSetup = TestSetup
   { setupK              :: SecurityParam
   , setupTestConfig     :: TestConfig
   , setupEpochSize      :: EpochSize
+    -- ^ Note: we don't think this value actually matters, since this test
+    -- overrides the leader schedule.
   , setupNodeJoinPlan   :: NodeJoinPlan
   , setupLeaderSchedule :: LeaderSchedule
   , setupSlotLength     :: SlotLength
```

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/Praos.hs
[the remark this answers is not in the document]
edsko: No. It should be using an `EpochInfo`, not an `EpochSize`. An` EpochSize` does not allow any kind of looking ahead, or indeed, looking back, past the era -- an `EpochInfo Identity` is an assumption that things are static -- but could still have different values across eras. An `EpochSize` on the other hand is an assumption things are static, _and have the same value in all eras_. Admittedly one could ask quite how much a ledger _should_ look past its boundaries, but at least with an `EpochInfo` we can _set_ some bounds; with `EpochSize` we cannot.

ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
edsko: One thing that John emphasizes a _lot_ is that the test we run _should not depend on the execution of that test_. Everything that we're going to throw at the test should be determined _a priori_, so that it is inspectable, reproducible, and crucially, shrinkable. Of course, achieving this for the kinds of tests that we want to do is not easy, but we should strive hard hard to get there.

edsko:
I think you might have a lost a commit that I pushed yesterday that fixed some CI warnings. Feel free to merge this after CI is green.

nfrisby:
> I think you might have a lost a commit that I pushed yesterday that fixed some CI warnings. Feel free to merge this after CI is green.

Silly me, yes. I've recovered and squashed it in, with another rebase.

Will do.

ouroboros-consensus/ouroboros-consensus-mock/test/Test/ThreadNet/Praos.hs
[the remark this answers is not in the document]
nfrisby: Ah, I had the scope wrong. I now see that this TODO comment was implying a change the mock Praos protocol, not just a change to the module dedicated to testing it.

I opened PR #2225.

nfrisby:
Well, my squash seems to have confused CI to cause it to retain a cached error :(

So I've force pushed now just the unsquashed cherry-pick. 🤞

nfrisby:
bors merge

PR #2228, opened 2020-06-08
Fix broken q-s-m tests and all bugs that appeared

mrBliss:
In #1739, I made some changes as some of our q-s-m infrastructure had been adopted by quickcheck-state-machines. Unfortunately, because of confusion with different types adopted in quickcheck-state-machines than the ones we use, our postconditions started checking the result of the implementation with the result of the *implementation* instead of the result of the model 🤦.

Revert the changes that led to this and fix all bugs there were hidden because of it.

### DIFF
--- a/ouroboros-consensus/test-storage/Test/Ouroboros/Storage/TestBlock.hs
+++ b/ouroboros-consensus/test-storage/Test/Ouroboros/Storage/TestBlock.hs
@@ -169,7 +169,7 @@ data EBB =
 
 instance Hashable EBB where
   hashWithSalt s (EBB epoch)  = hashWithSalt s (unEpochNo epoch)
-  hashWithSalt s RegularBlock = hashWithSalt s (0 :: Int)
+  hashWithSalt s RegularBlock = hashWithSalt s (-1 :: Int)
 
 data TestBody = TestBody {
       tbForkNo  :: !Word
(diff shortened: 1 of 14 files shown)


mrBliss:
Draft because there is still at least one bug in the ChainDB tests and the FS tests still need to be fixed (`renameFile`).

Also, I expect the FS tests to start failing on darwin and mingw32. We might have to disable those again (preferably just the problematic test by an OS check, instead of the *whole* testsuite in Nix, like before) or fix them.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/TestBlock.hs
```diff
@@ -314,6 +316,59 @@ testBlockToBlockInfo tb = BlockInfo {
   where
     TestHeader{..} = testHeader tb
 
+{-------------------------------------------------------------------------------
+  Real chain length
+-------------------------------------------------------------------------------}
+
+-- | In chain selection, we use 'BlockNo' as a proxy for the block length.
+-- This is entirely correctly, except for those dreadful EBBs, which share
```
mrBliss, line 324:
```suggestion
-- This is entirely correct, except for those dreadful EBBs, which share
```

mrBliss:
bors merge

PR #2231, opened 2020-06-09
OpCert replacement test

nc6:
This adds a test of operational certificate replacement in Shelley.

- Extracts some more general Shelley configuration.
- Construct a node restart schedule for opcert replacement. This deliberately replaces some opcerts not on the correct slot, so we expect some nodes to be unable to lead in some slots.
- This entails making some changes to the "rekeying" infrastructure, because we need access to the slot.
- RealPBFT tests are updated accordingly.

At the moment we allow any slot to `CannotLead`, rather than deducing explicitly what slots cannot lead from the restart schedule. This PR has been dragging for a while, so I'd rather get some testing in now and then improve later.

### DIFF
--- a/ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
+++ b/ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
@@ -32,6 +32,7 @@ import           Test.Tasty
 import           Test.Tasty.QuickCheck
 
 import           Cardano.Crypto.Seed (mkSeedFromBytes)
+import           Cardano.Slotting.EpochInfo
 import           Cardano.Slotting.Slot
 
 import           Ouroboros.Network.Block (SlotNo (..))
@@ -871,6 +872,8 @@ prop_simple_real_pbft_convergence TestSetup
       , slotLength
       , txGenExtra = ()
       }
+    epochInfo :: EpochInfo Identity
+    epochInfo = fixedSizeEpochInfo epochSize
 
     testOutput =
         runTestNetwork testConfig testConfigB TestConfigMB
@@ -891,7 +894,7 @@ prop_simple_real_pbft_convergence TestSetup
                           ]
                         Ref.Nondeterministic{} -> Set.empty
                   in Set.lookupGT s nominalSlots
-              , rekeyUpd      = mkRekeyUpd genesisConfig genesisSecrets
+              , rekeyUpd      = mkRekeyUpd genesisConfig genesisSecrets epochInfo
               , rekeyFreshSKs =
                   let prj  = Crypto.hashVerKey . Crypto.deriveVerKeyDSIGN
                       acc0 =   -- the VKs of the operational keys at genesis
@@ -1288,15 +1291,18 @@ genNodeRekeys params nodeJoinPlan nodeTopology numSlots@(NumSlots t)
 mkRekeyUpd
   :: Genesis.Config
   -> Genesis.GeneratedSecrets
+  -> EpochInfo Identity
+  -> CoreNodeId
   -> ProtocolInfo (ChaChaT m) ByronBlock
-  -> EpochNo
+  -> SlotNo
   -> Crypto.SignKeyDSIGN Crypto.ByronDSIGN
-  -> Maybe (TestNodeInitialization m ByronBlock)
-mkRekeyUpd genesisConfig genesisSecrets pInfo eno newSK =
+  -> Maybe (TestNodeInitialization ByronBlock)
+mkRekeyUpd genesisConfig genesisSecrets epochInfo _ pInfo sno newSK =
   case pInfoLeaderCreds of
     Nothing              -> Nothing
     Just (isLeader, mfs) ->
       let PBftIsLeader{pbftCoreNodeId} = isLeader
+          eno = runIdentity $ epochInfoEpoch epochInfo sno
           genSK = genesisSecretFor genesisConfig genesisSecrets pbftCoreNodeId
           isLeader' = updSignKey genSK configBlock isLeader (coerce eno) newSK
           pInfo' = pInfo { pInfoLeaderCreds = Just (isLeader', mfs) }
--- a/ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
+++ b/ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
@@ -1296,7 +1296,7 @@ mkRekeyUpd
   -> ProtocolInfo (ChaChaT m) ByronBlock
   -> SlotNo
   -> Crypto.SignKeyDSIGN Crypto.ByronDSIGN
-  -> Maybe (TestNodeInitialization ByronBlock)
+  -> Maybe (TestNodeInitialization m ByronBlock)
 mkRekeyUpd genesisConfig genesisSecrets epochInfo _ pInfo sno newSK =
   case pInfoLeaderCreds of
     Nothing              -> Nothing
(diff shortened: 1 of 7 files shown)


ouroboros-consensus-byron-test/test/Test/ThreadNet/RealPBFT.hs
```diff
@@ -1288,15 +1291,18 @@ genNodeRekeys params nodeJoinPlan nodeTopology numSlots@(NumSlots t)
...
+  -> SlotNo
   -> Crypto.SignKeyDSIGN Crypto.ByronDSIGN
   -> Maybe (TestNodeInitialization m ByronBlock)
-mkRekeyUpd genesisConfig genesisSecrets pInfo eno newSK =
+mkRekeyUpd genesisConfig genesisSecrets epochInfo _ pInfo sno newSK =
   case pInfoLeaderCreds of
     Nothing              -> Nothing
     Just (isLeader, mfs) ->
       let PBftIsLeader{pbftCoreNodeId} = isLeader
+          eno = runIdentity $ epochInfoEpoch epochInfo sno
```
nfrisby, line 1305: On some level, this should be `futureSlotToEpoch` instead of `fixedSizeEpochInfo`. But that will only be necessary if `RealPBFT` is updated to include a hard fork. So it's moot until then.

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -2,124 +2,83 @@
...
-    pure TestSetup
-      { setupD
-      , setupK
-      , setupTestConfig
-      }
-
-  -- TODO shrink
 
 tests :: TestTree
 tests = testGroup "RealTPraos"
```
nfrisby, line 26: This diff somewhat undoes my recent PR by moving the generator back into the `Infra.Shelley`. I assume that's so it can be shared with `ouroboros-consensus-cardano:test:test`. That seems very reasonable for now, but I anticipate that we wil eventually want separate test setup generators for those two.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -268,3 +307,195 @@ mkProtocolRealTPraos genesis CoreNode { cnDelegateKey, cnVRF, cnKES, cnOCert } =
...
+  => Seed
+  -> ShelleyGenesis c
+  -> [CoreNode c]
+  -> Rekeying m (ShelleyBlock c)
+ocertRekeying
+  initSeed
+  ShelleyGenesis {sgSlotsPerKESPeriod}
+  coreNodes
+  = Rekeying
+  { rekeyOracle = \_ sn -> Just sn
```
nfrisby, line 453: @nc6 Would you revisit the reason you mentioned for this being a valid `rekeyOracle` and at it as a comment here?

nfrisby: Err, I now see `tniCrucialTxs = []` below. So that's probably why the oracle doesn't matter. This rekey doesn't require posting anything on the chain, so timeliness/promptness isn't a concern like it was for Byron.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -268,3 +307,195 @@ mkProtocolRealTPraos genesis CoreNode { cnDelegateKey, cnVRF, cnKES, cnOCert } =
...
+  , rekeyUpd = \(CoreNodeId nodeIdx) oldProtocolInfo (SlotNo slotNo) kesKey ->
+      let
+        currentKESPeriod = fromIntegral $ slotNo `div` sgSlotsPerKESPeriod
```
nfrisby, line 456: `ouroboros-consensus-cardano:test:test` will eventually use a the HFC. Will this computation yield the correct period when the Shelley chain doesn't start at 0?

ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -268,3 +307,195 @@ mkProtocolRealTPraos genesis CoreNode { cnDelegateKey, cnVRF, cnKES, cnOCert } =
...
+testTuning :: TestTuning
+testTuning = TestTuning
+  { rotateKESKeys = False }
+
+genShelleyTestConfig
+  :: TestTuning
+  -> Gen ShelleyTestConfig
+genShelleyTestConfig TestTuning {rotateKESKeys} = do
+    initSeed <- arbitrary
+    k <- SecurityParam <$> Gen.elements [5,10]
```
nfrisby, line 336: I'd suggest `choose (2, 10)` for `k`.

nfrisby:
I haven't review the `cardano` bit since it's out-of-date (doesn't compile).

I looked over `Infra.Shelley` and added a few comments. It generally makes sense, but I'm not particularly familiar with the KES management. @mrBliss ?

The other modules LGTM.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -394,9 +391,6 @@ runThreadNetwork ThreadNetworkArgs
   where
     _ = keepRedundantConstraint (Proxy @(Show (LedgerView (BlockProtocol blk))))
 
-    epochInfo :: EpochInfo m
-    epochInfo = fixedSizeEpochInfo epochSize
-
```
mrBliss, line 399: @nfrisby you're working on testing hard forks, in which the epoch size will no longer be static. Do you think there will be an oracular `EpochInfo`? If so, wouldn't it live here, in a central location?

ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -142,14 +163,32 @@ genCoreNode startKESPeriod = do
       seed <- mkSeedFromBytes <$> getRandomBytes sz
       pure $ go seed
 
+genOCert
```
mrBliss, line 166: Since this is pure, I'd call it `mkOCert`.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -268,3 +307,195 @@ mkProtocolRealTPraos genesis CoreNode { cnDelegateKey, cnVRF, cnKES, cnOCert } =
...
+                    ( fromIntegral np
+                    , (fromIntegral np, ix + 1)
+                    )
+                  )
+                  (firstKESPeriod, 0)
+          pure
+            [ (SlotNo slot, CoreNodeId coreNodeIdx)
+              | slot <- takeWhile (< t) ((kesPeriodStartSlots !!) <$> periods)
+            ]
+        pure
```
mrBliss, line 433: It's hard (at least for me) to see with which lines (415-418) these align. Extracting the code in the `forM` into a local helper will make this easier to read for me :slightly_smiling_face:

mrBliss:
@nc6 I seem to remember you were running into problems with invalid KES keys in the tests of this PR (or was that somewhere else?). #2300 was merged today, in which @edsko and I refactor the way the `HotKey` is handled (no more `error` in case of an invalid key!). Maybe that helps?

mrBliss:
@nc6 Do you think you'll have a chance to take a look at this?

nc6:
I'm going to close this, since it's at least two refactors out of date and I am unlikely to have the chance to look at it.

PR #2248, opened 2020-06-11
Introduce TranslateNetworkProtocolVersion

edsko:
The HFC needs _versions_ for each eras, but the _translation_ can only be
specified at the top-level, this is not compositional.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs
@@ -4,82 +4,85 @@
 
 module Ouroboros.Consensus.Node.NetworkProtocolVersion
   ( HasNetworkProtocolVersion(..)
+  , TranslateNetworkProtocolVersion(..)
+    -- * Re-exports
+  , NodeToNodeVersion(..)
+  , NodeToClientVersion(..)
   ) where
 
 import           Data.List.NonEmpty (NonEmpty (..))
 import           Data.Proxy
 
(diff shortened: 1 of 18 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs
```diff
@@ -74,6 +67,17 @@ class ( Show (NodeToNodeVersion   blk)
...
+  -- | Translate to network-layer type
+  nodeToNodeProtocolVersion
+    :: Proxy blk -> NodeToNodeVersion blk -> N.NodeToNodeVersion
+
+  -- | Translate to network-layer type
+  nodeToClientProtocolVersion
+    :: Proxy blk -> NodeToClientVersion blk -> N.NodeToClientVersion
+
+  -- Defaults
+
```
mrBliss, line 80: Should we make `N.NodeToClientV_2` the default, now that the `LocalStateQuery` protocol has been enabled?

edsko: :shrug:

edsko:
bors merge

edsko:
bors cancel

ouroboros-consensus/src/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs
edsko: Defaults are now gone, and V2 is now the enabled for the mock ledger.

edsko:
bors merge

PR #2250, opened 2020-06-12
Don't give the illusion that we can version hashes

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -241,6 +242,12 @@ defaultCodecs ccfg version = Codecs {
     dec :: SerialiseNodeToNode blk a => forall s. Decoder s a
     dec = decodeNodeToNode ccfg version
 
+    encHash :: HeaderHash blk -> Encoding
+    encHash = Serialise.encode . toRawHash (Proxy @blk)
+
+    decHash :: forall s. Decoder s (HeaderHash blk)
+    decHash = fromRawHash (Proxy @blk) <$> Serialise.decode
```
mrBliss, line 249: It's the third time I see these now, we should define them near `ConvertRawHash`

ouroboros-consensus-byron-test/src/Ouroboros/Consensus/ByronDual/Node/Serialisation.hs
```diff
@@ -99,11 +99,6 @@ instance DecodeDisk DualByronBlock (AnnTip DualByronBlock) where
 
 instance SerialiseNodeToNodeConstraints DualByronBlock
 
--- | @'HeaderHash' 'DualByronBlock'@
-instance SerialiseNodeToNode DualByronBlock ByronHash where
-  encodeNodeToNode _ _ = encodeByronHeaderHash
-  decodeNodeToNode _ _ = decodeByronHeaderHash
```
mrBliss, line 105: Lets' get rid of `encodeByronHeaderHash` and `decodeByronHeaderHash` as well.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
edsko: Ah, okay, will do.

ouroboros-consensus-byron-test/src/Ouroboros/Consensus/ByronDual/Node/Serialisation.hs
edsko: These are still used internally.

edsko:
bors merge

Issue #2184, opened 2020-06-01
newTimeout is not safe on `32`-bit archs


coot:
As noted by @njd42 :

> On a 32 bit machine with a timeout more than ~35min in the future defaultRegisterDelay would be invoked which, in turn, makes use of newTimeout. This calls GHC.registerTimeout where the intrinsic 32 bit constraint rears its ugly head again.

coot:
We haven't fixed `newTimeout` bug in the pr input-output-hk/io-sim#57 addresses safety of `MonadTimer` primitives in the `si-timers` package.

Issue #1416, opened 2020-01-07
DeltaQ measures


coot:
We should finish DeltaQ architecture, and use it in
* peer2peer governor (this is tracked in #1415)
* BlockFetch (this is tracked in #469)

PR #2251, opened 2020-06-12
Keep Alive protocol

karknu:


ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -162,7 +162,11 @@ data NodeToNodeProtocols appType bytes m a b = NodeToNodeProtocols {
 
     -- | tx-submission mini-protocol
     --
-    txSubmissionProtocol :: RunMiniProtocol appType bytes m a b
+    txSubmissionProtocol :: RunMiniProtocol appType bytes m a b,
+
+    -- | keep-alive mini-protocol
+    --
+    keepAliveProtocol :: RunMiniProtocol appType bytes m a b
```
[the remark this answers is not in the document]
karknu: We will need a new version for this protocol, I just haven't considered it yet.
This is weeks from being useful so it will not make `NodeToNodeV_2`.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -251,6 +253,7 @@ type EdgeStatusVar m = StrictTVar m EdgeStatus
 -- each node.
 runThreadNetwork :: forall m blk.
                     ( IOLike m
+                    , MonadTime m
```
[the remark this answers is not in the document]
karknu: I didn't know about `MonadMonotonicTime`, but it looks like it is exactly what is needed. And we should probably go through the networking parts of ouroboros-network and switch `MonadTime` to `MonadMonotonicTime`.

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
```diff
@@ -269,3 +275,21 @@ gsvRequestResponseDuration PeerGSV{outboundGSV, inboundGSV}
         gsvTrailingEdgeArrive outboundGSV reqSize
      <> gsvTrailingEdgeArrive inboundGSV respSize
 
+
+defaultGSV :: PeerGSV
+defaultGSV = PeerGSV { outboundGSV, inboundGSV }
+  where
+    default_g = 3600 -- 1h, start with an unreasonable large G.
```
semanticphilosopher, line 282: This needs to be `min (div maxBound 4, 3600)` so that we don't get bitten by 32ness later

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
```diff
@@ -269,3 +275,21 @@ gsvRequestResponseDuration PeerGSV{outboundGSV, inboundGSV}
...
+    default_s = 2e-6 -- 5Mbps.
```
semanticphilosopher, line 283: sorry to nit pick - but that is 4Mpbs

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
karknu: There isn't a maxBound for DiffTime.

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
karknu: I was lazy and stole it from https://github.com/input-output-hk/ouroboros-network/blob/6507ec6eb08e75925e9c3072a11db84e226bf706/ouroboros-network/src/Ouroboros/Network/BlockFetch.hs#L289

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
coot: That's the problem, you can still declare to large `DiffTime` which will not fit into `Int32`. `threadDelay` is safe but `registerDelay` is not.

coot: ref: #2184

semanticphilosopher:
I was going to suggest we used a technique such as https://github.com/input-output-hk/ouroboros-network/blob/6507ec6eb08e75925e9c3072a11db84e226bf706/io-sim-classes/src/Control/Monad/Class/MonadTimer.hs#L125-L136 <https://github.com/input-output-hk/ouroboros-network/blob/6507ec6eb08e75925e9c3072a11db84e226bf706/io-sim-classes/src/Control/Monad/Class/MonadTimer.hs#L125-L136>

Perhaps export maxDelay as an attribute of MonadTImer ?

karknu:
That is a great idea. I will define a local `maxG` for now.

coot:
Part of #1416

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
karknu: This is changed to use to old hard coded g value (500ms). The reason for that is that we have to have a reasonable value in case the `KeepAlive` protocol isn't supported.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -130,6 +134,16 @@ data Handlers m peer blk = Handlers {
         -> peer
         -> TxSubmissionServerPipelined (GenTxId blk) (GenTx blk) m ()
 
+    , hKeepAliveClient
+        :: NodeToNodeVersion
+        -> StdGen
+        -> ScheduledStop m
+        -> peer
+        -> (StrictTVar m (M.Map peer PeerGSV))
```
mrBliss, line 142:
```suggestion
        -> StrictTVar m (Map peer PeerGSV)
```

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -37,8 +37,10 @@ import           Control.Monad (forever)
 import           Control.Monad.Class.MonadTimer (MonadTimer)
 import           Control.Tracer
 import           Data.ByteString.Lazy (ByteString)
+import qualified Data.Map.Strict as M
```
mrBliss, line 40:
```suggestion
import           Data.Map.Strict (Map)
```

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -130,6 +134,16 @@ data Handlers m peer blk = Handlers {
...
 
+    , hKeepAliveClient
+        :: NodeToNodeVersion
+        -> StdGen
+        -> ScheduledStop m
+        -> peer
+        -> (StrictTVar m (M.Map peer PeerGSV))
+        -> KeepAliveInterval
+        -> StrictTVar m (Maybe Time)
+        -> KeepAliveClient m ()
```
mrBliss, line 145: Does it have to take all these arguments? Compared to the other handlers, it takes a lot of them. I would expect that the only extra arguments needed are the ones shared with other handlers. The other ones can be given in `mkHandlers`. For example, `ScheduledStop m` and `KeepAliveInterval` are just static and can be moved from `aKeepAliveClient` to `hKeepAliveClient` (or are you planning on making them dynamic or shared?)

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -960,11 +961,14 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
                                                   -- interval which doesn't play nice with
                                                   -- blockfetch descision interval.
+                , bfcSalt                   = 0
                 }
             }
 
       nodeKernel <- initNodeKernel nodeArgs
       let mempool = getMempool nodeKernel
+      let kaRng = case seed of
+                       (Seed s) -> mkStdGen s
```
mrBliss, line 971:
```suggestion
                       Seed s -> mkStdGen s
```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -215,10 +216,13 @@ run runargs@RunNodeArgs{..} =
...
                         btime
                         chainDB
+                        bfsalt
       nodeKernel <- initNodeKernel nodeArgs
       rnNodeKernelHook registry nodeKernel
 
-      let ntnApps = mkNodeToNodeApps   nodeArgs nodeKernel
+
+      kaRng <- newStdGen
+      let ntnApps = mkNodeToNodeApps   nodeArgs nodeKernel kaRng
```
mrBliss, line 225: Should we make the `kaRng` part of `nodeArgs`? Rename it to `keepAliveRng`.

karknu:
> This branch includes a commit that's in master.

I know, I needed it for testing which was done with a 1.18.0 node.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
karknu: `ScheduledStop m` will be  used by all protocols when the new connection manager is merged. The keepalive protocol was developed on  top of that branch, which is why it already has it.
Do you have `aKeepAliveClient` and  `hKeepAliveClient` mixed up? 
Anyway the root of cause of `hKeepAliveClient` arguments' is `Ouroboros.Network.KeepAlive.keepAliveClient` and moving arguments between `aKeepAliveClient` and `hKeepAliveClient` wont change that.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
karknu: The tip-sample protocol also needs a StdGen.
I think we should add a StdGen to `nodeArgs` and then protocols that need pseudo random numbers can just split it.

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientRegistry.hs
```diff
@@ -156,12 +181,57 @@ bracketSyncWithFetchClient (FetchClientRegistry _ctxVar
...
+
+    -- It is possible for the keepAlive client to keep running even without a fetch client, but
+    -- a fetch client shouldn't run without a keepAlive client.
+    unregister :: m ()
+    unregister = do
+      fetchclient_m <- atomically $ do
+        fetchclients <- readTVar keepRegistry
+        case Map.lookup peer fetchclients of
+             Nothing -> do
+               -- If the fetch client is already dead we remove ourself directly.
```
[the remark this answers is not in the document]
karknu: I don't think "it ourself" is correct, that would imply that we're removing something left by the fetch client. We are removing the PeerGSV which was inserted by the register function.

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientRegistry.hs
```diff
@@ -156,12 +181,57 @@ bracketSyncWithFetchClient (FetchClientRegistry _ctxVar
...
+               modifyTVar dqRegistry $ \m ->
+                 assert (peer `Map.member` m) $
+                 Map.delete peer m
+               return Nothing
+             Just rc -> return $ Just rc
+      case fetchclient_m of
+           Nothing -> return ()
+           Just (tid, doneVar) -> do
+             -- Cancel the fetch client.
+             throwTo tid AsyncCancelled
```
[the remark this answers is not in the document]
karknu: Will change it.

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientRegistry.hs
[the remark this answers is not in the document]
coot: what about `remove PeerGSV ourself`?

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -510,10 +525,26 @@ mkApps kernel Tracers {..} Codecs {..} genChainSyncTimeout Handlers {..} =
...
+      let kacApp = case version of
+                        -- Version 1 doesn't support keep alive protocol but Blockfetch
+                        -- still requires a PeerGSV per peer.
+                        NodeToNodeV_1 -> \_ -> forever (threadDelay 1000) >> return ((), Nothing)
+                        NodeToNodeV_2 -> \_ -> forever (threadDelay 1000) >> return ((), Nothing)
+                        _             -> \dqCtx -> do
+                          runPeerWithLimits
+                            nullTracer
+                            cKeepAliveCodec
+                            (byteLimitsKeepAlive (const 0)) -- TODO: Real Bytelimits, see #1727
```
coot, line 543: I closed #1727 recently, did I jumped ahead?

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
```diff
@@ -8,43 +8,90 @@ module Ouroboros.Network.KeepAlive
...
+      endTime <- getMonotonicTime
+      startTime_m <- atomically $ readTVar startTimeV
+      case startTime_m of
+           Just startTime -> do
+               let rtt = diffTime endTime startTime
+                   sample = fromSample startTime endTime payloadSize
+               gsv' <- atomically $ do
+                   m <- readTVar dqCtx
+                   assert (peer `M.member` m) $ do
+                     let m' = M.adjust (\a -> a <> sample) peer m
```
coot, line 84: You could use [updateLookupWithKey](https://hackage.haskell.org/package/containers-0.6.3.1/docs/Data-Map-Strict.html#v:updateLookupWithKey) to both update the dictionary and get the update value.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
```diff
@@ -8,43 +8,90 @@ module Ouroboros.Network.KeepAlive
...
+    -> StrictTVar m (Maybe Time)
     -> KeepAliveClient m ()
-keepAliveClient shouldStopSTM KeepAliveInterval { keepAliveInterval } =
+keepAliveClient tracer shouldStopSTM peer dqCtx KeepAliveInterval { keepAliveInterval } startTimeV =
     SendMsgKeepAlive go
   where
-    decisionSTM :: TVar m Bool
+    payloadSize = 2
+
+    decisionSTM :: Lazy.TVar m Bool
```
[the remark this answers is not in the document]
coot: I found the answer: it comes from `registerDelay`.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
karknu: Yes, #1727 has not been implemented and the "TODO" is present on all NodeToNode protocols.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
```diff
@@ -8,43 +8,90 @@ module Ouroboros.Network.KeepAlive
...
-      shouldStopSTM
+    decisionSTM delayVar = do
+       shouldStop <- shouldStopSTM
+       case shouldStop of
+            Stop -> return Stop
+            Run  -> do
+              done <- Lazy.readTVar delayVar
+              if done
+                 then return Run
+                 else retry
```
coot, line 71:
```suggestion
              Lazy.readTVar delayVar >>= check
              return Run
```
one could also use `$>`, but it requires importing it first.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
```diff
@@ -8,43 +8,90 @@ module Ouroboros.Network.KeepAlive
...
        , MonadTimer m
+       , Ord peer
        )
-    => ScheduledStop m
+    => Tracer m (TraceKeepAliveClient peer)
+    -> ScheduledStop m
+    -> peer
+    -> (StrictTVar m (M.Map peer PeerGSV))
     -> KeepAliveInterval
+    -> StrictTVar m (Maybe Time)
```
coot, line 53: Do we need to access the above `StrictTVar` outside of the `keepAliveClient`?  If not I think we can remove it. The client can thread `Maybe Time` without using a `StrictTVar`.

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
```diff
@@ -242,13 +242,48 @@ gsvTrailingEdgeArrive (GSV g s v) bytes =
...
+-- sample's interval over the time constant will suffice. The relative error of
+-- this numerical approximation is, for our use case, small. Eg 1/50 (20s
+-- between samples with a 1000s time constant) has a relative error of 1%. The
+-- expected typical range of this relative error is between 5% (ratio of 1/10),
+-- to 0.5% (1/100).
+--
+-- Given the inherent measurement noise in this measurement, the use of the
+-- approximation is well justified. We choose (reaonably aribtarily) 1000s as
+-- the time constant, it is unclear if this should be a configuration variable
+-- or not.
```
coot, line 276: I would add a note that this `Semigroup` is non-commutative.

The new value must come first, which is not the order that is used in the `keepAliveClient`.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
karknu: You're right it isn't needed.

mrBliss:
> > This branch includes a commit that's in master.
> 
> I know, I needed it for testing which was done with a 1.18.0 node.

It makes reviewing this PR harder for me, because that commit touches many consensus files, but I have to ignore these changes. Moreover, I believe it's the cause of the merge conflicts.

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
karknu: Will add it to the comment.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
karknu: I find my version more readable.

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
coot: I think we also need to fix it [here](https://github.com/input-output-hk/ouroboros-network/pull/2251/files#diff-e8529b129ff6f1539758a2ef6947310bR83)

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
```diff
@@ -8,43 +8,90 @@ module Ouroboros.Network.KeepAlive
...
+      case startTime_m of
+           Just startTime -> do
+               let rtt = diffTime endTime startTime
+                   sample = fromSample startTime endTime payloadSize
+               gsv' <- atomically $ do
+                   m <- readTVar dqCtx
+                   assert (peer `M.member` m) $ do
+                     let m' = M.adjust (\a -> a <> sample) peer m
+                     writeTVar dqCtx m'
+                     return $ (M.!) m' peer
```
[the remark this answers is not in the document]
karknu: Addressed by switching to updateLookupWithKey

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
karknu: Dang, you're right!

karknu:
> It makes reviewing this PR harder for me, because that commit touches many consensus files, but I have to ignore these changes. Moreover, I believe it's the cause of the merge conflicts.

The PR is intended to be reviewed one commit at the time, but in you're case I guess you're mostly interested in reviewing the changes at the consensus layer so that doesn't help. I've rebased on top of master and the extra commit is no longer there.

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
coot: :1st_place_medal:

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
coot: I re-opened it, I think I confused it with per-state time limits.

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientRegistry.hs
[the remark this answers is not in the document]
coot: it would be nice to clean this at some point...

ouroboros-network/src/Ouroboros/Network/DeltaQ.hs
coot: I wonder how this affects your tests ...

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -510,10 +524,25 @@ mkApps kernel Tracers {..} Codecs {..} genChainSyncTimeout Handlers {..} =
...
+                            timeLimitsKeepAlive
+                            channel
+                            $ keepAliveClientPeer
+                            $ hKeepAliveClient version (neverStop (Proxy :: Proxy m)) them dqCtx
+                                (KeepAliveInterval 10)
```
mrBliss, line 543: Nitpicking, we prefer to indent with 2 spaces from `case`, not 5.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -911,6 +913,8 @@ runThreadNetwork systemTime ThreadNetworkArgs
                  (ledgerState <$>
                     ChainDB.getCurrentLedger chainDB)
 
+      let kaRng = case seed of
+                       Seed s -> mkStdGen s
```
mrBliss, line 916:
```suggestion
      let kaRng = case seed of
                    Seed s -> mkStdGen s
```
Nitpicking, we prefer to indent with 2 spaces from `case`, not 5.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -54,6 +54,8 @@ import qualified Data.Set as Set
 import qualified Data.Typeable as Typeable
 import           Data.Void (Void)
 import           GHC.Stack
+import           System.Random (mkStdGen)
+import qualified Test.QuickCheck.Exception as QC
```
mrBliss: Is this even used?

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -213,9 +215,12 @@ run runargs@RunNodeArgs{..} =
                         rnTraceConsensus
                         btime
                         chainDB
+                        bfsalt
+                        keepAliveRng
       nodeKernel <- initNodeKernel nodeArgs
       rnNodeKernelHook registry nodeKernel
 
+
```
mrBliss, line 223:
```suggestion

```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -205,6 +205,8 @@ run runargs@RunNodeArgs{..} =
                       (pure $ BackoffDelay 60) -- see 'BackoffDelay'
                       (ledgerState <$>
                          ChainDB.getCurrentLedger chainDB)
+      bfsalt <- randomIO -- Per-node specific value used by blockfetch when ranking peers.
+      keepAliveRng <- newStdGen
```
mrBliss, line 209: Let's create these in `mkNodeArgs` instead of here.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
karknu: I think that `case` is a special case. But I've no desire to fight an indentation war in consensus.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
karknu: No, I suspect it was a missmerge during rebase to master.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
mrBliss: Everybody has his/her own style, the style we follow in consensus is not my own style either (I prefer 4 spaces everywhere, actually).

PR #2252, opened 2020-06-12
Use `SomeBlock` for `Query`

edsko:
This paves the way for the HFC (unlike `Some`, `SomeBlock` is useable inside
`NS`).


edsko:
bors merge

PR #2261, opened 2020-06-15
Move Ouroboros.Consensus.Util.Random to Test.Util.Random

### LINKED ISSUE #2259: Move Ouroboros.Consensus.Util.Random to the testsuite
`Ouroboros.Consensus.Util.Random` is only needed by test code. We should move it to `test-infra` so that nobody accidentally uses it in the implementation code. Quoting Duncan:
> 99% of potential uses of `MonadRandom` in `IO` would be a self-DoS

mrBliss:
Fixes #2259.

`blockProductionIOLike` from `Ouroboros.Consensus.Node.BlockProduction`
depended on it, but as it is only used in the ThreadNet tests, move that
function to `Test.ThreadNet.Util.BlockProduction`.


mrBliss:
bors merge

PR #2265, opened 2020-06-15
VolatileDB: keep NestedCtxt in memory

### LINKED ISSUE #2260: Volatile DB: keep NestedCtxt in memory

mrBliss:
Fixes #2260.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/VolDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/VolDB.hs
@@ -182,6 +182,7 @@ openDB args@VolDbArgs{..} = do
       , maxBlocksPerFile = volBlocksPerFile
       , tracer           = volTracer
       , parser           = blockFileParser args
+      , prefixLen        = reconstructPrefixLen (Proxy @(Header blk))
       }
 
 -- | For testing purposes
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/VolDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/VolDB.hs
@@ -63,6 +63,7 @@ import           Control.Monad (join)
 import           Control.Tracer (Tracer, nullTracer)
 import           Data.Bifunctor (first)
 import qualified Data.ByteString.Lazy as Lazy
+import           Data.ByteString.Short (ShortByteString)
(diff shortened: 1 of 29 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/VolDB.hs
```diff
@@ -497,26 +503,36 @@ blockFileParser VolDbArgs{..} =
...
-      (decodeDisk volCodecConfig)
+      decodeNestedCtxtAndBlock
       volCheckIntegrity
       volValidation
+  where
+    prefixLen :: PrefixLen
+    prefixLen = reconstructPrefixLen (Proxy @(Header blk))
+
+    -- Also extract the prefix from the given bytestring
+    decodeNestedCtxtAndBlock
```
edsko, line 514: This requires a comment if nothing else. I guess here we are in control of that bytestring, and so I _guess_ it really must _just_ be the block? In general though this is certainly not the case for those `ByteString -> ..` decoders, which get their compositionality from the `ByteSpan`s in the decoded value, right?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl.hs
```diff
@@ -249,24 +248,29 @@ putBlockImpl :: forall m blockId. (IOLike m, Ord blockId)
...
       if Map.member bbid currentRevMap then
         lift $ lift $ traceWith tracer $ BlockAlreadyHere bbid
       else do
-        bytesWritten <- lift $ lift $ hPut hasFS currentWriteHandle builder
+        -- The builder will likely consist of a single chunk as the annotation
+        -- of the block is used here, which is strict. Related: #1215.
+        let bytes = BS.toLazyByteString builder
+        bytesWritten <- lift $ lift $ hPutAll hasFS currentWriteHandle bytes
         fileIsFull <- state $ updateStateAfterWrite bytesWritten
+                                                    (takePrefix prefixLen bytes)
```
edsko, line 264: I wonder if we should do this the other way around. Suppose the `Builder` does _not_ correspond to something already in memory. The way you do it now first writes it to disk, forcing the entire thing into memory (because you're holding on to the start), and _then_ gets the first few bytes. if you did it the other way around, you'd only force the first chunk in memory, and GC can start to collect stuff when the file is being written. (Would need to be sure of course that we also force the short bytestring.)

mrBliss:
bors merge

PR #2266, opened 2020-06-15
Use ShortByteString for OneEraHash and ConvertRawHash

mrBliss:
Since the Byron and Shelley hash types are `ShortByteString`s under the hood,
switch over the representation of `OneEraHash` from `Strict.ByteString` to
`ShortByteString` too.

To avoid the intermediary `Strict.ByteString` in the conversion, add
`toShortRawHash` and `fromShortRawHash` to `ConvertRawHash`.

This avoids the memory overhead that a regular strict `ByteString` has.
According to the [Haddocks of `ShortByteString`][SBS], using a
`ByteString` (unshared) for a 32 byte hash (which is the case for Byron and
Shelley) requires 72 + 32 bytes, but a `ShortByteString` requires only 32 + 32
bytes. This can make a noticeable difference in memory usage as we're keeping
lots of hashes in memory:

* O(k) in the in-memory indices of the VolatileDB (hash and prev hash)
* O(cached epochs * epoch size) in the secondary index caches of the
  ImmutableDB
* ...

[SBS]: http://hackage.haskell.org/package/bytestring-0.10.10.0/docs/Data-ByteString-Short.html#g:2

### DIFF
--- a/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Block.hs
+++ b/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Block.hs
@@ -75,10 +75,10 @@ mkByronHash :: CC.ABlockOrBoundaryHdr ByteString -> ByronHash
 mkByronHash = ByronHash . CC.abobHdrHash
 
 instance ConvertRawHash ByronBlock where
-  toRawHash   _ = CC.hashToBytes . unByronHash
-  fromRawHash _ = ByronHash . CC.unsafeHashFromBytes
-  hashSize    _ = fromIntegral $ Crypto.hashDigestSize
-                                   (error "proxy" :: Crypto.Blake2b_256)
+  toShortRawHash   _ = CC.abstractHashToShort . unByronHash
+  fromShortRawHash _ = ByronHash . CC.unsafeAbstractHashFromShort
+  hashSize         _ = fromIntegral $ Crypto.hashDigestSize
+                                        (error "proxy" :: Crypto.Blake2b_256)
 
 {-------------------------------------------------------------------------------
   Block
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Block.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Block.hs
@@ -50,7 +50,7 @@ import           Cardano.Prelude (NoUnexpectedThunks (..))
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.HeaderValidation
 import           Ouroboros.Consensus.Storage.Common (BinaryBlockInfo (..))
-import           Ouroboros.Consensus.Util (hashFromBytesE)
+import           Ouroboros.Consensus.Util (hashFromBytesShortE)
 import           Ouroboros.Consensus.Util.Condense
 
 import qualified Shelley.Spec.Ledger.BlockChain as SL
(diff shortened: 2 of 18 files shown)


ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Block.hs
```diff
@@ -73,6 +73,10 @@ mkByronHash :: CC.ABlockOrBoundaryHdr ByteString -> ByronHash
 mkByronHash = ByronHash . CC.abobHdrHash
 
 instance ConvertRawHash ByronBlock where
+  -- TODO if @cardano-ledger@ exposes @ShortByteString@ variants of
+  -- 'CC.hashToBytes' and 'CC.unsafeHashFromBytes', we can use them to
+  -- override 'toShortRawHash' and 'fromShortRawHash' to avoid going through a
+  -- 'ByteString'.
```
mrBliss, line 79: Byron is already using `ShortByteString`, but it just doesn't expose it. See https://github.com/input-output-hk/cardano-ledger-specs/blob/master/byron/crypto/src/Cardano/Crypto/Hashing.hs

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Block.hs
```diff
@@ -79,6 +79,7 @@ instance Condense (ShelleyHash c) where
   condense = show . unShelleyHash
 
 instance Crypto c => ConvertRawHash (ShelleyBlock c) where
+  -- TODO use 'ShortByteString' for 'Hash'
```
mrBliss, line 82: We should switch the [`Hash` type in `cardano-ledger-specs`](https://github.com/input-output-hk/cardano-base/blob/c388d765566ad5e6141c586425bcd7b4efbfec8a/cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs#L68) to use `ShortByteString`.

ouroboros-consensus/src/Ouroboros/Consensus/Block/Abstract.hs
```diff
@@ -131,25 +133,43 @@ headerPoint = castPoint . blockPoint
...
   hashSize    :: proxy blk -> Word32
 
+  {-# MINIMAL hashSize
+            , (toRawHash | toShortRawHash)
+            , (fromRawHash | fromShortRawHash) #-}
+
 encodeRawHash :: ConvertRawHash blk
               => proxy blk -> HeaderHash blk -> Encoding
-encodeRawHash p = Serialise.encode . toRawHash p
+encodeRawHash p = Serialise.encode . toShortRawHash p
```
mrBliss, line 168: When/if we decide to use `ShortByteString` for our hashes, we should audit consensus for other uses of `toRawHash` and `fromRawHash` that should be replaced with their `ShortByteString` variants.

ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Block.hs
mrBliss: Done in https://github.com/input-output-hk/cardano-ledger-specs/pull/1581.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Block.hs
mrBliss: Done in https://github.com/input-output-hk/cardano-base/pull/144

ouroboros-consensus/src/Ouroboros/Consensus/Block/Abstract.hs
mrBliss: Done

edsko:
Glad to have this major performance penalty gone finally :smirk:

mrBliss:
bors ping

mrBliss:
bors merge

PR #2267, opened 2020-06-15
Base FetchMode on the number of slots behind

### LINKED ISSUE #2028: Return the right FetchMode regardless the expected chain density
Fix this TODO: https://github.com/input-output-hk/ouroboros-network/blob/f9469fa3885aefdfa852216a66c4e319a0d19c11/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs#L268

Plan:
* Bump 5 to 20
* Express it in terms of slots instead of blocks

The API isn't quite right, consensus shouldn't be making the decision here. We can give the network layer the right information, i.e., how many slots we're behind, to base *their* decision on.

mrBliss:
Fixes #2028.

Simply base the current `FetchMode` on the number of slots behind our chain is
behind the current one instead of the number of blocks it is behind. For the
latter we'd need to know the expected density for the expected block type. Use
20 as a value that should be fine for both BFT and Praos.


mrBliss:
bors merge

Issue #2215, opened 2020-06-04
Add more ghc warning options to network packages.


coot:
Consensus is using:
```
                       -Wall
                       -Wcompat
                       -Wincomplete-uni-patterns
                       -Wincomplete-record-updates
                       -Wpartial-fields
                       -Widentities
                       -Wredundant-constraints
                       -Wmissing-export-lists
```

coot:
Done in #2303

PR #2268, opened 2020-06-15
Enable NodeToNodeV_2 and NodeToClientV_3

### LINKED ISSUE #1633: Enable NodeToNodeV_2
> In https://github.com/input-output-hk/ouroboros-network/pull/1632 we prepared consensus to be able to instantiate the chain sync protocol with a sized header. We did however not enable it yet, as we wanted to bundle this with other network protocol changes.
>
> Marking as medium priority because we are now able to enable this at any point without breaking compatibility, and moreover it is an optimization only, enabling the block fetch client to make better decisions.

**UPDATE:** We'll need `NodeToNodeV_2` for two things:

1. `ByronNodeToNodeVersion2`, which enables the block size hints to be sent along Byron headers, used by the ChainSync protocol, as described above.
2. `CardanoNodeToNodeVersion2`, which enables the hard-fork compatible format of the `CardanoBlock`. See https://github.com/input-output-hk/ouroboros-network/blob/f9469fa3885aefdfa852216a66c4e319a0d19c11/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L210

It seems like the network team are also planning to enable the [`MsgKThxBye` message](https://github.com/input-output-hk/ouroboros-network/blob/02ff4eaeb8c4fa47a6b6d99cbd1219f209de3c05/ouroboros-network/src/Ouroboros/Network/Protocol/TxSubmission/Type.hs#L231) with V2. There's also the [upcoming `KeepAlive` protocol](https://github.com/input-output-hk/ouroboros-network/pull/2251/) that will need to be enabled at some point.

Similarly, we'll need `NodeToNodeV_3` for [`CardanoNodeToNodeVersion3`](https://github.com/input-output-hk/ouroboros-network/blob/f9469fa3885aefdfa852216a66c4e319a0d19c11/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L216).

mrBliss:
Fixes #1633.

This is required to let a `cardano-node` send Shelley "things" across the network after hard forking.

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
@@ -207,7 +207,7 @@ instance TPraosCrypto sc => TranslateNetworkProtocolVersion (CardanoBlock sc) wh
 
   nodeToNodeProtocolVersion _ = \case
       CardanoNodeToNodeVersion1 -> NodeToNodeV_1
-      CardanoNodeToNodeVersion2 -> error "NodeToNodeV_2"
+      CardanoNodeToNodeVersion2 -> NodeToNodeV_2
       v                         -> error $ "unsupported version: " <> show v
 
   nodeToClientProtocolVersion _ = \case
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
@@ -213,7 +213,7 @@ instance TPraosCrypto sc => TranslateNetworkProtocolVersion (CardanoBlock sc) wh
   nodeToClientProtocolVersion _ = \case
       CardanoNodeToClientVersion1 -> NodeToClientV_1
       CardanoNodeToClientVersion2 -> NodeToClientV_2
-      CardanoNodeToClientVersion3 -> error "NodeToClientV_3"
+      CardanoNodeToClientVersion3 -> NodeToClientV_3
       v                           -> error $ "unsupported version: " <> show v
 
 {-------------------------------------------------------------------------------
--- a/ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
@@ -26,6 +26,8 @@ data NodeToClientVersion
     = NodeToClientV_1
     | NodeToClientV_2
     -- ^ added local-query mini-protocol
+    | NodeToClientV_3
+    -- ^ enabled @CardanoNodeToClientVersion3@
   deriving (Eq, Ord, Enum, Bounded, Show, Typeable)
 
 -- | We set 16ths bit to distinguish `NodeToNodeVersion` and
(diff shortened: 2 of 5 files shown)


mrBliss:
@coot This doesn't break network compatibility in any way, right? So this could be safely merged?

If you want to include other changes in those versions, like `MsgKThxBye`, you'll have to be quick :slightly_smiling_face:

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -207,13 +207,13 @@ instance TPraosCrypto sc => TranslateNetworkProtocolVersion (CardanoBlock sc) wh
 
   nodeToNodeProtocolVersion _ = \case
       CardanoNodeToNodeVersion1 -> NodeToNodeV_1
-      CardanoNodeToNodeVersion2 -> error "NodeToNodeV_2"
+      CardanoNodeToNodeVersion2 -> NodeToNodeV_1
```
coot, line 210: Is this right? Shouldn't it be set to `NodeToNodeV_2` which is added in this PR.

ouroboros-network/src/Ouroboros/Network/Protocol/TxSubmission/Type.hs
```diff
@@ -228,7 +228,7 @@ instance Protocol (TxSubmission txid tx) where
 
   exclusionLemma_NobodyAndServerHaveAgency TokDone tok = case tok of {}
 
-{-# WARNING MsgKThxBye "MsgKThxBye: should only be used with not yet introduced NodeToNodeV_2" #-}
+{-# WARNING MsgKThxBye "MsgKThxBye: should only be used with not yet introduced NodeToNodeV_3" #-}
```
coot, line 231: I would rather change this to: `MsgKThxBye`: must not be used with `NodeToNodeV_1`. `tx-sumbission` can take care of it for some time now, it's only matter for backward compatibility against a node that was build before introducing it.

mrBliss: Ah, I did not know that. Will update.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
mrBliss: Of course :facepalm:

ouroboros-network/src/Ouroboros/Network/Protocol/TxSubmission/Type.hs
coot: The warning is only left to remind us this when we will start using `MsgKThxBye`.

mrBliss:
bors merge

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
```diff
@@ -40,13 +42,15 @@ nodeToClientVersionCodec = CodecCBORTerm { encodeTerm, decodeTerm }
     where
       encodeTerm NodeToClientV_1 = CBOR.TInt 1
       encodeTerm NodeToClientV_2 = CBOR.TInt (2 `setBit` nodeToClientVersionBit)
+      encodeTerm NodeToClientV_3 = CBOR.TInt (3 `setBit` nodeToClientVersionBit)
```
mrBliss, line 45: @coot bors failed because I had forgotten to add this line. https://github.com/input-output-hk/ouroboros-network/issues/2215 would have prevented this mistake :slightly_smiling_face:

mrBliss:
bors merge

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
coot: I'll try to add the flags this week.

mrBliss: :+1:  (It happened again. Uggh, this is annoying, I'll enable the flags locally so I can fix all errors once and for all.)

mrBliss:
bors merge

Issue #2204, opened 2020-06-04
Chain-Sync ingress limits


coot:
We should use max header size limit which leaves on the chain.  Currently we set limits for the lifetime of a process, and we'd like to change it to set them when connection starts.

@mrBliss we would like to have a way to access the current max header size.  
@dcoutts mentioned that it would be also good check if it is enforced by the
ledger.

mrBliss:
> @mrBliss we would like to have a way to access the current max header size.

Can do. I can add `maxHeaderSize :: LedgerState blk -> Word32` to `LedgerSupportsMempool`. You can get a `LedgerState` using `ChainDB.getCurrentLedger` + `NodeKernel.getChainDB`. Where exactly do you need access to it?

> @dcoutts mentioned that it would be also good check if it is enforced by the
> ledger.

I would expect so (I know for a fact that it is the case for Shelley, i.e., we call [`chainChecks`](https://github.com/input-output-hk/cardano-ledger-specs/blob/59f8c946b231d4d2c01f974ba4ddb76334a5eb7e/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Chain.hs#L226)). Check with the ledger team for Byron.

coot:
We will need to access it when we start a connection; passing `maxHeaderSizeM :: m Word32` through `DiffusionArguments` would work. I think the diffusion is run in `Ouroboros.Consensus.Node`.

mrBliss:
Now that #2275 is merged, you have [`maxHeaderSize :: LedgerState blk -> Word32`](https://github.com/input-output-hk/ouroboros-network/blob/cc4bb6195d08e5519c3428f6617f3e04c565f65b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/CommonProtocolParams.hs#L14) at your disposal.

coot:
Great, thanks.

PR #2275, opened 2020-06-16
Introduce CommonProtocolParams

mrBliss:
Using this class you can query the ledger for the current max
header/transaction size.

This supersedes the `maxTxSize` method of `LedgerSupportsMempool`.

Moreover, we fix a bug in the implementation of `maxTxSize` for `ByronBlock`:
it was returning the max block size instead of the max transaction size!


mrBliss:
@coot This is the consensus side of #2204. When this PR is merged, I'm going to remove the consensus label from that issue.

mrBliss:
bors merge

PR #2278, opened 2020-06-17
Enable NodeToNodeV_2 for Byron

mrBliss:

PR #2281, opened 2020-06-17
Only enable the Byron-to-Shelley hard fork for NodeToNodeV_3

mrBliss:
Now that we enabled `NodeToNodeV_2` for Byron in #2278, we can no longer
enable the hard fork with the same version number, so we need `NodeToNodeV_3`.

Why not? Because if #2278 makes it into a release that is deployed on mainnet
without the release being ready for Byron-to-Shelley, then V2 should not
enable the hard fork (which it did before this commit).


mrBliss:
We decided to go for the other option, which is to revert #2278.

PR #2283, opened 2020-06-18
Revert "Enable NodeToNodeV_2 for Byron"

mrBliss:
This reverts commit 69c39d57f33565f9acf34f18e1640ba40e08a6e9.

If this change ends up being released and deployed before Byron-to-Shelley is
ready, then we'll need a NodeToNodeV_3 to enable Byron-to-Shelley.

See #2281 for the alternative approach.


mrBliss:
bors merge

PR #2286, opened 2020-06-18
coot/trailing bytes 2

coot:
- network-mux: trailing bytes
- ouroboros-network-framework & ouroboros-network: trailing bytes
- ouroboros-consensus: trailing bytes

### DIFF
--- a/network-mux/src/Network/Mux.hs
+++ b/network-mux/src/Network/Mux.hs
@@ -220,7 +220,11 @@ miniProtocolJob tracer egressQueue
         writeTVar miniProtocolStatusVar StatusIdle
         putTMVar completionVar $ Right result
         buf <- readTVar miniProtocolIngressQueue
-        writeTVar miniProtocolIngressQueue $ BL.append remainder buf
+        case remainder of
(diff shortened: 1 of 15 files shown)


network-mux/src/Network/Mux.hs
```diff
@@ -220,7 +220,11 @@ miniProtocolJob tracer egressQueue
         writeTVar miniProtocolStatusVar StatusIdle
         putTMVar completionVar $ Right result
         buf <- readTVar miniProtocolIngressQueue
-        writeTVar miniProtocolIngressQueue $ BL.append remainder buf
+        case remainder of
+          Just trailing ->
+            writeTVar miniProtocolIngressQueue $ BL.append trailing buf
```
karknu, line 225: I think you should move the reading of `buf` to inside the `Just` case.

coot: Yes; even better just use `modifyTVar`, we use strict `TVar`'s so this should be ok.

coot:
bors merge

Issue #2235, opened 2020-06-09
Randomize initial nonce in Praos tests


nfrisby:
In mock `Praos` and `RealTPraos`, the genesis block is currently initialized with the same fixed nonce every time. If I understand correctly, that means the leader schedule for the first epoch is similarly fixed. Especially once we start with statistical testing (eg for CG/CP violation rates), we'll want our samples to be more independent than that.

Thus this Issue is to have the tests' QuickCheck generators determine the initial nonce.

PR #2289, opened 2020-06-18
Shelley/Cardano: take the initial nonce as a parameter

### LINKED ISSUE #2005: Shelley: choose the right initial nonce
Fix this TODO:
https://github.com/input-output-hk/ouroboros-network/blob/eec4a136bb51d882ab4cf4de552f1ce978459f56/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs#L209

mrBliss:
Fixes #2005.

This nonce will be used to construct Shelley's initial `PrtclState`, both when
starting a fresh Shelley chain and when forking from Byron to Shelley (when
translating the Byron `ChainDepState` to the Shelley one).

We store the initial nonce in the `TPraosParams`, which is part of the
`ConsensusConfig` for `TPraos`. We need it here, at run-time, because we need
it when translating the Byron `ChainDepState` to the Shelley one.

`protocolInfoShelley` and `protocolInfoCardano` (as well as
`ProtocolRealTPraos` and `ProtocolCardano`) now take a `Nonce` argument that
will be used as the initial nonce. Typically the `Nonce` passed to these
functions (constructors) should be derived from the hash of the Shelley
Genesis config JSON file. Moreover, this allows us to choose a different
initial nonce for testing purposes, as required for #2235.

NOTE: up until now we have used `SL.NeutralNonce` as the initial nonce for
Shelley. When a different nonce is picked in `cardano-node`, i.e., one derived
from the hash of the Shelley Genesis config JSON file, it would cause a hard
fork.

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano.hs
@@ -142,6 +142,12 @@ data Protocol (m :: * -> *) blk p where
   -- | Run TPraos against the real Shelley ledger
   ProtocolRealTPraos
     :: ShelleyGenesis TPraosStandardCrypto
+    -> Nonce
+       -- ^ The initial nonce, typically derived from the hash of Genesis
+       -- config JSON file.
+       --
+       -- WARNING: chains using different values of this parameter will be
+       -- mutually incompatible.
     -> ProtVer
     -> Natural -- ^ Max major protocol version
     -> Maybe (TPraosLeaderCredentials TPraosStandardCrypto)
@@ -157,6 +163,12 @@ data Protocol (m :: * -> *) blk p where
     -> Maybe PBftLeaderCredentials
        -- Shelley
     -> ShelleyGenesis TPraosStandardCrypto
+    -> Nonce
+       -- ^ The initial nonce for the Shelley era, typically derived from the
+       -- hash of Shelley Genesis config JSON file.
+       --
+       -- WARNING: chains using different values of this parameter will be
+       -- mutually incompatible.
     -> ProtVer -- TODO unify with 'Update.ProtocolVersion' (2 vs 3 numbers)
     -> Natural -- ^ Max major protocol version
     -> Maybe (TPraosLeaderCredentials TPraosStandardCrypto)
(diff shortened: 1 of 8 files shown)


mrBliss:
Quoting https://github.com/input-output-hk/cardano-ledger-specs/pull/1558#issuecomment-645993144:
> @dcoutts decided to use the hash of the genesis JSON file instead. That hash can be computed without needing to run the right version of the code from this repo (e.g., in 10 years, it will still be easy to compute the hash, maybe when this repo is long gone older_man). It is also much easier for users to compute.

I expect this means that we, consensus, can't compute that genesis hash anymore, but that we will receive it from `cardano-node` (it will be a CLI/config parameter, like Byron?) as part of `ProtocolRealTPraos` and `ProtocolCardano` (instead of hiding the ability to override the nonce from the node).

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
```diff
@@ -192,14 +192,19 @@ data TPraosParams = TPraosParams {
     , tpraosMaxLovelaceSupply :: !Word64
       -- | Testnet or mainnet?
     , tpraosNetworkId         :: !SL.Network
+      -- | Initial nonce used for the TPraos protocol state. Typically this is
+      -- derived from the hash of the Shelley genesis config JSON file, but
+      -- different values may be used for testing purposes.
+    , tpraosInitialNonce      :: !SL.Nonce
```
mrBliss, line 198: Mention why we store it here, link to issue #2326.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano.hs
```diff
@@ -157,6 +160,9 @@ data Protocol (m :: * -> *) blk p where
     -> Maybe PBftLeaderCredentials
        -- Shelley
     -> ShelleyGenesis TPraosStandardCrypto
+    -> Nonce
+     -- ^ The initial nonce for the Shelley era, typically derived from the
+     -- hash of Shelley Genesis config JSON file.
```
mrBliss, line 165: Mention the danger of a hard fork here

mrBliss:
bors merge

PR #2291, opened 2020-06-18
More information in evolveKey's error messages

mrBliss:
A lot of people on the testnet are having problems with generating the KES
keys for the right period, see
<https://github.com/input-output-hk/cardano-node/issues/1235>.

Make the error message they get more helpful by including the current (= last)
period of the KES key, and the period we're evolving to.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
@@ -97,17 +97,20 @@ evolveKESKeyIfNecessary updateForgeState (SL.KESPeriod kesPeriod) = do
   where
     -- | Evolve the given key so that its KES period matches @kesPeriod@.
     evolveKey :: HotKey c -> HotKey c
-    evolveKey (HotKey oldPeriod outdatedKey) = go outdatedKey oldPeriod kesPeriod
-      where
-        go !sk c t
-          | t < c
-          = error "Asked to evolve KES key to old period"
-          | c == t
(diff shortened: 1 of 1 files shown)


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -97,17 +97,20 @@ evolveKESKeyIfNecessary updateForgeState (SL.KESPeriod kesPeriod) = do
...
+        = error $
+            "Asked to evolve KES signature key at period " <>
+            show kesPeriodOfKey <> "  to old period " <> show kesPeriod
+        | otherwise
+        = case KES.updateKES () key kesPeriodOfKey of
+            Just key' -> evolveKey (HotKey (kesPeriodOfKey + 1) key')
+            Nothing   -> error $
+              "Could not evolve KES signature key to period " <>
+              show kesPeriod <> ", its current period " <>
+              show kesPeriodOfKey <> " is its last"
```
mrBliss, line 113: An alternative is to throw a new type of exception for this, but the annoying thing is that we'd need to reintroduce `RunNode.nodeExceptionIsFatal` (removed in #2243) so that we can mark this exception as fatal in the `ErrorPolicy`. The goal is to empty out `RunNode` :slightly_frowning_face:.

Hmm, actually, exceptions thrown by `error` are not marked as fatal in the error policy (not our choice). So how come the current `error` is causing the node to shut down (see https://github.com/input-output-hk/cardano-node/issues/1235) :thinking:

mrBliss:
We're going for a better solution, see #2299.

PR #2293, opened 2020-06-18
shelley-test: epoch size of 10k/f not 10k

nfrisby:
The Shelley era should use an epoch size of `10k/f`, not just `10/k`.


ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -188,6 +188,10 @@ mkGenesisConfig pVer k d slotLength maxKESEvolutions coreNodes =
     , sgStaking               = initialStake
     }
   where
+    -- the reciprocol of the active slot coefficient
```
mrBliss, line 191:
```suggestion
    -- the reciprocal of the active slot coefficient
```

PR #2296, opened 2020-06-19
Update comments/names in the HardFork code

nfrisby:
I think the only potentially objectionable content is the "rename defaultSafeZone to noLowerBoundSafeZone" commit. Let me know and I'll drop it.


edsko:
bors merge

Issue #2127, opened 2020-05-21
Catch pure exceptions thrown by the ledger


mrBliss:
If the ledger-side implementation of a pure ledger method (e.g., `applyLedgerBlock`) throws a pure exception (e.g., `Map.!: given key is not an element in the map`), we should catch it instead of crashing the node (or letting the `ErrorPolicy` handle it).

Such a protective boundary should be put into place in the ChainDB.

coot:
@mrBliss I am wondering why did it crashed the node.  Was it the case that the consensus error policy used `Throw` in this case to bring the node down?

mrBliss:
> @mrBliss I am wondering why did it crashed the node. Was it the case that the consensus error policy used `Throw` in this case to bring the node down?

It happened in a background thread of the ChainDB that adds blocks, which *should* cause the node to be shut down.

We have no policy for pure exceptions, as requested by Duncan, so that `error`s in per-connection threads won't shut down the whole node.

coot:
Duncan posted this log in slack
```
[dunky:cardano.node.IpSubscription:Error:52] [2020-05-20 22:18:34.66 UTC] [String "Application Exception: 127.0.0.1:3001 SubscriberError {seType = SubscriberWorkerCancelled, seMessage = \"SubscriptionWorker exiting\", seStack = []}",String "[127.0.0.1:3001,127.0.0.1:3003]",String "WithIPList SubscriptionTrace",String "LocalAddresses {laIpv4 = Just 0.0.0.0:0, laIpv6 = Just [::]:0, laUnix = Nothing}"]
[dunky:cardano.node.IpSubscription:Info:52] [2020-05-20 22:18:34.66 UTC] [String "Closed socket to 127.0.0.1:3001",String "[127.0.0.1:3001,127.0.0.1:3003]",String "WithIPList SubscriptionTrace",String "LocalAddresses {laIpv4 = Just 0.0.0.0:0, laIpv6 = Just [::]:0, laUnix = Nothing}"]
cardano-node: ExceptionInLinkedThread "ThreadId 35" Map.!: given key is not an element in the map
CallStack (from HasCallStack):
  error, called at libraries/containers/Data/Map/Internal.hs:610:17 in containers-0.6.0.1:Data.Map.Internal
```

[...]

Issue #1966, opened 2020-04-17
Catch any exceptions thrown during decoding data from disk


mrBliss:
When any of the databases decode data from disk, they "graciously" handle failure (`fail`, `DeserialiseFailure`, `Left`, etc.). For example, the ImmutableDB and VolatileDB will truncate invalid blocks.

When a decoder throws an impure exception instead of a proper failure (`fail`), the exception is propagated instead of handled as "invalid data". When this happens while serving a request for a connection, the error policy will just restart the connection thread in question. However, when this happens on startup, e.g., during validation or recovery of the databases, the node will crash with that exception because the databases don't even get the chance to handle this invalid data. On the next startup(s), the same will happen unless the user intervenes.

Decoders should never throw impure exceptions, but bugs happen, see https://github.com/input-output-hk/cardano-base/pull/103.

We should catch such exceptions during decoding. Ideally this should be done in a central place, because it would be too easy to forget to do it in some decoder.

PR #2301, opened 2020-06-22
Handle pure exceptions

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/2127 and https://github.com/input-output-hk/ouroboros-network/issues/1966
In this pr we try to catch all unexpected exceptions that can appear, especially while decoding a blockor applying a block to the ledger db (ignoring async exceptions). Ideally of course we would not have any unexpected exceptions, but this can't be avoided since we use many other libraries, which may have partial functions, there are always bugs etc.
There are many reasons we may want to catch such exceptions:
- There is already a handler for specific exceptions, but an unexpected exception may bypass the handler. For example, we already truncate a file when decoding fails. But if there is some `error "..."` in the decoder, the exception will bypass this handler. The same can happen with an unexpected exception while we apply a block to the ledger db. We now make sure that in case of any error proper handling will occur
- An unexpected exception triggered from a client thread may bubble up and kill the whole node. This could happen at the addBlockRunner, which is responsible for adding blocks to the VolatileDB and is linked with the main thread of the node. With this pr any unexpected exception that can appear while adding a new block, will be caught, returned and rethrown to the threads that use the client db api (mainly BLockFetch threads).
- We may finally catch an error so that we provide a proper log with more context, before letting it bubble up. Especially for these unknown errors that we catch here, additional contect can be very helpful for debugging.


kderme:
> The fact that this can happen in multiple places means that the handler should live at a lower level, i.e., in a single place instead of in multiple.

@mrBliss I think an additional handler can be added at `ledgerDbPush`, which is the lower level monadic function. `ledgerDbPushMany` should then stop when a block application fails and `ledgerValidateCandidate` will only return the `ValidatedChainDiff` that we managed to validate, ignoring any faulty blocks and any block after it. So this handler will handle errors both when the chainDB get initiated and when a new block is applied eg by the `BlockFetch`.

With that being said I think the handlers I add here for the `addBlockRunner` still serve a purpose since they can catch more exceptions, like pure exceptions while adding the block to the volatileDB and they help keep the node alive.

Thoughts on this?

abailly-iohk:
@dnadales  This PR is 2.5 years old, should it not be closed?

dnadales:
@kderme I'm sorry once more that yet another PR from you did not got merged. I'm closing this one since it's quite old.

PR #2305, opened 2020-06-23
cardano: fixup counting in rsShelleySlots ThreadNet property

nfrisby:
It's a bit of a corner case, but I caught this bug when running thousands of tests.

As a result, it was requiring Shelley blocks if we _merely reached_ the Shelley era. But if there are just a few slots in the Shelley era, it might get unlucky and have no active slots. This PR corrects it so that it doesn't fail when there are just a few unlucky Shelley slots.

I had previously noticed I wasn't seeing `Optional`s in the `tabulate`, but I thought I just unlucky -- I hadn't run large numbers. With the fix, I now see `Optional`s.


nfrisby:
CI now failing with "unable to allocate memory". I will patiently wait until master advances and I'll rebase onto it.

PR #2308, opened 2020-06-23
Fix building github pages

coot:


.github/workflows/github-page.yml
```diff
@@ -44,6 +44,8 @@ jobs:
       # checkout git repository
       - name: checkout ouroboros-network repository
         run: |
+          mkdir ouroboros-network
+          cd ouroboros-network
```
mrBliss, line 48: This change doesn't match the PR's title at all

coot:
Was fixed in another PR.

PR #2311, opened 2020-06-24
Backport DNS fixes

coot:
- Issue #1893
- Issue #1873

Patches:
- peer-selection: publicRootPeersProvider
- peer-selection: localRootPeersProvider
- peer-selection: Resource
- peer-selection: Use MonadSTM.Strict in RootPeersDNS
- peer-selection: use timeout in RootPeersDNS

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
@@ -31,8 +31,7 @@ import           Data.Map.Strict (Map)
 import           Control.Exception (Exception (..), IOException)
 import           Control.Monad (when, unless)
 import           Control.Monad.Class.MonadAsync
--- TODO: use `MonadSTM.Strict`
-import           Control.Monad.Class.MonadSTM
+import           Control.Monad.Class.MonadSTM.Strict
 import           Control.Monad.Class.MonadTime
 import           Control.Monad.Class.MonadTimer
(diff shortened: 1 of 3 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -50,6 +55,89 @@ data DomainAddress = DomainAddress {
...
+      _ -> do
+        rs <- DNS.makeResolvSeed resolvConf
+        DNS.withResolver rs (pure . constantResource)
+  where
+    go :: FilePath -> TVar IO TimedResolver -> Resource DNS.Resolver
+    go filePath resourceVar = Resource $ do
+      r <- atomically (readTVar resourceVar)
+      case r of
+        NoResolver -> do
+          !modTime <- getModificationTime filePath
```
karknu, line 126: `getModificationTime` can throw `isDoesNotExistError` if the resolvConf file is missing, which will happen on OSX during temporary network failures. How will your code deal with that?

coot: I pushed a fix which catches `IOException`s and `DNSError`s.  Detection of network configuration changes would not workk on Windows, where the network config is not read from a file, but returned by a win32 api call.  I tested this on Linux and Windows. Check the updated branch.

coot:
Requires #2298 to be merged first.

coot:
bors merge

PR #2312, opened 2020-06-24
Fix #2310

### LINKED ISSUE #2310: Ledger view forecast across hard fork boundary is incorrect
Suppose we are near the transition point, and we are constructing a forecast that crosses the transition from some era A to era B. We do this by asking era A for the ledger view for the slot we're trying to construct a forecast for, and then translating that to a ledger view for era B:

```haskell
forecastNotFinal :: forall f blk blk' blks.
                    Translate f blk blk'
                 -> Current (AnnForecast f) blk
                 -> Forecast (HardForkLedgerView_ f (blk ': blk' ': blks))
forecastNotFinal g (Current start AnnForecast{..}) =
    Forecast (forecastAt annForecast) $ \for ->
      translateIf annForecastNext for <$> forecastFor annForecast for
```

That is not correct. We should instead ask era A for it's _final_ ledger view, and then translate _that_ to the final slot. 
This is important, because the safe zone of era A might be shorter than the one in era B, and the current approach means that our forecast range is determined by the safe zone of era A. 

Marking this as medium priority because fixing it is not absolutely essential: it just means that our forecast range is more limited than we'd like it to be. Certainly should be workable for a testnet. However, also assigned it to myself and to the current milestone, thinking about a fix now.

edsko:


nfrisby:
I've confirmed locally on my WIP branch that this patch fixes the repro and moreover the test I'm developing has now passed more than 6000 QC tests involving this patch. LGTM in that sense.

edsko:
Created https://github.com/input-output-hk/ouroboros-network/pull/2312 for constructing tests that target this more directly.

edsko:
bors merge

PR #2315, opened 2020-06-24
handle exceptions while decoding

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/1966

This makes sure that a file is always truncated even in case of some unexpected exception.


abailly-iohk:
@dnadales This PR is 2.5 years old, should it not be closed?

PR #2316, opened 2020-06-25
Set stabilityWindow to 4k/f

nc6:
As per https://github.com/input-output-hk/cardano-ledger-specs/issues/1525


nc6:
Yes, this was what Peter Gazi suggested. Unfortunately it was on a video call so there's not much of a paper trail...

@dcoutts was in the meeting and could also probably confirm :-)

nc6:
Going to close this, since we would probably want to change only the epoch structure, not the window size

PR #2318, opened 2020-06-25
Fix forecasting abstraction, and add forecasting tests.

### LINKED ISSUE #2314: Create standalone tests for the HFC forecast infrastructure
These tests should

* Expose the bug reported in #2310 and fixed in #2312 
* Make essential use of the `SlotNo` parameter in `TranslateForecast`
* Highlight the fact that when we take the "final" forecast of one era to then translate to the next, we need to use the final slot in that era, not the first slot in the next (just like we ensure we never have a Byron ledger state in which the major protocol version is 2).

edsko:
Closes #2314.


mrBliss (approved):
Great, nice to see this finally tested. Good thing you finished them, otherwise we probably wouldn't have found the bug.

edsko:
bors merge

edsko:
bors merge

PR #2327, opened 2020-06-26
Cleanup consensus imports

edsko:
We now consistently only import from `Ouroboros.Consensus.Block` for block
related stuff, including things we used to import from `network` and/or
`cardano-base`.

This also renames `At` to `NotOrigin` (at least within consensus) to avoid the
frequent clashes with our use of `At` in the tests.

We no longer export the crazy `Block` type from `network` but only export the
pattern synonyms; and `RealPoint` doesn't use `Block` at all anymore.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/TestBlock.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/TestBlock.hs
@@ -75,9 +75,6 @@ import           Test.QuickCheck hiding (Result)
 import           Cardano.Crypto.DSIGN
 import           Cardano.Prelude (NoUnexpectedThunks)
 
-import           Ouroboros.Network.Block (ChainHash (..), HeaderHash,
-                     SlotNo (..))
-import qualified Ouroboros.Network.Block as Block
 import           Ouroboros.Network.MockChain.Chain (Chain (..), Point)
 import qualified Ouroboros.Network.MockChain.Chain as Chain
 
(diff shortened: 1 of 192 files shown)


ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/TestBlock.hs
```diff
@@ -167,7 +164,7 @@ instance Condense TestHash where
 
 data TestBlock = TestBlock {
       tbHash  :: TestHash
-    , tbSlot  :: Block.SlotNo
+    , tbSlot  :: SlotNo
```
mrBliss, line 167: :relieved: This has bothered me before.

ouroboros-consensus/src/Ouroboros/Consensus/Block/Abstract.hs
```diff
@@ -26,6 +27,29 @@ module Ouroboros.Consensus.Block.Abstract (
...
+    -- * Re-export basic definitions from @ouroboros-network@
+  , blockMeasure
+  , BlockMeasure -- opaque
+  , blockPoint
+  , castHash
+  , castPoint
+  , ChainHash(..)
+  , HasHeader(..)
+  , HeaderHash
+  , Point(GenesisPoint, BlockPoint)
```
mrBliss, line 39: :+1:

ouroboros-consensus/src/Ouroboros/Consensus/Block/Abstract.hs
```diff
@@ -26,6 +27,29 @@ module Ouroboros.Consensus.Block.Abstract (
...
+  , pointHash
+  , pointSlot
+  , StandardHash
+    -- * Re-export basic definitions from @cardano-base@
```
mrBliss, line 43: Actually, `cardano-slotting`, which lives in the `slotting` dir of `cardano-base` :grin:  (who needs package and folder names to match? :roll_eyes:)

edsko: Let's say it's the repo name :)

edsko:
bors merge

PR #2328, opened 2020-06-26
Avoid needing instance for `GetPrevHash` for both headers and blocks

edsko:
Make `headerPrevHash` the primitive, and `blockPrevHash` derived. This means we only need to give/require the instance for `GetPrevHash` for headers, and don't need to duplicate it for blocks (which is now just a single function `blockPrevHash`).


edsko:
bors merge

PR #2329, opened 2020-06-26
Introduce HeaderFields

edsko:
`HasHeader` now only has a single function:

```haskell
getHeaderFields :: HeaderFields b
```

The other functions are derived. This has two advantages:

* "Forwarding" instances now become significantly simpler (forwarding from a block to a header, or from a derived block to the block it derives from, such as in `DualBlock`)
* We can now construct an `AnchoredFragment` of `HeaderFields`, because `HeaderFields` _itself_ has an `HasHeader` instance.


edsko:
bors merge

Issue #2156, opened 2020-05-27
Get rid of EBBs once and for all


edsko:
This would be quite a large refactoring, but it would simplify a large part of the codebase, and means that going forward we can forget about EBBs entirely.

If we didn't care about preserving history, we could have rewritten the existing chain, rewriting nothing but the prev-hashes of every block, skipping any EBBs. We _do_ want to preserve history, and so we cannot do this, but we can do the next best thing: we can introduce a hashmap mapping the hash of every EBB to its predecessor. This way we can do a local translation (in the `HasHeader` instance for `ByronBlock`) so that if we have blocks `A - EBB - B`, we rewrite the prev-hash of `B` on the fly to be `A` rather than the EBB. Now we can pretend EBBs don't exist, we don't store them, we don't send them, we have no `SlotNo` clashes, we have no `BlockNo` clashes, soooo many things in the consensus layer and storage layer could be made simpler, at the cost of one static value and `HasHeader` instance for a block type that is _anyway_ no longer going to be used.

mrBliss:
Possible plan:

To transition to an EBB-free chain, there will need to be a transition period in which we are both able to serve an EBB-contaminated chain and an EBB-free chain. Old Byron nodes (NodeToNodeV1/2 and NodeToClientV2) still expect EBBs and that's what they'll get. However, new Byron-to-Shelley nodes (NodeToNodeV3 and NodeToClient3) should not receive EBBs anymore. Once we have switched to Shelley, we will know that only nodes that are capable of handling an EBB-free chain are active, after which we can start making more invasive changes to get rid of EBBs once and for all. E.g., migrate the database, remove all EBB-related edge cases, etc.

This means that new Byron-to-Shelley nodes should both be capable of serving an EBB-contaminated chain (to remain compatible with old Byron nodes) and serving an EBB-free chain (to new Byron-to-Shelley nodes). The ChainDB will need to provide EBB-contaminated Iterators and Readers as well as EBB-free ones. The latter can be implemented on top of the former as it knows which blocks are EBBs (it can use  the `GetIsEBB` block component internally).

Something else we should check: there is a staging net, which might also contain EBBs. To keep that working, we'd need to include its EBB mapping in our hard-coded one. Alternatively, they can rewrite it to be EBB-free :grin:.

If we want to use the Shelley release as the opportunity to start this process, we should do the following before the Shelley release:

* Hard-code the EBB mapping as discussed in the ticket description
[...]

PR #2335, opened 2020-06-27
Provide context in `GetPrevHash`

edsko:
Critically, this provides all the plumbing so that in the end we can have

```haskell
-- | Translate from a raw Byron prev-hash to 'ChainHash'
--
-- This takes the 'CodecConfig' so that we can "translate away" EBBs (#2156).
fromByronPrevHash :: CodecConfig ByronBlock
                  -> Maybe CC.HeaderHash
                  -> ChainHash ByronBlock
```

(even if we don't yet take advantage of this).

### DIFF
--- a/cardano-client/src/Cardano/Client/Subscription.hs
+++ b/cardano-client/src/Cardano/Client/Subscription.hs
@@ -42,8 +42,8 @@ import           Ouroboros.Network.Protocol.Handshake.Version (DictVersion,
                      Versions, foldMapVersions)
 import qualified Ouroboros.Network.Snocket as Snocket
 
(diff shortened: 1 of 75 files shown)


edsko:
Note that using this _may_ require a patch to the Byron ledger implementation; see https://github.com/input-output-hk/ouroboros-network/issues/2156#issuecomment-650534213 .

edsko:
https://github.com/input-output-hk/ouroboros-network/pull/2336 shows a a proof of concept slightly alternative implementation. I prefer that one, but haven't pushed it fully through yet. That PR will probably obsolete this one.

edsko:
Obsoleted by #2336.

Issue #2292, opened 2020-06-18
Test hard fork combinator near rollbacks


nfrisby:
From https://github.com/input-output-hk/ouroboros-network/pull/2066#pullrequestreview-430336258

> Modify that tests (or set up a second one) so that we are making sure we are testing what happens near boundary bounds. Rollback across transition points, ledger forecast across transition points, that sort of thing. This will require some careful tweaking of the leadership schedule I think in order to get precisely the rollbacks that we expect. This might be a bit awkward to set up but it's important, because it's the only way we can really test what happens at those transition boundaries.

As of PR #2277, `ouroboros-consensus-cardano:test:test` tests a reasonably realistic HFC.

This Issue is to make that test more strenuous by ensuring rollback etc near the transition point.

PR #2337, opened 2020-06-29
network: also make the ChainSync CanAwait timeout configurable

nfrisby:
This PR prepares for #2292. My WIP branch for that induces rollbacks by enforcing short network partitions by delaying `MsgRollForward` messages for several slots. Since that message is a valid response in the `CanAwait` state, the tests will also need to relax/de-activate that timeout. This PR doesn't change the timeout. It only exposes it as a parameter.


nfrisby:
bors merge

Issue #2210, opened 2020-06-04
TipSample protocol


coot:
Implement `WarmPeer` protocol for sampling tip information from a warm peer.

PR #2340, opened 2020-06-29
TipSample mini-protocol

coot:


coot:
Related to #2210

ouroboros-network/src/Ouroboros/Network/Protocol/TipSample/Type.hs
```diff
@@ -0,0 +1,80 @@
...
+      -- | Server response with its current tip.
+      --
+      MsgCurrentTip :: tip -> Message (TipSample tip slotNo)
+                                       StTipRequest StIdle
+
+
+      -- | Get first tip after given slot.  It can block.
+      --
+      MsgGetTipAfterSlotNo :: slotNo -> Message (TipSample tip slotNo)
+                                                StIdle (StBusy BlockUntilSlot)
```
coot, line 49: `MsgGetTipAfterSlotNo` does not see forks, it just waits until we pass a given `SlotNo`:  it can happen that a server replies and before even it gets the next request, a fork happens and now it will block until the next given slot is passed.  I think this is ok. The measurements will include how well the node is placed in the network (if a node needs to follow less forks to get to a given slot, it have a greater chance to serve us the final block slightly earlier).

ouroboros-network/src/Ouroboros/Network/Protocol/TipSample/Type.hs
```diff
@@ -0,0 +1,80 @@
...
+    BlockUntilTipChange :: TipRequestKind
+
+data TokTipRequest (tipRequestKind :: TipRequestKind) where
+    TokBlockUntilSlot      :: TokTipRequest BlockUntilSlot
+    TokBlockUntilTipChange :: TokTipRequest BlockUntilTipChange
+
+-- | `tip-sample` protocols, desigined for established peers to sample tip from
+-- upstream peers.
+--
+data TipSample tip slotNo where
```
coot, line 24: ^ `slotNo` can be a concrete type (`SlotNo`).

ouroboros-network/src/Ouroboros/Network/Protocol/TipSample/Type.hs
```diff
@@ -0,0 +1,100 @@
...
+import Network.TypedProtocol.Pipelined (N (..), Nat)
+import Cardano.Slotting.Slot (SlotNo)
+
+import Ouroboros.Network.Util.ShowProxy
+
+
+-- | There are three of blocking requests: awiat until slot, await until
+-- a tip will change; the third one has a dedicated type:
+-- @'StFollowTip' :: 'TipSample' tip@.
+--
```
karknu, line 22: The comment doesn't match the current code. There is only one request, MsgFollowTip.
I don't think the types `TipRequestKind` and `TokTipRequest` are needed anymore.

ouroboros-network/src/Ouroboros/Network/Protocol/TipSample/Codec.hs
```diff
@@ -0,0 +1,168 @@
...
+  where
+    encode :: forall (pr  :: PeerRole)
+                     (st  :: TipSample tip)
+                     (st' :: TipSample tip).
+              PeerHasAgency pr st
+           -> Message (TipSample tip) st st'
+           -> CBOR.Encoding
+    encode (ClientAgency TokIdle) (MsgFollowTip n slotNo) =
+      encodeListLen 2
+        <> encodeWord 1
```
karknu, line 57: Why do you start encoding at 1 instead of at 0?

ouroboros-network/src/Ouroboros/Network/Protocol/TipSample/Server.hs
```diff
@@ -0,0 +1,102 @@
...
+    --
+    runTipSampleServer :: m (ServerStIdle tip m a)
+  }
+
+-- | In 'StIdle' protocol state the server does not have agency, it is await for
+-- one of:
+--
+-- * 'MsgGetTipAfterSlotNo'
+-- * 'MsgGetTipAfterTip'
+-- * 'MsgDone'
```
karknu, line 40: Messages to wait for are `MsgFollowTip` and `MsgDone`.

ouroboros-network/src/Ouroboros/Network/Protocol/TipSample/Codec.hs
coot: Probably a left over; I'll fix this to start at `0`.

coot:
bors merge

PR #2343, opened 2020-06-30
Add analysis that shows EBBs and predecessors

edsko:


ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -185,6 +191,7 @@ instance TPraosCrypto c => HasAnalysis (CardanoBlock c) where
   blockTxSizes blk = case blk of
     Cardano.BlockByron b    -> blockTxSizes b
     Cardano.BlockShelley sh -> blockTxSizes sh
+  knownEBBs = undefined
```
kderme, line 194: @edsko I guess I should also fix this?

edsko: Oh, sorry, I forgot about that. I can fix it if you want, or you can give it a go? It would only require a type translation/injection.

edsko: Whichever you prefer, let me know.

PR #2348, opened 2020-06-30
Shelley: prefer self-produced blocks during chain selection

### LINKED ISSUE #1286: Make sure that we prefer own blocks over incoming blocks
When an upstream node and we ourselves are both elected leader, in the (admittedly unlikely) scenario that we receive the upstream block _before_ we manage to produce our own, I think the current chain DB does not guarantee that we will prefer our own block -- but it should; this is a clear violation of the Ouroboros spec.

mrBliss:
Fixes #1286.

* Include `SelfIssued` in `TPraosChainSelectView`, use that to prefer the self
  issued tip in case the tips are produced in the same slot.

* Store the block issuer verification key (new, strict `BlockIssuerVKey` data
  type) in the `BlockConfig` so that we have access to it when we need to
  extract the 'SelectView' for a header.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
@@ -1,10 +1,12 @@
+{-# LANGUAGE DataKinds      #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric  #-}
 {-# LANGUAGE DerivingVia    #-}
 {-# LANGUAGE TypeFamilies   #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 module Ouroboros.Consensus.Shelley.Ledger.Config (
-    BlockConfig (..)
+    BlockIssuerVKey (..)
+  , BlockConfig (..)
   , mkShelleyBlockConfig
   , CodecConfig (..)
   ) where
@@ -20,6 +22,7 @@ import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.BlockchainTime
 
 import qualified Shelley.Spec.Ledger.Genesis as SL
+import qualified Shelley.Spec.Ledger.Keys as SL
 import qualified Shelley.Spec.Ledger.PParams as SL (ProtVer)
 
 import           Ouroboros.Consensus.Shelley.Ledger.Block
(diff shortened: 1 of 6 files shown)


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
```diff
@@ -293,26 +296,61 @@ data instance ConsensusConfig (TPraos c) = TPraosConfig {
...
---   - All chains are considered equally preferable
-data TPraosChainSelectView c = ChainSelectView {
+-- We order between chains as follows:
+--
+-- 1. By chain length, with longer chains always preferred.
+-- 2. If the tip of each chain has the same slot number, we prefer the one tip
+--    that we produced ourselves.
+-- 3. If the tip of each chain was issued by the same agent, then we prefer
+--    the chain whose tip has the highest ocert issue number.
+-- 4. By the leader value of the chain tip, with lower values preferred.
```
nc6, line 323: If the tips are produced by the same agent, then I would expect them to have identical leader values in the same slot. So I guess this change is to prefer chains with tips in different slots?

mrBliss: Ah, because the VRF doesn't change when a new ocert is issued.

So the previous order was correct, but the new one is also correct, as it doesn't matter in the end. I find this order a bit easier to understand, though: as the last tie breaker, use the leader value, all other comparisons are more important.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
```diff
@@ -28,23 +31,36 @@ import           Ouroboros.Consensus.Shelley.Ledger.Block
...
+  deriving anyclass (NoUnexpectedThunks)
+
 data instance BlockConfig (ShelleyBlock c) = ShelleyConfig {
       -- | The highest protocol version this node supports. It will be stored
       -- the headers of produced blocks.
       shelleyProtocolVersion :: !SL.ProtVer
     , shelleySystemStart     :: !SystemStart
     , shelleyNetworkMagic    :: !NetworkMagic
     , shelleyProtocolMagicId :: !ProtocolMagicId
+    , shelleyBlockIssuerVKey :: !(BlockIssuerVKey c)
```
mrBliss, line 48: Comment: When chain selection is comparing two fragments, it will prefer the fragment with a tip signed by this key (provided that the `BlockNo`s and `SlotNo`s of the two tips are equal). For core nodes, this should be set to the verification key corresponding to the node's signing key, to make sure we prefer self-issued blocks. For non core codes, this can be set to `NotABlockIssuer`.

mrBliss:
bors merge

PR #2349, opened 2020-07-01
Introduce queries about an era that can be answered in any era

### LINKED ISSUE #2346: Byron: query the transition point
Add a query to the Byron ledger that returns the epoch of the Shelley [transition](https://github.com/input-output-hk/ouroboros-network/blob/595a4c2b3385b5dac214ba74501d1980e6d14bd6/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs#L130), when known (`Maybe EpochNo`).

Eventually, this should become a query of the HFC, as we'll want such a query for each transition.

mrBliss:
Fixes #2346.

* Split `HardForkQuery` up into:
  1. `QueryIfCurrent`: queries about a certain era that only can be answered
     when the current ledger is in the respective era.
  2. `QueryAnytime`: queries about a certain era that can be answered when the
     current ledger is in *any* era. No queries about the first era can be
     asked to keep the HFC with a single ledger isomorphic to just the ledger.

     An example query is `EraStart`, which returns the start epoch/time/slot
     of an era.

* The pattern synonyms for `CardanoQuery` were renamed from `QueryByron` and
  `QueryShelley` to `QueryIfCurrentByron`, `QueryIfCurrentShelley`, and
  `QueryAnytimeShelley`.

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs
@@ -44,7 +44,7 @@ module Ouroboros.Consensus.Cardano.Block (
   , OneEraTipInfo (TipInfoByron, TipInfoShelley)
     -- * Query
   , CardanoQuery
-  , Query (QueryByron, QueryShelley)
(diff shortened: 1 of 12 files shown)


mrBliss:
bors merge

PR #2352, opened 2020-07-01
Introduce tickChainDepState

edsko:


ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Protocol.hs
```diff
@@ -236,14 +283,14 @@ rewind k p =
...
-       -> HardForkChainDepState xs
+       -> Ticked (HardForkLedgerView xs)
+       -> Ticked (HardForkChainDepState xs)
        -> Except (HardForkValidationErr xs) (HardForkChainDepState xs)
 update cfg@HardForkConsensusConfig{..}
-       (Ticked slot ledgerView)
        (OneEraValidateView view)
-       chainDepState =
+       (Ticked slot ledgerView)
+       (Ticked _slot chainDepState) =
```
edsko, line 293: We should not be aligning here. Instead to the same as in `check`.

mrBliss: Done

edsko:
bors merge

PR #2362, opened 2020-07-07
Rewrite the golden tests

### LINKED ISSUE #2130: Create golden tests for CardanoBlock
At the moment, we have golden tests for Byron, and for Shelley, but really we don't care about either of other ledgers, and instead want to guarantee compatibility only for the One True Chain, which will be `Cardano` block, currently equal to `ByronBlock`, but soon equal to `HardForkBlock '[ByronBlock, ShelleyBlock]` (which then should indeed be compatible, where applicable, to `ByronBlock`).

mrBliss:
Closes #2130.

Rewrite the golden tests to be file based (yet again, after #1855). This time,
we use hard-coded paths relative to the root of the project. This makes
updating the results much easier than before. Previously, you had to run the
test with an empty expected list and then copy-paste the new output. Now, you
can just delete the file and run the test again, no more copy-pasting needed.

We follow a template-based approach like we did for the roundtrip tests: for
each ledger, we now only have to construct a value of the `Examples` record
and feed that to the test runner instead of having to redefine tons of tests
for each ledger.

Moreover, we now exhaustively test *all* serialisation format, not just the
default encoders, but all *versions* (`NodeToNode` and `NodeToClient`) as
well.


ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
```diff
@@ -132,6 +202,13 @@ exampleApplyTxErr =
     $ STS.UtxowFailure
     $ STS.InvalidWitnessesUTXOW [SL.VKey 1]
 
+exampleAnnTip :: AnnTip (Block ShortHash)
+exampleAnnTip = AnnTip {
+      annTipSlotNo  = SlotNo 0
+    , annTipBlockNo = BlockNo 0
```
mrBliss, line 208: Use other values than 0

ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Examples.hs
```diff
@@ -53,10 +70,80 @@ secParam = SecurityParam 2
...
+exampleHeader :: Header ByronBlock
+exampleHeader = getHeader exampleBlock
+
+exampleEBBHeader :: Header ByronBlock
+exampleEBBHeader = getHeader exampleEBB
+
+exampleAnnTip :: AnnTip ByronBlock
+exampleAnnTip = AnnTip {
+      annTipSlotNo  = SlotNo 0
+    , annTipBlockNo = BlockNo 0
```
mrBliss, line 143: Non-zero values

ouroboros-consensus-cardano/test/Test/Consensus/Cardano/Examples.hs
```diff
@@ -0,0 +1,344 @@
...
+    , exampleQueryAndResult = combineEras (Proxy @SomeResult)        (coerce Golden.exampleQueryAndResult) someResultByron     someResultShelley
+    , exampleAnnTip         = combineEras (Proxy @AnnTip)                    Golden.exampleAnnTip          annTipByron         annTipShelley
+    , exampleLedgerState    = combineEras (Proxy @LedgerState)               Golden.exampleLedgerState     ledgerStateByron    ledgerStateShelley
+    , exampleChainDepState  = combineEras (Proxy @WrapChainDepState) (coerce Golden.exampleChainDepState)  chainDepStateByron  chainDepStateShelley
+    , exampleExtLedgerState = combineEras (Proxy @ExtLedgerState)            Golden.exampleExtLedgerState  extLedgerStateByron extLedgerStateShelley
+    }
+
+-- | Extend the examples from all the individual eras with multi-era examples,
+-- e.g., applying a transaction to the wrong era.
+examples :: Golden.Examples (CardanoBlock Crypto)
```
mrBliss, line 251: Use `[]` for `Examples` so we can provide a `Monoid` instance and use it here.

ouroboros-consensus-cardano/test/Test/Consensus/Cardano/Golden.hs
```diff
@@ -0,0 +1,34 @@
...
+
+tests :: TestTree
+tests = goldenTest_all codecConfig goldenDir examples
+  where
+    goldenDir = "ouroboros-consensus-cardano" </> "test" </> "golden"
+
+instance ToDirectory (HardForkNodeToNodeVersion (CardanoEras sc)) where
+  toDirectory v = case v of
+    CardanoNodeToNodeVersion1 -> "CardanoNodeToNodeVersion1"
+    CardanoNodeToNodeVersion2 -> "CardanoNodeToNodeVersion2"
```
mrBliss, line 26: Check whether we don't have all combinations now

mrBliss: We don't, for example `HardForkNodeToNodeDisabled (WrapNodeToNodeVersion ByronNodeToNodeVersion2)` is not a sensible combination.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/Serialisation/Golden.hs
```diff
@@ -0,0 +1,440 @@
...
+      -> ExceptT CBOR.DeserialiseFailure (ST.Lazy.ST s) FlatTerm
+    provideInput bs
+      | Strict.null bs = return []
+      | otherwise      = do
+          next <- lift $ ST.Lazy.strictToLazyST $ do
+              -- This will always be a 'Partial' here because decodeTermToken
+              -- always starts by requesting initial input. Only decoders that
+              -- fail or return a value without looking at their input can give
+              -- a different initial result.
+              CBOR.Partial k <- CBOR.deserialiseIncremental CBOR.decodeTermToken
```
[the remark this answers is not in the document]
mrBliss: Ah, this is because `ST.Lazy.ST` implements `MonadFail`.

ouroboros-consensus-cardano/test/Test/Consensus/Cardano/Examples.hs
```diff
@@ -67,37 +69,80 @@ import           Test.Consensus.Cardano.Generators (toTelescope')
...
-    aux inj exs = K (coerce . unK . apFn inj <$> getExample exs)
+    eraInjections :: NP (Examples -.-> K (Examples (CardanoBlock Crypto)))
+                        (CardanoEras Crypto)
+    eraInjections =
+           fn (K . injExamplesByron)
+        :* fn (K . injExamplesShelley)
+        :* Nil
+
+-- Generalise in #2368
+injExamplesByron :: Examples ByronBlock -> Examples (CardanoBlock Crypto)
```
mrBliss, line 90: I now have two such injections functions instead of one generic one, because these two are much simpler (and not too wide :grin:). When we have #2368 we could replace them again by a single generic one.

I still use `combineEras`, so adding an era will break the tests.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/Serialisation/Golden.hs
```diff
@@ -12,7 +13,12 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 module Test.Util.Serialisation.Golden (
     Examples (..)
-  , ToDirectory (..)
+  , emptyExamples
```
mrBliss, line 16: No need to export it

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/Util/Serialisation/Golden.hs
```diff
@@ -396,45 +436,49 @@ goldenTest_SerialiseNodeToClient codecConfig goldenDir Examples {..} =
...
--- unless the list is empty, then print "[]"
-pprintList :: Show a => [a] -> String
-pprintList [] = "[]"
-pprintList (x:ys) = intercalate "\n" $
-    [ "[ " <> show x ] <>
-    [ ", " <> show y | y <- ys ] <>
-    [ "]" ]
+-- Handles CBOR-in-CBOR.
+showFlatTermDiff :: FlatTerm -> FlatTerm -> String
+showFlatTermDiff a b
```
mrBliss, line 476: Show what a HFC header looks like

ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Examples.hs
```diff
@@ -131,16 +137,30 @@ exampleBlock =
 exampleEBB :: ByronBlock
 exampleEBB = forgeEBB cfg (SlotNo 0) (BlockNo 0) GenesisHash
 
+exampleSerialisedBlock :: Serialised ByronBlock
+exampleSerialisedBlock = Serialised "\130\SOH<BLOCK>"
+
+exampleSerialisedEBB :: Serialised ByronBlock
+exampleSerialisedEBB = Serialised "\130\NUL<EBB>"
```
mrBliss, line 144: Do we actually need those bytes there? If so, add a comment

ouroboros-consensus-cardano/test/Test/Consensus/Cardano/Examples.hs
```diff
@@ -128,12 +173,46 @@ codecConfig = CardanoCodecConfig Byron.codecConfig Shelley.codecConfig
   Additional injections
 -------------------------------------------------------------------------------}
 
+serialisedBlockByron :: Serialised ByronBlock -> Serialised (CardanoBlock Crypto)
+serialisedBlockByron (Serialised bytes) = Serialised bytes
+
+serialisedBlockShelley ::
+     Serialised (ShelleyBlock Crypto)
+  -> Serialised (CardanoBlock Crypto)
+serialisedBlockShelley (Serialised bytes) = Serialised ("\130\STX" <> bytes)
```
mrBliss, line 182: Don't prepend bytes, comment why

mrBliss:
bors merge

PR #2363, opened 2020-07-07
Configurable BlockFetch parameters

karknu:


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -145,6 +146,12 @@ data RunNodeArgs blk = RunNodeArgs {
       --
       -- Use 'defaultClockSkew' when unsure.
     , rnMaxClockSkew :: ClockSkew
+
+      -- | Maximum concurrent downloads during bulk syncing.
+    , rnMaxConcurrencyBulkSyn :: Word
```
mrBliss, line 151:
```suggestion
    , rnMaxConcurrencyBulkSync :: Word
```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -145,6 +146,12 @@ data RunNodeArgs blk = RunNodeArgs {
...
+
+      -- | Maximum requests in flight during bulk syncing.
```
mrBliss, line 153: Comment is incorrect

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -145,6 +146,12 @@ data RunNodeArgs blk = RunNodeArgs {
...
+    , rnMaxConcurrencyDeadline :: Word
```
mrBliss, line 154: What about `rnCustomMiniprotocolParameters :: MiniProtocolParameters -> MiniProtocolParameters`. This approach allows us to define the defaults in this module instead of deferring that responsibility to `cardano-node` while allowing them to be overridden in `cardano-node`.

karknu: `MiniProtocolParameters` represents hard limits and are not really configurable. I'm not touching them in this PR.

mrBliss: Ah, I just realised that `BlockBlockFetchConfiguration` is not a subset of `MiniProtocolParameters`.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -145,6 +146,12 @@ data RunNodeArgs blk = RunNodeArgs {
       --
       -- Use 'defaultClockSkew' when unsure.
     , rnMaxClockSkew :: ClockSkew
+
```
mrBliss, line 149: The principle I originally had in mind when we introduced `RunNodeArgs` is to use these fields for parameters for which we *can't* pick default values, e.g., `NetworkMagic`, `ProtocolInfo`, etc. For `NodeArgs` and `ChainDbArgs`, we can, so we provide functions to customise them without having to force the user (i.e., `cardano-node`) to pick the defaults (we know better which defaults to pick).

Over the time `RunNodeArgs` has been changed by multiple people and the principle no longer holds. (I want to refactor/clean up this mess at some point).

That's the reason why I prefer a "customise" function instead of requiring concrete values.

mrBliss:
bors merge

PR #2366, opened 2020-07-07
ChainDB.StateMachine: relax precondition on closing

### LINKED ISSUE #1533: Chain DB q-s-m: relax precondition on reopening

mrBliss:
Fixes #1533.

This precondition was no longer necessary after #2228, specifically 13743aed.

### DIFF
--- a/ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
+++ b/ouroboros-consensus/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
@@ -944,13 +944,6 @@ precondition Model {..} (At cmd) =
      -- with iterators that the model allows. So we only test a subset of the
      -- functionality, which does not include error paths.
      Stream from to           -> isValidIterator from to
-     -- Make sure we don't close (and reopen) when there are other chains
-     -- equally preferable or even more preferable (which we can't switch to
-     -- because they fork back more than @k@) than the current chain in the
-     -- ChainDB. We might pick another one than the current one when
-     -- reopening, which would bring us out of sync with the model, for which
-     -- reopening is a no-op (no chain selection). See #1533.
-     Close                    -> Not equallyOrMorePreferableFork
      Reopen                   -> Not $ Boolean (Model.isOpen dbModel)
      -- To be in the future, @blockSlot blk@ must be greater than @slot@.
      --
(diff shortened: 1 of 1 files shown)


mrBliss:
bors merge

PR #2373, opened 2020-07-09
Increase bfcMaxConcurrencyDeadline

karknu:
Inrease bfcMaxConcurrencyDeadline from one to four.
This lowers the risk of getting stuck fetching a block
from a slow peer in case of slot collision.


mrBliss:
It would be even better if the (default) parameters for all network-related things were all defined in the network code, not in consensus code.

karknu:
> It would be even better if the (default) parameters for all network-related things were all defined in the network code, not in consensus code.

I'm trying really hard to fix this issue with out changing anything related to how configuration is handled.

mrBliss:
> > It would be even better if the (default) parameters for all network-related things were all defined in the network code, not in consensus code.
> 
> I'm trying really hard to fix this issue with out changing anything related to how configuration is handled.

I understand that, but I can still dream, can't I? :slightly_smiling_face:.

If all network-related things were defined in the network code, you wouldn't even have to ask consensus for a review :slightly_smiling_face:.

karknu:
Superseded by #2363

PR #2374, opened 2020-07-09
Add the fake initial snapshot to the snapshots.

nc6:
This is needed to allow the initial stake pools to continue producing
blocks into the second epoch.

HERE BE DRAGONS: this function is not suitable for production use!

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
@@ -253,6 +253,10 @@ protocolInfoShelley genesis initialNonce maxMajorPV protVer mbCredentials =
     --
     -- This function embodies a little more logic than ideal. We might want to
     -- move it into `cardano-ledger-specs.`
+    --
+    -- HERE BE DRAGONS! This function is intended to help in testing. It should
+    -- not be called with anything other than 'emptyGenesisStaking' in
+    -- production.
     registerGenesisStaking :: SL.ChainState c -> SL.ChainState c
     registerGenesisStaking cs@(SL.ChainState {chainNes = oldChainNes} ) = cs
         { SL.chainNes = newChainNes }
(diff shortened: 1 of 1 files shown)


mrBliss:
> HERE BE DRAGONS: this function is not suitable for production use!

But now it *will be* used by any pure Shelley node. Or is that not "production"?

What are the consequences of using it?

The warning should really live in the code, not just in this commit message.

nc6:
It will not be used unless `registerGenesisStaking` is invoked. But you're right, I should add this comment onto that function!

mrBliss:
> It will not be used unless `registerGenesisStaking` is invoked. But you're right, I should add this comment onto that function!

But `registerGenesisStaking` *is currently* called in `protocolInfoShelley`.

nc6:
Hmm, true. OK, I have added a comment that explains the use, but I think a better solution would be to make this optional in `ShelleyGenesis` and then avoid doing any manipulation. I've added an issue at https://github.com/input-output-hk/cardano-ledger-specs/issues/1633

deepfire:
Ohh, but then pool block creation stops when the second epoch starts...

deepfire:
Works just fine now!

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
```diff
@@ -253,6 +253,10 @@ protocolInfoShelley genesis initialNonce maxMajorPV protVer mbCredentials =
     --
     -- This function embodies a little more logic than ideal. We might want to
     -- move it into `cardano-ledger-specs.`
+    --
+    -- HERE BE DRAGONS! This function is intended to help in testing. It should
+    -- not be called with anything other than 'emptyGenesisStaking' in
+    -- production.
```
mrBliss, line 259: So production /= Shelley testnets.

nc6:
bors retry

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
nc6: On shelley testnets it should be called with `emptyGenesisStaking`

PR #2377, opened 2020-07-09
Fix problems with paths to golden folders

mrBliss:
The Nix CI was not finding them, creating the golden results each time.
Moreover, `cabal test` was making the directories relative to each subpackage
whereas `cabal run` was making the directories relative to the root.

Copy a solution found by our friends from cardano-wallet.


mrBliss:
bors merge

mrBliss:
bors merge

PR #2378, opened 2020-07-09
shelley-test: allow d=0 in ThreadNet tests generator

nfrisby:


edsko:
bors merge

nfrisby:
FYI, this PR may be to blame for the failure https://hydra.iohk.io/build/3393090/nixlog/1. That build was for a Draft PR #2379 that at the time extended this PR's commits, but it's not obvious to me which commits are to blame for the failure.

I anticipate it's this PR. I'll investigate today.

PR #2379, opened 2020-07-09
Update the d protocol parameter during RealTPraos ThreadNet tests

### LINKED ISSUE #1770: Shelley: protocol testing with lowering `d`
@mrBliss 's current integration work and tests a fixed value for `d` (#1695). The Shelley testnet will want to start reducing `d` fairly quickly (after all, a major reason for the testnet is to test staking). It is therefore important that we test soon that we can lower this parameter. Doing so is however non-trivial as the ledger rules impose restrictions on precisely when this can happen, and so the test infrastructure will need to make sure that these restrictions are satisfied before submitting updates to change `d`.

Somewhat ironically, this is a lower priority for the mainnet, as there it will probably take longer before we start reducing `d`.

nfrisby:
Fixes #1770.

Opening as Draft PR to gather input before refactoring/tidying. It seems to be working. A few specific questions:

  1) ~Any advice on avoiding the limit of `MockKES 10`? It seems like the `RealTPraos` test always fails with a KES-related `CannotLead` if it runs for more than 200 slots.~ @mrBliss explained, thanks!

  2) Did I miss any simpler solutions? This is my first foray into Shelley transaction creation and Shelley ledger state inspection.

  3) We can't lower `k` because that risks CP violations. But we need to reach the second epoch in order to test this update proposal (`epochSize = 10k/f = 20k` here, and `k>=5` seems to avoid CP violations so far). So I increased `numSlots`, which makes these already slow tests even slower. Any better ideas?

cc: @edsko

### DIFF
--- a/ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
+++ b/ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
@@ -13,9 +13,11 @@ module Test.ThreadNet.Infra.Shelley (
     CoreNode(..)
   , CoreNodeKeyInfo(..)
   , DecentralizationParam(..)
+  , KesConfig(..)
   , coreNodeKeys
   , genCoreNode
   , mkGenesisConfig
+  , mkKesConfig
   , mkLeaderCredentials
   , mkProtocolRealTPraos
   , tpraosSlotLength
@@ -35,8 +37,8 @@ import           Test.QuickCheck
 import           Cardano.Binary (toCBOR)
 import           Cardano.Crypto.DSIGN.Class (DSIGNAlgorithm (..), SignKeyDSIGN,
                      signedDSIGN)
-import           Cardano.Crypto.KES.Class (SignKeyKES, deriveVerKeyKES,
-                     genKeyKES)
+import           Cardano.Crypto.KES.Class (KESAlgorithm, SignKeyKES,
+                     deriveVerKeyKES, genKeyKES, totalPeriodsKES)
 import           Cardano.Crypto.Seed (mkSeedFromBytes)
 import qualified Cardano.Crypto.Seed as Cardano.Crypto
 import           Cardano.Crypto.VRF.Class (SignKeyVRF, deriveVerKeyVRF,
@@ -49,6 +51,7 @@ import           Ouroboros.Consensus.Node.ProtocolInfo
 import           Ouroboros.Consensus.Util.IOLike
 
 import           Test.Util.Orphans.Arbitrary ()
+import           Test.Util.Slots (NumSlots (..))
 import           Test.Util.Time (dawnOfTime)
 
 import qualified Shelley.Spec.Ledger.Address as SL
@@ -192,6 +195,38 @@ mkLeaderCredentials CoreNode { cnDelegateKey, cnVRF, cnKES, cnOCert } =
       }
 
 {-------------------------------------------------------------------------------
+  KES configuration
+-------------------------------------------------------------------------------}
+
+-- | Currently @'maxEvolutions' * 'slotsPerEvolution'@ is the max number of
+-- slots the test can run without needing new ocerts.
+--
+-- TODO This limitation may be lifted by PR #2107, see
+-- <https://github.com/input-output-hk/ouroboros-network/issues/2107>.
+data KesConfig = KesConfig
+  { maxEvolutions     :: Word64
+  , slotsPerEvolution :: Word64
+  }
+
+-- | A 'KesConfig' that will not require more evolutions than this test's
+-- crypto @c@ allows
+mkKesConfig
+  :: forall proxy c. KESAlgorithm c
+  => proxy c -> NumSlots -> KesConfig
+mkKesConfig prx (NumSlots t) = KesConfig
+    { maxEvolutions
+    , slotsPerEvolution = divCeiling t maxEvolutions
+    }
+  where
+    maxEvolutions = fromIntegral $ totalPeriodsKES prx
+
+    -- | Like 'div', but rounds-up.
+    divCeiling :: Integral a => a -> a -> a
+    divCeiling n d = q + min 1 r
+      where
+        (q, r) = quotRem n d
+
+{-------------------------------------------------------------------------------
   TPraos node configuration
 -------------------------------------------------------------------------------}
 
@@ -204,10 +239,10 @@ mkGenesisConfig
   -> SecurityParam
   -> DecentralizationParam
   -> SlotLength
-  -> Word64  -- ^ Max KES evolutions
+  -> KesConfig
   -> [CoreNode c]
   -> ShelleyGenesis c
-mkGenesisConfig pVer k d slotLength maxKESEvolutions coreNodes =
+mkGenesisConfig pVer k d slotLength kesCfg coreNodes =
     ShelleyGenesis {
       -- Matches the start of the ThreadNet tests
       sgSystemStart           = dawnOfTime
@@ -216,14 +251,8 @@ mkGenesisConfig pVer k d slotLength maxKESEvolutions coreNodes =
     , sgActiveSlotsCoeff      = recip recipF
     , sgSecurityParam         = maxRollbacks k
     , sgEpochLength           = EpochSize (10 * maxRollbacks k * recipF)
-      -- TODO maxKESEvolutions * sgSlotsPerKESPeriod = max number of slots the
-      -- test can run without needing new ocerts. The maximum number of slots
-      -- the tests run now is 200 and the mock KES supports 10 evolutions, so
-      -- 10 * 20 == 200 is enough.
-      -- We can relax this in:
-      -- <https://github.com/input-output-hk/ouroboros-network/issues/2107>
-    , sgSlotsPerKESPeriod     = 20
-    , sgMaxKESEvolutions      = maxKESEvolutions
+    , sgSlotsPerKESPeriod     = slotsPerEvolution kesCfg
+    , sgMaxKESEvolutions      = maxEvolutions     kesCfg
     , sgSlotLength            = getSlotLength slotLength
     , sgUpdateQuorum          = quorum
     , sgMaxLovelaceSupply     = maxLovelaceSupply
--- a/ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
+++ b/ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
@@ -16,10 +16,13 @@ module Test.ThreadNet.Infra.Shelley (
   , KesConfig(..)
   , coreNodeKeys
   , genCoreNode
+  , incrementMinorProtVer
+  , mkEpochSize
   , mkGenesisConfig
   , mkKesConfig
   , mkLeaderCredentials
   , mkProtocolRealTPraos
+  , mkSetDecentralizationParamTxs
   , tpraosSlotLength
   ) where
 
@@ -37,6 +40,7 @@ import           Test.QuickCheck
 import           Cardano.Binary (toCBOR)
 import           Cardano.Crypto.DSIGN.Class (DSIGNAlgorithm (..), SignKeyDSIGN,
                      signedDSIGN)
+import           Cardano.Crypto.Hash.Class (hashWithSerialiser)
 import           Cardano.Crypto.KES.Class (KESAlgorithm, SignKeyKES,
                      deriveVerKeyKES, genKeyKES, totalPeriodsKES)
 import           Cardano.Crypto.Seed (mkSeedFromBytes)
@@ -59,12 +63,15 @@ import qualified Shelley.Spec.Ledger.BaseTypes as SL
 import qualified Shelley.Spec.Ledger.Coin as SL
 import qualified Shelley.Spec.Ledger.Credential as SL
 import qualified Shelley.Spec.Ledger.Crypto as SL
+import qualified Shelley.Spec.Ledger.Genesis as SL
 import qualified Shelley.Spec.Ledger.Keys as SL
 import qualified Shelley.Spec.Ledger.OCert as SL
 import qualified Shelley.Spec.Ledger.PParams as SL
+import qualified Shelley.Spec.Ledger.Tx as SL
 import qualified Shelley.Spec.Ledger.TxData as SL
 
-import           Ouroboros.Consensus.Shelley.Ledger (ShelleyBlock)
+import           Ouroboros.Consensus.Shelley.Ledger (GenTx (..), ShelleyBlock,
+                     mkShelleyTx)
 import           Ouroboros.Consensus.Shelley.Node
 import           Ouroboros.Consensus.Shelley.Protocol
 import           Ouroboros.Consensus.Shelley.Protocol.Crypto (DSIGN)
@@ -98,6 +105,13 @@ instance Arbitrary DecentralizationParam where
 
 tpraosSlotLength :: SlotLength
 tpraosSlotLength = slotLengthFromSec 2
+
+-- | The reciprocal of the active slots coefficient
+--
+-- To simplify related calculations throughout the test suite, we require this
+-- to be a natural number.
+recipActiveSlotsCoeff :: Num a => a
+recipActiveSlotsCoeff = 2   -- so f = 0.5
 
 {-------------------------------------------------------------------------------
   CoreNode secrets/etc
@@ -230,6 +244,9 @@ mkKesConfig prx (NumSlots t) = KesConfig
   TPraos node configuration
 -------------------------------------------------------------------------------}
 
+mkEpochSize :: SecurityParam -> EpochSize
+mkEpochSize (SecurityParam k) = EpochSize $ 10 * k * recipActiveSlotsCoeff
+
 -- | Note: a KES algorithm supports a particular max number of KES evolutions,
 -- but we can configure a potentially lower maximum for the ledger, that's why
 -- we take it as an argument.
@@ -248,9 +265,9 @@ mkGenesisConfig pVer k d slotLength kesCfg coreNodes =
       sgSystemStart           = dawnOfTime
     , sgNetworkMagic          = 0
     , sgNetworkId             = networkId
-    , sgActiveSlotsCoeff      = recip recipF
+    , sgActiveSlotsCoeff      = recip recipActiveSlotsCoeff
     , sgSecurityParam         = maxRollbacks k
-    , sgEpochLength           = EpochSize (10 * maxRollbacks k * recipF)
+    , sgEpochLength           = mkEpochSize k
     , sgSlotsPerKESPeriod     = slotsPerEvolution kesCfg
     , sgMaxKESEvolutions      = maxEvolutions     kesCfg
     , sgSlotLength            = getSlotLength slotLength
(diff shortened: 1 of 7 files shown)


nfrisby:
Example output includes:

```
      Updating d (250 in total):
      51.2% No
      25.2% Yes, LT
      23.6% Yes, GT
--
      Updating d (250 in total):
      57.6% No
      25.6% Yes, LT
      16.8% Yes, GT
--
      Updating d (250 in total):
      59.2% No
      21.2% Yes, LT
      19.6% Yes, GT
--
      Updating d (250 in total):
      52.4% No
      24.4% Yes, LT
      23.2% Yes, GT
```

ouroboros-consensus-shelley-test/src/Test/ThreadNet/TxGen/Shelley.hs
```diff
@@ -43,17 +44,26 @@ import           Test.ThreadNet.Infra.Shelley
 
 data ShelleyTxGenExtra h = ShelleyTxGenExtra
   { -- | Generator environment.
-    stgeGenEnv :: Gen.GenEnv h
+    stgeGenEnv  :: Gen.GenEnv h
+    -- | Generate not transactions before this slot.
```
mrBliss, line 48:
```suggestion
    -- | Generate no transactions before this slot.
```

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -14,25 +15,33 @@ import           Test.Tasty.QuickCheck
 
 import           Cardano.Crypto.Hash (ShortHash)
 import qualified Cardano.Crypto.KES.Class as KES
+import           Cardano.Slotting.EpochInfo.Impl (fixedSizeEpochInfo)
```
mrBliss, line 18:
```suggestion
import           Cardano.Slotting.EpochInfo (fixedSizeEpochInfo)
```

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -14,25 +15,33 @@ import           Test.Tasty.QuickCheck
...
 import           Test.Util.Orphans.Arbitrary ()
+import           Test.Util.Slots (NumSlots (..))
 
 import qualified Shelley.Spec.Ledger.BaseTypes as SL
 import qualified Shelley.Spec.Ledger.OCert as SL
+import qualified Shelley.Spec.Ledger.LedgerState as SL
 import qualified Shelley.Spec.Ledger.PParams as SL
 
 import           Ouroboros.Consensus.Shelley.Ledger (ShelleyBlock)
+import qualified Ouroboros.Consensus.Shelley.Ledger as SO
```
mrBliss, line 44: `SO`? I have used `Shelley` as the qualifier for this import, but maybe adding the two extra imports to the import list is less redundant.

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -45,47 +54,85 @@ import           Test.ThreadNet.Util.Seed (runGen)
 
 type Crypto = TPraosMockCrypto ShortHash
 
+-- | A number in @[0,10]@.
+newtype Tenths = Tenths Word64
```
mrBliss, line 58: When I read `Tenths`, I think about `1/10, 2/10, ..., 10/10`.

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -110,18 +163,47 @@ prop_simple_real_tpraos_convergence TestSetup
...
       , nodeJoinPlan = trivialNodeJoinPlan numCoreNodes
       , nodeRestarts = noRestarts
-      , txGenExtra   = ShelleyTxGenExtra $ mkGenEnv coreNodes
+      , txGenExtra   = ShelleyTxGenExtra
+        { stgeGenEnv  = mkGenEnv coreNodes
+        , stgeStartAt =
+            SlotNo $ if includingDUpdateTx then 1 else 0
+            -- We don't generate any transactions before the transaction
+            -- carrying the proposal because they might consume its inputs
+            -- before it does, thereby rendering it invalid.
```
mrBliss, line 172: Ah, I was wondering why you needed that, ok.

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
nfrisby: "Tenth" behaves like a unit; compare to `newtype Seconds = Seconds Word64`, for example.

Does that observation influence your preference?

mrBliss: > Does that observation influence your preference?

Sorry, no :grinning:

nfrisby: No sweat off my back :) I will change it.

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
nfrisby: Done, thanks.

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
nfrisby: Done, thanks.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/TxGen/Shelley.hs
nfrisby: Done, thanks.

nfrisby:
Thanks for the review, @mrBliss. I've addressed everything and cleaned up the commits.

I think it's ready for final Approval and merge.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -44,29 +54,68 @@ import           Ouroboros.Consensus.Node.ProtocolInfo
...
 import qualified Test.Shelley.Spec.Ledger.Generator.Core as Gen
 
+{-------------------------------------------------------------------------------
+  The decentralization parameter
+-------------------------------------------------------------------------------}
+
+-- | A suitable value for the @d@ protocol parameter
+--
+-- In the range @0@ to @1@, inclusive. Beware the misnomer: @0@ means fully
+-- decentralized, and @1@ means fully centralized.
```
mrBliss, line 88: Thanks for including this, I always forget this :slightly_smiling_face:

ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -158,36 +207,67 @@ mkLeaderCredentials CoreNode { cnDelegateKey, cnVRF, cnKES, cnOCert } =
...
+-- slots the test can run without needing new ocerts.
+--
+-- TODO This limitation may be lifted by PR #2107, see
+-- <https://github.com/input-output-hk/ouroboros-network/issues/2107>.
+data KesConfig = KesConfig
+  { maxEvolutions     :: Word64
+  , slotsPerEvolution :: Word64
+  }
+
+-- | A 'KesConfig' that will not require more evolutions than this tests's
```
mrBliss, line 224:
```suggestion
-- | A 'KesConfig' that will not require more evolutions than this test's
```

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -56,36 +68,88 @@ data TestSetup = TestSetup
...
+
+maxK :: Word64
+maxK = 10   -- More than this wastes execution time
+
+-- | In particular, choose a 'numSlots' appropriate for this test
+genRealTPraosTestConfig :: SecurityParam -> Gen TestConfig
+genRealTPraosTestConfig setupK = do
+    tc <- arbitrary
+
+    ensureMultipleEpoches <- frequency [(2, pure True), (1, pure False)]
```
mrBliss, line 100:
```suggestion
    ensureMultipleEpochs <- frequency [(2, pure True), (1, pure False)]
```

ouroboros-consensus-shelley-test/test/Main.hs
```diff
@@ -11,6 +11,10 @@ main = defaultMain tests
 
 tests :: TestTree
 tests =
+  testGroup "shelley"
+  [ Test.ThreadNet.RealTPraos.tests
+  ]
+  `asTypeOf`
```
mrBliss, line 17: Forgot to remove?

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -56,36 +68,88 @@ data TestSetup = TestSetup
...
+    pure $ if not ensureMultipleEpoches then tc else
+      let mkEpochSize :: SecurityParam -> Word64
+          mkEpochSize k = 10 * maxRollbacks k * recipActiveSlotsCoeff
```
mrBliss, line 103: We have this in `mkGenesisConfig` too, maybe promote it to a top-level helper?

ouroboros-consensus-shelley-test/test/Main.hs
nfrisby: Gah, this is the second PR in which I've let that slip in :(

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
nfrisby: Done, thanks!

nfrisby:
> 
building of '/nix/store/vv1wx2py1l00km71h9xh5wpj0acqvwag-ouroboros-consensus-shelley-test-test-test-0.1.0.0-check' timed out after 900 seconds of silence

Ugh. This could be demonstration of the risk of the relative slow down this PR introduces. Though I'm definitely surprised to see 15 minutes. Maybe CI is hammered?

nfrisby:
As of this latest push, I have adjusted the nightly builds of RealTPraos. It runs less tests than it used to (400 instead of 1500 per core), because now 25% of those tests add 1 or 2 epochs-worth of slots, which can be quite a bit more slots than the usual 0-100. If `k=10`, then 2 additional epochs is an additional `2 * 10 * k / f = 400` slots on top of the standard 0-100. We only add epochs to the nightly tests.

We add those epochs because (a) it's important to test multiple epochs; (b) we're interested in testing the net when `d` is changed by a PPU, which requires multiple epochs; (c) Shelley epochs are long (10k/f, with f=0.5 in this test); and (d) we can't shorten them by lowering `k` (min 5 in this test) because that makes CP violations too likely.

This diff also handles some more obstacles found during this work:

  (1) The `test-infra: add and use TxGenFailure exception` commit. The cardano-ledger-specs utxo generator that RealTPraos reuses occasionally gives up by calling `QC.discard`. This was causing tests failures/inflation of run times (because it restarted the RealTPraos test). For now, this diff catches and ignores that exception unless it happens 1000 times in a row (different seed each time). On the 1000th failure, we raise a fatal exception. So far I've only seen it fail 80 times in a row, but I really don't know how to judge if even 1000 is sufficient. @uroboros is working on making the generator better at avoiding this case and/or making it easier to control what it does when it gives up.

  (2) The `shelley-test: introduce KesConfig record in ThreadNet infra` commit. When we run for so many extra slots, we were exhausting the mock KES limits. This diff adds code to set the evolution interval based on the number of slots so that the KES limits are always out of reach of a particular test.

  (3) The default behavior of the Shelley utxo generator may generate arbitrary protocol parameter updates. Those can clobber the particular update that this test introduces to set `d`. So we instruct the generator to exclude PPUs if we're introducing our own blessed PPU that changes `d`.

  (4) Every Shelley transaction, even just a PPU, must consume an input. It was easiest to consume one of the inputs made available by the genesis block. Therefore we disable the generation of any txs during slot 0, so that we can be sure that the input is available for our blessed PPU that changes `d`.

nfrisby:
I've just now kicked off https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/168 as another experiment to check if I estimated the run times correctly.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -637,8 +640,38 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
         -- we generate different transactions in each slot.
-        let txs = runGen (nodeSeed `combineWith` unSlotNo curSlotNo) $
-                    testGenTxs numCoreNodes curSlotNo cfg txGenExtra ledger
+        let genTxs i =
+                runGen
+                  (nodeSeed `combineWith` unSlotNo curSlotNo `combineWith` i)
+                  (testGenTxs numCoreNodes curSlotNo cfg txGenExtra ledger)
+
+        -- These TxGen generators fail by invoking 'QC.discard'. We try many
+        -- times before actually giving up. When we give up, it fatal, not just
```
mrBliss, line 649:
```suggestion
        -- times before actually giving up. When we give up, it's fatal, not just
```

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -637,8 +640,38 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
+                case lr of
+                  Right _ -> pure txs
+                  Left  _ ->
+                      if acc < accLimit then loop (acc + 1) else
+                      MonadThrow.throwM (TxGenFailure accLimit)
+              where
+                seqList :: forall a. [a] -> ()
+                seqList = \case
+                    []   -> ()
+                    x:xs -> x `seq` seqList xs
```
mrBliss, line 663: There exists [`forceElemsToWHNF`](https://github.com/input-output-hk/cardano-prelude/blob/eee27bd6697d5ab1b943ee9c50ee638bde009c81/src/Cardano/Prelude/Strict.hs#L17) for this.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -637,8 +640,38 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
+                runGen
+                  (nodeSeed `combineWith` unSlotNo curSlotNo `combineWith` i)
+                  (testGenTxs numCoreNodes curSlotNo cfg txGenExtra ledger)
+
+        -- These TxGen generators fail by invoking 'QC.discard'. We try many
+        -- times before actually giving up. When we give up, it fatal, not just
+        -- a discard.
+        let loop acc = do
+                let txs = genTxs acc
+                lr <- tryJust isDiscard $ MonadThrow.evaluate $ seqList txs
```
mrBliss, line 653: `discard :: a` is nasty, making you do things like this :cry:

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -637,8 +640,38 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
+                    x:xs -> x `seq` seqList xs
+
+                isDiscard :: Exn.SomeException -> Maybe ()
+                isDiscard e = if QC.isDiscard e then Just () else Nothing
+
+                -- At time of writing, a RealTPraos test run maxed out at about
+                -- 85 consecutive failures here.
+                --
+                -- TODO Better justification.
+                accLimit = 1000 :: Int
```
mrBliss, line 672: Fine for me, and it's only temporary.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -637,8 +640,38 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
-                    testGenTxs numCoreNodes curSlotNo cfg txGenExtra ledger
+        let genTxs i =
+                runGen
+                  (nodeSeed `combineWith` unSlotNo curSlotNo `combineWith` i)
+                  (testGenTxs numCoreNodes curSlotNo cfg txGenExtra ledger)
+
+        -- These TxGen generators fail by invoking 'QC.discard'. We try many
+        -- times before actually giving up. When we give up, it fatal, not just
+        -- a discard.
+        let loop acc = do
```
mrBliss, line 651: I had the impression `acc` was an accumulator of transactions, but it's actually more of a seed, right? Maybe a different name and type signature would clarify it.

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -121,28 +204,71 @@ prop_simple_real_tpraos_convergence TestSetup
...
+    inclPPUs :: WhetherToGeneratePPUs
+    inclPPUs =
+        -- We don't generate any other updates, since doing so might
+        -- accidentally supplant the bespoke update that these tests are
+        -- expecting.
+        --
+        -- The transaction this test introduces causes all nodes to propose the
+        -- same parameter update. It'd technically be OK if some nodes then
+        -- changed their proposal to a different update, as long as at least
+        -- @Quorum@-many nodes were still proposing this test's original update
```
mrBliss, line 227: `sgUpdateQuorum` is currently 1, but since we're asserting that the genesis config must be valid, it trips over the recently introduced https://github.com/input-output-hk/cardano-ledger-specs/blob/c453fa6fce99ed3cb09d129c829ddbf0980a642c/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/Genesis.hs#L303, which I fix here https://github.com/input-output-hk/ouroboros-network/pull/2456/commits/54c71ce0836016d7309494984796204064101715. This is fine because all your nodes propose the same update, correct?

mrBliss:
@nfrisby You can bors it whenever you think it is ready. Because this only affects testing, this can go straight to master, no need to wait until after tagging the release.

ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
nfrisby: Yep, the transaction associate the same proposal to every genesis key hash. And we're preventing the random generator from changing any of those associations by preventing it from introducing any updates.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
nfrisby: Done!

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
nfrisby: Yeah :(

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
nfrisby: Thanks!

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
nfrisby: Thanks!

PR #2380, opened 2020-07-10
Don't combine ticked LedgerView and ChainDepState

edsko:
This was leading to unnecessary `error` calls. Instead, if protocols need the ticked ledger view, they can include it in the ticked chain dep state.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
@@ -271,6 +271,12 @@ instance PBftCrypto c => ChainSelection (PBft c) where
 instance HasChainIndepState (PBft c)
   -- Use defaults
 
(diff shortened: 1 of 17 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -271,6 +271,13 @@ instance PBftCrypto c => ChainSelection (PBft c) where
 instance HasChainIndepState (PBft c)
   -- Use defaults
 
+-- Ticking has no effect on the PBFtState, but we do need the ticked ledger view
+data instance Ticked (PBftState c) = TickedPBftState {
+      tickedPBftLedgerView :: Ticked (LedgerView (PBft c))
+    , getTickedPBftState   :: PBftState c
+    }
+
+
```
edsko, line 280: :scream:

edsko:
bors merge

PR #2382, opened 2020-07-10
Do not let d=0 in HFC ThreadNet test (ie revert #2378)

nfrisby:
I created this commit via `git revert -m c12c9cd03`

> This reverts commit c12c9cd03e1909c7ad514fbf8404b37e8e2e90fd, reversing changes made to fc548489331cf6a5fab148b1b121e3bf6a6b184f.

---

I've debugged the observed failure https://hydra.iohk.io/build/3393090/nixlog/1 and concluded that `d=0` is not currently supported (at least in the HFC test). Thus this PR reverts PR #2378.

The test runs for exactly two epochs, one Byron and one Shelley. There are no leaders in the Shelley era. I only ever see empty overlay schedules (as expected, since `d=0`) and an empty pool distribution (unexpected, at least by me) when I add tracing to the TPraos leadership check.

```
LedgerView
  { lvProtParams = ...
  , lvOverlaySched = fromList []
  , lvPoolDistr = PoolDistr {unPoolDistr = fromList []}
  , lvGenDelegs = ...
  }
```

(It also fails in the same way, against `origin/master`, if I let it run for exactly 3 epochs.)

### DIFF
--- a/ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
+++ b/ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
@@ -56,7 +56,7 @@ data TestSetup = TestSetup
 
 instance Arbitrary TestSetup where
   arbitrary = do
-    setupD <- (/10)         <$> choose   (0, 10)
+    setupD <- (/10)         <$> choose   (1, 10)
     setupK <- SecurityParam <$> elements [5, 10]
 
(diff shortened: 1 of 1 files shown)


nfrisby:
Closing without merging. I've confused myself. It seems that `d=0` is OK in `Test.ThreadNet.RealTPraos`, but problematic in `Test.ThreadNet.Cardano`.

PR #2375, opened 2020-07-09
generalize db-converter

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/2369

PR #2390, opened 2020-07-13
Avoid recomputing the ShelleyLedgerConfig

mrBliss:
In #2375, we discovered that validating a Byron chain took 10x as long using
`CardanoBlock` than using `ByronBlock`.

Looking at the profiling report, we see it being dominated by:

```
exp' src/Shelley/Spec/NonIntegral.hs:(171,1)-(175,43) 52.1 56.7
ln'  src/Shelley/Spec/NonIntegral.hs:(159,1)-(162,27) 22.7 26.1
```

Which is strange, as we're validating a Byron-only chain! It turns out that
these two functions are called by `mkActiveSlotCoeff`, which is called as part
of `mkShelleyLedgerConfig`, which we are calling in `completeLedgerConfig`. This
happens for *each* block.

To fix this, we replace the partial ledger config of Shelley with a non-partial
one containing a dummy `EpochInfo Identity`. In `completeLedgerConfig` we
replace the dummy `EpochInfo Identity` with the correct one. This means we only
call `mkShelleyLedgerConfig` once and cache its result.


mrBliss:
bors merge

PR #2395, opened 2020-07-13
Fix stack build

mrBliss:
BuildKite was apparently failing for master.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -11,8 +11,10 @@ packages:
   - ./ouroboros-consensus/ouroboros-consensus-test-infra
   - ./ouroboros-consensus/ouroboros-consensus-mock
   - ./ouroboros-consensus-byron
+  - ./ouroboros-consensus-byron-test
   - ./ouroboros-consensus-byronspec
   - ./ouroboros-consensus-shelley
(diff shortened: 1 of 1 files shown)


mrBliss:
I actually don't think enabling `stack-rebuild` for all branches is the right thing to do. It will take ages to do a full stack build (no caching whatsoever :roll_eyes:), it will slow down all our PRs, it will incur extra load on CI, etc. All that while we don't officially support Stack.

We could fix Stack build issues on-demand, i.e., when they're noticed on master :flushed:.

mrBliss:
Ok, that's it. The `stack-rebuild` job has been running for almost 30 minutes and it's only at 94/212. The `buildkite` job is required for merging and I don't want to be forced to wait at least one hour to merge each PR.

mrBliss:
bors merge

mrBliss:
It's still not fixed, I guess because of the lack of the `-external-libsodium-vrf` flag. Maybe a Stack user that cares about it (and doesn't mind waiting +1h for each build) can fix it.

mrBliss:
BTW, not even `cardano-node` is doing `stack-rebuild` in CI.

PR #2399, opened 2020-07-14
HFC compatibility layer

edsko:
This makes it possible to submit the same query languages whether or not the HFC is in use.


edsko:
bors merge

PR #2358, opened 2020-07-03
Refactor network/consensus versioning

mrBliss:
Fixes #2309.

Previously, we had a list of supported `BlockNodeToNodeVersion`s and a method
to translate them to a `NodeToNodeVersion`. This meant that there were two
layers of "support": (1) include the version in `BlockNodeToNodeVersion`
and (2) return a non-`error` for it in the translation function.

Replace these methods by a map from `NodeToNodeVersion` to
`BlockNodeToNodeVersion`. Similarly for `NodeToClient`. An important insight
is that not each network version requires a consensus-side block version. For
example, enabling the `LocalStateQuery` `NodeToClient` protocol requires a new
network version, but no block version, as the serialisation format doesn't
change. A new block version is only needed when the serialisation changes, so
not when protocols get added/removed or the protocol messages are
added/removed.

This means we don't need separate Byron (nor Cardano) versions for
`NodeToClientV_2`, as the serialisation format didn't change, so remove these
redundant versions and map `NodeToClientV_2` to `ByronNodeToClientVersion1`.

The `mostRecentSupportedNodeToNode` method was only used for the ThreadNet
tests to make sure we were testing the most recent version. Remove this method
in favour of randomly picking a supported version in the tests. The
`Test.ThreadNet.Util.NodeToNodeVersion` was added to aid in picking a version.

* Rewrite `foldMapVersions` and `combineVersions` to work for any `Foldable`
  instead of a `NonEmpty`, as we now use a `Map` instead of a `NonEmpty` for
  the supported versions.

* Rewrite `Cardano.Client.Subscription` to take a `CodecConfig` +
  `NetworkMagic` instead of an entire `TopLevelConfig`. Producing a
  `TopLevelConfig` is a lot of work for clients. The whole point of
  introducing the `CodecConfig` was to make it easier for clients.

PR #2401, opened 2020-07-14
Keep-Alive 15.x

karknu:


ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/NetworkProtocolVersion.hs
```diff
@@ -19,6 +19,9 @@ data ByronNodeToNodeVersion =
 
     -- | We send headers /with/ a size hint
   | ByronNodeToNodeVersion2
+
+    -- | KeepAlive protocol
+  | ByronNodeToNodeVersion3
```
mrBliss, line 24: This is not needed, in #2358 we decoupled binary serialisation format changes from protocol changes.

PR #2416, opened 2020-07-16
Generalize chain selection

### LINKED ISSUE #2118: Generalize `ChainSelection` for `HardForkProtocol`
The definition as it stands uses `BlockNo` when the two tips are from different eras. This would mean that we cannot compare certificate numbers when transitioning from Shelley to Shelley-MC or Goguen.

This is not necessary until after Shelley, so marking this as `enhancement`.

edsko:
Closes #2118.


edsko:
bors merge

PR #2417, opened 2020-07-16
Remove stack-rebuild BuildKite job

### LINKED ISSUE #2394: Buildkite cannot build the master branch
All the recent builds fail at https://buildkite.com/input-output-hk/ouroboros-network/builds?branch=master

Builds on branches seem OK for some reason.

The failure seems related to the `stack` config versus somewhat recent additional packages/deps like `sop-core` or `small-steps`.

mrBliss:
Fixes #2394.

We do not support Stack officially and it has been broken for a while. We were
only running the job on master, not on PRs, because it takes ages (at least 1
hour), so it will inevitably break again after fixing it.

Our neighbours in `cardano-node` and `cardano-ledger-specs` don't have this CI
job either, so I see no point in incurring the extra CI cost just to have the
master branch marked with a red cross.


mrBliss:
bors merge

PR #2421, opened 2020-07-16
Further updates from cardano-ledger-specs.

nc6:
- Merge ProtocolMagicId and NetworkMagic in Genesis. I could have pushed
this merging further up the chain, but have a preference for more
minimal diffs this close to release :-)

- This incorporates changes to Byron address witnessing, so we are no
longer calling into Byron code to do this.


nc6:
bors merge

PR #2423, opened 2020-07-16
Validate the Shelley genesis config

mrBliss:
Use the validation function introduced in
<https://github.com/input-output-hk/cardano-ledger-specs/pull/1644>.


mrBliss:
bors merge

PR #2431, opened 2020-07-17
Inject transactions from older eras

### LINKED ISSUE #2371: Mempool: inject transactions from older eras
The HFC should automatically inject transactions from one era into the next in the mempool (parameterized by a function to do so, obviously, between every consecutive pair of eras). The point of doing this is that there is client code out there that submits transactions but does not follow the chain (at least, not directly), but instead uses services such as the REST API to see if their transaction made it -- so those would continue to work even if we move to Goguen, and they haven't upgraded.

edsko:
Closes #2371.


edsko:
bors merge

PR #2438, opened 2020-07-19
Export distribForgeState

kantp:
This exports `distribForgeState` from
`Ouroboros.Consensus.HardFork.Combinator.Forge`, which is needed in order to
trace the KES metrics after the Shelley transition in the Cardano
protocol (input-output-hk/cardano-node#1448).


kantp:
bors merge

Issue #2433, opened 2020-07-17
Enforce a minimum size on `inFlightBytesLowWatermark`


karknu:
We need to make sure that `inFlightBytesLowWatermark` is larger then 2 two blocks so that the Blockfetch protocol has a chance to run in pipelining mode..

Issue #2430, opened 2020-07-17
Blockfetch can get stuck fetching data from a slow peer


karknu:
There is a problem where Blockfetch will get stuck fetching data from a far-away peer.
The number of peers we’re fetching data from is controlled by MaxConcurrencyBulkSync, which is set to 1.
If a node starts to use a slow peer (for example because it was listed first in the topology file) it will never move away from it. Any possible requests from new peers will be stomped by MaxConcurrencyBulkSync , even if they could complete faster based on deltaQ estimates.
The only way we switch peers if there is enough network disruption that the outstanding request queue completely drain.

PR #2441, opened 2020-07-20
Blockfetch improvements needed before KeepAlive client can be enabled.

### LINKED ISSUE #2434: Limit the rate we make new blockfetch descisions
Limit blockfetch decision making to 100hz.

karknu:
Fixes #2434 #2433 and #2430 which are required for #1416.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -913,6 +913,9 @@ runThreadNetwork systemTime ThreadNetworkArgs
                   bfcMaxConcurrencyBulkSync = 1
                 , bfcMaxConcurrencyDeadline = 2
                 , bfcMaxRequestsInflight    = 10
+                , bfcDecisionLoopInterval   = 0.0 -- Mock testsuite can use sub-second slot
+                                                  -- interval which doesn't play nice with
+                                                  -- blockfetch descision interval.
                 }
             }
 
--- a/ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
+++ b/ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
@@ -47,6 +47,7 @@ import           Ouroboros.Network.BlockFetch.DeltaQ
                    ( PeerGSV(..), SizeInBytes
                    , PeerFetchInFlightLimits(..)
                    , calculatePeerFetchInFlightLimits
+                   , comparePeerGSV
                    , estimateResponseDeadlineProbability
                    , estimateExpectedResponseDuration )
 
@@ -93,10 +94,11 @@ data FetchMode =
   deriving (Eq, Show)
 
 
-type PeerInfo header extra =
+type PeerInfo header peer extra =
        ( PeerFetchStatus header,
          PeerFetchInFlight header,
          PeerGSV,
+         peer,
          extra
        )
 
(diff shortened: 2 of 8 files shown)


ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -91,10 +92,11 @@ data FetchMode =
   deriving (Eq, Show)
 
 
-type PeerInfo header extra =
+type PeerInfo header peer extra =
```
coot, line 95: What about changing this into a record with strict fields.

karknu: I'm not a fan of the "swizzel" functions used to access the different fields of the PeerInfo type, but I think that such a change is outside the scope of this PR.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -917,11 +943,20 @@ fetchRequestDecision FetchDecisionPolicy {
...
+  = Left $ FetchDeclineConcurrencyLimit
+             fetchMode maxConcurrentFetchPeers
+
+    -- If we're at the concurrency limit refuse any additional peers.
   | peerFetchReqsInFlight == 0
   , let maxConcurrentFetchPeers = case fetchMode of
                                     FetchModeBulkSync -> maxConcurrencyBulkSync
                                     FetchModeDeadline -> maxConcurrencyDeadline
-  , nConcurrentFetchPeers >= maxConcurrentFetchPeers
+  , nConcurrentFetchPeers == maxConcurrentFetchPeers
```
coot, line 959: Could you elaborate why this change above?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -854,11 +863,32 @@ fetchRequestDecisions fetchDecisionPolicy fetchMode chains =
...
                        peerFetchReqsInFlight)
       $ chains
 
 
+    -- Order the peers based on current PeerGSV. The top performing peers will be
+    -- permitted to go active even if we're above the desired maxConcurrentFetchPeers
+    -- which will cause us to switch smoothly from a slower to faster peers.
+    -- When switching from slow to faster peers we will be over the configured limit, but
+    -- PeerGSV is expected to be updated rather infrequently so the set of prefered peers should
+    -- be stable during 10s of second.
```
coot, line 876: Looks good :)

ouroboros-network/test/Ouroboros/Network/BlockFetch/Examples.hs
```diff
@@ -128,7 +128,7 @@ blockFetchExample1 decisionTracer clientStateTracer clientMsgTracer
           decisionTracer clientStateTracer
           (sampleBlockFetchPolicy1 blockHeap currentChainHeaders candidateChainHeaders)
           registry
-          (BlockFetchConfiguration 2 1 10)
+          (BlockFetchConfiguration 2 1 10 0.01)
```
coot, line 131: I would use the named fields to set up the fields, e.g.
```
BlockFetchConfiguration {
    bfcMaxConcurrencyBulkSync = 2,
    bfcMaxConcurrencyDeadline = 1,
    bfcMaxRequestsInflight = 10,
    bfcDecisionLoopInterval = 0.01
  }
```
This way it's easier to figure out the meaning of each value.

ouroboros-network/src/Ouroboros/Network/BlockFetch/State.hs
```diff
@@ -67,19 +75,24 @@ fetchLogicIterations decisionTracer clientStateTracer
...
+      stateFingerprint' <- fetchLogicIteration
         decisionTracer clientStateTracer
         fetchDecisionPolicy
         fetchTriggerVariables
         fetchNonTriggerVariables
         stateFingerprint
+      end <- getMonotonicTime
+      let delta = diffTime end start
+      -- Limit descision making to once every decisionLoopInterval.
+      when (delta < decisionLoopInterval fetchDecisionPolicy) $ threadDelay delta
```
nfrisby, line 94: As you mentioned on our Slack thread: `threadDelay (decisionLoopInterval fetchDecisionPolicy - delta)`.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -913,6 +913,9 @@ runThreadNetwork systemTime ThreadNetworkArgs
                   bfcMaxConcurrencyBulkSync = 1
                 , bfcMaxConcurrencyDeadline = 2
                 , bfcMaxRequestsInflight    = 10
+                , bfcDecisionLoopInterval   = 0.0 -- Mock testsuite can use sub-second slot
```
nfrisby, line 916: We decided to set it to `0` for now, which disables the rate limiter during the ThreadNet tests.

If this interval is greater than the slot length, then this will lead to test failures. In this case, some BF thread would eventually sleep for a duration that spans the next slot onset, but the tests currently assume that the nodes are more responsive than that -- roughly: "all nodes are fully synced by the end of every slot".

See https://input-output-rnd.slack.com/archives/CFKLUH4R0/p1595255785454400 for the discussion.

PR #2443, opened 2020-07-21
test-infra: uniform distribution for Seed

nfrisby:
The `arbitrary :: Int` is not uniform; it favors values close to 0.

I'm not sure this diff ultimately matters, but this seems closer to our intentions.


nfrisby:
Since we have time to be patient for this one, I'm running the nightlies on it: https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/162

PR #2444, opened 2020-07-21
iosim: fixup MonadEvaluate IOSim instance

nfrisby:
This diff follows-up on PR #2043.

In particular, the behavior of `runSim $ evaluate $ error "boom" :: Either Failure ()` is now `Left _`, but was the same as `error "boom"` before this commit.

### DIFF
--- a/io-sim/src/Control/Monad/IOSim.hs
+++ b/io-sim/src/Control/Monad/IOSim.hs
@@ -71,6 +71,7 @@ import qualified System.IO.Error as IO.Error (userError)
 import           Control.Monad (when)
 import           Control.Monad.ST.Lazy
 import qualified Control.Monad.ST.Strict as StrictST
+import           Control.Monad.ST.Lazy.Unsafe (unsafeIOToST)
 import           Data.STRef.Lazy
 
 import qualified Control.Monad.Catch as Exceptions
@@ -121,6 +122,7 @@ data SimA s a where
   Throw        :: SomeException -> SimA s a
   Catch        :: Exception e =>
                   SimA s a -> (e -> SimA s a) -> (a -> SimA s b) -> SimA s b
+  Evaluate     :: a -> (a -> SimA s b) -> SimA s b
 
   Fork         :: SimM s () -> (ThreadId -> SimA s b) -> SimA s b
   GetThreadId  :: (ThreadId -> SimA s b) -> SimA s b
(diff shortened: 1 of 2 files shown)


io-sim/src/Control/Monad/IOSim.hs
```diff
@@ -777,6 +779,18 @@ schedule thread@Thread{
                                                (CatchFrame handler k ctl) }
       schedule thread' simstate
 
+    Evaluate expr k -> do
+      mbWHNF <- unsafeIOToST $ try $ evaluate expr
```
nfrisby, line 783: I'm disappointed by `unsafeIOToST` here -- is there a safe alternative?

io-sim/src/Control/Monad/IOSim.hs
```diff
@@ -777,6 +779,18 @@ schedule thread@Thread{
...
+      case mbWHNF of
+        Left e -> do
+          -- schedule this thread to immediately raise the exception
+          let thread' = thread { threadControl = ThreadControl (Throw e) ctl }
+          schedule thread' simstate
+        Right whnf -> do
+          -- continue with the resulting WHNF
+          let thread' = thread { threadControl = ThreadControl (k whnf) ctl }
+          schedule thread' simstate
```
nfrisby, line 792: Do these two alternatives recursively call `schedule` correctly? I admit that I mostly copied from similar constructors' cases, without considering all of the context that I'm not altering.

io-sim/src/Control/Monad/IOSim.hs
coot: I don't think there's a way to avoid `unsafe` function when we want to use `evaluate` (which, I think, is a bit magical how it deals with exceptions).

io-sim/src/Control/Monad/IOSim.hs
coot: This is allright.

nfrisby:
Thanks for responding to my questions.

> Could you add a test?

Pushed!

-----

FYI, this function was already defined by `io-sim` but wasn't being used anywhere (there's an occurrence, but it's commented out). I needed it to debug/work-around some exception-related behavior on work-in-progress branch.

io-sim/src/Control/Monad/IOSim.hs
nfrisby: I agree :( thanks for confirming.

PR #2445, opened 2020-07-21
Update index-state and nix updates

newhoggy:


nix/default.nix
```diff
@@ -2,28 +2,30 @@
 , crossSystem ? null
 , config ? {}
 , sourcesOverride ? {}
+, gitrev ? null
```
coot, line 5: I don't see it used anywhere, is it really needed?

newhoggy: Not used.  Removed.  Thanks!

mrBliss:
@coot Any reason not to merge this (or are we waiting until after the hard fork)?

coot:
@nc6 are you sure we can already merge PRs?

PR #2446, opened 2020-07-21
Branch to collect post-fork improvements

edsko:
Any PRs that change the consensus layer, unless they are high priority bug fixes that should be included on mainnet prior to the hard fork transition, should be opened against this one (rather than against `master`). We will collect them here, and after the transition has happened, merge them into master. This is just a temporary measure to make sure that we can respond quickly should any problem require urgent fixes.


edsko:
bors merge

PR #2457, opened 2020-07-23
Default bfcMaxConcurrencyDeadline should be 1

karknu:


mrBliss:
Out of curiosity, why should it be 1 instead of 2?

karknu:
A `bfcMaxConcurrencyDeadline` > 1 is acceptable for stake pool operators that are fine with wasting some bandwidth on fetching blocks from multiple sources in order to get it as fast as possible.
The default config should be suitable for normal clients.

PR #2458, opened 2020-07-23
Mention cause and point when revalidating a header fails

mrBliss:
I ran into this when validating the mainnet_candidate with the `db-validator` tool. Thanks to the improved error message I discovered that it was because of an incorrect value used for PBftSignatureThreshold, which is hard-coded to the default of 0.22 in the db-validator.

Bonus: base16 `Show` instance of `OneEraHash`.


mrBliss:
bors merge

Issue #2440, opened 2020-07-20
Simplify LedgerDB after `snapEvery` changes to 1


kderme:
If we decide to keep all recent snapshots in memory, the ledgerDB can be simplified:
- `ledgerDbSnapEvery` field can be removed
- `rollback` doesn't need to apply any new blocks.

Issue #1936, opened 2020-04-10
Keep _every_ snapshot of the ledger state in memory


edsko:
At the moment, the ledger DB has a `snapEvery` parameter that tells it how often to store a snapshot of the ledger state in memory. When the ledger DB is then asked for a historical snapshot, it might have to replay some blocks, which are stored on disk. This is relatively expensive, but should also occur relatively infrequently. Ticket #1026 is about figuring out a good value for that `snapEvery` parameter, and also suggests that we might to special case some snapshots near the tip (see also https://github.com/input-output-hk/ouroboros-network/issues/1026#issuecomment-572945358). 

It is worth considering if we could just store _every_ snapshot. This would make the design of the ledger DB significantly simpler, would avoid the need for any kinds of special cases, and as a bonus, also allows simplifications in other parts of the code base; in particular, see https://github.com/input-output-hk/ouroboros-network/issues/1935#issuecomment-611950826 .

mrBliss:
@intricate Can you try to figure out how much memory keeping the last `k` snapshots of the ledger would require (just set `snapEvery` to 1) compared to the current policy (with mainnet)? You can measure the memory by profiling or by using the heap size stuff.

kderme:
Some initial results I got by using `computeHeapSize`. Probably it would make sense to compare this with the number of transactions on these blocks and test for bigger chains. However the main issue is that I met 4 closures types that are not supported: `THUNK_STATIC, FUN_STATIC, MUT_ARR_PTRS_FROZEN_CLEAN,  MUT_ARR_PTRS_FROZEN_DIRTY` which I'm not sure I supported the correct way.

![Figure_1-1](https://user-images.githubusercontent.com/11467473/87584645-7f600b80-c6e6-11ea-8e28-7d50cff1ceb4.png)

kderme:
Eventuall these closure types disappeared when I added a bang pattern for the result of `ledgerDbPush` before calling `computeHeapSize`. These are the results I got for a small part of mainnet, starting from the same snapshot at block 3a76f8bf33d854e221d958543d42293ab8ce94b4573a7b1e5d70cd345e8a9bf3.

![full](https://user-images.githubusercontent.com/11467473/87900211-5c25bb00-ca5c-11ea-97aa-f83541f22ca9.png)

![zoom2](https://user-images.githubusercontent.com/11467473/87900259-87a8a580-ca5c-11ea-8fdd-2d49e6fe7ee8.png)
[...]

PR #2459, opened 2020-07-23
Keep every snapshot in memory and remove the concept of snapEvery

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/2440
I tried to
- remove some tests which became trivial.
- update the documentation.
- simplify some constraints

but it's possible I missed some of those.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
@@ -318,15 +318,11 @@ getCurrent LgrDB{..} = readTVar varDB
 getCurrentState :: IOLike m => LgrDB m blk -> STM m (ExtLedgerState blk)
 getCurrentState LgrDB{..} = LedgerDB.ledgerDbCurrent <$> readTVar varDB
 
-getPastState :: (IOLike m, UpdateLedger blk, LedgerSupportsProtocol blk)
+getPastState :: (IOLike m, HasHeader blk)
              => LgrDB m blk -> Point blk -> m (Maybe (ExtLedgerState blk))
 getPastState LgrDB{..} p = do
     db <- atomically $ readTVar varDB
-    LedgerDB.defaultResolveBlocks resolveBlock $
-      LedgerDB.ledgerDbPast
-        (extLedgerCfgFromTopLevel (lgrTopLevelConfig args))
-        (pointToWithOriginRealPoint p)
-        db
+    return $ LedgerDB.ledgerDbPast (pointToWithOriginRealPoint p) db
 
 -- | PRECONDITION: The new 'LedgerDB' must be the result of calling either
 -- 'LedgerDB.ledgerDbSwitch' or 'LedgerDB.ledgerDbPushMany' on the current
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
@@ -56,18 +56,15 @@ module Ouroboros.Consensus.Storage.LedgerDB.InMemory (
    , ledgerDbPush'
    , ledgerDbPushMany'
    , ledgerDbSwitch'
-   , ledgerDbPast'
    ) where
 
 import           Prelude hiding (mod, (/))
-import qualified Prelude
 
 import           Codec.Serialise (Serialise (..))
 import           Codec.Serialise.Decoding (Decoder)
 import qualified Codec.Serialise.Decoding as Dec
 import           Codec.Serialise.Encoding (Encoding)
 import qualified Codec.Serialise.Encoding as Enc
-import           Control.Monad ((>=>))
 import           Control.Monad.Except hiding (ap)
 import           Control.Monad.Reader hiding (ap)
 import           Data.Foldable (toList)
@@ -98,184 +95,65 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
 --
 -- The ledger DB looks like
 --
--- > anchor |> blocks and snapshots <| current
+-- > anchor |> snapshots <| current
 --
 -- where @anchor@ records the oldest known snapshot and @current@ the most
--- recent.
+-- recent. The anchor is the oldest point we can roll back to.
 --
--- As an example, suppose we have @snapEvery = 4@ (i.e., we will take a
--- snapshot every 4 blocks) and @k = 6@. The ledger DB grows as illustrated
--- below, where we indicate the anchor, number of blocks, the stored
--- blocks/snapshots, and the current ledger; the oldest block we can roll back
--- to is marked.
+-- We take a snapshot after each block is applied and keep in memory a window
+-- of the last @k@ snapshots. We have verified empirically (#1936) that the
+-- overhead of keeping @k snapshots in memory is small, i.e., about 5%
+-- compared to keeping a snapshot every 100 blocks. This is thanks to sharing
+-- between consecutive snapshots.
 --
--- > anchor |> #   [ blocks ]                                      <| tip
+-- As an example, suppose we have @k = 6@. The ledger DB grows as illustrated
+-- below, where we indicate the anchor number of blocks, the stored snapshots,
+-- and the current ledger.
+--
+-- > anchor |> #   [ snapshots ]                   <| tip
 -- > ---------------------------------------------------------------------------
--- > *G     |> (0) [ ]                                             <| G
--- > *G     |> (1) [ B1]                                           <| L1
--- > *G     |> (2) [ B1,  B2]                                      <| L2
--- > *G     |> (3) [ B1,  B2,   B3]                                <| L3
--- > *G     |> (4) [ B1,  B2,   B3,  L4]                           <| L4
--- > *G     |> (5) [ B1,  B2,   B3,  L4,  B5]                      <| L5
--- > *G     |> (6) [ B1,  B2,   B3,  L4,  B5,  B6]                 <| L6
--- >  G     |> (7) [*B1,  B2,   B3,  L4,  B5,  B6,  B7]            <| L7
--- >  G     |> (8) [ B1, *B2,   B3,  L4,  B5,  B6,  B7,  L8]       <| L8
--- >  G     |> (9) [ B1,  B2,  *B3,  L4,  B5,  B6,  B7,  L8,  B9]  <| L9   (*)
--- > *L4    |> (6) [ B5,  B6,   B7,  L8,  B9,  B10]                <| L10  (**)
--- >  L4    |> (7) [*B5,  B6,   B7,  L8,  B9,  B10, B11]           <| L11
--- >  L4    |> (8) [ B5, *B6,   B7,  L8,  B9,  B10, B11, L12]      <| L12
--- >  L4    |> (9) [ B5,  B6,  *B7,  L8,  B9,  B10, B12, L12, B13] <| L13
--- > *L8    |> (6) [ B9,  B10,  B12, L12, B13, B14]                <| L14
+-- > G      |> (0) [ ]                             <| G
+-- > G      |> (1) [ L1]                           <| L1
+-- > G      |> (2) [ L1,  L2]                      <| L2
+-- > G      |> (3) [ L1,  L2,  L3]                 <| L3
+-- > G      |> (4) [ L1,  L2,  L3,  L4]            <| L4
+-- > G      |> (5) [ L1,  L2,  L3,  L4,  L5]       <| L5
+-- > G      |> (6) [ L1,  L2,  L3,  L4,  L5,  L6]  <| L6
+-- > L1     |> (6) [ L2,  L3,  L4,  L5,  L6,  L7]  <| L7
+-- > L2     |> (6) [ L3,  L4,  L5,  L6,  L7,  L8]  <| L8
+-- > L3     |> (6) [ L4,  L5,  L6,  L7,  L8,  L9]  <| L9   (*)
+-- > L4     |> (6) [ L5,  L6,  L7,  L8,  L9,  L10] <| L10
+-- > L5     |> (6) [*L6,  L7,  L8,  L9,  L10, L11] <| L11
+-- > L6     |> (6) [ L7,  L8,  L9,  L10, L11, L12] <| L12
+-- > L7     |> (6) [ L8,  L9,  L10, L12, L12, L13] <| L13
+-- > L8     |> (6) [ L9,  L10, L12, L12, L13, L14] <| L14
 --
--- The ledger DB must guarantee we must at all times be able to roll back @k@
+-- The ledger DB must guarantee that at all times we are able to roll back @k@
 -- blocks. For example, if we are on line (*), and roll back 6 blocks, we get
 --
--- > G |> [B1, B2, B3]
---
--- from which we can /compute/ @G |> [B1, B2, B3] <| L3@ by re-applying
--- @snapEvery - 1@ blocks. However, as soon as we pushed one more block @(**)@,
--- and then roll back 6 blocks, we end up with
---
--- > L4 |> [] <| L4
---
--- This representation has the following properties:
---
--- * The distance between snapshots is at most @snapEvery@. This implies that
---   rolling back involves re-applying at most @snapEvery - 1@ blocks
---   (note that @snapEvery >= 1@).
---
--- * This implies that the number of (references to) blocks we store will vary
---   between @k@ and @k + snapEvery - 1@ (unless we are near genesis).
---
--- * When the number of blocks reaches @k + snapEvery@, we can drop the first
---   @snapEvery@ blocks, shifting up the anchor.
---
--- * The average number of blocks we have to reapply on a rollback is given by
---
---   >    average [0 .. snapEvery - 1]
---   > == ((snapEvery - 1) * snapEvery / 2) / snapEvery
---   > == (snapEvery - 1) / 2
---
---   (Obvious) boundary case: if @snapEvery == 1@, the average number is zero.
---
--- * The number of snapshots we store is in the range
---
---   > [1 + floor(k / snapEvery), 1 + ceiling(k / snapEvery)]
---
---   If @snapEvery@ divides @k@, the number is precisely @1 + k / snapEvery@.
---
--- * Picking a suitable value of @snapEvery@ for a given @k@ is a tradeoff
---   between cost of reapplying blocks and the cost of storing more snapshots.
---   The latter is primarily the cost of less opportunity for garbage
---   collection, which is hard to quantify abstractly. This should probably
---   be determined empirically.
---
--- Some example boundary cases:
---
--- * Suppose @k = 4@ and @snapEvery = 1@:
---
---   > *G  |> []               <| G
---   > *G  |> [L1]             <| L1
---   > *G  |> [L1, L2]         <| L2
---   > *G  |> [L1, L2, L3]     <| L3
---   > *G  |> [L1, L2, L3, L4] <| L4
---   > *L1 |> [L2, L3, L4, L5] <| L5
---   > *L2 |> [L3, L4, L5, L6] <| L6
---
---   Note that in this case the number of blocks will always be @k@ (unless we
---   are close to genesis), and the anchor is always the oldest point we can
---   roll back to.
---
--- * If @k = 1@ and @snapEvery = 4@, we get
---
---   > *G  |> [ ]                  <| G
---   > *G  |> [ B1]                <| L1
---   >  G  |> [*B1,  B2]           <| L2
---   >  G  |> [ B1, *B2,  B3]      <| L3
---   >  G  |> [ B1,  B2, *B3, L4]  <| L4
---   > *L4 |> [ B5]                <| L5
---   >  L4 |> [*B5,  B6]           <| L6
---   >  L4 |> [ B5, *B6,  B7]      <| L7
---   >  L4 |> [ B5,  B6, *B7, L8]  <| L8
---   > *L8 |> [*B9]                <| L9
---
---   (The minimum distance between the current ledger and the maximum rollback
---   is @k@.)
---
--- * If @k = 0@ and @snapEvery = 4@, we get
---
---   > *G  |> [ ]                  <| G
---   >  G  |> [*B1]                <| L1
---   >  G  |> [ B1, *B2]           <| L2
---   >  G  |> [ B1,  B2, *B3]      <| L3
---   > *L4 |> [ ]                  <| L4
---   >  L4 |> [*B5]                <| L5
---   >  L4 |> [ B5, *B6]           <| L6
---   >  L4 |> [ B5,  B6, *B7]      <| L7
---   > *L8 |> [ ]                  <| L8
---
---   @k = 0@ is the only case where the list might be empty. Of course, this is
---   not a particularly useful configuration.
---
--- * If @k = 1@ and @snapEvery = 1@, we get
---
---   > *G  |> [ ]   <| G
---   > *G  |> [ L1] <| L1
---   > *L1 |> [ L2] <| L2
---   > *L2 |> [ L3] <| L3
---
--- * If @k = 0@ and @snapEvery = 1@, we get
---
---   > *G  |> [ ] <| G
---   > *L1 |> [ ] <| L1
---   > *L2 |> [ ] <| L2
---
--- * Finally, if @k = snapEvery = k4@, we get
---
---   > *G  |> [ ]                             <| G
---   > *G  |> [ B1]                           <| L1
---   > *G  |> [ B1,  B2]                      <| L2
---   > *G  |> [ B1,  B2,  B3]                 <| L3
---   > *G  |> [ B1,  B2,  B3, L4]             <| L4
---   >  G  |> [*B1,  B2,  B3, L4, B5]         <| L5
---   >  G  |> [ B1, *B2,  B3, L4, B5, B6]     <| L6
---   >  G  |> [ B1,  B2, *B3, L4, B5, B6, B7] <| L7
---   > *L4 |> [ B5,  B6,  B7, L8]             <| L8
---   >  L4 |> [*B5,  B6,  B7, L8, B9]         <| L9
+-- > L3 |> []
 data LedgerDB l r = LedgerDB {
-      -- | The ledger state at the tip of the chain
-      ledgerDbCurrent :: !l
-
       -- | Older ledger states
-    , ledgerDbBlocks  :: !(StrictSeq (Checkpoint l r))
+      ledgerDbCheckpoints :: !(StrictSeq (Checkpoint l r))
 
       -- | Information about the state of the ledger /before/
-    , ledgerDbAnchor  :: !(ChainSummary l r)
+    , ledgerDbAnchor      :: !(ChainSummary l r)
 
       -- | Ledger DB parameters
-    , ledgerDbParams  :: !LedgerDbParams
+    , ledgerDbParams      :: !LedgerDbParams
     }
   deriving (Show, Eq, Generic, NoUnexpectedThunks)
 
-data LedgerDbParams = LedgerDbParams {
-      -- | Take a snapshot every @n@ blocks
-      --
-      -- Must be @>= 1@.
-      ledgerDbSnapEvery     :: !Word64
-
+newtype LedgerDbParams = LedgerDbParams {
       -- | Security parameter (maximum rollback)
-    , ledgerDbSecurityParam :: !SecurityParam
+      ledgerDbSecurityParam :: SecurityParam
     }
   deriving (Show, Eq, Generic, NoUnexpectedThunks)
 
 -- | Default parameters
---
--- TODO: We should decide empirically what a good @snapEvery@ value is.
--- <https://github.com/input-output-hk/ouroboros-network/issues/1026>
 ledgerDbDefaultParams :: SecurityParam -> LedgerDbParams
-ledgerDbDefaultParams (SecurityParam k) = LedgerDbParams {
-      ledgerDbSnapEvery     = 100
-    , ledgerDbSecurityParam = SecurityParam k
+ledgerDbDefaultParams securityParam = LedgerDbParams {
+      ledgerDbSecurityParam = securityParam
     }
 
 {-------------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -98,184 +95,67 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
 --
 -- The ledger DB looks like
 --
--- > anchor |> blocks and snapshots <| current
+-- > anchor |> snapshots <| current
 --
 -- where @anchor@ records the oldest known snapshot and @current@ the most
--- recent.
+-- recent. The anchor is the oldest point we can rollback.
```
mrBliss, line 101:
```suggestion
-- recent. The anchor is the oldest point we can roll back to.
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -98,184 +95,67 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
...
 --
--- As an example, suppose we have @snapEvery = 4@ (i.e., we will take a
--- snapshot every 4 blocks) and @k = 6@. The ledger DB grows as illustrated
--- below, where we indicate the anchor, number of blocks, the stored
--- blocks/snapshots, and the current ledger; the oldest block we can roll back
--- to is marked.
+-- We take snapshots after each block is applied and keep in memory a window
+-- of the last k snapshots. We have verified empirically that the overhead of
+-- keeping k snapshots in memory is small and doesn't exceed 9MiB. That's
+-- because there is a lot of sharing between consecutive snapshots.
```
mrBliss, line 106:
```suggestion
-- We take a snapshot after each block is applied and keep in memory a window
-- of the last @k@ snapshots. We have verified empirically (#1936) that the
-- overhead of keeping @k snapshots in memory is small, i.e., about 5%
-- compared to keeping a snapshot every 100 blocks. This is thanks to sharing
-- between consecutive snapshots.
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -98,184 +95,67 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
...
+-- > G     |> (5) [ L1,  L2,  L3,  L4,  L5]       <| L5
+-- > G     |> (6) [ L1,  L2,  L3,  L4,  L5,  L6]  <| L6
+-- > L1    |> (6) [ L2,  L3,  L4,  L5,  L6,  L7]  <| L7
+-- > L2    |> (6) [ L3,  L4,  L5,  L6,  L7,  L8]  <| L8
+-- > L3    |> (6) [ L4,  L5,  L6,  L7,  L8,  L9]  <| L9   (*)
+-- > L4    |> (6) [ L5,  L6,  L7,  L8,  L9,  L10] <| L10
+-- > L5    |> (6) [*L6,  L7,  L8,  L9,  L10, L11] <| L11
+-- > L6    |> (6) [ L7,  L8,  L9,  L10, L11, L12] <| L12
+-- > L7    |> (6) [ L8,  L9,  L10, L12, L12, L13] <| L13
+-- > L8    |> (6) [ L9,  L10, L12, L12, L13, L14] <| L14
```
mrBliss, line 128: Align `|>`, `[`, and `<|` with those in the header on line 112.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -98,184 +95,67 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
...
 --
 -- The ledger DB must guarantee we must at all times be able to roll back @k@
 -- blocks. For example, if we are on line (*), and roll back 6 blocks, we get
```
mrBliss, line 131:
```suggestion
-- The ledger DB must guarantee that at all times we are able to roll back @k@
-- blocks. For example, if we are on line (*), and roll back 6 blocks, we get
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -98,184 +95,67 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
...
-      -- | Take a snapshot every @n@ blocks
-      --
-      -- Must be @>= 1@.
-      ledgerDbSnapEvery     :: !Word64
-
-      -- | Security parameter (maximum rollback)
-    , ledgerDbSecurityParam :: !SecurityParam
+newtype LedgerDbParams = LedgerDbParams {
+    -- | Security parameter (maximum rollback)
+    ledgerDbSecurityParam :: SecurityParam
```
mrBliss, line 151:
```suggestion
      -- | Security parameter (maximum rollback)
      ledgerDbSecurityParam :: SecurityParam
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -478,53 +352,45 @@ applyBlock cfg ap db = case ap of
...
-ledgerDbToList LedgerDB{..} = map cpToPair $ toList ledgerDbBlocks
+ledgerDbToList :: LedgerDB l r -> [(r, l)]
+ledgerDbToList LedgerDB{..} = map cpToPair $ toList ledgerDbSnapshots
 
 -- | All snapshots currently stored by the ledger DB (new to old)
 --
--- For each snapshot we also return the distance from the tip
-ledgerDbSnapshots :: forall l r. LedgerDB l r -> [(Word64, l)]
-ledgerDbSnapshots LedgerDB{..} = go 0 ledgerDbBlocks
+-- This also include the anchor. For each snapshot we also return the distance
```
mrBliss, line 370:
```suggestion
-- -- This also includes the snapshot at the anchor. For each snapshot we also return the distance
```

(but wrapped to 80 chars)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -549,157 +413,88 @@ shiftAnchor toRemove ChainSummary{..} = ChainSummary {
...
-rollback :: forall m l r b. (ApplyBlock l b, ResolvesBlocks r b m)
-         => FullBlockConfig l b
-         -> Word64
+rollback :: forall l r.
+            Word64
          -> LedgerDB l r
-         -> m (Maybe (LedgerDB l r))
-rollback _   0 db = return $ Just db
-rollback cfg n db = rollbackTo cfg (\_anchor -> go) db
+         -> Maybe (LedgerDB l r)
```
mrBliss, line 495: Pure, perfect :slightly_smiling_face:

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -771,7 +564,7 @@ ledgerDbSwitch :: (ApplyBlock l b, ResolvesBlocks r b m, c)
                -> LedgerDB l r
                -> m (Either ExceededRollback (LedgerDB l r))
 ledgerDbSwitch cfg numRollbacks newBlocks db = do
-    mRolledBack <- rollback cfg numRollbacks db
+    let mRolledBack = rollback numRollbacks db
     case mRolledBack of
```
mrBliss, line 568:
```suggestion
    case rollback numRollbacks db of
```
and remove `do` from L566.

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/LedgerDB/InMemory.hs
```diff
@@ -407,20 +366,13 @@ instance Arbitrary SwitchSetup where
 instance Arbitrary LedgerDbParams where
   arbitrary = do
       k <- choose (0, 6)
-      snapEvery <- choose (1, 8)
       return $ LedgerDbParams {
-            ledgerDbSnapEvery     = snapEvery
-          , ledgerDbSecurityParam = SecurityParam k
+          ledgerDbSecurityParam = SecurityParam k
```
mrBliss, line 370:
```suggestion
            ledgerDbSecurityParam = SecurityParam k
```

ouroboros-consensus/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -469,10 +469,10 @@ runMock cmd initMock =
         -- The snapshot that the real implementation will write to disk
         --
         -- The real implementation keeps the length of the blocks between @k@
-        -- and @k + snapEvery 1@ (provided that there are enough blocks). The
-        -- function 'ledgerDbCountToPrune' computes how many (old) blocks should
-        -- be dropped to get back into that range; the last (most recent) block
-        -- to be dropped will become the new anchor. It is the anchor that gets
+        -- and @k + 1@ (provided that there are enough blocks). The function
```
mrBliss, line 472: Is `k + 1` correct? Oh, is that after pushing a block before `ledgerDbCountToPrune`?

kderme: Yes, it's actually k. I bit confusing, because the implementation uses this function for the in memory snapshots while here it's used to create a disk snapshot.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -318,15 +318,13 @@ getCurrent LgrDB{..} = readTVar varDB
 getCurrentState :: IOLike m => LgrDB m blk -> STM m (ExtLedgerState blk)
 getCurrentState LgrDB{..} = LedgerDB.ledgerDbCurrent <$> readTVar varDB
 
-getPastState :: (IOLike m, UpdateLedger blk, LedgerSupportsProtocol blk)
+getPastState :: (IOLike m, StandardHash blk)
```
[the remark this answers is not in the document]
kderme: How come we prefer this instead of the smaller constraint?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -98,184 +95,68 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
...
---   >  L4 |> [*B5,  B6,  B7, L8, B9]         <| L9
+-- > L3 |> []
 data LedgerDB l r = LedgerDB {
       -- | The ledger state at the tip of the chain
-      ledgerDbCurrent :: !l
+      ledgerDbCurrent   :: !l
 
       -- | Older ledger states
-    , ledgerDbBlocks  :: !(StrictSeq (Checkpoint l r))
+    , ledgerDbSnapshots :: !(StrictSeq (Checkpoint l r))
```
mrBliss, line 140:
```suggestion
    , ledgerDbCheckpoints :: !(StrictSeq (Checkpoint l r))
```
Now the field name matches the element name.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -478,53 +353,45 @@ applyBlock cfg ap db = case ap of
...
+ledgerDbToList LedgerDB{..} = map cpToPair $ toList ledgerDbSnapshots
 
 -- | All snapshots currently stored by the ledger DB (new to old)
 --
--- For each snapshot we also return the distance from the tip
-ledgerDbSnapshots :: forall l r. LedgerDB l r -> [(Word64, l)]
-ledgerDbSnapshots LedgerDB{..} = go 0 ledgerDbBlocks
+-- This also includes the snapshot at the anchor. For each snapshot we also
+-- return the distance from the tip.
+ledgerDbOldStates :: forall l r. LedgerDB l r -> [(Word64, l)]
```
mrBliss, line 373: Rename this to `ledgerDbSnapshots`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -98,184 +95,68 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
...
---   >  G  |> [*B1,  B2,  B3, L4, B5]         <| L5
---   >  G  |> [ B1, *B2,  B3, L4, B5, B6]     <| L6
---   >  G  |> [ B1,  B2, *B3, L4, B5, B6, B7] <| L7
---   > *L4 |> [ B5,  B6,  B7, L8]             <| L8
---   >  L4 |> [*B5,  B6,  B7, L8, B9]         <| L9
+-- > L3 |> []
 data LedgerDB l r = LedgerDB {
       -- | The ledger state at the tip of the chain
-      ledgerDbCurrent :: !l
+      ledgerDbCurrent   :: !l
```
mrBliss, line 137: We don't really need to store this separately anymore, we can just look at the last element of `ledgerDbCheckpoints`. You can turn this into a function. When the `StrictSeq` is empty, use `csLedger` of `ledgerDbAnchor`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -548,158 +413,89 @@ shiftAnchor toRemove ChainSummary{..} = ChainSummary {
...
          -> LedgerDB l r
-         -> m (Maybe (LedgerDB l r))
-rollback _   0 db = return $ Just db
-rollback cfg n db = rollbackTo cfg (\_anchor -> go) db
+         -> Maybe (LedgerDB l r)
+rollback 0 db = Just db
+rollback n db = rollbackTo (\_anchor -> go) db
   where
     go :: StrictSeq (Checkpoint l r) -> Maybe (StrictSeq (Checkpoint l r))
     go blocks =
```
mrBliss, line 501: Rename `blocks` to `checkpoints`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -770,9 +564,8 @@ ledgerDbSwitch :: (ApplyBlock l b, ResolvesBlocks r b m, c)
                -> [Ap m l r b c]  -- ^ New blocks to apply
                -> LedgerDB l r
                -> m (Either ExceededRollback (LedgerDB l r))
```
mrBliss, line 566: The `ResolvesBlocks r b m` constraint on `ledgerDbSwitch` should go, it was only required to resolve blocks when rolling back, which should no longer be necessary. Note the `c` argument, which can still imply `ResolvesBlocks r b m` when the list of `Ap`s contains a `Re/ApplyRef`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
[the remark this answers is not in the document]
mrBliss: Yeah, good question. It's what we always do, but that's not a good answer :thinking:

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -98,184 +95,65 @@ import           Ouroboros.Consensus.Util.CBOR (decodeWithOrigin,
...
---   >  G  |> [ B1, *B2,  B3, L4, B5, B6]     <| L6
---   >  G  |> [ B1,  B2, *B3, L4, B5, B6, B7] <| L7
---   > *L4 |> [ B5,  B6,  B7, L8]             <| L8
---   >  L4 |> [*B5,  B6,  B7, L8, B9]         <| L9
+-- > L3 |> []
 data LedgerDB l r = LedgerDB {
-      -- | The ledger state at the tip of the chain
-      ledgerDbCurrent :: !l
-
       -- | Older ledger states
```
mrBliss, line 136: Now that `ledgerDbCurrent` is no longer a field, we only have these "older ledger states", which includes the current one. So rename it to:
```suggestion
      -- | Ledger state checkpoints
```

mrBliss:
Just as a sanity check, I tried this out using `cardano-node`:

`cardano-node`: e8fb486e44aceb4b9962d7b3b76849becd510820
`ouroboros-network`: ec74d1c463b6d18bdca2b714abe730c1d237cc32 + with and without cherry picking c4d44357e2a3c12b30188f305f9c4be6521e6271

Using a database recently synced with mainnet, in the Shelley era. I disabled syncing. Very important: the VolatileDB must contain a chain of at least `k` blocks, otherwise we won't have `k` snapshots in memory.

Trace output:

```
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:16.18 UTC] Opened imm db with immutable tip at (Point 4564640, "35b10144ee477ba9f80389fbcdbb4f5ebd7678e1750dbaa2583eeff00de69466") and epoch 211                                                                                                                                                                [desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:16.46 UTC] Opened vol db
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:19.91 UTC] Replaying ledger from snapshot DiskSnapshot 28 at (Point 4563080, "2c85088fa88973c87cd0f2aca3aebab30ab9719e72e2b784d2395b7df692071e")                                                                      
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:20.23 UTC] Replayed block: slot SlotNo {unSlotNo = 4563100} of At (SlotNo {unSlotNo = 4564640})                             [desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:22.92 UTC] block replay progress (%) = 100.0                                                               
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:24.87 UTC] before next, messages elided = 4563119                                                                           [desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:24.87 UTC] Replayed block: slot SlotNo {unSlotNo = 4564640} of At (SlotNo {unSlotNo = 4564640})                             
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:24.87 UTC] Opened lgr db                                                                                                    [desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:19:51.55 UTC] Opened db with immutable tip at (Point 4564640, "35b10144ee477ba9f80389fbcdbb4f5ebd7678e1750dbaa2583eeff00de69466") and tip (Point 4608280, "64455962d499039d109e642aafce53408df6c9dcc0f8f726074a837b739682c7")  
```

I looked at the memory usage of `cardano-node` with `htop`.

Without this PR: **911 MB RAM**
With this PR: **7717 MB RAM** (!!!)

I repeated this a few times and always got similar results. Trying with `snapEvery = 1` instead of this PR gave similar results.

This means we should definitely *not* merge this PR now.

Thoughts:
* The measurements in #1936 were done using the Byron ledger, not the Shelley (or Cardano) one.
* I suspect that for some reason, there is much less sharing in the Shelley ledger than in the Byron one.
* Recently there has been a lot transactions in the chain, record numbers. This might explain some extra memory usage, but not this much.
* This is without https://github.com/input-output-hk/cardano-ledger-specs/pull/1722/, which fixes some thunks. I'll measure again with this fix in, but I'm not expecting much of it.

mrBliss:
As a sanity check, I repeated my measurements using mainnet at the end of the Byron:

Trace output:

```
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:16.18 UTC] Opened imm db with immutable tip at (Point 4564640, "35b10144ee477ba9f80389fbcdbb4f5ebd7678e1750dbaa2583eeff00de69466") and epoch 211
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:16.46 UTC] Opened vol db
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:19.91 UTC] Replaying ledger from snapshot DiskSnapshot 28 at (Point 4563080, "2c85088fa88973c87cd0f2aca3aebab30ab9719e72e2b784d2395b7df692071e")
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:20.23 UTC] Replayed block: slot SlotNo {unSlotNo = 4563100} of At (SlotNo {unSlotNo = 4564640})
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:22.92 UTC] block replay progress (%) = 100.0
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:24.87 UTC] before next, messages elided = 4563119
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:24.87 UTC] Replayed block: slot SlotNo {unSlotNo = 4564640} of At (SlotNo {unSlotNo = 4564640})
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:16:24.87 UTC] Opened lgr db
[desktop:cardano.node.ChainDB:Info:5] [2020-08-10 10:19:51.55 UTC] Opened db with immutable tip at (Point 4564640, "35b10144ee477ba9f80389fbcdbb4f5ebd7678e1750dbaa2583eeff00de69466") and tip (Point 4608280, "64455962d499039d109e642aafce53408df6c9dcc0f8f726074a837b739682c7")
```

Without this PR: **305 MB RAM**
With this PR: **455-465 MB RAM**

Not the catastrophic increase we saw for Shelley, but still more than expected from #1936. That's 150 MB more instead of the expected <10 MB :thinking:. I'm not sure how the Haskell heap grows, maybe that plays a role here?

I'm wondering whether we're overlooking something :thinking:.

mrBliss:
@kderme Could you repeat the experiments from #1936 with the Shelley ledger?

kderme:
With https://github.com/input-output-hk/ouroboros-network/commit/8e7b5e20ed12974a16a32ffcb17941b9a14f9775 and cherry-picked https://github.com/input-output-hk/cardano-ledger-specs/pull/1775 I still see very big memory usage.

snapEvery = 1:
```
 4,105,555,200 bytes maximum residency (47 sample(s))
 104,871,740,160 bytes allocated in the heap
 92,482,133,216 bytes copied during GC

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0     97207 colls,     0 par   38.548s  38.702s     0.0004s    0.0033s
  Gen  1        47 colls,     0 par   36.904s  40.847s     0.8691s    4.6105s

  MUT     time   60.596s  ( 62.544s elapsed)
  GC      time   75.452s  ( 79.549s elapsed)
  Total   time  136.048s  (142.094s elapsed)
```
snapEvery = 100:
```
 831,904,064 bytes maximum residency (120 sample(s))
 104,871,360,408 bytes allocated in the heap
 91,016,416,384 bytes copied during GC

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0     97134 colls,     0 par   38.149s  38.292s     0.0004s    0.0030s
  Gen  1       120 colls,     0 par   20.708s  21.219s     0.1768s    0.6442s

  MUT     time   61.557s  ( 63.407s elapsed)
  GC      time   58.857s  ( 59.510s elapsed)
  Total   time  120.415s  (122.918s elapsed)
```

snapEvery = k:
```
 682,901,344 bytes maximum residency (127 sample(s))
 104,871,422,048 bytes allocated in the heap
 90,606,678,416 bytes copied during GC

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0     97127 colls,     0 par   37.810s  37.952s     0.0004s    0.0035s
  Gen  1       127 colls,     0 par   18.535s  18.971s     0.1494s    0.4656s

  MUT     time   59.905s  ( 61.707s elapsed)
  GC      time   56.345s  ( 56.922s elapsed)
  Total   time  116.251s  (118.630s elapsed)

```

kderme:
It looks like what caused the issue was the delegationTransition fixed on https://github.com/input-output-hk/cardano-ledger-specs/pull/1779. Creating a new reward Map (with over 40k entries) probably killed sharing between consecutive ledger.

snapEvery = 1, without delegation fix:
every-1-no-fix-2


With the fix cherry-picked, the graph looks much different

snapEvery = 1, with delegation fix:

every-1-fix


This looks similar with
snapEvery = 100, with delegation fix:

every-100-fix

kderme:
Some points:
- All these results are with https://github.com/input-output-hk/cardano-ledger-specs/pull/1775 cherry-picked
- The memory difference between snapEvery=1 vs 100 (with the fix) happens because of `utxoInductive` (dark green on snapEvery=1 graph). Maybe something we can optimize or see if it's reasonable.
- I'd suggest we delay this a bit more, until all fixes are in place and also fixes on epoch boundaries, to make sure we don't cause big memory spikes

mrBliss:
> It looks like what caused the issue was the delegationTransition fixed on [input-output-hk/cardano-ledger-specs#1779](https://github.com/input-output-hk/cardano-ledger-specs/pull/1779). Creating a new reward Map (with over 40k entries) probably killed sharing between consecutive ledger.
> 
> [..]

Great, thanks for this investigation! Also nice to see I unknowingly already fixed the issue :slightly_smiling_face: 


> Some points:
> 
>     * All these results are with [input-output-hk/cardano-ledger-specs#1775](https://github.com/input-output-hk/cardano-ledger-specs/pull/1775) cherry-picked
> 
>     * The memory difference between snapEvery=1 vs 100 (with the fix) happens because of `utxoInductive` (dark green on snapEvery=1 graph). Maybe something we can optimize or see if it's reasonable.

I believe it corresponds to [this line](https://github.com/input-output-hk/cardano-ledger-specs/blob/d0f7eded325c4e4b36e44b4b49fd75b0c83f00f3/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Utxo.hs#L334):
```haskell
      { _utxo = eval ((txins txb ⋪ utxo) ∪ txouts txb),
```
which is the main expected contributor to the memory growth, i.e., UTxO changes. Unless that [`eval`](https://github.com/input-output-hk/cardano-ledger-specs/blob/d0f7eded325c4e4b36e44b4b49fd75b0c83f00f3/semantics/executable-spec/src/Control/Iterate/SetAlgebraInternal.hs#L1106) is doing something suboptimal, there's not much we can do.

> 
>     * I'd suggest we delay this a bit more, until all fixes are in place and also fixes on epoch boundaries, to make sure we don't cause big memory spikes

Which fixes do you mean? https://github.com/input-output-hk/cardano-ledger-specs/pull/1779 has been merged and https://github.com/input-output-hk/cardano-ledger-specs/pull/1775 should not affect sharing, right? Do you mean https://github.com/input-output-hk/cardano-ledger-specs/pull/1785?

I would actually be interested in seeing the impact of an epoch transition on the memory usage (with https://github.com/input-output-hk/cardano-ledger-specs/pull/1785 applied).

I'm relieved that we can still go through with this simplification after all :relieved:. But I agree, let's first optimise the overlay schedule so we can reuse the memory freed by that optimisation for this simplification.

kderme:
On epoch boundaries the results also seem similar. This is using the db-analyser, starting from a snapshot at the very beginning of epoch 208 and validating up to a very recent block:

snapEvery = 100:
every-100

snapEvery = 1:
every-1-boundaries

Both cases report a maximum residency of 2,500,000,000 bytes and I see max 4GB used by top/ps (What's this PINNED memory?).
Total time is slightly higher on snapEvery = 1 (around 10 secs). I believe this is because of more time spent on gc:

snapEvery = 100:
```
MUT     time   18.213s  ( 56.391s elapsed)
GC      time   64.126s  ( 34.119s elapsed)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0      4246 colls,  4246 par    9.601s   5.887s     0.0014s    0.0353s
  Gen  1       146 colls,   145 par   131.197s  66.556s     0.4559s    1.3432s
```

snapEvery = 1:
```
MUT     time   13.467s  ( 60.153s elapsed)
GC      time   83.366s  ( 44.040s elapsed)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0      4240 colls,  4240 par   11.467s   7.066s     0.0017s    0.0429s
  Gen  1       164 colls,   163 par   165.699s  83.906s     0.5116s    1.3530s


```

mrBliss:
> On epoch boundaries the results also seem similar. This is using the db-analyser, starting from a snapshot at the very beginning of epoch 208 and validating up to a very recent block:
> 
> snapEvery = 100:
> [..]
> 
> snapEvery = 1:
> [..]
> 
> Both cases report a maximum residency of 2,500,000,000 bytes and I see max 4GB used by top/ps (What's this PINNED memory?).

Good, that's what I was expecting/hoping: there can be at most one epoch boundary transition in a range of `k` blocks, so `snapEvery` should not matter.

PINNED memory is typically the bytes in `ByteString`s. As explained on Slack, my hypothesis is that these might be deserialised Byron addresses.

> Total time is slightly higher on snapEvery = 1 (around 10 secs). I believe this is because of more time spent on gc:
> 
> snapEvery = 100:
> 
> ```
> MUT     time   18.213s  ( 56.391s elapsed)
> GC      time   64.126s  ( 34.119s elapsed)
> 
>                                      Tot time (elapsed)  Avg pause  Max pause
>   Gen  0      4246 colls,  4246 par    9.601s   5.887s     0.0014s    0.0353s
>   Gen  1       146 colls,   145 par   131.197s  66.556s     0.4559s    1.3432s
> ```
> 
> snapEvery = 1:
> 
> ```
> MUT     time   13.467s  ( 60.153s elapsed)
> GC      time   83.366s  ( 44.040s elapsed)
> 
>                                      Tot time (elapsed)  Avg pause  Max pause
>   Gen  0      4240 colls,  4240 par   11.467s   7.066s     0.0017s    0.0429s
>   Gen  1       164 colls,   163 par   165.699s  83.906s     0.5116s    1.3530s
> ```

More things to traverse, so more time, that's expected. Note that profiling skews this result, running with `-s` without profiling, the result should be better. Also, that epoch boundary transition is generating tons of garbage, so it makes sense that GC is doing a lot of work.

kderme:
The results after the fix https://github.com/input-output-hk/ouroboros-network/pull/2532 look pretty similar again:

before pr:
every-100

after pr:
pr

I also tested what happened if we never call prune (keep all ledger on memory). This shows that sharing is pretty good, if we take into account that by the end there are ~90000 ledger states stored.
no-prune


I also tested times and memory with psrecord (no ghc profiling involved) and the pr doesn't create any visible difference:

plot1

mrBliss:
Great, let's merge!

mrBliss:
bors merge

mrBliss (commented):
bors merge

mrBliss (approved):
bors merge

mrBliss:
bors ping

mrBliss:
bors merge

PR #2461, opened 2020-07-23
cabal.project: use multiple subdirs

mrBliss:


cabal.project
```diff
@@ -120,132 +120,44 @@ source-repository-package
...
-  --sha256: 02v9bd95vjal7yp96b59dgap2k53c2lrg9vxw6d62cxzw8n635y6
-  subdir: test
-
-source-repository-package
-  type: git
-  location: https://github.com/input-output-hk/cardano-base
-  tag: 7d795c3040ea7785812efa1c97864bbb41b15d3e
-  --sha256: 130i0yj4y9br1m2bhisi6wni3f40i31nfhg878hv0kwi17chl9sz
-  subdir: binary
+  subdir: . test
```
mrBliss, line 130: For some reason put these two on two separate lines (like below) didn't work. Strange, because it seems to work in https://github.com/input-output-hk/plutus/blob/32790fd3344e11fa9f7d8a8a26ed8ef519621abd/cabal.project#L174-L175 (unless they don't actually depend on these packages).

mrBliss:
I think we'll need some help with CI.

cabal.project
nc6: I think plutus still uses the stack project to determine their CI

nc6:
Made some updates to this that should hopefully get it building in CI...

PR #2464, opened 2020-07-24
Optimise translating the Byron UTxO to the Shelley UTxO

mrBliss:
Instead of going from compact Byron Tx -> normal Byron Tx -> normal Shelley Tx
-> compact Shelley Tx, go from compact Byron Tx -> compact Shelley Tx.

The crux is to avoid deserialising and reserialising the address in the `TxOut`.

I measured the time it took to translate the Byron ledger to the Shelley ledger
when syncing mainnet_candidate2:

Before: 3.7s
After:  1.2s

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -15,6 +15,9 @@ module Ouroboros.Consensus.Cardano.CanHardFork (
     TriggerHardFork (..)
   , ByronPartialLedgerConfig (..)
   , ShelleyPartialLedgerConfig (..)
+    -- * Exported for testing purposes
+  , translateTxIdByronToShelley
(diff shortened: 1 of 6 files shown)


mrBliss:
Still a draft until https://github.com/input-output-hk/cardano-ledger-specs/pull/1708 has been merged.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -271,35 +273,34 @@ translatePointByronToShelley = \case
     GenesisPoint   -> GenesisPoint
     BlockPoint s h -> BlockPoint s (translateHeaderHashByronToShelley h)
 
+
+-- | We use the same hasing algorithm so we can unwrap and rewrap the bytes.
```
mrBliss, line 277:
```suggestion
-- | We use the same hashing algorithm so we can unwrap and rewrap the bytes.
```

mrBliss:
bors merge

PR #2465, opened 2020-07-24
Split Qry into Qry and Expr

### LINKED ISSUE #2463: Make `Qry` showable
This will make debugging easier. 

Mind that this will mean we need to do things a bit differently; the current use of `QBind` will make it impossible to show the `Qry` as it is currently defined.

edsko:
While `Qry` still provides a monadic interface, most of the actual calculation happens in the interpreter for `Expr`, and any past-horizon exceptions come from the evaluation of an `Expr`. Unlike `Qry`, `Expr` uses PHOAS, which means that we can actually `Show` expressions, despite them containing functions. This means that we can include the offending expression in the exception.

This is also a strict generalization: previously, an entire `Qry` was always evaluated against a single era; with the split, every `Expr` is evaluated against a single era, but every `Expr` embedded in a `Qry` can be evaluated against a _different_ era.

Closes #2463.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Types.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Types.hs
@@ -30,6 +30,7 @@ import           Control.Exception (assert)
 import           Data.Fixed
 import           Data.Time (NominalDiffTime, UTCTime, addUTCTime, diffUTCTime)
 import           GHC.Generics (Generic)
+import           Quiet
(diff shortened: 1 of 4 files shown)


edsko:
Marking this as a draft request, not because it's not ready, but because by rights it should have been opened against #2446 . This is open against `master` instead so that it's easier for the wallet folks to cherry-pick it on top of their checkout of `ouroboros-consensus`.

edsko:
Closes https://github.com/input-output-hk/ouroboros-network/issues/2463 .

PR #2468, opened 2020-07-27
ErrorPolicy: ProtocolLimitFailure

coot:


coot:
As advised by @njd42 We'll wait with merging after hard fork; we have a proof now that the system can recovers now.

coot:
bors merge

PR #2469, opened 2020-07-27
Shelley: reduce per transaction overhead and add fixed overhead

mrBliss:
Also assert that we never underestimate the size of the block body.


mrBliss:
bors merge

mrBliss:
bors merge

PR #2475, opened 2020-07-28
tip-sample appliction

coot:
This PR implements tip-sample application and `TipRegistry` which allows to
extract resutlts.

- tip-sample: TipFragment
- tip-sample: client application
- tip-sample: server application
- tip-sample: TipRegistry
- tip-sample: TipFragment tests
- tip-sample: TipRegistry tests
- ChainFragment: export lookupBySlotFT
- tip-sample: tipSampleClient integration test

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/TipSample/Server.hs
+++ b/ouroboros-network/src/Ouroboros/Network/TipSample/Server.hs
@@ -1,0 +1,65 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Ouroboros.Network.TipSample.Server where
+
+import           Control.Monad.Class.MonadSTM
+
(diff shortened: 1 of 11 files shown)


ouroboros-network/src/Ouroboros/Network/TipSample/Server.hs
```diff
@@ -0,0 +1,65 @@
...
+import           Ouroboros.Network.Block (Tip (..))
+import           Ouroboros.Network.Protocol.TipSample.Server
+
+
+-- | 'tip-sample' server application.
+--
+tipServer :: forall header m.
+             MonadSTM m
+          => STM m (Tip header)
+          -- ^ 'ChainDB' 'getCurrentSlot' method.
```
karknu, line 23: Is the comment or the variable name correct?
That is is this a `getCurrentSlot` method or a `getCurrentTip` method?

coot: it should be `getCurrentTip`.

coot:
```suggestion
          -- ^ 'ChainDB' 'getCurrentTip' method.
```

PR #2481, opened 2020-07-30
Improve nodeBlockFetchSize estimates and test them

### LINKED ISSUE #2480: Shelley: incorrect nodeBlockFetchSize
@karknu Reported that the first Shelley block on mainnet has a size of 5 bytes according to `nodeBlockFetchSize`:
```
2020-07-29 21:45:06.715300804 UTC: ThreadId 256: ConnectionId {localAddress = 192.168.1.134:40009, remoteAddress = 18.177.103.105:3001} CompletedBlockFetch responseTime 0.250252965s size 5 ChainRange from At (SlotNo {unSlotNo = 4492800}) to At (SlotNo {unSlotNo = 4492800})
```

That's because we mistakenly report the block *body* size instead of the block size:
https://github.com/input-output-hk/ouroboros-network/blob/8a1fd0659189358e145005d10a4dbf77cbe6c02f/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs#L348

The `nodeBlockFetchSize` is used by the BlockFetchClient to know how many bytes to expect when downloading the block corresponding to the header. Until very recently, we were hard-coding it to 2000 for Byron, so we are not enforcing it to be correct (yet).

mrBliss:
Fixes #2480.

Return the block size instead of the body size for Shelley.

Take CBOR-in-CBOR and the EBB tag into account for Byron.

Include a test in the serialisation roundtrip test skeleton.


mrBliss:
bors merge

ouroboros-consensus/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block.hs
```diff
@@ -112,8 +112,19 @@ data SimpleBlock' c ext ext' = SimpleBlock {
...
+  encode (SimpleBlock hdr body) = mconcat [
+        CBOR.encodeListLen 2
+      , encode hdr
+      , encode body
+      ]
+  decode = do
+      CBOR.decodeListLenOf 2
+      hdr  <- decode
+      body <- decode
+      return (SimpleBlock hdr body)
```
mrBliss, line 127: I went for this instead of the generic instance so it's easier to estimate the size of the block.

mrBliss:
bors merge

mrBliss:
bors merge

mrBliss:
(Yes, I'm using CI to run my testsuites. I compiled locally with `-O0` which is too slow to run all the tests.)

mrBliss:
bors merge

PR #2484, opened 2020-07-31
Upper bound for CBOR-in-CBOR tag is 7 instead of 5 bytes

mrBliss:
Apparently we sometimes generate blocks that are 65536 bytes large or more,
requiring a 7 byte CBOR-in-CBOR tag (65535 would only require 5 bytes), failing
the test added in #2481, as our upper bound assumed 5 bytes.


mrBliss:
bors merge

PR #2489, opened 2020-07-31
 ChainDB: write sync to VolDB, async chain selection 

### LINKED ISSUE #2487: Chain DB: addBlock queue ineffective
@karknu has discovered that the BlockFetch client is able to quickly add the first block of the Shelley era to the ChainDB, but the second one takes 11s!

```
MsgRequestRange ChainRange from At (SlotNo {unSlotNo = 4492800}) to At (SlotNo {unSlotNo = 4492800})
MsgStartBatch ChainRange from At (SlotNo {unSlotNo = 4492800}) to At (SlotNo {unSlotNo = 4492800})
MsgBlock recved
MsgBlock verified after 0.000023s
MsgBlock written to disk after 0.085022s (0.085045s)
CompletedBlockFetch responseTime 0.172265s size 1013 ChainRange from At (SlotNo {unSlotNo = 4492800}) to At (SlotNo {unSlotNo = 4492800})

MsgRequestRange ChainRange from At (SlotNo {unSlotNo = 4492840}) to At (SlotNo {unSlotNo = 4492840})
MsgStartBatch ChainRange from At (SlotNo {unSlotNo = 4492840}) to At (SlotNo {unSlotNo = 4492840})
MsgBlock recved
MsgBlock verified after 0.000019s
MsgBlock written to disk after 11.361727s (11.361746s)
CompletedBlockFetch responseTime 11.522864s size 1013 ChainRange from At (SlotNo {unSlotNo = 4492840}) to At (SlotNo {unSlotNo = 4492840})
```

One might think that the first block was quick to validate, but the second slow. This would be counter-intuitive because the first block triggers the transition, requiring an expensive translation of the ledger state. Validating the second block should be quick.

Note that when the BlockFetch client adds a block to the ChainDB, it should only block until the block has been written to disk, not until chain selection has been performed for that block.

With some extra tracing, we see:
```
addFetchedBlock At (SlotNo {unSlotNo = 4492800}) 0.096076772s
chainSelection: 16.176468239s
addFetchedBlock At (SlotNo {unSlotNo = 4492840}) 15.617731605s
```
This means that actually the chain selection for the first block is taking long, not the second. Adding the second block is blocked by the first block still being processed.

What's actually going on is the following:

* The BlockFetch client adds the blocks it downloaded to a queue in the ChainDB with a maximum size of 10. This maximum size is there to provide back pressure, otherwise the number of blocks in memory could grow without bound.
* A background thread in the ChainDB processes the blocks in this queue one by one in a loop. In each iteration of the loop, it writes the block to disk (unless we intentionally ignore it), delivers [the promise](https://github.com/input-output-hk/ouroboros-network/blob/04a02d9130f79b876ab0907e5a63b358123fd5a1/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs#L365) the BlockFetch client is waiting on, then performs chain selection for the block, and finally delivers another [promise](https://github.com/input-output-hk/ouroboros-network/blob/04a02d9130f79b876ab0907e5a63b358123fd5a1/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs#L378) indicating the block has been processed.
* While the background thread is still doing chain selection for the first block, the second block has been added to the queue and the BlockFetch client is waiting for it to be written to disk, which won't happen until after the first block has been fully processed.

This means that the effective overlap or pipelining is limited to 1 block, not the configured 10.

To fix this, there could be a separate queue for each step, i.e., one for writing blocks to disk and one for doing chain selection for blocks.

However, the more queues, the more time lost on synchronising things and overhead. The shorter the actual steps (writing to disk, chain selection) take, the more overhead there will be. So adding the extra queue is not guaranteed to speed things up in *all cases*. Likely for this case, but for bulk chain sync of mostly empty Byron blocks, it *might* slow things down.

My plan: in practice we *always* wait for the block to be written to disk, we never want to just add the block to the queue without any extra waiting. We can synchronously add the block to the VolatileDB and *only then* add the block to the queue with blocks awaiting chain selection. This would also allow reordering out of order blocks in that queue using, e.g., an `OrdPSQ`.

mrBliss:
Fixes #2487.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
@@ -15,7 +15,6 @@
 module Ouroboros.Consensus.Storage.ChainDB.Impl.ChainSel
   ( initialChainSelection
   , addBlockAsync
-  , addBlockSync
   , chainSelectionForBlock
     -- * Exported for testing purposes
   , olderThanK
@@ -213,22 +212,6 @@ initialChainSelection immDB volDB lgrDB tracer cfg varInvalid varFutureBlocks
 -- the next startup, a correct in-memory state will be reconstructed from the
 -- file system state.
 addBlockAsync
-  :: forall m blk. (IOLike m, HasHeader blk)
-  => ChainDbEnv m blk
-  -> blk
-  -> m (AddBlockPromise m blk)
-addBlockAsync CDB { cdbTracer, cdbBlocksToAdd } =
-    addBlockToAdd (contramap TraceAddBlockEvent cdbTracer) cdbBlocksToAdd
-
--- | Add a block to the ChainDB, /synchronously/.
---
--- This is the only operation that actually changes the ChainDB. It will store
--- the block on disk and trigger chain selection, possibly switching to a
--- fork.
---
--- When the slot of the block is > the current slot, a chain selection will be
--- scheduled in the slot of the block.
-addBlockSync
   :: forall m blk.
      ( IOLike m
      , GetPrevHash blk
@@ -236,12 +219,11 @@ addBlockSync
      , InspectLedger blk
      , HasHardForkHistory blk
      , VolDbSerialiseConstraints blk
-     , HasCallStack
      )
   => ChainDbEnv m blk
-  -> BlockToAdd m blk
-  -> m ()
-addBlockSync cdb@CDB {..} BlockToAdd { blockToAdd = b, .. } = do
+  -> blk
+  -> m (AddBlockPromise m blk)
+addBlockAsync cdb@CDB {..} b = do
     (isMember, invalid, curChain) <- atomically $ (,,)
       <$> VolDB.getIsMember               cdbVolDB
       <*> (forgetFingerprint <$> readTVar cdbInvalid)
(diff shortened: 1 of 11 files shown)


mrBliss:
This doesn't even compile, but I'm dumping this here before my leave, so it won't be forgotten.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -408,59 +408,51 @@ data InvalidBlockInfo blk = InvalidBlockInfo
...
--- to implement 'AddBlockPromise'.
-data BlockToAdd m blk = BlockToAdd
-  { blockToAdd            :: !blk
-  , varBlockWrittenToDisk :: !(StrictTMVar m Bool)
-    -- ^ Used for the 'blockWrittenToDisk' field of 'AddBlockPromise'.
-  , varBlockProcessed     :: !(StrictTMVar m (Point blk))
+-- | FIFO queue used to perform chain selection on blocks asynchronously.
+-- Blocks are read from this queue by a background thread, which performs
+-- chain selection for each block synchronously.
+newtype ChainSelectionQueue m blk = ChainSelectionQueue (TBQueue m (BlockToProcess m blk))
```
mrBliss, line 417: To automatically reorder blocks that arrived out of order, we could, for example, turn this into a `(T)(M)Var` containing an `OrdPSQ` ordered on the `SlotNo`.

mrBliss: Actually, something simple like a `Map` (ordered by the `RealPoint`s) or just a list + `insertBy` is already enough.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -444,6 +419,15 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr = do
...
 
     if
+      -- If the block is already in the current chain, ignore it. As we add
+      -- blocks to the VolatileDB before adding them to the chain selection
+      -- queue, it can often happen that when we did chain selection for a
+      -- block, we already extended its candidate chain with its successors
+      -- (that are in the queue) as they are already part of the VolatileDB.
+      | AF.withinFragmentBounds (blockPoint hdr) curChain -> do
+        trace $ IgnoreBlockAlreadyInChain p
+        return tipPoint
```
mrBliss, line 429: Without this case, this PR would significantly slow down sync speeds (certainly for Byron), as this happens very often, resulting in the expensive `TrySwitchToAFork` case.

mrBliss:
@karknu It would be great if you do some benchmarks with and without this PR, to see whether this actually improves sync speed.

I'm not expecting miracles, but for Shelley we'll now have a queue of 10 blocks instead of just one, so that should help smooth over the very slow initial Shelley block. This might also help with the slow end-of-epoch blocks.

We must also make sure this doesn't accidentally slow down bulk syncing in the Byron era. I fear that we will sometimes hit the `TrySwitchToAFork` case when the queue is full, causing a slow down.

As mentioned in one of my inline comments, this PR can help with blocks arriving out-of-order, but I haven't tested this at all.

karknu:
@mrBliss I will take a look at it tomorrow.

mrBliss:
@kderme observed that this PR again adds blocks concurrently to the VolatileDB, whereas we previously did it from a single thread. We have to think about the consequences of this.

mrBliss:
Obsoleted by #2721.

PR #2490, opened 2020-07-31
Update cardano-ledger-specs

nfrisby:
The first commit integrates with the upstream PRs input-output-hk/cardano-ledger-specs#1735 and input-output-hk/cardano-ledger-specs#1739, which generalize the contexts and instances necessary for using the cardano-ledger-specs test case generators.

The second commit removes a workaround. PR input-output-hk/cardano-ledger-specs#1731 changed the Shelley transaction generator to no longer fail with `QC.discard` and moreover to attempt a given number of retries before failing fatally. That supplants our temporary workaround.

~It also generalizes the `TPraosMockCrypto` definition to allow for a separate address hashing algorithm, which in turn allows Test.Util.ThreadNet.Cardano to satisfy the assumption currently made by the Byron-to-Shelley translation that the key hashes in the bootstrap addresses have the same bit length as those in normal Shelley addresses.~

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
@@ -54,9 +54,6 @@ import qualified Data.Set as Set
 import qualified Data.Typeable as Typeable
 import           Data.Void (Void)
 import           GHC.Stack
-import qualified Test.QuickCheck.Exception as QC
-
-import           Cardano.Prelude (forceElemsToWHNF)
 
 import qualified Ouroboros.Network.AnchoredFragment as AF
 import           Ouroboros.Network.BlockFetch (BlockFetchConfiguration (..))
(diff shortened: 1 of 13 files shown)


nfrisby:
I just now rebased onto `origin/master`.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/Network.hs
```diff
@@ -640,38 +637,10 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
-                --
-                -- Revisit this scheme after Issue
-                -- input-output-hk/cardano-ledger-specs#1689
-                attemptLimit = 1000 :: Int
-
-        txs <- loop (1 :: Int)
+        let txs = runGen
+                (nodeSeed `combineWith` unSlotNo curSlotNo)
+                (testGenTxs numCoreNodes curSlotNo cfg txGenExtra ledger)
+
```
mrBliss, line 643: :relieved:

mrBliss:
> I just now rebased onto `origin/master`.

Bors when you want

Issue #2460, opened 2020-07-23
Block-specific arguments for db-analyser and db-validator


mrBliss:
The `ProtocolInfo` for Byron, Shelley, and Cardano hard-code some parameters:
```
    protocolInfoShelley
      genesis
      NeutralNonce -- TODO -- <-
      2000
      (SL.ProtVer 0 0) -- <-
      Nothing
```
In particular: `(Just $ PBftSignatureThreshold 0.22)` for Byron, which is different from the value used on `mainnet_candidate` (0.9), making it impossible to use these tools for `mainnet_candidate` or any other nets that use different parameters than mainnet.

We already have two Byron-specific arguments:
```
  , requiresNetworkMagic :: w ::: Bool            <?> "Expecto patronum?"
  , genesisHash          :: w ::: Hash CB.Raw     <?> "Expected genesis hash"
```
We shouldn't keep adding all arguments in one place, we should start making such arguments specific to the block type. We can use [`optparse-applicative` *commands*](https://github.com/pcapriotti/optparse-applicative#commands) for this.

Additionally: there is quite a bit of code duplicated between the two, e.g., the `protocolMagicId` business, `BlockType`, the parser for block-specific configuration arguments, etc. Common parts should become library code shared between the two executables.

kderme:
Maybe related to the hard-coded params, I sometimes get errors like
```
TraceInitChainSelEvent (InitChainSelValidation (InvalidBlock (ExtValidationErrorHeader (HeaderProtocolError 
(HardForkValidationErrFromEra S (Z (WrapValidationErr {unwrapValidationErr = ChainTransitionError [OverlayFailure (VRFKeyBadNonce (Nonce "81e47a19e6b29b0a65b9591762ce5143ed30d0261e5d24a3201752506b20f15c") (SlotNo {unSlotNo 
= 4492800}) (Nonce "81e47a19e6b29b0a65b9591762ce5143ed30d0261e5d24a3201752506b20f15c") (CertifiedVRF 
{certifiedOutput = OutputVRF {getOutputVRFBytes = 
"6\236Sx\209\245\EOT\SUBY\235\141\150\230\GS\233o\tP\251A\180\159\245\DC1\247\188\DEL\209\t\212\&8>\GS$\190p4\23
0t\156f\DC2p\r\213\206\176\198ew\184\138\EM\174(k\DC3!\209[\206\SUB\183\&6"}, certifiedProof = CertPraosVRF 
[...]

PR #2493, opened 2020-08-03
Use commads for db-validator

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/2460

~This probably breaks a nightly test.~ fixed

An example to use it with mainnet is the following:
```
cabal new-exec db-validator -- --db /path/to/db cardano 
--genesisHash 5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb 
--configByron /path/to/configuration/mainnet/mainnet-byron-genesis.json --threshold 0.22 
--configShelley /path/to/configuration/mainnet/mainnet-shelley-genesis.json 
--nonce 1a3be38bcbb7911969283716ad7aa550250226b76a61fc51cc9a9a35d9276d81 --verbose

```

PR #2501, opened 2020-08-10
Merge db-validator and db-analyser

kderme:
Continues https://github.com/input-output-hk/ouroboros-network/pull/2493 and closes https://github.com/input-output-hk/ouroboros-network/issues/2460

With this pr we merge the two executables: db-validator and db-analyser into one. They had already a lot of code duplication, so this simplifies things a lot. The executable works by default like a validator, unless an analsisName flag is enabled (ie --show-slot-block-no).


ouroboros-consensus-cardano/tools/db-validator/Main.hs
```diff
@@ -40,25 +32,45 @@ import qualified Ouroboros.Consensus.Storage.ChainDB as ChainDB
...
+    case blockType cmdLine of
+      ByronBlock   args -> validateOrAnalyse cmdLine args
+      ShelleyBlock args -> validateOrAnalyse cmdLine args
+      CardanoBlock args -> validateOrAnalyse cmdLine args
+
+warnUnusedFlags :: CmdLine -> IO ()
+warnUnusedFlags CmdLine{..} = case (analysis, onlyImmDB) of
+  (Just _, True)
+    -> putStrLn "Warning: Flag onlyImmDB is ignored on analysis mode"
+  _ -> return ()
```
mrBliss, line 55: Actually, couldn't we just let analyses take this flag into account too? In other words, analyse the ImmDB or the ChainDB?

ouroboros-consensus-cardano/tools/db-validator/Class.hs
```diff
@@ -1,11 +1,17 @@
...
-module Analysis (HasAnalysis (..), mkTopLevelConfig) where
+{-# LANGUAGE FlexibleInstances      #-}
+{-# LANGUAGE NamedFieldPuns         #-}
+{-# LANGUAGE OverloadedStrings      #-}
+{-# LANGUAGE RecordWildCards        #-}
+{-# LANGUAGE ScopedTypeVariables    #-}
+{-# LANGUAGE TypeApplications       #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+
+module Class (
```
mrBliss, line 9: I'm not a fan of the name `Class`, it's rather vague, it contains type classes, I suppose.

What if you spread the instances over `Block.Byron`, `Block.Shelley`, and `Block.Cardano`?

ouroboros-consensus-cardano/tools/db-validator/Main.hs
kderme: True. I've been working on extending the analyser for the whole chainDB and make this flag useful.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -167,3 +167,4 @@ executable db-validator
                        -fno-ignore-asserts
                        -threaded
                        -rtsopts
+                       "-with-rtsopts=-T -I0 -N2 -A16m"
```
kderme, line 170: This is now same as cardano-node. It allows similar results, when the db-validator is used to count performance.

ouroboros-consensus-cardano/tools/db-validator/Main.hs
```diff
@@ -86,9 +89,22 @@ parseCmdLine = CmdLine
             long "onlyImmDB"
           , help "Validate only the immutable DB (e.g. do not do ledger validation)"
           ])
+    <*> parseValidationPolicy
```
mrBliss, line 92: Good idea, I ran into this too the last few days.

ouroboros-consensus-cardano/tools/db-validator/Block/Byron.hs
```diff
@@ -0,0 +1,122 @@
...
+    mkProtocolInfo ByronBlockArgs {..} = do
+      config <- openGenesisByron configFileByron genesisHash requiresNetworkMagic
+      return $ mkByronProtocolInfo config threshold
+    countTxOutputs = aBlockOrBoundary (const 0) countTxOutputsByron
+    blockHeaderSize = fromIntegral .
+      aBlockOrBoundary blockBoundaryHeaderSize blockHeaderSizeByron
+    blockTxSizes = aBlockOrBoundary (const []) blockTxSizesByron
+    knownEBBs = const Byron.knownEBBs
+
+type ByronBlockArgs = Args ByronBlock
```
mrBliss, line 57: ~Is this type used/needed? It saves just one character, that's not worth the extra indirection for me. It's not that we have to be careful to maintain backwards compatibility. Moreover, it's easier to align multiple lines of `Args x` than `xArgs` :slightly_smiling_face:.~

UPDATE: I just the similar synonyms for Shelley and Cardano, which do save a lot of characters. Then we should keep this one too, for consistency, of course. So you can ignore this comment.

ouroboros-consensus-cardano/tools/db-validator/HasAnalysis.hs
```diff
@@ -0,0 +1,25 @@
...
+import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.Node.ProtocolInfo
+import           Ouroboros.Consensus.Storage.ChainDB.Serialisation (SizeInBytes)
+
+{-------------------------------------------------------------------------------
+  HasAnalysis
+-------------------------------------------------------------------------------}
+
+class GetPrevHash blk => HasAnalysis blk where
+    data Args blk
```
mrBliss, line 20: Thinking out loud: what if we add:
```haskell
    argsParser :: proxy blk -> Parser (Args blk)
```
What do you think?

ouroboros-consensus-cardano/tools/db-validator/Main.hs
mrBliss: Do you intend to include that in this PR before it's merged?

mrBliss: Idea: if we can make `analyse` work with the ChainDB too, then validation would be just another type of analysis, a no-op analysis. This means we can truly merge `analyse` and `validate`! In which case, I would call the tool `db-analyser` instead of `db-validator`. Great, let's do this, this perfectly matches the title of this PR :slightly_smiling_face:.

ouroboros-consensus-cardano/tools/db-analyser/Block/Byron.hs
```diff
@@ -0,0 +1,152 @@
...
+
+mkByronProtocolInfo :: Genesis.Config
+                    -> Maybe PBftSignatureThreshold
+                    -> ProtocolInfo IO ByronBlock
+mkByronProtocolInfo genesisConfig signatureThreshold =
+    protocolInfoByron
+      genesisConfig
+      signatureThreshold
+      (Update.ProtocolVersion 1 0 0)
+      (Update.SoftwareVersion (Update.ApplicationName "db-analyse") 2)
```
mrBliss, line 151:
```suggestion
      (Update.SoftwareVersion (Update.ApplicationName "db-analyser") 2)
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,109 @@
...
+      flag' VolDB.NoValidation $ mconcat [
+          long "vol-no-validation"
+        , help "Don't validate blocks of the Volatile db"
+        ]
+    , flag' VolDB.ValidateAll $ mconcat [
+          long "vol-validate-all"
+        , help "Validate all blocks of the Volatile db"
         ]
-    , pure Nothing
     ]
```
mrBliss, line 107: Let's simplify this into `parseFullValidation :: Parser Bool`, which defaults to `False`, but when `--full-validation` is specified, the `ImmDB.ValidateMostRecentChunk` and `VolDB.ValidateAll` policies are used (similar to [what we do for the real node](https://github.com/input-output-hk/ouroboros-network/blob/aecfe77de784e6c75108172802ee59d2d7087eaf/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L184)). You can use "Validate all blocks on disk" as the `help` text.

Hmm, I'd still like to default it to `True` in case of `OnlyValidation` without losing the ability to disable it explicitly. Maybe we need `Maybe True` for that, with `Nothing` as default and `--no-full-validation` for `Just False`.

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -326,35 +116,80 @@ parseAnalysis = asum [
...
+          chainDbArgs = args' {
+              ChainDB.cdbImmValidation = immValidationPolicy
+            , ChainDB.cdbVolValidation = volValidationPolicy
+            }
+          (immDbArgs, _, _, _) = fromChainDbArgs chainDbArgs
+      if onlyImmDB then
+        ImmDB.withImmDB immDbArgs $ \immDB -> do
+          runAnalysis analysis cfg (Left immDB) registry
+          immDbTipPoint <- ImmDB.getPointAtTip immDB
+          putStrLn $ "DB tip: " ++ show immDbTipPoint
```
mrBliss, line 174:
```suggestion
          putStrLn $ "ImmDB tip: " ++ show immDbTipPoint
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -326,35 +116,80 @@ parseAnalysis = asum [
...
+      else
+        ChainDB.withDB chainDbArgs $ \chainDB -> do
+          runAnalysis analysis cfg (Right chainDB) registry
+          chainDbTipPoint <- atomically $ ChainDB.getTipPoint chainDB
+          putStrLn $ "DB tip: " ++ show chainDbTipPoint
```
mrBliss, line 179:
```suggestion
          putStrLn $ "ChainDB tip: " ++ show chainDbTipPoint
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
-          ])
-    <*> strOption (mconcat [
+    <$> strOption (mconcat [
             long "db"
-          , help "Path to the chain DB (parent of \"immutable\" directory)"
+          , help "Path to the chain DB"
           , metavar "PATH"
           ])
-    <*> parseBlockType
+    <*> flag False True (mconcat [
```
mrBliss, line 73: Use [`switch`](http://hackage.haskell.org/package/optparse-applicative-0.16.0.0/docs/Options-Applicative.html#v:switch), same for the flag below.

mrBliss (approved):

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -326,35 +116,108 @@ parseAnalysis = asum [
...
+    <*> parseMaybe (PBftSignatureThreshold <$> thresholdParser)
+  where
+    thresholdParser = option auto (mconcat [
+            long "threshold"
+          , help "PBftSignatureThreshold"
+          , metavar "THRESHOLD"
+          ])
+=======
+parseByronType = ByronBlock <$> argsParser Proxy
+>>>>>>> a6370e735... Make Parser part of the HasAnalysis
```
mrBliss, line 158: Merge conflict

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
-          ]))
-    <*> flag True False (mconcat [
-            long "testnet"
-          , help "The DB contains blocks from testnet rather than mainnet"
-          ])
-    <*> strOption (mconcat [
+    <$> strOption (mconcat [
             long "db"
-          , help "Path to the chain DB (parent of \"immutable\" directory)"
+          , help "Path to the chain DB"
```
mrBliss, line 70:
```suggestion
          , help "Path to the Chain DB"
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
           , metavar "PATH"
           ])
-    <*> parseBlockType
+    <*> switch (mconcat [
+            long "verbose"
+          , help "Enable verbose logging"
+          ])
+    <*> switch (mconcat [
+            long "onlyImmDB"
+          , help "Validate only the immutable DB (e.g. do not do ledger validation)"
```
mrBliss, line 79:
```suggestion
          , help "Validate only the Immutable DB (e.g. do not do ledger validation)"
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
-        , help "A Byron network"
-        ]
-    , flag' (Just Shelley) $ mconcat [
-          long "shelley"
-        , help "A Shelley network"
+parseValidationPolicy :: Parser (Maybe ValidateBlocks)
+parseValidationPolicy = parseMaybe $ asum [
+      flag' ValidateAll $ mconcat [
+          long "validate-all"
+        , help "Validate all blocks of the volatile and immutable db"
```
mrBliss, line 89:
```suggestion
        , help "Validate all blocks of the Volatile and Immutable DB"
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
         ]
-    , flag' (Just Cardano) $ mconcat [
-          long "cardano"
-        , help "A Byron-to-Shelley network"
+    , flag' MinimumValidation $ mconcat [
+          long "minimum-validation"
+        , help "Validate a minimum part of the volatile and immutable db"
```
mrBliss, line 93:
```suggestion
        , help "Validate a minimum part of the Volatile and Immutable DB"
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -326,35 +116,80 @@ parseAnalysis = asum [
...
         , help "Show all EBBs and their predecessors"
         ]
+    , pure OnlyValidation
     ]
 
+blockTypeParser :: Parser BlockType
+blockTypeParser = subparser $ mconcat
+  [ command "byron"   (info (parseByronType   <**> helper) (progDesc "byron command"  ))
+  , command "shelley" (info (parseShelleyType <**> helper) (progDesc "shelley command"))
+  , command "cardano" (info (parseCardanoType <**> helper) (progDesc "cardano command"))
```
mrBliss, line 126: I'm trying out the tool and with `--help`, it looks like this:
```
[..]
Available commands:
  byron                    byron command
  shelley                  shelley command
  cardano                  cardano command
```
Better would be:
```
[..]
Available commands:
  byron                    Analyse a Byron-only DB
  shelley                  Analyse a Shelley-only DB
  cardano                  Analyse a Cardano DB
```

ouroboros-consensus-cardano/tools/db-analyser/Block/Shelley.hs
```diff
@@ -0,0 +1,85 @@
...
+      Nothing
+
+countOutputs :: Shelley.Crypto c => SL.Tx c -> Int
+countOutputs tx = length $ SL._outputs $ SL._body tx
+
+parseShelleyArgs :: Parser ShelleyBlockArgs
+parseShelleyArgs = ShelleyBlockArgs
+    <$> strOption (mconcat [
+            long "configShelley"
+          , help "Path to config file."
```
mrBliss, line 75:
```suggestion
          , help "Path to config file"
```

ouroboros-consensus-cardano/tools/db-analyser/Block/Shelley.hs
```diff
@@ -0,0 +1,85 @@
...
+          , metavar "PATH"
+          ])
+    <*> asum [ Nonce  <$> parseNonce
+             , pure NeutralNonce]
+  where
+    parseNonce = strOption (mconcat [
+            long "nonce"
+          , help "initial nonce"
```
mrBliss, line 83:
```suggestion
          , help "Initial nonce, i.e., hash of the genesis config file"
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
-          long "byron"
-        , help "A Byron network"
-        ]
-    , flag' (Just Shelley) $ mconcat [
-          long "shelley"
-        , help "A Shelley network"
+parseValidationPolicy :: Parser (Maybe ValidateBlocks)
+parseValidationPolicy = parseMaybe $ asum [
+      flag' ValidateAll $ mconcat [
+          long "validate-all"
```
mrBliss, line 88:
```suggestion
          long "validate-all-blocks"
```
Otherwise it might be confusing to users that want to use this tool to do ledger validation.

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
-      flag' (Just Byron) $ mconcat [
-          long "byron"
-        , help "A Byron network"
-        ]
-    , flag' (Just Shelley) $ mconcat [
-          long "shelley"
-        , help "A Shelley network"
+parseValidationPolicy :: Parser (Maybe ValidateBlocks)
+parseValidationPolicy = parseMaybe $ asum [
+      flag' ValidateAll $ mconcat [
```
mrBliss, line 87:
```suggestion
      flag' ValidateAllBlocks $ mconcat [
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
+          long "validate-all"
+        , help "Validate all blocks of the volatile and immutable db"
         ]
-    , flag' (Just Cardano) $ mconcat [
-          long "cardano"
-        , help "A Byron-to-Shelley network"
+    , flag' MinimumValidation $ mconcat [
```
mrBliss, line 91:
```suggestion
    , flag' MinimumBlockValidation $ mconcat [
```

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,307 +1,97 @@
...
+          long "minimum-validation"
```
mrBliss, line 92:
```suggestion
          long "minimum-block-validation"
```

mrBliss (commented):
I have been trying out the new tool and found some more nitpicks :grin:. If you prefer, I can fix them myself, let me know.

BTW, most of the commits should be squashed before merging.

kderme:
> I have been trying out the new tool and found some more nitpicks . If you prefer, I can fix them myself, let me know.
> 
> BTW, most of the commits should be squashed before merging.

They seem reasonable. I'm fixing them and squashing

mrBliss:
Great! I think we're ready to bors

kderme:
bors merge

PR #2502, opened 2020-08-10
Extend Cardano ThreadNet tests so that Shelley era contains active stake pools

### LINKED ISSUE #2388: Introduce registered stakepools early in the Shelley era in Cardano ThreadNet tests
At the moment, the tests fail with a (de)centralization parameter of `d=0`, because at the moment, only the overlay slots of the Shelley era have leaders.

nfrisby:
Fixes #2388.

The big commit could be decomposed a little, but the excess diff I'm thinking of is mostly minor (whitespace/renaming/adding one param to `testGenTxs` etc) and localized -- it's obvious when you see it. Please ping me if the decomposition would be worthwhile by making your review that much easier.

cc: @edsko FYI

### DIFF
--- a/ouroboros-consensus-cardano/test/Test/Consensus/Cardano/MockCrypto.hs
+++ b/ouroboros-consensus-cardano/test/Test/Consensus/Cardano/MockCrypto.hs
@@ -1,0 +1,45 @@
+{-# LANGUAGE DataKinds    #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+module Test.Consensus.Cardano.MockCrypto (
+    BlockCompatByron
+  , TPraosMockCryptoCompatByron
+  ) where
+
+import           Cardano.Crypto.DSIGN (Ed25519DSIGN)
+import           Cardano.Crypto.Hash (Blake2b_224, HashAlgorithm)
+import           Cardano.Crypto.KES (MockKES)
+
+import           Test.Cardano.Crypto.VRF.Fake (FakeVRF)
+
+import           Ouroboros.Consensus.Shelley.Ledger (ShelleyBlock)
+import           Ouroboros.Consensus.Shelley.Protocol.Crypto (TPraosCrypto)
+
+import           Shelley.Spec.Ledger.Crypto (Crypto (..))
+
+-- | A mock replacement for 'TPraosStandardCrypto' that is compatible with
+-- bootstrapping from Byron
+--
+-- This mocks more components than does
+-- 'Test.Consensus.Shelley.MockCrypto.TPraosMockCrypto'. This prevents the
+-- @cardano-ledger-specs@ generators from being re-used. Currently, this is not
+-- an obstacle for example in the Cardano ThreadNet tests.
+--
+-- NOTE: The "Ouroboros.Consensus.Cardano.CanHardFork" translation currently
+-- assumes that @ADDRHASH@ has the same bit size as Byron address hashes (ie
+-- 224); that's why we use 'Blake2b_224' here.
+--
+-- NOTE: The @shelley-spec-ledger@ package currently requires that @'DSIGN' ~
+-- 'Ed25519DSIGN' in order to use Byron bootstrap witnesses.
+data TPraosMockCryptoCompatByron h
+
+instance HashAlgorithm h => Crypto (TPraosMockCryptoCompatByron h) where
+  type ADDRHASH (TPraosMockCryptoCompatByron h) = Blake2b_224
+  type DSIGN    (TPraosMockCryptoCompatByron h) = Ed25519DSIGN
+  type HASH     (TPraosMockCryptoCompatByron h) = h
+  type KES      (TPraosMockCryptoCompatByron h) = MockKES 10
+  type VRF      (TPraosMockCryptoCompatByron h) = FakeVRF
+
+instance HashAlgorithm h => TPraosCrypto (TPraosMockCryptoCompatByron h)
+
+type BlockCompatByron h = ShelleyBlock (TPraosMockCryptoCompatByron h)
--- a/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
+++ b/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
@@ -13,9 +13,15 @@ module Test.ThreadNet.Cardano (
 
 import           Control.Exception (assert)
 import           Control.Monad (guard, replicateM)
+import           Control.Monad.Identity (runIdentity)
+import           Control.Monad.Reader (runReaderT)
+import           Data.Functor ((<&>))
 import           Data.List ((!!))
 import qualified Data.Map as Map
+import           Data.Maybe (isJust, maybeToList)
 import           Data.Proxy (Proxy (..))
+import           Data.Set (Set)
+import qualified Data.Set as Set
 import           Data.Word (Word64)
 import           GHC.Generics (Generic)
 import           Numeric.Natural (Natural)
@@ -24,11 +30,11 @@ import           Test.QuickCheck
 import           Test.Tasty
 import           Test.Tasty.QuickCheck
 
-import           Cardano.Slotting.Slot (EpochNo, EpochSize (..), SlotNo (..))
+import           Cardano.Slotting.EpochInfo
+import           Cardano.Slotting.Slot (EpochNo (..), EpochSize (..),
+                     SlotNo (..))
 
 import           Cardano.Crypto.Hash.Blake2b (Blake2b_256)
-
-import qualified Ouroboros.Network.MockChain.Chain as MockChain
 
 import           Ouroboros.Consensus.BlockchainTime
 import           Ouroboros.Consensus.Config.SecurityParam
@@ -43,7 +49,10 @@ import           Ouroboros.Consensus.Util.IOLike (IOLike)
 import           Ouroboros.Consensus.HardFork.Combinator.Serialisation.Common
                      (isHardForkNodeToNodeEnabled)
 
+import qualified Cardano.Chain.Common as CC.Common
 import qualified Cardano.Chain.Genesis as CC.Genesis
+import           Cardano.Chain.ProtocolConstants (kEpochSlots)
+import           Cardano.Chain.Slotting (unEpochSlots)
 import qualified Cardano.Chain.Update as CC.Update
 
 import           Ouroboros.Consensus.Byron.Ledger.Block (ByronBlock)
@@ -53,8 +62,10 @@ import           Ouroboros.Consensus.Byron.Node
 import qualified Shelley.Spec.Ledger.BaseTypes as SL
 import qualified Shelley.Spec.Ledger.Genesis as SL
 import qualified Shelley.Spec.Ledger.OCert as SL
+import qualified Shelley.Spec.Ledger.OverlaySchedule as SL
 import qualified Shelley.Spec.Ledger.PParams as SL
 
+import           Ouroboros.Consensus.Shelley.Ledger (mkShelleyGlobals)
 import           Ouroboros.Consensus.Shelley.Node
 import qualified Ouroboros.Consensus.Shelley.Protocol as Shelley
 import           Ouroboros.Consensus.Shelley.Protocol.Crypto (KES, TPraosCrypto)
@@ -69,7 +80,7 @@ import qualified Test.ThreadNet.Infra.Byron as Byron
 import qualified Test.ThreadNet.Infra.Shelley as Shelley
 import           Test.ThreadNet.Network (CalcMessageDelay (..), NodeOutput (..),
                      TestNodeInitialization (..))
-import           Test.ThreadNet.TxGen.Cardano ()
+import           Test.ThreadNet.TxGen.Cardano (CardanoTxGenExtra (..))
 import           Test.ThreadNet.Util.Expectations (NumBlocks (..))
 import           Test.ThreadNet.Util.NodeJoinPlan (trivialNodeJoinPlan)
 import           Test.ThreadNet.Util.NodeRestarts (noRestarts)
@@ -94,7 +105,8 @@ data Partition = Partition SlotNo NumSlots
   deriving (Show)
 
 partitionExclusiveUpperBound :: Partition -> SlotNo
-partitionExclusiveUpperBound (Partition s (NumSlots d)) = Util.addSlots d s
+partitionExclusiveUpperBound (Partition s (NumSlots dur)) =
+    Util.addSlots dur s
 
 -- | The varying data of this test
 --
@@ -107,6 +119,10 @@ data TestSetup = TestSetup
   , setupD                 :: Shelley.DecentralizationParam
   , setupHardFork          :: Bool
     -- ^ whether the proposal should trigger a hard fork or not
+  , setupInitialNonce      :: SL.Nonce
+    -- ^ the initial Shelley 'SL.ticknStateEpochNonce'
+    --
+    -- We vary it to ensure we explore different leader schedules.
   , setupK                 :: SecurityParam
   , setupPartition         :: Partition
   , setupSlotLengthByron   :: SlotLength
@@ -119,9 +135,18 @@ data TestSetup = TestSetup
 instance Arbitrary TestSetup where
   arbitrary = do
     setupD <- arbitrary
-                -- TODO Issue 2388 prevents `d=0` in this test.
+                -- The decentralization parameter cannot be 0 in the first
+                -- Shelley epoch, since stake pools can only be created and
+                -- delegated to via Shelley transactions.
                 `suchThat` ((/= 0) . Shelley.decentralizationParamToRational)
-    setupK <- SecurityParam <$> choose (2, 6)
+    setupK <- SecurityParam <$> choose (4, 6)
+                -- If k < 4, common prefix violations become too likely in
+                -- Praos mode.
+
+    setupInitialNonce <- frequency
+      [ (1, pure SL.NeutralNonce)
+      , (9, SL.mkNonceFromNumber <$> arbitrary)
+      ]
 
     setupSlotLengthByron   <- arbitrary
     setupSlotLengthShelley <- arbitrary
@@ -130,7 +155,7 @@ instance Arbitrary TestSetup where
     let TestConfig{numCoreNodes, numSlots} = setupTestConfig
 
     setupByronLowerBound <- arbitrary
-    setupHardFork        <- frequency [(9, pure True), (1, pure False)]
+    setupHardFork        <- frequency [(49, pure True), (1, pure False)]
     setupPartition       <- genPartition numCoreNodes numSlots setupK
 
     setupVersion         <- genVersionFiltered
@@ -141,6 +166,7 @@ instance Arbitrary TestSetup where
       { setupByronLowerBound
       , setupD
       , setupHardFork
+      , setupInitialNonce
       , setupK
       , setupPartition
       , setupSlotLengthByron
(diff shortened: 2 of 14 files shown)


ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
```diff
@@ -1,15 +1,308 @@
...
+        , SL._poolPubKey = Shelley.mkKeyHash poolSK
+        , SL._poolRAcnt  =
+            SL.RewardAcnt Shelley.networkId $ Shelley.mkCredential poolSK
+        , SL._poolRelays = StrictSeq.empty
+        , SL._poolVrf    = Shelley.mkKeyHashVrf vrfSK
+        }
+
+-----
+
+ejectShelleyNS :: NS f (CardanoEras c) -> Maybe (f (ShelleyBlock c))
```
nfrisby, line 286: @mrBliss Are there pre-existing functions that replace/simplify these `ejectShelleyNS`, `ejectShelleyTickedLedgerState`, and `getUTxOShelley` functions I'm defining here?

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/MockCrypto.hs
```diff
@@ -36,3 +40,29 @@ instance HashAlgorithm h => Crypto (TPraosMockCrypto h) where
 instance HashAlgorithm h => TPraosCrypto (TPraosMockCrypto h)
 
 type Block h = ShelleyBlock (TPraosMockCrypto h)
+
+-- | Like 'TPraosMockCrypto' except mocks fewer components in order to be
```
nfrisby, line 44: Should this crypto bundle be relocated to `ouroboros-consensus-cardano/src/Test/*`? (That file hierarchy does not yet exist.)

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/MockCrypto.hs
```diff
@@ -36,3 +40,29 @@ instance HashAlgorithm h => Crypto (TPraosMockCrypto h) where
...
+-- generators from being used. Currently, this is not an obstacle for example
+-- in the Cardano ThreadNet tests.
+--
+-- NOTE: The "Ouroboros.Consensus.Cardano.CanHardFork" translation currently
+-- assumes that @ADDRHASH@ has the same bit size as Byron address hashes (ie
+-- 224); that's why we use 'Blake25_224' here.
+--
+-- NOTE: The @shelley-spec-ledger@ package currently requires that @'DSIGN' ~
+-- 'Ed25519DSIGN'.
+data TPraosMockCryptoCompatByron h
```
mrBliss, line 57: Some thoughts about parameterising mock crypto by `h` (also `TPraosMockCrypto`): in #2503, I rewrite the examples used by the golden tests to use `TPraosStandardCrypto`, so no more `TPraosMockCrypto` there. I believe the serialisation roundtrip tests still use `ShortHash` for performance reasons and for generator reuse/ease (IIRC). The Shelley ThreadNet tests also use `ShortHash` for performance. Any other things I'm missing? All other things uses Mock* with `Blake2b_256`, I suppose. So actually there's no need to parameterise `TPraosMockCryptoCompatByron` over `h`, correct? Nevertheless, I think it's fine to leave it as-is.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/MockCrypto.hs
```diff
@@ -36,3 +40,29 @@ instance HashAlgorithm h => Crypto (TPraosMockCrypto h) where
...
+-- | Like 'TPraosMockCrypto' except mocks fewer components in order to be
+-- compatible with bootstrapping from Byron
+--
+-- Note that mocking less components does prevent the @cardano-ledger-specs@
+-- generators from being used. Currently, this is not an obstacle for example
+-- in the Cardano ThreadNet tests.
+--
+-- NOTE: The "Ouroboros.Consensus.Cardano.CanHardFork" translation currently
+-- assumes that @ADDRHASH@ has the same bit size as Byron address hashes (ie
+-- 224); that's why we use 'Blake25_224' here.
```
mrBliss, line 53:
```suggestion
-- 224); that's why we use 'Blake2b_224' here.
```

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
```diff
@@ -107,6 +119,10 @@ data TestSetup = TestSetup
   , setupD                 :: Shelley.DecentralizationParam
   , setupHardFork          :: Bool
     -- ^ whether the proposal should trigger a hard fork or not
+  , setupInitialNonce      :: SL.Nonce
+    -- ^ the initial Shelley 'SL.ticknStateEpochNonce'
+    --
+    -- We vary it too ensure we explore different leader schedules.
```
mrBliss, line 125:
```suggestion
    -- We vary it to ensure we explore different leader schedules.
```

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
```diff
@@ -530,20 +577,57 @@ prop_simple_cardano_convergence TestSetup
...
+        shift x = Set.mapMonotonic (Util.addSlots (unEpochSize x))
+
+        -- The relative offset of the overlay slots in each Shelley epoch.
+        overlayOffsets :: Set SlotNo
+        overlayOffsets =
+            Map.keysSet $
+            runIdentity $ flip runReaderT shelleyGlobals $
+            SL.overlaySchedule
+              @Crypto
+              (EpochNo 0)   -- Ignoring this input; see uses of @shift@ above
```
mrBliss, line 618: I don't understand, if you're shifting all slots by the epoch size, why not use `EpochNo 1` here?

Ah, you're shifting by `epochSizeByron` first. A cleaner solution would be to pass the correct `EpochInfo` to `shelleyGlobals`. You would have to construct a correct `Summary` and pass that to [`snapshotEpochInfo`](https://github.com/input-output-hk/ouroboros-network/blob/3a83c64ce76fb995b9557d3eb55d3c7e1c2fc251/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/EpochInfo.hs#L48).

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
```diff
@@ -1,15 +1,308 @@
...
+data CardanoTxGenExtra c = CardanoTxGenExtra
+  { ctgeByronGenesisKeys :: GeneratedSecrets
+  , ctgeNetworkMagic     :: Byron.NetworkMagic
+  , ctgeShelleyCoreNodes :: [Shelley.CoreNode c]
+  }
+
+instance ( TPraosCrypto c
+           -- These equalities allow use of Byron bootstrap addresses in the
+           -- Shelley era, since @shelley-spec-ledger@ requires Ed25519 and the
+           -- current Byron-to-Shelley translation requires _224.
```
mrBliss, line 77:
```suggestion
           -- current Byron-to-Shelley translation requires Blake2b_224.
```
Or was the omission intentional?

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
```diff
@@ -1,15 +1,308 @@
...
+-- It returns 'Nothing' if the core node does not have any utxo in its
+-- 'byronAddr' (eg if this transaction has already been applied).
+migrateUTxO :: forall c.
+               (TPraosCrypto c, DSIGN c ~ Ed25519DSIGN)
+            => MigrationInfo c
+            -> SlotNo
+            -> LedgerConfig (CardanoBlock c)
+            -> LedgerState (CardanoBlock c)
+            -> Maybe (GenTx (CardanoBlock c))
+migrateUTxO migrationInfo curSlot lcfg lst
```
mrBliss, line 153: I must say that I'm not that familiar with all this ledger business. Best to let @nc6 review this part.

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
mrBliss: There's is the [`LedgerStateShelley`](https://github.com/input-output-hk/ouroboros-network/blob/3a83c64ce76fb995b9557d3eb55d3c7e1c2fc251/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs#L494) pattern synonym, but that's not for a *ticked* ledger state, and I don't want to add such a synonym as I don't want other packages to mess with ticked ledger states.

No, I don't think there are any pre-existing functions you can use to replace/simplify these three.

Why "eject"? Is that an existing naming convention? It makes me think of "reject", which would do the opposite (so in some sense, "eject" makes sense). I would go for something like "select" instead.

nfrisby: We agreed to follow https://hackage.haskell.org/package/generics-sop-0.5.1.0/docs/Generics-SOP.html#t:Ejection

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
nfrisby: It was, but no need to be coy. Clarified, thanks!

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
nfrisby: Fixed, thanks.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/MockCrypto.hs
nfrisby: Fixed, thanks.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/MockCrypto.hs
nfrisby: Yeah, `git grep -e 'TPraosMockCrypto' -- '*hs' | grep -vE -e "TPraosMockCrypto[')]" -e 'TPraosMockCrypto (h|ShortHash)'` confirms that your list is exhaustive for `TPraosMockCrypto`.

And this PR only uses the new `TPraosMockCryptoCompatByron` at type `Blake2b_256`.

But I agree the parameterization is fine as-is. Some constraint contexts would be simpler without it, but it's not burdensome to currently justify the extra diff now, in my opinion. In fact, some functions and certainly some types are clearer/safer when paramterized over `h`, in my opinion.

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
nfrisby: We agreed that the `Summary` is not easy to create here, so to avoid inflating the diff eta, we compromised with a smaller refactor:

```
        -- The overlay slots in the ith Shelley epoch.
        overlayOffsets :: Word64 -> Set SlotNo
        overlayOffsets i =
            -- Shift to account for the initial Byron era.
            Set.mapMonotonic (Util.addSlots numByronSlots) $

            Map.keysSet $
            runIdentity $ flip runReaderT shelleyGlobals $
            SL.overlaySchedule
              @Crypto
              (EpochNo i)   -- NB 0 <-> the first /Shelley/ epoch
              genesisKeyHashes
              (sgProtocolParams genesisShelley)   -- notably contains setupD
```

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
nfrisby: Yeah, that's what I had in mind when I added @nc6 as a reviewer: this part and then just a general FYI.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/MockCrypto.hs
nfrisby: Done, thanks!

nfrisby:
I force pushed updated commits: they now extend `origin/master` and reflect Thomas' comments.

I think it's ready for merge, as long as @nc6 doesn't see any problems/simplification opportunities in the transaction created in the new lines of `ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs`.

nfrisby:
I rebased onto `origin/master` -- no conflicts, everything built locally 👍

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
```diff
@@ -1,15 +1,308 @@
...
+            pick (SL.TxOut addr _) =
+                addr == SL.AddrBootstrap (SL.BootstrapAddress byronAddr)
+
+        -- Total held by 'byronAddr'
+        pickedCoin :: SL.Coin
+        pickedCoin =
+            sum $ fmap (\(SL.TxOut _ coin) -> coin) picked
+
+        fee        :: SL.Coin
+        fee        = 0
```
nc6, line 169: I guess you just have `minFee` and `poolDeposit` set to 0?

ouroboros-consensus-cardano/test/Test/Consensus/Cardano/MockCrypto.hs
```diff
@@ -0,0 +1,45 @@
...
+-- NOTE: The @shelley-spec-ledger@ package currently requires that @'DSIGN' ~
+-- 'Ed25519DSIGN'.
+data TPraosMockCryptoCompatByron h
+
+instance HashAlgorithm h => Crypto (TPraosMockCryptoCompatByron h) where
+  type ADDRHASH (TPraosMockCryptoCompatByron h) = Blake2b_224
+  type DSIGN    (TPraosMockCryptoCompatByron h) = Ed25519DSIGN
+  type HASH     (TPraosMockCryptoCompatByron h) = h
+  type KES      (TPraosMockCryptoCompatByron h) = MockKES 10
+  type VRF      (TPraosMockCryptoCompatByron h) = FakeVRF
```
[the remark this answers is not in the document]
nfrisby: Good to know, thanks. My general preference is to mock as much as possible. Is there a better alternative? Or maybe the real `VRF` is relatively simple and not worth mocking? I'm unfamiliar.

ouroboros-consensus-cardano/test/Test/Consensus/Cardano/MockCrypto.hs
```diff
@@ -0,0 +1,45 @@
...
+-- 'Test.Consensus.Shelley.MockCrypto.TPraosMockCrypto'. This prevents the
+-- @cardano-ledger-specs@ generators from being re-used. Currently, this is not
+-- an obstacle for example in the Cardano ThreadNet tests.
+--
+-- NOTE: The "Ouroboros.Consensus.Cardano.CanHardFork" translation currently
+-- assumes that @ADDRHASH@ has the same bit size as Byron address hashes (ie
+-- 224); that's why we use 'Blake2b_224' here.
+--
+-- NOTE: The @shelley-spec-ledger@ package currently requires that @'DSIGN' ~
+-- 'Ed25519DSIGN'.
```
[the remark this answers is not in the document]
nfrisby: Ah, my edits lost a key phrase: "in order to support Byron witnesses". I'll add it back.

For example: https://github.com/input-output-hk/cardano-ledger-specs/blob/dae9940d110d8671a154e5f9946195c315837905/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/Address/Bootstrap.hs#L198-L202

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
```diff
@@ -528,20 +577,56 @@ prop_simple_cardano_convergence TestSetup
...
+        takeWhile (isJust . Set.lookupLT (SlotNo t)) $
+        map overlayOffsets [0..]
+      where
+        NumSlots t = numSlots
+
+        -- The overlay slots in the ith Shelley epoch.
+        overlayOffsets :: Word64 -> Set SlotNo
+        overlayOffsets i =
+            -- Shift to account for the initial Byron era.
+            Set.mapMonotonic (Util.addSlots numByronSlots) $
```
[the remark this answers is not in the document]
nfrisby: Yep; it seemed quite a bit more direct this way instead of building a sophisticated `epochInfo` (by building a HardFork `Summary` eg).

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
```diff
@@ -1,15 +1,308 @@
...
+  , ctgeNetworkMagic     :: Byron.NetworkMagic
+  , ctgeShelleyCoreNodes :: [Shelley.CoreNode c]
+  }
+
+instance ( TPraosCrypto c
+           -- These equalities allow use of Byron bootstrap addresses in the
+           -- Shelley era, since @shelley-spec-ledger@ requires Ed25519 and the
+           -- current Byron-to-Shelley translation requires a 224-bit hash.
+         , ADDRHASH c ~ Blake2b_224
+         , DSIGN    c ~ Ed25519DSIGN
```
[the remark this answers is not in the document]
nfrisby: (See other comment thread.)

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
nfrisby: Yeah, the `Cardano` ThreadNet uses the `ouroboros-shelley-test` genesis config

https://github.com/input-output-hk/ouroboros-network/blob/83baf7eb58907137d50605fc2c85fd4c51ae68e0/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs#L446-L447

which boils down to `SL.emptyPParams`.

https://github.com/input-output-hk/ouroboros-network/blob/62e57b9673e0140c87c6edd6ca8fb7b813c9f8d9/ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs#L293

ouroboros-consensus-cardano/test/Test/ThreadNet/TxGen/Cardano.hs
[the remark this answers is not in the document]
nfrisby: Ah, I just noticed that the comment above explains it in natural language.

nfrisby:
@nc6 Thanks for your eyes! I expanded comments a bit in response to (most of?) your questions/observations.

nfrisby:
Doh! I currently have terrrrible wifi, so I'm unable to fetch the new deps to update my nix-shell to build locally. So I've pushed up a one line change (adding `import qualified Shelley.Spec.Ledger.OverlaySchedule as SL`). I'm hoping that suffices.

If not, I'll have to wait until the next chance I get an hour or so to kill while keeping the sleeping puppy company.

ouroboros-consensus-cardano/test/Test/Consensus/Cardano/MockCrypto.hs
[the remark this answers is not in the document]
mrBliss: One thing to keep in mind is that real VRF will make the tests slower.

mrBliss:
> Doh! I currently have terrrrible wifi, so I'm unable to fetch the new deps to update my nix-shell to build locally. So I've pushed up a one line change (adding `import qualified Shelley.Spec.Ledger.OverlaySchedule as SL`). I'm hoping that suffices.
> 
> If not, I'll have to wait until the next chance I get an hour or so to kill while keeping the sleeping puppy company.

You don't have to feel obligated to finish this PR during your vacation, on the contrary, ignore it completely until you're back :slightly_smiling_face:

nfrisby:
bors retry

PR #2503, opened 2020-08-11
Update dependency on cardano-ledger-specs and rewrite examples

mrBliss:
`cardano-ledger-specs` no longer exports the examples that we used for our
golden tests. Take this as an opportunity to rewrite the examples ourselves
using real crypto instead of the mock crypto used by the examples in
`cardano-ledger-specs`. This also means our golden tests won't break anymore
when examples are changed in `cardano-ledger-specs` without actually making
breaking changes to the serialisation format.


mrBliss:
bors merge

PR #2507, opened 2020-08-12
mux-demo: remove dependency on io-sim

mrBliss:
The `mux-demo` doesn't depend on the `io-sim` package.

I believe this is the only reason `cardano-node` needs to list `io-sim` as a
`source-repository-package`. In `cardano-node`, the tests are disabled for all
dependencies, but we can't disable building executables in dependencies, e.g.,
`mux-demo`.

### DIFF
--- a/network-mux/network-mux.cabal
+++ b/network-mux/network-mux.cabal
@@ -136,7 +136,6 @@ executable mux-demo
                        directory,
                        network-mux,
                        io-sim-classes,
-                       io-sim,
                        contra-tracer,
                        stm,
 


mrBliss:
bors merge

mrBliss:
Uggh, I was wrong. `ouroboros-network` has an internal `ouroboros-protocol-tests` library which also depends on `io-sim`, but Cabal seems to build all internal libraries, even when they're only needed for tests. This means `cardano-node` will still need to list `io-sim` as a `source-repository-package`.

PR #2512, opened 2020-08-12
Shelley: avoid expensive equality check of the LedgerView

mrBliss:
We check whether the `LedgerView` changed after (re)applying each block, to see
whether we should store a snapshot of the old one. The `LedgerView` includes the
overlay schedule (a map with 432,000 entries!) and the pool distribution (a map
with an entry for each stakepool). Comparing these two data structures for
equality is very costly. Since the overlay schedule and the pool distribution
only change when transitioning to a new epoch, we can use the epoch number as a
proxy for the equality check of those two data structures.

The other two data structures, i.e., `lvProtParams` and `lvGenDelegs`, are small
and can be checked time.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
@@ -260,16 +260,14 @@ applyHelper f cfg blk
     newShelleyState <- f globals oldShelleyState (shelleyBlockRaw blk)
 
     let history'
-          -- TODO how expensive is this check?
-          | SL.currentLedgerView oldShelleyState ==
-            SL.currentLedgerView newShelleyState
-          = history
-          | otherwise
+          | ledgerViewChanged oldShelleyState newShelleyState
           = History.snapOld
               (SL.securityParameter globals)
               (blockSlot blk)
               (SL.currentLedgerView oldShelleyState)
               history
+          | otherwise
+          = history
 
     return ShelleyLedgerState {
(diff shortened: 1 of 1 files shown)


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -279,6 +277,20 @@ applyHelper f cfg blk
...
+    ledgerViewChanged :: SL.ShelleyState c -> SL.ShelleyState c -> Bool
+    ledgerViewChanged old new
+      | SL.nesEL old /= SL.nesEL new
+        -- When the epoch changed, we know the overlay schedule and the pool
+        -- distribution will have changed. This allows us to avoid checking
+        -- the equality of these huge data structures.
+      = True
+      | otherwise
+      = SL.lvProtParams oldLedgerView /= SL.lvProtParams newLedgerView ||
+        SL.lvGenDelegs  oldLedgerView /= SL.lvGenDelegs  newLedgerView
```
mrBliss, line 289: @nc6 Do these two values ever change within an epoch or also only at the epoch boundary?

If the latter, then we can improve/simplify things even further.

nc6: Protocol parameters only change at an epoch boundary. Genesis delegates may change within an epoch.

mrBliss: Great, updated. Thanks!

mrBliss:
bors merge

PR #2515, opened 2020-08-14
Unify the VolatileDB and VolDB 

mrBliss:
Make the VolatileDB aware of the `blk`, simplifying a lot of things and making
the VolDB wrapper redundant.

Now the VolatileDB can keep the nested context in memory as a
`SomeBlock (NestedCtxt Header) blk` instead of a `ShortByteString`, which has to
be decoded each time.

Reorganise the VolatileDB along the way.


mrBliss:
Draft because when I do the same for the ImmutableDB/ImmDB, I might reconsider some decisions I made. Not urgent anyway.

mrBliss:
Closed in favour of #2534.

Issue #2520, opened 2020-08-18
Introduce header revalidation


mrBliss:
As mentioned in https://github.com/input-output-hk/cardano-ledger-specs/pull/1785#issuecomment-675035681, the majority of block replay time is spent in C functions doing crypto checks, see the two top of the profiling report:
```
COST CENTRE   SRC                                               %time  %alloc
verify        src/Cardano/Crypto/VRF/Praos.hs:(388,1)-(397,33)   46.9    0.1
verify        Crypto/PubKey/Ed25519.hs:(107,1)-(114,30)           6.3    0.0
```

Both are part of `validateHeader`, which accounts for 63.2% of the time and 11.6% of the allocations.

Note that this is when we're *reapplying blocks*, so we know these blocks *and thus their headers* are already valid. While we already distinguish between block body [application](https://github.com/input-output-hk/ouroboros-network/blob/cfb465726c9d023ad3ab2a3030d847df27787c3d/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Abstract.hs#L63) and [reapplication](https://github.com/input-output-hk/ouroboros-network/blob/cfb465726c9d023ad3ab2a3030d847df27787c3d/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Abstract.hs#L76), we don't do the same thing for block [header validation](https://github.com/input-output-hk/ouroboros-network/blob/cfb465726c9d023ad3ab2a3030d847df27787c3d/ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs#L468).

Header validation consists of three parts:
https://github.com/input-output-hk/ouroboros-network/blob/cfb465726c9d023ad3ab2a3030d847df27787c3d/ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs#L474-L492

1. `validateEnvelope`, which is cheap, but could be skipped during header revalidation. Maybe leave it in as a basic sanity check?
[...]

PR #2521, opened 2020-08-18
Introduce header revalidation to speed up Shelley block revalidation

mrBliss:
Implements the plan described in #2520.

My block revalidation test went from 1m52s to 45s, which nicely matches the
expected speed-up: header validation took ~60% of the time.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
@@ -343,6 +343,27 @@ instance PBftCrypto c => ConsensusProtocol (PBft c) where
     where
       params = pbftWindowParams cfg
 
+  reupdateChainDepState cfg@PBftConfig{..}
+                        toValidate
+                        slot
+                        (TickedPBftState (TickedPBftLedgerView dms) state) =
+      case toValidate of
+        PBftValidateBoundary hash ->
+          appendEBB cfg params slot hash state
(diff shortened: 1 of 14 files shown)


mrBliss:
Draft because https://github.com/input-output-hk/cardano-ledger-specs/pull/1789 hasn't been merged yet.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -343,6 +343,27 @@ instance PBftCrypto c => ConsensusProtocol (PBft c) where
...
+                        toValidate
+                        slot
+                        (TickedPBftState (TickedPBftLedgerView dms) state) =
+      case toValidate of
+        PBftValidateBoundary hash ->
+          appendEBB cfg params slot hash state
+        PBftValidateRegular PBftFields{pbftIssuer} _ _ ->
+          case Bimap.lookupR (hashVerKey pbftIssuer) dms of
+            Nothing ->
+              error $ show $ PBftNotGenesisDelegate
```
mrBliss, line 356: I'm not too happy about these `error . show` combos I have here and in other places, but we basically had the same `error` in `cantBeError` (in `Extended.hs`).

I do like the approach taken in `cardano-ledger-specs`, i.e., skipping specific checks (we would need to introduce some kind of predicate for this) can be done by flipping a switch.

mrBliss:
bors merge

mrBliss:
> LGTM! Merged in c-l-s as well

I'll update `cardano-node` next.

PR #2525, opened 2020-08-20
ChainSyncClient: gracefully terminate if the chain is uninteresting

### LINKED ISSUE #2499: Avoid throwing exceptions in mini protocols
From @coot :

The network-mux library, for some time, is capable of returning values from each mini-protocol. We would like you guys to move from throwing exceptions (which was necessary in the previous version of mux) to gracefully finish a protocol and return.  AFAIK that's only important for the chain-sync client when it decides that the remote chain is too old, in which case now it should not throw, but gracefully return (i.e. reach the termination state of the protocol).  A clean exit allows us to restart mini-protocols on the same bearer (what the peer-to-peer governor will do when it decides to promote the peer back as a hot one).

I think you don't need to worry about synchronisation between all the protocols, e.g. if chain-sync exists, the governor should ask all the other protocols to terminate (I need to review if it actually does it though), and it's probably the best place to handle that.

This should work with the current mux compatibility layer that we are using. The small change that we'd need to do is to handle the return value and run ErrorPolicies on it to actually suspend the peer. This means this could go to master. Alternatively, it could be just merged with coot/connection-manager, and go to master with all the other changes.

When throwing is ok? When a protocol is violated. One thing to remember is when one throws from the context of a protocol it will always bring the bearer down, and since we will run it in bidirectional mode it will  shutdown both responder and initiator whether the exception was thrown from initiator or responder side.

mrBliss:
Fixes #2499 for the ChainSyncClient.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE ExistentialQuantification  #-}
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE MultiWayIf                 #-}
 {-# LANGUAGE NamedFieldPuns             #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
@@ -24,6 +25,7 @@ module Ouroboros.Consensus.MiniProtocol.ChainSync.Client (
     Consensus
   , chainSyncClient
   , bracketChainSyncClient
+  , ChainSyncClientResult (..)
   , ChainSyncClientException (..)
   , ChainDbView (..)
   , defaultChainDbView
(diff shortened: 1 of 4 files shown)


mrBliss:
@coot The CI failures are due to the base branch of this PR.

dcoutts (approved):
LGTM

coot:
> @coot The CI failures are due to the base branch of this PR.

That's ok, it's still work in progress.

Do the consensus clients check what in this branch is called `ControlMessage`  (and what in master is called `ScheduledStop`)? It seems not: [link](https://github.com/input-output-hk/ouroboros-network/blob/coot/connection-manager/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs#L585).

 The peer-to-peer governor relies on clients to respond to it. Could you add it to this PR?

```
data ControlMessage =
    -- | Continue operation.
      Continue

    -- | Hold on, e.g. do not sent messages until resumed.  This is not used for
    -- any hot protocol.
    --
    | Quiesce

    -- | The client is expected to terminate as soon as possible.
    --
    | Terminate
  deriving (Eq, Show)
```

The consensus protocols will only ever see `Continue` or `Terminate`.

coot:
I merged and resolved conflicts with my local changes, hopefully `coot/connection-manager` builds now on CI.

coot:
> Do the consensus clients check what in this branch is called ControlMessage (and what in master is called ScheduledStop)? It seems not: link.

I put this in a separate issue: #2528

PR #2533, opened 2020-08-24
Avoid keeping the AVVM balances in memory

mrBliss:
See the descriptions of the individual commits for more details.

### DIFF
--- a/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node.hs
+++ b/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node.hs
@@ -159,21 +159,26 @@ protocolInfoByron genesisConfig mSigThresh pVer sVer mLeader =
     ProtocolInfo {
         pInfoConfig = TopLevelConfig {
             topLevelConfigProtocol = PBftConfig {
-                pbftParams = byronPBftParams genesisConfig mSigThresh
+                pbftParams = byronPBftParams compactedGenesisConfig mSigThresh
               }
           , topLevelConfigBlock = FullBlockConfig {
-                blockConfigLedger = genesisConfig
-              , blockConfigBlock  = mkByronConfig genesisConfig pVer sVer
-              , blockConfigCodec  = mkByronCodecConfig genesisConfig
+                blockConfigLedger = compactedGenesisConfig
+              , blockConfigBlock  = mkByronConfig compactedGenesisConfig pVer sVer
+              , blockConfigCodec  = mkByronCodecConfig compactedGenesisConfig
               }
           }
       , pInfoInitLedger = ExtLedgerState {
+            -- Important: don't pass the compacted genesis config to
+            -- 'initByronLedgerState', it needs the full one, including the AVVM
+            -- balances.
             ledgerState = initByronLedgerState genesisConfig Nothing
           , headerState = genesisHeaderState S.empty
           }
       , pInfoBlockForging =
           return . byronBlockForging <$> mLeader
       }
+  where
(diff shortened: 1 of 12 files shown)


ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node.hs
```diff
@@ -159,21 +159,26 @@ protocolInfoByron genesisConfig mSigThresh pVer sVer mLeader =
     ProtocolInfo {
         pInfoConfig = TopLevelConfig {
             topLevelConfigProtocol = PBftConfig {
-                pbftParams = byronPBftParams genesisConfig mSigThresh
+                pbftParams = byronPBftParams compactedGenesisConfig mSigThresh
               }
           , topLevelConfigBlock = FullBlockConfig {
                 blockConfigLedger = genesisConfig
```
kderme, line 165:
```suggestion
                blockConfigLedger = compactedGenesisConfig
```

This can also be `compactedGenesisConfig` right? Or else we still keep a reference to the balances through the `PartialLedgerConfig`.

kderme (changes requested):

ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node.hs
mrBliss: Indeed! Good catch, thanks

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -154,34 +146,34 @@ type LedgerDB blk = LedgerDB.LedgerDB (ExtLedgerState blk) (RealPoint blk)
...
-data LgrDbArgs m blk = forall h. Eq h => LgrDbArgs {
+data LgrDbArgs m blk = LgrDbArgs {
       lgrTopLevelConfig :: TopLevelConfig blk
-    , lgrHasFS          :: HasFS m h
+    , lgrHasFS          :: SomeHasFS m
     , lgrParams         :: LedgerDbParams
     , lgrDiskPolicy     :: DiskPolicy
     , lgrGenesis        :: m (ExtLedgerState blk)
     , lgrTracer         :: Tracer m (TraceEvent (RealPoint blk))
     , lgrTraceLedger    :: Tracer m (LedgerDB blk)
```
kderme, line 156: Not related with the pr, but do we ever use this tracer?

kderme (approved):

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
mrBliss: Good question. At first I thought that we indeed no longer use it. However, when I looked at the PR that introduced it (#1156), I realised that we're not using `lgrTraceLedger` but we *are* using `cdbTraceLedger`. I suppose Edsko introduced it to be able to write [`toChainDbArgs`](https://github.com/input-output-hk/ouroboros-network/blob/1c8924b856baf809f193798ff686cdb2c0ed25d2/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Args.hs#L214). So we need we can't simply remove `lgrTraceLedger` without extra changes.

At some point in the future, I'd like to clean up the argument handling of the ChainDB (and get rid of those nasty `error`s :roll_eyes:). We can get rid of `lgrTraceLedger` then. A prerequisite for this is #2534.

mrBliss:
bors merge

PR #2547, opened 2020-08-26
Update source dependencies to GHC 8.10 compatible versions

mrBliss:


cabal.project
```diff
@@ -123,136 +123,151 @@ source-repository-package
...
   tag: 2547ad1e80aeabca2899951601079408becbc92c
   --sha256: 1p2kg2w02q5w1cvqzhfhqmxviy4xrzada3mmb096j2n6hfr20kri
+
+-- The r0 revision of this quickcheck-state-machine-0.7 on Hackage adds a lower
+-- bound on the text package: >=1.2.4.0. However, 1.2.4.0 doesn't support GHC
+-- 8.10, 1.2.4.1 will, but that hasn't been released yet. In the meantime,
+-- depend on the git version which is identical, but without the lower bound on
+-- text. See
+-- https://github.com/advancedtelematic/quickcheck-state-machine/issues/371
+source-repository-package
```
nc6, line 266: Rather than vendoring q-s-m, we could instead just use `allow-older` to override the constraint on `text`:

```
allow-older: quickcheck-state-machine-0.7:text
```

nc6 (approved):
Looks good. Made one suggestion.

Also I can't seem to build this with GHC8.10 - I get issues with Win32-network. Presumably you don't see this?

mrBliss:
> Also I can't seem to build this with GHC8.10 - I get issues with Win32-network. Presumably you don't see this?

That's "intentional". This PR just makes the source dependencies compatible with 8.10, but not yet the network layer (including Win32-network). That will be done by #2539.

cabal.project
mrBliss: Good idea!

mrBliss:
bors merge

PR #2546, opened 2020-08-26
Remove LedgerView history

mrBliss:
Fixes #1935.

See the individual commits for detailed descriptions.

PR #2541, opened 2020-08-26
Rename RunOrStop as ControlMessage

coot:
This patch is pulled out from `coot/connection-manager` branch to avoid future
merge conflicts.

PR #2548, opened 2020-08-26
Port #2525 and #2531 to master

mrBliss:
#2525 and #2531 were merged in the https://github.com/input-output-hk/ouroboros-network/tree/coot/connection-manager. However, that branch won't be merged for another few weeks. In the meantime, the merge conflicts with the changes made in those two PRs will accumulate. For example, I know that #2546 will cause a merge conflict.

Now that @coot has kindly merged #2541 (which those PRs depended on) in master, we can port #2525 and #2532 to master so we can avoid the upcoming merge conflicts. After this PR is merged, the `coot/connection-manager` branch should be rebased onto master.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -101,7 +101,7 @@ data Handlers m peer blk = Handlers {
       hChainSyncClient
         :: NodeToNodeVersion
         -> StrictTVar m (AnchoredFragment (Header blk))
-        -> ChainSyncClientPipelined (Header blk) (Tip blk) m Void
+        -> ChainSyncClientPipelined (Header blk) (Tip blk) m ChainSyncClientResult
         -- TODO: we should consider either bundling these context parameters
         -- into a record, or extending the protocol handler representation
         -- to support bracket-style initialisation so that we could have the
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -101,6 +101,7 @@ import           Ouroboros.Consensus.Storage.ChainDB.Serialisation
 data Handlers m peer blk = Handlers {
       hChainSyncClient
         :: NodeToNodeVersion
+        -> ControlMessageSTM m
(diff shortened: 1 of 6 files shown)


mrBliss:
Ah, #2525 makes the ThreadNet tests loop/hang. I did not realise that before, because the branch in which I merged #2525 didn't compile cleanly on CI (so CI never ran these tests) and I never ran them myself :flushed:.

mrBliss:
> Ah, #2525 makes the ThreadNet tests loop/hang. I did not realise that before, because the branch in which I merged #2525 didn't compile cleanly on CI (so CI never ran these tests) and I never ran them myself flushed.

I solved the issue with help from @nfrisby, see #2553.

mrBliss:
@coot @dcoutts @nfrisby Can somebody approve/review this PR? It ports two already reviewed PRs to master. However, because the branch in which they were merged had build errors in CI, we never detected a bug in the original PR, which @nfrisby and I fixed.

@coot This means your `coot/connection-manager` branch still contains that bug, so best to merge this PR soon and rebase your branch on top of master.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -352,36 +354,44 @@ showTracers tr = Tracers {
...
+  -> m (a, Maybe bytes)
+
 -- | Applications for the node-to-node protocols
 --
 -- See 'Network.Mux.Types.MuxApplication'
 data Apps m peer bCS bBF bTX bKA a = Apps {
       -- | Start a chain sync client that communicates with the given upstream
       -- node.
-      aChainSyncClient    :: NodeToNodeVersion -> peer -> Channel m bCS -> m (a, Maybe bCS)
+      aChainSyncClient    :: App m peer bCS a
```
nfrisby, line 371: This means we'll not see the actual return value of the client, even though all the upstream code changes now make it available. (The `a` is shared by the other protocols, and none of those return anything. So it's `()`/`Void`/`void`/etc or at the very least a bit misleading.) The CS return value is recorded in a tracer invoked within the client, so it'll at least be logged, which seems like the primary use-case, though.

Just an observation.

mrBliss: Yeah, that's indeed not so nice. That `Apps` business is actually written by the network team. Unfortunately it lives in the consensus layer :grimacing:.

mrBliss:
bors merge

PR #2550, opened 2020-08-26
Fix current failures on nightly Cardano ThreadNet tests

### LINKED ISSUE #2549: Cardano ThreadNet failing with Common Prefix violations
The nightlies started failing on Monday https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/206#849139a3-7913-4372-9425-80bd9ff04150

It's the Cardano test. The assert failure present in the logs corresponds to a Common Prefix violation. This is too consistent to be "unlucky".

I think PR #2502 (merged Monday) is far and away the most likely culprit. It extends the Cardano ThreadNet tests so that the reach a Shelley epoch era in which stakepools are actually leading. In retrospect, that PR should have lowered the active slot coefficient from `f=0.5`. For low enough `d`, such a high `f` leads to wedges and hence CP violations.

This task is to alter this test (eg lower `f`, raise the minimum `k`, etc) to make CP violations extremely unlikely. This is not entirely obvious how to do without making the tests extremely expensive (eg just set `k` very high, which linearly increases the number of slots necessary to reach an epoch with active stakepools).

Also, these failures should be proper `property` failures, not just an `assert`.

nfrisby:
Fixes #2549 (CP violations in Cardano net). I can't predict how soon we'll see a CP violation again, but I'd suspect this change will afford us weeks -- in my local Mock Praos testing, I've never seen `k=8 f=0.2` fail during 300 slots, after >=10,000 of test runs.

So I think these numbers are worth a merge now. For example, they passed one nightly:

https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/208#872eb9dc-c494-4328-9dba-c871157739b6

We can reopen if we see similar failure relatively soon. Also, I hope to revisit once my more general related line of work converges on some extrapolatable probabilities.

### DIFF
--- a/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
+++ b/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
@@ -496,6 +496,7 @@ prop_simple_cardano_convergence TestSetup
         Shelley.mkGenesisConfig
           (SL.ProtVer shelleyMajorVersion 0)
           setupK
+          activeSlotCoeff
           setupD
           setupSlotLengthShelley
           (Shelley.mkKesConfig (Proxy @(KES Crypto)) numSlots)
@@ -769,6 +770,9 @@ mkProtocolCardanoAndHardForkTxs
   Constants
 -------------------------------------------------------------------------------}
 
+activeSlotCoeff :: Rational
+activeSlotCoeff = 0.5
+
 maxMajorPVShelley :: Natural
 maxMajorPVShelley = 100   -- arbitrary
 
@@ -892,7 +896,7 @@ byronEpochSize (SecurityParam k) =
     unEpochSlots $ kEpochSlots $ CC.Common.BlockCount k
 
 shelleyEpochSize :: SecurityParam -> Word64
-shelleyEpochSize = unEpochSize . Shelley.mkEpochSize
+shelleyEpochSize k = unEpochSize $ Shelley.mkEpochSize k activeSlotCoeff
 
 isShelley :: CardanoBlock c -> Bool
 isShelley = \case
--- a/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
+++ b/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
@@ -88,6 +88,7 @@ import qualified Test.ThreadNet.Util.NodeTopology as Topo
 import           Test.ThreadNet.Util.Seed (runGen)
 import qualified Test.Util.BoolProps as BoolProps
 import           Test.Util.HardFork.Future
+import           Test.Util.Nightly
 import           Test.Util.Orphans.Arbitrary ()
 import           Test.Util.Slots (NumSlots (..))
 
@@ -343,10 +344,25 @@ genPartition (NumCoreNodes n) (NumSlots t) (SecurityParam k) = do
 
     pure $ Partition (SlotNo firstSlotIn) (NumSlots dur)
 
+-- | Run relatively fewer tests
+--
+-- These tests are slow, so we settle for running fewer of them in this test
+-- suite since it is invoked frequently (eg CI for each push).
+twoFifthsTestCount :: QuickCheckTests -> QuickCheckTests
+twoFifthsTestCount (QuickCheckTests n) = QuickCheckTests $
+    if 0 == n then 0 else
+    max 1 $ (2 * n) `div` 5
+
 tests :: TestTree
 tests = testGroup "Cardano ThreadNet" $
-    [ testProperty "simple convergence" $ \setup ->
-          prop_simple_cardano_convergence setup
+    [ let name = "simple convergence" in
+      askIohkNightlyEnabled $ \enabled ->
+      if enabled
+      then testProperty name $ \setup ->
+             prop_simple_cardano_convergence setup
+      else adjustOption twoFifthsTestCount $
+           testProperty name $ \setup ->
+             prop_simple_cardano_convergence setup
     ]
 
 prop_simple_cardano_convergence :: TestSetup -> Property
--- a/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
+++ b/ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
@@ -367,21 +367,7 @@ prop_simple_cardano_convergence TestSetup
     tabulatePartitionDuration $
     tabulateFinalIntersectionDepth $
     tabulatePartitionPosition $
-    prop_general_semisync PropGeneralArgs
-      { pgaBlockProperty       = const $ property True
-      , pgaCountTxs            = fromIntegral . length . extractTxs
-      , pgaExpectedCannotForge = noExpectedCannotForges
-      , pgaFirstBlockNo        = 0
-      , pgaFixedMaxForkLength  = Just maxForkLength
-      , pgaFixedSchedule       =
-          -- the leader schedule isn't fixed because the Shelley leader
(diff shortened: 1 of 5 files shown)


ouroboros-consensus-shelley-test/test/Test/ThreadNet/RealTPraos.hs
```diff
@@ -75,6 +75,9 @@ minK = 5   -- Less than this increases risk of CP violations
 maxK :: Word64
 maxK = 10   -- More than this wastes execution time
 
+activeSlotCoeff :: Rational
+activeSlotCoeff = 0.5   -- TODO this is high
```
nfrisby, line 79: I'd like to merge with this TODO in place. This is the same value that was there. It's been unproblematic so far (because the RealTPraos test properties don't check for CP violations -- they are expected with such a high `f`), but I'd like to revisit it later.

ouroboros-consensus/ouroboros-consensus-test-infra/src/Test/ThreadNet/General.hs
```diff
@@ -850,27 +850,26 @@ prop_general_internal syncity pga testOutput =
   Final chains properties
 -------------------------------------------------------------------------------}
 
--- | What was the most number of blocks needed to be dropped from a final chain
```
nfrisby, line 853: This function was written in the context of `Cardano.hs` but was moved to the more general `General.hs` prematurely. This diff makes it actually generally enough to live in this module.

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
```diff
@@ -361,36 +362,40 @@ prop_simple_cardano_convergence TestSetup
...
       { initSeed
       , numCoreNodes
       , numSlots
       } = setupTestConfig
 
+    pga = PropGeneralArgs
+        { pgaBlockProperty       = const $ property True
+        , pgaCountTxs            = fromIntegral . length . extractTxs
+        , pgaExpectedCannotForge = noExpectedCannotForges
+        , pgaFirstBlockNo        = 1
```
nfrisby, line 388: This was 0, but should have been 1. These tests involve so many blocks that it doesn't really matter, but 1 is correct.

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
```diff
@@ -138,9 +138,10 @@ instance Arbitrary TestSetup where
                 -- Shelley epoch, since stake pools can only be created and
                 -- delegated to via Shelley transactions.
                 `suchThat` ((/= 0) . Shelley.decentralizationParamToRational)
-    setupK <- SecurityParam <$> choose (4, 6)
-                -- If k < 4, common prefix violations become too likely in
-                -- Praos mode.
+    setupK <- SecurityParam <$> choose (8, 10)
+                -- If k < 8, common prefix violations become too likely in
+                -- Praos mode for thin overlay schedules (ie low d), even for
+                -- f=0.2.
```
mrBliss, line 144: Could you measure how much longer the (non-nightly) tests now take?

nfrisby:
Talked to Thomas. Two todos:

1) Measure the change in run-time of the normal invocation, not nightly. If we don't accept Common Prefix violations and we want to reach later epochs, then we have to set `k` and `f` such that the epoch size is large and we have to run for correspondingly many slots. So it's "Common Prefix violations or relatively few slots to execute: choose one." I lowered the nightly reps to balance to preserve its total run time, but I didn't yet for the normal, non-nightly invocation.

Based on this diff, I think the non-nightly for default 100 iterations will take 4.5 minutes, which is too high for a local dev.

2) Why didn't this fail? https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/209#11427e76-5b83-48a3-8839-ebbf9d3daea6

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
```diff
@@ -770,8 +771,14 @@ mkProtocolCardanoAndHardForkTxs
...
 
+-- | The active slot coefficient, @f@.
+--
+-- Some of these tests includes Shelley epochs in which stakepools are actually
+-- leading. In that case, the @k@, @d@, and @f@ parameters and the length of
+-- any scheduled network partitions need to be balanced so that Common Prefix
+-- violations (in particular, wedges) are extremely unlikely.
 activeSlotCoeff :: Rational
-activeSlotCoeff = 0.5
+activeSlotCoeff = 0.2   -- chosen to match mainnet
```
mrBliss, line 793: I'm not sure I understand "chosen to match mainnet", because mainnet uses 0.05. Or is that "relative"?

mrBliss (approved):

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
nfrisby: Yeah, typo on my part. Thanks for catching.

ouroboros-consensus-cardano/test/Test/ThreadNet/Cardano.hs
nfrisby: The ThreadNet tests change as follows:

```
BEFORE   107.57s   96.51s    103.83s   99.63s

AFTER    223.75s   219.25s   229.55s   252.89s
```

So it's about 130% slower (which is similar to the nightly changes I made).

Option A) I'll set the non-nightly to use only 40 instead of 100.

Option B) Alter the test setup to reach the 3rd epoch less often in the non-nightly run.

I'm favoring A, for simplicity.

mrBliss: Yeah, A is fine for now.

nfrisby:
That rebase addressed Thomas's comments and picks up master changes (automatic, no conflicts). In non-nightly mode, the Cardano ThreadNet tests now run 40% of the requested number of cases (the default 100 induces 40 runs). That returns its default run-time to the ~100s from before this PR.

Commit 8bb7258 last night unexpectedly passed the tests. It's diff doesn't include any upstream changes, and is quite larger. Instead of combing it, I've just initiated another run of it: https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/210

Maybe 8bb7258  just got (relatively) lucky; if this run includes a failure, then I'll be satisfied that we haven't overlooked anything and I'll merge this Approve PR. Thanks, Thomas.

nfrisby:
Hmm. https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/210 passed :( But I reviewed the relevant diff and nothing is jumping out at me. So I'm trying again: https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/211 narrows to just Cardano, so it's going to have about twice as much opportunity to fail now. If this also passes, then either the Monday and Tuesday night commits each got unlucky once and the Wednesday night commit will have effectively gotten lucky 4 times in a row. In that case, I'll merge this anyway and open a ticket to follow-up later.

Also, I'll rebase this PR as soon bors manages to merge PR 2551 (reorg `-test-infra` and `-mock`) -- I'm expecting git to follow the renames and introduce no conflicts 🤞 .

nfrisby:
PR 2551 is stuck on a (Windows?) Hydra failure I don't understand.

https://hydra.iohk.io/jobset/Cardano/ouroboros-network-pr-2551#tabs-errors

See https://input-output-rnd.slack.com/archives/CAP8NM7N0/p1598557979036000 too -- thanks @johnalotoski !

So contrary to the original plan, I'm going to merge this PR first and hope @mrBliss can fix up #2551 tomorrow.

PR #2551, opened 2020-08-27
Relocate -test-infra and -mock packages

mrBliss:
Previously, `ouroboros-consensus-test-infra` (test infrastructure for consensus
tests) and `ouroboros-consensus-mock` (mock ledger integration + tests) were
subdirectories of the `ouroboros-consensus` package because the test suites of
`ouroboros-consensus` depended on the code in the two subdirectories and the
code in the two subdirectories depended on the library of `ouroboros-consensus`.

Both subdirectories also *doubled as packages*, so that other packages could
depend on them (e.g., `-cardano` depended on `-mock`, `-byron-test` depends on
`-test-infra`, etc.).

Having a directory both be a separate package and an extra source code directory
of an other package is annoying: you have to list the modules in the directory
and its dependencies in *both* cabal files. Inevitably, one of these gets out of
date. In fact, I bet that *both of them* were incomplete. It also makes it hard
to trim dependencies with tools like `weeder`. Moreover, GHCi always complained
because of `-Wmissing-home-modules`.

To fix this, do the same trick we did for `-byron(-test)` and `-shelley(-test)`,
namely, introduce a separate `-test` package:

* Move `./ouroboros-consensus/ouroboros-consensus-test-infra` to
  `./ouroboros-consensus-test` and turn it into a library.
* Split `./ouroboros-consensus/ouroboros-consensus-mock` into
  `./ouroboros-consensus-mock` and `./ouroboros-consensus-mock-test`.
* Move the testsuites of `ouroboros-consensus` to the new
  `ouroboros-consensus-test` package.
* Let both `ouroboros-consensus-test` and `ouroboros-consensus-mock(-test)`
  depend on `ouroboros-consensus`.
* Let `ouroboros-consensus-mock-test` depend on the `ouroboros-consensus-test` library.
* Let the testsuites in `ouroboros-consensus-test` depend on the
  `ouroboros-consensus-mock` library.

* For consistency, split `ouroboros-consensus-cardano` into
  `ouroboros-consensus-cardano` and `ouroboros-consensus-cardano-test`.

Now there is no more sharing of source code directories.


cabal.project
```diff
@@ -27,57 +28,27 @@ constraints:
   hedgehog >= 1.0,
   bimap >= 0.4.0
 
-package Win32-network
-  tests: True
+tests: True
```
mrBliss, line 31: Like in `cardano-node`: just enable all tests and disable those that we don't want (actually not needed anymore in Cabal 3.4! `source-repository-package`s are treated as real external packages instead of extra internal "source code". I'm keeping the code for the benefit of the users of 3.2.)

Now we can't forget to enable the tests of a newly introduced package.

mrBliss (commented):
When this PR and #2539 are merged, it'll finally be easy to run `weeder` :slightly_smiling_face:

nfrisby:
> Now there is no more sharing of source code directories.

🙌!

nfrisby:
bors retry

mrBliss:
I believe the CI problems are due to the fact that `ouroboros-consensus-test` does not contain a library, only test suites. See https://hydra.iohk.io/build/3965678/nixlog/3.

If I split `ouroboros-consensus-mock` into `ouroboros-consensus-mock` and `ouroboros-consensus-mock-test`, I can turn `ouroboros-consensus-test-infra` into the library of the `ouroboros-consensus-test` package. Remember that the tests in `ouroboros-consensus-test` depend on `ouroboros-consensus-mock` and `ouroboros-consensus-mock-test` on the `ouroboros-consensus-test` library :dizzy_face:.

mrBliss:
bors merge

mrBliss:
bors merge

cabal.project
mrBliss: My current theory is that this change is causing the evaluation errors on CI:
```
in job ‘x86_64-w64-mingw32.shell.x86_64-linux’:
The Haskell package set does not contain the package: binary-search (build dependency).
If you are using Stackage, make sure that you are using a snapshot that contains the package. Otherwise you may need to update the Hackage snapshot you are using, usually by updating haskell.nix.
```
My idea comes from https://github.com/input-output-hk/cardano-node/blob/77d7e394fc7747114df8e8121a31afcabc91c06e/cabal.project#L46, although there they *disable* instead of *enable* the tests.

mrBliss: Holy crap, that was it :roll_eyes:. I have to list each component explicitly for some reason. That was totally obvious from that error message :roll_eyes:.

PR #2555, opened 2020-08-28
Micro optimisations

mrBliss:
See the individual commit messages for more details.

Together, these reduce the time to sync the first 40k slots from a local `cardano-node` from ~45s to ~37s.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/AnchoredFragment.hs
+++ b/ouroboros-network/src/Ouroboros/Network/AnchoredFragment.hs
@@ -189,7 +189,13 @@ anchorIsGenesis Anchor{}      = False
 -- In other words, this would be the block immediately /before/ the other blocks
 -- in the fragment.
 anchorFromBlock :: HasHeader block => block -> Anchor block
-anchorFromBlock b = Anchor (blockSlot b) (blockHash b) (blockNo b)
+anchorFromBlock b = Anchor sno hash bno
+  where
+    HeaderFields {
+        headerFieldSlot    = sno
+      , headerFieldBlockNo = bno
+      , headerFieldHash    = hash
+      } = getHeaderFields b
 
 -- | Compute which 'Point' this anchor corresponds to
 anchorToPoint :: Anchor block -> Point block
--- a/ouroboros-network/src/Ouroboros/Network/AnchoredFragment.hs
+++ b/ouroboros-network/src/Ouroboros/Network/AnchoredFragment.hs
@@ -569,17 +569,25 @@ anchorNewest :: forall block. HasHeader block
              => Word64  -- ^ @n@
              -> AnchoredFragment block
              -> AnchoredFragment block
-anchorNewest n c = case c of
-    Empty _       -> c
-    _ | n >= len  -> c
-      | otherwise -> dropOldest (len - n) c
+anchorNewest n c
+    | toDrop <= 0
+    = c
+    | toDrop < 5
+      -- Hybrid approach: microbenchmarks have shown that a linear drop is
+      -- faster when the number of elements is small. For a larger number of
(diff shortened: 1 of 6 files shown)


ouroboros-network/src/Ouroboros/Network/AnchoredFragment.hs
```diff
@@ -540,17 +569,25 @@ anchorNewest :: forall block. HasHeader block
...
+    toDrop = len - fromIntegral n
 
-    dropOldest :: Word64 -> AnchoredFragment block -> AnchoredFragment block
-    dropOldest _ (Empty a) = Empty a
-    dropOldest 0 c'        = c'
-    dropOldest m (_ :< c') = dropOldest (m - 1) c'
+    linearDrop :: Int -> AnchoredFragment block -> AnchoredFragment block
+    linearDrop !_ (Empty a) = Empty a
+    linearDrop !0 c'        = c'
+    linearDrop !m (_ :< c') = linearDrop (m - 1) c'
```
coot, line 590: Do you need a bang patter if one of the pattern matches is a literal `0`?

mrBliss: Indeed not, but I added it for consistency and alignment. It's now clear that we're strict in the first argument.

ouroboros-network/src/Ouroboros/Network/AnchoredFragment.hs
```diff
@@ -785,13 +822,17 @@ mapAnchoredFragment f (AnchoredFragment a c) =
...
-       => (block -> Bool)  -- ^ Filtering predicate
-       -> AnchoredFragment block
-       -> [AnchoredFragment block]
-filter p = filterWithStop p (const False)
+filter ::
+     forall block. HasHeader block
+  => (block -> Bool)  -- ^ Filtering predicate
+  -> AnchoredFragment block
+  -> [AnchoredFragment block]
+filter p = filterWithStopSpec p (const False)
```
mrBliss, line 830: Oh, oops, this should be `filterWithStop`, this is a left over from testing. Will fix.

coot (approved):
Nice work, I would only add a test that compares `filterWithStop` against `filterWithStopSpec`.

mrBliss:
> Nice work, I would only add a test that compares `filterWithStop` against `filterWithStopSpec`.

Done.

mrBliss:
bors merge

mrBliss:
> # Canceled

Uggh, what's your problem?

mrBliss:
bors merge

dcoutts (commented):
Nice.

This reminds me that I wanted to change the representation of the final stages of the fetch fragments from a list of finger tree based sequences to just a list. We don't actually gain much from the use of a sequence, and the costs of rebuilding them is larger. There are also I think some opportunities for lazyness we could get from using lists in that stage.

mrBliss:
> This reminds me that I wanted to change the representation of the final stages of the fetch fragments from a list of finger tree based sequences to just a list. We don't actually gain much from the use of a sequence, and the costs of rebuilding them is larger. There are also I think some opportunities for lazyness we could get from using lists in that stage.

My first idea was also to move to a different representation than fragments, but I couldn't immediately come up with a good way to handle https://github.com/input-output-hk/ouroboros-network/blob/65be742e2063e069f6144f07387fcc0f61e16324/ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientState.hs#L331-L333
There, we want to know the last element of the list so we can decide to join them. Using lists there means a linear, but possibly lazy, traversal.

It would also have meant a much more invasive change, whereas this PR just improves a single function.

But indeed, this could be further improved.

PR #2556, opened 2020-08-28
Remove Typeable constraints from MuxPeer

mrBliss:
The `Typeable` constraints in question are not used, but are missing in
`cardano-node`.


mrBliss:
bors merge

Issue #2506, opened 2020-08-11
Off-by-one error in TPraos history


nc6:
As reported by @nfrisby :+1: 
First:
The anchor is the earliest slot to which we can roll back. It should correspond to the first slot with an entry in the historical state.
https://github.com/input-output-hk/ouroboros-network/blob/62e57b9673e0140c87c6edd6ca8fb7b813c9f8d9/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/State.hs#L45-L57
Second: the sequence of historical states -- which includes the current state -- never has more than k entries.
https://github.com/input-output-hk/ouroboros-network/blob/62e57b9673e0140c87c6edd6ca8fb7b813c9f8d9/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs#L403-L407
So we can only roll back to the first entry, which would involve dropping at most k-1 blocks.
Is my reasoning here sound?
If so, that is off-by-one with respect to the semantics of k:
k == 1: we can roll back at most one block
https://github.com/input-output-hk/ouroboros-network/blob/62e57b9673e0140c87c6edd6ca8fb7b813c9f8d9/ouroboros-consensus/src/Ouroboros/Consensus/Config/SecurityParam.hs#L15-L24
This is in some sense a minor discrepancy, considering k=2160 in production. But still, is it intentional/recorded anywhere?
Edit: Maybe a fix is as simple as instead pruning to a length of k+1, since the sequence includes the current state, IIUC.

mrBliss:
I was mistaken, #2546 does not fix this issue. I got confused because `TPraosState` also keeps snapshots, but that's the `ChainDepState` of Shelley, not the `LedgerState` of Shelley.

I will fix it now, by maintaining removing the ability to rewind a `ChainDepState` and keeping the snapshots in the `HeaderState` instead of having to reimplement the snapshots for each ledger (Byron as well).

Note that the `HeaderState` will still contain snapshots, while we keep the last `k` `HeaderState`s around as part of the LedgerDB, so that's again snapshots within snapshots. I have some ideas on how to get rid of this too, but it's less straightforward than with the `LedgerState`, because we also use the `HeaderState` in the ChainSyncClient, where the `HeaderState` can be more than `k` blocks ahead of any `HeaderState` in the LedgerDB, but we might still need to roll it back `k` blocks when the upstream peer does so.

[...]

PR #2575, opened 2020-09-02
Remove rewindChainDepState and simplify the HeaderState

### LINKED ISSUE #2571: Remove rewindChainDepState
The `HeaderState` is defined as follows:
```haskell
data HeaderState blk = HeaderState {
      -- | Protocol-specific state
      headerStateConsensus :: !(ChainDepState (BlockProtocol blk))

      -- | The most recent @k@ tips
    , headerStateTips      :: !(StrictSeq (AnnTip blk))

      -- | Tip /before/ 'headerStateTips'
    , headerStateAnchor    :: !(WithOrigin (AnnTip blk))
    }
```

To roll back the `HeaderState`, we roll back `headerStateTips` and call `rewindChainDepState` on `headerStateConsensus`. This means that `ChainDepState` must maintain snapshots of itself in order to facilitate rollback. Each ledger with non-trivial `ChainDepState` must do snapshotting, which is error-prone (#2506, also see the many bugs related to `PBftState`).

We can simplify this by reusing the snapshots made of the `HeaderState` for the `ChainDepState`. Then this logic is implemented in a single place and doesn't have to be reimplemented for each ledger. It also simplifies the `ConsensusProtocol` type class by removing the `rewindChainDepState` method (also removing the `Serialise (HeaderHash hdr)` wart).

---

Note that the `HeaderState` will still contain snapshots, while we keep the last `k` `HeaderState`s around as part of the LedgerDB, so that's again snapshots within snapshots. I have some ideas on how to get rid of this too, but it's less straightforward than with the `LedgerState` (which we did in #2546), because we also use the `HeaderState` in the ChainSyncClient, where the `HeaderState` can be more than `k` blocks ahead of any `HeaderState` in the LedgerDB, but we might still need to roll it back `k` blocks when the upstream peer does so.

My plan is to store `HeaderState`s without snapshots in the LedgerDB and use the LedgerDB's snapshots to create a range of `k` `HeaderState`s, which can then be used as the starting point for the ChainSyncClient.

mrBliss:
Fixes #2571, #2383, and #2506.

**WARNING:** this breaks binary compatibility of Byron and Shelley ledger
snapshots. The ledger will have to be replayed from genesis!

Old situation
-------------

The `HeaderState` is defined as follows:

```haskell
data HeaderState blk = HeaderState {
      -- | Protocol-specific state
      headerStateConsensus :: !(ChainDepState (BlockProtocol blk))

      -- | The most recent @k@ tips
    , headerStateTips      :: !(StrictSeq (AnnTip blk))

      -- | Tip /before/ 'headerStateTips'
    , headerStateAnchor    :: !(WithOrigin (AnnTip blk))
    }
```

To roll back the `HeaderState`, we roll back `headerStateTips` and call
`rewindChainDepState` on `headerStateConsensus`, which should restore the
`ChainDepState` to what it was for any of the last `k` headers. This means that
`ChainDepState` must itself maintain snapshots in order to facilitate rollback.

Each ledger with a non-trivial `ChainDepState` must implement its own
snapshotting, which is error-prone (#2506, also see the many bugs related to
`PBftState`).

What's worse is that we're dealing with *snapshots within snapshots*: the
`HeaderState` is part of the `ExtLedgerState` maintained by the `LedgerDB`, so
we store a snapshot for each of the last `k` `HeaderState`s. Each of those `k`
`HeaderState`s stores the last `k` tips and a `ChainDepState`, which has
snapshots going back `k` blocks. This means that after syncing `2k` blocks, the
oldest snapshot of the `HeaderState` in the `LedgerDB` will keep contain `k`
tips, of which all but one are older than `k` anyway. Similarly for the
`ChainDepState` stored in that `HeaderState`, it will contain information older
than `k`.

New situation
-------------

Simplify this by not letting `HeaderState` and `ChainDepState` be rewindable.
Instead, reuse the snapshots made of the `HeaderState` in the `LedgerDB` for the
`ChainDepState`. This snapshotting logic is already in place and will not have
to be reimplemented for each ledger.

The new `HeaderState` looks like this:

```haskell
data HeaderState blk = HeaderState {
      headerStateTip   :: !(WithOrigin (AnnTip blk))
    , headerStateState :: !(ChainDepState (BlockProtocol blk))
    }
```

We remove the `rewindChainDepState` method and the need for the `ChainDepState`
to maintain snapshots itself. This significantly simplifies the implementation
for Byron (`PBftState`) and Shelley (`TPraosState`). So much that they're now
almost trivial. As a bonus, it removes the `Serialise (HeaderHash hdr)` wart
that we needed because of EBBs.

That's it for header validation when done as part of block validation by the
`ChainDB`: we simply reuse the snapshots in the `LedgerDB` to validate the
header just like we do for the block body.

However, it's a *different story* for header validation when done by the
ChainSync client. The important difference is that the candidate header chain
might be *ahead* of our own chain or *on a different fork*, so that we can't use
the snapshots in the `LedgerDB` to validate the incoming headers. For example,
when bulk syncing from scratch in the Byron era, we can look `2k` slots ahead
and thus might have to validate roughly `2k` headers with an empty `LedgerDB`.
When the upstream node is more than `k` headers ahead of us and rolls back `k`
headers, we must be able to support it. This means we need another kind of
snapshotting for the `HeaderState` when used by the ChainSync client.

Hence, we introduce:

```haskell
data HeaderStateHistory blk = HeaderStateHistory {
      -- | The last, right-most, element in the sequence contains the current
      -- state, see 'headerStateHistoryCurrent'.
      headerStateHistorySnapshots :: !(StrictSeq (HeaderState blk))

      -- | Header state /before/ the oldest in 'headerStateHistorySnapshots'.
      -- This is the oldest tip we can roll back to.
    , headerStateHistoryAnchor    :: !(HeaderState blk)
    }
```

This type is only used by the ChainSync client (so not again /snapshots withing
snapshots/). We already had a separate `validateHeader'` function for the
ChainSync client, so we modify that one to maintain a `HeaderStateHistory`.

When the ChainSync client connects to an upstream node, it populates its
`HeaderStateHistory` using the last `k` snapshots of the `LedgerDB` with the new
`ChainDB.getHeaderStateHistory` function. Any newly synced headers are appended
to the history. When we trim the synced fragment, we trim the
`HeaderStateHistory` in the same way, keeping them both aligned at all
times (checked by a new invariant).


ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
```diff
@@ -172,98 +191,141 @@ tickHeaderState :: ConsensusProtocol (BlockProtocol blk)
...
+-- rewinding that header state again by @j@ where @i + j > k@ is not possible
+-- and will yield 'Nothing'.
+rewindHeaderStateHistory ::
+     forall blk. (BlockSupportsProtocol blk, HasAnnTip blk)
+  => Point blk
+  -> HeaderStateHistory blk -> Maybe (HeaderStateHistory blk)
+rewindHeaderStateHistory p HeaderStateHistory {..} =
+    case Seq.dropWhileR rolledBack headerStateHistorySnapshots of
+        Empty
+          | pointToWithOriginRealPoint p /=
```
nfrisby, line 312: Simplify guard to `rolledBack headerStateHistoryAnchor`?

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
```diff
@@ -172,98 +191,141 @@ tickHeaderState :: ConsensusProtocol (BlockProtocol blk)
...
-castHeaderState :: ( Coercible (ChainDepState (BlockProtocol blk ))
-                               (ChainDepState (BlockProtocol blk'))
-                   , TipInfo blk ~ TipInfo blk'
-                   )
-                => HeaderState blk -> HeaderState blk'
-castHeaderState HeaderState{..} = HeaderState{
-      headerStateConsensus = coerce headerStateConsensus
-    , headerStateTips      = castSeq castAnnTip $ headerStateTips
-    , headerStateAnchor    = fmap    castAnnTip $ headerStateAnchor
+-- | Trim the 'HeaderStateHistory' to the given size, dropping the oldest
```
nfrisby, line 252: `headerStateHistoryTrim 5 x` will result in a `HeaderStateHistory` data structure that ultimately contains 6 header states, right? (It's +1 because I'm counting the anchor state.) That makes this comment a little ambiguous.

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
mrBliss: Nice! :pray:

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
```diff
@@ -102,24 +101,11 @@ instance CanMock era => Arbitrary (Point (ShelleyBlock era)) where
...
-         -> WithOrigin SlotNo
-         -> TPraosState era
-         -> Gen (TPraosState era)
-      go steps prevSlot st
-        | 0 <- steps = return st
-        | otherwise  = do
-          let slot = withOrigin (SlotNo 0) succ prevSlot
-          newPrtclState <- arbitrary
-          go (steps - 1) (NotOrigin slot) (TPraosState.append slot newPrtclState st)
+      TPraosState lastSlot <$> arbitrary
```
nfrisby, line 108: I'm surprised that this generator can choose `lastSlot` independently of `tpraosStateChainDepState`.

Is it only used in properties testing low-level things like serialization eg?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -488,9 +485,9 @@ getFilteredDelegationsAndRewardAccounts ss creds =
   Serialisation
 -------------------------------------------------------------------------------}
 
-serialisationFormatVersion0 :: VersionNumber
-serialisationFormatVersion0 = 0
-
+-- | 'serialisationFormatVersion0' used to include the 'LedgerViewHistory', but
```
nfrisby, line 488: Use `@` instead of `'` to prevent these Haddock references to `*Version0` from emitting warnings.

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
```diff
@@ -172,98 +191,145 @@ tickHeaderState :: ConsensusProtocol (BlockProtocol blk)
...
     castSeq :: (a -> b) -> StrictSeq a -> StrictSeq b
-    castSeq f = Seq.fromList . map f . toList
+    castSeq f = Seq.toStrict . fmap f . Seq.getSeq
 
--- | Rewind the header state
+-- | Rewind the header state history
 --
--- This involves 'rewindChainState', and so inherits its PRECONDITION that the
--- target point must have been previously applied.
+-- PRECONDITION: the point to rewind to must correspond to a header (or
```
nfrisby, line 295: Does an EBB count as a "header" here? I think so, since this is used in ChainSync client.

I would vote for explicitly stating whether EBBs are relevant to "header" or "block" just about everywhere the words are used in a comment, sadly.

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
```diff
@@ -172,98 +191,145 @@ tickHeaderState :: ConsensusProtocol (BlockProtocol blk)
...
+-- rewinding that header state again by @j@ where @i + j > k@ is not possible
+-- and will yield 'Nothing'.
+rewindHeaderStateHistory ::
+     forall blk. (BlockSupportsProtocol blk, HasAnnTip blk)
+  => Point blk
+  -> HeaderStateHistory blk -> Maybe (HeaderStateHistory blk)
+rewindHeaderStateHistory p HeaderStateHistory {..} =
+    case Seq.dropWhileR rolledBack headerStateHistorySnapshots of
+        Empty
+          | rolledBack headerStateHistoryAnchor
```
nfrisby, line 316: Should this alternative assert that the slot part of `p` is less than (<= b/c of EBBs?) the slot part of `anchor`?

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
```diff
@@ -310,8 +302,8 @@ instance PBftCrypto c => ConsensusProtocol (PBft c) where
                       slot
                       (TickedPBftState (TickedPBftLedgerView dms) state) =
       case toValidate of
```
nfrisby, line 304: This is a nice simplification to see. EBBs were only tracked by PBFT state to enable rewind. No more rewind means no more EBB code 🙌

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/State.hs
```diff
@@ -23,17 +23,11 @@ module Ouroboros.Consensus.Protocol.PBFT.State (
     -- * Construction
   , empty
   , append
-  , appendEBB
-  , rewind
     -- * Queries
   , countSignatures
-  , countInWindow
```
nfrisby, line 30: Was dead code:

```
$ git grep -e countInWindow origin/master -- '*hs'
origin/master:ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/State.hs:  , countInWindow
origin/master:ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/State.hs:countInWindow :: PBftState c -> Word64
origin/master:ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/State.hs:countInWindow PBftState{..} = size inWindow
```

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/State.hs
```diff
@@ -360,155 +202,35 @@ tipSlot cs@PBftState{..} = lastSignedSlot cs `max` ebbsMax ebbs
...
-        preAnchor  = preAnchor'
-      , postAnchor = postAnchor'
-      , preWindow  = preWindow'
-      , inWindow   = inWindow'
+append ::
+     forall c. PBftCrypto c
+  => WindowSize
+  -> PBftSigner c
+  -> PBftState c -> PBftState c
+append n signer@(PBftSigner _ gk) PBftState{..} =
```
nfrisby, line 217: Minor: this might be more obviously correct if we split it into a grow step and shrink step.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/State.hs
```diff
@@ -580,32 +274,39 @@ fromList k n (anchor, signers, ebbs) =
...
+      Foldable.foldl'
+        (\acc (PBftSigner slot key) -> Map.insertWith (<>) key [slot] acc)
+        Map.empty
+    . inWindow
+
+decompact :: PBftCrypto c => Map (PBftVerKeyHash c) [SlotNo] -> PBftState c
+decompact =
+      fromList
+    . sortOn pbftSignerSlotNo
+      -- We use a difference list to avoid repeated (7 lists in total)
```
nfrisby, line 288: Minor: maybe just use `concatMap`, since the outer list is always just 7?

nfrisby:
I'm going to rephrase this part to make sure I understand and setup a question below.

> However, it's a _different story_ for header validation when done by the
> ChainSync client. The important difference is that the candidate header chain
> might be _ahead_ of our own chain or _on a different fork_, so that we can't use
> the snapshots in the `LedgerDB` to validate the incoming headers.

Before this PR, the ChainSync client included the server's `HeaderState` and could rewind that header state in order to handle a `RollBackward` message.

After this PR, `HeaderState` is no longer rewindable. Thus the ChainSync client now maintains a `HeaderState` for each point on _their_ fragment (which includes the anchor = intersection with our fragment).

How long can their fragment be? That's limited only by the availability of a ledger view. The necessary ledger view is acquired by looking up the past ledger view corresponding to the intersection of our fragment and their fragment and then forecasting that view ahead to the slot determined by each `RollForward`. So the maximum forecast determines the upper bound on the length of their fragment. For Shelley, at least, this is `stabilityWindow`, which is currently 3k/f=129600 (plus 1 for the anchor) on mainnet.

That would be a lot, per peer. But we can reduce it via a few typical assumptions.

  - In a healthy net, the Common Prefix invariant ensures a node is essentially never more than `k` blocks behind any other node, so that lowers the worse-case to just 2160 (plus 1 for the anchor) -- except for during bulk-sync.
  - Moreover, the peer's chain -- even if it's that long -- probably won't be fully dense (it's ~5% dense with the current f=0.05 value).

So the technical worst-case (100% density, full stability window behind) is 129601 headers & header-states per peer, but the realistic worst-case (100% density, k behind) is 2161, and the typical worst-case is actually ~109 (f density, k behind). The average case will likely be in the tens per peer, I think, assuming a reasonable chain propagation delay (ie _message delay_).

(During bulk sync, assuming ChainSync is faster than BlockFetch, will each upstream peer's candidate fragment likely grow to `3k/f * f = 3k = 6480`?)

Except for the extreme cases, a significant portion of that additional footprint is likely balanced by this PR's removal of snapshots from each individual `HeaderState`.

Please correct me here if these cocktail-napkin calculations are off the mark. If all of the above is correct, maybe it deserves to be included in a large comment somewhere.

nfrisby:
I've assessed all the diffs, and nothing looked egregious. I spent significant time on some diffs, but not all. Minor comments throughout (several EBB related), plus a high-level question regarding overhead.

This is big enough that I think a phone call to discuss it would be justified, so I'm not clicking Approve yet. Feel free to schedule that phone call with me, if you agree.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -320,7 +318,7 @@ translateChainDepStateByronToShelley
   -> PBftState bc
   -> TPraosState era
 translateChainDepStateByronToShelley TPraosConfig { tpraosParams } pbftState =
-    TPraosState.empty (PBftState.tipSlot pbftState) $
```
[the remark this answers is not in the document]
mrBliss: Good point. In practice it's impossible that the last Byron slot is occupied by an EBB:
* On mainnet we stopped producing EBBs a while before the transition
* The transition happens at the start of an epoch, so if the last was occupied by an EBB, it must have been the EBB the start of the previous epoch. This means the previous epoch must have been empty, which is a violation of the "`k` blocks per `2k` slots" property.

I'll add a comment.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
mrBliss: Exactly, it's only used for the serialisation roundtrip tests.

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
mrBliss: Done.

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
mrBliss: Done. I have added `headerStatePoint` in the process, which is already an improvement.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT.hs
mrBliss: Indeed, this made me very happy and was a sign to me that this is a good refactoring :slightly_smiling_face:.

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/State.hs
mrBliss: Done

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/State.hs
mrBliss: Yeah, you're right, premature optimisation, etc. Each list of the 7 lists will have 2160/7 elements. This is not a known hotspot, so simplicity trumps performance.

mrBliss:
@nfrisby Thanks for going through this :pray: 

I have add a large comment to the ChainSync client discussing what you astutely pointed out below.

> I'm going to rephrase this part to make sure I understand and setup a question below.
> 
> > However, it's a _different story_ for header validation when done by the
> > ChainSync client. The important difference is that the candidate header chain
> > might be _ahead_ of our own chain or _on a different fork_, so that we can't use
> > the snapshots in the `LedgerDB` to validate the incoming headers.
> 
> Before this PR, the ChainSync client included the server's `HeaderState` and could rewind that header state in order to handle a `RollBackward` message.
> 
> After this PR, `HeaderState` is no longer rewindable. Thus the ChainSync client now maintains a `HeaderState` for each point on _their_ fragment (which includes the anchor = intersection with our fragment).
> 
> How long can their fragment be? That's limited only by the availability of a ledger view. The necessary ledger view is acquired by looking up the past ledger view corresponding to the intersection of our fragment and their fragment and then forecasting that view ahead to the slot determined by each `RollForward`. So the maximum forecast determines the upper bound on the length of their fragment. For Shelley, at least, this is `stabilityWindow`, which is currently 3k/f=129600 (plus 1 for the anchor) on mainnet.
> 
> That would be a lot, per peer. But we can reduce it via a few typical assumptions.
> 
>     * In a healthy net, the Common Prefix invariant ensures a node is essentially never more than `k` blocks behind any other node, so that lowers the worse-case to just 2160 (plus 1 for the anchor) -- except for during bulk-sync.
> 
>     * Moreover, the peer's chain -- even if it's that long -- probably won't be fully dense (it's ~5% dense with the current f=0.05 value).
> 
> 
> So the technical worst-case (100% density, full stability window behind) is 129601 headers & header-states per peer, but the realistic worst-case (100% density, k behind) is 2161, and the typical worst-case is actually ~109 (f density, k behind). The average case will likely be in the tens per peer, I think, assuming a reasonable chain propagation delay (ie _message delay_).
> 
> (During bulk sync, assuming ChainSync is faster than BlockFetch, will each upstream peer's candidate fragment likely grow to `3k/f * f = 3k = 6480`?)

I'm ignoring a 100% density for TPraos and always sticking to `f`. So indeed a look-ahead of `3k` headers. In fact, since "their" fragment shares the anchor point with "our" fragment, which is `k` headers long (unless near genesis), the length of "their" fragment and header state history is `4k = 8640` headers in the worst case. (For more details, see my new comment in the ChainSync client.)

This is indeed significant.

A possible solution is to "pause" rolling forward when the number of headers on `theirFrag` reaches some limit. `theirFrag` will always include the last `k` headers of our current chain, so the limit should at least be `k`. To allow pipelining by the ChainSync client, is should be more than `k`, but it doesn't have to be `8k`. Slightly related: #423.
cc: @edsko @dcoutts 


> 
> Except for the extreme cases, a significant portion of that additional footprint is likely balanced by this PR's removal of snapshots from each individual `HeaderState`.
> 
> Please correct me here if these cocktail-napkin calculations are off the mark. If all of the above is correct, maybe it deserves to be included in a large comment somewhere.

Indeed, as I mention above, I have added a large comment in the ChainSync client.

> I've assessed all the diffs, and nothing looked egregious. I spent significant time on some diffs, but not all. Minor comments throughout (several EBB related), plus a high-level question regarding overhead.
> 
> This is big enough that I think a phone call to discuss it would be justified, so I'm not clicking Approve yet. Feel free to schedule that phone call with me, if you agree.

Since @edsko is back on Monday, I was planning to wait until then to review it with him (one less thing he has to catch up on). If you like, you're welcome to join the call in which we review it, let me know.

ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
```diff
@@ -283,15 +281,8 @@ instance Arbitrary (PBftState PBftByronCrypto) where
...
-          st' <-
-            if ebb then do
-              headerHashBytes <-
-                PBftState.headerHashBytes (Proxy @ByronBlock) <$> arbitrary
-              return $ PBftState.appendEBB k windowSize slot headerHashBytes st
-            else do
-              newSigner <- PBftState.PBftSigner slot <$> arbitrary
-              return $ PBftState.append k windowSize newSigner st
+          newSigner <- PBftState.PBftSigner slot <$> arbitrary
+          let st' = PBftState.append windowSize newSigner st
```
mrBliss, line 285: Use `PBfState.fromList`

ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node/Serialisation.hs
```diff
@@ -65,7 +65,7 @@ instance EncodeDisk ByronBlock (PBftState PBftByronCrypto) where
   encodeDisk _ = encodeByronChainDepState
 -- | @'ChainDepState' ('BlockProtocol' 'ByronBlock')@
 instance DecodeDisk ByronBlock (PBftState PBftByronCrypto) where
-  decodeDisk ccfg = decodeByronChainDepState (getByronSecurityParam ccfg)
+  decodeDisk _ = decodeByronChainDepState
```
mrBliss, line 68: Extend the golden test infrastructure to be able to test backwards compatibility of things. Use in combination with `Versioned`? Create an issue.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
```diff
@@ -315,13 +318,41 @@ newtype instance Ticked (SL.LedgerView era) = TickedPraosLedgerView {
...
+data TPraosState era = TPraosState {
+      tpraosStateLastSlot      :: !(WithOrigin SlotNo)
+    , tpraosStateChainDepState :: !(SL.ChainDepState era)
+    }
+  deriving (Generic, Show, Eq)
+
+instance Era era => NoUnexpectedThunks (TPraosState era)
+
+serialisationFormatVersion0 :: VersionNumber
+serialisationFormatVersion0 = 0
```
mrBliss, line 334: 1

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
```diff
@@ -362,6 +365,34 @@ isValid :: forall blk. LedgerSupportsProtocol blk
...
+    volChain = volatileChain (configSecurityParam cfg) id m
+
+    headerStates :: Map (Point blk) (HeaderState blk)
+    headerStates = snd $
+        foldl'
+          (\(st, states) blk ->
+              let st' = tickThenReapply (extLedgerCfgFromTopLevel cfg) blk st
+              in (st', Map.insert (blockPoint blk) (headerState st') states))
+          (initLedger, Map.singleton GenesisPoint (headerState initLedger))
+          (Chain.toOldestFirst (currentChain m))
```
mrBliss, line 394: Move something like this (`Chain -> HeaderStateHistory`?) in `Test.Util.Chain` (or `HeaderStateHistory` itself?) and share it with the ChainSyncClient test.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Protocol.hs
```diff
@@ -107,7 +106,6 @@ instance CanHardFork xs => ConsensusProtocol (HardForkProtocol xs) where
   checkIsLeader         = check
   updateChainDepState   = update
   reupdateChainDepState = reupdate
-  rewindChainDepState _ = rewind
```
mrBliss, line 109: Remove snapshots from `PastState` and `retractToSlot`.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -185,6 +185,68 @@ bracketChainSyncClient tracer ChainDbView { getIsInvalidBlock } varCandidates
 -- TODO #467 if the 'theirTip' that they sent us is on our chain, just
 -- switch to it.
 
+
+-- = Candidate fragment size
+-- -----------------------
```
mrBliss, line 190: Extra `-`

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -185,6 +185,68 @@ bracketChainSyncClient tracer ChainDbView { getIsInvalidBlock } varCandidates
...
+-- >              V         V           V
+-- > 'ourFrag':   A-H-H-H-H-H-H-H-H-...-H
+-- >                     \
+-- > 'theirFrag':         H-H-H-H-...   ...   ...
+-- >                    ^
+-- >                    |
+-- >           most recent intersection (between A and the tip)
+--
+-- Note that the 'ourFrag' and 'theirFrag' share anchors /at all times/. In the
+-- figure above, the first two headers on 'ourFrag' are thus also on
```
mrBliss, line 220:
```suggestion
-- figure above, the first three headers on 'ourFrag' are thus also on
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
```diff
@@ -174,6 +175,10 @@ data ChainDB m blk = ChainDB {
       -- returned.
     , getPastLedger      :: Point blk -> STM m (Maybe (ExtLedgerState blk))
 
+      -- | Get a 'HeaderStateHistory' populated with the 'HeaderState's of the
+      -- last @k@ blocks of the current chain.
+    , getHeaderStateHistory :: STM m (HeaderStateHistory blk)
```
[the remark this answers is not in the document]
mrBliss: See #2594 and #2595.

ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node/Serialisation.hs
mrBliss: See #2596.

edsko (approved):
Very nice, always nice when a single refactoring (ledger DB) leads to so many simplifications.

mrBliss:
bors merge

Issue #2479, opened 2020-07-30
Keep older ledger snapshots


mrBliss:
It is likely that we'll break the compatibility of the ledger snapshots of Shelley. This will require replaying the whole chain from genesis, including the entire Byron era.

If we were to keep older snapshots around, e.g., one every 50 epochs, then we could limit the required replay whenever we break the Shelley snapshots.

This will require a change in the snapshot policy, because we currently delete older snapshots. We'll have to remember not to delete some of them.

---

**UPDATE:** https://github.com/input-output-hk/ouroboros-network/pull/2733 already implements part of this issue: name snapshots after their slot number, allow named snapshots that won't be deleted, etc.

This issue is now only about making the snapshot policy more flexible.

It should also be possible to manually specify via the CLI of which slots to take snapshots. This means that we should also make snapshots during ledger replay.

mrBliss:
Ideas:

* Instead of numbering snapshots, use the corresponding slot number as the name. That makes it much clearer of what it is a snapshot. Note that EBBs have the same slot number as the block after it. I think that's fine, as long as we just use the slot number as a name and never rely on it being the right slot number.

* Allow suffixing the snapshot with a name, in which case it will never be deleted (and excluded from the on-disk snapshot limit). For example `4492799.last_ByronBlock`. This would come in handy for node operators when the Shelley ledger format changes, it would avoid a replay from genesis. It would also be useful for testing/benchmarking/profiling.

mrBliss:
More ideas:

* Allow for a linear number of snapshots in the length of the chain, e.g., one every `x` epochs`.

* Make a snapshot at the transition.

mrBliss:
Relevant code:

* https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
* https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs

kderme:
I was wondering if we can tell the real slot of a disk snapshot in a fast way (preferably without having to parse the whole snapshot). I think this can be done, since we store a `ChainSummary` and not just a ledgerState, so by tweaking the CBOR parsers we could tell the slot by reading only a few bytes.

[...]

PR #2576, opened 2020-09-02
Keep older ledger snapshots

kderme:
https://github.com/input-output-hk/ouroboros-network/issues/2479

That's at a very early state, but some remarks so far:
- The filename for Genesis  is a special case. I'm using "genesis_..".
- If a slot already exists when we try to write a snapshot, we now just avoid writing it again. I think this is what this comment is refering to:
`-- TODO avoid taking multiple snapshots corresponding to the same tip.`

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
@@ -21,6 +21,7 @@ module Ouroboros.Consensus.Storage.LedgerDB.OnDisk (
     -- * Write to disk
   , takeSnapshot
   , trimSnapshots
+  , mkDiskSnapshot
     -- * Low-level API (primarily exposed for testing)
   , DiskSnapshot -- opaque
   , deleteSnapshot
@@ -37,7 +38,8 @@ import           Control.Monad.Except
 import           Control.Tracer
 import qualified Data.Bifunctor.TH as TH
 import qualified Data.List as List
-import           Data.Maybe (mapMaybe)
+import           Data.Maybe (isNothing, mapMaybe)
+import           Data.Ord
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import           Data.Word
@@ -203,7 +205,9 @@ initLedgerDB replayTracer
                              s
         case ml of
           Left err -> do
-            deleteSnapshot hasFS s
+            when (isNothing (dsSuffix s)) $
+              -- We don't delete named snapshots, even if we couldn't parse them
+              deleteSnapshot hasFS s
             traceWith tracer $ InvalidSnapshot s err
             tryNewestFirst (acc . InitFailure s err) ss
           Right (r, l, replayed) ->
@@ -274,9 +278,15 @@ initStartingWith tracer conf streamAPI initDb = do
 -- testing purposes, 'takeSnapshot' returns the block reference corresponding
 -- to the snapshot that we wrote.
 --
--- NOTE: This is a lower-level API that unconditionally takes a snapshot
--- (i.e., independent from whether this snapshot corresponds to a state that
--- is more than @k@ back).
+-- NOTE: This is a lower-level API that takes a snapshot independent from
+-- whether this snapshot corresponds to a state that is more than @k@ back. It
+-- doesn't take a snapshot if a snapshot at the same slot already exists on
+-- disk or if the tip is at genesis.
+--
+-- Note that an EBB can have the same slot number as the block after it. This
+-- doesn't matter. The one block difference in the ledger state doesn't warrant
+-- an additional snapshot. The slot number in the name of the snapshot is only
+-- indicative, we don't rely on it being correct.
 --
 -- TODO: Should we delete the file if an error occurs during writing?
 takeSnapshot :: forall m l r h. MonadThrow m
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -317,13 +324,18 @@ trimSnapshots tracer hasFS DiskPolicy{..} = do
 -------------------------------------------------------------------------------}
 
 -- | On disk snapshots are numbered monotonically
-newtype DiskSnapshot = DiskSnapshot Int
+data DiskSnapshot = DiskSnapshot {
+    dsSlotNo :: !(WithOrigin SlotNo)
+  , dsSuffix :: !(Maybe String)
+  }
```
mrBliss, line 330: Let's document it:
```suggestion
data DiskSnapshot = DiskSnapshot {
      -- | We name snapshots after the slot number the ledger state corresponds
      -- to. This gives an indication of how recent the snapshot is.
      --
      -- Note that the snapshot names are only indicative, we don't rely on them
      -- having the correct slot number. We only use the names to determine the
      -- order in which we try them. If a user chooses to rename their
      -- snapshots, we'll just try them in a different order. That order might
      -- be suboptimal, i.e., trying older snapshots requiring more replay
      -- first.
      dsSlotNo :: SlotNo
      -- | Snapshots can optionally have a suffix, separated by the slot number
      -- with an underscore, e.g., @4492799_last_Byron@. This suffix acts as
      -- metadata for the operator of the node. Snapshots with a suffix will
      -- /not be trimmed/.
    , dsSuffix :: Maybe String
    }
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -317,13 +324,18 @@ trimSnapshots tracer hasFS DiskPolicy{..} = do
 -------------------------------------------------------------------------------}
 
 -- | On disk snapshots are numbered monotonically
-newtype DiskSnapshot = DiskSnapshot Int
+data DiskSnapshot = DiskSnapshot {
+    dsSlotNo :: !(WithOrigin SlotNo)
```
mrBliss, line 328: There's no point in making a snapshot of the ledger state at genesis, we are given this state at startup.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -353,24 +365,43 @@ writeSnapshot hasFS encLedger encRef ss cs = do
...
+    pref = case dsSlotNo of
+      Origin       -> "genesis"
+      NotOrigin sl -> show $ unSlotNo sl
+    suff = case dsSuffix of
+      Nothing  -> ""
+      Just str -> "_" ++ str
 
 snapshotFromPath :: String -> Maybe DiskSnapshot
-snapshotFromPath = fmap DiskSnapshot . readMaybe
+snapshotFromPath fileName = fmap (flip DiskSnapshot fixedSuff) (readSlot pref)
```
mrBliss, line 392: Using `do` and record fields might be more readable here.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -276,24 +276,31 @@ initStartingWith tracer conf streamAPI initDb = do
...
              -> HasFS m h
              -> (l -> Encoding)
              -> (r -> Encoding)
-             -> LedgerDB l r -> m (DiskSnapshot, WithOrigin r)
-takeSnapshot tracer hasFS encLedger encRef db = do
-    ss <- nextAvailable <$> listSnapshots hasFS
-    writeSnapshot hasFS encLedger encRef ss oldest
-    traceWith tracer $ TookSnapshot ss (csTip oldest)
-    return (ss, csTip oldest)
+             -> (r -> SlotNo)
```
mrBliss, line 288: Hmm, this is a sign of a leaky abstraction. The `LedgerDB` is parametric in `r`, yet here we say that it must have a `SlotNo`.

I think it would be slightly better if we turn this into `(r -> DiskSnapshot)`. This will also make it easier to later automatically (vs. manually by the user) give a suffix to snapshots. This also simplifies `mkIfAvailable` (I think it can go): if the `DiskSnapshot` for `tip` is in the list of snapshots, don't create one. When a named snapshot with the same slot number already exists, so be it, we'll later trim the one we created away anyway.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -276,24 +276,31 @@ initStartingWith tracer conf streamAPI initDb = do
...
+      Nothing -> return (Nothing, csTip oldest)
+      Just ss -> do
+        writeSnapshot hasFS encLedger encRef ss oldest
+        traceWith tracer $ TookSnapshot ss (csTip oldest)
+        return (Just ss, csTip oldest)
   where
     oldest :: ChainSummary l r
     oldest = ledgerDbAnchor db
 
+    tip = csTip oldest
```
mrBliss, line 302: BTW, I see 3 other uses of `csTip oldest` in the function above.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -353,24 +365,43 @@ writeSnapshot hasFS encLedger encRef ss cs = do
     encode :: ChainSummary l r -> Encoding
     encode = encodeChainSummary encLedger encRef
 
--- | Delete snapshot from disk
-deleteSnapshot :: HasCallStack => HasFS m h -> DiskSnapshot -> m ()
-deleteSnapshot HasFS{..} = removeFile . snapshotToPath
+-- | Delete snapshot from disk, unless it has a suffix
+deleteSnapshot :: (HasCallStack, Monad m) => HasFS m h -> DiskSnapshot -> m ()
+deleteSnapshot HasFS{..} ds@DiskSnapshot {..} =
+    when (isNothing dsSuffix) $ removeFile $ snapshotToPath ds
```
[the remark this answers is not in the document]
kderme: We also call it when we fail to use a snapshot at `tryNewestFirst`. Do we still want to delete it in this case case, even though it has a suffix?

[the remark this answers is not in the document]
mrBliss: I see. I don't think we should delete it in that case. I suppose we can do a similar check in `tryNewestFirst`.

It *is* important to do the check in `trimSnapshots`, because I think it can change which snapshots get deleted.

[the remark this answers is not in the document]
kderme: > It is important to do the check in trimSnapshots, because I think it can change which snapshots get deleted.

I don't understand this part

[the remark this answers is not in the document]
mrBliss: For example, say you want to keep 2 *non-named* snapshots on disk. You already have two named snapshots that are newer than all non-named snapshot. The LedgerDB just created its third non-named snapshot 140 and now trims it snapshots:
```
1000_shelley
200_favourite_snapshot
140
120
100
```
In `trimSnapshots`, you'll find there are 5 snapshots on disk, now you  drop the first two and try to delete the remaining three: `140`, `120`, and `100`. You end up without non-named snapshots instead of 2.

If you filter first in `trimSnapshots` on "non-named", you'll drop `140` and `120` from the list and delete `100`. Now you end up with 2 non-named snapshots.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -353,24 +365,43 @@ writeSnapshot hasFS encLedger encRef ss cs = do
...
 
 -- | List on-disk snapshots, most recent first
 listSnapshots :: Monad m => HasFS m h -> m [DiskSnapshot]
 listSnapshots HasFS{..} =
     aux <$> listDirectory (mkFsPath [])
   where
     aux :: Set String -> [DiskSnapshot]
-    aux = List.sortBy (flip compare) . mapMaybe snapshotFromPath . Set.toList
+    aux = List.sortBy (flip compare) . mapMaybe snapshotFromPath  . Set.toList
```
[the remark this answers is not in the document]
kderme: I noticed that this changed a bit which snapshots are deleted. ie if we have
100
100_a
200

previously it would compare based on the Ord instance of `DiskSnapshot`  (not just the slot) and it would delete the 100. now since they have the same slot, it's undeterministic (on my manual tests it deletes nothing).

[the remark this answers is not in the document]
mrBliss: We can change the `Ord` instance to prefer a named snapshot over an unnamed one with the same slot number (and document it!).

For deleting, we should filter out the named ones first as I said elsewhere, so it shouldn't play a role there.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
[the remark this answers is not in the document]
kderme: So we want `onDiskNumSnapshots` to indicate the number of  non-named snapshots? That's not how this pr works

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -274,56 +278,88 @@ initStartingWith tracer conf streamAPI initDb = do
 -- testing purposes, 'takeSnapshot' returns the block reference corresponding
 -- to the snapshot that we wrote.
 --
--- NOTE: This is a lower-level API that unconditionally takes a snapshot
--- (i.e., independent from whether this snapshot corresponds to a state that
--- is more than @k@ back).
+-- NOTE: This is a lower-level API that takes a snapshot independent from
+-- whether this snapshot corresponds to a state that is more than @k@ back. It
+-- doesn't take a snapshot if a snapshot at the same slot already exists on
+-- disk or the tip is the genesis.
```
mrBliss, line 284:
```suggestion
-- disk or if the tip is at genesis.
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -274,56 +278,88 @@ initStartingWith tracer conf streamAPI initDb = do
...
-    writeSnapshot hasFS encLedger encRef ss oldest
-    traceWith tracer $ TookSnapshot ss (csTip oldest)
-    return (ss, csTip oldest)
+             -> (r -> DiskSnapshot)
+             -> LedgerDB l r
+             -> m (Maybe DiskSnapshot, WithOrigin r)
+takeSnapshot tracer hasFS encLedger encRef toSnap db = case tip of
+    Origin      -> return (Nothing, tip)
+    NotOrigin r -> do
+      let snapShot = toSnap r
```
mrBliss, line 303: For consistency:
```suggestion
      let snapshot = toSnap r
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -274,56 +278,88 @@ initStartingWith tracer conf streamAPI initDb = do
...
 trimSnapshots :: Monad m
               => Tracer m (TraceEvent r)
               -> HasFS m h
               -> DiskPolicy
-              -> m [DiskSnapshot]
+              -> m ()
 trimSnapshots tracer hasFS DiskPolicy{..} = do
     snapshots <- listSnapshots hasFS
+    -- We ignore named snaspshots
+    let snapshots' = filter (isNothing . dsSuffix) snapshots
```
mrBliss, line 328:
```suggestion
    -- We ignore named snapshots
    snapshots <- filter (isNothing . dsSuffix) <$> listSnapshots hasFS
```
Now we can't accidentally refer to `snapshots` instead of `snapshots'`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -274,56 +278,88 @@ initStartingWith tracer conf streamAPI initDb = do
...
---
--- The deleted snapshots are returned.
+-- snapshots are stored on disk. The oldest snapshots are deleted. Snapshots
+-- with a suffix are not deleted.
 trimSnapshots :: Monad m
               => Tracer m (TraceEvent r)
               -> HasFS m h
               -> DiskPolicy
-              -> m [DiskSnapshot]
+              -> m ()
```
mrBliss, line 324: What's the reason for this change?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -357,20 +393,40 @@ writeSnapshot hasFS encLedger encRef ss cs = do
...
+    pref = show $ unSlotNo dsSlotNo
+    suff = case dsSuffix of
+      Nothing  -> ""
+      Just str -> "_" ++ str
 
 snapshotFromPath :: String -> Maybe DiskSnapshot
-snapshotFromPath = fmap DiskSnapshot . readMaybe
+snapshotFromPath fileName = do
+    slot <- readSlot pref
+    Just $ DiskSnapshot slot fixedSuff
```
mrBliss, line 419: Embrace the monad :grin: 
```suggestion
    return $ DiskSnapshot slot fixedSuff
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
kderme: It wasn't used anywhere

mrBliss:
I have rebased and changed this PR in #2733.

PR #2578, opened 2020-09-03
mock: equalStakeDist should calculate relative stake instead of absolute

### LINKED ISSUE #2577: Mock Praos implementation leads too often
I just ran a small experiment. Though it was built on a chain of several experimental commits, I'm pretty sure they were unrelated.

The experiment ran the Mock Praos ThreadNet tests 1000 times with a many variables fixed (numSlots=300 numNodes=2 k=2 f=0.2 etc). I plotted the resulting histogram of number of active slots (counted as any slot in which any node forged a block). The histogram looked to my eye to have a center located much closer to 0.350 than to the specified 0.200.

This Issue is to double-check this more carefully. (I repeated the experiment more carefully on the RealTPraos tests and the histogram seemed correctly centered on 0.200.)

Follow-up: I set f=0.5 and observed a center closer to 0.750 than to the specified 0.500.

nfrisby:
Fixes #2577.

Is this the appropriate fix? I'm not certain if relative stake is always the intended semantics of a `StakeDist`.

### DIFF
--- a/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Stake.hs
+++ b/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Stake.hs
@@ -46,6 +46,8 @@ data StakeHolder =
 -------------------------------------------------------------------------------}
 
 -- | In the mock setup, only core nodes have stake
+--
+-- INVARIANT: The rationals should sum to 1.
 newtype StakeDist = StakeDist { stakeDistToMap :: Map CoreNodeId Rational }
   deriving (Show, Eq, Serialise, NoUnexpectedThunks)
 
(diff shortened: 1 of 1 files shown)


ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Stake.hs
```diff
@@ -78,14 +80,17 @@ totalStakes addrDist = foldl f Map.empty
...
+    Map.fromList $
+    mapMaybe (nodeStake . snd) $
+    Map.toList ad
   where
     nodeStake :: NodeId -> Maybe (CoreNodeId, Rational)
     nodeStake (RelayId _) = Nothing
-    nodeStake (CoreId i)  = Just (i, 1)
+    nodeStake (CoreId i)  = Just (i, recip (fromIntegral n))
+
+    n = sum $ map (length . nodeStake) $ Map.elems ad
```
nfrisby, line 93: Is `sum . map length` instead of `length . filter isJust`  too cheeky?

nfrisby:
The corresponding test for this requires statistical testing, which is a separate/broader WIP.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Stake.hs
erikd: I think so. `length . filter isJust` is compiled validated documentation.

mrBliss (approved):
The [spec](https://hydra.iohk.io/build/3926827/download/1/ledger-spec.pdf) (section 17 on page 107) says:

> σ ∈ [0, 1] Stake proportion

So this PR makes sense. If this is enough to fix #2577, perfect.

nfrisby:
I rebased to latest tip and also removed the correct-but-unintentional mutually recursive.

PR #2579, opened 2020-09-03
Fix bug in runQuery: evaluate all Exprs against the same era

mrBliss:
Fixes https://github.com/input-output-hk/cardano-db-sync/issues/276

In #2465, we changed the implementation of `runQuery` to try to evaluate each
`Expr` in the `Query` against any era, accepting the result of the first era in
which it succeeds. This means that different `Expr`s within the same `Query` can
be evaluated in different eras.

Fix it by trying to evaluate each `Expr` in the `Query` against the same era,
trying the next era as soon as one `Expr` fails to be evaluated.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Qry.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Qry.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE RankNTypes                 #-}
 {-# LANGUAGE RecordWildCards            #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Qry.hs
```diff
@@ -533,7 +545,7 @@ instance Show (ClosedExpr a) where
     where
       go :: Int  -- How many variables are already in scope?
          -> Int  -- Precedence
-         -> Expr Var a -> ShowS
+         -> Expr Var b -> ShowS
```
mrBliss, line 552: Needed because I enabled `ScopedTypeVariables` for `runQuery`. The `a` in this `go` is *not* the `a` from `Show (ClosedExpr a)`.

erikd (approved):
It has tests so its probably right,

erikd:
bors merge

PR #2582, opened 2020-09-03
test: show mock chain tips instead of entire chains in failures

nfrisby:
Some of our failures now include hundreds of blocks. 'show' for hundreds of
blocks for each of 2-5 nodes creates hundreds of megabytes of output, most of
which is entirely useless. That much output makes the logs difficult to
interpret and unnecessarily bloats the artifacts from the nightly build run.

If that info is actually required for a debugging session, the unlucky dev can
revert this commit. (Though I would recommend projecting out the relevant parts
of the blocks and/or subarrays of the chain instead of dumping the whole
thing.)

### DIFF
--- a/ouroboros-consensus-test/src/Test/ThreadNet/General.hs
+++ b/ouroboros-consensus-test/src/Test/ThreadNet/General.hs
@@ -408,6 +408,7 @@ noExpectedCannotForges _ _ _ = False
 prop_general ::
   forall blk.
      ( Condense blk
+     , Condense (HeaderHash blk)
      , Eq blk
      , RunNode blk
      )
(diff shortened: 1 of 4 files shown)


mrBliss (approved):
LGTM. I know the pain. Emacs really can't handle long lines, so whenever this happened, my Emacs hangs :roll_eyes:.

nfrisby:
Yep, that's exactly my most constant pain point with this as well.

nfrisby:
That bors failure seems like a hiccup and also unrelated to this diff. Retrying.

nfrisby:
bors retry

mrBliss:
Hydra is being clever and reused the previous result :grin:. I typically ask @nc6 to let Hydra rerun the job in question.

PR #2585, opened 2020-09-04
Trace the success cases of block forging

mrBliss:
If another chain freeze event (CAD-1747) ever reoccurs, this will help with
find out where in the forge process it went wrong.

Reorder the constructors of `TraceForgeEvent` so that they are in the same order
as the flowchart in the docstring. Indent them correctly.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -308,6 +308,8 @@ forkBlockForging maxTxCapacityOverride IS{..} blockForging =
               trace failure
               exitEarly
 
+        trace $ TraceBlockContext currentSlot bcBlockNo bcPrevPoint
+
         -- Get ledger state corresponding to bcPrevPoint
         --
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -303,11 +304,14 @@ forkBlockForging maxTxCapacityOverride IS{..} blockForging =
             mkCurrentBlockContext currentSlot
                 <$> ChainDB.getCurrentChain chainDB
           case eBlkCtx of
-            Right blkCtx -> return blkCtx
+            Right blkCtx -> do
+              return blkCtx
```
deepfire, line 308: Is the idea here to improve visual consistency?

deepfire (approved):
LGTM!

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
mrBliss: Oh, that's an accident, I first added the trace message in this case. I'll revert.

mrBliss:
bors merge

PR #2601, opened 2020-09-11
Update dependency on ledger-specs, remove overlay schedule

mrBliss:
Updates after:
<https://github.com/input-output-hk/cardano-ledger-specs/pull/1849> which breaks
the ledger snapshot compatibility.


mrBliss:
Draft because https://github.com/input-output-hk/cardano-ledger-specs/pull/1849 hasn't merged yet.

ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
```diff
@@ -630,26 +629,28 @@ prop_simple_cardano_convergence TestSetup
...
+          -> EpochNo
+          -> Set SlotNo
+        overlaySlotsInEpoch ei d epoch = runIdentity $ do
+            epochSize <- epochInfoSize  ei epoch
+            firstSlot <- epochInfoFirst ei epoch
+            let slotsInEpoch =
+                   [firstSlot .. Util.addSlots (unEpochSize epochSize) firstSlot]
+            return
+              $ Set.fromList
+              $ filter (SL.isOverlaySlot firstSlot d) slotsInEpoch
```
mrBliss, line 653: I have asked that ledger-specs provides a function like this (not a `Set`, but a list). This is just a very naive (linear!) implementation.

@nfrisby This is just used as a label, not a check. So what if I got it wrong?

JaredCorduan: > I have asked that ledger-specs provides a function like this (not a `Set`, but a list). This is just a very naive (linear!) implementation.

That's what I was expecting to add to ledger-specs, is it too naive? ie just filtering the epoch slots with `isOverlaySlot`.

Alternatively, you can invert `isOverlaySlot` and compute them directly. which would be `floor (x / d)`. I thought since it was testing it wouldn't matter too much, and having `isOverlaySlot` as the single source of truth would be nice.

mrBliss: Since this is just testing, either way is fine. As long as `overlaySlotsInEpoch` (feel free to pick another name) is defined in ledger-specs and not here. If it's a naive linear filter, then there should be a warning in the docstring so that nobody would ever consider using it for non-testing code.

nfrisby: > So what if I got it wrong?

To validate, I'd run the test with `--iohk-enable-nightly-tests` and with 200 iterations (~30hr on a CI box) or more(depending on how much time you have. As long as you see `True` and `False` for `label_hadActiveNonOverlaySlots`, then it's behaving as expected. (This test is ultimately statistical, so I left it as a manual and occasional check for now, as we work out our statistical testing plan.)

ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
```diff
@@ -630,26 +629,28 @@ prop_simple_cardano_convergence TestSetup
         overlayOffsets i =
             -- Shift to account for the initial Byron era.
             Set.mapMonotonic (Util.addSlots numByronSlots) $
-
```
nfrisby, line 633: While you're in here, would you fix the typo on line 628/627?

```
significantly +more+ involved
```

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
```diff
@@ -412,7 +412,11 @@ instance TPraosCrypto era => ConsensusProtocol (TPraos era) where
               -> Nothing
     where
       chainState = tickedTPraosStateChainDepState cs
-      lv         = getTickedPraosLedgerView (tickedTPraosStateLedgerView cs)
+      lv         = getTickedPraosLedgerView $ tickedTPraosStateLedgerView cs
+      d          = SL._d $ SL.lvProtParams lv
+      asc        = tpraosLeaderF $ tpraosParams cfg
+      firstSlot  = firstSlotOfEpochOfSlot (tpraosEpochInfo cfg) slot
+      gkeys      = Map.keys dlgMap
```
nfrisby, line 419: Does the order of the `gkeys` list matter? Is _ascending_ correct for this use in `cardano-ledger-specs:...lookupIn...`?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/Util.hs
```diff
@@ -28,3 +29,12 @@ isNewEpoch ei newSlot referenceWO = runIdentity $ do
     reference = fromWithOrigin genesisSlotNo referenceWO
     -- TODO
     genesisSlotNo = SlotNo 0
+
+-- | Return the first slot in the epoch of the given slot.
+firstSlotOfEpochOfSlot ::
```
nfrisby, line 34: I anticipate that this could fit in a more general `Util` module, if you're interested in searching for other occurrences of this pattern. I'm sure I've written it multiple times, but I'm not sure those were merged/have survived.

nfrisby (commented):
LGTM, with some minor comments. I have not Approved because of the Draft status.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
JaredCorduan: I don't have the full context here, but the ordering on the `gkeys` does matter for the overlay schedule. `cardano-ledger-specs` is just using the `Ord` instance on key hashes.

mrBliss: I based myself on the following line: https://github.com/input-output-hk/cardano-ledger-specs/pull/1849/files#diff-3e7e2e6fbe599db2044bd7f16e5bf9a9R264
But indeed, good point. @JaredCorduan it would be good if `lookupInOverlaySchedule` took a set of them, instead of a list, that would solve the issue.

JaredCorduan: @mrBliss is it okay if I still lookup elements by index? Such as using `Data.Set.elemAt`? (I would check the size first, though, since `elemAt` is partial.)  the old overlay schedule did indeed depend on this ordering.

JaredCorduan: @mrBliss I am going to merge https://github.com/input-output-hk/cardano-ledger-specs/pull/1849 now, just since it is a bit big and it would be nice to have merged. But I am happy to do a quick follow up PR to make `lookupInOverlaySchedule` take a set of genesis keys once I understand the issue!

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/Util.hs
nc6: Could add it to `cardano-slotting`

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
nc6: I've raised https://github.com/input-output-hk/cardano-ledger-specs/issues/1856 for this, but would like to get this PR merged first in order to unblock changes in `cardano-node` for the release.

nc6 (approved):

Issue #2471, opened 2020-07-27
Implement singleEraTransition for Shelley


mrBliss:
Related to #2452.

mrBliss:
For documentation of this update process, see https://github.com/input-output-hk/ouroboros-network/pull/1741/

edsko:
In order to avoid the same problem that we had with Byron (#2455), we could take the following approach:

* Blocks that came in within an epoch after the `4k/f` voting deadline are not relevant.
* Since there are gaps between blocks, we are probably only sure that there will be no more relevant block when we have seen the first block after the deadline.
* If we _count_ how many blocks we have seen post deadline, and we have reached `k` of them, we know that that last pre-* deadline block won't be rolled back anymore.
* At this point we can look at the ledger state and see which proposals we accepted in the voting period, if any, and notify the HFC is one of them indicates a transition.

Issue #2455, opened 2020-07-22
Byron should use block no, not slot no, to decide whether or not to report transition to HFC


edsko:

edsko:
We decided that we will try to clean this up _after_ the hard fork, and accept the risk of the non-monotonicity of the time info for now.

edsko:
When we do address this, we should add a property (both for Byron and for Shelley) that the transition point reported to the HFC matches the point of the adoption of the update proposal in the ledger. It doesn't _technically_ matter, especially if the HFC takes over and does the transition _before_ the update proposal would have been accepted, but the other direction _is_ more troublesome (where, say, we'd have a bit of Byron at version 2.0.0 and then suddenly Shelley), and either way if it doesn't line up it's just confusing.

edsko:
We also have to be careful of course that when we do change it, we don't inadvertently introduce a hard fork by interpreting the chain differently. The events on mainnet, as far as protocol updates go, are listed at https://gist.github.com/edsko/75fc60669676dd498b519c77f08d0dc4 .

edsko:
A summary of the relevant events is

```
Slot     Event
--------------------------------------------------------------------------------
322876   Update to 0.1.0 confirmed
327196   Update to 0.1.0 stably confirmed
327617   Update to 0.1.0 endorsements: 1
327720   Update to 0.1.0 endorsements: 2
327724   Update to 0.1.0 endorsements: 3
327727   Update to 0.1.0 candidate
332047   Update to 0.1.0 stable candidate, scheduled for epoch 16.
345600   Update to 0.1.0 complete [note: first slot of epoch 16]

1784895  Update to 0.2.0 confirmed 
1789215  Update to 0.2.0 stably confirmed, endorsements: 1
1789216  Update to 0.2.0 endorsements: 2
1789217  Update to 0.2.0 endorsements: 3
1789219  Update to 0.2.0 candidate
1793539  Update to 0.2.0 stable candidate, scheduled for epoch 84
1814400  Update to 0.2.0 complete [note: first slot of epoch 84]

3781110  Update to 1.0.0 confirmed 
3785430  Update to 1.0.0 stably confirmed
3786431  Update to 1.0.0 endorsements: 1
3786460  Update to 1.0.0 endorsements: 2
3786462  Update to 1.0.0 endorsements: 3
3786463  Update to 1.0.0 candidate
3790783  Update to 1.0.0 stable candidate, scheduled for epoch 176
[...]

PR #2604, opened 2020-09-15
Add BlockNo and transition info to Byron/Shelley ledger state

edsko:
This consists of 4 separate steps, each hopefully easily reviewable commits:

1. Add BlockNo to the Byron ledger state and update it when we apply a block
2. Rename the fields of the Shelley ledger state to be more consistent with
   the Byron ledger, and also pave the way for the next commit. This 
   commit makes no semantic changes.
3. Add BlockNo to the Shelley ledger state, updating it again on block 
   application.
4. Finally, add a field to both the Byron and the Shelley ledger that will 
   record some additional information about the transition (critically, it 
   will include the `BlockNo` that we need to be stable before reporting
   anything to the HFC). This field is mostly a placeholder right now, but since
   'transition not yet known' is a valid value _anyway_, that is okay: we 
   can fill in the details later and not break any snapshots.

This paves the way for #2455 and #2471; since it breaks the snapshots, it 
would be good to include this PR now that we're breaking them _anyway_, to 
avoid having to do that again in the next release.


dcoutts (approved):
LGTM though I don't totally get the final patch. At least I can see it's not _doing_ anything.

edsko:
bors merge

PR #2616, opened 2020-09-18
Update index state

newhoggy:


newhoggy:
I'm looking for some guidance on how to fix this compiler error:

```
src/Test/ThreadNet/TxGen/Shelley.hs:88:53: error:
    Not in scope: ‘Gen.tryGenTx’
    Neither ‘Test.Shelley.Spec.Ledger.Generator.Constants’,
            ‘Test.Shelley.Spec.Ledger.Generator.Core’ nor ‘Test.Shelley.Spec.Ledger.Generator.Utxo’ exports ‘tryGenTx’.
   |
88 |     either (const Nothing) (Just . mkShelleyTx) <$> Gen.tryGenTx
   |                                                     ^^^^^^^^^^^^

src/Test/ThreadNet/TxGen/Shelley.hs:132:13: error:
    Not in scope: ‘Gen.genTxRetries’
    Neither ‘Test.Shelley.Spec.Ledger.Generator.Constants’,
            ‘Test.Shelley.Spec.Ledger.Generator.Core’ nor ‘Test.Shelley.Spec.Ledger.Generator.Utxo’ exports ‘genTxRetries’.
    |
132 |           , Gen.genTxRetries     = 1000
    |             ^^^^^^^^^^^^^^^^
```

ouroboros-consensus-cardano-test/src/Test/ThreadNet/TxGen/Cardano.hs
```diff
@@ -11,6 +11,7 @@ module Test.ThreadNet.TxGen.Cardano (
     CardanoTxGenExtra (..),
   ) where
 
+import           Cardano.Ledger.Val ((<->))
```
mrBliss, line 14: Can you move this down to where it used to be? Even though it doesn't match the `Shelley.Spec.Ledger` namespace, it is defined in the same package, and we *try* (but often fail) to group imports per IOHK package (not for general hackage packages).

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
```diff
@@ -27,6 +27,7 @@ module Ouroboros.Consensus.Shelley.Node (
   , validateGenesis
   ) where
 
+import           Cardano.Ledger.Val ((<->))
```
mrBliss, line 30: Same

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
```diff
@@ -378,7 +380,7 @@ instance TPraosCrypto era => ConsensusProtocol (TPraos era) where
 
   checkIsLeader cfg TPraosCanBeLeader{..} slot cs = do
       -- First, check whether we're in the overlay schedule
-      case SL.lookupInOverlaySchedule firstSlot gkeys d asc slot of
+      case SL.lookupInOverlaySchedule firstSlot (S.fromList gkeys) d asc slot of
```
mrBliss, line 383: Let's undo this change and change L421 to:
```haskell
      gkeys      = Map.keysSet dlgMap
```

mrBliss:
I'll take over this PR.

mrBliss (approved):

network-mux/network-mux.cabal
```diff
@@ -33,7 +33,7 @@ library
                        binary          >=0.8 && <0.11,
                        bytestring      >=0.10 && <0.11,
                        containers      >=0.5 && <0.7,
-                       network         >=3.1 && <3.2,
+                       network         >=3.1 && <3.1.2,
```
mrBliss, line 36: @newhoggy What's the reason for this change? Does the new 3.1.2 not work somehow? It seems to compile just fine for me.

mrBliss:
FYI, I split off the ledger update in #2644 and merged it so that Duncan can tag the release. I rebased this branch on master and now it only contains the index update (and updates of the other dependencies).

network-mux/network-mux.cabal
newhoggy: It doesn't build on Windows.

See https://github.com/haskell/network/issues/484

newhoggy:
> I'll take over this PR.

Oops, I force pushed over your PR.  I forced pushed it back.

newhoggy:
This looks good.  Let's merge.

network-mux/network-mux.cabal
mrBliss: I see. We have many dependencies on the `network` package. I'm wondering whether it would make more sense to add a constraint to `cabal.project` instead of changing the constraint in *one* `.cabal` file. @coot Your opinion?

mrBliss:
@dcoutts Says he can't seem to build with the index update. Let's wait a little bit until the node is tagged before we merge this.

network-mux/network-mux.cabal
mrBliss: Actually, you're right, the upper bound should be here, because it is indeed only `network-mux` that relies on the function that went missing. And if we put the constraint in `cabal.project`, then all downstream repos should add that same constraint to their `cabal.project`, instead of automatically inheriting it through the `network-mux.cabal`.

mrBliss:
1.21 has been tagged and I need this in order to depend on the recently released `nothunks` library, so I'm merging it.

mrBliss:
bors merge

coot (approved):
LGTM

PR #2618, opened 2020-09-18
Export constructors for TraceKeepAliveClient

karknu:


coot:
I think they should also be exported from `Ouroboros.Network.NodeToNode` module.

karknu:
> I think they should also be exported from `Ouroboros.Network.NodeToNode` module.

Why?
`Ouroboros.Network.NodeToNode` doesn't currently import `Ouroboros.Network.KeepAlive` and other "application level" tracers are not exported such as TraceTxSubmissionInbound, TraceTxSubmissionOutBound and blockfetch.

coot:
ok, let's leave this as is.

coot (approved):
LGTM

PR #2624, opened 2020-09-21
BlockFetch protocol limit

coot:
The block fetch limit is set according to:  https://docs.google.com/spreadsheets/d/1W0PJCYPan8smSg_CnuG-fQVFOnb7DzQhm3W1LuVTZR4/edit?pli=1#gid=0

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
@@ -285,24 +285,22 @@ nodeToNodeProtocols MiniProtocolParameters {
         -- block-fetch client can pipeline at most 'blockFetchPipeliningMax'
         -- blocks (currently '10').  This is currently hard coded in
         -- 'Ouroboros.Network.BlockFetch.blockFetchLogic' (where
-        -- @maxInFlightReqsPerPeer = 10@ is specified).  In the future the
+        -- @maxInFlightReqsPerPeer = 100@ is specified).  In the future the
         -- block fetch client will count bytes rather than blocks.  By far
         -- the largest (and the only pipelined message) in 'block-fetch'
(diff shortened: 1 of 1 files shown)


ouroboros-network/src/Ouroboros/Network/NodeToNode.hs
```diff
@@ -285,24 +285,22 @@ nodeToNodeProtocols MiniProtocolParameters {
...
         --
-        -- So the overall limit is `10 * 2_097_154 = 20_971_540` (i.e. aroudn
-        -- '20Mb'), we add 10% safety margin:
+        -- In the byron era this limit was set to `10 * 2MB`, we keep the more
+        -- relaxed limit here.
         --
-        maximumIngressQueue = addSafetyMargin $
-          fromIntegral blockFetchPipeliningMax * 2_097_154
+        maximumIngressQueue = addSafetyMargin $ fromIntegral $
+          max 20_000_000 (blockFetchPipeliningMax * 64_000)
```
karknu, line 303: The limit is 0xffff, not 64000 bytes.

coot: Right, I'll fix this.

semanticphilosopher (commented):
I have a concern as to how this may effect bulk download. Correct me if I'm wrong, but this represents the total _data_ in flight during a bulk block fetch; I assume that we are ack-ing the blocks (probably implicitly by asking for more).  This makes it analogous to the role of the TCP congestion window (from a performance perspective).  My concern is will 6.4Mb of inflight data be sufficient to hide latency to achieve reasonable throughput for peers that are "reasonably" distant (i.e not in same data centre, but on same continent).

coot:
bors merge

njd42 (approved):
lgtm

coot:
bors merge

PR #2625, opened 2020-09-21
Fixes to the networking spec

AndrewWestberg:
Updating the timestamp section of the mux protocol based on telegram comments from @karknu . 
Updating the 3.2 table to include Transaction Submission protocol for NtN communication. The paragraph above it describes it and I've seen those messages in a NtN wireshark capture.
Also fixing some typographical issues.

### DIFF
--- a/docs/network-spec/mux.tex
+++ b/docs/network-spec/mux.tex
@@ -54,7 +54,7 @@ Table~\ref{segment-header} shows the layout of the data segments of the multiple
 The segment header contains the following data:
 \begin{description}
 \item[Transmission Time]
-  The transmission time is a time stamp based the wall clock of the peer with a
+  The transmission time is a time stamp based the lower 32 bits of the sender's monotonic clock with a
   resolution of one microsecond.
 \item[Mini Protocol ID] The unique ID of the mini protocol as in Table~\ref{mini-protocol-id}.
 \item[Payload Length] The payload length is the size of the segment payload in Bytes.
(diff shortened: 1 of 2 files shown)


docs/network-spec/mux.tex
```diff
@@ -180,7 +180,7 @@ \section{Life Cycle of a Connection}
   1  & DeltaQ                                & Yes  & Yes \\ \hline
   2  & Chain-Sync instantiated to headers    & Yes  & No \\ \hline
   3  & Block-Fetch                           & Yes  & No  \\ \hline
-  4  & Transaction-Submission                & No   & Yes  \\ \hline
+  4  & Transaction-Submission                & Yes  & Yes  \\ \hline
```
coot, line 183: We have two separate `mini-protocols` for transaction submission:
* `LocalTxSubmission` - which is only used for `node-to-client`
* `TxSubmission` - only used by the `node-to-node` protocol.

They are quite a bit different, thus I suggest we have two entries in this table.

coot (changes requested):

docs/network-spec/mux.tex
AndrewWestberg: Do these share the same protocol id?

coot: `TxSubmission` has id `4`
`LocaTxSubmission` has id `6`

AndrewWestberg: Fixed

coot (approved):
LGTM

coot:
bors merge

PR #2629, opened 2020-09-22
Mux restart tests

karknu:


karknu:
Implements #2607

network-mux/test/Test/Mux.hs
```diff
@@ -1219,9 +1222,62 @@ dummyAppToChannel DummyApp {daAction, daRunTime} = \_ -> do
...
+
+dummyRestartingAppToChannel :: forall a m.
+                     ( MonadAsync m
+                     , MonadCatch m
+                     , MonadTimer m
+                     )
+                  => (DummyApp, a)
+                  -> (Channel m -> m ((DummyApp, a), Maybe BL.ByteString))
+dummyRestartingAppToChannel (app, r) = \_ -> do
+    threadDelay $ daRunTime app
```
coot, line 1255: We're not using the channel, shouldn't we run some simple protocol, e.g. `req-response`.

I think It would be interesting to see that if there are left overs bytes the mux will well behave when restarting a mini-protocol.

karknu: Since we never read any data there is always a left over byte in the channel.
Anyway a simple req-rsp protocol would be a more realistic test.

coot: There's `Test.Mux.ReqResp` which does not even rely on `typed-protocols`.

network-mux/test/Test/Mux.hs
```diff
@@ -1219,9 +1222,62 @@ dummyAppToChannel DummyApp {daAction, daRunTime} = \_ -> do
...
+              )
+            => MuxBearer m
+            -> DummyApp
+            -> m ()
+triggerApp bearer app = do
+    let chan = muxBearerAsChannel bearer (daNum app) InitiatorDir
+    traceWith verboseTracer $ "app waiting " ++ (show $ daNum app)
+    threadDelay (daStartAfter app)
+    traceWith verboseTracer $ "app starting " ++ (show $ daNum app)
+    send chan $ BL.singleton 0xa5
```
coot, line 1278: why just a single byte, not the whole conversation?

network-mux/test/Test/Mux.hs
```diff
@@ -1270,6 +1317,152 @@ prop_mux_start_mX apps runTime = do
...
+                      -> m Bool
+    runRestartingApps _ ops | M.null ops = return True
+    runRestartingApps mux ops = do
+        appResult <- atomically $ foldr (<|>) retry $ M.elems ops
+        case appResult of
+             Left _ -> return False
+             Right (app, 0) -> do
+                 runRestartingApps mux $ M.delete (daNum app) ops
+             Right (app, restarts) -> do
+                 op <- runMiniProtocol mux (daNum app) ResponderDirectionOnly StartOnDemand
```
coot, line 1404: If we add a delay before call to `runMiniProtocol` we can check how restarting works if there are outstanding bytes in the ingress queue.

network-mux/test/Test/Mux.hs
coot: I see, it is not using `Test.Mux.ReqResp` protocol, which I hoped for.  When we are using `dummyAppToChannel` we never ready any bytes from mux ingress queue.  If we use a proper protocol the tests would be more comprehensive.

Another testing strategy that could be useful is using `IOSim` and tracing to ensure that the mux is in the right state on restarts. We do that for other components: block-fetch, connection-manager, peer-to-peer governor.

network-mux/test/Test/Mux.hs
karknu: The case with left over bytes in the queues is covered by the testcases `prop_mux_snd_recv` and `prop_mux_snd_recv_bi` using the reqresp protocol. Complicating these tests by adding a reqresp protocol would not increase test coverage.

network-mux/test/Test/Mux.hs
karknu: See https://github.com/input-output-hk/ouroboros-network/pull/2629#discussion_r500214138 . ReqResp byte tests are covered in other testcases.

network-mux/test/Test/Mux.hs
karknu: There are always bytes in the ingress queue in this test case.

network-mux/test/Test/Mux.hs
coot: What if the left over bytes are only mistreated after restarting a protocol?

karknu: `prop_mux_snd_recv` and `prop_mux_snd_recv_bi` restarts the protocol a random number of times.

coot (approved):
LGTM

Issue #2472, opened 2020-07-27
ThreadNet rewrite


mrBliss:
@nfrisby has been wanting to rewrite the ThreadNet tests, now that we'll have some time in August to address tech debt, we should consider it.

Requests from Edsko and me:

* Generate all test inputs upfront, nothing should depend on running the test. E.g., generated transactions should not depend on the ledger state at that slot, which we don't yet know upfront, etc.
* Proper shrinking, depends on the bullet above.
* Per connection version negotiation (`BlockNodeToNodeVersion`).
  * Different nodes should have different lower bounds, depends on #2364.
* Heterogeneous networks: different block types, e.g., `ByronBlock` vs. `CardanoBlock`, `ShelleyBlock` vs. `ShelleyHFC`.
* Run nodes the same way that real nodes run, i.e., #2032, #1390.
* Replace unit tests with labels.
* Performance. Good starting point: profile the HFC A/B test, which does barely anything and see what the low hanging fruit is.
* Structure and write the code in such a way that it is easier to understand and modify by other people in the team. Deadlines and time difference have made this kind of more “collective ownership” difficult, but we should attempt to improve this situation.

@nfrisby Feel free to add to/modify this list in any way you see fit.

PR #2635, opened 2020-09-24
Avoid filling the ThreadNet mempools with random transactions

### LINKED ISSUE #2581: Workaround the crucial tx + full mempool + blocked forge scenario
This error was first seen in https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/220#6223fd68-d16b-4dc4-b4ad-3f652711e36a -- it's the `insufficient growth` error in RealTPraos. (The other is Issue 2506 again, whose fix is currently under review.)

PR #2543 changed the timing of the ledger access, which made block forging "faster", which meant that a node's crucial tx for a slot it leads would not necessarily be included in the block it forged. So PR 2543 also introduce explicit synchronization so that the node won't forge until the crucial transaction had entered the mempool.

This new failure is due to the mempool becoming full, so the `addTxs` for the crucial transaction blocks for multiple slots, and that in turn blocks the forge. This adds a "message delay" that the sensitive ThreadNet properties currently identify as an unexpected slow down in the chain growth (eg it sees that node 1 lead slot 446, but the corresponding new block didn't arrive until slot 460.)

This Issue is Done once we prevent this scenario from causing test failures.

nfrisby:
Fixes #2581. At least as stop-gap until PR #2472.

I don't like estimating the transaction sizes and/or throttling the size of a batch of randoms. It all seems much too pessimistic.

For example, for the repro I was working with, the average tx per block fell from 10-20 to 2-3. These are random transactions, so they're not particularly important, but more seems better than less regarding code coverage.

### DIFF
--- a/ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns         #-}
 {-# LANGUAGE ConstraintKinds      #-}
 {-# LANGUAGE DataKinds            #-}
 {-# LANGUAGE DeriveAnyClass       #-}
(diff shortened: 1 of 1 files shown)


nfrisby:
The only attractive alternative we've considered is the discussion on #2581. That path would add to the Mempool API so the node can check if a sequence of hypothetical transactions would be valid or invalid regardless of the current size of the mempool. If I understand correctly, this is already loosely planned functionality. That functionality would let us mitigate #2581 because we're fairly certain that the mempool only fills up far after all the crucial transactions are relevant (in RealTPraos at least; not sure about RealPBFT, but we haven't seen errors related to full mempools there). So even if the mempool were full, we'd still be able to see immediately that the crucial tx would be rejected (eg because it's already on the chain).

I proved that out on this branch, for comparison. https://github.com/input-output-hk/ouroboros-network/tree/nfrisby/issue-2581-precheck-crucials

nfrisby:
Closed, favoring PR #2643 instead.

PR #2636, opened 2020-09-25
Fix deltaq's g calculation

karknu:


njd42 (approved):
Looks good: do we have any separate tests for the EWMA properties? It would be good to have higher degree of trust in that primitive (as we are likely to use it elsewhere).

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
```diff
@@ -136,6 +137,67 @@ instance Arbitrary NetworkDelay where
...
+    (c_aid, s_aid) <- runKeepAliveClientAndServer (NetworkDelay nd) seed verboseTracer controlMessageSTM
+                          registry clientId (KeepAliveInterval keepAliveInterval)
+    threadDelay 15 -- enough time to send one, and only one keep-alive
+    dqLive <- atomically $ readPeerGSVs registry
+
+    atomically $ writeTVar controlMessageV Terminate
+    void $ wait c_aid
+    void $ wait s_aid
+
+    -- XXX Must be larger than the KeepAliveInterval timeout or we leak threads in the SIM
```
coot, line 172: I have an impression that this would help: https://github.com/input-output-hk/ouroboros-network/pull/2631 (`registerDelay` without unsupervised threads). It might be worth trying.

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
```diff
@@ -136,6 +137,67 @@ instance Arbitrary NetworkDelay where
...
+    gCheck (GSV g _ _) = nd == g * 2
+
+-- Test that a single sample causes g estimated to be set correctly.
+prop_keepAlive_ping_pong :: NetworkDelay -> Int -> Property
+prop_keepAlive_ping_pong nd seed = do
+    let (_output, r_e) = (selectTraceEventsSay &&& traceResult True)
+                             (runSimTrace $ prop_keepAlive_ping_pongM nd seed)
+    ioProperty $ do
+        -- printf "new testcase %s\n" (show nd)
+        -- mapM_ (printf "%s\n") _output
```
coot, line 195: ^ left overs from debugging?

coot (changes requested):
Using `IOSim` and traces we gain access to internal state changes.  This way we can verify that at each step the computation is in bounds (or any other useful properties), rather than just when the `keep-alive` terminates.

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
karknu: It is intentionally left there.

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
karknu: If the PR gets merged before this one, I'll rebase and try it.

karknu:
> Using `IOSim` and traces we gain access to internal state changes. This way we can verify that at each step the computation is in bounds (or any other useful properties), rather than just when the `keep-alive` terminates.

Great suggestion!
I've updated the testcase to do that.

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
```diff
@@ -149,99 +147,56 @@ prop_keepAlive_convergenceM
...
-    gsvCheck :: PeerGSV -> Bool
-    gsvCheck PeerGSV{outboundGSV, inboundGSV} =
-        gCheck outboundGSV && gCheck inboundGSV
-
-    gCheck :: GSV -> Bool
-    gCheck (GSV g _ _) =
-        let low = 0.95 * nd / 2
-            high = 1.05 * nd / 2 in
-        g >= low && g <= high
+    return ()
```
coot, line 172: This `return ()` is not neaded.

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
```diff
@@ -149,99 +147,56 @@ prop_keepAlive_convergenceM
...
-                       , MonadMonotonicTime m
-                       , MonadTimer m
-                       , Eq (Async m ())
-                       )
-                    => Tracer m (WithThreadAndTime a) -> Tracer m a
-threadAndTimeTracer tr = Tracer $ \s -> do
-    !now <- getMonotonicTime
-    !tid <- myThreadId
-    traceWith tr $ WithThreadAndTime now (show tid) s
+prop_keepAlive_convergence nd seed =
```
coot, line 177: Nice :tada:

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
```diff
@@ -149,99 +147,56 @@ prop_keepAlive_convergenceM
...
                           registry clientId (KeepAliveInterval keepAliveInterval)
     threadDelay $ timeConstant * keepAliveInterval
-    dqLive <- atomically $ readPeerGSVs registry
 
     atomically $ writeTVar controlMessageV Terminate
     void $ wait c_aid
     void $ wait s_aid
 
     -- XXX Must be larger than the KeepAliveInterval timeout or we leak threads in the SIM
     threadDelay (keepAliveInterval + 128)
```
coot, line 171: I think you only need to use this because `runPeerWithLimits` will leak a thread (on Windows and `IOSim`).  If this is the case #2631 will fix it.  It's worth adding this as a comment.

coot (approved):
LGTM, just two minor comments.

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
karknu: Will remove

ouroboros-network/test/Test/Ouroboros/Network/KeepAlive.hs
karknu: I'll add a reverence to it.

PR #2642, opened 2020-09-29
Only tick the header state when checking for leadership

mrBliss:
A block producer checks each slot whether it should produce a block. One of the
steps in this process is to tick the `ExtLedgerState`, which can be rather
expensive. For example, creating the rewards update after 4k/f slots for Shelley
takes multiple seconds.

Instead, only tick the `ChainDepState`, which is all we need to check whether
we're the leader. To do this, we need the ticked ledger view, which is exactly
what `forecastFor` was giving us.

Only if we're actually the leader, do we have to tick the actual `LedgerState`,
which involves the expensive rewards update computation, because we need it to
get transactions from the Mempool that are valid w.r.t. it, etc.

However, changes in the ledger are required because `forecastFor` is implemented
using `futureLedgerView` (from the ledger), which calls the `TICK` rule,
which *does* compute the expensive rewards update, but its result is discarded
using the `view` function (that produces a ledger view from the ledger state).
But as the ledger state type is strict in all its fields, the computation is
forced. Note that simply making the field lazy is not so nice, because we want
the state data type to be strict. Instead, a better idea solution is to split up
the `TICK` rule in two: one transition for the information needed by
`futureLedgerView` and one transition for the information that
`futureLedgerView` doesn't need, i.e., the rewards update, etc.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -58,6 +58,7 @@ import           Ouroboros.Consensus.Ledger.SupportsProtocol
 import           Ouroboros.Consensus.Mempool
 import           Ouroboros.Consensus.Node.Run
 import           Ouroboros.Consensus.Node.Tracers
+import           Ouroboros.Consensus.Protocol.Abstract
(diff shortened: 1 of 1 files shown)


edsko (approved):

mrBliss:
bors merge

PR #2643, opened 2020-09-29
ThreadNet: unblock forge when the mempool is full if the crucial txs would be invalid

### LINKED ISSUE #2581: Workaround the crucial tx + full mempool + blocked forge scenario

nfrisby:
Fixes #2581.

In our current ThreadNet tests, either the crucial transactions are only relevant "early" in the test (eg RealTPraos, Cardano, etc) or else we've never seen the mempools filling up cause a Issue #2581-like failure (eg RealPBFT).

Based on the assumption that our crucial txs are used up "early" enough, this PR fixes Issue #2581. "Later" when the mempools have become full, the nodes are still able to immediately determine that the crucial txs are invalid (since they're already on the chain, etc). This unblocks the forge soon enough to not disturb the consensus behavior in the ThreadNet.

### DIFF
--- a/ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
@@ -41,6 +41,7 @@ import           Control.Monad.Class.MonadTimer (MonadTimer)
 import qualified Control.Monad.Except as Exc
 import           Control.Tracer
 import qualified Data.ByteString.Lazy as Lazy
+import           Data.Either (isRight)
 import           Data.Functor.Identity (Identity)
 import qualified Data.List as List
 import qualified Data.List.NonEmpty as NE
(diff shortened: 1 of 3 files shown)


mrBliss (approved):
:+1:

mrBliss:
bors merge

PR #2644, opened 2020-09-29
Update dependency on ledger-specs

mrBliss:
Many Shelley imports are now redundant as `Shelley.Spec.Ledger.API` exports much
more. Unfortunately, still not everything we need. Explicitly list which imports
are missing from the API.


dcoutts (approved):
LGTM!

mrBliss:
bors merge

PR #2648, opened 2020-09-30
Make `Exactly` a newtype

edsko:
This is an internal refactoring only, but it means that Exactly can be given
sane `Functor`, `Foldable`, `Transversable`, as well as `Show` and `Eq`
instances (without any complicated `All ( .. Compose ...)` constraints). A lot
more convenient to work with.


mrBliss (approved):

edsko:
bors merge

PR #2664, opened 2020-10-05
Replace NoUnexpectedThunks with NoThunks

mrBliss:
The code in `cardano-prelude` has been released as a separate library on
Hackage: https://hackage.haskell.org/package/nothunks

### DIFF
--- a/ouroboros-consensus-byron/ouroboros-consensus-byron.cabal
+++ b/ouroboros-consensus-byron/ouroboros-consensus-byron.cabal
@@ -61,6 +61,7 @@ library
                      , formatting        >=6.3   && <6.4
                      , mtl               >=2.2   && <2.3
                      , serialise         >=0.2   && <0.3
+                     , nothunks
                      , text              >=1.2   && <1.3
 
                      , ouroboros-network
--- a/ouroboros-consensus-cardano-test/test/Main.hs
+++ b/ouroboros-consensus-cardano-test/test/Main.hs
@@ -1,5 +1,7 @@
 module Main (main) where
 
+import           Cardano.Crypto.Libsodium (sodiumInit)
+
 import           Test.Tasty
 import           Test.Util.Nightly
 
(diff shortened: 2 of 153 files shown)


mrBliss:
Still building it locally after a rebase. I will also run all tests with the nothunks check on.

ouroboros-consensus-cardano-test/test/Main.hs
```diff
@@ -9,7 +11,7 @@ import qualified Test.Consensus.Cardano.Serialisation (tests)
 import qualified Test.ThreadNet.Cardano (tests)
 
 main :: IO ()
-main = defaultMainWithIohkNightly tests
+main = sodiumInit >> defaultMainWithIohkNightly tests
```
mrBliss, line 14: For the non-test code: call `sodiumInit` here: https://github.com/input-output-hk/cardano-node/blob/512a82974fb9b5745d1152a7ab0623f02e8fe45e/cardano-cli/src/Cardano/CLI/TopHandler.hs#L69

edsko (approved):

ouroboros-consensus-byron/ouroboros-consensus-byron.cabal
```diff
@@ -61,6 +61,7 @@ library
                      , formatting        >=6.3   && <6.4
                      , mtl               >=2.2   && <2.3
                      , serialise         >=0.2   && <0.3
+                     , nothunks
```
mrBliss, line 64: FYI, we don't need bounds on `nothunks` as we are compatible with any *currently released* version (0.1.0.0 and 0.1.1.0). When ["Derive NoThunks for Void"](https://github.com/input-output-hk/nothunks/commit/6fd1cf7a3c77607ba190881b48b219dd50dc5ef7) becomes part of a new release on Hackage and our index-state is updated to a date *after* the release, *then* we'll need to remove our orphan `NoThunks Void` instance from `Ouroboros.Consensus.Util.Orphans` and set that release as the new lower bound.

mrBliss:
bors merge

PR #2667, opened 2020-10-06
db-analyser: clean up

mrBliss:
* Fix mainnet vs. testnet bug. Classic boolean blindness: in one place the
  `Bool` meant "testnet" and in another "mainnet". Fix it by using
  `RequiresNetworkMagic` instead of `Bool`.

* Pass the initial ledger to the analyses. This opens the door for analyses that
  do block and/or header validation.

* Replace `IORef`s by threading some state around.

* Parameterise `processAll` over `BlockComponent`. Now, for example,
  `showHeaderSize` no longer needs to decode the blocks, it just receives the slot
  number and the header size from the indices.

* Introduce `AnalysisEnv`.

* Drop unused dependency on `cardano-slotting`.


edsko (approved):

mrBliss:
bors merge

PR #2678, opened 2020-10-09
Fix local address

coot:
Since we are still in unidirectional mode, we should always pick the
local address using `anyIPv4Addr` or `anyIPv6Addr`.

The bug was introduced in d1a56a5aaf.


karknu (approved):
LGTM.

coot:
bors merge

PR #2679, opened 2020-10-09
Use the real Allegra and Mary era tags

### LINKED ISSUE #2668: Use real era tags for Allegra and Mary
In #2666 we add(ed) the Allegra and Mary eras to `CardanoBlock`. These are identical to the Shelley era (as if Shelley hard-forks to Shelley to Shelley):

```
type AllegraEra c = Era.Shelley c
type MaryEra    c = Era.Shelley c
```

This means we don't have to do any translations or conversions yet between Shelley-based eras.

When `cardano-ledger-specs` exports the right `AllegraEra` and `MaryEra` tags (see https://jira.iohk.io/browse/CAD-1999), we should start using them as the real era tags.

This means we'll have to do some translations (`LedgerView`, `LedgerState`, `ShelleyGenesis`, ...) for which we can use https://github.com/input-output-hk/cardano-ledger-specs/pull/1893. Note that all instances of the `TranslateEra` class that can be defined in `cardano-ledger-specs`, should be defined there. We will also have some consensus-specific ones.

We will also have to do cross-era chain selection using preferably [`SelectSameProtocol`](https://github.com/input-output-hk/ouroboros-network/blob/53a6ccbe3998764fc1a2cf42ef6fcafc60f817ec/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Protocol/ChainSel.hs#L49). Currently the protocol is parameterised over the era too, so `BlockProtocol (ShelleyBlock (ShelleyEra c) ~ BlockProtocol (ShelleyBlock (AllegraEra c))` is not true because `TPraos (ShelleyEra c) ~ TPraos (AllegraEra c)` is not true. We can use [`CustomChainSel`](https://github.com/input-output-hk/ouroboros-network/blob/53a6ccbe3998764fc1a2cf42ef6fcafc60f817ec/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Protocol/ChainSel.hs#L57) with some translations to work around it until this is fixed in the ledger itself, by not parameterising the protocol-specific types over the era, but just over the crypto (see https://jira.iohk.io/browse/CAD-2000).

mrBliss:
Fixes #2668.

Previously, the Allegra and Mary eras were exactly the same as the Shelley era,
so no translations between the eras were needed.

In this commit, we use the actual, different `AllegraEra` and `ShelleyEra` tags.
This means we need translations between the eras.


mrBliss:
Draft because this depends on https://github.com/input-output-hk/cardano-ledger-specs/pull/1902.

mrBliss:
Rebased and nearly ready. This includes #2716 until merged.

The implementation is done, but the tests are still blocked by one thing (AFAIK, unless the current compile errors are hiding other problems): the generators for the serialisation tests in the ledger only work for the Shelley era, not for the [Allegra and Mary eras](https://github.com/input-output-hk/cardano-ledger-specs/blob/32f492b27ac291f181d5194a13126660e259b1c2/shelley/chain-and-ledger/shelley-spec-ledger-test/src/Test/Shelley/Spec/Ledger/Utils.hs#L141-L142).

Because of how our generators are written (see #2709), it is PITA to try to work around this.

edsko (approved):
Just waiting for @uroboros :roll_eyes: :grimacing:

uroboros:
> Just waiting for @uroboros 🙄 😬

coming soon! ish!

ouroboros-network/test/Ouroboros/Network/NodeToNode/Version/Test.hs
```diff
@@ -18,16 +18,11 @@ tests = testGroup "Ouroboros.Network.NodeToNode.Version"
...
-
-    shrink NodeToNodeV_4 = [NodeToNodeV_3]
-    shrink NodeToNodeV_3 = [NodeToNodeV_2]
-    shrink NodeToNodeV_2 = [NodeToNodeV_1]
-    shrink NodeToNodeV_1 = []
+    arbitrary = arbitraryBoundedEnum
+
+    shrink v
+      | v == minBound = []
+      | otherwise     = [pred v]
```
mrBliss, line 25: This way it's future-proof

mrBliss:
bors merge

Issue #2680, opened 2020-10-12
Investigate failing assertion in TxSubmission.Inbound


mrBliss:
The following assertion is failed in the nightly consensus tests:
https://github.com/input-output-hk/ouroboros-network/blob/6a2b8922856460cccdc7015d33a14ebd28696d14/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs#L201-L204
For more details, see https://input-output-rnd.slack.com/archives/GCW2S19CM/p1601951773003900 and in particular https://input-output-rnd.slack.com/archives/GCW2S19CM/p1602269445020300?thread_ts=1601951773.003900&cid=GCW2S19CM

Bisecting has shown that the failure was introduced in https://github.com/input-output-hk/ouroboros-network/pull/2644/, which updates the ledger dependency. Among the changes to the ledger was an update to the transaction generator that we use in the failing consensus test. I don't think there is a bug in that generator or that the ledger is to blame, but that the change in behaviour of the generator is just triggering the bug that was there all along.

To reproduce, see https://github.com/input-output-hk/ouroboros-network/tree/nfrisby/debug-nightly

PR #2681, opened 2020-10-12
Temporarily disable Shelley's transaction generator

mrBliss:
It can be re-enabled as soon as #2680 has been resolved.

Bonus: this speeds up the Shelley ThreadNet tests, which spend most of their
time in this horribly slow transaction generator, see
https://github.com/input-output-hk/cardano-ledger-specs/issues/1901.

### DIFF
--- a/.buildkite/slow-ThreadNet-tests.sh
+++ b/.buildkite/slow-ThreadNet-tests.sh
@@ -25,8 +25,8 @@ set -euo pipefail
 rows=(
     # From the slowest individual invocation ...
     '1 Cardano    1500'  # ~70 minutes per invocation
-    '2 RealTPraos 200'   # ~30 minutes per invocation (but high variance)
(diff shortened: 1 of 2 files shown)


nfrisby:
I've spawned https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/262 to validate this diff (and to see how much speedup we get)

nfrisby (approved):
LGTM.

The nightly (see my above comment) shows the tests are about 9 times faster than they were (I had the old timings -- ie the denominator in that ratio -- lying around, but they were derived from a somewhat recent nightly build.)

So we can increase master's `2*200 + 4*100 = 800` to `9*800 = 7200 = 2*2000 + 8*400`. That's what this diff does, ✔️.

mrBliss:
The regular nightly test passed too: https://buildkite.com/input-output-hk/ouroboros-network-nightly/builds/263#b775acea-9167-4828-9854-64c3b9dc7d3c.

PR #2691, opened 2020-10-19
Handshake Improvements.

karknu:
Change acceptableVersion so that it also returns the "acceptable" data
agreed upon by the client and server.
Update the client so that it checks that response is acceptable.
Change runApplication to take the result of the version negotiation
instead of a local and remote version of protocol options.

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Client.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Client.hs
@@ -28,13 +28,14 @@ import           Ouroboros.Network.Protocol.Handshake.Version
 handshakeClientPeer
   :: Ord vNumber
   => VersionDataCodec extra CBOR.Term vNumber agreedOptions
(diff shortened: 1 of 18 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Client.hs
```diff
@@ -56,11 +57,13 @@ handshakeClientPeer VersionDataCodec {encodeData, decodeData, getAgreedOptions}
                 Done TokDone (Left (HandshakeError $ HandshakeDecodeError vNumber err))
 
               Right vData' ->
-                -- TODO: we should check that we agree on received @vData'@,
-                -- this might be less trivial than testing for equality.
-                Done TokDone $ Right $ ( runApplication (versionApplication version) vData vData'
-                                       , getAgreedOptions (versionExtra version) vNumber vData'
-                                       )
+                case acceptVersion (versionExtra version) vData vData' of
```
coot, line 60: Thanks for finishing that `TODO`!.

coot (approved):
LGTM

PR #2692, opened 2020-10-19
ledger state query to return everything possible

JaredCorduan:
This (draft) PR extends the data returned by the ledger state query so that it returns the entire `ShelleyState`.

On the downside, it makes an already big state dump even bigger. It adds three maps approximately the size of the number of registered stake pools, and a map that is approximately the size of the number of registered stake credentials. On the upside, it will let us debug some things where we are currently in the dark. In particular, we'll be able to see the block counts, the pool distribution, and the reward update.

thoughts?

### DIFF
--- a/ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
+++ b/ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
@@ -73,7 +73,7 @@ instance CanMock era => Arbitrary (SomeBlock Query (ShelleyBlock era)) where
     , pure $ SomeBlock GetCurrentPParams
     , pure $ SomeBlock GetProposedPParamsUpdates
     , pure $ SomeBlock GetStakeDistribution
-    , pure $ SomeBlock GetCurrentEpochState
+    , pure $ SomeBlock GetCurrentShelleyState
     , (\(SomeBlock q) -> SomeBlock (GetCBOR q)) <$> arbitrary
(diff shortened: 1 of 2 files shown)


mrBliss (approved):
Code changes: LGTM

Making the change: if it helps debugging, sure. This does mean we're kind of changing the API, because we're now sending more data back, which clients might not be able to decode. It just means `cardano-node` needs to be updated, but the type checker will lead the way. Are you aware of any other users of this query?

Anyway, I'm not really worried. Let's do it, but if somebody starts complaining, I'll refer them to you :smile:

mrBliss (changes requested):
Actually, let me try to version these queries properly. I have two other queries I need to add, I'll include this one too and require a new version number.

(Requesting changes so that I include these changes in a PR with a new version).

JaredCorduan:
> Making the change: if it helps debugging, sure. This does mean we're kind of changing the API, because we're now sending more data back, which clients might not be able to decode. It just means `cardano-node` needs to be updated, but the type checker will lead the way. Are you aware of any other users of this query?

indeed it is changing the API, but I was encouraged by the comment above it:
```
   -- | Only for debugging purposes, we don't guarantee binary compatibility.
   -- Moreover, it is huge.
```
I am not aware of any other users. Unless you include folks who use the output of the ledger state dump from the CLI, there are some community tools that use it, but I suspect they'll actually be very happy to have access to the reward updates.

> Anyway, I'm not really worried. Let's do it, but if somebody starts complaining, I'll refer them to you

fair enough :smile:

PR #2693, opened 2020-10-20
Update index-state to 2020-10-01

newhoggy:


cabal.project
```diff
@@ -124,28 +124,29 @@ package small-steps-test
...
 
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-prelude
-  tag: 3376feb66f94e4a12ad0b51efa2f252cc7f52967
-  --sha256: 0m6f0m8dsyji1l4c4n8yr7axk8mljwfajk7qp07zaw5fzrsd4abf
-  subdir: test
+  tag: a5519e09958ad1605ed438d26dd7aad39167d0f9
+  --sha256: 03v46yn5bnkmwcm1zwihjhqvma4ssh3s1s1bfdizvq18y1janwf1
+  subdir: cardano-prelude-test
```
mrBliss, line 143: Merge these to and use two `subdir`s, like in the rest of the file :slightly_smiling_face:

cabal.project
```diff
@@ -170,6 +171,7 @@ source-repository-package
     shelley/chain-and-ledger/dependencies/non-integer
     shelley/chain-and-ledger/executable-spec
     shelley/chain-and-ledger/shelley-spec-ledger-test
+    shelley-ma/impl
```
mrBliss, line 174: Is this one needed already? I was only going to add it when we really use it (#2679).

cabal.project
```diff
@@ -156,8 +157,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/cardano-ledger-specs
-  tag: d5eaac6c4b21a8e69dc3a5503a72e3c3bfde648e
-  --sha256: 1lzwfi6bc7z995s345ij6aachsrmhmrgm71060z6rvk1w97b3jqk
+  tag: c4aab5045977ab2bf45a27f5804cfcbe2509fc5e
+  --sha256: 0n3n4vkaznrbaqhbh3iqlgfglfwxg04bzag8q76xy1qz0l8v2093
```
mrBliss, line 161: Any breaking changes?

mrBliss:
@newhoggy I'm curious why you are pushing this change. `cardano-node` already uses the new index-state. Is it just for consistency?

It's just that it's not ideal when people unrelated to the repo or dependency, update a dependency that might cause lots of breaking changes (see the CI failures). I'm currently working on a dependency update of `cardano-ledger-specs` in #2679, but first three more PRs (https://github.com/input-output-hk/cardano-ledger-specs/pull/1922, https://github.com/input-output-hk/cardano-ledger-specs/pull/1902, https://github.com/input-output-hk/cardano-ledger-specs/pull/1923) need to be merged in `cardano-ledger-specs`.

Unless you can drop the dependency bump of `cardano-ledger-specs`, this PR will have to wait until the four PRs I mentioned are merged.

cabal.project
newhoggy: I only added it because the build failed.

newhoggy:
Thanks for your advice.  Fortunately, I the earlier decision to make the new version of `cardano-prelude` backwards compatible means I ought to be able to do as you suggest and not bump `cardano-ledger-spec`.  I'll try that now and see how it goes.

mrBliss (approved):

mrBliss:
bors merge

PR #2694, opened 2020-10-20
Add more queries and start versioning them

### LINKED ISSUE #2690: Add query that returns the current era
It would be convenient for the node cli to know what era we're in.

We can add a constructor to [`QueryHardFork`](https://github.com/input-output-hk/ouroboros-network/blob/a09f209bb6bbb1a1a9a5b31d671da0189001e175/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs#L276).

mrBliss:
Changes:

* Shelley: add `GetGenesisConfig` returning `SL.ShelleyGenesis era`.

* Shelley: add `DebugNewEpochState` returning `SL.NewEpochState era`. The `Debug-` prefix indicates that it is intended for debugging and that we don't guarantee binary compatibility.

* Shelley: add `DebugChainDepState` returning `SL.ChainDepState (EraCrypto era)`.

* HFC: add `GetCurrentEra` to `QueryHardFork` returning `EraIndex xs`. An `EraIndex` is basically a number indicating in which era the node's current ledger is. One can, for example, use `eraIndexToInt :: EraIndex xs -> Int`.

* Rename the `GetCurrentEpochState` query to `DebugEpochState` for consistency.

* When running in Cardano mode, the new queries are only available when the new `CardanoNodeToClientVersion3` is used. The new `NodeToClientV_4` network version was added for this. Unfortunately, we can't enforce this statically, only at run-time, i.e., when sending the queries. Still, it's better to get an exception when sending a new query to a node that doesn't support it than getting *no response at all* because the receiving node failed to decode it.

* When running in pure Shelley mode, the new queries are only available when the new `ShelleyNodeToClientVersion2` is used. 

* `SomeBlock` was renamed to `SomeSecond`, which is a more accurate name.


edsko (approved):
Should we use versioning?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -457,6 +457,11 @@ data instance Query (ShelleyBlock era) :: Type -> Type where
   GetGenesisConfig
     :: Query (ShelleyBlock era) (SL.ShelleyGenesis era)
 
+  -- | Only for debugging purposes, we don't guarantee binary compatibility.
+  -- Moreover, it is huge.
+  GetCurrentShelleyState
+    :: Query (ShelleyBlock era) (SL.ShelleyState era)
```
mrBliss, line 467: Rename `Get-` to `Debug-`, same for `GetCurrentEpochState`.

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
```diff
@@ -30,7 +30,9 @@ data NodeToClientVersion
     | NodeToClientV_2
     -- ^ added local-query mini-protocol
     | NodeToClientV_3
-    -- ^ enabled @CardanoNodeToClientVersion3@
+    -- ^ enabled @CardanoNodeToClientVersion2@
+    | NodeToClientV_4
```
mrBliss, line 34: @coot FYI, this means we're going to start using `NodeToClientV_4` in the next release.

ouroboros-consensus-test/test-consensus/Test/Consensus/HardFork/Combinator.hs
```diff
@@ -376,16 +376,22 @@ instance CanHardFork '[BlockA, BlockB] where
...
+      )
 
 versionN2C :: BlockNodeToClientVersion TestBlock
-versionN2C = HardForkNodeToClientEnabled $
-                  EraNodeToClientEnabled ()
-               :* EraNodeToClientEnabled ()
-               :* Nil
+versionN2C =
+    HardForkNodeToClientEnabled
+      HardForkSpecificNodeToClientVersion1
```
mrBliss, line 390: Use `maxBound` here (add the required instance)

ouroboros-consensus-test/test-consensus/Test/Consensus/HardFork/Combinator.hs
```diff
@@ -376,16 +376,22 @@ instance CanHardFork '[BlockA, BlockB] where
   hardForkInjectTxs = InPairs.mk2 injectTx_AtoB
 
 versionN2N :: BlockNodeToNodeVersion TestBlock
-versionN2N = HardForkNodeToNodeEnabled $
-                  EraNodeToNodeEnabled ()
-               :* EraNodeToNodeEnabled ()
-               :* Nil
+versionN2N =
+    HardForkNodeToNodeEnabled
+      HardForkSpecificNodeToNodeVersion1
```
mrBliss, line 381: Same

ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
```diff
@@ -181,45 +181,53 @@ arbitraryNodeToNode injByron injShelley injAllegra injMary = oneof
     -- V2.
     , (\(WithVersion versionByron b) versionShelley versionAllegra versionMary ->
           WithVersion
-            (HardForkNodeToNodeEnabled (
-                 EraNodeToNodeEnabled versionByron
+            (HardForkNodeToNodeEnabled
+              HardForkSpecificNodeToNodeVersion1
```
mrBliss, line 185: `maxBound`

ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
```diff
@@ -310,45 +318,54 @@ arbitraryNodeToClient injByron injShelley injAllegra injMary = oneof
     -- Byron + HardFork enabled.
     , (\(WithVersion versionByron b) versionShelley versionAllegra versionMary ->
           WithVersion
-            (HardForkNodeToClientEnabled (
-                 EraNodeToClientEnabled versionByron
+            (HardForkNodeToClientEnabled
+              -- We don't bother with version 1, version 2 is a superset
+              HardForkSpecificNodeToClientVersion2
```
mrBliss, line 323: Same

edsko (approved):
Nice, I'm glad we went ahead with the versioning, feels more principled now.

mrBliss:
Draft because this is waiting for https://github.com/input-output-hk/cardano-ledger-specs/pull/1927. Unfortunately, a bunch of changes will be necessary to consensus (and the ledger) before we can update our dependency on `cardano-ledger-specs` to include #1927. Very likely #2679 will need to be merged first.

JaredCorduan:
@mrBliss it's not straight forward to add the nonces to a query, right? i.e. the epoch nonce, the candidate nonce, etc. I know consensus doesn't use the `ChainState`, but those items would also be great to query if it's easy (but I seem to remember that it's not...)

https://github.com/input-output-hk/cardano-ledger-specs/blob/4bcc2edbb9dfa9fe867a04c52da16c6284905947/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Chain.hs#L130-L138

mrBliss:
> @mrBliss it's not straight forward to add the nonces to a query, right? i.e. the epoch nonce, the candidate nonce, etc. I know consensus doesn't use the `ChainState`, but those items would also be great to query if it's easy (but I seem to remember that it's not...)
> 
> https://github.com/input-output-hk/cardano-ledger-specs/blob/4bcc2edbb9dfa9fe867a04c52da16c6284905947/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Chain.hs#L130-L138

Indeed not straightforward. As you allude to, consensus' ledger state only contains the [`ShelleyState`](https://github.com/input-output-hk/ouroboros-network/blob/95c9cd29ed889abe170ab2e9c6e8523b02432f03/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L180). 

We do store the [`ChainDepState`](https://github.com/input-output-hk/cardano-ledger-specs/blob/bc0d9f4e8a5fa1850d406351a4535cc2837206f7/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/API/Protocol.hs#L207) as well, but we consider it "protocol state", not "ledger state". The query functionality is bundled in the `QueryLedger` class, so I'm a tiny bit hesitant to change this. I'll think about this.

mrBliss:
> > @mrBliss it's not straight forward to add the nonces to a query, right? i.e. the epoch nonce, the candidate nonce, etc. I know consensus doesn't use the `ChainState`, but those items would also be great to query if it's easy (but I seem to remember that it's not...)
> > https://github.com/input-output-hk/cardano-ledger-specs/blob/4bcc2edbb9dfa9fe867a04c52da16c6284905947/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/STS/Chain.hs#L130-L138
> 
> Indeed not straightforward. As you allude to, consensus' ledger state only contains the [`ShelleyState`](https://github.com/input-output-hk/ouroboros-network/blob/95c9cd29ed889abe170ab2e9c6e8523b02432f03/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L180).
> 
> We do store the [`ChainDepState`](https://github.com/input-output-hk/cardano-ledger-specs/blob/bc0d9f4e8a5fa1850d406351a4535cc2837206f7/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/API/Protocol.hs#L207) as well, but we consider it "protocol state", not "ledger state". The query functionality is bundled in the `QueryLedger` class, so I'm a tiny bit hesitant to change this. I'll think about this.

@JaredCorduan After discussing this with Edsko, we have come up with a good approach to allow you to query the `ChainDepState` too. I'll start implementing it.

JaredCorduan:
> @JaredCorduan After discussing this with Edsko, we have come up with a good approach to allow you to query the `ChainDepState` too. I'll start implementing it.

That's really great news! Thank you :)

edsko (approved):
Slightly sad the Right Thing :tm: didn't work out but this is easier for sure.

mrBliss:
bors merge

Issue #2684, opened 2020-10-14
Handle `bind` failure


karknu:
If bind fails for the responder socket we should log it and exit with an error.

Issue #2519, opened 2020-08-18
Silent failure on opening of unix domain socket


newhoggy:
I tried opening a Unix domain socket filename with an invalid filename where it was longer than the maximum of `104` characters.  In the REPL, it fails like this:

```
λ> import Prelude
λ> import qualified Ouroboros.Network.NodeToClient as NTC
λ> import qualified Ouroboros.Network.Snocket as Snocket
λ> let path = "/private/var/folders/zh/ln41q4zs52x2fd61rxccmq640000gn/T/temp/chairman/test-13dfca22944230a2/socket/mocket" :: String
λ> NTC.withIOManager $ \iocp -> do { sn <- pure $ Snocket.localSnocket iocp path; sd <- Snocket.open sn (Snocket.addrFamily sn $ Snocket.localAddressFromPath path); Snocket.bind sn sd $ Snocket.localAddressFromPath path; Snocket.listen sn sd}
*** Exception: pokeSockAddr: path is too long
CallStack (from HasCallStack):
  error, called at Network/Socket/Types.hsc:1077:40 in ntwrk-3.1.1.1-58eabeff:Network.Socket.Types
```

In `cardano-node`, the process continues with no evidence of failure except that the socket file cannot be found.

My expectation is that the node should fail to start.

PR #2696, opened 2020-10-21
Handle exiting threads in diffusion layer

karknu:
If a thread spawned by diffusion exits for any reason we kill
all other threads and return. This is done to ensure that the node isn't
left running in a semi-functional state, for example with the client
side working but the server has exited due to a configuration error.

Example output:
`cardano-node: Network.Socket.bind: unsupported operation (Can't assign requested address)⏎`

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Diffusion.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Diffusion.hs
@@ -171,24 +171,32 @@ runDataDiffusion tracers
         Async.withAsync (cleanNetworkMutableState networkLocalState) $ \cleanLocalNetworkStateThread ->
 
           -- fork server for local clients
-          Async.withAsync (runLocalServer iocp networkLocalState) $ \_ ->
+          Async.withAsync (runLocalServer iocp networkLocalState) $ \localServerThread ->
 
               -- fork ip subscription
-              Async.withAsync (runIpSubscriptionWorker snocket networkState lias) $ \_ ->
+              Async.withAsync (runIpSubscriptionWorker snocket networkState lias) $ \ipSubThread ->
 
                 -- fork dns subscriptions
-                withAsyncs (runDnsSubscriptionWorker snocket networkState lias <$> daDnsProducers) $ \_ ->
+                withAsyncs (runDnsSubscriptionWorker snocket networkState lias <$> daDnsProducers) $ \dnsSubThreads ->
 
                   case daDiffusionMode of
                     InitiatorAndResponderDiffusionMode ->
                       -- fork servers for remote peers
-                      withAsyncs (runServer snocket networkState . fmap Socket.addrAddress <$> addresses) $ \_ ->
-                        -- If any other threads throws 'cleanNetowrkStateThread' and
-                        -- 'cleanLocalNetworkStateThread' threads will will finish.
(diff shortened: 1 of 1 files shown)


karknu:
Fixes #2684 .

coot (approved):
LGTM

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -319,9 +327,10 @@ runDataDiffusion tracers
             sd
             (daLocalResponderApplication applications)
             localErrorPolicy
+          return ()
```
coot, line 330: Do we need this `return`? There's `void` applied to previous monadic action.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -350,11 +359,12 @@ runDataDiffusion tracers
             sd
             (daResponderApplication applications)
             remoteErrorPolicy
+          return ()
```
coot, line 362: The same here.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -369,13 +379,13 @@ runDataDiffusion tracers
         , spErrorPolicies          = remoteErrorPolicy
         , spSubscriptionTarget     = daIpProducers
         }
-      (daInitiatorApplication applications)
+      (daInitiatorApplication applications) >> return ()
```
coot, line 382: It would be cleaner to write:
```
void $ NodeToNode.ipSubscriptionWorker ...
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -391,7 +401,7 @@ runDataDiffusion tracers
         , spErrorPolicies          = remoteErrorPolicy
         , spSubscriptionTarget     = dnsProducer
         }
-      (daInitiatorApplication applications)
+      (daInitiatorApplication applications) >> return ()
```
coot, line 404: The same here.

coot (changes requested):
Some minor comments.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
karknu: No, the `return` isn't needed. Will remove it.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
karknu: Same reply ;)

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
karknu: 👍

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
karknu: 👍

karknu:
This should fix #2519 too.

coot (approved):
LGTM

PR #2701, opened 2020-10-22
cardano-ping as a demo

karknu:
A mux demo that uses ouroboros-network's keepalive protocol
to mimic the ping command.

### DIFF
--- a/network-mux/demo/cardano-ping.hs
+++ b/network-mux/demo/cardano-ping.hs
@@ -1,0 +1,343 @@
+{-# LANGUAGE BangPatterns      #-}
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE NamedFieldPuns    #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards   #-}
+{-# LANGUAGE RankNTypes        #-}
+
+module Main (main) where
+
+import qualified Codec.CBOR.Encoding as CBOR
+import qualified Codec.CBOR.Decoding as CBOR
+import qualified Codec.CBOR.Read     as CBOR
+import qualified Codec.CBOR.Write    as CBOR
+import           Control.Exception
+import           Control.Monad (replicateM, when, unless)
+import           Control.Monad.Class.MonadAsync
+import           Control.Monad.Class.MonadTime
+import           Control.Monad.Class.MonadTimer hiding (timeout)
+import           Control.Tracer (nullTracer)
+import           Data.Aeson hiding (Options, json)
+import           Data.ByteString.Lazy (ByteString)
+import qualified Data.ByteString.Lazy as BL
+import qualified Data.ByteString.Lazy.Char8 as C8 (putStrLn)
+import           Data.Maybe (fromMaybe, isNothing)
+import           Data.Word
+import           Data.TDigest
+import           Data.Text (unpack)
+import           Network.Socket ( AddrInfo)
+import qualified Network.Socket as Socket
+import           Text.Printf
+import           System.Environment (getArgs)
+import           System.Exit
+import           System.Console.GetOpt
+import           System.IO (hFlush, hPutStr, stderr, stdout)
+
+import           Network.Mux.Bearer.Socket
+import           Network.Mux.Types
+import           Network.Mux.Timeout
+
+mainnetMagic :: Word32
+mainnetMagic = 764824073
+
+handshakeNum ::  MiniProtocolNum
+handshakeNum = MiniProtocolNum 0
+
+keepaliveNum :: MiniProtocolNum
+keepaliveNum = MiniProtocolNum 8
+
+data Flag = HostF String | PortF String | MagicF String | QuietF | JsonF deriving Show
+
+optionDescriptions :: [OptDescr Flag]
+optionDescriptions = [
+    Option "h" ["host"]  (ReqArg HostF "host")  "hostname/ip, e.g relay.iohk.example",
+    Option "m" ["magic"] (ReqArg MagicF "magic") ("magic, defaults to " ++ show mainnetMagic),
+    Option "j" ["json"]  (NoArg JsonF ) "json output flag",
+    Option "p" ["port"]  (ReqArg PortF "port") "portnumber, e.g 1234",
+    Option "q" ["quiet"] (NoArg QuietF ) "quiet flag, csv/json only output."
+  ]
+
+data Options = Options {
+      host  :: Maybe String
+    , port  :: String
+    , magic :: Word32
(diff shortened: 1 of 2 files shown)


network-mux/network-mux.cabal
```diff
@@ -161,3 +161,36 @@ executable mux-demo
...
+                       network,
+                       serialise,
+                       tdigest,
+                       text
+  if os(windows)
+    buildable:         False
+  else
+    buildable:         True
+  default-language:    Haskell2010
+  ghc-options:         -threaded
```
coot, line 189: Maybe we can use the [common](https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-section-common-common) stanza for common `ghc` options?  It requires`cabal-version: 2.2`.

network-mux/network-mux.cabal
```diff
@@ -161,3 +161,36 @@ executable mux-demo
...
+executable cardano-ping
+  hs-source-dirs:      demo, test
+  main-is:             cardano-ping.hs
+  other-modules:       Test.Mux.ReqResp
+  build-depends:       base,
+                       directory,
+                       network-mux,
+                       io-sim-classes,
+                       contra-tracer,
+                       stm,
```
coot, line 175: What do you need from `stm` package that's not in `io-sim-classes`?

network-mux/network-mux.cabal
```diff
@@ -161,3 +161,36 @@ executable mux-demo
...
+
+                       binary,
+                       bytestring,
+                       cborg,
+                       network,
+                       serialise,
+                       tdigest,
```
coot, line 182: :+1:

network-mux/demo/cardano-ping.hs
```diff
@@ -0,0 +1,342 @@
...
+handshakeNum = MiniProtocolNum 0
+
+keepaliveNum :: MiniProtocolNum
+keepaliveNum = MiniProtocolNum 8
+
+data Flag = HostF String | PortF String | MagicF String | QuietF | JsonF deriving Show
+
+options :: [OptDescr Flag]
+options = [
+    Option "h" ["h"]     (ReqArg HostF "host")  "hostname/ip, e.g relay.iohk.example",
```
coot, line 51: Just neat picking: I'd use`['h']` rather than `"h"`.   I would also make the long option `--host` rather than `--h`, there's already `-h`.

network-mux/demo/cardano-ping.hs
```diff
@@ -0,0 +1,341 @@
...
+
+main :: IO ()
+main = do
+    args <- getArgs
+    let (actions, _, _ ) = getOpt RequireOrder options args
+        host = getHost actions
+        port = getPort actions
+        quiet = getQuiet actions
+        magic = getMagic actions
+        jsonF = getJson actions
```
coot, line 91: Another way is to use a record type with a monoid instance and `foldMap` over `getArgs` results. It might be a bit shorter and more elegant: it would need the record type definition, `Semigroup` and `Monoid` instances and `foldMap` accumulator function.

network-mux/demo/cardano-ping.hs
```diff
@@ -0,0 +1,341 @@
...
+        putStrLn "                           timestamp,                 host,  cookie,  sample,  median,     p90,    mean,     min,     max,     std"
+
+    caids <- mapM (async . pingClient quiet magic jsonF) addresses
+    mapM_ wait caids
+
+data NodeToNodeVersion = NodeToNodeVersionV1 Word32
+                       | NodeToNodeVersionV2 Word32
+                       | NodeToNodeVersionV3 Word32
+                       | NodeToNodeVersionV4 Word32 Bool
+                       deriving (Eq, Ord, Show)
```
coot, line 105: So you are not using dependently typed version data :).

network-mux/network-mux.cabal
karknu: cabal-version 2.2 shouldn't be a problem, we already require a far newer version.

network-mux/network-mux.cabal
karknu: Nothing, will remove.

network-mux/demo/cardano-ping.hs
karknu: Will fix the long option.

network-mux/demo/cardano-ping.hs
karknu: Reworked to use `foldr`

network-mux/demo/cardano-ping.hs
```diff
@@ -46,56 +47,55 @@ keepaliveNum = MiniProtocolNum 8
...
-        quiet = getQuiet actions
-        magic = getMagic actions
-        jsonF = getJson actions
+    let (flags, _, _ ) = getOpt RequireOrder optionDescriptions args
+        options = foldr buildOptions defaultOpts flags
         hints = Socket.defaultHints { Socket.addrSocketType = Socket.Stream }
-    addresses <- Socket.getAddrInfo (Just hints) (Just host) (Just port)
 
-    unless jsonF $
+    when (isNothing $ host options) $ error "Specify host/ip with '-h <hostname>'"
```
coot, line 91: It would be nicer to `putStrlLn` and then throw `ExitFailure 1 :: ExitCode` (or whatever code that fits better).

coot (approved):
Just a minor comment, otherwise pretty nice!

network-mux/demo/cardano-ping.hs
karknu: :+1:

njd42 (commented):
I see Marcin is handling the low level questions, I'll ask some of the higher level ones. 

Are you seeing this purely as a "demo" or as a utility - i.e it is used by people to create measuremetn statistics (e.g. pipe the output into something that plots it)?

karknu:
> I see Marcin is handling the low level questions, I'll ask some of the higher level ones.
> 
> Are you seeing this purely as a "demo" or as a utility - i.e it is used by people to create measuremetn statistics (e.g. pipe the output into something that plots it)?

I see it as a simple tool for troubleshooting and a way to get some quick statistics for plotting. However it is not meant to be something that monitors nodes 24x7 (it doesn't reconnect).

When `Cardano.Api` is ready we should implement a `cardano-ping` with it, at which point this implementation will become just a demo.

coot:
> Are you seeing this purely as a "demo" or as a utility - i.e it is used by people to create measuremetn statistics (e.g. pipe the output into something that plots it)?

devops are keen to use it to:

* to trouble should when other people come to them with misconfigured nodes (running on wrong network, e.g. test-net vs main-net)
* to substitute `nc` in their _topology health check_

PR #2704, opened 2020-10-23
Shelley: only keep a compact genesis config in memory

mrBliss:
Some fields in the Shelley genesis config are only needed at start-up. There's
no point in keeping them in memory during the rest of the execution. In some
testing setups, e.g., for transaction benchmarking, the genesis config can be
quite large.

This also means that the `GetGenesisConfig` query (#2694) would only return the
compact genesis config, not the original full one.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
@@ -11,6 +11,9 @@ module Ouroboros.Consensus.Shelley.Ledger.Config (
   , mkShelleyBlockConfig
   , CodecConfig (..)
   , StorageConfig (..)
+  , CompactGenesis -- opaque
+  , getCompactGenesis
+  , compactGenesis
   ) where
 
 import           Data.Map.Strict (Map)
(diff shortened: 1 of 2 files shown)


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
```diff
@@ -90,3 +93,26 @@ data instance StorageConfig (ShelleyBlock era) = ShelleyStorageConfig {
...
   deriving (Generic, NoThunks)
+
+{-------------------------------------------------------------------------------
+  Compact genesis
+-------------------------------------------------------------------------------}
+
+-- | Compact variant of 'SL.ShelleyGenesis' with some fields erased that are
+-- only used on start-up and that should not be kept in memory forever.
+--
+-- For example, the 'sgInitialFunds' field.
```
mrBliss, line 104: State which two fields are cleared.

edsko (approved):
Not terribly keen on having this "record with some fields erased" approach. Surely it would be cleaner to have a different type that makes it clear that we don't have values for these fields?

mrBliss:
> Not terribly keen on having this "record with some fields erased" approach. Surely it would be cleaner to have a different type that makes it clear that we don't have values for these fields?

That would indeed be cleaner. I guess `ShelleyGenesis` shouldn't have contained these fields in the first place. But it does, and they're part of the hash of the file, and that's part of the chain, ...

Also, we don't want consensus to be in the business of defining genesis configs for a ledger.

mrBliss:
bors merge

PR #2705, opened 2020-10-26
json imp. and count argument

karknu:
Spit out a propper json array of samples.
Add a `-c` argument which limits the number of pings to send

### DIFF
--- a/network-mux/demo/cardano-ping.hs
+++ b/network-mux/demo/cardano-ping.hs
@@ -14,13 +14,14 @@ import qualified Codec.CBOR.Write    as CBOR
 import           Control.Exception
 import           Control.Monad (replicateM, when, unless)
 import           Control.Monad.Class.MonadAsync
+import           Control.Monad.Class.MonadSTM.Strict
 import           Control.Monad.Class.MonadTime
 import           Control.Monad.Class.MonadTimer hiding (timeout)
-import           Control.Tracer (nullTracer)
+import           Control.Tracer (Tracer (..), nullTracer, traceWith)
 import           Data.Aeson hiding (Options, json)
 import           Data.ByteString.Lazy (ByteString)
 import qualified Data.ByteString.Lazy as BL
-import qualified Data.ByteString.Lazy.Char8 as C8 (putStrLn)
+import qualified Data.ByteString.Lazy.Char8 as BSC (pack, putStr)
 import           Data.Maybe (fromMaybe, isNothing)
 import           Data.Word
 import           Data.TDigest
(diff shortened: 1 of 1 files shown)


network-mux/demo/cardano-ping.hs
```diff
@@ -20,7 +20,7 @@ import           Control.Tracer (nullTracer)
 import           Data.Aeson hiding (Options, json)
 import           Data.ByteString.Lazy (ByteString)
 import qualified Data.ByteString.Lazy as BL
-import qualified Data.ByteString.Lazy.Char8 as C8 (putStrLn)
+import qualified Data.ByteString.Lazy.Char8 as C8 (putStr)
```
coot, line 23: A more commonly used name is `BSC`.

network-mux/demo/cardano-ping.hs
```diff
@@ -320,23 +326,26 @@ pingClient Options{quiet, magic, json} peer = bracket
            then return (msBlob sdu, t_e)
            else nextMsg bearer timeoutfn ptclNum
 
-    keepAlive :: MuxBearer IO -> TimeoutFn IO -> String -> TDigest 5 -> Word16 -> IO ()
+    keepAlive :: MuxBearer IO -> TimeoutFn IO -> String -> TDigest 5 -> Word32 -> IO ()
+    keepAlive _ _ _ _ cookie | cookie == maxCount = return ()
     keepAlive bearer timeoutfn peerStr td cookie = do
-        !t_s <- write bearer timeoutfn $ wrap keepaliveNum InitiatorDir (keepAliveReq cookie)
+        when (cookie > 0 && json) $ putStrLn ","
```
coot, line 332: This is not very elegant, but I don't have an idea how to make it nicer.

coot (approved):

network-mux/demo/cardano-ping.hs
karknu: Will change.

network-mux/demo/cardano-ping.hs
karknu: It also doesn't work when we're pinging multiple hosts. Will think of another way.

network-mux/demo/cardano-ping.hs
```diff
@@ -269,8 +302,8 @@ toStatPoint ts host cookie sample td =
     stddev' = fromMaybe 0 (stddev td)
 
 
-pingClient :: Options -> AddrInfo -> IO ()
-pingClient Options{quiet, magic, json} peer = bracket
+pingClient :: (LogMsg -> IO ()) -> Options -> AddrInfo -> IO ()
```
coot, line 305: Why not just use `Tracer IO LogMsg`?

karknu: I have no confident in `Tracer`.

coot: You are using it!

karknu: Any similarity with `Tracer` was unintentional.

coot: `Tracer` was modelled exactly after this case and  `-> a` is **the cannonical** example of a contravariant functor.

karknu: I would never have seen it. Will make the adjustment.

Issue #2368, opened 2020-07-07
Generalise Isomorphic to more than one era


mrBliss:
Motivating examples: `Test.Consensus.Cardano.Examples` and `Test.Consensus.Cardano.Generators`.

edsko:
Could we perhaps use this to lift the restrictions on `QueryAnyTime` / `QueryHradFork`?

edsko:
Can then maybe also simplify `Test.Consensus.Cardano.ByronCompatibility`, and instead of giving instances per type, just give one instancefor the newtype wrapper `ByronToCardano` (and one instance in the other direction).

PR #2709, opened 2020-10-27
Cardano: write some generators generically

mrBliss:
The Cardano generators are currently mostly written by enumerating all eras.
This approach doesn't scale as the number of eras will keep growing.

To improve this, add `Arbitrary` instances for a few era-generic types, e.g.,
`NS`, `Telescope`, ..., and use those.

There are still a bunch of generators that will need a similar treatment.
Related issue: #2368.


edsko (approved):
Reviewed over meet.

mrBliss:
bors merge

PR #2711, opened 2020-10-28
Requests tx in order.

karknu:
There is a problem with the way the TxSubmission protocol tracks which txs has been accepted. The only mechanics provided is a simple window size update. This means that if the server decides to fetch txs in an arbitrary order it is very hard for the client to keep track of which txid's it has sent to the client.

By requesting Tx in the order they where presented (if the server is interested in them) the client can easily keep track of a limited number of TXs (rather than all TXs sent since the start of the connection), and the server can verify that the client does indeed send newish txs.

Requesting tx in order also means that they get added to the mempool faster and thus propagate through the network quicker.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
+++ b/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
@@ -13,7 +13,7 @@ module Ouroboros.Network.TxSubmission.Inbound (
     TxSubmissionProtocolError(..),
   ) where
 
-import           Data.Foldable (foldl')
+import           Data.Foldable (foldl', toList)
 import qualified Data.List.NonEmpty as NonEmpty
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
@@ -71,6 +71,7 @@ data TraceTxSubmissionInbound txid tx = TraceTxSubmissionInbound --TODO
 data TxSubmissionProtocolError =
        ProtocolErrorTxNotRequested
      | ProtocolErrorTxIdsNotRequested
+     | ProtocolErrorTxIdNotNew
   deriving Show
 
 instance Exception TxSubmissionProtocolError where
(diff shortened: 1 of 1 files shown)


karknu:
Fixes #2680 in a controversial way.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -253,11 +256,18 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
+            txidsAndSizeSeq = Seq.fromList txids
+            txidsSet = Set.fromList (map fst txids)
+            unAckSet = Set.fromList $ toList $ unacknowledgedTxIds st
 
         unless (Seq.length txidsSeq <= fromIntegral reqNo) $
           throwM ProtocolErrorTxIdsNotRequested
 
+        -- We are requesting *new* txids from the peer. There shouldn't
+        -- be any overlap with txids that we still haven't processed.
+        unless (Set.intersection txidsSet unAckSet == Set.empty) $
```
mrBliss, line 268: `Set.null`?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -350,12 +360,21 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
         }
       where
+
+        -- TODO: Add to StrictSeq.
+        partition :: (a -> Bool) -> StrictSeq a -> (StrictSeq a, StrictSeq a)
+        partition cond = foldl' part (Seq.empty, Seq.empty)
+          where
+            part (ls, rs) e
+              | cond e    = ((Seq.|>) ls e, rs)
+              | otherwise = (ls, (Seq.|>) rs e)
```
mrBliss, line 370: Alternatively:
```suggestion
        partition :: (a -> Bool) -> StrictSeq a -> (StrictSeq a, StrictSeq a)
        partition cond xs = (Seq.toStrict ls, Seq.toStrict rs)
          where
            (ls, rs) = LazySeq.partition cond xs
```
with: `import qualified Data.Sequence as LazySeq`
I'm not sure which one is more efficient.

karknu:
This doesn't fix #2680 it just makes it harder to trigger.

PR #2712, opened 2020-10-28

mrBliss:
The main reason is to pull in:
https://github.com/input-output-hk/cardano-ledger-specs/pull/1943


deepfire (approved):
bors r+

Issue #2688, opened 2020-10-15
Rate limit invadlid tx


coot:
Add rate limiting of transactions based on their validy history.

gufmar:
This might means the same?
https://github.com/input-output-hk/cardano-node/issues/2020

Issue #2682, opened 2020-10-12
Test submitting duplicate txids using tx-submission application


coot:
Related to #2680

Issue #1736, opened 2020-03-03
Validate tx sizes which are coming with `MsgReplyTxIds`


coot:
When a transaction arrives in `MsgReplyTx` we should pass tx sizes along side the `tx`s.  This will require changes in the codec to allow that.  See 'Annotated` in `cardano-binary` package.

Issue #1721, opened 2020-02-28
Ouroboros.Network.TxSubmission.txSubmissionInbound - request txs in order


coot:
As described in todo
```
        -- TODO: This implementation is deliberately naive, we pick in an
        -- arbitrary order and up to a fixed limit. This is to illustrate
        -- that we can request txs out of order. In the final version we will
        -- try to pick in-order and only pick out of order when we have to.
        -- We will also uses the size of txs in bytes as our limit for
        -- upper and lower watermarks for pipelining. We'll also use the
        -- amount in flight and delta-Q to estimate when we're in danger of
        -- becomming idle, and need to request stalled txs.
        --
```

PR #2718, opened 2020-10-30
Add support for duplicate txids.

karknu:
When the server asks for new txids from the client it can't assume that
those txids doesn't already exist in its buffer.

The TxSubmission test suite has been updated so that it doesn't generate
unique txids anymore.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
+++ b/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
@@ -13,7 +13,7 @@ module Ouroboros.Network.TxSubmission.Inbound (
     TxSubmissionProtocolError(..),
   ) where
 
-import           Data.Foldable (foldl')
+import           Data.Foldable (foldl', toList)
 import qualified Data.List.NonEmpty as NonEmpty
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
@@ -39,8 +39,6 @@ import           Ouroboros.Network.NodeToNode.Version (NodeToNodeVersion)
 import           Ouroboros.Network.Protocol.TxSubmission.Server
 import           Ouroboros.Network.TxSubmission.Mempool.Reader
                      (MempoolSnapshot (..), TxSubmissionMempoolReader (..))
-
-
 
 -- | The consensus layer functionality that the inbound side of the tx
 -- submission logic requires.
@@ -265,14 +263,12 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
         -- the mempool. This prevents us from requesting their corresponding
         -- transactions again in the future.
         let st' = st {
-          requestedTxIdsInFlight = requestedTxIdsInFlight st - reqNo,
-          unacknowledgedTxIds    = unacknowledgedTxIds st <> txidsSeq,
-          availableTxids         = availableTxids st <> txidsMap
+          requestedTxIdsInFlight = requestedTxIdsInFlight st - reqNo
         }
         mpSnapshot <- atomically mempoolGetSnapshot
         continueWithStateM
           (serverIdle n)
-          (acknowledgeTxIdsInMempool st' mpSnapshot)
+          (acknowledgeTxIdsInMempool st' txidsSeq txidsMap mpSnapshot)
 
       CollectTxs txids txs -> do
         -- To start with we have to verify that the txs they have sent us do
@@ -301,7 +297,7 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
             -- combined with the fact that we request txs out of order means
             -- our bufferedTxs has to track all the txids we asked for, even
             -- though not all have replies.
-            bufferedTxs' = bufferedTxs st <> txIdsRequestedWithTxsReceived
+            btxsReq = bufferedTxs st <> txIdsRequestedWithTxsReceived
 
             -- We have to update the unacknowledgedTxIds here eagerly and not
             -- delay it to serverReqTxs, otherwise we could end up blocking in
@@ -311,20 +307,26 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
             -- Check if having received more txs we can now confirm any (in
             -- strict order in the unacknowledgedTxIds sequence).
             (acknowledgedTxIds, unacknowledgedTxIds') =
-              Seq.spanl (`Map.member` bufferedTxs') (unacknowledgedTxIds st)
+              Seq.spanl (`Map.member` btxsReq) (unacknowledgedTxIds st)
 
             -- If so we can submit the acknowledged txs to our local mempool
-            txsReady = foldr (\txid r -> maybe r (:r) (bufferedTxs' Map.! txid))
+            txsReady = foldr (\txid r -> maybe r (:r) (btxsReq Map.! txid))
                              [] acknowledgedTxIds
 
             -- And remove acknowledged txs from our buffer
-            bufferedTxs'' = foldl' (flip Map.delete)
-                                   bufferedTxs' acknowledgedTxIds
+            btxsAcked = foldl' (flip Map.delete)
+                                   btxsReq acknowledgedTxIds
 
-        _ <- mempoolAddTxs txsReady
+        _writtenTxids <- mempoolAddTxs txsReady
+
+        -- If we are acknowleding transactions that are still in unacknowledgedTxIds'
+        -- we need to re-add them so that we also can acknowledge them again later.
+        -- This will happen incase of duplicate txids within the same window.
+        let live =  filter (`elem` unacknowledgedTxIds') $ toList acknowledgedTxIds
+            btxsReAdded = forceElemsToWHNF $ btxsAcked <> (Map.fromList (zip  live (repeat Nothing)))
 
         continueWithStateM (serverIdle n) st {
-          bufferedTxs         = bufferedTxs'',
+          bufferedTxs         = btxsReAdded,
           unacknowledgedTxIds = unacknowledgedTxIds',
           numTxsToAcknowledge = numTxsToAcknowledge st
                               + fromIntegral (Seq.length acknowledgedTxIds)
@@ -338,46 +340,62 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
     -- mempool.
     --
     acknowledgeTxIdsInMempool :: ServerState txid tx
+                              -> StrictSeq txid
+                              -> Map txid TxSizeInBytes
                               -> MempoolSnapshot txid tx idx
                               -> ServerState txid tx
-    acknowledgeTxIdsInMempool st MempoolSnapshot{mempoolHasTx} =
+    acknowledgeTxIdsInMempool st txidsSeq _ _ | Seq.null txidsSeq  = st
+    acknowledgeTxIdsInMempool st txidsSeq txidsMap MempoolSnapshot{mempoolHasTx} =
         -- Return the next 'ServerState'
         st {
           availableTxids      = availableTxids',
           bufferedTxs         = bufferedTxs'',
-          unacknowledgedTxIds = unacknowledgedTxIds',
+          unacknowledgedTxIds = unacknowledgedTxIds'',
           numTxsToAcknowledge = numTxsToAcknowledge st
                               + fromIntegral (Seq.length acknowledgedTxIds)
         }
       where
-        -- Divide the available txs in two: those that are already in the
-        -- mempool and those that are not. We'll request some txs from the
+
+        (ignoredTxids, availableTxidsMp) =
+              Map.partitionWithKey
+                (\txid _ -> mempoolHasTx txid)
+                txidsMap
+
+        (_ignoredTxidsU, availableTxidsU) =
+              Map.partitionWithKey
+                (\txid _ -> elem txid (unacknowledgedTxIds st))
+                txidsMap
+
+        -- Divide the new txids in two: those that are already in the
+        -- mempool or in flight and those that are not. We'll request some txs from the
         -- latter.
-        (ignoredTxids, availableTxids') =
-          Map.partitionWithKey
-            (\txid _ -> mempoolHasTx txid)
-            (availableTxids st)
+        availableTxids' = availableTxids st <> Map.intersection availableTxidsMp availableTxidsU
 
         -- The txs that we intentionally don't request, because they are
         -- already in the mempool, need to be acknowledged.
         --
         -- So we extend bufferedTxs with those txs (so of course they have
         -- no corresponding reply).
-        bufferedTxs' = forceElemsToWHNF $
-                       bufferedTxs st
+        bufferedTxs' = bufferedTxs st
                     <> Map.map (const Nothing) ignoredTxids
+
+        unacknowledgedTxIds' = unacknowledgedTxIds st <> txidsSeq
 
         -- Check if having decided not to request more txs we can now
         -- confirm any txids (in strict order in the unacknowledgedTxIds
         -- sequence). This is used in the 'numTxsToAcknowledge' below
         -- which will then be used next time we SendMsgRequestTxIds.
         --
-        (acknowledgedTxIds, unacknowledgedTxIds') =
-          Seq.spanl (`Map.member` bufferedTxs') (unacknowledgedTxIds st)
+        (acknowledgedTxIds, unacknowledgedTxIds'') =
+          Seq.spanl (`Map.member` bufferedTxs') unacknowledgedTxIds'
 
-        -- If so we can remove acknowledged txs from our buffer
-        --
-        bufferedTxs'' = foldl' (flip Map.delete)
+
+        -- If so we can remove acknowledged txs from our buffer provided that they
+        -- are not still in unacknowledgedTxIds''. This happens in case of duplicate
+        -- txids.
+        bufferedTxs'' = forceElemsToWHNF $ foldl' (\m txid -> if elem txid unacknowledgedTxIds''
+                                              then m
+                                              else Map.delete txid m)
                                 bufferedTxs' acknowledgedTxIds
 
     serverReqTxs :: forall (n :: N).
--- a/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
+++ b/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
@@ -268,7 +268,7 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
         mpSnapshot <- atomically mempoolGetSnapshot
         continueWithStateM
           (serverIdle n)
-          (acknowledgeTxIdsInMempool st' txidsSeq txidsMap mpSnapshot)
+          (acknowledgeTxIds st' txidsSeq txidsMap mpSnapshot)
 
       CollectTxs txids txs -> do
         -- To start with we have to verify that the txs they have sent us do
@@ -297,7 +297,7 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
             -- combined with the fact that we request txs out of order means
             -- our bufferedTxs has to track all the txids we asked for, even
             -- though not all have replies.
-            btxsReq = bufferedTxs st <> txIdsRequestedWithTxsReceived
+            bufferedTxs1 = bufferedTxs st <> txIdsRequestedWithTxsReceived
 
             -- We have to update the unacknowledgedTxIds here eagerly and not
             -- delay it to serverReqTxs, otherwise we could end up blocking in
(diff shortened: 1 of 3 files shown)


karknu:
Fixes #2680 and #2682.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -321,10 +319,16 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
                                    bufferedTxs' acknowledgedTxIds
 
-        _ <- mempoolAddTxs txsReady
+        _writtenTxids <- mempoolAddTxs txsReady
+
+        -- If we are acknowleding transactions that are still in unacknowledgedTxIds'
+        -- we need to re-add them so that we also can acknowledge them again later.
+        -- This will happen incase of duplicate txids within the same window.
+        let live =  filter (`elem` unacknowledgedTxIds') $ toList acknowledgedTxIds
+            !bufferedTxs''' = bufferedTxs'' <> (Map.fromList (zip  live (repeat Nothing)))
```
mrBliss, line 328:
```suggestion
            !bufferedTxs''' = bufferedTxs'' <> Map.fromList (zip live (repeat Nothing))
```
Also, the moment we need three primes (`'''`) might be a sign that we should switch to a different naming scheme. Either by numbering or by using different, more descriptive names.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -338,25 +342,36 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
     -- mempool.
     --
     acknowledgeTxIdsInMempool :: ServerState txid tx
+                              -> (StrictSeq txid)
+                              -> (Map txid TxSizeInBytes)
```
mrBliss, line 346:
```suggestion
                              -> StrictSeq txid
                              -> Map txid TxSizeInBytes
```

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -367,17 +382,24 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
                        bufferedTxs st
                     <> Map.map (const Nothing) ignoredTxids
 
+
```
mrBliss, line 385:
```suggestion
```

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -367,17 +382,24 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
-          Seq.spanl (`Map.member` bufferedTxs') (unacknowledgedTxIds st)
+        (acknowledgedTxIds, unacknowledgedTxIds'') =
+          Seq.spanl (`Map.member` bufferedTxs') unacknowledgedTxIds'
 
-        -- If so we can remove acknowledged txs from our buffer
-        --
-        bufferedTxs'' = foldl' (flip Map.delete)
+
+        -- If so we can remove acknowledged txs from our buffer provided that they
+        -- are not still in unacknowledgedTxIds''. This happens incase of duplicate
```
mrBliss, line 398:
```suggestion
        -- are not still in unacknowledgedTxIds''. This happens in case of duplicate
```

mrBliss (commented):
What's the status of this PR?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
coot: I am more and more leaning toward using unicode `bufferedTxs₃` (this makes `agda` code easier to read - though harder to type).

karknu:
> What's the status of this PR?

From my point of view it is ready and waiting on reviews.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
karknu: I'll figure out a better naming scheme that doesn't require unicode.

mrBliss:
@nfrisby can you take a look at this?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
coot: The revolution is still to happen :D ! Unicode has its WATs: how's that possible that there's not supscript `b,c,d`.

coot:
Rebasing on top of `mater`  should fix `Windows` CI build error.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -311,20 +307,26 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
+                                   btxsReq acknowledgedTxIds
+
+        _writtenTxids <- mempoolAddTxs txsReady
 
-        _ <- mempoolAddTxs txsReady
+        -- If we are acknowleding transactions that are still in unacknowledgedTxIds'
+        -- we need to re-add them so that we also can acknowledge them again later.
+        -- This will happen incase of duplicate txids within the same window.
+        let live =  filter (`elem` unacknowledgedTxIds') $ toList acknowledgedTxIds
+            btxsReAdded = forceElemsToWHNF $ btxsAcked <> (Map.fromList (zip  live (repeat Nothing)))
```
mrBliss, line 326: Does this `forceElemsToWHNF` force the thunks? Because the type of `bufferedTxs` is `Map txid (Maybe tx)`, so even forcing a `Maybe tx` to WHNF won't force the `tx` to WHNF (unless it already is).

What we in consensus typically do is replace `Maybe` with a strict variant, often specific to the use case. The constructor names would then also explain what `Nothing` and `Just` mean.

Actually, could you try removing the `forceElemsToWHNF` and replace `<>` with `Map.union`? I think the problem might actually be that `Map.Strict` is the same type as `Map.Lazy` and that the `Semigroup` instance of `Map.Lazy` is being used here.

karknu: Removing `forceElemsToWHNF` and replacing `<>` with `Map.union` seems to still work. However I seem to be unable to trigger a "thunk invariance failure" at all so I'm not going to address this.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -311,20 +307,26 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
-                                   bufferedTxs' acknowledgedTxIds
+            btxsAcked = foldl' (flip Map.delete)
+                                   btxsReq acknowledgedTxIds
+
+        _writtenTxids <- mempoolAddTxs txsReady
 
-        _ <- mempoolAddTxs txsReady
+        -- If we are acknowleding transactions that are still in unacknowledgedTxIds'
+        -- we need to re-add them so that we also can acknowledge them again later.
+        -- This will happen incase of duplicate txids within the same window.
```
nfrisby, line 324: > duplicate txids within the same window.

Is there a simple-enough example you could record in this code comment of how this might happen? I don't recall; it's been too long since I touched that repro we had a while ago.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -338,46 +340,62 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
       where
-        -- Divide the available txs in two: those that are already in the
-        -- mempool and those that are not. We'll request some txs from the
+
+        (ignoredTxids, availableTxidsMp) =
+              Map.partitionWithKey
+                (\txid _ -> mempoolHasTx txid)
+                txidsMap
+
+        (_ignoredTxidsU, availableTxidsU) =
```
nfrisby, line 364: Use `Map.filterWithKey`?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -338,46 +340,62 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
+          Seq.spanl (`Map.member` bufferedTxs') unacknowledgedTxIds'
 
-        -- If so we can remove acknowledged txs from our buffer
-        --
-        bufferedTxs'' = foldl' (flip Map.delete)
+
+        -- If so we can remove acknowledged txs from our buffer provided that they
+        -- are not still in unacknowledgedTxIds''. This happens in case of duplicate
+        -- txids.
+        bufferedTxs'' = forceElemsToWHNF $ foldl' (\m txid -> if elem txid unacknowledgedTxIds''
```
nfrisby, line 396: Worth creating a set for use when checking membership? As I mentioned elsewhere, I don't have a sense of what the expected sizes will be.

nfrisby (commented):
This is tricky stuff! I'm glad you figured out the bug.

I apologize for only responding two weeks later now :(

I left a bunch of comments throughout, some of which are minor. Mostly about code clarity, suggestive names, etc.

The potentially blockers are the questions about efficiency (some expressions went from logarithmic to linear eg) but I don't know the relative sizes involved. Have you considered those trade-offs? If so, it'd be valuable to record your thoughts in the comments.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
```diff
@@ -188,8 +174,8 @@ txSubmissionOutbound tracer maxUnacked TxSubmissionMempoolReader{..} _version co
 
           MempoolSnapshot{mempoolLookupTx} <- atomically mempoolGetSnapshot
 
-          let txidxs  = [ Map.lookup txid unackedMap | txid <- txids ]
-              txidxs' = catMaybes txidxs
+          let txidxs  = [ find (\(t,_) -> t == txid) unackedSeq | txid <- txids ]
```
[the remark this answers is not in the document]
karknu: `unackedMap` has to be removed since it is keyed on `txid` and can't cope with duplicate txid in the same window.
The tx window size is 10. It is not configurable and requires a new protocol version to change.

ouroboros-network/test/Test/Ouroboros/Network/TxSubmission.hs
```diff
@@ -261,26 +267,103 @@ instance Arbitrary a => Arbitrary (LargeNonEmptyList a) where
...
+            ) in
+    ioProperty $ do
+        tr' <- evaluateTrace tr
+        case tr' of
+             SimException e trace -> do
+                return $ counterexample (intercalate "\n" $ show e : trace) False
+             SimDeadLock trace -> do
+                 return $ counterexample (intercalate "\n" $ "Deadlock" : trace) False
+             SimReturn (inmp, outmp) _trace -> do
+                 -- printf "Log: %s\n" (intercalate "\n" _trace)
```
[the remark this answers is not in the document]
karknu: I prefer to keep it.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
karknu: The tx windows size is 10 so I don't think a Set is worth it.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -265,14 +263,12 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
         -- the mempool. This prevents us from requesting their corresponding
         -- transactions again in the future.
         let st' = st {
-          requestedTxIdsInFlight = requestedTxIdsInFlight st - reqNo,
```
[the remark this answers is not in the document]
karknu: Perhaps `acknowledgeTxIds` is more suitable. The function acknowledges txids that are either in mempool or that we've learned of already but not yet comited.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -311,20 +307,26 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
 
             -- If so we can submit the acknowledged txs to our local mempool
-            txsReady = foldr (\txid r -> maybe r (:r) (bufferedTxs' Map.! txid))
+            txsReady = foldr (\txid r -> maybe r (:r) (btxsReq Map.! txid))
                              [] acknowledgedTxIds
 
             -- And remove acknowledged txs from our buffer
-            bufferedTxs'' = foldl' (flip Map.delete)
-                                   bufferedTxs' acknowledgedTxIds
+            btxsAcked = foldl' (flip Map.delete)
```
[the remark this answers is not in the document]
karknu: Will change to `bufferedTxsN`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
karknu: This is the simplest example that quickcheck could generate. I don't think it is concise enough to put it in a code comment.
Consider the following list of TXs and a window size greater than or equal to five.
[Tx 0, Tx 1, Tx 2, Tx 3, Tx 2]

 1. The server requests 3 new TxIds.
 2. The client responds with [TxId 0, TxId 1, TxId 2].
 3. The server requests [Tx 0, Tx 1].
 4. The server acks [TxId 0, TxId 1] and request 2 new TxIds.
 5. The client responds with [Tx 0, Tx 1]
 6. The server reqest Tx 2.
 7. The client responds with [TxId 3, TxId 2 (dupe)].
 8. The client responds with Tx 2.
 9. The server request Tx 3
10. The server acks [TxId 2, TxId 3] and request two new TxIds. Note that only the TxId 2 from msg 2 is acked, not the dupe TxId2 from msg 7.
11. The client responds with no more Txids.
12. The server will trigger an assertion failure (line 199) because it has failed to acknowledge the 2 (dupe) TxId.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -311,20 +307,26 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
+                                   btxsReq acknowledgedTxIds
+
+        _writtenTxids <- mempoolAddTxs txsReady
 
-        _ <- mempoolAddTxs txsReady
+        -- If we are acknowleding transactions that are still in unacknowledgedTxIds'
+        -- we need to re-add them so that we also can acknowledge them again later.
+        -- This will happen incase of duplicate txids within the same window.
+        let live =  filter (`elem` unacknowledgedTxIds') $ toList acknowledgedTxIds
```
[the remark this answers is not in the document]
karknu: Will fix.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
karknu: Will change.

ouroboros-network/test/Test/Ouroboros/Network/TxSubmission.hs
```diff
@@ -261,26 +267,103 @@ instance Arbitrary a => Arbitrary (LargeNonEmptyList a) where
...
+                          return $ map getTxId inmp === take (length inmp) (map getTxId $
+                              filter getTxValid outUniqueTxIds)
+                      (False, False)
+                           -- If we are presented with a stream of valid and invalid Txs with
+                           -- duplicate txids we're content with completing the protocol
+                           -- without error.
+                           -> return $ property True
+
+
+-- TODO: Belongs in iosim.
```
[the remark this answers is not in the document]
karknu: I have no interest in including `go` directly into my property. I prefer to keep it in a form that makes it easy to cut-and-paste into every future test case I write.

karknu:
> This is tricky stuff! I'm glad you figured out the bug.
> 
> I apologize for only responding two weeks later now :(
> 
> I left a bunch of comments throughout, some of which are minor. Mostly about code clarity, suggestive names, etc.
> 
> The potentially blockers are the questions about efficiency (some expressions went from logarithmic to linear eg) but I don't know the relative sizes involved. Have you considered those trade-offs? If so, it'd be valuable to record your thoughts in the comments.

Thank you for your review. The current window size is 10, so I have not considered complexity.
The most obvious performance improvement would be to fetch requests in order (#1721). The arbitrary fetching order increases the time from when we receive a new Tx to when we can commit it to the mempool by several RTTs.
There are other TxSubmission related issues that I hope to get to address soon such as #2688, #2615, and #1736.
However this PR is only for duplicate TX id support.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
nfrisby: Thanks, that's helpful.

Hmm. Could the comment say "duplicate txids can arise merely as a consequence of pipelining the current BF protocol"? Do you think that is an accurate summary?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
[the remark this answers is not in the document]
nfrisby: `acknowledgeTxIds` sounds good to me 👍. Having recently revisited the documentation on the state fields, the fact that this involves updating `bufferedTxs` now seems unsurprising to me.

ouroboros-network/test/Test/Ouroboros/Network/TxSubmission.hs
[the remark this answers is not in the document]
nfrisby: With that feedback, I'll refine my suggestion to define a combinator like this one. The main refinement is that it doesn't necessitate the `[String]` space-leak and so seems like a good candidate for inclusion in `io-sim`.

```
foldSayResultTrace ::
     (Time -> ThreadId -> String -> a -> a)   -- ^ handle an 'EventSay'
  -> (a -> Either Failure b -> c)  -- ^ handle end of trace
  -> (a -> SomeException -> c)  -- ^ handle pure exception thrown by definition of trace
  -> a
  -> Trace b
  -> c
```

Main point: Whatever the details end up being, I agree that a special combinator focusing only on `say` and termination events would be helpful. It could be another `TraceSay` data type with its own fold function or just the Church encoding as above, no personal preference. For example, I'm mainly envisioning `a ~ c ~ Test.QuickCheck.Property`.

nfrisby:
> The current window size is 10, so I have not considered complexity.

I think "the window size is small" resolves all my complexity concerns. I'm tempted to record that in a comment as relevant contextual information. Your call 👍.

Last question: do you want to factor out the `io-sim` wrapper before merging this PR?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
karknu: In the example above pipelining permitted the duplicate txids to trigger an assertion, but pipelining didn't create the duplicate Tx id. The client decided to send a duplicate Tx id within the same window. I think in theory this could happen in case the client sent us a new TxId, rolled back the a fork without that Tx, fetched the Tx again from another peer and then sent us the same TxId again.

It is also possible to trigger this bug if the Txs aren't fetched in order (they are fetched based on `Ord` TxId, not in the order they where presented), in away that doesn't depend on pipelining.

karknu:
> > The current window size is 10, so I have not considered complexity.
> 
> I think "the window size is small" resolves all my complexity concerns. I'm tempted to record that in a comment as relevant contextual information. Your call 👍.

I added a comment regarding the windows size.

> 
> Last question: do you want to factor out the `io-sim` wrapper before merging this PR?

I expect more resistance/comments on adding `evaluateTrace` to io-sim so I'm fine with just having it as a TODO.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
mrBliss: I still think that `Map.union` is cleaner than `forceElemsToWHNF` (no need to force thunks if you don't even introduce them).

nfrisby (approved):
OK, yeah, a max window size of 10 seems like it'll truncate any of these potential algorithmic complexity issues. Thanks! LGTM! Approving.

mrBliss:
Any reason not to bors this?

karknu:
> Any reason not to bors this?

I'm waiting for either @dcoutts or @coot . Hoping for 2021.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -337,47 +339,63 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
+                (\txid _ -> mempoolHasTx txid)
+                txidsMap
+
+        availableTxidsU =
+              Map.filterWithKey
+                (\txid _ -> not $ elem txid (unacknowledgedTxIds st))
+                txidsMap
+
+        -- Divide the new txids in two: those that are already in the
+        -- mempool or in flight and those that are not. We'll request some txs from the
```
coot, line 370: This comment seems out of place now, I think it ought to be at the top of this hunk.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -337,47 +339,63 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
-        -- mempool and those that are not. We'll request some txs from the
+
+        (ignoredTxids, availableTxidsMp) =
+              Map.partitionWithKey
+                (\txid _ -> mempoolHasTx txid)
+                txidsMap
+
+        availableTxidsU =
+              Map.filterWithKey
+                (\txid _ -> not $ elem txid (unacknowledgedTxIds st))
```
coot, line 366: One could use `notElem` here.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -337,47 +339,63 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
 
         -- The txs that we intentionally don't request, because they are
         -- already in the mempool, need to be acknowledged.
         --
         -- So we extend bufferedTxs with those txs (so of course they have
         -- no corresponding reply).
-        bufferedTxs' = forceElemsToWHNF $
-                       bufferedTxs st
+        bufferedTxs' = bufferedTxs st
                     <> Map.map (const Nothing) ignoredTxids
```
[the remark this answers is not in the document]
karknu: Yes, `bufferedTxs` contains txs that are inflight. We don't want them overwritten with `Nothing` in case of a duplicate txid.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -337,47 +339,63 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
-          Seq.spanl (`Map.member` bufferedTxs') (unacknowledgedTxIds st)
+        (acknowledgedTxIds, unacknowledgedTxIds'') =
+          Seq.spanl (`Map.member` bufferedTxs') unacknowledgedTxIds'
 
-        -- If so we can remove acknowledged txs from our buffer
-        --
-        bufferedTxs'' = foldl' (flip Map.delete)
+
+        -- If so we can remove acknowledged txs from our buffer provided that they
+        -- are not still in unacknowledgedTxIds''. This happens in case of duplicate
```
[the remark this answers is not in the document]
karknu: `are still not` doesn't have the same meaning as `are not still`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -311,20 +307,26 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
+                                   bufferedTxs1 acknowledgedTxIds
+
+        _writtenTxids <- mempoolAddTxs txsReady
 
-        _ <- mempoolAddTxs txsReady
+        -- If we are acknowleding transactions that are still in unacknowledgedTxIds'
+        -- we need to re-add them so that we also can acknowledge them again later.
+        -- This will happen incase of duplicate txids within the same window.
+        let live = filter (`elem` unacknowledgedTxIds') $ toList acknowledgedTxIds
+            bufferedTxs3 = forceElemsToWHNF $ bufferedTxs2 <> (Map.fromList (zip live (repeat Nothing)))
```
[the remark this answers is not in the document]
karknu: Will move it to `bufferedTxs2`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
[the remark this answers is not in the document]
coot: I see, that's ok then.

PR #2721, opened 2020-11-02
ChainDB: let the BlockFetch client add blocks asynchronously

### LINKED ISSUE #2487: Chain DB: addBlock queue ineffective

mrBliss:
Fixes #2487.

Currently, the effective queue size when adding blocks to the ChainDB is 1 (for
why, see #2487). In this commit, we let the BlockFetch client add blocks fully
asynchronously to the ChainDB, which restores the effective queue size to the
configured value again, e.g., 10.

The BlockFetch client will no longer wait until the block has been written to
the VolatileDB (and thus also not until the block has been processed by chain
selection). The BlockFetch client can just hand over the block and continue
downloading with minimum delay. To make this possible, we change the behaviour
of `getIsFetched` and `getMaxSlotNo` to account for the blocks in the queue,
otherwise the BlockFetch client might try to redownload already-fetched blocks.

This is an alternative to #2489, which let the BlockFetch client write blocks to
the VolatileDB synchronously. The problem with that approach is that multiple
threads are writing to the VolatileDB, instead of a single background thread. We
have relied on the latter to simplify the VolatileDB w.r.t. consistency after
incomplete writes.


mrBliss:
Draft because the impact on bulk sync speed should be measured.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -146,18 +146,15 @@ getBlockComponent ::
...
-        case pointHash p of
-          BlockHash hash -> f hash
-          GenesisHash    -> False
+getIsFetched CDB{..} = do
+    checkBlocksToAdd <- memberBlocksToAdd cdbBlocksToAdd
+    checkVolDb       <- VolatileDB.getIsMember cdbVolatileDB
+    return $ \pt ->
+      case pointToWithOriginRealPoint pt of
+        Origin -> False
+        NotOrigin pt' -> checkBlocksToAdd pt' || checkVolDb (realPointHash pt')
```
mrBliss, line 157: Race condition: when a block has been removed but not yet written to the VolatileDB, this will return `False`. Possible solution: in the background thread, peek the block, process it, and then remove it from the queue.

edsko: Good thing _someone_ is paying attention :roll_eyes:

amesgen:

Niols:
Also, no work is lost because https://github.com/IntersectMBO/ouroboros-consensus/pull/1179 contains a variant of this fix.

PR #2726, opened 2020-11-03
network: refactor with Map.intersectionWith in Handshake server

nfrisby:
The code was using `List.intersect` to cope with the differing map codomains, but `Map.intersectionWith` seems more direct.

The comment caught my eye and prompted this quick PR. HTH.


nfrisby:
I'd guess that `lookupMax . intersectionWith (,)` does as little work as possible, but I'm not totally sure. Might be worthwhile to confirm that this diff doesn't introduce a bit more work in the presumably common case of the common version being within one or two.

coot:
I don't think that performance will ever be an issue here.  Using `Map.`intersectWith` is indeed nicer, and easier to grasp what's going on.

coot (approved):
LGTM

nfrisby:
I'm sorry for the delay; I was confused. My rebase just now to current `master` was automatic. borsing now. 👍 

bors r+

PR #2731, opened 2020-11-04
Update dependency on cardano-base

mrBliss:
Breaking change: https://github.com/input-output-hk/cardano-base/pull/179

We were using `VerKeyDSIGN MockDSIGN` as the `PBftVerKeyHash` for
`PBftMockCrypto`. Since `VerKeyDSIGN` no longer is allowed to implement `Ord`,
we can no longer derive `Ord` from the underlying `VerKeyDSIGN` type. Instead,
we derive `Ord` *via* the underlying `Word64` type.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/Crypto.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/Crypto.hs
@@ -1,7 +1,11 @@
+{-# LANGUAGE DeriveAnyClass         #-}
+{-# LANGUAGE DeriveGeneric          #-}
+{-# LANGUAGE DerivingVia            #-}
 {-# LANGUAGE FlexibleContexts       #-}
 {-# LANGUAGE FlexibleInstances      #-}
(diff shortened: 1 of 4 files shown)


nc6 (approved):

mrBliss:
bors merge

ouroboros-consensus/src/Ouroboros/Consensus/Protocol/PBFT/Crypto.hs
```diff
@@ -1,7 +1,11 @@
+{-# LANGUAGE DeriveAnyClass         #-}
+{-# LANGUAGE DeriveGeneric          #-}
+{-# LANGUAGE DerivingVia            #-}
 {-# LANGUAGE FlexibleContexts       #-}
 {-# LANGUAGE FlexibleInstances      #-}
 {-# LANGUAGE KindSignatures         #-}
 {-# LANGUAGE ScopedTypeVariables    #-}
+{-# LANGUAGE StandaloneDeriving     #-}
```
mrBliss, line 8: Oh, I didn't need this in the end :flushed:

PR #2732, opened 2020-11-04
Clarify chain selection

edsko:
Previously the `ChainSelection` class provided two method: `preferCandidate`, comparing a candidate chain to the current chain, and `compareCandidates`, comparing two candidate chains. However, there is no need for this generality: it suffices to say that _when comparing the current chain a candidate, and they are equally preferable, we prefer the current chain_. This is now defined once and for all outside of the class. In other words, `preferCandidate` is now defined in terms of `compareChains`, which replaces `compareCandidates` (which can be called with either two candidate chains, or with the current chain and a candidate).

We then do the same for `compareAnchoredCandidates`/`preferAnchoredCandidate`, but here the situation is a bit more subtle, and we need to change the precondition we rely on. The benefit however is that the new `compareAnchoredCandidates` is now entirely symmetric, simplifying the contract, and -- critically -- simplifying the formal description of chain selection in the Consensus Report.

### DIFF
--- a/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
+++ b/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs
@@ -795,7 +795,7 @@ validChains cfg m bs =
     sortChains $ chains bs
   where
     sortChains :: [Chain blk] -> [Chain blk]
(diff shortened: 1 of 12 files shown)


mrBliss (approved):
LGTM

mrBliss:
bors merge

PR #2733, opened 2020-11-05
Allow permanent, named ledger snapshots and number based on slot

mrBliss:
* Use the slot number of the ledger state as the number of the snapshot. This
  makes it more obvious to which block a snapshot corresponds.

  Note that we don't rely on these numbers being correct, we use them only to
  determine the order (descending) in which we try them when opening the ledger
  DB.

* Allow named, *permanent* snapshots. A snapshot can be given a name by
  appending a `_` and any string (even empty). Such snapshots are just like
  other automatically created snapshots, except that they will *never* be
  removed during a periodic trim of the snapshots (by default we only keep 2
  temporary snapshots around).

  For example, if you have a snapshot of the final Byron state lying around, you
  can call it, e.g., `4492799_last_Byron` and you don't have to worry about it
  being deleted.

* We no longer create a snapshot of the genesis ledger state. Not much point in
  doing that anyway.

### DIFF
--- a/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
+++ b/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
@@ -166,7 +166,7 @@ data Success ss =
     Unit ()
   | MaybeErr (Either (ExtValidationError TestBlock) ())
   | Ledger (ExtLedgerState TestBlock)
-  | Snapped (ss, Point TestBlock)
+  | Snapped (Maybe (ss, RealPoint TestBlock))
   | Restored (MockInitLog ss, ExtLedgerState TestBlock)
   deriving (Show, Eq, Functor, Foldable, Traversable)
 
@@ -178,22 +178,22 @@ newtype Resp ss = Resp (Success ss)
   Pure model
 -------------------------------------------------------------------------------}
 
--- The mock ledger records the blocks and ledger values (new to old)
+-- | The mock ledger records the blocks and ledger values (new to old)
 type MockLedger = [(TestBlock, ExtLedgerState TestBlock)]
 
--- | We simply enumerate snapshots
+-- | We use the slot number of the ledger state as the snapshot number
 --
 -- We only keep track of this to be able to give more meaningful statistics
 -- about generated tests. The mock implementation doesn't actually " take "
--- any snapshots (instead it stores the legder state at each point).
-newtype MockSnap = MockSnap Int
+-- any snapshots (instead it stores the ledger state at each point).
+newtype MockSnap = MockSnap Word64
   deriving (Show, Eq, Ord, Generic, ToExpr)
 
 -- | State of all snapshots on disk
 --
 -- In addition to the state of the snapshot we also record the tip of the chain
 -- at the time we took the snapshot; this is important for 'mockMaxRollback'.
-type MockSnaps = Map MockSnap (Point TestBlock, SnapState)
+type MockSnaps = Map MockSnap (RealPoint TestBlock, SnapState)
 
 -- | Mock implementation
 --
@@ -212,9 +212,6 @@ data Mock = Mock {
       -- See also 'applyMockLog', 'mockMaxRollback'.
     , mockRestore :: Point TestBlock
 
-      -- | Counter to assign 'MockSnap's
-    , mockNext    :: Int
-
       -- | Ledger DB params
       --
       -- We need the ledger DB params only to compute which snapshots the real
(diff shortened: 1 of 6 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -368,14 +409,37 @@ trimSnapshots tracer hasFS DiskPolicy{..} = do
   Internal: reading from disk
 -------------------------------------------------------------------------------}
 
--- | On disk snapshots are numbered monotonically
-newtype DiskSnapshot = DiskSnapshot Int
+data DiskSnapshot = DiskSnapshot {
+      -- | Snapshots are numbered. We will try the snapshots with the highest
+      -- number first.
+      --
+      -- When creating a snapshot, we use the slot number the ledger state it
```
mrBliss, line 416:
```suggestion
      -- When creating a snapshot, we use the slot number of the ledger state it
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -368,14 +409,37 @@ trimSnapshots tracer hasFS DiskPolicy{..} = do
...
+      -- corresponds to as the snapshot number. This gives an indication of how
+      -- recent the snapshot is.
+      --
+      -- Note that the snapshot names are only indicative, we don't rely on the
+      -- snapshot number matching the slot number of the corresponding ledger
+      -- state. We only use the snapshots numbers to determine the order in
+      -- which we try them. If a user chooses to rename their snapshots, we'll
```
mrBliss, line 423: Delete the sentence starting with "If ..", we already assume the user is not messing with the database.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -368,14 +409,37 @@ trimSnapshots tracer hasFS DiskPolicy{..} = do
...
+      -- just try them in a different order. That order might be suboptimal,
+      -- i.e., trying older snapshots requiring more replay first.
+      dsNumber :: Word64
+      -- | Snapshots can optionally have a suffix, separated by the snapshot
```
mrBliss, line 427: Blank line for Edsko.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -113,14 +116,20 @@ data StreamAPI m blk = StreamAPI {
...
-  -> (Point blk -> e)  -- ^ Error when tip not found
-  -> a                 -- ^ Starting point when tip /is/ found
-  -> (blk -> a -> m a) -- ^ Update function for each block
+  -> Point blk             -- ^ Starting point for streaming
+  -> (RealPoint blk -> e)  -- ^ Error when tip not found
+  -> a                     -- ^ Starting point when tip /is/ found
+  -> (blk -> a -> m a)     -- ^ Update function for each block
   -> ExceptT e m a
 streamAll StreamAPI{..} tip notFound e f = ExceptT $
     streamAfter tip $ \case
```
mrBliss, line 125: Let the `StreamAPI` return a `Left (RealPoint blk)` instead.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -374,16 +370,18 @@ runMock cmd initMock =
...
-        -- be dropped will become the new anchor. It is the anchor that gets
-        -- written to disk.
-        snapped :: Point TestBlock
-        snapped
+        -- The real implementation keeps @k@ snapshots in memory (provided that
+        -- there are enough blocks). The function 'ledgerDbCountToPrune'
+        -- computes how many (old) blocks/snapshots should be dropped to get
+        -- back into that range; the last (most recent) block to be dropped will
+        -- become the new anchor. It is the anchor that gets written to disk,
+        -- i.e., the ledger snapshot @k@ blocks back from the tip.
```
mrBliss, line 382: Say that `ledgerDbCountToPrune` tells you how many snapshots too many we have, excluding the anchor. Scratch that, just rewrite it without that function.

edsko (approved):
Nice. Who wrote that inscrutable test code :angry: :grimacing:

mrBliss:
bors merge

PR #2738, opened 2020-11-10
Cardano: generalise golden example injection using Nary

### LINKED ISSUE #2368: Generalise Isomorphic to more than one era

mrBliss:
Fixes #2368.

The new `Inject` class provides a convenient and uniform way to inject things
into the hard fork combinator, i.e., turn an `f blk` into an `f (HardForkBlock
xs)`.

Use the `Inject` class to write the inject the golden test examples from an era
into `HardForkBlock`.

Some golden test results involving `HardForkState` change because we use the
proper `Bound`s.


edsko (approved):
Glad that that injection stuff worked out so cleanly!

mrBliss:
bors merge

PR #2740, opened 2020-11-11
Golden tests for Allegra and Shelley

### LINKED ISSUE #2669: Golden tests for Allegra and Mary
When #2668 is done, we should think about which golden tests make sense for Allegra and Mary. Most things will be identical to the Shelley ones, though.

mrBliss:
Fixes #2669.

* Generalise the examples in `Test.Consensus.Shelley.Examples` to be parametric
  in the era where possible.

* Provide separate examples of types that are era-specific, i.e., `TxBody`. Turn
  the top-level set of all examples into an era-parametric function taking the
  `TxBody` as an argument.

* Use that function to create `Examples` records for Shelley, Allegra, and Mary.

Note that we don't have golden tests of Allegra and Mary as part of
`ouroboros-consensus-shelley-test`, only as part of
`ouroboros-consensus-cardano-test` as we only care about `CardanoBlock`.

Many examples have been decoupled, e.g., the example `HeaderHash` is no longer
computed from the example block, so that changes to the block won't cause the
hash to change anymore. This is why the golden test results have changed in this
commit.


edsko (approved):

mrBliss:
bors merge

mrBliss:
bors merge

PR #2746, opened 2020-11-16
Rudimentary diffusion socket logging

newhoggy:


mrBliss:
FYI, if this branch is just for yourself, you can just push the branch without opening a PR.

newhoggy:
I made some changes and am submitting a proper PR.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -38,28 +41,28 @@ import           Ouroboros.Network.Protocol.Handshake.Version
...
-import           Ouroboros.Network.Subscription.Ip
+import           Ouroboros.Network.NodeToNode
+                     (AcceptConnectionsPolicyTrace (..),
+                     AcceptedConnectionsLimit (..), DiffusionMode (..),
+                     NodeToNodeVersion (..), NodeToNodeVersionData,
+                     RemoteAddress)
+import qualified Ouroboros.Network.NodeToNode as NodeToNode
+import           Ouroboros.Network.Socket (ConnectionId (..),
+                     NetworkMutableState, NetworkServerTracers (..),
+                     cleanNetworkMutableState, newNetworkMutableState)
```
coot, line 55: Please preserve the original import style, and make sure the diff is minimal.  This will ensure that updating other branches when this is merged will be easy.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -79,25 +82,26 @@ data DiffusionTracers = DiffusionTracers {
...
       -- ^ Trace rate limiting of accepted connections
+    , dtSocketTracer           :: Tracer IO SocketTrace
     }
 
 
 -- | Network Node argumets
 --
 data DiffusionArguments = DiffusionArguments {
-      daIPv4Address  :: Maybe (Either Socket.Socket AddrInfo)
+      daIPv4Address              :: Maybe (Either Socket.Socket AddrInfo)
```
coot, line 92: the extra spaces seems unnecessary

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -277,7 +281,7 @@ runDataDiffusion tracers
         -- Return an IPv4 address with an emphemeral portnumber if we use IPv4
         anyIPv4Addr :: SockAddr -> Maybe SockAddr
         anyIPv4Addr (Socket.SockAddrInet _ _) = Just (Socket.SockAddrInet 0 0)
-        anyIPv4Addr _ = Nothing
+        anyIPv4Addr _                         = Nothing
```
coot, line 284: Please don't mix stylistic changes in the same commit, this will only make rebasing other branches harder.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -38,28 +41,28 @@ import           Ouroboros.Network.Protocol.Handshake.Version
...
+                     cleanNetworkMutableState, newNetworkMutableState)
 import           Ouroboros.Network.Subscription.Dns
+import           Ouroboros.Network.Subscription.Ip
 import           Ouroboros.Network.Subscription.Worker (LocalAddresses (..))
 import           Ouroboros.Network.Tracers
 
+data SocketTrace = SocketTrace
+  { socketTraceMessage :: Text
+  , socketTracePath    :: Maybe FilePath
+  } deriving (Eq, Show)
```
coot, line 64: The name is misleading, what about `DiffusionInitiatilizationTracer`, a bit long but self explaining (which will be important in a config file).

I think it's better to use a sum type instead of bare `Text`.  Then just looking at the constructors one can see what output it can generate.  A more detailed textuual representation should be defined in `cardano-node` (in the `Cardano.Tracing.OrphanedInstances.Network`).

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -154,9 +158,8 @@ runDataDiffusion tracers
                                     , daAcceptedConnectionsLimit
                                     , daDiffusionMode
                                     }
-                 applications@DiffusionApplications { daErrorPolicies } =
-    withIOManager $ \iocp -> do
-
+                 applications@DiffusionApplications { daErrorPolicies } = do
+  withIOManager $ \iocp -> do
```
coot, line 162: this also seems unnecessary

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +307,28 @@ runDataDiffusion tracers
                Left sd -> do
                    a <- Socket.getSocketName sd
                    case a of
-                        (Socket.SockAddrUnix path) ->
+                        (Socket.SockAddrUnix path) -> do
+                          traceWith dtSocketTracer $ SocketTrace "Creating local server socket for unix path" (Just path)
                           return (sd, Snocket.localSnocket iocp path)
                         _                          ->
                             -- TODO: This should be logged.
```
coot, line 314: Could you trace this event.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +307,28 @@ runDataDiffusion tracers
...
         )
         (\(sd,sn) -> Snocket.close sn sd) -- We close the socket here, even if it was provided for us.
         (\(sd,sn) -> do
 
           case daLocalAddress of
                Left _ -> pure () -- If a socket was provided it should be ready to accept
-               Right a -> do
-                 Snocket.bind sn sd $ Snocket.localAddressFromPath a
+               Right path -> do
+                 traceWith dtSocketTracer $ SocketTrace "Binding to socket for path: " (Just path)
```
coot, line 330: `Binding socket to path: ...` would sound better.

coot: Note: you can also log the socket file descriptor with `show sd` or if you just need the file descriptor you can use `socketFd sd :: IO CInt`.

coot (dismissed):

newhoggy:
Oops, sorry, I had `stylish-haskell` turned on.  Will revert the stylistic changes.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: I wasn't able to find `socketFd` so I went with `show`.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: Done!

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
coot: Sorry, it's `socketToFd`

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -60,6 +62,19 @@ import           Ouroboros.Network.Subscription.Dns
 import           Ouroboros.Network.Subscription.Worker (LocalAddresses (..))
 import           Ouroboros.Network.Tracers
 
+data DiffusionInitializationTracer
+  = DiffusionInitiatilizationRunServer
```
coot, line 66: I'd be in favour of shorter names, without `DiffusionInitialisation` prefix.  We can require that this module is imported qualified (that's my plan anyway).  The prefix will be in the logs because we log the tracer name, so no need to have it twice.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -60,6 +62,19 @@ import           Ouroboros.Network.Subscription.Dns
...
+  | DiffusionInitiatilizationRunLocalServer
+  | DiffusionInitiatilizationCreatingLocalServerSocketForUnixPath FilePath
```
coot, line 68: ^ I would call this `SystemdSocket`

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +319,30 @@ runDataDiffusion tracers
                Left sd -> do
                    a <- Socket.getSocketName sd
                    case a of
-                        (Socket.SockAddrUnix path) ->
+                        (Socket.SockAddrUnix path) -> do
+                          traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationCreatingLocalServerSocketForUnixPath path
                           return (sd, Snocket.localSnocket iocp path)
-                        _                          ->
+                        unsupportSocket -> do
                             -- TODO: This should be logged.
```
coot, line 326: You can remove this line now :)

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +319,30 @@ runDataDiffusion tracers
                Left sd -> do
                    a <- Socket.getSocketName sd
                    case a of
-                        (Socket.SockAddrUnix path) ->
+                        (Socket.SockAddrUnix path) -> do
+                          traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationCreatingLocalServerSocketForUnixPath path
                           return (sd, Snocket.localSnocket iocp path)
-                        _                          ->
+                        unsupportSocket -> do
```
coot, line 325:
```suggestion
                        unsupportedAddr -> do
```

`getSocketName` returns `SockAddr` not a socket.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +319,30 @@ runDataDiffusion tracers
...
                             -- TODO: This should be logged.
+                            traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationUnsupportedSocketType (show unsupportSocket)
```
coot, line 327: I would call this `UnsupportedSystemdLocalSocket (unsupportedAddr  :: SockAddr)`

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +319,30 @@ runDataDiffusion tracers
                Left sd -> do
                    a <- Socket.getSocketName sd
                    case a of
-                        (Socket.SockAddrUnix path) ->
+                        (Socket.SockAddrUnix path) -> do
+                          traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationCreatingLocalServerSocketForUnixPath path
```
coot, line 323:
```suggestion
                          traceWith dtDiffusionInitializationTracer $ CreatingLocalServerSocket a
```
By including `a :: SockAddr` the message we don't need to have `ForUnixPath` prefix.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +319,30 @@ runDataDiffusion tracers
...
                             -- TODO: This should be logged.
+                            traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationUnsupportedSocketType (show unsupportSocket)
                             throwIO UnsupportedLocalSocketType
 #endif
-               Right a -> do
-                   let sn = Snocket.localSnocket iocp a
-                   sd <- Snocket.open sn (Snocket.addrFamily sn $ Snocket.localAddressFromPath a)
+               Right path -> do
+                   let sn = Snocket.localSnocket iocp path
+                   traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationCreateLocalServerSocketForSnocketPath path
```
coot, line 332:
```suggestion
                   traceWith dtDiffusionInitializationTracer $ CreateLocalServerSocket path
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +319,30 @@ runDataDiffusion tracers
...
+                   sd <- Snocket.open sn (Snocket.addrFamily sn $ Snocket.localAddressFromPath path)
+                   traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationCreatedLocalServerSocketForSnocketPath path
```
coot, line 334:
```suggestion
                   traceWith dtDiffusionInitializationTracer $ CreatedLocalServerSocket path
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +319,30 @@ runDataDiffusion tracers
...
         )
         (\(sd,sn) -> Snocket.close sn sd) -- We close the socket here, even if it was provided for us.
         (\(sd,sn) -> do
 
           case daLocalAddress of
                Left _ -> pure () -- If a socket was provided it should be ready to accept
-               Right a -> do
-                 Snocket.bind sn sd $ Snocket.localAddressFromPath a
+               Right path -> do
+                 traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationBindingToSocket path (show sd)
```
coot, line 343:
```suggestion
                 traceWith dtDiffusionInitializationTracer $ BindingSocket (show sd) path
```

by including socket first it will be more natural to read the log message `BindingSocket <socket: 13> /some/path`

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,24 +319,30 @@ runDataDiffusion tracers
...
+                 Snocket.bind sn sd $ Snocket.localAddressFromPath path
+                 traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationListeningToSocket path (show sd)
```
coot, line 345:
```suggestion
                 traceWith dtDiffusionInitializationTracer $ ListeningSocket (show sd) path
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -336,20 +359,24 @@ runDataDiffusion tracers
          )
 
     runServer :: SocketSnocket -> NetworkMutableState SockAddr -> Either Socket.Socket SockAddr -> IO ()
-    runServer sn networkState address =
+    runServer sn networkState address = do
+      traceWith dtDiffusionInitializationTracer DiffusionInitiatilizationRunServer
```
coot, line 363: I would pus this message down, just before calling `withServer`. It would be nice to include the address in the message.  You can get it from `Either Socket SockAddr` using `getSocketName :: Socket -> IO SockAddr` (to get the address of the systemd socket).

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -336,20 +359,24 @@ runDataDiffusion tracers
...
       bracket
         (
           case address of
                Left sd -> return sd
-               Right a -> Snocket.open sn (Snocket.addrFamily sn a)
+               Right path -> do
+                 traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationCreatingServerSocket path
```
coot, line 369:
```suggestion
                 traceWith dtDiffusionInitializationTracer $ CreatingServerSocket addr
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -336,20 +359,24 @@ runDataDiffusion tracers
...
     runServer :: SocketSnocket -> NetworkMutableState SockAddr -> Either Socket.Socket SockAddr -> IO ()
-    runServer sn networkState address =
+    runServer sn networkState address = do
+      traceWith dtDiffusionInitializationTracer DiffusionInitiatilizationRunServer
       bracket
         (
           case address of
                Left sd -> return sd
-               Right a -> Snocket.open sn (Snocket.addrFamily sn a)
+               Right path -> do
```
coot, line 368:
```suggestion
               Right addr -> do
```

This is a more likely to be `IPv4` or `IPv6` address than a unix path.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -336,20 +359,24 @@ runDataDiffusion tracers
...
         )
         (Snocket.close sn) -- We close the socket here, even if it was provided for us.
         (\sd -> do
 
           case address of
                Left _ -> pure () -- If a socket was provided it should be ready to accept
-               Right a -> do
-                 Snocket.bind sn sd a
+               Right path -> do
+                 traceWith dtDiffusionInitializationTracer $ DiffusionInitiatilizationBindingServerSocket path
```
coot, line 378:
```suggestion
                 traceWith dtDiffusionInitializationTracer $ BindingServerSocket addr
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -336,20 +359,24 @@ runDataDiffusion tracers
...
+                 Snocket.open sn (Snocket.addrFamily sn path)
         )
         (Snocket.close sn) -- We close the socket here, even if it was provided for us.
         (\sd -> do
 
           case address of
                Left _ -> pure () -- If a socket was provided it should be ready to accept
-               Right a -> do
-                 Snocket.bind sn sd a
+               Right path -> do
```
coot, line 377:
```suggestion
               Right addr -> do
```

coot (dismissed):
A bit more nitpicking ;).

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
coot: All fields in this type should be strict.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: Looks like `socketToFd` is only available on `network-3.1.2.0`, but we are stuck on `network-3.1.0.1` for the moment.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
mrBliss: BTW, `type FilePath = String = [Char]`, so adding a bang won't guarantee that there are no thunks, it will only force the first `(:)` constructor, but *not* its arguments.

coot: For `FilePath` this is ok, as it will be evaluated by the network library; But this is the reason why turning other `String`s to `CInt` and use `socketToFd` to obtain the file descriptor number, rather than use `show socket` (which would never be evaluated by our code).

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -60,9 +62,22 @@ import           Ouroboros.Network.Subscription.Dns
 import           Ouroboros.Network.Subscription.Worker (LocalAddresses (..))
 import           Ouroboros.Network.Tracers
 
+data DiffusionInitializationTracer
+  = RunServer
+  | RunLocalServer
+  | CreatingSystemdSocketForUnixPath !FilePath
+  | CreateSystemdSocketForSnocketPath !FilePath
+  | CreatedSystemdSocketForSnocketPath !FilePath
+  | BindingToSocket !FilePath !String
```
coot, line 71: Use `CInt` for socket descriptor, you can get it using [socketToFd](https://hackage.haskell.org/package/network-3.1.2.0/docs/Network-Socket.html#v:socketToFd).  Then the show instance should show it as `<socket: 12>`.

coot: `BindingLocalSocket` would sound better for me.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -60,9 +62,22 @@ import           Ouroboros.Network.Subscription.Dns
...
+  | ListeningToSocket !FilePath !String
```
coot, line 72: What about `ListeningLocalSocket`: by calling `listen` the socket is in listening mode or listening socket is a common phrase too.

coot (changes requested):

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: Unfortunately, I don't have access to `socketToFd` because `ouroboros-network` is still using `network-3.1.0.1` but `socketToFd` was introduced in `network-3.1.2.0`

newhoggy: This PR shows that `network-3.1.2.0` does not work for this project on Windows: https://github.com/input-output-hk/ouroboros-network/pull/2767

karknu:
I think you should add the windows' `UnsupportedReadySocket` error too.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -291,7 +307,7 @@ runDataDiffusion tracers
     runLocalServer :: IOManager
                    -> NetworkMutableState LocalAddress
                    -> IO ()
-    runLocalServer iocp networkLocalState =
+    runLocalServer iocp networkLocalState = do
```
karknu, line 310: Is this `do` needed?

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -336,22 +358,26 @@ runDataDiffusion tracers
          )
 
     runServer :: SocketSnocket -> NetworkMutableState SockAddr -> Either Socket.Socket SockAddr -> IO ()
-    runServer sn networkState address =
+    runServer sn networkState address = do
```
karknu, line 361: and this `do`?

karknu (approved):
LGTM. Just some minor things.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -60,9 +62,22 @@ import           Ouroboros.Network.Subscription.Dns
...
+  = RunServer
+  | RunLocalServer
+  | CreatingSystemdSocketForUnixPath !FilePath
+  | CreateSystemdSocketForSnocketPath !FilePath
+  | CreatedSystemdSocketForSnocketPath !FilePath
+  | BindingLocalSocket !FilePath !String
+  | ListeningLocalSocket !FilePath !String
+  | CreatingServerSocket !SockAddr
+  | BindingServerSocket !SockAddr
+  | UnsupportedSystemdSocket !SockAddr
```
coot, line 75: This is ambiguous, what about
```suggestion
  | UnsupportedLocalSystemdSocket !SockAddr
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -60,9 +62,22 @@ import           Ouroboros.Network.Subscription.Dns
 import           Ouroboros.Network.Subscription.Worker (LocalAddresses (..))
 import           Ouroboros.Network.Tracers
 
+data DiffusionInitializationTracer
+  = RunServer
+  | RunLocalServer
+  | CreatingSystemdSocketForUnixPath !FilePath
+  | CreateSystemdSocketForSnocketPath !FilePath
+  | CreatedSystemdSocketForSnocketPath !FilePath
+  | BindingLocalSocket !FilePath !String
```
coot, line 71:
```suggestion
  | ConfiguringLocalSocket !FilePath !String
```
The `Snocket.bind` does not just `bind`s, it also configures the socket.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -60,9 +62,22 @@ import           Ouroboros.Network.Subscription.Dns
...
+  | ListeningLocalSocket !FilePath !String
+  | CreatingServerSocket !SockAddr
+  | BindingServerSocket !SockAddr
```
coot, line 74:
```suggestion
  | ConfiguringServerSocket !SockAddr
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -302,26 +318,32 @@ runDataDiffusion tracers
...
         )
         (\(sd,sn) -> Snocket.close sn sd) -- We close the socket here, even if it was provided for us.
         (\(sd,sn) -> do
 
           case daLocalAddress of
                Left _ -> pure () -- If a socket was provided it should be ready to accept
-               Right a -> do
-                 Snocket.bind sn sd $ Snocket.localAddressFromPath a
+               Right path -> do
+                 traceWith dtDiffusionInitializationTracer $ BindingLocalSocket path (show sd)
```
coot, line 341:
```suggestion
                 traceWith dtDiffusionInitializationTracer $ ConfiguringLocalSocket path (show sd)
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -336,22 +358,26 @@ runDataDiffusion tracers
...
         )
         (Snocket.close sn) -- We close the socket here, even if it was provided for us.
         (\sd -> do
 
           case address of
                Left _ -> pure () -- If a socket was provided it should be ready to accept
-               Right a -> do
-                 Snocket.bind sn sd a
+               Right addr -> do
+                 traceWith dtDiffusionInitializationTracer $ BindingServerSocket addr
```
coot, line 376:
```suggestion
                 traceWith dtDiffusionInitializationTracer $ ConfiguringServerSocket addr
```

coot (dismissed):
Please also change the use of `show socket`.  You can use this snippet:

```
-- | Socket file descriptor.
--
newtype SocketFD = SocketFD { getSocketFD :: CInt }

instance Show SocketFD where
  show fd = "<socket: " ++ show (getSocketFD fd) ++ ">"

socketFD :: Socket -> IO SocketFD
socketFD = fmap SocketFD . socketToFd
```
Only type constructor of `SocketFD` needs to be exported.  You should use it in `DiffusionInitializationTracer` instead of the `String` (where you use `show socket` pattern).

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: Fixed!

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -65,6 +67,26 @@ import           Ouroboros.Network.Subscription.Dns
...
+instance Show SocketFD where
+  show fd = "<socket: " ++ show (getSocketFD fd) ++ ">"
+
+data DiffusionInitializationTracer
+  = RunServer
+  | RunLocalServer
+  | CreatingSystemdSocketForUnixPath !FilePath
+  | CreateSystemdSocketForSnocketPath !FilePath
+  | CreatedSystemdSocketForSnocketPath !FilePath
+  | ConfiguringLocalSocket !FilePath !LocalSocket
```
coot, line 83: ^ I think you want to use `FileDescriptor` instead of `LocalSocket`.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -65,6 +67,26 @@ import           Ouroboros.Network.Subscription.Dns
...
+  | ListeningLocalSocket !FilePath !LocalSocket
```
coot, line 84: ^ The same here.

coot (changes requested):

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: Done!

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: Done!

newhoggy:
```
[macbook-:cardano.node.DiffusionInitializationTracer:Info:43] [2020-11-25 01:52:25.98 UTC] [String "test-059f56f3348ccd9e/socket/node-0",String "CreateSystemdSocketForSnocketPath"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:43] [2020-11-25 01:52:25.98 UTC] [String "test-059f56f3348ccd9e/socket/node-0",String "CreatedSystemdSocketForSnocketPath"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:47] [2020-11-25 01:52:25.98 UTC] [String "[::]:59250",String "CreatingServerSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:47] [2020-11-25 01:52:25.98 UTC] [String "[::]:59250",String "ConfiguringServerSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:43] [2020-11-25 01:52:25.98 UTC] [String "test-059f56f3348ccd9e/socket/node-0",String "<file-descriptor: 29>",String "ConfiguringLocalSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:46] [2020-11-25 01:52:25.98 UTC] [String "0.0.0.0:59250",String "CreatingServerSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:46] [2020-11-25 01:52:25.98 UTC] [String "0.0.0.0:59250",String "ConfiguringServerSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:47] [2020-11-25 01:52:25.98 UTC] [String "RunServer"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:46] [2020-11-25 01:52:25.98 UTC] [String "RunServer"]
```

newhoggy:
This is how it looks now:

```
[macbook-:cardano.node.DiffusionInitializationTracer:Info:43] [2020-11-25 01:52:25.98 UTC] [String "test-059f56f3348ccd9e/socket/node-0",String "CreateSystemdSocketForSnocketPath"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:43] [2020-11-25 01:52:25.98 UTC] [String "test-059f56f3348ccd9e/socket/node-0",String "CreatedSystemdSocketForSnocketPath"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:47] [2020-11-25 01:52:25.98 UTC] [String "[::]:59250",String "CreatingServerSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:47] [2020-11-25 01:52:25.98 UTC] [String "[::]:59250",String "ConfiguringServerSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:43] [2020-11-25 01:52:25.98 UTC] [String "test-059f56f3348ccd9e/socket/node-0",String "<file-descriptor: 29>",String "ConfiguringLocalSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:46] [2020-11-25 01:52:25.98 UTC] [String "0.0.0.0:59250",String "CreatingServerSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:46] [2020-11-25 01:52:25.98 UTC] [String "0.0.0.0:59250",String "ConfiguringServerSocket"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:47] [2020-11-25 01:52:25.98 UTC] [String "RunServer"]
[macbook-:cardano.node.DiffusionInitializationTracer:Info:46] [2020-11-25 01:52:25.98 UTC] [String "RunServer"]
```

ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
```diff
@@ -280,7 +280,7 @@ type LocalHandle = Socket
 #endif
 
 -- | System dependent LocalSnocket type
-newtype LocalSocket  = LocalSocket { getLocalHandle :: LocalHandle }
+newtype LocalSocket  = LocalSocket { getLocalHandle :: LocalHandle } deriving (Eq, Show)
```
coot, line 283: You shouldn't need those instance, but adding them is a good idea.  I'll add this to #2772.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -65,6 +69,26 @@ import           Ouroboros.Network.Subscription.Dns
 import           Ouroboros.Network.Subscription.Worker (LocalAddresses (..))
 import           Ouroboros.Network.Tracers
 
+-- | Socket file descriptor.
+--
+newtype SocketFD = SocketFD { getSocketFD :: CInt } deriving Eq
```
coot, line 74: You can remove `SocketFD` now, there's `FileDescriptor` for that purpose.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -310,14 +337,18 @@ runDataDiffusion tracers
                Left sd -> do
                    a <- Socket.getSocketName sd
                    case a of
-                        (Socket.SockAddrUnix path) ->
+                        (Socket.SockAddrUnix path) -> do
+                          traceWith dtDiffusionInitializationTracer $ CreatingSystemdSocketForUnixPath path
```
coot, line 344: `Snocket.localSnocket` is a pure operation which just wraps various socket / named pipes operation in a record, it cannot fail.  Still the message makes sense, but maybe it should be something like: `UsingSystemdSocket path`.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -310,14 +337,18 @@ runDataDiffusion tracers
...
-                        _unsupportedAddr ->
+                        unsupportedAddr -> do
+                          traceWith dtDiffusionInitializationTracer $ UnsupportedLocalSystemdSocket unsupportedAddr
                           throwIO UnsupportedLocalSocketType
 #endif
                Right addr -> do
                    let sn = Snocket.localSnocket iocp addr
+                   traceWith dtDiffusionInitializationTracer $ CreateSystemdSocketForSnocketPath addr
                    sd <- Snocket.open sn (Snocket.addrFamily sn $ Snocket.localAddressFromPath addr)
+                   traceWith dtDiffusionInitializationTracer $ CreatedSystemdSocketForSnocketPath addr
```
coot, line 354: This two messages should not mention `Systemd`, after all this socket is created here and not delivered by `systemd`. What about: `CreatedLocalSocket addr`.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -310,14 +337,18 @@ runDataDiffusion tracers
...
+                          traceWith dtDiffusionInitializationTracer $ CreatingSystemdSocketForUnixPath path
                           return (LocalSocket sd, Snocket.localSnocket iocp path)
-                        _unsupportedAddr ->
+                        unsupportedAddr -> do
+                          traceWith dtDiffusionInitializationTracer $ UnsupportedLocalSystemdSocket unsupportedAddr
                           throwIO UnsupportedLocalSocketType
 #endif
                Right addr -> do
                    let sn = Snocket.localSnocket iocp addr
+                   traceWith dtDiffusionInitializationTracer $ CreateSystemdSocketForSnocketPath addr
```
coot, line 352: think you are more interested to see a log message when `Snocket.open` fails, than searching for `Create` / `Created` messages.  If `Snocket.open` fails and throws an exception, `runLocalServer` will fail, and then `runDataDiffusion` will fail as well.  As a result the node will stop with an exception.  Isn't it enough? If you think so too, then I suggest we remove `CreateSystemdSocketForSnocketPath`.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -329,9 +360,12 @@ runDataDiffusion tracers
           case daLocalAddress of
                Left _ -> pure () -- If a socket was provided it should be ready to accept
                Right path -> do
+                 traceWith dtDiffusionInitializationTracer . ConfiguringLocalSocket path =<< localSocketFileDescriptor sd
```
coot, line 366: I never liked `=<<`, but it actually make sense here.  Maybe I should change my mind about it.

coot: Just break the line in two, it's too long, e.g.
```
traceWith dtDiffusionInitializationTracer . ConfiguringLocalSocket path
  =<< localSocketFileDescriptor sd
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -329,9 +360,12 @@ runDataDiffusion tracers
...
                  Snocket.bind sn sd $ Snocket.localAddressFromPath path
+                 traceWith dtDiffusionInitializationTracer . ListeningLocalSocket path =<< localSocketFileDescriptor sd
```
coot, line 368: Same here.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -65,6 +69,26 @@ import           Ouroboros.Network.Subscription.Dns
...
+-- | Socket file descriptor.
+--
+newtype SocketFD = SocketFD { getSocketFD :: CInt } deriving Eq
+
+instance Show SocketFD where
+  show fd = "<socket: " ++ show (getSocketFD fd) ++ ">"
+
+data DiffusionInitializationTracer
+  = RunServer
+  | RunLocalServer
```
coot, line 81: I would add `SockAddr` and `LocalAddress` on both `RunServer` and `RunLocalServer`. These are probably the only two messages that should be logged with `Notice` severity, all others are either `Info` or `Debug` messages so it makes sense to show the addresses on both of them.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -350,17 +384,21 @@ runDataDiffusion tracers
         (
           case address of
                Left sd -> return sd
-               Right a -> Snocket.open sn (Snocket.addrFamily sn a)
+               Right addr -> do
+                 traceWith dtDiffusionInitializationTracer $ CreatingServerSocket addr
```
coot, line 391: The same note as for the other `Creating...` message applies, e.g. if things go wrong the diffusion will error.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -350,17 +384,21 @@ runDataDiffusion tracers
...
+                 Snocket.open sn (Snocket.addrFamily sn addr)
         )
         (Snocket.close sn) -- We close the socket here, even if it was provided for us.
         (\sd -> do
 
           case address of
                Left _ -> pure () -- If a socket was provided it should be ready to accept
                Right addr -> do
+                 traceWith dtDiffusionInitializationTracer $ ConfiguringServerSocket addr
```
coot, line 400: Similarly for this message, if `bind` errors diffusion stops => node stops. The error should be logged by the node.

coot (commented):
A few more comments

ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
coot: I pushed this to #2772.  I used `deriving Show via Quiet LocalSocket` which will give us slightly nicer `Show` instance.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: I wasn't able to get a `SockAddr` or `LocalAddress` for these points of logging, so I used `FileDescriptor` instead.  Please advise.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: This is mostly because I don't trust that node will actually trust that stop with an exception under all circumstances.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
coot: For `runServer` you can do this:
```
          addr <- case address of
               Left sd -> getLocalName sn sd
               Right addr -> do
                 traceWith dtDiffusionInitializationTracer $ ConfiguringServerSocket addr
                 Snocket.bind sn sd addr
                 return addr
                 Snocket.listen sn sd

          traceWith dtDiffusionInitializationTracer (RunServer addr)
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
coot: Then wrap `runDataDiffusion` top level action with:
```
handle (\(e :: SomeExcpetion) -> withTracer dtDiffusionInitializationTracer (DiffusionErrored e) >> throwIO e $ do ... -- the runDataDiffusion logic follows.
```
You'll see not only that the diffusion broke, but also why.

You could test what node does, just before it calls `runDataDiffusion` you could call `throwIO (userError "HALT")`. For a longer term, we could use `io-sim-classes` package in the `node` package.  This would allow to testing node using `IOSim` and catch errors which you are hunting.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
coot: [Snocket haddoc documentation](https://input-output-hk.github.io/ouroboros-network/ouroboros-network-framework/Ouroboros-Network-Snocket.html#t:Snocket) could help you find the solution (though it could be improved).

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: The reason I hesitate is I don't know how all the code fits together.

For example:

* are threads forked somewhere, so the exception doesn't propagate to the top-level
* do exceptions get caught somewhere between here and there and don't get rethrown
* or maybe one of the socket functions block and never return
* or maybe there's an error that's not an exception, but it's ignored, so logging on an exception doesn't help

Not knowing the code well, I can't make these judgement calls.  I would be happy to remove any excess logging if they prove to be unfruitful when the failure occurs and the logging doesn't help.

But I your suggestion to add a top-level wrapper seems useful regardless.

coot: The sequence seems this: `main` calls `Cardano.Node.Run.runNode` which wraps `Ouroboros.Consensus.Node.run` in `withShutdownHandler`.  The latter call is using `race_` between the main action (`Ouroboros.Consensus.Node.run`) and some monitoring thread.  `Ouroboros.Consensus.Node.run` tail calls `Ouroboros.Network.Diffusion.runDataDiffusion`. All of them are done without using fork, threads are spawned by `Ouroboros.Consensus.Node.run` (for some consensus functionality) and `runDataDiffusion`.  The `main` installs `topLevelExceptionHandler`which seems fine.

`runDataDiffusion` is using `withAsync` to fork all threads and awaits on all asyncs using `waitAnyCancel` (using `waitAny` would be fine too).  It seems to me that exceptions are handled correctly up to `main`.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -208,6 +229,10 @@ runDataDiffusion tracers
                             ] ++ dnsSubThreads
 
   where
+    traceException :: IO a -> IO a
+    traceException f = catch f $ \(e :: SomeException) -> do
+      traceWith dtDiffusionInitializationTracer (DiffusionErrored e) 
+      throwIO e
```
coot, line 238: Please leave an empty line between definitions. We even use two empty lines at times, especially when we use a single empty line between different cases of the same definition.  This makes things much easier to read.

coot (approved):
Thanks, I left one layout comment, otherwise it can be merged.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
newhoggy: Done!

newhoggy:
bors merge

PR #2747, opened 2020-11-16
Throw 'ErrorCall's and 'FatalError's.

coot:
For `node-to-client` applications throw `ErrorCall` and `FatalError` (from `protolude`).

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
@@ -82,7 +82,9 @@ module Ouroboros.Network.NodeToClient (
   , HandshakeTr
   ) where
 
-import           Control.Exception (IOException)
+import           Cardano.Prelude (FatalError)
+
+import           Control.Exception (ErrorCall, IOException)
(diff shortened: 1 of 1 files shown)


ouroboros-network/src/Ouroboros/Network/NodeToClient.hs
```diff
@@ -584,6 +586,16 @@ networkErrorPolicies = ErrorPolicies
       , ErrorPolicy
           $ \(_ :: IOManagerError)
                 -> Just Throw
+
+        -- Using 'error' throws.
+      , ErrorPolicy
+          $ \(_ :: ErrorCall)
+                -> Just Throw
+
+        -- Using panic throws.
```
mrBliss, line 595: For consistency with the case above, I would:
```suggestion
        -- Using 'panic' throws.
```

coot: Done, thanks.

karknu (approved):
LGTM.

dcoutts (approved):
Good for now. Will be good to move to a different scheme for clients once we have time.

coot:
bors merge

PR #2753, opened 2020-11-17
Use a single set of credentials for all Shelley-based eras 

mrBliss:
We add the ability to share a single KES key by multiple Shelley-based eras.

Also, instead of a different thread per era, the hard fork combinator can now
combine `BlockForging` records from multiple eras into one `BlockForging` record
which picks the appropriate `BlockForging` based on the current era.

This means that the KES key shared by the Shelley-based eras will only be tried
to be evolved once, instead of once for each Shelley-based era (in the previous
approach, all threads would try to evolve it, but all but one would be an
identity operation). This should also improve tracing: instead of tracing
messages for each era, only messages of the current era are traced.

Note that multiple credentials *per era* (i.e. to benchmark multiple stakepools
on one node) will still result in multiple forging threads.

We refactor the protocol parameters for Shelley/Allegra/Mary to handle the
shared key: we now have a `ProtocolParamsShelleyBased` record which contains
everything needed for a Shelley-based chain, i.e., the genesis config and the
credentials. This record should be combined with the right era-specific ones.
This also paves the way for adding a Mary-only mode.


edsko (approved):
This was trickier than anticpated!

mrBliss:
bors merge

PR #2754, opened 2020-11-17
Add a pure Mary mode

mrBliss:
Note that this mode is not tested by consensus like the others, i.e., we only
test the Mary era as part of the Cardano blockchain, not on its own.


mrBliss:
Draft because this depends on #2753.

All commits in this branch except for the last were created for other reasons. Only the last commit was needed to add a pure Mary mode. @deepfire as you can see, the changes in consensus are trivial. All the hard work will have to be done in the node by yourself or the node team.

edsko (approved):

mrBliss:
bors merge

deepfire:
Thank you @mrBliss !

PR #2768, opened 2020-11-19
Restructure Github Actions

newhoggy:
This restructures the Github Actions so that all the building happens first before any of the tests are run.

This is beneficial because previously a test failure would fail CI before we even know all the projects build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,10 +82,6 @@ jobs:
         sudo apt-get -y remove --purge software-properties-common
         sudo apt-get -y autoremove
 
-    - name: Cabal config
-      run: |
-        cabal user-config init -a "http-transport: plain-http" -a "store-dir: C:\\SR" -f -v3
-
     - name: Set up temp directory
       run: |
         echo 'TMPDIR=${{ runner.temp }}'  >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -121,108 +117,84 @@ jobs:
...
       uses: actions/checkout@v2
 
+    # Build dependencies
+    - name: Build dependencies
+      run: cabal --builddir="$CABAL_BUILDDIR" configure --enable-tests
+
     - name: Use cabal.project.local.windows
-      run: cp ./cabal.project.local.ci.windows ./cabal.project.local
+      run: |
+        cat ./cabal.project.local.ci.windows >> ./cabal.project.local
```
coot, line 126: Is there something wrong with `cp`?

.github/workflows/build.yml
```diff
@@ -121,108 +117,84 @@ jobs:
     - name: checkout ouroboros-network repository
       uses: actions/checkout@v2
 
+    # Build dependencies
+    - name: Build dependencies
+      run: cabal --builddir="$CABAL_BUILDDIR" configure --enable-tests
```
coot, line 122: Where does `$CABAL_BUILDDIR` is set?

coot (commented):
LGTM, just some minor questions / comments.

.github/workflows/build.yml
newhoggy: It is set by the step on line `22`

.github/workflows/build.yml
newhoggy: I do a `configure` in he previous step, so here I'm appending to whatever `local` file that `configure` produced.

coot: I see, thanks for pointing this out.

coot (approved):
LGTM

newhoggy:
bors merge

PR #2777, opened 2020-11-27
socketFileDescriptor: use unsafeFdSocket

coot:
Even if socket is garbage collected / closed and the file descriptor is reused by the system, `FileDescriptor` type does not make it easy to be used with any low level functions in `base` which take file descriptor as an argument, so one should not be able to act on a wrong file descriptor.  See [unsafeFdSocket](https://hackage.haskell.org/package/network-3.1.2.1/docs/Network-Socket.html#v:unsafeFdSocket).


karknu (approved):
LGTM.

coot:
bors merge

PR #2778, opened 2020-11-27
Remove ChainFragment

mrBliss:
`ChainFragment`s were hardly ever used anymore, only in some tests. We use
`AnchoredFragment`s almost everywhere. There are very few situations in which
you *don't* want an anchor.

This means that `ChainFragment`s are only really there to make implementing
`AnchoredFragment`s easier, i.e., by wrapping them. However, we can simply
inline them, dropping tons of code in the process.

In the process, simplify the chain and fragment generators. Drop a few pattern
synonyms. Stop using block number 0 for genesis and remove `legacyHeadBlockNo`.


mrBliss:
I added the `prop_sliceRange` that I forgot about. There were no other tests for the `ChainFragment` that needed to be ported over.

edsko (approved):
Nice, good that cleaned up a bit.

mrBliss:
bors merge

PR #2779, opened 2020-11-27
db-converter: drop path and path-io dependencies

mrBliss:
Since we use `canonicalizePath`, we know we are dealing with absolute paths so
there is not much confusion anymore. I don't think this is worth the two extra
dependencies that are not used anywhere else.


edsko (approved):

mrBliss:
bors merge

PR #2783, opened 2020-12-01
Capture the logic for tryAddTxs in a pure function + data type

mrBliss:
Represent the control flow explicitly using the `TryAddTxs` data type, returned
by `pureTryAddTxs`.

Provide a pure interpreter for this control flow, `runTryAddTxs`, which can be
used in the tests.

Provide a concurrent, STM-based interpreter for this control flow,
`implTryAddTxs`, which is the real implementation.


edsko (approved):
Nice, separating out the logic from the effect. I like it.

PR #2784, opened 2020-12-01
Support for getting peers from ledger (not used)

karknu:
Support for randomly picking peers from the current ledger state.
The peers are randomly choosen from the relays registered on chain, and
the chance to pick a given stake pools relay is weighted by the amount
of stake they control.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -21,6 +21,7 @@ module Ouroboros.Consensus.NodeKernel (
   , getMempoolReader
   , getMempoolWriter
   , getPeersFromCurrentLedger
+  , getPeersFromCurrentLedgerAfterSlot
   ) where
 
 import           Control.Monad
@@ -693,3 +694,22 @@ getPeersFromCurrentLedger kernel p = do
       return
         $ map (second (fmap stakePoolRelayAddress))
         $ getPeers immutableLedger
+
+-- | Like 'getPeersFromCurrentLedger' but with a \"after slot number X\" condition.
+getPeersFromCurrentLedgerAfterSlot ::
+     forall blk localPeer m remotePeer.
+     ( IOLike m
+     , LedgerSupportsPeerSelection blk
+     , UpdateLedger blk
+     )
+  => NodeKernel m remotePeer localPeer blk
+  -> SlotNo
+  -> STM m (Maybe [(PoolStake, NonEmpty RelayAddress)])
+getPeersFromCurrentLedgerAfterSlot kernel slotNo =
+    getPeersFromCurrentLedger kernel afterSlotNo
+  where
+    afterSlotNo :: LedgerState blk -> Bool
+    afterSlotNo st =
+      case ledgerTipSlot st of
+        Origin        -> False
+        NotOrigin tip -> tip > slotNo
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
@@ -1,0 +1,145 @@
+{-# LANGUAGE BangPatterns               #-}
+{-# LANGUAGE GADTs                      #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE RecordWildCards            #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+
+module Ouroboros.Network.PeerSelection.LedgerPeers (
+    DomainAddress (..),
+    IP.IP (..),
+    LedgerPeersConsensusInterface (..),
+    RelayAddress (..),
+    PoolStake (..),
+    AccPoolStake (..),
+    TraceLedgerPeers (..),
+    pickPeers,
+    accPoolStake,
+
+    Socket.PortNumber
+    ) where
+
+
+import           Control.Monad.Class.MonadSTM
+import           Control.Tracer (Tracer, traceWith)
+import qualified Data.IP as IP
+import           Data.List (foldl')
+import           Data.List.NonEmpty (NonEmpty (..))
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Map.Strict as Map
+import           Data.Map.Strict (Map)
+import           Data.Ratio
+import           Data.Word
+import qualified Network.Socket as Socket
+import           System.Random
+
+import           Cardano.Slotting.Slot (SlotNo)
+import           Ouroboros.Network.PeerSelection.RootPeersDNS (DomainAddress (..))
+
+import           Text.Printf
+
+newtype LedgerPeersConsensusInterface m = LedgerPeersConsensusInterface {
+      lpGetPeers :: SlotNo -> STM m (Maybe [(PoolStake, NonEmpty RelayAddress)])
+    }
+
+-- | Trace LedgerPeers events.
+data TraceLedgerPeers =
+      PickedPeer !RelayAddress !AccPoolStake ! PoolStake
+      -- ^ Trace for a peer picked with accumulated and relative stake of its pool.
+    | PickedPeers !Word16 ![RelayAddress]
+      -- ^ Trace for the number of peers we wanted to pick and the list of peers picked.
+    | FetchingNewLedgerState !Int
+      -- ^ Trace for fetching a new list of peers from the ledger. Int is the number of peers
+      -- returned.
+
+
+instance Show TraceLedgerPeers where
+    show (PickedPeer addr ackStake stake) =
+        printf "PickedPeer %s ack stake %s ( %.04f) relative stake %s ( %.04f )"
+            (show addr)
+            (show $ unAccPoolStake ackStake)
+            (fromRational (unAccPoolStake ackStake) :: Double)
+            (show $ unPoolStake stake)
+            (fromRational (unPoolStake stake) :: Double)
+    show (PickedPeers n peers) =
+        printf "PickedPeers %d %s" n (show peers)
+    show (FetchingNewLedgerState cnt) =
+        printf "Fetching new ledgerstate, %d registered pools"
+            cnt
+
+
+data RelayAddress = RelayAddressDomain DomainAddress
+                  | RelayAddressAddr IP.IP Socket.PortNumber
+                  deriving (Show, Eq, Ord)
+
+-- | The relative stake of a stakepool in relation to the total amount staked.
+-- A value in the [0, 1] range.
+--
+newtype PoolStake = PoolStake { unPoolStake :: Rational }
+  deriving (Eq, Fractional, Num, Ord, Show)
+
+-- | The accumulated relative stake of a stake pool, like PoolStake but it also includes the
+-- relative stake of all preceding pools. A value in the range [0, 1].
+--
+newtype AccPoolStake = AccPoolStake { unAccPoolStake :: Rational }
+    deriving (Eq, Num, Ord)
+
+-- | Convert a list of pools with stake to a Map keyed on the accumulated stake.
+-- Consensus provides a list of pairs of relative stake and corresponding relays for all usable
+-- registered pools.
+-- By creating a Map keyed on the `AccPoolStake` that is the sum of the pool's relative stake and
+-- the stake of all preceding pools we can support weighted random selection in
+-- O(log n) time by taking advantage of Map.lookupGE (returns the smallest key greater or equal
+-- to the provided value).
+--
+accPoolStake :: [(PoolStake, NonEmpty RelayAddress)]
+             -> Map AccPoolStake (PoolStake, NonEmpty RelayAddress)
+accPoolStake pl =
+    let pl' = reRelativeStake pl
+        ackList = foldl' fn [] pl' in
+    Map.fromList ackList
+  where
+    fn :: [(AccPoolStake, (PoolStake, NonEmpty RelayAddress))]
+       -> (PoolStake, NonEmpty RelayAddress)
+       -> [(AccPoolStake, (PoolStake, NonEmpty RelayAddress))]
+    fn [] (s, rs) =
+        [(AccPoolStake (unPoolStake s), (s, rs))]
+    fn ps (s, !rs) =
+        let accst = AccPoolStake (unPoolStake s)
+            as = fst $ head ps
+            !acc = as + accst in
+        (acc, (s, rs)) : ps
+
+-- | Not all stake pools have valid \/ usable relay information. This means that we need to
+-- recalculate the relative stake for each pool.
+--
+reRelativeStake :: [(PoolStake, NonEmpty RelayAddress)]
+                -> [(PoolStake, NonEmpty RelayAddress)]
+reRelativeStake pl =
+    let total = sum $ map fst pl in
+    map (\(s, rls) -> (s / total, rls)) pl
+
+-- | Try to pick n random peers.
+pickPeers :: forall m. Monad m
+          => StdGen
+          -> Tracer m TraceLedgerPeers
+          -> Map AccPoolStake (PoolStake, NonEmpty RelayAddress)
+          -> Word16
+          -> m (StdGen, [RelayAddress])
+pickPeers inRng _ pools _ | Map.null pools = return (inRng, [])
+pickPeers inRng tracer pools cnt = go inRng cnt []
+  where
+    go :: StdGen -> Word16 -> [RelayAddress] -> m (StdGen, [RelayAddress])
+    go rng 0 picked = return (rng, picked)
+    go rng n picked =
+        let (r :: Word64, rng') = random rng
+            d = maxBound :: Word64
+            x = fromIntegral r % fromIntegral d in
+        case Map.lookupGE (AccPoolStake x) pools of
+             -- XXX We failed pick a peer. Shouldn't this be an error?
+             Nothing -> go rng' (n - 1) picked 
+             Just (ackStake, (stake, relays)) -> do
+                 let (ix, rng'') = randomR (0, NonEmpty.length relays - 1) rng'
+                     relay = relays NonEmpty.!! ix
+                 traceWith tracer $ PickedPeer relay ackStake stake
+                 go rng'' (n - 1) (relay : picked)
--- a/ouroboros-network/test/Test/LedgerPeers.hs
+++ b/ouroboros-network/test/Test/LedgerPeers.hs
@@ -1,0 +1,162 @@
+{-# LANGUAGE BangPatterns      #-}
+{-# LANGUAGE NamedFieldPuns    #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes        #-}
+
+module Test.LedgerPeers where
+
+import           Control.Exception (SomeException (..))
+import           Control.Monad.Class.MonadAsync
+import           Control.Monad.Class.MonadFork
+import           Control.Monad.Class.MonadSay
+import           Control.Monad.Class.MonadTime
+import           Control.Monad.Class.MonadThrow
+import           Control.Monad.IOSim
+import           Control.Tracer (showTracing, Tracer (..), traceWith)
+import           Data.List (foldl', intercalate)
+import           Data.List.NonEmpty (NonEmpty (..))
+import           Data.Word
+import           Data.Ratio
+import           System.Random
+
+import           Ouroboros.Network.PeerSelection.LedgerPeers
+
+import           Test.Tasty (TestTree, testGroup)
+import           Test.Tasty.QuickCheck (testProperty)
+import           Test.QuickCheck
+import           Text.Printf
+
+tests :: TestTree
+tests = testGroup "LedgerPeers"
+  [ testProperty "Pick 100%" prop_pick100
+  , testProperty "Pick" prop_pick
+  ]
+
+newtype ArbitraryRelayAddress = ArbitraryRelayAddress RelayAddress
+
+instance Arbitrary ArbitraryRelayAddress where
+    arbitrary = do
+      ArbitraryRelayAddress <$> elements [ RelayAddressAddr (read "1.1.1.1") 1234
+               , RelayAddressDomain (DomainAddress "relay.iohk.example" 1234)
+               ]
+
+data StakePool = StakePool {
+      spStake :: !Word64
+    , spRelay :: NonEmpty RelayAddress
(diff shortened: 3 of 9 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,188 @@
...
+import           Ouroboros.Network.Subscription.Ip
+import           Ouroboros.Network.Subscription.Dns
+
+import           Text.Printf
+
+newtype LedgerPeersConsensusInterface m = LedgerPeersConsensusInterface {
+      lpGetPeers :: SlotNo -> STM m (Maybe [(PoolStake, NonEmpty RelayAddress)])
+    }
+
+data TraceLedgerPeers =
```
coot, line 50: Could you add haddocks, including to each constructor, e..g. what is the `Word16` in `PickedPeers`, or the `Int` in `FetchingNewlEdgerState`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,188 @@
...
+
+data RelayAddress = RelayAddressDomain DomainAddress
+                  | RelayAddressAddr IP.IP Socket.PortNumber
+                  deriving (Show, Eq, Ord)
+
+-- | The relative stake of the stakepool. A value in the [0, 1] range.
+type PoolStake = Rational
+
+-- | The relative stake of the stakepool and all preceeding pools. A value in the range [0, 1].
+type AckPoolStake = Rational
```
coot, line 77: Wouldn't it be safer to have two `newyptes` instead of two types?

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,188 @@
...
+
+-- | Convert a list of pools with stake to a Map keyed on the ackumulated stake.
+--
+ackPoolStake :: [(PoolStake, NonEmpty RelayAddress)]
+             -> Map Rational (PoolStake, NonEmpty RelayAddress)
+ackPoolStake pl =
+    let pl' = reRelativeStake pl
+        ackList = snd $ foldl' (\(as, ps) (s, rs) -> (as + s, (as + s, (s, rs)):ps)) (0, []) pl' in
```
coot, line 85: ^ the problem with `foldl'` and deep tuples is that tuples are lazy, so strictness of `foldl'` does not help much.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,188 @@
...
+pickPeers inRng _ pools _ | Map.null pools = return (inRng, [])
+pickPeers inRng tracer pools cnt = go inRng cnt []
+  where
+    go :: StdGen -> Word16 -> [RelayAddress] -> m (StdGen, [RelayAddress])
+    go rng 0 picked = return (rng, picked)
+    go rng n picked =
+        let (r :: Word64, rng') = random rng
+            d = maxBound :: Word64
+            x = fromIntegral r % fromIntegral d
+            !pick_m = Map.lookupGE x pools in
```
coot, line 112: No need for a bang, if you write:
```
case Map.lookupGE x pools of
  ...
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,188 @@
...
+        case pick_m of
+             Nothing -> go rng' (n - 1) picked -- XXX We failed pick a peer. Shouldn't this be an error?
+             Just (ackStake, (stake, relays)) -> do
+                 let (ix, rng'') = randomR (0, NonEmpty.length relays - 1) rng'
+                     relay = relays NonEmpty.!! ix
```
coot, line 117: We might pick the same `relay` multiple times, is that ok?

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,188 @@
...
+        let (ipTarget, dnsTargets) = foldl' peersToSubTarget
+                                                 (IPSubscriptionTarget [] 0 , []) $ nub pickedPeers
+        ipAid <- async $ runIP ipTarget
+
+        dnsAids <- sequence [async $ runDns peer | peer <- dnsTargets]
+
+        -- let it run for 10 minutes, then repeat with a new set of random peers.
+        ttlAid <- async (threadDelay 600)
+
+        _ <- waitAnyCancel (ttlAid : ipAid : dnsAids)
```
coot, line 168: It would be nicer to use `race_`, or
```
withAsync (runIP ..) $ \ippAid ->
  withAsyncs ([runDns peer | peer <- dnsTargets]) $ \dnsAids ->
    race_ (threadDelay 600)
               (waitAny ...)
```
This is safe if an async exception hits the thread, or if any of the thread errors.
I think we have `withAsyncs` somewhere in the code.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
karknu: I think that should be fine, there could also be overlap due to different domain names resolving to the same address. We may also pick a peer that the governor already knows about, or the peer may be unreachable.
The governor asks LedgerPeers for some fresh IP addresses, if some of them turn out to be unsuitable it can simply ask again.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
karknu: Yes, it would be bad if they where mixed up.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
karknu: Would the insertion of them into a strict Map below force their evaluation?

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
karknu: I agree that `withAsync` is the proper way to do this. But this code is unused and the connection-manager aware version will not  start any IP or DNS subscription works.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -693,3 +694,22 @@ getPeersFromCurrentLedger kernel p = do
       return
         $ map (second (fmap stakePoolRelayAddress))
         $ getPeers immutableLedger
+
+-- | Like `getPeersFromCurrentLedger` but with a "after slot number X" condition.
```
mrBliss, line 698: The following is valid Haddock:
```suggestion
-- | Like 'getPeersFromCurrentLedger' but with a \"after slot number X\" condition.
```

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -693,3 +694,22 @@ getPeersFromCurrentLedger kernel p = do
...
+     forall blk localPeer m remotePeer.
+     ( IOLike m
+     , LedgerSupportsPeerSelection blk
+     , UpdateLedger blk
+     )
+  => NodeKernel m remotePeer localPeer blk
+  -> SlotNo
+  -> STM m (Maybe [(PoolStake, NonEmpty RelayAddress)])
+getPeersFromCurrentLedgerAfterSlot kernel slotNo =
+  getPeersFromCurrentLedger kernel afterSlotNo
```
mrBliss, line 709:
```suggestion
    getPeersFromCurrentLedger kernel afterSlotNo
```

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -693,3 +694,22 @@ getPeersFromCurrentLedger kernel p = do
...
+  where
+    afterSlotNo :: LedgerState blk -> Bool
+    afterSlotNo st =
+      case ledgerTipSlot st of
+        Origin        -> False
+        NotOrigin tip -> tip >  slotNo
```
mrBliss, line 715:
```suggestion
      case ledgerTipSlot st of
        Origin        -> False
        NotOrigin tip -> tip > slotNo
```

mrBliss (commented):
Consensus changes look good to me, except for some stylistic issues.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,203 @@
...
+        printf "Fetching new ledgerstate, %d registered pools"
+            cnt
+
+
+data RelayAddress = RelayAddressDomain DomainAddress
+                  | RelayAddressAddr IP.IP Socket.PortNumber
+                  deriving (Show, Eq, Ord)
+
+-- | The relative stake of the stakepool. A value in the [0, 1] range.
+newtype PoolStake = PoolStake { unPoolStake :: Rational } deriving (Eq, Fractional, Num, Ord, Show)
```
coot, line 83: Let's keep line short for those who split screens or use two screens;
```suggestion
newtype PoolStake = PoolStake { unPoolStake :: Rational }
  deriving (Eq, Fractional, Num, Ord, Show)
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,203 @@
...
+
+-- | The relative stake of the stakepool and all preceeding pools. A value in the range [0, 1].
+newtype AckPoolStake = AckPoolStake { unAckPoolStake :: Rational } deriving (Eq, Num, Ord)
```
coot, line 86:
```suggestion
newtype AckPoolStake = AckPoolStake { unAckPoolStake :: Rational }
    deriving (Eq, Num, Ord)
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,203 @@
...
+ackPoolStake pl =
+    let pl' = reRelativeStake pl
+        ackList = snd $ foldl' fn (AckPoolStake 0, []) pl' in
+    Map.fromList ackList
+  where
+    fn :: (AckPoolStake, [(AckPoolStake, (PoolStake, NonEmpty RelayAddress))])
+       -> (PoolStake, NonEmpty RelayAddress)
+       -> (AckPoolStake, [(AckPoolStake, (PoolStake, NonEmpty RelayAddress))])
+    fn (as, ps) (s, rs) =
+        let ackst = AckPoolStake (unPoolStake s) in
```
coot, line 101: Nice, I like that it's now visible that we coerce `PoolStake` to `AckPoolStake`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,203 @@
...
+    show (FetchingNewLedgerState cnt) =
+        printf "Fetching new ledgerstate, %d registered pools"
+            cnt
+
+
+data RelayAddress = RelayAddressDomain DomainAddress
+                  | RelayAddressAddr IP.IP Socket.PortNumber
+                  deriving (Show, Eq, Ord)
+
+-- | The relative stake of the stakepool. A value in the [0, 1] range.
```
coot, line 82:
```suggestion
-- | The relative stake of the stakepool. A value in the [0, 1] range.
--
```
Just for readability sake.

It's relative to what? Stake of all the pools on the chain? I'd be explicit about it.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,203 @@
...
+newtype PoolStake = PoolStake { unPoolStake :: Rational } deriving (Eq, Fractional, Num, Ord, Show)
+
+-- | The relative stake of the stakepool and all preceeding pools. A value in the range [0, 1].
```
coot, line 85:
```suggestion
-- | The relative stake of the stakepool and all preceeding pools. A value in the range [0, 1].
--
```

As above, I would be explicit about relative to what? And I'd explain how this differs from `PoolStake`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
coot: I would change the type prefix of the type from `Ack` to `Acc` as in `Accumulatae`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,203 @@
...
+  where
+    fn :: (AckPoolStake, [(AckPoolStake, (PoolStake, NonEmpty RelayAddress))])
+       -> (PoolStake, NonEmpty RelayAddress)
+       -> (AckPoolStake, [(AckPoolStake, (PoolStake, NonEmpty RelayAddress))])
+    fn (as, ps) (s, rs) =
+        let ackst = AckPoolStake (unPoolStake s) in
+        (as + ackst, (as + ackst, (s, rs)) : ps)
+
+-- | Not all stake pools have valid/usable relay information. This means that we need to
+-- recalculate the relative stake for each pool.
```
coot, line 105:
```suggestion
-- | Not all stake pools have valid \/ usable relay information. This means that we need to
-- recalculate the relative stake for each pool.
--
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
coot: That's ok, thanks for clarification.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,203 @@
...
+pickPeers inRng tracer pools cnt = go inRng cnt []
+  where
+    go :: StdGen -> Word16 -> [RelayAddress] -> m (StdGen, [RelayAddress])
+    go rng 0 picked = return (rng, picked)
+    go rng n picked =
+        let (r :: Word64, rng') = random rng
+            d = maxBound :: Word64
+            x = fromIntegral r % fromIntegral d in
+        case Map.lookupGE (AckPoolStake x) pools of
+             Nothing -> go rng' (n - 1) picked -- XXX We failed pick a peer. Shouldn't this be an error?
```
coot, line 129:
```suggestion
             -- XXX We failed pick a peer. Shouldn't this be an error?
             Nothing -> go rng' (n - 1) picked 
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,203 @@
...
+             Just (ackStake, (stake, relays)) -> do
+                 let (ix, rng'') = randomR (0, NonEmpty.length relays - 1) rng'
+                     relay = relays NonEmpty.!! ix
+                 traceWith tracer $ PickedPeer relay ackStake stake
+                 go rng'' (n - 1) (relay : picked)
+
+-- | Example of how how peers from ledger could be used.
+-- In reality it will be implemented using the new connection-manager and governor.
```
coot, line 137:
```suggestion
-- | Example of how how peers from ledger could be used.
-- In reality it will be implemented using the new connection-manager and governor.
--
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
coot: I see your point, but I won't feel good merging this to master;  if I'd be a reviewer and stumbled upon such code, I would have more reservations for the rest of reviewing work.   There's an an easy fix too.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
coot: No, that's the same problem.  Haskell evaluates to weak head normal form, which means it will not pass through a lazy constructor.  You need to use `seq` or `!` to force the accumulators, e.g
```
foldl' (\(as, ps) (s, rs) -> 
            let !as' = as + s
                !ps' = (s, rs) : ps
            in (as', (as' , ps'))
        (0, []) pl'
```
And we also increased sharing by using `as'` twice.

coot: why do we need to use `as'` twice in the result?

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
karknu: The offending code has been removed.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
karknu: We don't need to use `as'` twice, if the folding function written so that it can handle the `[]`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
coot: :100:

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,140 @@
...
+    let pl' = reRelativeStake pl
+        ackList = foldl' fn [] pl' in
+    Map.fromList ackList
+  where
+    fn :: [(AccPoolStake, (PoolStake, NonEmpty RelayAddress))]
+       -> (PoolStake, NonEmpty RelayAddress)
+       -> [(AccPoolStake, (PoolStake, NonEmpty RelayAddress))]
+    fn [] (s, rs) =
+        [(AccPoolStake (unPoolStake s), (s, rs))]
+    fn ps (s, rs) =
```
coot, line 101:
```suggestion
    fn ps (s, !rs) =
```
`s` is evaluated when we force `acc`, `rs` isn't.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -0,0 +1,140 @@
...
+-- | The accumulated relative stake of a stake pool, like PoolStake but it also includes the
+-- relative stake of all preceding pools. A value in the range [0, 1].
+--
+newtype AccPoolStake = AccPoolStake { unAccPoolStake :: Rational }
+    deriving (Eq, Num, Ord)
+
+-- | Convert a list of pools with stake to a Map keyed on the accumulated stake.
+--
+accPoolStake :: [(PoolStake, NonEmpty RelayAddress)]
+             -> Map AccPoolStake (PoolStake, NonEmpty RelayAddress)
```
coot, line 90: I don't understand the significance of the order in `[(PoolStake, NonEmpty RelayAddress)]`.  The more to the right of the list is a pool the less`AccPoolStake` it will have.  What's the assumption / semantics of the list?

ouroboros-network/test/Test/LedgerPeers.hs
```diff
@@ -0,0 +1,162 @@
...
+            map (\p -> ( PoolStake (fromIntegral (spStake p) % fromIntegral totalStake)
+                       , spRelay p)) sps
+
+-- | A pool with 100% stake should allways be picked.
+prop_pick100 :: Word16
+             -> Property
+prop_pick100 seed =
+    let rng = mkStdGen $ fromIntegral seed
+        sps = [ (1, RelayAddressAddr (read "1.1.1.1") 1  :| [])
+              , (0, RelayAddressAddr (read "0.0.0.0") 0  :| [])
```
coot, line 77: if `RelayAddress` was abstract / polymorhpic over addresses & dns names, we wouldn't need to use real ip addresses here.

ouroboros-network/test/Test/LedgerPeers.hs
```diff
@@ -0,0 +1,162 @@
...
+             SimException e trace -> do
+                 return $ counterexample (intercalate "\n" $ show e : trace) False
+             SimDeadLock trace -> do
+                 return $ counterexample (intercalate "\n" $ "Deadlock" : trace) False
+             SimReturn (_, peers) _trace -> do
+                 -- printf "Log: %s\n" (intercalate "\n" _trace)
+                 return $ peers === [ RelayAddressAddr (read "1.1.1.1") 1 ]
+
+
+prop_pick :: LedgerPools
```
coot, line 93: Add a haddock which says that this property ensures that `length` of `pickPeers` is equal to `count`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
karknu: Consensus provides an ordered list, but it has no effect on the picking scheme.

ouroboros-network/test/Test/LedgerPeers.hs
karknu: I prefer using a concrete type in the test suite to having to add two extra parameters to `RelayAddress` in the code.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
coot: I think I got it now; it would be good if there was a paragraph how the picking scheme works, certainly that would help me read the code.

karknu: Agreed. will add something.

coot: Ok, let me know when you'll be done; Apart from this it's _LGTM_.

coot (approved):
LGTM

PR #2785, opened 2020-12-01
Allow for limited wallclock rollback

### LINKED ISSUE #2781: Allow for limited wallclock  rollback
`BlockchainTime` does not allow the time to go back one slot or more because of system clock changes (e.g., NTP sync). In the Byron era, slots were 20s, so this hardly every happened. Now, in the Shelley era, slots are 1s, and the clock going back one second or more because of an NTP sync is *much more* likely. Shutting down the node with a `SystemClockMovedBack` exception in such cases is rather drastic.

We should allow the clock to roll back up to some configurable number of seconds without throwing a fatal exception. When the rollback is within the configured maximum, we should trace a warning but *leave the slot unchanged*. When the rollback is more than the configured maximum, we will still throw the fatal `SystemClockMovedBack`. This is important, because if time moved too far back, we'd have to move blocks from the ImmutableDB to the VolatileDB.

mrBliss:
Fixes #2781.

By default, we will allow the system clock to roll back three seconds to account for time changes because of NTP. When this happens, `TraceSystemClockMovedBack` will be traced. Not that the current slot won't decrease, we'll just stay in the same slot longer.

When the system clock rolls back more than three seconds, we still shut down with the fatal `SystemClockMovedBack` exception.

### DIFF
--- a/ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
+++ b/ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
@@ -30,6 +30,7 @@ import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.BlockchainTime
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.ResourceRegistry
+import           Ouroboros.Consensus.Util.Time
 
 import           Test.Util.Orphans.Arbitrary ()
 import           Test.Util.Orphans.IOLike ()
@@ -85,7 +86,7 @@ prop_delayNextSlot TestDelayIO{..} =
         tdioStart  <- pickSystemStart
         let time = defaultSystemTime tdioStart nullTracer
         atStart    <- fst <$> getWallClockSlot  time tdioSlotLen
-        nextSlot   <-         waitUntilNextSlot time tdioSlotLen atStart
+        nextSlot   <-         waitUntilNextSlot time tdioSlotLen maxClockRewind atStart
         afterDelay <- fst <$> getWallClockSlot  time tdioSlotLen
         assertEqual "atStart + 1" (atStart + 1) afterDelay
         assertEqual "nextSlot"    nextSlot      afterDelay
@@ -96,6 +97,11 @@ prop_delayNextSlot TestDelayIO{..} =
         pick :: UTCTime -> SystemStart
         pick = SystemStart . Time.addUTCTime (negate tdioStart')
 
+    -- Will only be needed when the system clock rolls back during the execution
+    -- of this test, which is rather unlikely.
+    maxClockRewind :: NominalDiffTime
+    maxClockRewind = secondsToNominalDiffTime 20
+
 {-------------------------------------------------------------------------------
   Test delay using mock time
 -------------------------------------------------------------------------------}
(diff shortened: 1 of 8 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/HardFork.hs
```diff
@@ -47,15 +47,37 @@ import           Ouroboros.Consensus.Util.Time
 -- incur computational overhead.)
 newtype BackoffDelay = BackoffDelay NominalDiffTime
 
+-- | Events traced by 'hardForkBlockchainTime'
+data HardForkBlockchainTimeEvent =
+    -- | The current slot is unknown
+    UnknownCurrentSlot RelativeTime HF.PastHorizonException
+
+    -- | The system clock moved back a bit, but less than 'hfbtMaxClockRewind',
+    -- e.g., because of an NTPC sync. This is acceptable.
```
mrBliss, line 56:
```suggestion
    -- e.g., because of an NTP sync. This is acceptable.
```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -272,11 +274,17 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
...
-                      ex)
+                (\case
+                   UnknownCurrentSlot t ex ->
+                     TraceCurrentSlotUnknown
+                       (fromRelativeTime systemStart t)
+                       ex
+                   SystemClockMovedBackABit oldT newT ->
+                     TraceSystemClockMovedBack
+                       (fromRelativeTime systemStart oldT)
+                       (fromRelativeTime systemStart newT))
```
mrBliss, line 285: Merge the two types and fmap `fromRelativeTime systemStart` over it.

ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
```diff
@@ -96,14 +97,20 @@ prop_delayNextSlot TestDelayIO{..} =
         pick :: UTCTime -> SystemStart
         pick = SystemStart . Time.addUTCTime (negate tdioStart')
 
+    -- Will only be needed when the system clock rolls back during the execution
+    -- of this test, which is rather unlikely.
+    maxClockRewind :: NominalDiffTime
+    maxClockRewind = secondsToNominalDiffTime 3
```
mrBliss, line 103: 20s

ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
```diff
@@ -140,39 +147,42 @@ scheduleCountSkips (Schedule (t:ts)) = go t ts
...
-      -- (which might not be the successor of the previous)
-      | now' >  now = (now :) <$> go (need - 1) ss now'
-
-      -- If time went backwards, we should see an exception
-      | otherwise   = throwError (now, now')
+    -- This let's us treat the schedule as an infinite stream of offsets
+    advanceSchedule :: Schedule -> (Fixed E1, Schedule)
+    advanceSchedule (Schedule ss) =
+        case ss of
+          []    -> (0.0, Schedule [0.0])
```
mrBliss, line 161: Move this up and replace with an error case, add a comment about the local invariant.

ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
```diff
@@ -252,20 +269,26 @@ prop_delayNoClockShift =
     withMaxSuccess 1 $ ioProperty $ do
       now   <- getCurrentTime
       slots <- originalDelay $
-                 testOverrideDelay (SystemStart now) (slotLengthFromMillisec 100) 5
+                 testOverrideDelay
+                   (SystemStart now)
+                   (slotLengthFromMillisec 100)
+                   (secondsToNominalDiffTime 3)
```
mrBliss, line 275: 20s

edsko (approved):

mrBliss:
bors merge

mrBliss:
bors merge

PR #2799, opened 2020-12-09

mrBliss:
This brings in https://github.com/input-output-hk/cardano-ledger-specs/pull/2049

Also improve the shrinking of `CardanoApplyTxErr`. These shrinkers were added to help chase down the failure fixed in the PR above. Better to leave them in, in case it happens again.


edsko (approved):

mrBliss:
bors merge

PR #2802, opened 2020-12-10

mrBliss:
This brings in https://github.com/input-output-hk/cardano-ledger-specs/pull/2052


edsko (approved):
I'm getting a strong sense of deja vu.

mrBliss:
bors merge

Issue #1818, opened 2020-03-17
Test failure on Windows


coot:
The following error was encountered: [github-actions](https://github.com/input-output-hk/ouroboros-network/pull/1808/checks?check_run_id=514178898)
```
socket send receive IPv4:                                  FAIL (0.07s)
      *** Failed! Exception: 'MuxError {errorType = MuxIOException wsaWaitForCompletion (recvBuf) 64: invalid argument (The specified network name is no longer available.), errorMsg = "(recv errored)", errorStack = []}' (after 9 tests):
```

coot:
Fixed by #1627

PR #2806, opened 2020-12-11
Build consensus in GitHub actions

newhoggy:
* Add MacOS builds
* Add `consensus` builds as separate jobs
* Introduce `CACHE_VERSION` to make it easy to "invalidate" the cabal store cache
* Work around `setup-haskell`, `ghcup` bugs
* Build with `cabal-3.4-rc4` for better caching and faster builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -140,8 +140,6 @@ jobs:
         cat >> cabal.project.local <<EOF
         package cardano-crypto-praos
           flags: -external-libsodium-vrf
-        package ouroboros-consensus-shelley-test
-          flags: -external-libsodium-vrf
         EOF
 
         cat ./cabal.project.local
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,8 +13,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        subset: [network, consensus]
         ghc: ["8.6.5", "8.10.2"]
         os: [ubuntu-latest, windows-latest]
+        exclude:
+          - os: windows-latest
+            subset: consensus
 
     steps:
     - uses: actions/checkout@v2
@@ -29,6 +33,9 @@ jobs:
 
         echo "CABAL_BUILDDIR=$CABAL_BUILDDIR"
         echo "CABAL_BUILDDIR=$CABAL_BUILDDIR" >> $GITHUB_ENV
+
+    - name: Set cache version
+      run: echo "CACHE_VERSION=9w76Z3Q"          >> $GITHUB_ENV
 
     - name: Install pkgconfiglite
       if: matrix.os == 'windows-latest'
@@ -59,12 +66,14 @@ jobs:
       if: matrix.os == 'macos-latest'
       run: brew install libsodium
 
-    - name: Configure to use libsodium
+    - name: Install build environment
+      if: matrix.os == 'ubuntu-latest'
       run: |
-        cat >> cabal.project <<EOF
-        package cardano-crypto-praos
-          flags: -external-libsodium-vrf
-        EOF
+        sudo apt-get update
+        sudo apt-get -y install libsodium23 libsodium-dev
+        sudo apt-get -y install libsystemd0 libsystemd-dev
+        sudo apt-get -y remove --purge software-properties-common
+        sudo apt-get -y autoremove
 
     - name: Select optimal cabal version
       run: |
@@ -86,15 +95,6 @@ jobs:
           's|C:/GitLabRunner/builds/2WeHDSFP/0/ghc/ghc/inplace/mingw/bin/ld.exe|C:/ProgramData/chocolatey/lib/ghc.8.10.2/tools/ghc-8.10.2/mingw/bin/ld.exe|g' \
           C:/ProgramData/chocolatey/lib/ghc.8.10.2/tools/ghc-8.10.2/lib/settings
 
-    - name: Install build environment
-      if: matrix.os == 'ubuntu-latest'
-      run: |
-        sudo apt-get update
-        sudo apt-get -y install libsodium23 libsodium-dev
-        sudo apt-get -y install libsystemd0 libsystemd-dev
-        sudo apt-get -y remove --purge software-properties-common
-        sudo apt-get -y autoremove
-
     - name: Set up temp directory
       run: |
         echo 'TMPDIR=${{ runner.temp }}'  >> $GITHUB_ENV
@@ -105,23 +105,23 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ${{ steps.setup-haskell.outputs.cabal-store }}
-        key: cabal-store-${{ matrix.ghc }}-${{ matrix.os }}
+        key: cabal-store-${{ env.CACHE_VERSION }}-${{ matrix.ghc }}-${{ matrix.os }}
     # Cache parts of dist-newstyle (C:\dist)
-    - name: cache buiddir [cache]
+    - name: cache build directory [cache]
       uses: actions/cache@v2
       with:
         path: C:\dist\cache
-        key: dist-cache-${{ matrix.ghc }}-${{ matrix.os }}
-    - name: cache buiddir [src]
+        key: dist-cache-${{ env.CACHE_VERSION }}-${{ matrix.ghc }}-${{ matrix.os }}
+    - name: cache build directory [src]
       uses: actions/cache@v2
       with:
         path: C:\dist\src
-        key: dist-src-${{ matrix.ghc }}-${{ matrix.os }}
-    - name: cache builddir [packagedb]
+        key: dist-src-${{ env.CACHE_VERSION }}-${{ matrix.ghc }}-${{ matrix.os }}
+    - name: cache build directory [packagedb]
       uses: actions/cache@v2
       with:
         path: C:\dist\packagedb
-        key: dist-packagedb-${{ matrix.ghc }}-${{ matrix.os }}
+        key: dist-packagedb-${{ env.CACHE_VERSION }}-${{ matrix.ghc }}-${{ matrix.os }}
 
     - name: update Hackage index
       run: cabal v2-update
@@ -136,6 +136,14 @@ jobs:
     - name: Use cabal.project.local.windows
       run: |
         cat ./cabal.project.local.ci.windows >> ./cabal.project.local
+
+        cat >> cabal.project.local <<EOF
+        package cardano-crypto-praos
+          flags: -external-libsodium-vrf
+        package ouroboros-consensus-shelley-test
+          flags: -external-libsodium-vrf
+        EOF
+
         cat ./cabal.project.local
 
     - name: Install happy
@@ -156,28 +164,80 @@ jobs:
 
     # ntp-client
     - name: ntp-client [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build ntp-client
 
     - name: io-sim-classes [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build io-sim-classes
 
     - name: io-sim [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build io-sim
 
     - name: typed-protocols [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build typed-protocols
 
     - name: typed-protocols-examples [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build typed-protocols-examples
 
     - name: network-mux [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build network-mux
 
     - name: ouroboros-network-framework [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-network-framework
 
     - name: ouroboros-network [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-network
+
+    - name: ouroboros-consensus [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus
+
+    - name: ouroboros-consensus-mock [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-mock
+
+    - name: ouroboros-consensus-test [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-test
+
+    - name: ouroboros-consensus-mock-test [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-mock-test
+
+    - name: ouroboros-consensus-byron [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-byron
+
+    - name: ouroboros-consensus-byronspec [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-byronspec
+
+    - name: ouroboros-consensus-byron-test [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-byron-test
+
+    - name: ouroboros-consensus-shelley [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-shelley
+
+    - name: ouroboros-consensus-shelley-test [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-shelley-test
+
+    - name: ouroboros-consensus-cardano [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-cardano
+
+    - name: ouroboros-consensus-cardano-test [build]
+      if: matrix.subset == 'consensus'
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus-cardano-test
 
     #
     # Test network packages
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -174,6 +179,9 @@ jobs:
     - name: ouroboros-network [build]
       run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-network
 
+    - name: ouroboros-consensus [build]
+      run: cabal --builddir="$CABAL_BUILDDIR" build ouroboros-consensus
+
```
mrBliss, line 184: *If* we decide to build consensus in GitHub Actions, we should build all of it, not just one package. A manual topological sort of the dependencies:

* `ouroboros-consensus`
* `ouroboros-consensus-mock`
* `ouroboros-consensus-test`
* `ouroboros-consensus-mock-test`
* `ouroboros-consensus-byron`
* `ouroboros-consensus-byronspec`
* `ouroboros-consensus-byron-test`
* `ouroboros-consensus-shelley`
* `ouroboros-consensus-shelley-test`
* `ouroboros-consensus-cardano`
* `ouroboros-consensus-cardano-test`

All these steps are sequential right? Can't we parallelise them? For example, why not `cabal build all` and then `cabal test all`? Is that just to see which component failed?

BTW, I'm not convinced that we should include consensus. We already have Hydra, why add another kind of CI? For network, it makes sense, as they want to test Windows-only code, which is not possible with Hydra. But for consensus, there is hardly any Windows-only code.
I guess one reason would be to test with GHC 8.6.5? One could say that the CI in `cardano-node` covers this already.

mrBliss:
I think something went wrong while rebasing?

newhoggy:
> I think something went wrong while rebasing?

It failed because `ouroboros-consensus` doesn't have tests.

.github/workflows/build.yml
newhoggy: `ghc-8.6.5` is a good reason.  I think the one if `cardano-node` doesn't cover this because it is in a separate repository and won't prevent code that doesn't compile on `ghc-8.6.5` from merging in this one.

It possible to only build `consensus` in Linux only as well.

mrBliss:
> > I think something went wrong while rebasing?
> 
> It failed because `ouroboros-consensus` doesn't have tests.

No, I'm talking about the three commits in this PR that are from master.

The tests of `ouroboros-consensus` live in `ouroboros-consensus-test` (`-test-consensus` and `-test-storage`).

.github/workflows/build.yml
newhoggy: I think the reason we don't do `build all` is because the repository includes `Win32-network` which doesn't build on POSIX.  That project really ought to be in its own repository.

newhoggy:
> No, I'm talking about the three commits in this PR that are from master

Ah yes.  I've removed them.  Not sure how they got there.

newhoggy:
I pushed some changes to make the consensus stuff build on Linux only.

.github/workflows/build.yml
```diff
@@ -136,6 +136,14 @@ jobs:
     - name: Use cabal.project.local.windows
       run: |
         cat ./cabal.project.local.ci.windows >> ./cabal.project.local
+
+        cat >> cabal.project.local <<EOF
+        package cardano-crypto-praos
+          flags: -external-libsodium-vrf
+        package ouroboros-consensus-shelley-test
+          flags: -external-libsodium-vrf
```
mrBliss, line 144: Are these two lines needed? Only `cardano-crypto-praos` has that flag.

.github/workflows/build.yml
```diff
@@ -30,6 +34,9 @@ jobs:
         echo "CABAL_BUILDDIR=$CABAL_BUILDDIR"
         echo "CABAL_BUILDDIR=$CABAL_BUILDDIR" >> $GITHUB_ENV
 
+    - name: Set cache version
+      run: echo "CACHE_VERSION=9w76Z3Q"          >> $GITHUB_ENV
```
mrBliss, line 38: Where does `9w76Z3Q` come from?

.github/workflows/build.yml
```diff
@@ -13,8 +13,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        subset: [network, consensus]
         ghc: ["8.6.5", "8.10.2"]
         os: [ubuntu-latest, windows-latest]
+        exclude:
+          - os: windows-latest
+            subset: consensus
```
mrBliss, line 21: Does it fail to build? If it doesn't, I would enable it.

.github/workflows/build.yml
newhoggy: Random generated string.  It's used as a way to "invalidate the cache".  For example if I want to know how long a full build takes, I change the string.

.github/workflows/build.yml
newhoggy: I would like to submit that as a separate PR.  Some further work is needed to set up `libsodium` for the Windows build.

newhoggy: I ended up fixing the Windows build.

.github/workflows/build.yml
```diff
@@ -156,29 +165,81 @@ jobs:
...
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build typed-protocols
 
     - name: typed-protocols-examples [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build typed-protocols-examples
 
     - name: network-mux [build]
+      if: matrix.subset == 'consensus'
       run: cabal --builddir="$CABAL_BUILDDIR" build network-mux
```
mrBliss, line 189: This is all used by the network layer too. Some of these might even be network-only (`typed-protocols-examples`?), maybe @coot can confirm?

.github/workflows/build.yml
```diff
@@ -188,27 +249,78 @@ jobs:
...
+    #   run: cabal --builddir="$CABAL_BUILDDIR" test ouroboros-consensus-shelley-test
+
+    # - name: ouroboros-consensus-cardano [test]
+    #   if: matrix.subset == 'consensus'
+    #   run: cabal --builddir="$CABAL_BUILDDIR" test ouroboros-consensus-cardano
+
+    # - name: ouroboros-consensus-cardano-test [test]
+    #   if: matrix.subset == 'consensus'
+    #   run: cabal --builddir="$CABAL_BUILDDIR" test ouroboros-consensus-cardano-test
+
```
mrBliss, line 323: Why are these disabled? Because they take too long?

mrBliss (commented):
Great that you separated consensus from network :+1: 

You mentioned the tests fail on MacOS. How?

BTW, I'm still not yet sure whether going all in for GitHub Actions is the right thing to do. This is yet another form of CI that we will have to maintain in addition to the Hydra one. In the past year, it has been rather flaky in my experience (mostly setting up the env, especially the Windows one).

.github/workflows/build.yml
newhoggy: Yep.  Nearly 2 hours I believe.

newhoggy: To see the test times see this draft PR: https://github.com/input-output-hk/ouroboros-network/pull/2831

newhoggy:
MacOS test failures can be seen here: https://github.com/input-output-hk/ouroboros-network/pull/2839

nfrisby:
> MacOS test failures can be seen here: #2839

Executive summary: A test involving IP sees a different result for v4 and v6.

nfrisby:
I chatted with John on Slack. Here's my summary.

He listed out two primary reasons for adding these builds to the this repository's GitHub Actions.

> Our networking code does not support running in WINE, which is an emulation of Windows.  Our Hydra builds do not **yet** provide a way to run tests on Windows proper.  In particular `cardano-node` test failures when run in WINE could point to an emulation problem so we want to see the same test failures in a Windows VM.  Github Actions provide us to a way to actually run our tests in a Windows VM.
  
> Github Actions provides us a way to build and test against multiple GHC versions, which is also another feature that our Hydra builds do not yet support.  We have downstream projects that are still on `ghc-8.6.5` and so rely on our libraries building on `ghc-8.6.5`.  Having CI check this compile version reduces the amount of risk that code is introduces that causes breakages downstream.  We anticipate further GHC upgrades in the future and not forcing the entire organisation to an atomic switch from one GHC version to another reduces upgrade risks.
  
He and I also agreed that it makes sense for him to maintain these as a Good Samaritan, but it'd be good to identify others who could proxy/takeover for him if need be. The Consensus Team hasn't owned the CI -- we just know who to ask for help when it hiccups in an interfering way. John suggested @rvl might be a good fit to start with.

Thus, we're fine with this PR proceeding. I'll review the details now. 👍

.github/workflows/build.yml
```diff
@@ -30,112 +34,102 @@ jobs:
...
       run: cabal v2-update
 
-    # checkout repository
-    - name: checkout ouroboros-network repository
-      uses: actions/checkout@v2
-
     - name: Cabal Configure
       run: cabal --builddir="$CABAL_BUILDDIR" configure --enable-tests
 
     - name: Use cabal.project.local.windows
```
nfrisby, line 124: Should be `Use cabal.project.local.ci.windows`, I think.

.github/workflows/build.yml
```diff
@@ -146,6 +140,14 @@ jobs:
...
+    - name: Print logs
+      if: always()
+      run: |
+        for x in $(find 'C:\Users\runneradmin\AppData\Roaming\cabal\logs' -name '*.log'); do
+          echo "==== $x ===="
+          cat "$x"
+        done
+
     #
     # Build & Test network packages
```
nfrisby, line 152: This header seems out-of-date.

.github/workflows/build.yml
```diff
@@ -156,29 +158,81 @@ jobs:
 
     # ntp-client
```
nfrisby, line 159: I would suggest this partitioning:

```

These are all subset='network':

    - name: Win32-network [build]
    - name: ntp-client [build]
    - name: io-sim-classes [build]
    - name: io-sim [build]
    - name: typed-protocols [build]
    - name: typed-protocols-examples [build]
    - name: network-mux [build]
    - name: ouroboros-network-framework [build]
    - name: ouroboros-network [build]

These are subset='consensus':

    - name: ouroboros-consensus [build]
    - name: ouroboros-consensus-mock [build]
    - name: ouroboros-consensus-test [build]
    - name: ouroboros-consensus-mock-test [build]
    - name: ouroboros-consensus-byron [build]
    - name: ouroboros-consensus-byronspec [build]
    - name: ouroboros-consensus-byron-test [build]
    - name: ouroboros-consensus-shelley [build]
    - name: ouroboros-consensus-shelley-test [build]
    - name: ouroboros-consensus-cardano [build]
    - name: ouroboros-consensus-cardano-test [build]

```

.github/workflows/build.yml
nfrisby: > To see the test times see this draft PR: #2831

They `consensus` cells of the matrix are both currently at ~220 minutes.

newhoggy: The specification for Github runners is here: https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners

.github/workflows/build.yml
newhoggy: Fixed.  Thanks!

.github/workflows/build.yml
newhoggy: Fixed.  Thanks!

.github/workflows/build.yml
newhoggy: After reading this comment, I realised I mis-classified some of the packages under `consensus` rather than `network` which is now fixed.  Is this what you meant?

.github/workflows/build.yml
nfrisby: @newhoggy I glanced at the `Win32-network.cabal` and it looks like it disables most things when not built on windows. Moreover, `cabal v2-build all` works in my `nix-shell`.

Could you point me towards one of the `Win32-network`-on-POSIX build errors?

.github/workflows/build.yml
nfrisby: > GitHub hosts Linux and Windows runners on Standard_DS2_v2 virtual machines in Microsoft Azure with the GitHub Actions runner application installed.

> For more information about the Standard_DS2_v2 machine resources, see "[Dv2 and DSv2-series](https://docs.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series)" in the Microsoft Azure documentation.

> DSv2-series sizes run on Intel® Xeon® Platinum 8272CL (Cascade Lake), Intel® Xeon® 8171M 2.1GHz (Skylake) or the the Intel® Xeon® E5-2673 v4 2.3 GHz (Broadwell) or the Intel® Xeon® E5-2673 v3 2.4 GHz (Haswell) processors with Intel Turbo Boost Technology 2.0 and use premium storage.

> Size | vCPU | Memory: GiB | Temp storage (SSD) GiB | Max data disks | Max cached and temp storage throughput: IOPS/MBps (cache size in GiB) | Max uncached disk throughput: IOPS/MBps | Max NICs | Expected network bandwidth (Mbps)
> -- | -- | -- | -- | -- | -- | -- | -- | --
> Standard_DS2_v2 | 2 | 7 | 14 | 8 | 8000/64 (86) | 6400/96 | 2 | 1500

.github/workflows/build.yml
```diff
@@ -30,185 +34,288 @@ jobs:
...
 
     - name: typed-protocols-examples [test]
+      if: matrix.subset == 'network'
       run: cabal --builddir="$CABAL_BUILDDIR" run typed-protocols-examples:tests
 
     - name: network-mux [test]
+      if: matrix.subset == 'network'
       run: cabal --builddir="$CABAL_BUILDDIR" run network-mux:test
 
     # issue: #1818
```
nfrisby, line 264: Would you elaborate a bit on this line comment?

The last GitHub comment on Issue #1818 is

> Fixed by PR #1627

PR #1627 has lots of comments, that I didn't read yet. Did it actually resolve #1818? EG I'd guess no, because of your line comment here.

.github/workflows/build.yml
```diff
@@ -30,185 +34,288 @@ jobs:
...
     - name: ourobors-network-framework [test]
+      if: matrix.subset == 'network'
       run: cabal --builddir="$CABAL_BUILDDIR" run ouroboros-network-framework:test -- -p '$0 != "typed-protocols.Socket.socket send receive IPv4"' 
 
     - name: ouroboros-network [test]
+      if: matrix.subset == 'network'
       run: cabal --builddir="$CABAL_BUILDDIR" run ouroboros-network:test
 
+    # Test network packages
```
nfrisby, line 273: -> `Test consensus packages`

.github/workflows/build.yml
nfrisby: Yep, the `if: matrix.subset ==` guards all look as expected to me now. 👍

.github/workflows/build.yml
nfrisby: ^^^ So the machine's CPU is relatively underpower, with 2 relatively meager cores.

I'll do an experiment to see what portion of the time is spent on my laptop VM (not too much bigger than the Azure machine) in the ThreadNet tests and report back here.

.github/workflows/build.yml
newhoggy: I don't know the history to #1818.  The comment already existed and I'm just preserving it.  @coot  Can you comment?

.github/workflows/build.yml
newhoggy: Oops.  Thanks!

.github/workflows/build.yml
nfrisby: Ah, I see, my mistake. Thanks for clarifying.

.github/workflows/build.yml
newhoggy: It actually builds now:

https://github.com/input-output-hk/ouroboros-network/pull/2868/checks?check_run_id=1697932929

Should we just switch to doing that?

nfrisby: Nice! :+1

Yeah, I think `v2-build all` is appropriate and will save some maintenance.

.github/workflows/build.yml
nfrisby: I ran the "consensus" test suites then I ran just their ThreadNet components. It's not quite apples to apples, but it's 11m23s for the whole thing, and 4m41s for just the ThreadNet parts. so about 40% time in ThreadNet.

```
ouroboros-consensus ouroboros-consensus-mock ouroboros-consensus-test ouroboros-consensus-mock-test ouroboros-consensus-byron ouroboros-consensus-byronspec ouroboros-consensus-byron-test ouroboros-consensus-shelley ouroboros-consensus-shelley-test ouroboros-consensus-cardano ouroboros-consensus-cardano-test
```

I used a fixed arbitrary QC replay seed. And I ran one test suites at a time, just like the GH action does.

I figured this on my laptop's VM would be loosely comparable to the GH action Azure machine. But then, compared to the numbers on the PR 2831 you shared above: mine are many times faster. EG the consensus 8.10.2 build, at https://github.com/input-output-hk/ouroboros-network/pull/2831/checks?check_run_id=1642016940, took ~200 minutes total. The breakdown of major/suspected summands there is

> build deps 30m + ouroboros-consensus-mock-test [test] 20m + ouroboros-consensus-byron-test [test] 60m + ouroboros-consensus-cardano-test [test] 60m + ouroboros-consensus-shelley-test [test] 7m = 177 minutes

I was careful to build them before running them. I peeked on the detailed logs on that experiment's PR and they also seem to have already been built before running for those times I quoted just above.

So... I think disabling the ThreadNets might save about 35% of the exec time or so? But that's based on the assumption that the reduction would be a similar percentage.

What do you think @newhoggy? Would that make a worthwhile difference? The ThreadNet tests are a "last line of defense" in some sense, but we at least run them via the Hydra push and merge CI and nightly.

.github/workflows/build.yml
newhoggy: Build all in done.

.github/workflows/build.yml
nfrisby: @newhoggy Here's a simpler option: let's just remove all of these lines and replace it with a single comment saying that the consensus tests take too long on GitHub's Azure machines.

I think I can Approve after that.

.github/workflows/build.yml
coot: I think this comment is stale and can be removed.  Early on we needed to filter out some tests, which does not happen any more.

nfrisby (approved):
Thanks for all the back and forth. LGTM.

I'm Approving, because it's not crucial, but I'm pretty sure @coot said somewhere (I forget) that the `issue 1818` comment can be dropped; it's resolved.

newhoggy:
bors merge

nfrisby:
Thanks for this, @newhoggy for introducing it and owning it (for the foreseeable future at least). I look forward to seeing it catch some Windows/8.6.5/etc regressions! :)

newhoggy:
Thanks for all your help getting this through.  It means in future, when there are down stream compile issues we can very quickly rule out upstream issues by pointing to recent successful CI builds.  😁

PR #2812, opened 2020-12-11
ChainSyncClient: derive ourTip from ourFrag

mrBliss:
Instead of storing `ourTip` in the `UnknownIntersectionState` and the
`KnownIntersectionState` state, just derive it from `ourFrag` when we need it.


edsko (approved):

mrBliss:
bors merge

PR #2803, opened 2020-12-11
Update index-state to 2020-12-11

newhoggy:

PR #2813, opened 2020-12-13
Stylish Haskell in Github Actions

newhoggy:


newhoggy:
https://github.com/input-output-hk/ouroboros-network/pull/2803 has been resolved.  We no longer need this work-around, but it can be something to consider in the future.

PR #2820, opened 2020-12-15
HFC: distinguish ForgeState "updated" from "unchanged"

mrBliss:
In the `BlockForging` combinator for the HFC, we were classifying a single-era
update of the forge state as "updated" even when it was "unchanged". This means
the node can no longer distinguish the two cases and that it might trace the
forge state every slot.


amesgen:

Issue #2614, opened 2020-09-17
Split ouroboros-network & ouroboros-consensus packages


coot:
`ouroboros-consensus` package is quite large, and whenever the networking team
needs to update it, we suffer from re-compiling a lot of code.  The parts that
depend on `ouroboros-network` are twoards the end of compilation pipeline.

I sugest introducion `ouroboros-consensus-node` package where the integration between
`ouroboros-network` and `ouroboros-cosnensus` is made.  We could also split
`ouroboros-network` so that the `ouroboros-cosnensus` package would only depend
on the mini protocols, chain fragments etc, but not on p2p-governor, diffusion
and other rather low level networking details.

edsko:
The goal would be to end up with the `ouroboros-consensus` main lib not depending on `-network` at all anymore.

edsko:
Network would be happy with reduced compilation times and I would be happy with improved independency and an improved understanding of what exactly we depend on, and where.

coot:
I'd be happy with cleaner dependency tree too :).

mrBliss:
I'm all for it!

mrBliss:
Some suggestions for names:

* `ouroboros-common`: for the modules that both network and consensus need, e.g., `Ouroboros.Network.Block`, `Ouroboros.Network.ChainFragment`, etc. We can use the `Ouroboros.*` or `Ouroboros.Common.*` module hierarchy.
* `ouroboros-node`: for the modules that combine network with consensus and the node-related stuff. We can use the `Ouroboros.Node` module hierarchy. I think it would be best if there is no dependency (in either direction) between `ouroboros-node` and the various ledger integration packages (`ouroboros-consensus-byron`, `-shelley`, `-cardano`, `-mock`, ...). This would be good for "low coupling" but also for build speed.

I'm looking forward to do this reorganisation! (@edsko just for the contribution stats :smile:)

In consensus, we could also consider splitting up the storage related-code into a separate package, but I think this will be detrimental to the build speed since the ChainDB relies on most of the non-storage code anyway (as it does chain selection and ledger validation). So if we split off storage in a separate package, cabal must first compile the non-storage consensus code before it can start on the storage package, whereas now it can parallelise this when the dependencies allow it.

mrBliss:
[...]

PR #2821, opened 2020-12-15
Repo reorganisation

mrBliss:
This is an abandoned attempt at #2614. This has bitrotted. No point in trying to update it, the reorganisation should be done afresh from master.

I'm dumping this to show the direction I was thinking about and some of the refactorings that would be needed.

Some ideas:
* A goal is that the network layer does not depend on `HasHeader`, `SlotNo`, etc. The network layer should parameterise over such things (this would make it possible to to switch to `LocalSlotNo`s as part of the EBB decontamination plan). The `AnchoredSeq` refactoring was a step in that direction. For convenience, the network tests can copy of `HasHeader`, which would move to consensus so that they don't have to be rewritten. The network test version of `HasHeader` can fold in `HasFullHeader`.
* A goal is that `ouroboros-byron/shelley/cardano/mock` don't depend on the network layer at all. To accomplish that, we'd have to think about the `SupportedNetworkProtocolVersion` instances, as they couple the network versions (defined in the network layer) to the consensus versions. Either by moving the network versions to the common package (strange), by moving these instances lower down the dependency tree, or ...


mrBliss:
As mentioned above, this PR is just a though dump. Feel free to close it when no longer useful.

amesgen:

PR #2822, opened 2020-12-15
Remove global ImmutableDB index cache

mrBliss:
This is an old attempt at input-output-hk/ouroboros-consensus#618. It only implemented step 1 of the proposal.


amesgen:

PR #2823, opened 2020-12-15
Ledger benchmarking

mrBliss:
This is my in-progress branch for ledger benchmarking before I got distracted by other things.

Benchmarking is just another pass of the `db-analyser`. The separate `analyse-results` tool should be folded into the benchmarking pass.

Only the three WIP commits are relevant, the others have been merged in master.


amesgen:

PR #2825, opened 2020-12-16
Address review comments on #2718

karknu:


ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -317,13 +317,15 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
             bufferedTxs2 = foldl' (flip Map.delete)
                                    bufferedTxs1 acknowledgedTxIds
 
-        _writtenTxids <- mempoolAddTxs txsReady
+            -- If we are acknowleding transactions that are still in unacknowledgedTxIds'
+            -- we need to re-add them so that we also can acknowledge them again later.
+            -- This will happen incase of duplicate txids within the same window.
+            live = filter (`elem` unacknowledgedTxIds') $ toList acknowledgedTxIds
+            bufferedTxs3 = forceElemsToWHNF $ bufferedTxs2 <>
+                               (Map.fromList (zip live (repeat Nothing)))
```
mrBliss, line 325: While you're at it, you could do:
```suggestion
            bufferedTxs3 = bufferedTxs2 `Map.union`
                             Map.fromList (zip live (repeat Nothing))
```

karknu: Does it work?

mrBliss: Never mind, the lazy `(<>)` is being used in a few other places and, as mentioned before, putting `Maybe`s in a map is a thunky thing to do.

(What we would do in consensus: not a use the lazy `Maybe` type as the values of the map and use the strict `Map.union` everywhere.)

coot: Using strict maybe would be more appropriate, `forceElemsToWHNF` will not evaluate what's below the lazy `Just` constructor, and we could just get rid of `forceElemsToWHNF`.

@mrBliss, doesn't `<>` use `Map.union` for `Map`?: it seems it [does](https://hackage.haskell.org/package/containers-0.6.4.1/docs/src/Data.Map.Internal.html#line-499).

coot: @karknu I would also break the `bufferedTxs1` into
```
            bufferedTxs1 = bufferedTxs st
                        <> txIdsRequestedWithTxsReceived
```

mrBliss: Oh, indeed `forceElemsToWHNF` doesn't help :facepalm:

> doesn't `<>` use `Map.union` for `Map`

Yeah, but that `<>` is defined `Data.Map.Lazy.union`, not `Data.Map.Strict.union`.

coot: > Yeah, but that <> is defined Data.Map.Lazy.union, not Data.Map.Strict.union.

Ouch :facepalm:, it's good to rembember.

coot: @karknu would you mind adding a patch to this PR to switch to strict maybe?

mrBliss: BTW, soon you can use `StrictMaybe` from `cardano-base`: https://github.com/input-output-hk/cardano-base/pull/198

karknu: @coot That seems like a risky change. `forceElemsToWHNF` and lazy maybe was used before I added support for duplicate txids.

coot: I don't think it's risky to make things more strict here, but let's wait for `cardano-base`.

coot (approved):
LGTM

PR #2835, opened 2020-12-29
Preparatory changes for ThreadNet rewrite

nfrisby:
This PR is the first of a few. These are relatively minor lower-level changes that I've needed for the ongoing ThreadNet rewrite. Since it's beginning to mature, I'm going to open a few PRs that contain a few of these changes.

They're easy to split off, if you'd like to review them separately. Let me know.

The main change in this set of commits is that we add the ability to _suppress_ a node's selection in a slot in the leader schedule. This is the core functionality enabling the "throttler" that prevents common prefix violations in the new ThreadNet tests despite small `k` and relatively large `f`. See `ForgeStateUpdateSuppressed`.

### DIFF
--- a/ouroboros-consensus-test/src/Test/ThreadNet/Util/NodeTopology.hs
+++ b/ouroboros-consensus-test/src/Test/ThreadNet/Util/NodeTopology.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DeriveGeneric       #-}
+{-# LANGUAGE DerivingVia         #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
@@ -18,7 +20,9 @@ import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Set (Set)
 import qualified Data.Set as Set
+import           GHC.Generics (Generic)
 import           GHC.Stack (HasCallStack)
+import           Quiet (Quiet (..))
 import           Test.QuickCheck
 
 import           Ouroboros.Consensus.Node.ProtocolInfo
@@ -46,8 +50,10 @@ import           Ouroboros.Consensus.Util.Orphans ()
 -- connected components during the execution, but the base topology is
 -- connected.
 --
-newtype NodeTopology = NodeTopology (Map CoreNodeId (Set CoreNodeId))
-  deriving (Eq, Show)
+newtype NodeTopology =
+    NodeTopology {unNodeTopology :: Map CoreNodeId (Set CoreNodeId)}
+  deriving (Eq, Generic)
+  deriving (Show) via Quiet NodeTopology
 
 instance Condense NodeTopology where
   condense top@(NodeTopology m)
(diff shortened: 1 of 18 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
```diff
@@ -218,20 +211,8 @@ checkShouldForge BlockForging{..}
 -------------------------------------------------------------------------------}
 
 -- | The result of updating something, e.g., the forge state.
-data UpdateInfo updated unchanged failed =
-    Updated      updated
-  | Unchanged    unchanged
+data UpdateInfo updated failed =
+    Updated updated
+    -- ^ The update have induced no change.
```
mrBliss: *has

ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
```diff
@@ -218,20 +211,8 @@ checkShouldForge BlockForging{..}
 -------------------------------------------------------------------------------}
 
 -- | The result of updating something, e.g., the forge state.
-data UpdateInfo updated unchanged failed =
-    Updated      updated
-  | Unchanged    unchanged
```
mrBliss, line 223: Just an FYI: this *might* affect the tracing of the KES info in the node. However, I think that because the HFC has no "unchanged" case (see https://github.com/input-output-hk/ouroboros-network/pull/2820), this change is probably invisible.

I guess this is more of a warning for when people start complaining (again?) about the node tracing the KES metrics every leadership check (= every slot) instead of when they changed. It used to be the latter (traced when changed), but because they want them traced at startup so that the RTView can pick them up (communicates via tracing), it was changed to be traced at every leadership check. Otherwise the RTView would have to wait until they actually change (after slotsPerKESPeriod = 129600 slots = 36 hours).

ouroboros-consensus-test/src/Test/ThreadNet/Util/NodeTopology.hs
```diff
@@ -18,9 +20,12 @@ import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Set (Set)
 import qualified Data.Set as Set
+import           GHC.Generics (Generic)
 import           GHC.Stack (HasCallStack)
 import           Test.QuickCheck
 
+import           Quiet (Quiet (..))
```
mrBliss, line 27: According to the style guide, you can group this import with the other imports above it :slightly_smiling_face:

mrBliss (approved):
LGTM, easy to review :+1:

ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
nfrisby: I'm confused as to how you even saw that line: it's not present on `4b22efadc`, which is -- and I think always has been -- the tip of this PR.

ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
nfrisby: Thanks for the note. (We have agreed to deal with this later, whenever it comes up, if it does.)

ouroboros-consensus-test/src/Test/ThreadNet/Util/NodeTopology.hs
nfrisby: Fixed! Thanks.

Issue #2725, opened 2020-11-02
Support trace event emission from within STM in `io-sim`


nfrisby:
This feature can aid in debugging when a transaction is triggering `retry` more or less often than expected.

Right now, only `Debug.Trace.trace` etc can be used. But if `io-sim` provided an interface to allow `MonadSay` and/or similar within an `STM m` computation, then the ultimate `Trace` could contain both the emitted message and also the surrounding events (eg which `TVar`s changed to incur the latest evaluation of the transaction).

EG I recall times where I needed to print the "fingerprint" portion of a transaction in order to determine why it was not finishing as often as I had expected. I think this would have helped with that.

The class instances for `STM IO` could be either no-ops or `TypeError`s.

`String` is the minimum viable argument, but `Dynamic` might also make sense (cf `Say` and `Output` from `io-sim:SimA`).

nfrisby:
@mrBliss and I discussed the motivation for this on PR #2387.

I came up with three motivators for `traceSTM :: Dynamic -> io-sim:STM`.

   * single-event-stream. I want to limit the number of factors that influence the order of events I see when eg debugging. With `saySTM` and `traceSTM`, the order will be determined by the `io-sim` scheduler (in turn: the program's internal synchronizing etc) and nothing else.
   
  * post-processable. I may want to post-process the event stream instead of viewing the full events. For example, I could ask "how many times had this transaction called `retry` before the timeout fired?" Or a `QuickCheck.counterexample` could report the state within an STM transaction (ie which of several `STM.check` calls is it blocked on, or even what is the latest value it had read from some `TVar`, etc).
 
  * even-despite-exceptions. I may want to include the result of post-processing in a `QuickCheck.counterexample` call for a failure that was triggered by an exception. (EG Maybe I'm debugging what appears to be a divergent loop and I introduce a `timeout` exception in order to reveal some state within the loop).

mrBliss:
As @nfrisby said: this is not about making it possible to trace from `STM m`, but only to trace from the IO simulator's `STM` monad (see the three motivators). In particular, the goal is *not* to make it possible to trace from the real `STM` monad.

When you find yourself needing to trace from within a real (non-test) `STM` transaction, there are (at least) two alternatives:

1. Let the transaction return the trace message(s), which can then be traced after the transaction succeeded by a regular tracer.
[...]

PR #2837, opened 2020-12-29
Preparatory changes for ThreadNet rewrite (PR number 2)

nfrisby:
This PR is the second of a few. These are relatively minor lower-level changes that I've needed for the ongoing ThreadNet rewrite. Since it's beginning to mature, I'm going to open a few PRs that contain a few of these changes.

This set of commits primary alteration is a configurable timeout for the response to ChainSync's `MsgFindIntersect` message. In the tests create large delays that are difficult to predict, so we simply suspend this timeout in the test (it is currently `shortWait` in the implementation).

We also provide some additional STM primitives that have been useful in tracing/debuging STM-based deadlocks in `io-sim`.

Lastly, there is a slightly more granular export that allows additional reuse in the rewritten ThreadNet code.

### DIFF
--- a/io-sim/src/Control/Monad/IOSim/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSim/Internal.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GADTSyntax                 #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE MultiParamTypeClasses      #-}
 {-# LANGUAGE NamedFieldPuns             #-}
 {-# LANGUAGE RankNTypes                 #-}
@@ -24,6 +25,7 @@ module Control.Monad.IOSim.Internal (
   runIOSim,
   runSimTraceST,
   traceM,
+  traceSTM,
   STM,
   STMSim,
   SimSTM,
@@ -97,6 +99,9 @@ runIOSim (IOSim k) = k Return
 traceM :: Typeable a => a -> IOSim s ()
 traceM x = IOSim $ \k -> Output (toDyn x) (k ())
 
+traceSTM :: Typeable a => a -> STMSim s ()
+traceSTM x = STM $ \k -> OutputStm (toDyn x) (k ())
+
 data SimA s a where
   Return       :: a -> SimA s a
 
@@ -145,6 +150,9 @@ data StmA s a where
   WriteTVar    :: TVar s a ->  a -> StmA s b  -> StmA s b
   Retry        :: StmA s b
   OrElse       :: StmA s a -> StmA s a -> (a -> StmA s b) -> StmA s b
+
+  SayStm       :: String -> StmA s b -> StmA s b
+  OutputStm    :: Dynamic -> StmA s b -> StmA s b
 
 -- Exported type
 type STMSim = STM
@@ -313,6 +321,9 @@ instance MonadFork (IOSim s) where
   forkIO task        = IOSim $ \k -> Fork task k
   forkIOWithUnmask f = forkIO (f unblock)
   throwTo tid e      = IOSim $ \k -> ThrowTo (toException e) tid (k ())
+
+instance MonadSay (STMSim s) where
+  say msg = STM $ \k -> SayStm msg (k ())
 
 instance MonadSTMTx (STM s) where
   type TVar_      (STM s) = TVar s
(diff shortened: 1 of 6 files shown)


nfrisby:
The current ThreadNet rewrite doesn't rely on `liftStStm`, but I've included it with `say` and `traceSTM` to complete the analogy.

Name bikeshedding welcome now.

nfrisby:
@coot We could choose to let this close Issue #2725 if we wanted. My original concept there involved a class like `MonadSay` but for `Dynamic` instead of `String` eg. But that's ultimately not crucial.

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -146,6 +152,11 @@ data StmA s a where
   Retry        :: StmA s b
   OrElse       :: StmA s a -> StmA s a -> (a -> StmA s b) -> StmA s b
 
+  SayStm       :: String -> StmA s b -> StmA s b
+  OutputStm    :: Dynamic -> StmA s b -> StmA s b
+
+  LiftStStm    :: StrictST.ST s a -> (a -> StmA s b) -> StmA s b
```
mrBliss, line 158: What do you need this one for?

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -38,6 +40,7 @@ module Control.Monad.IOSim.Internal (
   Trace (..),
   TraceEvent (..),
   liftST,
+  liftStStm,
```
mrBliss, line 43: I understand that the `-Stm` suffix is consistent with the naming used in the `StmA` data type, but looking at this export list, `liftStStm` seems rather inconsistent: `ST` and `STM` are always uppercased in the other identifiers. Nevertheless, I admit that `liftSTSTM` is no good either and `liftStSTM` seems internally inconsistent :slightly_smiling_face:.

So because I can't think of a better name, I don't object.

However, as I comment below, why do you need it? If we can avoid it, we could also avoid the naming problem :smile:

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -146,6 +152,11 @@ data StmA s a where
   Retry        :: StmA s b
   OrElse       :: StmA s a -> StmA s a -> (a -> StmA s b) -> StmA s b
 
+  SayStm       :: String -> StmA s b -> StmA s b
+  OutputStm    :: Dynamic -> StmA s b -> StmA s b
```
mrBliss, line 156: You only need these for debugging right? And `Debug.Trace` is not an option in those settings (I guess you want a chronological trace)?

nfrisby: I use `SayStm` (via `say`) for debugging, yep. I basically provide it as an alternative to `debugTracer`.

I use `OutputStm` (via `traceSTM`) QuickCheck diagnostics (ie `tabulate` inputs and `Property`s that check event sequences) as well as for debugging.

nfrisby:
I can imagine scenarios in which we'd need `liftStStm`, but I've avoided them so far. (I had one at first, but I refactored.)

To simplify the PR, my latest force push removed `liftStStm` -- it's not hard to put back if we end up wanting/needing it.

io-sim/src/Control/Monad/IOSim/Internal.hs
nfrisby: I removed `liftStStm` in 30 Dec force push.

io-sim/src/Control/Monad/IOSim/Internal.hs
nfrisby: I removed `liftStStm` in 30 Dec force push.

io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -1278,12 +1288,14 @@ data StmStack s b a where
                    -> StmStack s b c
                    -> StmStack s a c
 
-execAtomically :: forall s a.
-                  TVarId
+execAtomically :: forall s a c.
+                  Time -> ThreadId -> Maybe ThreadLabel
```
coot, line 1292: I'd prefer to break it to get one arg per line.

coot (approved):
LGTM, just a minor stylistic comment.

io-sim/src/Control/Monad/IOSim/Internal.hs
nfrisby: Fixed, thanks for catching my sloppiness there.

nfrisby:
@mrBliss Marcin approved; you also OK with this diff?

mrBliss (approved):
> @mrBliss Marcin approved; you also OK with this diff?

Yes :slightly_smiling_face:

io-sim/src/Control/Monad/IOSim/Internal.hs
nfrisby: For future referents, we elaborated on the ticket; Issue #2725.

PR #2840, opened 2021-01-04
Set fill-column as emacs directory variable in ouroboros-consensus/docs

nfrisby:
This helps ensure that at least all emacs users' paragraph wraps will be idempotent.

### DIFF
--- a/ouroboros-consensus/.dir-locals.el
+++ b/ouroboros-consensus/.dir-locals.el
@@ -1,0 +1,5 @@
+;;; Directory Local Variables
+;;; For more information see (info "(emacs) Directory Variables")
+
+((nil
+  (fill-column . 80)))


ouroboros-consensus/docs/.dir-locals.el
```diff
@@ -0,0 +1,5 @@
+;;; Directory Local Variables
+;;; For more information see (info "(emacs) Directory Variables")
+
+((nil
+  (fill-column . 80)))
```
mrBliss, line 5: I was expecting `latex-mode` instead of `nil`, but indeed, `nil` is good if not better.

Thought: why not promote it up one directory level so that it also applies to the rest of the consensus code, i.e., the comments?

mrBliss (approved):

ouroboros-consensus/docs/.dir-locals.el
nfrisby: I wondered the same, so I've done so now.

PR #2843, opened 2021-01-05
Upgrade to ghc 8.10.3 in GitHub actions CI

newhoggy:
This PR is on top of https://github.com/input-output-hk/ouroboros-network/pull/2806

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,9 +121,24 @@ jobs:
     - name: Cabal Configure
       run: cabal --builddir="$CABAL_BUILDDIR" configure --enable-tests
 
-    - name: Use cabal.project.local.ci.windows
+    - name: Record dependencies
       run: |
-        cat ./cabal.project.local.ci.windows >> ./cabal.project.local
+        cat ${{ env.PLAN_JSON }} | jq -r '."install-plan"[].id' | sort | uniq > dependencies.txt
+
+    - uses: actions/cache@v2
+      name: Cache cabal store
+      with:
+        path: ${{ steps.setup-haskell.outputs.cabal-store }}
+        key: cache-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
+        restore-keys: |
+          cache-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
+          cache-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-
+
+    - name: Set up cabal.project.local
+      run: |
+        if [ "$OS" == Windows_NT ]; then
+          cat ./cabal.project.local.ci.windows >> ./cabal.project.local
+        fi
 
         cat >> cabal.project.local <<EOF
         package cardano-crypto-praos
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -30,185 +34,295 @@ jobs:
...
         echo "LIBSODIUM_PATH=$LIBSODIUM_PATH"
         echo "$LIBSODIUM_PATH" >> $GITHUB_PATH
 
-    - name: Install libsodium
+    - name: Install libsodium (MacOS)
       if: matrix.os == 'macos-latest'
       run: brew install libsodium
 
-    - name: Configure to use libsodium
+    - name: Install build environment
```
coot, line 77: I'd call it `Install libsodium (Linux)` to be consistent with other similar steps.

.github/workflows/build.yml
```diff
@@ -30,185 +34,295 @@ jobs:
...
+    - uses: actions/cache@v2
+      name: Cache cabal store
+      with:
+        path: ${{ steps.setup-haskell.outputs.cabal-store }}
+        key: cache-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
+        restore-keys: |
+          cache-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
+          cache-${{ env.CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-
+
+    - name: Use cabal.project.local.ci.windows
```
coot, line 131: Shouldn't we use this only on `Windows` as the name `cabal.project.local.ci.windows` suggests?

.github/workflows/build.yml
newhoggy: Done.  Thanks!

.github/workflows/build.yml
newhoggy: Fixed.  Thanks!

nfrisby:
I'm checking locally how difficult it'd be to avoid `-Wno-unused-matches`. Otherwise seems good, thanks 👍 .

nfrisby:
Ryan Scott has already patched `bifunctors` so that the `-unused-match` warnings are no longer generated by the splices added in this PR.

I don't know when `bifunctors` will cut a new release with that, and even if I did, I don't know when -- or even admittedly how -- we'd adopt that. But eventually we don't need to disable those warnings anymore 🙌.

(Well, I'm pretty sure we won't need the warnings. Like I said, I'm not sure how to even temporarily test our code with a newer `bifunctors` --- since I only use the nix env and I haven't earned much mastery of `haskell.nix` yet.)

.github/workflows/build.yml
```diff
@@ -121,9 +121,25 @@ jobs:
     - name: Cabal Configure
       run: cabal --builddir="$CABAL_BUILDDIR" configure --enable-tests
 
-    - name: Use cabal.project.local.ci.windows
+    - name: Record dependencies
       run: |
-        cat ./cabal.project.local.ci.windows >> ./cabal.project.local
```
coot, line 126: How do we set `-Werror` option, this used to be done in `cabal.project.local.ci.windows`?

newhoggy: The diff is a bit funny.  The `cat ./cabal.project.local.ci.windows >> ./cabal.project.local` is preserved, but is lower down.  I've added an `if` statement around it so it is only used for Windows.

PR #2847, opened 2021-01-06
[CAD-2455] Document what needs to be done to add another era

mrBliss:
In particular, a Shelley-based era. This can be used as a template for the
Alonzo era.

### DIFF
--- a/ouroboros-consensus/docs/AddingAnEra.md
+++ b/ouroboros-consensus/docs/AddingAnEra.md
@@ -1,0 +1,127 @@
+# Adding a new Shelley-based era to consensus
+
+This document walks through the steps required to add a new Shelley-based era to
+consensus and to add it as an extra era to the Cardano blockchain.
+
+Prior art upon which this is based:
+* [#2666](https://github.com/input-output-hk/ouroboros-network/pull/2666)
+* [#2679](https://github.com/input-output-hk/ouroboros-network/pull/2679)
+
+The steps are fairly straightforward. We have put things in place (in consensus
+and in the ledger) when adding the Allegra and Mary eras that should now make it
+much easier to add a new additional era.
+
+There are two main driving changes: adding the new era to
+`Ouroboros.Consensus.Shelley.Eras` and including that era in
+`Ouroboros.Consensus.Cardano.Block`. The compiler should point the rest of the
+way. It's mostly a matter of following the existing patterns by imitating the
+previous Shelley-based case. Be sure to run both the Shelley
+(`ouroboros-consensus-shelley-test`) and the Cardano
+(`ouroboros-consensus-cardano-test`) tests.
+
+For exhaustiveness, we give an overview of the changes needed. The new era we'll
+be adding is the Alonzo era, which comes after the Mary era.
+
+## Preparation
+
+* Locate the new tag in the ledger, e.g., `AlonzoEra`. This is an empty data
+  type that is used at the type level to indicate the era. The ledger should
+  have an instance of the `ShelleyBasedEra` class (the class defined in
+  `Shelley.Spec.Ledger.API`, do not confuse it with the class with the same name
+  in consensus) for this era. This class should provide all the instances
+  consensus integration will rely on.
+
+* Note that both the affected consensus libraries and test suites are parametric
+  in the Shelley-based era or in the general era. The test suite relies on the
+  necessary `Arbitrary`, `Show`, `Eq`, ... instances to be available for all of
+  the Cardano eras. This means that a new era can only be added to the Cardano
+  eras when the ledger provides all the necessary instances for the new era,
+  *including the instances for the test suite* (e.g., the serialisation
+  generators). Note that because the test suite is parametric/generic in the
+  eras, it is not easy to temporarily omit a single era from the test suite
+  until the required instances have been added. This means that we can't extend
+  the library with the new era without extending the test suite too. In the
+  past, this has been the main blocker: the library was ready, but the required
+  instances in the ledger were missing for our test suite(s).
+
+* Add the required dependencies on the ledger package (and its test package) to
+  the `cabal.project` file. You will have to add it to some other cabal files
+  too, let the compiler tell you which ones.
+
+## `ouroboros-consensus-shelley`
+
+* Define `StandardAlonzo` in `Ouroboros.Consensus.Shelley.Eras` and add any
+  missing instances to that module, update the export list appropriately.
+
+* In `Ouroboros.Consensus.Shelley.Node`, define `ProtocolParamsAlonzo` just like
+  the existing ones. In case the era in question needs extra parameters (e.g., a
+  new genesis config?), they can be added here.
+
+## `ouroboros-consensus-shelley-test`
+
+* In `Test.Consensus.Shelley.Examples`, define an `examplesAlonzo` similar to
+  how `examplesMary` is defined. If necessary, i.e., when new type families have
+  been created in the ledger (e.g., `TxOut`), the `examples` function might have
+  to take more arguments. This is where the golden test examples are defined,
+  but only the Shelley ones are tested as part of this test suite. The others
+  are only tested as part of the `ouroboros-consensus-cardano-test` test suite.
+
+## `ouroboros-consensus-cardano`
+
+* In `Ouroboros.Consensus.Cardano.Block`, include `AlonzoEra` in `CardanoEras`
+  and `ShelleyBasedEras`. Update all the pattern synonyms in the module with the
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/docs/AddingAnEra.md
```diff
@@ -0,0 +1,126 @@
...
+  necessary `Arbitrary`, `Show`, `Eq`, ... instances to be available for all of
+  the Cardano eras. This means that a new era can only be added to the Cardano
+  eras when the ledger provides all the necessary instances for the new era,
+  *including the instances for the test suite* (e.g., the serialisation
+  generators). Note that because the test suite is parametric/generic in the
+  eras, it is not easy to temporarily omit a single era from the test suite
+  until the required instances have been added. This means that we can't extend
+  the library with the new era without extending the test suite too. In the
+  past, this has been the main blocker: the library was ready, but the required
+  instances in the ledger were missing for our test suite(s).
```
mrBliss, line 44: I don't know whether the required testing instances are already in place. As mentioned in this paragraph, they have to be in order for us to complete the integration in consensus.

Thought: just like we added the `ShelleyBasedEra` class to the ledger, which implies all the instances we need for the consensus integration, we could add a similar class to the ledger, but which implies all the instances we need for the integration in the consensus tests.

ouroboros-consensus/docs/AddingAnEra.md
```diff
@@ -0,0 +1,126 @@
...
+
+The steps are fairly straightforward. We have put things in place (in consensus
+and in the ledger) when adding the Allegra and Mary eras that should now make it
+much easier to add a new additional era.
+
+There are two main driving changes: adding the new era to
+`Ouroboros.Consensus.Shelley.Eras` and including that era in
+`Ouroboros.Consensus.Cardano.Block`. The compiler should point the rest of the
+way. It's mostly a matter of following the existing patterns by imitating the
+previous Shelley-based case. Be sure to run both the Shelley
```
nfrisby, line 18: > The compiler should point the rest of the
way. It's mostly a matter of following the existing patterns by imitating the
previous Shelley-based case.

🙌

ouroboros-consensus/docs/AddingAnEra.md
```diff
@@ -0,0 +1,126 @@
...
+(`ouroboros-consensus-shelley-test`) and the Cardano
+(`ouroboros-consensus-cardano-test`) tests.
+
+For exhaustiveness, we give an overview of the changes needed. The new era we'll
+be adding is the Alonzo era, which comes after the Mary era.
+
+## Preparation
+
+* Locate the new tag in the ledger, e.g., `AlonzoEra`. The ledger should have an
```
nfrisby, line 27: _tag_ is just what the Consensus and Ledger Teams have agreed/come to call this concept right?

ouroboros-consensus/docs/AddingAnEra.md
```diff
@@ -0,0 +1,126 @@
...
+
+* In `Ouroboros.Consensus.Shelley.Node`, define `ProtocolParamsAlonzo` just like
+  the existing ones. In case the era in question needs extra parameters (e.g., a
+  new genesis config?), they can be added here.
+
+## `ouroboros-consensus-shelley-test`
+
+* In `Test.Consensus.Shelley.Examples`, define an `examplesAlonzo` similar to
+  how `examplesMary` is defined. If necessary, i.e., when new type families have
+  been created, the `examples` function might have to take more arguments. This
```
nfrisby, line 63: > when new type families have been created

"when new type families have been created in `cardano-ledger-specs` or `Ouroboros.Consensus.Shelley.Eras`"?

ouroboros-consensus/docs/AddingAnEra.md
```diff
@@ -0,0 +1,126 @@
...
+  is where the golden test examples are defined, but only the Shelley ones are
+  tested as part of this test suite. The others are only tested as part of the
+  `ouroboros-consensus-cardano-test` test suite.
+
+## `ouroboros-consensus-cardano`
+
+* In `Ouroboros.Consensus.Cardano.Block`, include `AlonzoEra` in `CardanoEras`
```
nfrisby, line 70: Tangent: what do you think about `type CardanoEras c = ByronBlock ': ShelleyBasedEras c` or similar? Did you already weigh the pros and cons?

ouroboros-consensus/docs/AddingAnEra.md
```diff
@@ -0,0 +1,126 @@
...
+  derive `maxMajorProtVer` from the new final era. Update
+  `protocolClientInfoCardano` too.
+
+* In `Ouroboros.Consensus.Cardano`, update the `ProtocolCardano` type synonym,
+  add the extra arguments needed for `protocolInfoCardano` to the
+  `ProtocolCardano` constructor. Update `protocolInfo` accordingly.
+
+## `ouroboros-consensus-cardano-test`
+
+* In `Test.Consensus.Cardano.Generators`, update `arbitraryNodeToNode`,
```
nfrisby, line 96: This was the first module where everything felt unfamiliar and it wasn't "obvious" to me how to follow your instructions. It's also where I ran out of time for today. I'll pick up here next time.

ouroboros-consensus/docs/AddingAnEra.md
mrBliss: Correct.

The tag for Alonzo can be found here: https://github.com/input-output-hk/cardano-ledger-specs/blob/da27cd3bc9ad4f67e361d011dc6b32a1e6849713/alonzo/impl/src/Cardano/Ledger/Alonzo.hs#L15-L16

ouroboros-consensus/docs/AddingAnEra.md
mrBliss: In `cardano-ledger-specs`. For example, they recently turned `TxOut` in a type family, IIRC. I'll update the text.

ouroboros-consensus/docs/AddingAnEra.md
mrBliss: I hadn't thought about that, good idea! I don't see any real cons (an extra indirection :roll_eyes:). I can push a commit.

ouroboros-consensus/docs/AddingAnEra.md
mrBliss: Hmm, I could do some more hand holding here, but it's mostly a matter of following the existing patterns. After you have picked up from here, we can reconsider.

ouroboros-consensus/docs/AddingAnEra.md
mrBliss: Ah, I just realised why we didn't do that. Compare for example Allegra: in `CardanoEras`  it's `ShelleyBlock (AllegraEra c)`, in `ShelleyBasedEras` it's `AllegraEra c`. This means we'd need a type-level `map`, e.g.
```haskell
type CardanoEras c = ByronBlock ': Map ShelleyBlock (ShelleyBasedEras c)
```
For simplicity (there are already too many people that don't understand type families, even though here it would just be an almost literate promotion from the value level to the type level), we didn't bother.

For reference:
```haskell
type CardanoEras c = ByronBlock ': Map ShelleyBlock (ShelleyBasedEras c)

type ShelleyBasedEras c = '[ShelleyEra c, AllegraEra c, MaryEra c]

type family Map f xs where
  Map _ '[]       = '[]
  Map f (x ': xs) = f x ': Map f xs
```

edsko: FWIW, the core concepts in `generics-sop` is parameterized over functors precisely to avoid working with type-level maps, as in practice that can get quite painful.

nfrisby: Yep, I missed that. I agree, with your decision. Maybe in another 5 years `Map` will be a non-issue...

ouroboros-consensus/docs/AddingAnEra.md
nfrisby: OK. I understand the pattern now. Especially if I just search for the name of a previous era, it seems easy enough to handle this module.

I think my hesitation is do mostly to feeling lost in this sea of instances in this module. I don't think there's much to be done there in advance of me spending to the time to understand them all.

Unless you can think of a pithy Haddock comment for the module header? Something about serialization tests, IIUC.

nfrisby (approved):
LGTM. Really helpful. Thank you, Thomas.

ouroboros-consensus/docs/AddingAnEra.md
mrBliss: I added a Haddock comment for the module header.

mrBliss:
bors merge

PR #2856, opened 2021-01-12
ListLike-4.7.3

coot:
'ListLike >= 4.7.4' is a better constraint than 'ListLike < 4.7.2', and it
works too.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -28,7 +28,7 @@ constraints:
   ip < 1.5,
   hedgehog >= 1.0,
   bimap >= 0.4.0,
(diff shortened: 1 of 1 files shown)


karknu (approved):
LGTM.
So there was something wrong with 4.7.3?

coot:
`4.7.2` did not build against `bytestring-0.10.12.0`: it introduced instances which also where introduced in `bytestring`.
It seems `4.7.3` works as well, I'll relax the constraint.

coot:
bors merge

PR #2859, opened 2021-01-12
readme: edits

coot:
Experiment with <details> and <summary> tags. Include all demo applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ the Cardano Shelley implementation:
   This document is a technical specification of the networking protocol.  It
   includes serialisation formats, necessary details of multiplexer and
   technical specifications of mini-protocols used by either _node-to-node_ and
-  _node-to-client_ flavors of the protocol. 
+  _node-to-client_ flavours of the protocol.
 
 * [Haddock documentation](https://input-output-hk.github.io/ouroboros-network/)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,6 +146,10 @@ nix-shell --run ./scripts/buildkite/check-stylish.sh
 ```
 
 ## Tests
+<details>
+<summary>
+Consensus test components.
+</summary>
```
mrBliss, line 152: I like this!

I was recently thinking: is this list of commands even useful to have here, I'm not sure we even maintain it. Personally, I wouldn't mind if it were deleted, but others might disagree.

coot: I don't mind removing both lists.

coot: I don't think anybody of us is regularly using `nix-build`, it's quite likely that things has changed since these commands where written in the `README.md` file.

README.md
```diff
@@ -30,7 +30,7 @@ the Cardano Shelley implementation:
   This document is a technical specification of the networking protocol.  It
   includes serialisation formats, necessary details of multiplexer and
   technical specifications of mini-protocols used by either _node-to-node_ and
-  _node-to-client_ flavors of the protocol. 
+  _node-to-client_ flavours of the protocol.
```
nfrisby, line 33: IOHK style guide says to favor (:grin:) US spellings. I'm mostly pointing that out just to tease --- adhering to that in this repo would require a big diff.

README.md
nfrisby: Yep, I agree. This seems especially sound if running these via `cabal` inside of `nix-shell` is close enough to indistinguishable from running them via `nix-build`. I would think so, except maybe for things like what kind of file handle `stdin` is (termtype etc). Likely fine.

nfrisby (approved):

README.md
coot: Thanks, I wasn't aware of that.  Do you have a link to that style guide?

coot:
bors merge

README.md
nfrisby: I've sent you a Google Docs like on Slack.

PR #2865, opened 2021-01-13
Draft MajorTypeFamilies.md

nfrisby:
This document is a snapshot of the major type families in the Consensus Layer. I find that having a working intuition for each of these type families in isolation helps me interpret various type expressions more rapidly. This document is intended to help people develop/refresh such an intuition.

### DIFF
--- a/ouroboros-consensus/docs/MajorTypeFamilies.md
+++ b/ouroboros-consensus/docs/MajorTypeFamilies.md
@@ -1,0 +1,166 @@
+# Major Type and Data Families in Consensus
+
+I started with
+
+```
+$ git show
+commit bf8579cc2ff2a7bc4ba23150eff659cfd1c6ccca (origin/master)
+[snip]
+Date:   Tue Jan 12 14:08:36 2021 +0000
+
+    Merge #2855
+[snip]
+$ git status -suno
+$ git grep -E \
+    -e '^(data|type) +family' \
+    -e '^ +(data|type)[^=]*$' \
+    origin/master -- 'ouroboros-consensus/src/*hs'
+```
+
+and then snipped out what seemed too incidental, too general, too obvious, and
+so on.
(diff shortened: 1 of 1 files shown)


nfrisby:
@mrBliss -- What do you think, is this is document attempting a fool's errand? If not, any egregious omissions?

I'm just looking for any course corrections before your escape :)

ouroboros-consensus/docs/MajorTypeFamilies.md
```diff
@@ -0,0 +1,166 @@
...
+Date:   Tue Jan 12 14:08:36 2021 +0000
+
+    Merge #2855
+[snip]
+$ git status -suno
+$ git grep -E \
+    -e '^(data|type) +family' \
+    -e '^ +(data|type)[^=]*$' \
+    origin/master -- 'ouroboros-consensus/src/*hs'
+```
```
mrBliss, line 18: That's seems like an exhaustive way of finding them :+1:

mrBliss (approved):
I'm a bit reluctant to duplicate information that will inevitably get out of date, but I have been working on the tech report, so :zipper_mouth_face:.

If this is helpful, then I don't object.

nfrisby:
Closing for now. I'll revisit this idea later. It'll be helpful to have the new members' opinions.

nfrisby:
I've had second thoughts, so I'm Closing this. I'll reconsider it with the Consensus Team at a later date.

nfrisby:
I don't think we should take on this burden of maintaining this document. Instead, we should:

  * make sure each family declaration has great Haddock

  * record a regexp or some other suggested mechanism for listing out the type families declarations
  
Basically, the only thing this document could really do is collate the descriptions into one place. The right `git grep` query might be able to effectively do that, eg.

PR #2866, opened 2021-01-13
Update `cardano-ledger-specs`

nc6:
Most of this should be fairly straightforward, but note that we have to drop the example which manually constructs an `ApplyTxError`, since it assumed more than is allowed about the structure of the error.

### DIFF
--- a/ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
+++ b/ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
@@ -72,6 +72,7 @@ import           Ouroboros.Consensus.Util.Time
 import qualified Cardano.Ledger.AuxiliaryData as SL (AuxiliaryDataHash (..))
 import qualified Cardano.Ledger.Core as Core
 import           Cardano.Ledger.Crypto (ADDRHASH, Crypto, DSIGN, HASH, VRF)
+import qualified Control.State.Transition.Extended as SL (PredicateFailure)
 import           Shelley.Spec.Ledger.API (StrictMaybe (..))
 import qualified Shelley.Spec.Ledger.API as SL
 import qualified Shelley.Spec.Ledger.BaseTypes as SL (Seed (..),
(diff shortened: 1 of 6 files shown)


ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
```diff
@@ -245,7 +251,7 @@ examples value txBody auxiliaryData = Golden.Examples {
     results = labelled [
           ("LedgerTip",              SomeResult GetLedgerTip (blockPoint blk))
         , ("EpochNo",                SomeResult GetEpochNo 10)
-        , ("EmptyPParams",           SomeResult GetCurrentPParams SL.emptyPParams)
+        , ("EmptyPParams",           SomeResult GetCurrentPParams def)
```
mrBliss, line 254: Just wondering: would it make sense to make it a `Monoid` so we can use `mempty` instead of `def`? `Monoid` has laws :slightly_smiling_face:

nc6: It might be reasonable for `PParams` to be a `Semigroup`, but a monoid? No. Precisely because the default instance would _not_ satisfy any monoid laws - the default values are default values, not a monoidal unit.

nc6: (`PParams' Maybe` _could_ be a monoid, but its unit would not be the thing we want here as a default)

mrBliss (approved):

nc6:
bors merge

nfrisby:
Sorry I didn't chime in here yet. Thanks to you both for making it happen.

@nc6 I have a minor question for future reference. Your PR description says

> we have to drop the example which manually constructs an ApplyTxError, since it assumed more than is allowed about the structure of the error.

but the diff doesn't seem to remove anything. Am I missing something or were you able to preserve it after all?

mrBliss:
> Sorry I didn't chime in here yet. Thanks to you both for making it happen.
> 
> @nc6 I have a minor question for future reference. Your PR description says
> 
> > we have to drop the example which manually constructs an ApplyTxError, since it assumed more than is allowed about the structure of the error.
> 
> but the diff doesn't seem to remove anything. Am I missing something or were you able to preserve it after all?

He did indeed remove it, but I restored it. You can see it when you look at the individual commits.

PR #2872, opened 2021-01-14
ChainDB: minor tidying in Reader.forward

nfrisby:
See the single commit message.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Follower.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Follower.hs
@@ -415,14 +415,15 @@ forward ::
   -> [Point blk]
   -> m (Maybe (Point blk))
 forward registry varFollower blockComponent CDB{..} = \pts -> do
-    -- NOTE: we don't use 'Query.getCurrentChain', which only returns the last
-    -- @k@ headers, because we want to see the headers that have not yet been
-    -- written to the ImmutableDB too.
-    (curChain, followerState) <- atomically $
-      (,) <$> readTVar cdbChain <*> readTVar varFollower
-    slotNoAtImmutableDBTip <- atomically $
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Reader.hs
```diff
@@ -415,14 +415,15 @@ forward ::
...
 forward registry varReader blockComponent CDB{..} = \pts -> do
-    -- NOTE: we don't use 'Query.getCurrentChain', which only returns the last
-    -- @k@ headers, because we want to see the headers that have not yet been
-    -- written to the ImmutableDB too.
-    (curChain, readerState) <- atomically $
-      (,) <$> readTVar cdbChain <*> readTVar varReader
-    slotNoAtImmutableDBTip <- atomically $
-      ImmutableDB.getTipSlot cdbImmutableDB
-    findFirstPointOnChain curChain readerState slotNoAtImmutableDBTip pts
+    -- NOTE: we use `cdbChain` instead of 'Query.getCurrentChain', which only
```
mrBliss, line 418:
```suggestion
    -- NOTE: we use 'cdbChain' instead of 'Query.getCurrentChain', which only
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Reader.hs
```diff
@@ -415,14 +415,15 @@ forward ::
...
+    -- returns the last @k@ headers, because we need to also see the headers
+    -- that happen to have not yet been GC'd over to the ImmutableDB.
```
mrBliss, line 420:
```suggestion
    -- that happen to have not yet been copied over to the ImmutableDB.
```
They are first copied and after a delay of 60s GC'd. It's the copying that is important here, not the GC'ing.

mrBliss (approved):
Code looks good, the comment needs a few fixes.

PR #2873, opened 2021-01-15
Fix lint warnings

newhoggy:
Everything in here is either fixing lint warnings or suppressing them, except for one useful addition which is a new `readTVarIO` function.

One alternate proposal is to add an `hlint.yaml` config to suppress all the lint warnings and not do any lint fixes, which helps me get around the problem where my IDE is flagging a lot of warnings which interferes with reading of the tooltips showing inferred types of expressions.

### DIFF
--- a/ouroboros-consensus-test/ouroboros-consensus-test.cabal
+++ b/ouroboros-consensus-test/ouroboros-consensus-test.cabal
@@ -170,6 +170,7 @@ test-suite test-consensus
                      , quiet
                      , serialise
                      , sop-core
+                     , stm
                      , tasty
                      , tasty-hunit
                      , tasty-quickcheck
--- a/ouroboros-consensus-test/test-consensus/Test/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus-test/test-consensus/Test/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -71,6 +71,9 @@ import           Test.Util.Orphans.IOLike ()
 import           Test.Util.TestBlock
 import           Test.Util.Tracer (recordingTracerTVar)
 
+{- HLINT ignore "Redundant $" -}
+{- HLINT ignore "Use list literal" -}
+
 {-------------------------------------------------------------------------------
   Top-level tests
 -------------------------------------------------------------------------------}
@@ -310,7 +313,7 @@ runChainSync securityParam (ClientUpdates clientUpdates)
       -- has thrown an exception or has gracefully terminated, so that at the
       -- end, we can read the chains in the states they were in when the
       -- exception was thrown.
-      stop <- fmap isJust $ atomically $ readTVar varClientResult
+      stop <- isJust <$> readTVarIO varClientResult
       unless stop $ do
         -- Client
         whenJust (Map.lookup tick clientUpdates) $ \chainUpdates ->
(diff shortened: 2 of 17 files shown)


ouroboros-consensus-test/ouroboros-consensus-test.cabal
```diff
@@ -170,6 +170,7 @@ test-suite test-consensus
                      , quiet
                      , serialise
                      , sop-core
+                     , stm
```
mrBliss, line 173: This should not be needed.

ouroboros-consensus-test/test-consensus/Test/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -71,13 +71,15 @@ import           Test.Util.Orphans.IOLike ()
...
+{- HLINT ignore "Use list literal" -}
+
 {-------------------------------------------------------------------------------
   Top-level tests
 -------------------------------------------------------------------------------}
 
 tests :: TestTree
 tests = testGroup "ChainSyncClient"
-    [ testProperty "chainSync"                 $ prop_chainSync
+    [ testProperty "chainSync"                   prop_chainSync
```
mrBliss, line 82: Please restore the `$`, we know it's redundant, but it's for consistency with the other lines below. 

You'll see that the main reason we have redundant `$`s throughout consensus is for consistency.

mrBliss (commented):
Sorry for looking at it while it's only a draft, but it's my last day, so I might not see the final PR :slightly_smiling_face: 

I know you mean well, but I find it kind of strange to start fixing HLint warnings in the code of other teams (this is not some repo without real owners, like `cardano-prelude` or `cardano-base`). Clearly, the people in the network and consensus teams are not using HLint (or at least not consistently), so inevitably new warnings will be introduced when the code changes. Are *you* going to keep fixing them? :slightly_smiling_face: 

I think the decision to use HLint should be made by the team(s) involved and it should then be enforced by CI.

ouroboros-consensus-test/ouroboros-consensus-test.cabal
newhoggy: Yep.  Not needed.  Removed.  Thanks!

ouroboros-consensus-test/test-consensus/Test/Consensus/MiniProtocol/ChainSync/Client.hs
newhoggy: I put the `$` back.  Thanks!

newhoggy:
> Sorry for looking at it while it's only a draft, but it's my last day, so I might not see the final PR 🙂
> 
> I know you mean well, but I find it kind of strange to start fixing HLint warnings in the code of other teams (this is not some repo without real owners, like `cardano-prelude` or `cardano-base`). Clearly, the people in the network and consensus teams are not using HLint (or at least not consistently), so inevitably new warnings will be introduced when the code changes. Are _you_ going to keep fixing them? 🙂
> 
> I think the decision to use HLint should be made by the team(s) involved and it should then be enforced by CI.

It was more a case of getting all these lint warning annotations in my IDE obscuring my attempts to inspect the types of things.

I'm not fussed about fixing any more lint warnings than enough, so I can figure out unfamiliar.

Just offering these fixes since I'm already making them, but having thought about it, if the team decides to never use `hlint`, it still would be nice to add an `hlint` config that disables all the rules so it doesn't get flagged by tools.

Also happy to add a `hlint` to CI even if it means starting with a config that has all rules disabled to minimise disruption.

PR #2874, opened 2021-01-15
Miscellaneous traces and counters

newhoggy:
This PR tracks the work done for https://github.com/input-output-hk/cardano-node/issues/2261.

There is also work being done on the `cardano-node` repository for the same ticket: https://github.com/input-output-hk/cardano-node/pull/2289

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
@@ -2,21 +2,21 @@
 {-# LANGUAGE EmptyDataDeriving         #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE FlexibleContexts          #-}
-{-# LANGUAGE PatternSynonyms           #-}
 {-# LANGUAGE ScopedTypeVariables       #-}
 {-# LANGUAGE StandaloneDeriving        #-}
 {-# LANGUAGE TypeApplications          #-}
 {-# LANGUAGE TypeFamilies              #-}
 {-# LANGUAGE UndecidableInstances      #-}
+
 module Ouroboros.Consensus.MiniProtocol.BlockFetch.Server
   ( blockFetchServer
     -- * Trace events
-  , TraceBlockFetchServerEvent
+  , TraceBlockFetchServerEvent(..)
     -- * Exceptions
   , BlockFetchServerException
   ) where
 
-import           Control.Tracer (Tracer)
+import           Control.Tracer (Tracer, traceWith)
 import           Data.Typeable (Typeable)
 
 import           Ouroboros.Network.Block (Serialised (..))
@@ -76,7 +76,7 @@ blockFetchServer
     -> NodeToNodeVersion
     -> ResourceRegistry m
     -> BlockFetchServer (Serialised blk) (Point blk) m ()
-blockFetchServer _tracer chainDB _version registry = senderSide
+blockFetchServer tracer chainDB _version registry = senderSide
   where
     senderSide :: BlockFetchServer (Serialised blk) (Point blk) m ()
     senderSide = BlockFetchServer receiveReq' ()
@@ -114,7 +114,8 @@ blockFetchServer _tracer chainDB _version registry = senderSide
     sendBlocks it = do
       next <- ChainDB.iteratorNext it
       case next of
-        IteratorResult blk     ->
+        IteratorResult blk     -> do
+          traceWith tracer TraceBlockFetchServerSendBlock
           return $ SendMsgBlock (withoutPoint blk) (sendBlocks it)
         IteratorExhausted      -> do
           ChainDB.iteratorClose it
@@ -129,7 +130,7 @@ blockFetchServer _tracer chainDB _version registry = senderSide
 -------------------------------------------------------------------------------}
 
 -- | Events traced by the Block Fetch Server.
-data TraceBlockFetchServerEvent blk
-   -- TODO no events yet. Tracing the messages send/received over the network
-   -- might be all we need?
+data TraceBlockFetchServerEvent blk =
+    -- | The server sent a block to the peer.
+    TraceBlockFetchServerSendBlock
   deriving (Eq, Show)
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE LambdaCase          #-}
-{-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies        #-}
 
@@ -118,7 +117,7 @@ chainSyncServerForFollower tracer chainDB flr =
         tip <- atomically $ ChainDB.getCurrentTip chainDB
         traceWith tracer $
           TraceChainSyncServerRead tip (point <$> update)
-        return $ Left $ sendNext tip (withoutPoint <$> update)
+        Left <$> sendNext tip update
       Nothing     -> return $ Right $ do
         -- Follower is at the head, we have to block and wait for the chain to
         -- change.
@@ -126,14 +125,16 @@ chainSyncServerForFollower tracer chainDB flr =
         tip    <- atomically $ ChainDB.getCurrentTip chainDB
         traceWith tracer $
           TraceChainSyncServerReadBlocked tip (point <$> update)
-        return $ sendNext tip (withoutPoint <$> update)
+        sendNext tip update
 
     sendNext :: Tip blk
-             -> ChainUpdate blk b
-             -> ServerStNext b (Point blk) (Tip blk) m ()
+             -> ChainUpdate blk (WithPoint blk b)
+             -> m (ServerStNext b (Point blk) (Tip blk) m ())
     sendNext tip update = case update of
-      AddBlock hdr -> SendMsgRollForward  hdr tip idle'
-      RollBack pt  -> SendMsgRollBackward pt tip idle'
+      AddBlock hdr -> do
+        traceWith tracer (TraceChainSyncRollForward (point hdr))
+        return $ SendMsgRollForward (withoutPoint hdr) tip idle'
+      RollBack pt  -> return $ SendMsgRollBackward pt tip idle'
 
     handleFindIntersect :: [Point blk]
                         -> m (ServerStIntersect b (Point blk) (Tip blk) m ())
@@ -141,9 +142,9 @@ chainSyncServerForFollower tracer chainDB flr =
       -- TODO guard number of points
       changed <- ChainDB.followerForward flr points
       tip     <- atomically $ ChainDB.getCurrentTip chainDB
-      return $ case changed of
-        Just pt -> SendMsgIntersectFound    pt tip idle'
-        Nothing -> SendMsgIntersectNotFound    tip idle'
+      case changed of
+        Just pt -> return $ SendMsgIntersectFound pt tip idle'
+        Nothing -> return $ SendMsgIntersectNotFound tip idle'
 
 {-------------------------------------------------------------------------------
   Trace events
@@ -156,4 +157,6 @@ chainSyncServerForFollower tracer chainDB flr =
 data TraceChainSyncServerEvent blk
   = TraceChainSyncServerRead        (Tip blk) (ChainUpdate blk (Point blk))
   | TraceChainSyncServerReadBlocked (Tip blk) (ChainUpdate blk (Point blk))
+  | TraceChainSyncRollForward       (Point blk)
+  | TraceChainSyncRollBackward      (Point blk)
   deriving (Eq, Show)
--- a/ouroboros-network/src/Ouroboros/Network/Counter.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Counter.hs
@@ -1,0 +1,8 @@
+module Ouroboros.Network.Counter (
+    Counter
+  ) where
+
+import Control.Tracer
+
+-- | Type alias to 
+type Counter m a = Tracer m a
--- a/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
+++ b/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
@@ -31,7 +31,7 @@ import           Control.Monad (unless)
 import           Control.Monad.Class.MonadSTM
 import           Control.Monad.Class.MonadSTM.Strict (checkInvariant)
 import           Control.Monad.Class.MonadThrow
-import           Control.Tracer (Tracer)
+import           Control.Tracer (Tracer, traceWith)
 
 import           Network.TypedProtocol.Pipelined (N, Nat (..))
 
@@ -63,7 +63,13 @@ data TxSubmissionMempoolWriter txid tx idx m =
        mempoolAddTxs :: [tx] -> m [txid]
     }
 
-data TraceTxSubmissionInbound txid tx = TraceTxSubmissionInbound --TODO
+data TraceTxSubmissionInbound txid tx =
+      -- | Transactions just about to be inserted.
+      TraceTxSubmissionCollected !Int
+      -- | Just accepted this many transactions.
+    | TraceTxSubmissionAccepted !Int
+      -- | Just rejected this many transactions.
+    | TraceTxSubmissionRejected !Int
   deriving Show
 
 data TxSubmissionProtocolError =
@@ -163,7 +169,7 @@ txSubmissionInbound
   -> TxSubmissionMempoolWriter txid tx idx m
   -> NodeToNodeVersion
   -> TxSubmissionServerPipelined txid tx m ()
-txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
+txSubmissionInbound tracer maxUnacked mpReader mpWriter _version =
     TxSubmissionServerPipelined $
       continueWithStateM (serverIdle Zero) initialServerState
   where
@@ -324,8 +330,13 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
             bufferedTxs3 = forceElemsToWHNF $ bufferedTxs2 <>
                                (Map.fromList (zip live (repeat Nothing)))
 
+        let !collected = length txs
 
-        _writtenTxids <- mempoolAddTxs txsReady
+        txidsAccepted <- mempoolAddTxs txsReady
+
+        let !accepted = length txidsAccepted
+        traceWith tracer $ TraceTxSubmissionAccepted accepted
+        traceWith tracer $ TraceTxSubmissionRejected (collected - accepted)
 
         continueWithStateM (serverIdle n) st {
           bufferedTxs         = bufferedTxs3,
--- a/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
+++ b/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
@@ -6,6 +6,8 @@
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
+{-# OPTIONS_GHC -Wno-partial-fields #-}
+
 module Ouroboros.Network.TxSubmission.Inbound (
     txSubmissionInbound,
     TxSubmissionMempoolWriter(..),
@@ -31,10 +33,11 @@ import           Control.Monad (unless)
 import           Control.Monad.Class.MonadSTM
 import           Control.Monad.Class.MonadSTM.Strict (checkInvariant)
 import           Control.Monad.Class.MonadThrow
-import           Control.Tracer (Tracer, traceWith)
+import           Control.Tracer (traceWith)
 
 import           Network.TypedProtocol.Pipelined (N, Nat (..))
 
+import           Ouroboros.Network.Counter (Counter)
 import           Ouroboros.Network.NodeToNode.Version (NodeToNodeVersion)
 import           Ouroboros.Network.Protocol.TxSubmission.Server
 import           Ouroboros.Network.TxSubmission.Mempool.Reader
@@ -63,14 +66,20 @@ data TxSubmissionMempoolWriter txid tx idx m =
        mempoolAddTxs :: [tx] -> m [txid]
     }
 
-data TraceTxSubmissionInbound txid tx =
-      -- | Transactions just about to be inserted.
-      TraceTxSubmissionCollected !Int
+data ProcessedTxCount = ProcessedTxCount {
       -- | Just accepted this many transactions.
-    | TraceTxSubmissionAccepted !Int
+      ptxcAccepted :: !Int
       -- | Just rejected this many transactions.
-    | TraceTxSubmissionRejected !Int
-  deriving Show
+    , ptxcRejected :: !Int
+    }
+  deriving (Eq, Show)
+
+data TraceTxSubmissionInbound txid tx =
+    -- | Number of transactions just about to be inserted.
+    TraceTxSubmissionCollected !Int
+    -- | Just processed transaction pass/fail breakdown.
+  | TraceTxSubmissionProcessed !ProcessedTxCount
+  deriving (Eq, Show)
 
 data TxSubmissionProtocolError =
        ProtocolErrorTxNotRequested
@@ -163,7 +172,7 @@ txSubmissionInbound
      , MonadSTM m
      , MonadThrow m
      )
-  => Tracer m (TraceTxSubmissionInbound txid tx)
+  => Counter m (TraceTxSubmissionInbound txid tx)
   -> Word16         -- ^ Maximum number of unacknowledged txids allowed
   -> TxSubmissionMempoolReader txid tx idx m
   -> TxSubmissionMempoolWriter txid tx idx m
@@ -331,12 +340,17 @@ txSubmissionInbound tracer maxUnacked mpReader mpWriter _version =
                                (Map.fromList (zip live (repeat Nothing)))
 
         let !collected = length txs
+        traceWith tracer $
+          TraceTxSubmissionCollected collected
 
         txidsAccepted <- mempoolAddTxs txsReady
 
         let !accepted = length txidsAccepted
-        traceWith tracer $ TraceTxSubmissionAccepted accepted
-        traceWith tracer $ TraceTxSubmissionRejected (collected - accepted)
+
+        traceWith tracer $ TraceTxSubmissionProcessed ProcessedTxCount {
+            ptxcAccepted = accepted
+          , ptxcRejected = collected - accepted
+          }
 
         continueWithStateM (serverIdle n) st {
           bufferedTxs         = bufferedTxs3,
--- a/ouroboros-network/ouroboros-network.cabal
+++ b/ouroboros-network/ouroboros-network.cabal
@@ -52,6 +52,7 @@ library
                        Ouroboros.Network.BlockFetch.Decision
                        Ouroboros.Network.BlockFetch.DeltaQ
                        Ouroboros.Network.BlockFetch.State
+                       Ouroboros.Network.Counter
                        Ouroboros.Network.DeltaQ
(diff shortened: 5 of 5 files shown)


nfrisby:
@newhoggy I think this is for input-output-hk/cardano-node#2261, is that right?

The existing "metric" pattern seems to do the accumulation of counters in `cardano-node:Cardano.Tracing.Metrics`. So the Consensus Layer should just emit the "I just now served a block" event, and the downstream "metric" will interpret that by atomically incrementing a counter. And so on for the similar metrics.

nfrisby:
As you noticed, there is no "I just now served a block" event currently emitted. However, we do trace every mini protocol message sent/received, via the following tracers.

https://github.com/input-output-hk/ouroboros-network/blob/389a00b5c2b89b48aa98a96ca10f72f085b6e32c/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs#L319-L326

And they can be set here. (I think you can ignore the `Serialised` variants. I don't yet know what `TxSub2` is about.)

https://github.com/input-output-hk/ouroboros-network/blob/389a00b5c2b89b48aa98a96ca10f72f085b6e32c/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L122-L126

So in a sense the metrics could use those. But maybe more precise events could be added. @dcoutts, what do you think?

nfrisby:
Are you counting roll forwards and rollbacks for the sake of https://jira.iohk.io/browse/CAD-2371? If so, I think you can instead use the existing `SwitchedToAFork` event. I elaborated in my comments on that Jira Issue -- I think it has sufficient info for that task.

newhoggy:
This PR is actually for https://jira.iohk.io/browse/CAD-2487.  Do my changes fit those use cases?

newhoggy:
I rebased the PR to an earlier point in history so I don't have to refactor anything when working on `cardano-node`.

nfrisby:
> I understood roll forward to mean a header was served in chain sync. Is that an accurate interpretation or am I way off base?

I think that's accurate, yes. (My suggestion of  `SwitchedToAFork` above was because I thought this PR might also relate to CAD-2371.)

The only caveat to that interpretation is that we might send the same header to the same peer multiple times. I think this is ultimately a caveat on the wording of the Issue  input-output-hk/cardano-node#2261: it's just slightly ambiguous.

My guess is that these metrics should indeed count each send of a repeated header, in which case I agree with your interpretation: every `RollForward` message contains exactly one block header. (A `RollBackward` includes neither a header nor a block -- it's just "point".)

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
```diff
@@ -107,17 +107,18 @@ blockFetchServer _tracer chainDB _version registry = senderSide
...
         -- its bounds are inclusive, so we don't have to check whether the
         -- iterator is empty.
-        Right it -> SendMsgStartBatch $ sendBlocks it
+        Right it -> SendMsgStartBatch $ sendBlocks 0 it
 
-    sendBlocks :: ChainDB.Iterator m blk (WithPoint blk (Serialised blk))
+    sendBlocks :: Int -> ChainDB.Iterator m blk (WithPoint blk (Serialised blk))
                -> m (BlockFetchSendBlocks (Serialised blk) (Point blk) m ())
-    sendBlocks it = do
+    sendBlocks blockCount it = do
```
karknu, line 114: Shouldn't this be `!blockCount`?

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -156,4 +163,8 @@ chainSyncServerForReader tracer chainDB rdr =
 data TraceChainSyncServerEvent blk
   = TraceChainSyncServerRead        (Tip blk) (ChainUpdate blk (Point blk))
   | TraceChainSyncServerReadBlocked (Tip blk) (ChainUpdate blk (Point blk))
+  | TraceChainSyncRollForward
```
coot, line 166: Shouldn't we log a bit more, e.g. `Point blk`? Similar question for other messages.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
newhoggy: Fixed.  Thanks!

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
newhoggy: I added `Tip`.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
nfrisby: I think we shouldn't do the counting in ChainSync server. That's an example of something that should happen downstream, in the metric logic that processes the "I served a block" messages. (See my similar comment here https://github.com/input-output-hk/ouroboros-network/pull/2874#issuecomment-762549330 )

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
```diff
@@ -107,17 +108,18 @@ blockFetchServer _tracer chainDB _version registry = senderSide
...
         -- its bounds are inclusive, so we don't have to check whether the
         -- iterator is empty.
-        Right it -> SendMsgStartBatch $ sendBlocks it
+        Right it -> SendMsgStartBatch $ sendBlocks 0 it
 
-    sendBlocks :: ChainDB.Iterator m blk (WithPoint blk (Serialised blk))
+    sendBlocks :: Int -> ChainDB.Iterator m blk (WithPoint blk (Serialised blk))
                -> m (BlockFetchSendBlocks (Serialised blk) (Point blk) m ())
-    sendBlocks it = do
+    sendBlocks !blockCount it = do
```
nfrisby, line 115: Please remove this count; the downstream tracer in `cardano-node` is doing the counting there.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -265,6 +269,7 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
         let st' = st {
           requestedTxIdsInFlight = requestedTxIdsInFlight st - reqNo
         }
+
```
nfrisby, line 272: Please remove this newline.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -324,6 +329,12 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
             bufferedTxs3 = forceElemsToWHNF $ bufferedTxs2 <>
                                (Map.fromList (zip live (repeat Nothing)))
 
+        writtenTxids <- mempoolAddTxs txsReady
```
nfrisby, line 332: This call already exists 7 lines below :)

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -324,6 +329,12 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
+
+        traceWith tracer $ TraceTxSubmissions
+          { traceTxsWritten = length writtenTxids
```
nfrisby, line 335: Please align the `=` signs

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -63,7 +63,11 @@ data TxSubmissionMempoolWriter txid tx idx m =
        mempoolAddTxs :: [tx] -> m [txid]
     }
 
-data TraceTxSubmissionInbound txid tx = TraceTxSubmissionInbound --TODO
+data TraceTxSubmissionInbound txid tx
+  = TraceTxSubmissions
+    { traceTxsCollected :: !Int
```
nfrisby, line 68: Please change it to the following, to match the style in this module.

```
data TraceTxSubmissionInbound txid tx =
    TraceTxSubmissions {
      traceTxsCollected :: !Int,
      traceTxsWritten   :: !Int
    }
  deriving Show
```

Thanks.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
```diff
@@ -107,17 +108,18 @@ blockFetchServer _tracer chainDB _version registry = senderSide
...
       next <- ChainDB.iteratorNext it
       case next of
         IteratorResult blk     ->
-          return $ SendMsgBlock (withoutPoint blk) (sendBlocks it)
+          return $ SendMsgBlock (withoutPoint blk) (sendBlocks (blockCount + 1) it)
```
nfrisby, line 119: I think you want to put the `traceWith` here: one event per block sent (no `Int`).

karknu (approved):
LGTM.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
```diff
@@ -129,7 +131,5 @@ blockFetchServer _tracer chainDB _version registry = senderSide
 -------------------------------------------------------------------------------}
 
 -- | Events traced by the Block Fetch Server.
-data TraceBlockFetchServerEvent blk
-   -- TODO no events yet. Tracing the messages send/received over the network
-   -- might be all we need?
+newtype TraceBlockFetchServerEvent blk = TraceBlockFetchServerBlockCount Int
```
nfrisby, line 134: Please remove the `Int`, add a docstring, use the following style, and use `data`. I know it has one constructor, but that's just a coincidence.

```
data TraceBlockFetchServerEvent blk =
    -- | The server sent a block to the peer.
    TraceBlockFetchServerSendBlock
```

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -1,6 +1,5 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE LambdaCase          #-}
-{-# LANGUAGE NamedFieldPuns      #-}
```
nfrisby, line 3: Nice catch!

nfrisby (dismissed):
@newhoggy I requested changes. Some are about content and some are about style. We're stingy about style on the Consensus Team, but you're not a Consensus Team member... so please let me know if you'd rather just have be re-style your commits once we agree on the content.

@coot I lost track of the Consensus-vs-Network divide and reviewed the whole diff. But please take a look at your modules too.

Edit: I reworded slightly to clarify what I meant to say about styling.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
coot: `Tip` will be confusing, on the server this will not be tip of the chain, I'd use just a `Point`.  @nfrisby what do you think?

dcoutts:
It is indeed the case that we should just trace the events, and not do any counting. The node top level is where we convert traces to metrics (including counters).

For the current purpose of counter metrics we do not need any additional info in the trace events, so we should avoid including anything that's not trivially cheap.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
newhoggy: Ah ok.  The reason I did the count rather than trace once per message was because I was concerned about the performance implications of emitting an event per block.  If that's not a concern, then it makes sense to do as you say and emit an event per block.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
newhoggy: Fixed.  Thanks!

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
newhoggy: Removed.  Thanks!

newhoggy:
> It is indeed the case that we should just trace the events, and not do any counting. The node top level is where we convert traces to metrics (including counters).
> 
> For the current purpose of counter metrics we do not need any additional info in the trace events, so we should avoid including anything that's not trivially cheap.
> 
> And yes, since the purpose is to get a metric that is a proxy for the load, then we're interested in all headers we serve, duplicates and all. It's all load. We should use a different message for roll back vs header served, but when we convert to a metric we may well just aggregate them for the load metric.

I've just removed the counting, but reading your message again, it's not clear if I should have done that.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
newhoggy: Thanks!

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
newhoggy: I was not able to find a way to get a `Point` for `TraceChainSyncRollForward`.  Can you suggest how I can get one?

nfrisby: `Ouroboros.Network.Block.getTipPoint` should do it.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -6,6 +6,8 @@
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
+{-# OPTIONS_GHC -Wno-partial-fields #-}
```
nfrisby, line 9: We avoid disabling this one; I'll comment again near the use case.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -63,7 +65,16 @@ data TxSubmissionMempoolWriter txid tx idx m =
        mempoolAddTxs :: [tx] -> m [txid]
     }
 
-data TraceTxSubmissionInbound txid tx = TraceTxSubmissionInbound --TODO
```
nfrisby, line 66: We avoid partial fields (I'm pretty sure that's as true for the Network Layer as it is from Consensus; @coot?)

You have a few options, but here's a relatively simple one.

```
data TraceTxSubmissionInbound txid tx =
      -- | Transactions just about to be inserted.
      TraceTxSubmissionCollected !Int
      -- | Just accepted this many transactions.
    | TraceTxSubmissionAccepted !Int
      -- | Just rejected this many transactions.
    | TraceTxSubmissionRejected !Int
```

You'd have two `traceWiths` back-to-back after the `mempoolAddTxs` call, which is a bit weird but not a giant problem. If you want to avoid that, I'd recommend defining a small single-constructor record type and passing that as thes ole argument to the merged `*Processed` event.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -118,32 +117,40 @@ chainSyncServerForReader tracer chainDB rdr =
...
+             -> m (ServerStNext b (Point blk) (Tip blk) m ())
     sendNext tip update = case update of
-      AddBlock hdr -> SendMsgRollForward  hdr tip idle'
-      RollBack pt  -> SendMsgRollBackward pt tip idle'
+      AddBlock hdr -> do
+        traceWith tracer (TraceChainSyncRollForward tip)
+        return $ SendMsgRollForward hdr tip idle'
+      RollBack pt  -> do
+        traceWith tracer (TraceChainSyncRollBackward tip)
+        return $ SendMsgRollBackward pt tip idle'
```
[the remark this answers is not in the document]
nfrisby: I don't see a commit from Karl. Did you mean Serge?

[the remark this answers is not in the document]
nfrisby: Also, I don't see any of those arguments on the `TraceChainSyncRollBackward`. (Which is fine as-is in my opinoin; I'm just clarifying that I don't follow your comment here.)

nfrisby:
The Consensus modules look good to me. Thanks for tightening down the diff 👍

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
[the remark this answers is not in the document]
newhoggy: Oops, yeah.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
newhoggy: This will cost has a non-zero runtime cost, but it's probably small and excellent idea for avoiding the warning.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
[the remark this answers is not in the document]
newhoggy: > Also, I don't see any of those arguments on the `TraceChainSyncRollBackward`. (Which is fine as-is in my opinion; I'm just clarifying that I don't follow your comment here.)

I misunderstood the diffs at the time I wrote the comment, but my comment about the extra fields were in error.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
newhoggy: Done!

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
newhoggy: Removed!

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -324,8 +332,15 @@ txSubmissionInbound _tracer maxUnacked mpReader mpWriter _version =
...
+        let !collected = length txs
+        traceWith tracer $
+          TraceTxSubmissionCollected collected
+
+        txidsAccepted <- mempoolAddTxs txsReady
 
-        _writtenTxids <- mempoolAddTxs txsReady
+        let !accepted = length txidsAccepted
+        traceWith tracer $ TraceTxSubmissionAccepted accepted
+        traceWith tracer $ TraceTxSubmissionRejected (collected - accepted)
```
coot, line 343: Maybe you'd like to only emit this when `collected > accepted`?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -63,7 +65,13 @@ data TxSubmissionMempoolWriter txid tx idx m =
        mempoolAddTxs :: [tx] -> m [txid]
     }
 
-data TraceTxSubmissionInbound txid tx = TraceTxSubmissionInbound --TODO
+data TraceTxSubmissionInbound txid tx =
+      -- | Transactions just about to be inserted.
```
coot, line 69:
```suggestion
      -- | Number of transactions just about to be inserted.
```

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
deepfire: @newhoggy, @nfrisby, I think in this case we could just drop the field names to avoid partial fields.

I'm really conscious of causing two traces in place of one.

nfrisby: > I'm really conscious of causing two traces in place of one.

I sympathize with that. I called it "weird but not a giant problem", but it does seem sloppy to me. I've assumed it wouldn't have a significant cost in a full node, but maybe my intuition is off for eg benchmarking use cases, maybe it adds up here and there, etc.

> I think in this case we could just drop the field names to avoid partial fields.

I would push back on `TraceTxSubmissionProcessed !Int !Int` -- that seems like a food gun to mix up the two `Int`s. So I think it would be worthwhile to introduce the

```
data ProcessedTxCount = ProcessedTxCount {
      ptcAccepted :: !Int
    , ptcRejected :: !Int
    }
```

coot:
@newhoggy I misunderstood this PR, possibly because the title is misleading.  So what you want to have are counters and we can leverage the contra-tracer to do that.  I'd suggest we have an enumeration type, each constructor would correspond to a counter, e.g.

```
data TxSubmissionCounters
  = CountCollected
  | CountAccepted
  | CountRejected
```

Then `txSubmissionInbound` would take additional argument `Tracer m TxSubmissionCounters`, we should use an alias
```
type Counter m a = Tracer m a
```
and use it instead of `Tracer`, you could add it to `Ouroboros.Network.Counter` or even better, `cardano-prelude` could provide it (or even `contra-tracer`: the argument to add it there would be to include a haddock on the type alise which describes how a `Tracer` can be used as a counter, this way we could sell it to `contra-tracer` package I think).

You would instrument it in a similar way to what @karknu suggested in slack, e.g.
```
newTxSubmissionCounters :: m (Counter m TxSubmissionCounters)
newTxSubmissionCounters = do
  collected <- newCounter -- e.g. `getCounter` from `ekg`...
  accepted <- newCounter
  rejected <- newCounter
  return $ Tracer \case
    CountCollected -> incCounter collected -- e.g. `inc` from `ekg` ...
    CountAccepted -> incCounter accepted
    CountRejected -> incCounter rejected
    ```



You should also create something similar to `Ouroboros.Consensus.Tracers` and pass the record of all counters around, @nfrisby can guide you.

**Comment by deepfire** [comment]
@coot, I think that we want to count transactions accepted/rejected, not messages (each of which corresponds to a number of those).

So yes, we have to use a counter, but not using the message counter transformer.

Hence my strong suggestion to coalesce the Accepted/Rejected into a single trace, to reduce volume.

File: ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs (hunk 12 above)
Comment (deepfire) [reply ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs:66]: > So I think it would be worthwhile to introduce the separate type

:+1: on that!

**Comment by coot** [comment]
@deepfire that's ok, I updated my comment.

**Comment by nfrisby** [comment]
@marcin After drafting a small wall of text, I realized that you're suggesting that we should use two different "event" types. One `data` type for a set of events for "an interesting thing happened" (like the `TraceEvent` data types we already have) and a one for "something relatively boring happened but we need to notify downstream" (like these new counters). Does that sound like a slight concretization of what you're suggesting? Thanks.

My gut tells me the two sets of events might overlap and that it won't be trivial to determine where any specific event should belong. So I'm tempted to just lump them into one event type and rely on the event-processing logic to distinguish them on a case-by-case basis. But maybe you three have more experience and are convinced that there will be no overlap and/or it'll be easy to distinguish which event belongs in which set.

**Comment by coot** [comment]
@nfrisby that's a good point.  What I am worried about is that when setting logging output off by passing `nullTracer` (which is possible though node configuration), one will accidentally remove metrics or vice versa.  Also tracing is causing quite a bit cpu usage, if we turn them more of them on this will cause more problems.  I don't mind if there are logging tracers that are used for statistics as well, I just prefer the default to be keep them separate, and if not put them on the logging side.

File: ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs (hunk 14 above)
Comment (newhoggy) [reply ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs]: Excellent idea.  Done!

File: ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs (hunk 15 above)
Comment (newhoggy) [reply ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs]: Fixed.  Thanks!

File: ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs (hunk 12 above)
Comment (newhoggy) [reply ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs:66]: Done.  Thanks!

**Comment by nfrisby** [comment]
> What I am worried about is that when setting logging output off by passing `nullTracer` (which is possible though node configuration), one will accidentally remove metrics or vice versa.

Ah, I see. That seems like a good reason to have two different inputs in the Consensus Layer's interface (say, "loggingTracers" vs "metricTracers" or maybe "countingTracers") , which at some point in the Consensus Layer implementation get combined via `<>`. They could reuse the same event type, which would avoid the concern I shared above about having to classify each event.

That distinction/duplication could be reflected in the `-node` UX too.

**Comment by newhoggy** [comment]
I've separated the counters from the tracers.  How does it look now?

File: ouroboros-consensus/src/Ouroboros/Consensus/Node.hs [hunk 16]
```diff
@@ -117,7 +118,10 @@ import           Ouroboros.Consensus.Storage.VolatileDB
 -- | Arguments expected from any invocation of 'runWith'
 data RunNodeArgs m addrNTN addrNTC blk = RunNodeArgs {
       -- | Consensus tracers
```
deepfire, line 120: Shouldn't this be retitled to `Consensus counters`?

newhoggy: Fixed.  Thanks!

deepfire:
@newhoggy, @nfrisby, I'm afraid that the `Counter` abstraction is making me feel ill at ease -- for several reasons:

1. It violates the principle of separation of concerns that the contravariant tracers establish -- the library user _just posts the information_ -- the policy (interpretation) happens elsewhere.

2. The separation of concerns allows you to concentrate the policy definition -- whereas if you violate it, you disperse that definition across all source points.

3. Another consequence is that tomorrow you can see someone else willing to enrich that trace -- and it will no longer be a `Counter` -- it already barely is (are two counters bundled together still a counter?  Should it be called `Counters` instead?).

cc @dcoutts

newhoggy:
I'll back out those changes and merge, then create a new PR for the refactoring because there is concern about getting the abstraction right and because the refactoring adds downstream refactoring risk.

newhoggy:
I've moved the refactoring here: https://github.com/input-output-hk/ouroboros-network/pull/2902

newhoggy:
bors merge

nfrisby:
I like the decision to spit the PR into two separate decisions. 👍 

Unfortunately, just to clarify for future reference, it looks like `Ouroboros.Network.Counter` was still included in this PR and used in `Ouroboros.Network.TxSubmission.Inbound`. It's all local, so not a huge problem.

But if the work on the follow-up PR #2902 doesn't proceed very quickly, let's go ahead and remove that essentially unused `Counter` stuff in a small fixup PR. Thanks.

ouroboros-network/src/Ouroboros/Network/Counter.hs
```diff
@@ -0,0 +1,8 @@
+module Ouroboros.Network.Counter (
+    Counter
+  ) where
+
+import Control.Tracer
+
+-- | Type alias to 
+type Counter m a = Tracer m a
```
deepfire, line 8: @newhoggy, does this still belong in this PR?

nfrisby: I assumed it was an accident in my previous comment. (Serge, this PR is already Merged.)

https://github.com/input-output-hk/ouroboros-network/pull/2874#issuecomment-768667380

newhoggy: Will be removed in a new PR.  Thanks!

PR #2884, opened 2021-01-19
Change package descriptions so that we don't run tests of your external deps

EncodePanda:
Currently, when running `cabal test all` we are running also tests of
our external dependencies. This commit modifies this behavior, tests
of our dependencies will be ignored.
This is unfortunately manual work - newly added dependencies will have to
adjust to this approach. Rumor has it that latest cabal fixes it, thus
if one day we will switch to cabal-3.4.0.0 why might get rid of all
those entries.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -101,7 +101,7 @@ package cardano-crypto-class
 package cardano-crypto-wrapper
   tests: False
 
-package cardano-prelude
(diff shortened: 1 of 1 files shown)


nfrisby (approved):
Yep, makes sense. Thanks 👍

PR #2886, opened 2021-01-19
Rollforward counter

newhoggy:
This is for "Counter, monotonically increasing, number of block headers by chain sync served" as mentioned in https://github.com/input-output-hk/cardano-node/issues/2261

Given we need load metrics, it is correct to count duplicate headers.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE LambdaCase          #-}
-{-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies        #-}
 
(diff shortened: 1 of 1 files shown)


disassembler:
bors try

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -156,4 +163,8 @@ chainSyncServerForReader tracer chainDB rdr =
 data TraceChainSyncServerEvent blk
   = TraceChainSyncServerRead        (Tip blk) (ChainUpdate blk (Point blk))
   | TraceChainSyncServerReadBlocked (Tip blk) (ChainUpdate blk (Point blk))
+  | TraceChainSyncRollForward
+  | TraceChainSyncRollBackward
```
nfrisby, line 167: Are we also planning metrics for the rest of these? I only see the `RollForward` event being used downstream.

nfrisby (approved):
@newhoggy My gut tells me to only add tracers as we need them. But on the other hand we make no stability promises about the tracer API, so we can remove them later if we decide they were dead weight after all.

Thanks for the PR.

nfrisby:
This is totally superceded by PR #2874, right? Let's close it if so.

coot (changes requested):
As I described [here](https://github.com/input-output-hk/ouroboros-network/pull/2874#issuecomment-766707113), I wouldn't mix traces with counters. I think it will be cleaner and it will give more option to control them at the top level if we keep them separate.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
newhoggy: I only intend to use `RollForward` for metrics at the moment.  Would you rather I delete the others?

newhoggy:
> This is totally superseded by PR #2874, right? Let's close it if so.

This is correct.  I will close.

PR #2888, opened 2021-01-20
Update to latest cardano-ledger-spec.

nc6:
This is something of a slapdash patch. Crucially, the latest changes in
the ledger are to support the generalisation of the update system for
privilege, and this means that the state of the update system will
change. At the moment, consensus drills deep into the update state to
work out when to execute a hard fork.

This PR locks down the update state to the Shelley update state. The
correct solution will be to work out exactly what consensus needs to
know and add an additional class to the API which provides this.

### DIFF
--- a/ouroboros-consensus-cardano-test/src/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs
+++ b/ouroboros-consensus-cardano-test/src/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs
@@ -43,7 +43,9 @@ import qualified Ouroboros.Consensus.HardFork.Combinator.Util.InPairs as InPairs
 import qualified Ouroboros.Consensus.HardFork.Combinator.Util.Tails as Tails
 import qualified Ouroboros.Consensus.HardFork.History as History
 
+import qualified Cardano.Ledger.Core as LC
 import qualified Cardano.Ledger.Era as SL
+import           Control.State.Transition (State)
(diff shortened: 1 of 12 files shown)


ouroboros-consensus-cardano-test/src/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs
```diff
@@ -130,11 +132,17 @@ type ShelleyBasedHardForkConstraints era1 era2 =
   , SL.TranslationContext era2 ~ ()
   )
 
-instance ShelleyBasedHardForkConstraints era1 era2
+instance ( ShelleyBasedHardForkConstraints era1 era2
+         , State (LC.EraRule "PPUP" era1) ~ SL.PPUPState era1
+         , State (LC.EraRule "PPUP" era2) ~ SL.PPUPState era2
+         )
```
nfrisby, line 138: The `StyleGuide.md` wants

```
instance ( Eq a
         , ..
         ) => C a where
```

That seems to be consistent with other instances already in this file, too. See for example

https://github.com/input-output-hk/ouroboros-network/blob/07b8d4ca6e22960d627411e1466a498110086d1e/ouroboros-consensus-cardano-test/src/Test/ThreadNet/Infra/ShelleyBasedHardFork.hs#L275

nc6:
Superseded by #2891

PR #2893, opened 2021-01-22
More explicit resource management in ChainSync client and server

### LINKED ISSUE #2892: Introduce the `bracketWithRegistry` combinator
During the post mortem for Issue #2870, we invented an idea to help prevent it in the future. That bug was due to confusion caused by the fact that we allocated a `registry` and also a narrower-manager (a `ChainDB` `Reader`/`Follower`) that in turn used that registry, but did not register itself in that registry, despite the registry and the manager having the same lifetime. We incorrectly assumed that `newReader` registered itself in the registry that it was passed.

So we're going to introduce a `ResourceRegistry` combinator for use in this situation, where a `bracket`-ed thing and a registry it needs share the same lifetime. This will let us be very explicit about that coincident lifetime, which should make it very obvious if future changes cause the lifetimes to become different.

nfrisby:
Fixes #2892.

### DIFF
--- a/ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
+++ b/ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
@@ -294,16 +294,18 @@ testOverrideDelay :: forall m. (IOLike m, MonadTime m, MonadDelay (OverrideDelay
                   -> Int  -- ^ Number of slots to collect
                   -> OverrideDelay m [SlotNo]
 testOverrideDelay systemStart slotLength maxClockRewind numSlots = do
-    withRegistry $ \registry -> do
-      time <- simpleBlockchainTime
-                registry
-                (defaultSystemTime systemStart nullTracer)
-                slotLength
-                maxClockRewind
+    bracketWithPrivateRegistry
+      (\registry -> simpleBlockchainTime
+                      registry
+                      (defaultSystemTime systemStart nullTracer)
+                      slotLength
+                      maxClockRewind)
+      (\_btime   -> pure ())
+      $ \btime   -> do
       slotsVar <- uncheckedNewTVarM []
       withWatcher
         "testOverrideDelay"
-        ( knownSlotWatcher time $ \slotNo -> do
+        ( knownSlotWatcher btime $ \slotNo -> do
             atomically $ modifyTVar slotsVar (slotNo :)
         ) $ do
         -- Wait to collect the required number of slots
--- a/ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
+++ b/ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
@@ -30,6 +30,7 @@ import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.BlockchainTime
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.ResourceRegistry
+import           Ouroboros.Consensus.Util.STM (withWatcher)
 import           Ouroboros.Consensus.Util.Time
 
 import           Test.Util.Orphans.Arbitrary ()
@@ -293,24 +294,23 @@ testOverrideDelay :: forall m. (IOLike m, MonadTime m, MonadDelay (OverrideDelay
                   -> Int  -- ^ Number of slots to collect
                   -> OverrideDelay m [SlotNo]
 testOverrideDelay systemStart slotLength maxClockRewind numSlots = do
-    result <- withRegistry $ \registry -> do
+    withRegistry $ \registry -> do
       time <- simpleBlockchainTime
                 registry
                 (defaultSystemTime systemStart nullTracer)
                 slotLength
                 maxClockRewind
       slotsVar <- uncheckedNewTVarM []
-      cancelCollection <-
-        onKnownSlotChange registry time "testOverrideDelay" $ \slotNo ->
-          atomically $ modifyTVar slotsVar (slotNo :)
-      -- Wait to collect the required number of slots
-      slots <- atomically $ do
-        slots <- readTVar slotsVar
-        when (length slots < numSlots) $ retry
-        return slots
-      cancelCollection
-      return $ reverse slots
-    return result
+      withWatcher
+        "testOverrideDelay"
+        ( knownSlotWatcher time $ \slotNo -> do
+            atomically $ modifyTVar slotsVar (slotNo :)
+        ) $ do
+        -- Wait to collect the required number of slots
+        atomically $ do
+          slots <- readTVar slotsVar
+          when (length slots < numSlots) $ retry
+          return $ reverse slots
 
 {-------------------------------------------------------------------------------
   Test-programmable time
--- a/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/API.hs
@@ -9,17 +9,15 @@
 module Ouroboros.Consensus.BlockchainTime.API (
     BlockchainTime(..)
   , CurrentSlot(..)
-  , onKnownSlotChange
+  , knownSlotWatcher
   ) where
 
 import           GHC.Generics (Generic)
-import           GHC.Stack
 import           NoThunks.Class (OnlyCheckWhnfNamed (..))
 
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Util.IOLike
-import           Ouroboros.Consensus.Util.ResourceRegistry
-import           Ouroboros.Consensus.Util.STM (Watcher (..), forkLinkedWatcher)
+import           Ouroboros.Consensus.Util.STM (Watcher (..))
 
 {-------------------------------------------------------------------------------
   API
@@ -54,27 +52,22 @@ data CurrentSlot =
   Derived functionality
 -------------------------------------------------------------------------------}
 
--- | Spawn a thread to run an action each time the slot changes
+-- | Watches for changes in the current slot
 --
 -- The action will not be called until the current slot becomes known
 -- (if the tip of our ledger is too far away from the current wallclock time,
--- we may not know what the current 'SlotId' is).
---
--- Returns a handle to kill the thread.
-onKnownSlotChange :: forall m. (IOLike m, HasCallStack)
-                  => ResourceRegistry m
-                  -> BlockchainTime m
-                  -> String            -- ^ Label for the thread
-                  -> (SlotNo -> m ())  -- ^ Action to execute
-                  -> m (m ())
-onKnownSlotChange registry btime label notify =
-      fmap cancelThread
-    $ forkLinkedWatcher registry label Watcher {
-          wFingerprint = id
-        , wInitial     = Nothing
-        , wNotify      = notify
-        , wReader      = getCurrentSlot'
-        }
+-- we may not know what the current 'SlotNo' is).
+knownSlotWatcher :: forall m. IOLike m
+                 => BlockchainTime m
+                 -> (SlotNo -> m ())  -- ^ Action to execute
+                 -> Watcher m SlotNo SlotNo
+knownSlotWatcher btime notify =
+    Watcher {
+        wFingerprint = id
+      , wInitial     = Nothing
+      , wNotify      = notify
+      , wReader      = getCurrentSlot'
+      }
   where
     getCurrentSlot' :: STM m SlotNo
     getCurrentSlot' = do
--- a/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/API.hs
@@ -19,7 +19,7 @@ import           NoThunks.Class (OnlyCheckWhnfNamed (..))
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.ResourceRegistry
-import           Ouroboros.Consensus.Util.STM (onEachChange)
+import           Ouroboros.Consensus.Util.STM (Watcher (..), forkLinkedWatcher)
 
 {-------------------------------------------------------------------------------
   API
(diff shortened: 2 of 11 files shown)


nfrisby:
@edsko This is what we talked about on Wednesday; let's scan it on our Friday call.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -587,6 +588,32 @@ releaseResources rr keys releaser = do
...
+     -- not themselves registered, since they're simply 'bracket'ed.
+  -> (a -> m ())
+     -- ^ Release the bracketed resources.
+  -> (ResourceRegistry m -> a -> m r)
+  -> m r
+bracketWithRegistry newA closeA body =
+    -- It would be OK to fuse these two calls to 'bracket', but it's simpler and
+    -- less error-prone this way.
+    withRegistry $ \registry ->
+      bracket (newA registry) closeA $ \a ->
```
kderme, line 614: I noticed this pr and #2892 while working on some changes at the ChainSync that we need for db-sync. I think in general blending brackets and registries can be tricky, basically because of the order of deallocation. Here `a` will be freed before anything else allocated with the registry. By using a simple registry instead, things are freed from younger to older.

kderme: I'm not sure if this is a real issue for Reader/Follower, but may be worth at least documenting it for `bracketWithRegistry`.

edsko: The point is that this is a _new_ registry: we are constructing something that for some reason needs a registry (example: the reader), but really the scope of that thing and the registry are coupled. Notice that _nothing else_ has access to this registry: the only thing that has access to it is the new resource itself. Hence, deallocating that resource before anything else in the registry is okay. 

What you are saying is true nonetheless: _if_ we then later allocate a new resource that somehow refers to this new resource (i.e., a reference in the reverse direction), we might be in trouble. That kind of "cyclic dependency" graph would indeed be potentially problematic.

kderme: Yes this is the usual case: newer resources reference older resources. That's why the younger to older deallocation that's done by Registry is the normal thing to do. And it's a kinda violated by the bracket.

edsko: Yes, you are correct. It would indeed be good to state the side condition here: those newer conditions should not be referring to the resources that is allocated by this new abstraction.

edsko: The point is that the lifetime of this new resource _is equal to_ the lifetime of the registry.

edsko: Good observation!

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -587,6 +588,32 @@ releaseResources rr keys releaser = do
...
+     -- ^ Allocate the bracketed resources. They depend on the registry but are
+     -- not themselves registered, since they're simply 'bracket'ed.
+  -> (a -> m ())
+     -- ^ Release the bracketed resources.
+  -> (ResourceRegistry m -> a -> m r)
+  -> m r
+bracketWithRegistry newA closeA body =
+    -- It would be OK to fuse these two calls to 'bracket', but it's simpler and
+    -- less error-prone this way.
+    withRegistry $ \registry ->
```
nfrisby, line 613: We can side-step the precondition that @kderme's noticed (nice catch!) by simply allocating the `a` in the registry. This combinator ensures it's the oldest thing in the registry, and so it will be deallocated last.

nfrisby: Also rename to `brackWithPrivateRegistry` and remove the `registry` argument from the `body`: none of our uses cases need it yet, so let's simplify at least for now.

The fact that none of our use cases need it yet suggested that perhaps `newFollower` should simply be replaced with `withFollower`, but it turns out that the QSM tests need the `newFollower` variant, since QSM inherently uses a flat sequence of commands: we'd have no hook into QSM in which to call `withFollower` with a scope other than "the whole test case".

ouroboros-consensus/src/Ouroboros/Consensus/Util/STM.hs
```diff
@@ -132,3 +102,83 @@ simStateT stVar (Sim k) = Sim $ \(StateT f) -> do
...
+  , wInitial     :: Maybe fp
+    -- | An action executed each time the fingerprint changes.
+  , wNotify      :: a -> m ()
+    -- | The variable to monitor.
+  , wReader      :: STM m a
+  }
+
+-- | An existentially quantified 'Watcher'.
+--
+-- It's sometimes convenient to forget the unnecessary type variables.
```
nfrisby, line 132: Eliminate `SomeWatcher`; just include the types in the use-case.

ouroboros-consensus/src/Ouroboros/Consensus/Util/STM.hs
```diff
@@ -132,3 +102,83 @@ simStateT stVar (Sim k) = Sim $ \(StateT f) -> do
...
+data SomeWatcher m where
+  SomeWatcher :: Eq fp => !(Watcher m a fp) -> SomeWatcher m
+
+-- | Execute a 'Watcher'
```
nfrisby, line 136: Add NOT EXPORTED to comment.

ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/API.hs
```diff
@@ -67,9 +67,14 @@ onKnownSlotChange :: forall m. (IOLike m, HasCallStack)
                   -> String            -- ^ Label for the thread
                   -> (SlotNo -> m ())  -- ^ Action to execute
                   -> m (m ())
-onKnownSlotChange registry btime label =
```
nfrisby, line 70: Switch this to be `Watcher`; the usesites will choose how to manage the thread.

ouroboros-consensus/src/Ouroboros/Consensus/Util/STM.hs
```diff
@@ -132,3 +102,83 @@ simStateT stVar (Sim k) = Sim $ \(StateT f) -> do
...
+spawnWatcher registry label watcher =
+    forkLinkedThread registry label $ runWatcher watcher
+
+-- | Spawn a new thread that runs a 'Watcher'
+--
+-- The thread is bracketed via 'withAsync' and 'link'ed.
+withWatcher :: forall m a fp r. (IOLike m, Eq fp, HasCallStack)
+            => String    -- ^ Label for the thread
+            -> Watcher m a fp
+            -> (Async m Void -> m r)
```
nfrisby, line 179: Drop `Async` argument from callback.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
nfrisby: Edsko and I discussed going farther and not passing the `registry` to the body. However, I just realized that the `a` type variable could be abused to pass the registry to the body.

So I'll still make no assumptions about orderings: we'll allocate the "bracketed resources" (via `newA`) in the registry.

nfrisby: I also just realized that the `newA` argument might itself allocate something in the registry. This would be weird, but allowed by is allowed by the types so far. I'm about to push commits that use `m (ResourceRegistry m -> a)` instead, which prevents that possibility at the cost of a slightly more refactoring.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -585,7 +586,49 @@ releaseResources rr keys releaser = do
 --
 -- See documentation of 'ResourceRegistry' for a detailed discussion.
 withRegistry :: (IOLike m, HasCallStack) => (ResourceRegistry m -> m a) -> m a
-withRegistry = bracket unsafeNewRegistry closeRegistry
+withRegistry body =
```
nfrisby, line 589: Too pedantic?

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -349,8 +349,10 @@ forkBlockForging
     -> BlockForging m blk
     -> m ()
 forkBlockForging maxTxCapacityOverride IS{..} blockForging =
-    void $ onKnownSlotChange registry btime threadLabel $
-        withEarlyExit_ . go
+    void
```
nfrisby, line 352: Contrary to existing module style? Or even global exception to style guide?

ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
```diff
@@ -293,24 +294,23 @@ testOverrideDelay :: forall m. (IOLike m, MonadTime m, MonadDelay (OverrideDelay
                   -> Int  -- ^ Number of slots to collect
                   -> OverrideDelay m [SlotNo]
 testOverrideDelay systemStart slotLength maxClockRewind numSlots = do
-    result <- withRegistry $ \registry -> do
+    withRegistry $ \registry -> do
       time <- simpleBlockchainTime
```
nfrisby, line 298: Check whether `simpleBlockchainTime` could be `withBlockchainTime` etc.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby, line 352: Indent `void`

nfrisby: Indent `void`; $ is binop

ouroboros-consensus/src/Ouroboros/Consensus/Util/STM.hs
```diff
@@ -3,14 +3,18 @@
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE DerivingStrategies         #-}
 {-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE GADTs                      #-}
```
nfrisby, line 6: Remove this

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -585,7 +586,49 @@ releaseResources rr keys releaser = do
...
+-- o The bracketed resources are older than everything in the registry.
+--
+-- In particular, this combinator closes the registry before releasing the
+-- bracketed resources, since any objects still in the registry will be younger
+-- than the bracketed resources and so may rely on them in their releasers.
+--
+-- See documentation of 'ResourceRegistry' for a more general discussion.
+bracketWithRegistry ::
+     (IOLike m, HasCallStack)
+  => m (ResourceRegistry m -> a)
```
nfrisby, line 614: Add TODO about Issue #2897; we can widen this argument once that's fixed.

edsko (approved):
Subtle stuff, as always with the registry. Thanks for your observation @kderme , and well-spotted on the more general problem @nfrisby . 

Peer-reviewed this over meet; good to go after some minor cleanup, the more general problem will be addressed separately.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -585,7 +586,49 @@ releaseResources rr keys releaser = do
...
+     -- ^ Allocate the bracketed resources. They depend on the registry but are
+     -- not themselves registered.
+  -> (a -> m ())
+     -- ^ Release the bracketed resources.
+  -> (ResourceRegistry m -> a -> m r)
```
nfrisby, line 619: Edsko clarified: we're removing this argument to the body because this combinator captures the intent of providing a bracketed thing that happens to need a registry. That registry is private. NB if the body also needs a registry, it can allocate its own. NB number 2: a prompt invocation of close for `a` could also close the registry.

Ideally we'd write `withFollower`, and `withFollower` would be defined via `withRegistry`, which would guarantee that it's private. This combinator is intended as a generalization of `withFollower` that can be used with any allocation function that expected a privately owned registry. (But we don't think an `ST` style existential type is worthwhile.)

Something like "This combinator provides a private repository. If the allocator does not leak the registry, then it will remain private."

ouroboros-consensus-test/test-consensus/Test/Consensus/BlockchainTime/Simple.hs
nfrisby: I did change this to `bracketWithPrivateRegistry`. It's a bit odd when there's no close function for the `btime`, but it's still better capturing the intent.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
nfrisby: > The point is that the lifetime of this new resource is equal to the lifetime of the registry.

For future reference, there are some minor caveats to that; see the comment on `bracketWithPrivateRegistry`.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
nfrisby: We backed out those new commits. We're allowing `newA` to allocate things in the registry. (I don't think it'd even _need_ to do that, since it could just close them in the `closeA`. But that might require extensions to the `a` interface.)

nfrisby: And because `newA` may allocate resources that'll be younger than the `a` bracketed here, we must track `a` in the registry; we can't close it separate from the registry because the registry has some older than `a` and some younger than `a` resources.

PR #2896, opened 2021-01-22
LocalStQuery and ChainSync clients: aditional monadic hooks

### LINKED ISSUE #2895: [FR] Allow Local State Client to idle in the acquired state
If this recvMsgAcquired
https://github.com/input-output-hk/ouroboros-network/blob/6cb9052bde39472a0555d19ade8a42da63d3e904/ouroboros-network/src/Ouroboros/Network/Protocol/LocalStateQuery/Client.hs#L71
had a wrapping `m` in the type, clients could:

1. Easily idle in the acquired state (and only-reacquire on tip-changes)
2. Log _after_ we're acquired, but _before_ we send a query.

## Context

Not absolutely certain the wallet would actually use (1), but (2) seems nice at least, and sounded like the change was simple.

## Workaround 

I think we can workaround by sending any query, suitably the GetEra one, and idling in the resulting `m` in
https://github.com/input-output-hk/ouroboros-network/blob/6cb9052bde39472a0555d19ade8a42da63d3e904/ouroboros-network/src/Ouroboros/Network/Protocol/LocalStateQuery/Client.hs#L100-L102

coot:
Fixes #2895 and incorporates #2011.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -674,7 +674,7 @@ chainSyncClient mkPipelineDecision0 tracer cfg
               (requestNext kis mkPipelineDecision' (Succ n) theirTip candTipBlockNo)
           (Succ n', (CollectOrPipeline, mkPipelineDecision')) ->
             CollectResponse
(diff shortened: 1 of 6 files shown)


karknu (approved):
LGTM.

coot:
bors merge

dcoutts (commented):
:+1:

PR #2899, opened 2021-01-27
Explicitly track relative age in the ResourceRegistry

### LINKED ISSUE #2897: Nested allocation leads to out-of-order ResourceRegistry keys
If the argument to `alloc` itself also allocates something in the registry, then the keys will no longer accurately reflect the relative age of the objects. This means the registry will deallocate them in the wrong order.

Optional task component: auditing the code base to see if this use pattern currently occurs.

We set medium priority because we haven't yet identified concrete misbehavior from this: we just identified a shortcoming of the internal library we're relying on to ensure that our cleanup handlers get to run correctly, which is a subtle thing. (EG releasing a lock when we're shutting down the DB, sending a goodbye message on a socket, etc.)

nfrisby:
Fixes #2897.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
@@ -56,6 +56,8 @@ import           Control.Monad
 import           Control.Monad.Reader
 import           Control.Monad.State.Strict
 import           Data.Bifunctor
+import           Data.Bimap (Bimap)
+import qualified Data.Bimap as Bimap
 import           Data.Either (partitionEithers)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
@@ -64,6 +66,7 @@ import           Data.Proxy
 import           Data.Set (Set)
 import qualified Data.Set as Set
 import           Data.Tuple (swap)
+import           Data.Word (Word64)
 import           GHC.Generics (Generic)
 import           NoThunks.Class (InspectHeapNamed (..), OnlyCheckWhnfNamed (..),
                      allNoThunks)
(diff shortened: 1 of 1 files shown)


nfrisby:
@edsko will review this live during Thursday's Meet. @EncodePanda, I'd like you to give it a pass too. It's self-contained and the module is very well documented, so you can look as deep as you like in your review pass.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -307,13 +325,28 @@ data RegistryState m = RegistryState {
       -- | Next available resource key
     , registryNextKey   :: !ResourceId
 
+      -- | The age of each currently allocated resource
+      --
+      -- We use a 'Bimap' so we can both efficiently forget a resource when it
```
nfrisby, line 330: Simplify this; just mention the ordering, let the forgetting part be implicit.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -296,6 +300,20 @@ deriving instance IOLike m => NoThunks (ResourceRegistry m)
   Internal: registry state
 -------------------------------------------------------------------------------}
 
+-- | The age of a resource is represented by how many resources were allocated
+-- before it
+--
+-- This data is not exposed by the 'ResourceRegistry' interface.
+newtype Age = YoungerThan Word64
```
nfrisby, line 307: Edsko favors the `maxBound` idea, let's try that.

ouroboros-consensus/src/Ouroboros/Consensus/Util/ResourceRegistry.hs
```diff
@@ -411,14 +442,26 @@ insertResource :: ResourceId
...
+      , registryAges      = Bimap.insert
+                              key
+                              (registryNextAge st)
+                              (registryAges st)
+      , registryNextAge   = nextYoungerAge (registryNextAge st)
       }
 
 -- | Remove resource from the registry (if it exists)
 removeResource :: ResourceId -> State (RegistryState m) (Maybe (Resource m))
-removeResource key = state $ \st ->
```
nfrisby, line 418: Try to use `state`, but with less point-free.

edsko (approved):
Nice, good to decouple these two concepts.

PR #2900, opened 2021-01-27
Clean-up the ChainSyncServers creation after recent modification to resource management

EncodePanda:
Please read through commit messages to understand the intent of this PR.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
@@ -26,7 +26,6 @@ import qualified Ouroboros.Consensus.Storage.ChainDB.API as ChainDB
 import           Ouroboros.Consensus.Storage.Serialisation
 
 import           Ouroboros.Consensus.Block
-import           Ouroboros.Consensus.Node.NetworkProtocolVersion
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.ResourceRegistry (ResourceRegistry)
 
@@ -56,9 +55,8 @@ chainSyncHeadersServer
     => Tracer m (TraceChainSyncServerEvent blk)
     -> ChainDB m blk
     -> Follower m blk (WithPoint blk (SerialisedHeader blk))
-    -> NodeToNodeVersion
     -> ChainSyncServer (SerialisedHeader blk) (Point blk) (Tip blk) m ()
-chainSyncHeadersServer tracer chainDB flr _version =
+chainSyncHeadersServer tracer chainDB flr =
     chainSyncServerForFollower tracer chainDB flr
 
 -- | Chain Sync Server for blocks for a given a 'ChainDB'.
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
@@ -59,9 +59,7 @@ chainSyncHeadersServer
     -> NodeToNodeVersion
     -> ChainSyncServer (SerialisedHeader blk) (Point blk) (Tip blk) m ()
 chainSyncHeadersServer tracer chainDB flr _version =
-    ChainSyncServer $
-      let ChainSyncServer server = chainSyncServerForFollower tracer chainDB flr in
-      server
+    chainSyncServerForFollower tracer chainDB flr
 
 -- | Chain Sync Server for blocks for a given a 'ChainDB'.
 --
@@ -75,9 +73,7 @@ chainSyncBlocksServer
     -> Follower m blk (WithPoint blk (Serialised blk))
     -> ChainSyncServer (Serialised blk) (Point blk) (Tip blk) m ()
 chainSyncBlocksServer tracer chainDB flr =
-    ChainSyncServer $
-      let ChainSyncServer server = chainSyncServerForFollower tracer chainDB flr in
-      server
+    chainSyncServerForFollower tracer chainDB flr
 
 -- | A chain sync server.
 --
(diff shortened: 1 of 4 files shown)


EncodePanda:
@karknu I've added you as a reviewer because you were doing the recent changes around the chainSyncServers

karknu (approved):
LGTM.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -44,42 +41,6 @@ chainSyncBlockServerFollower
...
-    -> NodeToNodeVersion
-    -> ChainSyncServer (SerialisedHeader blk) (Point blk) (Tip blk) m ()
-chainSyncHeadersServer tracer chainDB flr _version =
-    ChainSyncServer $
-      let ChainSyncServer server = chainSyncServerForFollower tracer chainDB flr in
-      server
-
--- | Chain Sync Server for blocks for a given a 'ChainDB'.
---
--- The local node-to-client protocol uses the chain sync mini-protocol with
```
nfrisby, line 69: Would you move this comment sentence to `chainSyncServerForFollower`? It'd be good to put the two driving examples there motivating why `b` does not equal `blk`.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -44,42 +41,6 @@ chainSyncBlockServerFollower
     -> m (Follower m blk (WithPoint blk (Serialised blk)))
 chainSyncBlockServerFollower chainDB registry = ChainDB.newFollower chainDB registry getSerialisedBlockWithPoint
 
--- | Chain Sync Server for block headers for a given a 'ChainDB'.
---
--- The node-to-node protocol uses the chain sync mini-protocol with chain
```
nfrisby, line 49: Would you move this comment sentence to `chainSyncServerForFollower`? It'd be good to put the two driving examples there motivating why `b` does not equal `blk`.

nfrisby:
@EncodePanda would you fixup the title? I just moments ago added you as a reviewer to PR #2903, which elaborates on this sort of meta-level concern; you can find some motivation/details there.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -114,7 +114,6 @@ data Handlers m peer blk = Handlers {
 
     , hChainSyncServer
         :: ChainDB.Follower m blk (ChainDB.WithPoint blk (SerialisedHeader blk))
-        -> NodeToNodeVersion
```
nfrisby, line 117: Marcin's PR 2090 introduced this argument to all Consensus mini protocol threads in commit 68d35f453c1b1bc1a6eb5ebc38f1bf8392f2be48.

I was looking at what the state was then, but then I realized I should just look at today's `master`. Turns out, none of our mini protocol threads (ie "handlers") use the version information.

That's based on inspection of this code and the definitions it calls to
https://github.com/input-output-hk/ouroboros-network/blob/ed24495e525931306a30de5827ea91caafbec528/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs#L171-L207

It looks like the NTC analogs don't even have the types for depending on the
version
https://github.com/input-output-hk/ouroboros-network/blob/ed24495e525931306a30de5827ea91caafbec528/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs#L88-L98

nfrisby: So my thought process is: let's remove c6764f7b6c11d020c7bc160ecf4bb31ac6dec7e4 from this PR, open a separate Issue, and ping Marcin on it. Without hearing his thoughts yet, my intuition is [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) says we should remove the version from all the NTN handlers.

nfrisby (changes requested):
Good catches! I'm happy you're inspecting so thoroughly.

I'm Requesting Changes b/c of the PR title and the separation of concerns (regarding removing the version argument).

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
EncodePanda: I have similar intuition. But if we remove `c6764f7b6c11d020c7bc160ecf4bb31ac6dec7e4` we will also have to remove `9388a549cb0ac01caf4fe643fd022e949c546556` - if we keep the version in `chainSyncHeadersServer` then it simply makes no sense of removing it in the next PR, because `chainSyncHeadersServer` gives nice facade for its usage in `Handlers`.

Maybe it's worth having the discussion about the version numbers being removed from `Handelres` (where they are not needed) now - in this PR?

EncodePanda: follow up after meeting with @nfrisby - we will keep this commit but we will not remove version argument from `Handlers`

nfrisby:
@EncodePanda FYI, the CI caught an unused import.

nfrisby:
Did you want to remove the arg from `chainSyncClient` too?

Oof, I forgot that the `TxSub` handlers are defined in the Network Layer. I saw your Slack conversation with Marcin -- it's not clear to me if that's the kind of diff that will incur the merge conflicts you two had discussed.

A few options come to mind:

1. we defer this PR until his `p2p` merge happens
1. we narrow this PR to just the consensus files
1. we get his Approval on your diff (which might also end up being one of the first two options here)

EncodePanda:
@nfrisby I've removed last commit, the one that removed `chainSyncHeadersServer` and `chainSyncBlockServer`. I've created a ticket #2942 (btw a low hanging fruit for newcommer) so that we won't forget to do it once `p2p-master` is in.

I've left changes in `ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs`, those should not block @coot when he eventually merges `p2p-master`

Please do final review of this PR so we can get it in :)

EncodePanda:
@nfrisby let's wrap this up, this way or the other :)

nfrisby (approved):
Reviewed over Meet 👍

PR #2901, opened 2021-01-27
NodeToClient ChainSync with block metadata

kderme:
This extends the NodeToClient protocol with extra metadata (an Interpreter) for slot to time conversions. This is meant to simplify db-sync (also maybe the wallet), since it will no longer need to retrieve time metadata from the LocalStateQuery.

The server keeps track of the latest Interpreter that was sent to the client and only sends a new one, when the current slot is outside the horizon. So it doesn't need to send an Interpreter with every block, but only when it is necessary.

There are some things I could improve:
- I didn't extend the `ShelleyNodeToClientVersion`, only the `NodeToClientVersion`. Probably I should do both?
- I more and more think that the HistoryFollower that I introduce, should actually be part of the Follower. There could be a dedicated BlockComponent to retrieve the metadata. This could reduce race conditions, between reading a block and reading from LedgerDB.
- The client code could become easier to use.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Follower.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Follower.hs
@@ -1,0 +1,148 @@
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Ouroboros.Consensus.HardFork.History.Follower (
+    -- * API
+    HistoryFollower (..)
+  , WithInterpreter (..)
+  , WithInterDefault
+  , mkWithInterpreter
+  , newHistoryFollower
+    -- * Client
+  , ClientHistoryFollower (..)
+  , WithInterClient
+  , newClientFollower
+  ) where
+
+import           Control.Monad.Class.MonadSTM.Strict hiding (newTVarIO)
+import           Data.Either (isRight)
+import           Data.SOP.Strict
+
+import           Ouroboros.Consensus.HardFork.Abstract
+import           Ouroboros.Consensus.HardFork.Combinator.Basics
+import           Ouroboros.Consensus.HardFork.Combinator.Ledger ()
+import qualified Ouroboros.Consensus.HardFork.History as History
+import           Ouroboros.Consensus.HardFork.History.Qry
+import           Ouroboros.Consensus.Ledger.Basics
+import           Ouroboros.Consensus.Ledger.Extended
+import           Ouroboros.Consensus.Storage.ChainDB
+import           Ouroboros.Consensus.Util
+import           Ouroboros.Consensus.Util.IOLike
+
+import           Ouroboros.Network.Block
+import           Ouroboros.Network.Point
+
+
+-- | The 'HistoryFollower' can be used to extend the 'Follower' with metadata
+-- which come from the ledger state. Currently this metadata is a History
+-- 'Interpreter' but it's technically possible to generalise this.
+--
+-- In the same way that a 'Follower' keeps track of the tip of a user, the
+-- 'HistoryFollower' keeps track and mirrors the Interpreter that a user has.
+-- Everytime a new 'Interpreter' is created and returned to the user, it is also
+-- cached. The api returns 'Nothing' when the cached 'Interpreter' is capable
+-- of serving the specified point, so the user doesn't need a new one.
+--
+newtype HistoryFollower m blk = HistoryFollower {
+      getInterpreter :: Point blk -> STM m (Maybe (Interpreter (HardForkIndices blk)))
+    }
+
+data WithInterpreter blk b = WithInterpreter {
+      payload     :: b
+    , interpreter :: Maybe (Interpreter (HardForkIndices blk))
+    } deriving Show
+
+type WithInterDefault blk = WithInterpreter blk (Serialised blk)
+
+mkWithInterpreter ::
+     b
+  -> Maybe (Interpreter (HardForkIndices blk))
+  -> WithInterpreter blk b
+mkWithInterpreter b mInter = WithInterpreter {
+      payload     = b
+    , interpreter = mInter
+    }
+
+instance (ShowProxy blk, ShowProxy b) => ShowProxy (WithInterpreter blk b) where
+    showProxy _ = mconcat [
+        "WithInterpreter "
+      , showProxy (Proxy :: Proxy blk)
+      , " "
+      , showProxy (Proxy :: Proxy b)
+      ]
+
+{-------------------------------------------------------------------------------
+  Implementation
+-------------------------------------------------------------------------------}
+
+newHistoryFollower ::
+     forall blk m.
+     ( IOLike m
+     , HasHardForkHistory blk
+     , IsLedger (LedgerState blk)
+     )
+  => LedgerConfig blk
+  -> ChainDB m blk
+  -> m (HistoryFollower m blk)
+newHistoryFollower cfg chainDB = do
+    varState <- newTVarIO Nothing
+    return $ HistoryFollower {
+        getInterpreter =
+          getInterpreterImpl cfg (ledgerState <$> getCurrentLedger chainDB) varState
+      }
+
+getInterpreterImpl ::
+     (IOLike m, HasHardForkHistory blk)
+  => LedgerConfig blk
+  -> STM m (LedgerState blk)
+  -> StrictTVar m (Maybe (Interpreter (HardForkIndices blk)))
+  -> Point blk
+  -> STM m (Maybe (Interpreter (HardForkIndices blk)))
+getInterpreterImpl cfg getState varInterpreter point = do
+    minter <- readTVar varInterpreter
+    case (minter, pointSlot point) of
+      (Just inter, At slot) | isSlotInHorizon slot inter -> return Nothing
+      _ -> do
+        st <- getState
+        let inter = History.mkInterpreter $ hardForkSummary cfg st
+        writeTVar varInterpreter $ Just inter
+        return $ Just inter
+
+-- | Run a sanity query to check if the slot is still in the Horizon.
+isSlotInHorizon :: SlotNo -> Interpreter xs -> Bool
+isSlotInHorizon slotNo inter = isRight $ interpretQuery inter query
+  where
+    query = (,) <$> slotToWallclock slotNo <*> slotToEpoch' slotNo
+
+{-------------------------------------------------------------------------------
+  Client History Follower
+-------------------------------------------------------------------------------}
+
+-- | This can be used by a client. Given a 'WithInterClient',
+-- 'getInterpreter' is the best effort to retrieve an 'Interpreter'.
+newtype ClientHistoryFollower m blk = ClientHistoryFollower {
+      getInterpreterClient :: WithInterClient blk
+                           -> STM m (Maybe (Interpreter (HardForkIndices blk)))
+    }
+
+newClientFollower :: MonadSTM m => m (ClientHistoryFollower m blk)
+newClientFollower = do
+    mInter <- newTVarIO Nothing
+    return $ ClientHistoryFollower $ getInterClientImpl mInter
+
+getInterClientImpl ::
+     MonadSTM m
+  => StrictTVar m (Maybe (Interpreter (HardForkIndices blk)))
+  -> WithInterClient blk
+  -> STM m (Maybe (Interpreter (HardForkIndices blk)))
+getInterClientImpl interVar withInter =
+    case interpreter withInter of
+      Just inter -> do
+        writeTVar interVar (Just inter)
+        return $ Just inter
+      Nothing -> do
+        mInter <- readTVar interVar
+        return mInter
+
+type WithInterClient blk = WithInterpreter blk blk
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Follower.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Follower.hs
@@ -16,13 +16,13 @@ module Ouroboros.Consensus.HardFork.History.Follower (
   ) where
 
 import           Control.Monad.Class.MonadSTM.Strict hiding (newTVarIO)
-import           Data.Either (isRight)
 import           Data.SOP.Strict
 
 import           Ouroboros.Consensus.HardFork.Abstract
 import           Ouroboros.Consensus.HardFork.Combinator.Basics
 import           Ouroboros.Consensus.HardFork.Combinator.Ledger ()
 import qualified Ouroboros.Consensus.HardFork.History as History
+import           Ouroboros.Consensus.HardFork.History.Caching
 import           Ouroboros.Consensus.HardFork.History.Qry
 import           Ouroboros.Consensus.Ledger.Basics
 import           Ouroboros.Consensus.Ledger.Extended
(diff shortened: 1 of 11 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -205,13 +213,31 @@ defaultCodecs ccfg version networkVersion = Codecs {
     dec :: SerialiseNodeToClient blk a => forall s. Decoder s a
     dec = decodeNodeToClient ccfg version
 
+    encInter :: WithInterDefault blk -> Encoding
```
kderme, line 216: Does this need a proper `SerialiseNodeToClient` instance and introduce a new `BlockNodeToClientVersion`?

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -89,61 +92,85 @@ chainSyncBlocksServer tracer chainDB flr =
...
-     forall m blk b.
-     ( IOLike m
-     , HasHeader (Header blk)
-     )
-  => Tracer m (TraceChainSyncServerEvent blk)
-  -> ChainDB m blk
-  -> Follower  m blk (WithPoint blk b)
-  -> ChainSyncServer b (Point blk) (Tip blk) m ()
-chainSyncServerForFollower tracer chainDB flr =
+chainSyncServerForReader
```
nfrisby, line 95: Please don't change the signature style. Consensus likes `::` at the end of the line.

nfrisby, line 95: Also, we recently renamed `Reader` to `Follower`. It looks like maybe your rebase lost that in this diff.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Follower.hs
```diff
@@ -0,0 +1,144 @@
...
+      , showProxy (Proxy :: Proxy b)
+      ]
+
+{-------------------------------------------------------------------------------
+  Implementation
+-------------------------------------------------------------------------------}
+
+newHistoryFollower ::
+    forall blk m.
+    ( IOLike m
```
nfrisby, line 78: A style deviation: would you indent this context so that it's farther to the right than the `=>`.

Like this:

```
       forall blk m.
       ( IOLike m
       , HasHardForkHistory blk
       , IsLedger (LedgerState blk)
       )
    => LedgerConfig blk
    -> ChainDB m blk
    -> m (HistoryFollower m blk)
```

If you'd like more details, please see "signatures" at https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Follower.hs
```diff
@@ -0,0 +1,144 @@
...
+      getInterpreterClient :: WithInterClient blk
+                           -> STM m (Maybe (Interpreter (HardForkIndices blk)))
+    }
+
+newClientFollower :: MonadSTM m => m (ClientHistoryFollower m blk)
+newClientFollower = do
+    mInter <- newTVarIO Nothing
+    return $ ClientHistoryFollower $ getInterClientImpl mInter
+
+getInterClientImpl :: MonadSTM m
```
nfrisby, line 131: Please indent the type on a newline after `::`, like you did in `newHistoryFollower`.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
nfrisby: Ugh. I see now that this module is already inconsistent with its `::` placement. Sorry for the confusion.

We do overall prefer `::` at the end of the line.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
```diff
@@ -98,4 +101,5 @@ class ( LedgerSupportsProtocol           blk
       , ShowProxy                (Header blk)
       , ShowProxy                 (Query blk)
       , ShowProxy           (TxId (GenTx blk))
+      , SListI (HardForkIndices blk)
```
nfrisby, line 104: Please maintain the big line of aligned `blk` vars.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -362,13 +407,15 @@ mkApps kernel Tracers {..} Codecs {..} Handlers {..} =
       bracketWithPrivateRegistry
         (chainSyncBlockServerFollower (getChainDB kernel))
         ChainDB.followerClose
-        $ \flr ->
+        $ \flr -> do
+          let cfg = topLevelConfigLedger $ getTopLevelConfig kernel
+          hflr <- newHistoryFollower cfg (getChainDB kernel)
```
nfrisby, line 412: Note to self: this is just a closure-bound `TVar`, so no need for `registry`, `bracket` etc.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Follower.hs
```diff
@@ -0,0 +1,144 @@
...
+newtype HistoryFollower m blk = HistoryFollower {
+      getInterpreter :: Point blk -> STM m (Maybe (Interpreter (HardForkIndices blk)))
+    }
+
+data WithInterpreter blk b = WithInterpreter {
+      payload     :: b
+    , interpreter :: Maybe (Interpreter (HardForkIndices blk))
+    } deriving Show
+
+type WithInterDefault blk = WithInterpreter blk (Serialised blk)
```
nfrisby, line 56: Is this "default" as in "not a client"? If so, I'd suggested `WithInterSerialised` or `WithInterNTN` (for node-to-node).

Also, more weakly, I'd suggesting spelling out `Inter` and `Interpreter`; `Inter` seems a bit ambiguous at first blush.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -89,61 +92,85 @@ chainSyncBlocksServer tracer chainDB flr =
...
       -- TODO guard number of points
       changed <- ChainDB.followerForward flr points
       tip     <- atomically $ ChainDB.getCurrentTip chainDB
       return $ case changed of
-        Just pt -> SendMsgIntersectFound    pt tip idle'
-        Nothing -> SendMsgIntersectNotFound    tip idle'
+        Just pt -> SendMsgIntersectFound pt tip idle'
+        Nothing -> SendMsgIntersectNotFound tip idle'
+
+chainSyncServerForReaderNoMeta
```
nfrisby, line 163: `Reader` to `Follower` here too.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -89,61 +92,85 @@ chainSyncBlocksServer tracer chainDB flr =
...
 
     handleFindIntersect :: [Point blk]
-                        -> m (ServerStIntersect b (Point blk) (Tip blk) m ())
+                        -> m (ServerStIntersect bmeta (Point blk) (Tip blk) m ())
     handleFindIntersect points = do
       -- TODO guard number of points
       changed <- ChainDB.followerForward flr points
       tip     <- atomically $ ChainDB.getCurrentTip chainDB
       return $ case changed of
-        Just pt -> SendMsgIntersectFound    pt tip idle'
```
nfrisby, line 145: Would you minimize the diff here?

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -4,7 +4,6 @@
 {-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE RecordWildCards       #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
-{-# LANGUAGE TypeApplications      #-}
```
nfrisby, line 7: Nice catch! I'm always impressed when people notice these things. Very thorough.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -131,7 +138,7 @@ mkHandlers NodeKernelArgs {cfg, tracers} NodeKernel {getChainDB, getMempool} =
 
 -- | Node-to-client protocol codecs needed to run 'Handlers'.
 data Codecs' blk serialisedBlk e m bCS bTX bSQ = Codecs {
-      cChainSyncCodec    :: Codec (ChainSync serialisedBlk (Point blk) (Tip blk))  e m bCS
+      cChainSyncCodec    :: Codec (ChainSync (WithInterpreter blk serialisedBlk) (Point blk) (Tip blk))  e m bCS
```
nfrisby, line 141: `WithInterpreterClient`?

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -61,8 +62,8 @@ chainSyncHeadersServer
     -> ChainSyncServer (SerialisedHeader blk) (Point blk) (Tip blk) m ()
 chainSyncHeadersServer tracer chainDB flr _version =
     ChainSyncServer $
-      let ChainSyncServer server = chainSyncServerForFollower tracer chainDB flr in
```
nfrisby, line 64: FYI we have another PR removing this function. So you'll probably have a rebase conflict soon.

nfrisby (changes requested):
I started with the rote stuff, since that's what jumped out at me: style, alignment breaks, mismerged rebased conflicts, anticipated rebase conflicts, etc, etc.

Sorry for the style being a bit inconsistent in `Client.hs` already; that was a surprise to me that I realized during the scan.

I'll do another pass soon where I review the actual substance of the diff. Thanks.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Follower.hs
kderme: > Is this "default" as in "not a client"? If so, I'd suggested WithInterSerialised or WithInterNTN (for node-to-node -- and WithInterClient would probably then best be WithInterNTC).

I used `WithInterDefault` following the existing naming `DefaultCodecs`, which is used by the server and contains the `Serialized` block. However it could make sence to change `Default` to `Server` in all these places. Note that the Interpreter is only involved with the NodeToClient protocol.

> Also, more weakly, I'd suggesting spelling out Inter as Interpreter; Inter seems a bit ambiguous at first blush.

In general I used the shorter `Inter` when the name started to become too long. But I'm fine with your suggestion.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
kderme: hls found it, I shouldn't take the credits :)

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
kderme: `Codecs'` is the general type (used by both client and server), so I think using the general `WithInterpreter` makes sense here. 

On another note `serialisedBlk` parameter is a bit misleading, since this can also be the blk.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
kderme: Fixed it. I tried to follow the guide for old and new signatures in this module.

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Server.hs
```diff
@@ -89,29 +89,28 @@ chainSyncBlocksServer tracer chainDB flr =
...
-     )
+     forall m blk b meta bmeta. (IOLike m, HasHeader (Header blk))
   => Tracer m (TraceChainSyncServerEvent blk)
   -> ChainDB m blk
   -> Follower  m blk (WithPoint blk b)
-  -> ChainSyncServer b (Point blk) (Tip blk) m ()
-chainSyncServerForFollower tracer chainDB flr =
+  -> (Point blk -> STM m meta)
+  -> (b -> meta -> bmeta)
+  -> ChainSyncServer bmeta (Point blk) (Tip blk) m ()
```
kderme, line 98: `bmeta` represents a pair of the block and the Interpreter

nfrisby:
I discussed this with Edsko. His comments were very much aligned with what you said here:

> I more and more think that the HistoryFollower that I introduce, should actually be part of the Follower. There could be a dedicated BlockComponent to retrieve the metadata. This could reduce race conditions, between reading a block and reading from LedgerDB.

Specifically, we also thought to add a `GetLedgerState :: BlockComponent blk (LedgerState blk)` here

https://github.com/input-output-hk/ouroboros-network/blob/bdaa81365e4369c5d284879af18a3865f4f8a261/ouroboros-consensus/src/Ouroboros/Consensus/Storage/Common.hs#L148-L167

https://github.com/input-output-hk/ouroboros-network/blob/bdaa81365e4369c5d284879af18a3865f4f8a261/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs#L191-L194

and rely on that in addition to `getSerialisedBlockWithPoint`.

_However_, as I'm typing this, I realized we won't necessarily have the `LedgerState` at hand for older blocks. Since `BlockComponent` is used for all sorts of `ChainDB` queries (not just `newFollower`), we should not add one that is often-enough unsatisfiable. My next thought was to extend the `newFollower` function

https://github.com/input-output-hk/ouroboros-network/blob/bdaa81365e4369c5d284879af18a3865f4f8a261/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs#L298-L301

to something like

```
    , newFollower ::
           forall b. ResourceRegistry m
        -> BlockComponent blk b
        -> m (Follower m blk (b, XXX))
```

where at first I thought `XXX` should be the most recent `LedgerState`. But I don't like that either, it seems too surprising. The most recent `LedgerState` would suffice, since that would let us create an `Interpreter` that should be able to handle any block we've ever seen, including whichever the follower last notified us of. So why not just set `XXX ~ Interpreter`? That might be right, but it also suggests that `Interpreter` should not be as coupled to the HFC as it currently is.

Your current approach -- of querying the meta information after each follower update --- seems fine, ultimately. Especially if you were to re-use Edko's `RunWithCachedSummary` (see the rest of this comment).

-----

He also noticed that your caching mechanism is very similar to the existing `HardFork.History.Caching` module, so hopefully you can reuse that directly (I haven't attempted to connect the dots yet, myself.)

https://github.com/input-output-hk/ouroboros-network/blob/963bdfb4d7383bcbbee5477b18dbc8b8d92ad8a5/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Caching.hs#L7

-----

I don't think the `HistoryFollower`/`RunWithCachedSummary` needs to be allocated in `NodeKernel`. It doesn't have any actual resources eg filehandles --- it's just a tvar. So you can do that allocation inside of `chainSyncBlocksServer` itself.

nfrisby:
OK. That was a big comment. Please let me know if that gives you something to work on or if you'd like to have a call to discuss. Thanks.

kderme:
> He also noticed that your caching mechanism is very similar to the existing HardFork.History.Caching module, so hopefully you can reuse that dMaybe (Interpreter xs))irectly (I haven't attempted to connect the dots yet, myself.)

It would be possible, but the existing caching would have to change. For example the type would have to become:
`cachedRunQuery :: forall a. Qry a -> STM m (Either PastHorizonException (a, Maybe (Interpreter xs)))`

Or using the type I introduce
`cachedRunQuery :: forall a. Qry a -> STM m (Either PastHorizonException (WithInterpreter blk b)`

kderme:
I started using Caching.hs (last commit), but I'm not sure how much it helps.

kderme:
I guess this is not happening. Closing.

PR #2902, opened 2021-01-27
Split counters from traces

newhoggy:


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -489,13 +495,15 @@ mkNodeKernelArgs
   keepAliveRng
   cfg
   initBlockForging
+  counters
```
nfrisby, line 498: Just to clarify my proposal, I'm suggesting:

```
return NodeKernelArgs
      { tracers = counters <> tracers
      , ...
```

instead of 

```
return NodeKernelArgs
      { counters
      , tracers
      , ...
```

I don't think we need to maintain the distinction inside our implementation and internal interfaces (so no new field in `NodeKernelArgs` etc).

nfrisby: (But Marcin and Serge haven't yet responded to that proposal, so I'm eg not yet expecting a corresponding update to this PR.)

newhoggy: I've created a new PR for the changes you've described here for comparison: https://github.com/input-output-hk/ouroboros-network/pull/2910

coot: > (But Marcin and Serge haven't yet responded to that proposal, so I'm eg not yet expecting a corresponding update to this PR.)

I think that's a good idea.

nfrisby:
@newhoggy Based on PR #2909, I'm wondering: can we close this PR? Thanks.

newhoggy:
Sure!

PR #2904, opened 2021-01-27
Newhoggy/git guide

newhoggy:


nfrisby:
I'm guessing this PR is just an experiment to debug the GH Action failures on my PR. Is that right, @newhoggy ?

newhoggy:
Yep.  I'm closing now because it appears to be a temporary glitch.

PR #2906, opened 2021-01-28
Install Haskell tools before checkout

newhoggy:
Install Haskell at first.  This achieves two outcomes:

* If the build were to fail in Haskell Setup due to flakiness, then fail earlier, so that the developer can wait less before discovering the error and re-triggering the build.
* The Haskell setup occurs before checkout which means it won't start cloning dependencies that were specified in `cabal.project`.  That way if the clone were to fail, then it will be in the explicit `cabal update` we have in our Github Actions template rather than in the Haskell setup, in which case we may be able to deal with that failure more easily.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,24 @@ jobs:
             subset: consensus
 
     steps:
-    - uses: actions/checkout@v2
+    - name: Select optimal cabal version
+      run: |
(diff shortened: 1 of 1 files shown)


nfrisby (approved):
LGTM

newhoggy:
Thanks!

newhoggy:
bors merge

PR #2915, opened 2021-02-01
ThreadNet test for a two-era chain: Allegra to Mary hardfork

### LINKED ISSUE #2846: Test the hard fork from Allegra to Mary
Run the ThreadNet tests for a net forking from Allegra to Mary. This can just be a copy-paste of [`Test.ThreadNet.ShelleyAllegra`](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus-cardano-test/test/Test/ThreadNet/ShelleyAllegra.hs). But ideally, the common code is generalised so it can be used by both (as there will follow many more hard forks, this would not be premature abstraction!). Alternatively, `ShelleyAllegra` is replaced with `AllegraMary`?

Ideally, we'd use the rewritten ThreadNet tests for this :slightly_smiling_face:

nfrisby:
Fixes #2846.

As for Shelley to Allegra, the Consensus Layer tests in this PR now also test Allegra to Shelley in a small ThreadNet test involving just those two eras.

Since our transaction generators for those are temporarily disabled (CAD-2119), the only transactions are the votes and `d`-reductions that the test itself injects.

The longer-term plan is for the ThreadNet rewrite to subsume this by simply testing the entirety of Cardano. The planned shrinker will be able to reduce a failing test case down to involve only a single faulty hardfork when that is indeed the bug (it will start in the era before the fork and schedule it to stop in the next).

### DIFF
--- a/ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
+++ b/ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
@@ -27,6 +27,7 @@ module Test.ThreadNet.Infra.Shelley (
   , mkLeaderCredentials
   , mkProtocolShelley
   , mkSetDecentralizationParamTxs
+  , mkAllegraSetDecentralizationParamTxs
   , mkVerKey
   , mkKeyPair
   , networkId
@@ -36,6 +37,7 @@ module Test.ThreadNet.Infra.Shelley (
 
 import           Control.Monad.Except (throwError)
 import qualified Data.ByteString as BS
+import           Data.Coerce (coerce)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Ratio (denominator, numerator)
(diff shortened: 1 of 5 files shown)


nfrisby:
Commit fcf33be just copies and renames the files. The next one makes the actual changes, so see it's diff.

Note that the commit history will be cleaned before this PR merges.

nfrisby:
@nc6 Would you skim the `mkAllegraSetDecentralizationParamTxs` function in `ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs`. Specifically, I'm wondering if there's already (parts of) a general interface in the ledger API that I could use to write this testing function in a more general way. Could it be used for both Allegra and Mary? Ideally, even Shelley itself too? Thanks.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/Infra/Shelley.hs
```diff
@@ -529,3 +536,121 @@ mkKeyHashVrf = SL.hashVerKeyVRF . deriveVerKeyVRF
...
+     ( HashAnnotated body era
+     , HashIndex body ~ EraIndependentTxBody
+     )
+  => body
+  -> Shelley.Spec.Ledger.Keys.Hash
+       (Cardano.Ledger.Era.Crypto era)
+       EraIndependentTxBody
+eraIndTxBodyHash' = coerce . hashAnnotated
+
+mkAllegraSetDecentralizationParamTxs ::
```
nfrisby, line 555: Comment here about this being a temporary bridge, soon to be replaced by the rewrite. In particular, we plan to replace this overall approach before Alonzo.

edsko (approved):
Reviewed over meet.

Issue #2919, opened 2021-02-02
Incorrect Pipelining Descisions for Chainsync Client


karknu:
Regarding `pipelineDecisionMax`, `pipelineDecisionMin`, `pipelineDecisionLowHighMark` . 

The decision for when to do `Request` and when to do `Pipelining` when there are `Zero` outstanding blocks is incorrect. The functions listed above only does a `Request` when there are `Zero` outstanding requests and the server is one block ahead of the client. The correct behavior is that a `Request` should be issued if both client and server have the same tip, that is they are both waiting for the next block to be forged.

nfrisby:
Does your PR #2917 fully resolve this?

karknu:
It does, but I intend to fix it in a separate PR.

nfrisby:
OK, got it. Thanks for clarifying.

karknu:
Fixed

PR #2920, opened 2021-02-03
Fix request/pipeling decision with 0 reqs

karknu:
The trigger for when we should use a Request rather than Pipeline
should be when the client and the server agree on the tip, that is they
are both waiting on the next block to be forged.

Previously the descision to do a Request only happened when the server
was exactly one block ahead of us and we had no outstanding requests.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Protocol/ChainSync/PipelineDecision.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Protocol/ChainSync/PipelineDecision.hs
@@ -104,10 +104,10 @@ pipelineDecisionMax :: Word32 -> Nat n -> WithOrigin BlockNo -> WithOrigin Block
                     -> PipelineDecision n
 pipelineDecisionMax omax n cliTipBlockNo srvTipBlockNo =
     case n of
-      Zero   -- We are at most one block away from the server's tip. We use
+      Zero   -- We are insync with the server's tip. We use
              -- equality so that this does not get triggered when we are ahead
              -- of the producer, and it will send us 'MsgRollBackward'.
-             | cliTipBlockNo `bnoPlus` 1 == srvTipBlockNo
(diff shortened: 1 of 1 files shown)


karknu:
Fixes #2919.

ouroboros-network/src/Ouroboros/Network/Protocol/ChainSync/PipelineDecision.hs
```diff
@@ -104,10 +104,10 @@ pipelineDecisionMax :: Word32 -> Nat n -> WithOrigin BlockNo -> WithOrigin Block
                     -> PipelineDecision n
 pipelineDecisionMax omax n cliTipBlockNo srvTipBlockNo =
     case n of
-      Zero   -- We are at most one block away from the server's tip. We use
+      Zero   -- We are at insync with the server's tip. We use
```
coot, line 107:
```suggestion
      Zero   -- We are insync with the server's tip. We use
```

ouroboros-network/src/Ouroboros/Network/Protocol/ChainSync/PipelineDecision.hs
```diff
@@ -143,10 +143,10 @@ pipelineDecisionMin :: Word32 -> Nat n -> WithOrigin BlockNo -> WithOrigin Block
                     -> PipelineDecision n
 pipelineDecisionMin omax n cliTipBlockNo srvTipBlockNo =
     case n of
-      Zero   -- We are at most one block away from the server's tip. We use
+      Zero   -- We are at insync with the server's tip. We use
```
coot, line 146:
```suggestion
      Zero   -- We are insync with the server's tip. We use
```

coot (approved):
LGTM

dcoutts (approved):

PR #2925, opened 2021-02-04
Clarify the argument records in Ouroboros.Consensus.Node

nfrisby:
This PR is a single small commit that touches only comments; see diff.


nfrisby:
I couldn't explain this well on our call, so I revisited the motivation and tried to improve the comments. HTH.

EncodePanda (approved):

nfrisby:
I rebased onto origin master. It was automatic.

Issue #2826, opened 2020-12-16
Take  snapshots of certain slots


mrBliss:
To help the ledger team debug issues, they'd like the ability to create a snapshot of the ledger state at a certain slot number.
#2479 is about extending the disk policy of the LedgerDB to allow taking snapshots of certain slots.

## Definition of Done:
* consensu layer exposes a new lower level flag allowing node operator to explicitly specify slot number for which ledger state's snapshot should be stored on disk
* modify `defaultDiskPolicy` so that it uses the above mentioned flag such that upon calling `onDiskShouldTakeSnapshot`, if the slot number meets the one provided via node operator, the snapshot is taken
* open up a branch on `cardano-node` where the new flag is integrated in the node

## Further improvements

Some other notes/remarks that might be included in this ticket on in follow-ups. After delivering this functionality we will most likely consult with ledger state team to clarify the next steps (if any). #KISS

Besides that, some more changes will be needed:

* As briefly mentioned in #2479. The slots might be older than the slot at the tip of the chain after opening, so the policy will never see that slot number and will never be able to take the snapshot. The user could manually delete blocks and resync them to work around this, but that's ugly. The proper solution is to add the ability to take snapshots during ledger replay.
* Even with the above implemented, we still wouldn't be able to make snapshots of slots before the "immutable slot" (the slot at the tip of the last snapshot). We could try to be smart and replay from an older snapshot or from genesis, but that's a bit more complicated. We could also give a warning and suggest the user to delete their on-disk snapshots, triggering a replay from genesis. This functionality should only really used by developers anyway.
* The snapshots consensus writes to disk are of `ExtLedgerState`, which wraps the ledger state as defined by the ledger team. It would probably be convenient for them if we could also write snapshots in the ledger format so that they can easily decode them. For Shelley, that ledger format would be `NewEpochState` (= our `LedgerState (ShelleyBlock era)`), but it might also be possible to transform the `ExtLedgerState` to the ledger's top-level `ChainState` format. Maybe the choice of format (consensus vs ledger) could be a flag?
* It would probably be easiest for the ledger team if they could use the node for this (so they can sync the necessary blocks at the same time), not the consensus-only `db-analyser` tool (which would be a natural place for us to add it). This means the node would take some extra CLI parameters that would have to be passed to consensus.

EncodePanda:
This is unlocked by https://github.com/input-output-hk/ouroboros-network/issues/2479.

EncodePanda:
[...]

PR #2928, opened 2021-02-08
Add absolute SlotNo parameter to LedgerDB OnDiskPolicy

### LINKED ISSUE #2479: Keep older ledger snapshots

EncodePanda:
Closes #2479, prepares necessary changes for #2826

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BlockArguments           #-}
 {-# LANGUAGE DisambiguateRecordFields #-}
 {-# LANGUAGE FlexibleContexts         #-}
 {-# LANGUAGE LambdaCase               #-}
@@ -205,7 +206,7 @@ openDBInternal args launchBgTasks = do
         testing = Internal
           { intCopyToImmutableDB       = getEnv  h Background.copyToImmutableDB
           , intGarbageCollect          = getEnv1 h Background.garbageCollect
-          , intUpdateLedgerSnapshots   = getEnv  h Background.updateLedgerSnapshots
+          , intUpdateLedgerSnapshots   = getEnv  h updateLedgerSnapshots
           , intAddBlockRunner          = getEnv  h Background.addBlockRunner
           , intKillBgThreads           = varKillBgThreads
           }
@@ -220,6 +221,10 @@ openDBInternal args launchBgTasks = do
   where
     tracer = Args.cdbTracer args
     (argsImmutableDb, argsVolatileDb, argsLgrDb, _) = Args.fromChainDbArgs args
+    updateLedgerSnapshots = \case
+      CDB{..} -> do
+        snapshotCandidate <- atomically $ LgrDB.oldestLedgerSnapshot <$> LgrDB.getCurrent cdbLgrDB
+        Background.updateLedgerSnapshots cdbLgrDB snapshotCandidate
 
 isOpen :: IOLike m => ChainDbHandle m blk -> STM m Bool
 isOpen (CDBHandle varState) = readTVar varState <&> \case
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
@@ -58,6 +58,7 @@ import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Config
 import           Ouroboros.Consensus.HardFork.Abstract
 import           Ouroboros.Consensus.Ledger.Abstract
+import           Ouroboros.Consensus.Ledger.Extended (ExtLedgerState)
 import           Ouroboros.Consensus.Ledger.Inspect
 import           Ouroboros.Consensus.Ledger.SupportsProtocol
 import           Ouroboros.Consensus.Protocol.Abstract
@@ -69,8 +70,8 @@ import           Ouroboros.Consensus.Util.ResourceRegistry
 import           Ouroboros.Consensus.Storage.ChainDB.API (BlockComponent (..))
 import           Ouroboros.Consensus.Storage.ChainDB.Impl.ChainSel
                      (addBlockSync)
-import           Ouroboros.Consensus.Storage.ChainDB.Impl.LgrDB
-                     (LgrDbSerialiseConstraints)
+import           Ouroboros.Consensus.Storage.ChainDB.Impl.LgrDB (LgrDB,
+                     LgrDbSerialiseConstraints)
 import qualified Ouroboros.Consensus.Storage.ChainDB.Impl.LgrDB as LgrDB
 import           Ouroboros.Consensus.Storage.ChainDB.Impl.Types
 import qualified Ouroboros.Consensus.Storage.ImmutableDB as ImmutableDB
@@ -216,6 +217,8 @@ copyToImmutableDB CDB{..} = withCopyLock $ do
 --   When enough blocks (depending on its config) have been replayed during
 --   startup, a snapshot of the replayed LgrDB will be written to disk at the
 --   start of this function.
+-- * We write the /oldest/ ledger state to disk because the intention is to only
+--   write ledger states to disk that we know to be immutable.
 --   NOTE: After this initial snapshot we do not take a snapshot of the LgrDB
 --   until the chain has changed again, irrespective of the LgrDB policy.
 -- * Schedule GC of the VolatileDB ('scheduleGC') for the 'SlotNo' of the most
@@ -244,13 +247,18 @@ copyAndSnapshotRunner
   -> GcSchedule m
   -> Word64 -- ^ Number of immutable blocks replayed on ledger DB startup
   -> m Void
-copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed =
-    if onDiskShouldTakeSnapshot Nothing replayed then do
-      updateLedgerSnapshots cdb
-      now <- getMonotonicTime
-      loop (Just now) 0
-    else
-      loop Nothing replayed
+copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed = do
+    snapshotCandidate <- atomically $ LgrDB.oldestLedgerSnapshot <$> LgrDB.getCurrent cdbLgrDB
+    case getTipSlot snapshotCandidate of
+      Origin ->
+        loop Nothing replayed
+      NotOrigin slotNo ->
+        if onDiskShouldTakeSnapshot Nothing replayed slotNo then do
+          updateLedgerSnapshots cdbLgrDB snapshotCandidate
+          now <- getMonotonicTime
+          loop (Just now) 0
+        else
+          loop Nothing replayed
   where
     SecurityParam k      = configSecurityParam cdbTopLevelConfig
     LgrDB.DiskPolicy{..} = LgrDB.getDiskPolicy cdbLgrDB
@@ -273,11 +281,16 @@ copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed =
       let distance' = distance + numToWrite
           elapsed   = (\prev -> now `diffTime` prev) <$> mPrevSnapshot
 
-      if onDiskShouldTakeSnapshot elapsed distance' then do
-        updateLedgerSnapshots cdb
-        loop (Just now) 0
-      else
-        loop mPrevSnapshot distance'
+      snapshotCandidate <- atomically $ LgrDB.oldestLedgerSnapshot <$> LgrDB.getCurrent cdbLgrDB
+      case getTipSlot snapshotCandidate of
+        Origin ->
+          loop mPrevSnapshot distance'
+        NotOrigin slotNo ->
+          if onDiskShouldTakeSnapshot elapsed distance' slotNo then do
+            updateLedgerSnapshots cdbLgrDB snapshotCandidate
+            loop (Just now) 0
+          else
+            loop mPrevSnapshot distance'
 
     scheduleGC' :: WithOrigin SlotNo -> m ()
     scheduleGC' Origin             = return ()
@@ -291,18 +304,23 @@ copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed =
             }
           gcSchedule
 
--- | Write a snapshot of the LedgerDB to disk and remove old snapshots
+-- | Take a snapshot of the provided Ledger State
+--
+-- Write a snapshot of the LedgerDB to disk and remove old snapshots
 -- (typically one) so that only 'onDiskNumSnapshots' snapshots are on disk.
+--
 updateLedgerSnapshots ::
-    ( IOLike m
+     ( IOLike m
      , LgrDbSerialiseConstraints blk
      , HasHeader blk
      , IsLedger (LedgerState blk)
      )
-  => ChainDbEnv m blk -> m ()
-updateLedgerSnapshots CDB{..} = do
-    void $ LgrDB.takeSnapshot  cdbLgrDB
-    void $ LgrDB.trimSnapshots cdbLgrDB
+  => LgrDB m blk
+  -> ExtLedgerState blk
+  -> m ()
+updateLedgerSnapshots lgrDB snapshotCandidate = do
+    void $ LgrDB.takeSnapshot  lgrDB snapshotCandidate
+    void $ LgrDB.trimSnapshots lgrDB
 
 {-------------------------------------------------------------------------------
   Executing garbage collection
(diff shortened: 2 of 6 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
```diff
@@ -205,7 +206,10 @@ openDBInternal args launchBgTasks = do
         testing = Internal
           { intCopyToImmutableDB       = getEnv  h Background.copyToImmutableDB
           , intGarbageCollect          = getEnv1 h Background.garbageCollect
-          , intUpdateLedgerSnapshots   = getEnv  h Background.updateLedgerSnapshots
+          , intUpdateLedgerSnapshots   = getEnv  h \case
```
EncodePanda, line 209: I have not spent much time getting familiar with this initialisation. I did what seemed reasonable but maybe there are some dragons here that I'm not aware of.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -70,7 +71,7 @@ import           Ouroboros.Consensus.Storage.ChainDB.API (BlockComponent (..))
 import           Ouroboros.Consensus.Storage.ChainDB.Impl.ChainSel
                      (addBlockSync)
 import           Ouroboros.Consensus.Storage.ChainDB.Impl.LgrDB
-                     (LgrDbSerialiseConstraints)
+                     (LgrDbSerialiseConstraints, LgrDB)
```
nfrisby, line 74: Does `stylish-haskell` catch this? I anticipate that upper case comes before lower case in its sorting.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -34,6 +35,7 @@ module Ouroboros.Consensus.Storage.ChainDB.Impl.LgrDB (
   , ValidateResult(..)
     -- * Previously applied blocks
   , getPrevApplied
+
```
nfrisby, line 38: Extraneous blank line.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -293,21 +295,31 @@ currentPoint = castPoint
              . ledgerState
              . LedgerDB.ledgerDbCurrent
 
+-- | Returns /the oldes/ ledger state currently in LedgerDB. This is just
+-- an alias to ledgerDbAnchor to give more semantic meaning
+getOldestLS :: LedgerDB' l -> ExtLedgerState l
```
nfrisby, line 300: I don't think we use the `LS` abbreviation anywhere else (right?), so let's continue spelling it out for now.

Also, since `getCurrent` from this module is an `STM` action, I suggest renaming this to `oldestLedgerSnapshot`. (That matches the scheme used by `currentPoint` eg.)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -293,21 +295,31 @@ currentPoint = castPoint
...
+  -> m (Maybe (DiskSnapshot, RealPoint blk))
+takeSnapshot LgrDB{ cfg, tracer, hasFS } toSnapshot =
+  wrapFailure (Proxy @blk) $ do
     LedgerDB.takeSnapshot
       tracer
       hasFS
       encodeExtLedgerState'
-      ledgerDB
+      toSnapshot
+
```
nfrisby, line 322: Extraneous blank line.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -12,10 +12,11 @@ module Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy (
 import           Data.Time.Clock (secondsToDiffTime)
 import           Data.Word
 import           NoThunks.Class (NoThunks, OnlyCheckWhnf (..))
-
+import           Cardano.Slotting.Slot (SlotNo)
 import           Control.Monad.Class.MonadTime
 
 import           Ouroboros.Consensus.Config.SecurityParam
+import           Cardano.Slotting.Slot (WithOrigin)
```
nfrisby, line 19: Combine this with the above import.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -12,10 +12,11 @@ module Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy (
 import           Data.Time.Clock (secondsToDiffTime)
 import           Data.Word
 import           NoThunks.Class (NoThunks, OnlyCheckWhnf (..))
-
+import           Cardano.Slotting.Slot (SlotNo)
```
nfrisby, line 15: The blank line was intentional: see "Import lists:" in https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md

`Control.Monad.Class.MonadTime` is (surprisingly) in the network layer's `io-sim-classes`.

It should be:

```
import NoThunks.Class (NoThunks, OnlyCheckWhnf (..))

import Cardano.Slotting.Slot (SlotNo, WithOrigin)

import Control.Monad.Class.MonadTime
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -61,7 +62,11 @@ data DiskPolicy = DiskPolicy {
       --   blocks had to be replayed.
       --
       -- See also 'defaultDiskPolicy'
-    , onDiskShouldTakeSnapshot :: Maybe DiffTime -> Word64 -> Bool
+    , onDiskShouldTakeSnapshot ::
+           Maybe DiffTime
+        -> Word64
+        -> WithOrigin SlotNo
```
nfrisby, line 68: The `Origin` case should always be a `False` decision; I don't anticipate ever wanting a snapshot for the genesis/initial ledger state. (This is why `OnDisk.takeSnapshot` is a no op in the `Origin` case.)

So my instinct is that this argument should be just `SlotNo`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -325,16 +325,11 @@ initStartingWith tracer cfg streamAPI initDb = do
...
   Write to disk
 -------------------------------------------------------------------------------}
-
--- | Take a snapshot of the /oldest ledger state/ in the ledger DB
---
--- We write the /oldest/ ledger state to disk because the intention is to only
--- write ledger states to disk that we know to be immutable. Primarily for
--- testing purposes, 'takeSnapshot' returns the block reference corresponding
--- to the snapshot that we wrote.
+-- | Take snapshot of the Ledger State that is provided as an argument to this
```
nfrisby, line 328: The Haddock syntax `'LedgerState'` would be good here.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
nfrisby: Looks like this testing interface is only used in the `test-storage` test suite. The QSM testing machinery there can generate a "take a snapshot" instruction as part of its test case. So I think you're implementation here seems correct (ie equivalent to what it was before).

If only to stick to the general pattern in this `openDBInternal` function, I suggest defining this lambda alongside `updateLedgerSnapshots`.

In fact, maybe the name `updateLedgerSnapshots` should _still_ be this function you've written here, ie that variable name should keep its former semantics. So we'd move this to `Background.hs` and rename what is currently there. I'm not sure what to call the new function (the one that is given the snapshot to write)... maybe `updateLedgerSnapshotsWith` `addLedgerSnapshot`?

(This comment got big, so let's chat about it on Tuesday.)

nfrisby (changes requested):
Logic looks good. My feedback is about `StyleGuide.md` stuff and variable name choices. And relocating one chunk of code.

Nice comments 👍

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -255,6 +258,10 @@ copyAndSnapshotRunner cdb@CDB{..} gcSchedule replayed =
     SecurityParam k      = configSecurityParam cdbTopLevelConfig
     LgrDB.DiskPolicy{..} = LgrDB.getDiskPolicy cdbLgrDB
 
+    getSlotNo :: ExtLedgerState blk -> WithOrigin SlotNo
```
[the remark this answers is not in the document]
EncodePanda: you mean 

```
ledgerTipSlot ::
     forall blk. UpdateLedger blk
  => LedgerState blk -> WithOrigin SlotNo
ledgerTipSlot = pointSlot . (ledgerTipPoint (Proxy @blk))
```

?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE BlockArguments #-}
```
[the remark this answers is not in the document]
EncodePanda: we agreed to keep BlockArguments :)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
EncodePanda: yes, stylish uses one ordering function, I'm not sure how it treats upper vs lower case, would have to check

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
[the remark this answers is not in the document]
nfrisby: I mixed up `Ouroboros.Consensus.Ledger.Basics` and `.Abstract`. My suggestion is actually to add the `.Basics` import in order to bring in `getTipSlot`.

(Apparently those two modules are separate to avoid a cyclic dependency -- I haven't yet investigated how to simplify that.)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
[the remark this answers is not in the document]
nfrisby: I don't recall reaching that agreement :) Admittedly: I had been awake for ~15 minutes.

However, I think it's moot in this module now: we pulled the `\case` out into its own local binding. That was the only use of `BlockArguments`, wasn't it?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
nfrisby: We decided on our call to just pull it out in a local binding; without doing any of the renaming I suggested above.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -244,17 +247,26 @@ copyAndSnapshotRunner
...
+          updateLedgerSnapshots cdbLgrDB snapshotCandidate
+          now <- getMonotonicTime
+          loop (Just now) 0
+        else
+          loop Nothing replayed
   where
     SecurityParam k      = configSecurityParam cdbTopLevelConfig
     LgrDB.DiskPolicy{..} = LgrDB.getDiskPolicy cdbLgrDB
 
+    -- getSlotNo :: ExtLedgerState blk -> WithOrigin SlotNo
```
nfrisby, line 266: There is commented out code

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -293,21 +294,31 @@ currentPoint = castPoint
              . ledgerState
              . LedgerDB.ledgerDbCurrent
 
+-- | Returns /the oldes/ ledger state currently in LedgerDB. This is just
```
nfrisby, line 297: Typo "oldes"

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -61,7 +64,11 @@ data DiskPolicy = DiskPolicy {
       --   blocks had to be replayed.
       --
       -- See also 'defaultDiskPolicy'
-    , onDiskShouldTakeSnapshot :: Maybe DiffTime -> Word64 -> Bool
```
nfrisby, line 64: The comment on this doesn't yet mention the new parameter

nfrisby (changes requested):
Lots of progress. I caught some typos etc.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
EncodePanda: removed in the commit it was introduced (rebased)

nfrisby:
Thanks for resolving all the comments!

I'm now blocking you on this. The last thing I want to do is double-check my recollection regarding the tests for this area of the code. I think your diff looks good (especially for a feature like this that isn't critical to core node functionality), but due diligence is for us to assess whether the existing tests are sufficient to accidental changes that might be slipping through (to ensure we didn't disturb the adjacent critical features). [This is what we discussed on the call, just writing it here for future reference.]

nfrisby:
This is the first LedgerDB PR I've reviewed, so part of my due diligence is to assess the existing tests.

Key observation: If I comment out the `intUpdateLedgerSnapshots` line, the ChainDB Q-S-M tests still pass. So that test is only checking whether or not taking a snapshot disrupts the rest of the ChainDB behavior.

https://github.com/input-output-hk/ouroboros-network/blob/1efb25d657ca24da00fdcefa17fb6c1eb6909b49/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs#L378-L385

We do also have the LedgerDB's own Q-S-M test, which checks behaviors related to snapshots.

https://github.com/input-output-hk/ouroboros-network/blob/1efb25d657ca24da00fdcefa17fb6c1eb6909b49/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs#L597-L617

Key observation: this repository does not test the snapshot policy itself, which is the primary aspect that this PR enriches/changes. So that's some existing tech debt -- I've now opened Issue #2937 to track it.

nfrisby (approved):
Approving, but please do squash the commits before merging. The primary commits in here already have some good messages. 👍 

Despite my immediately previous comment on this PR, in the context of the above tests already covering the crucial functional behavior of snaphots, and the relative simplicity of this PR's diff, I'm comfortable with Approving by inspection.

The only notable non-functional behavior change that's obvious to me is that we now obtain the (in-memory) snapshot candidate before checking whether to write whereas before we could check first and only obtain it when we needed it. I don't think this will be a noticeable performance hit, given the other steps in that loop.

@EncodePanda: We should do some manual validation once the downstream `cardano-node` exposes a configuration option that let's us exercise this new feature. In fact, maybe we should just make our own branch of that repo to do such validation ASAP, and maybe even end up doing the integration for them/at least bootstrapping it.

EncodePanda:
Squashed to a single commit with following message

```
Modify DiskPolicy so that onDiskShouldTakeSnapshot takes SlotNo as an argument

Disk policy will take SlotNo as an argument. For a moment a richer
information - WithOrigin SlotNo - was discussed, but that idea was
rejected. Probability that anyone will take snapshot of the genesis state
is close to zero. If our users will ever require such functionality,
we can always bring it back.

In order for this new policy to work properly we had to modify
takeSnapshot functionality. Previously takeSnapshot took a decision of
which ledger state to make snapshot off.
This change modifies this behavior. It is up to the caller of the
takeSnapshot function to decide which ledger state to save.
```

EncodePanda:
bors +r

PR #2930, opened 2021-02-10
Add a self-issued predicate to the Consensus interface

### LINKED ISSUE #2926: Provide a predicate testing whether specific identities issued a header's block
For a particular metric the Node Team wants to add, they need a means of classifying a header as "self-issued" or "other-issued".

Consensus currently doesn't provide an interface for that. Byron headers and Shelley headers both ultimately provide that "issuer" data, but differently.

So I'm envisioning that this Issue is done when we export something like this:

```
class HasTestIssuer blk where
  type IssuersOfInterest blk
  testIssuer :: IssuersOfInterest blk -> Header blk -> Bool

instance HasTestIssuer ByronBlock where
  -- hashes of verification key for signing key (delegatee? delegator? whatever bhIssuer contains)
  type IssuersOfInterest ByronBlock = Set KeyHash
  testIssuer = ... _bhIssuer ...

instance (... era ...) => HasTestIssuer (ShelleyBlock era) where
  type IssuersOfInterest (ShelleyBlock era) = Set (KeyHash 'BlockIssuer (Crypto era))
  testIssuer = ... bheaderVk ...

instance (... xs ...) => HasTestIssuer (HardForkBlock xs) where
  type IssuersOfInterest (HardForkBlock xs) = NP WrapIssuersOfInterest xs
  testIssuer = ... testIssuer ...   -- dispatched via the header's era tag
```

With the above, the node's tracer logic will have have a `IssuersOfInterest`, which can only be constructed at some code location that knows the concrete block type.

nfrisby:
Fixes #2926.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Block/SupportsMetrics.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Block/SupportsMetrics.hs
@@ -1,0 +1,33 @@
+-- | See 'BlockSupportsMetrics'.
+module Ouroboros.Consensus.Block.SupportsMetrics (
+    BlockSupportsMetrics (..)
+  , WhetherSelfIssued (..)
+  , isSelfIssuedConstUnknown
+  ) where
+
+import           Ouroboros.Consensus.Block.Abstract
+
+{-------------------------------------------------------------------------------
+  Supported blocks
+-------------------------------------------------------------------------------}
+
+-- | Whether a block was issued by a stakeholder currently forging on this node
+data WhetherSelfIssued =
+    IsSelfIssued
+  | IsNotSelfIssued
+    -- | We are unable to determine
+  | UnknownSelfIssued
+  deriving (Show, Eq)
+
+-- | Evidence that a block supports the metrics needed for business requirements
+--
+-- For example, we use 'isSelfIssued' to help Stake Pool Operators monitor how
+-- many of their forged blocks have not ended up on their chain.
+class BlockSupportsMetrics blk where
+  -- | See 'WhetherSelfIssued'.
+  isSelfIssued :: BlockConfig blk -> Header blk -> WhetherSelfIssued
+
+-- | Use this default for block types that cannot or do not yet support the
+-- predicate
+isSelfIssuedConstUnknown :: BlockConfig blk -> Header blk -> WhetherSelfIssued
+isSelfIssuedConstUnknown _cfg _hdr = UnknownSelfIssued
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
@@ -99,3 +99,8 @@ class ( LedgerSupportsProtocol           blk
       , ShowProxy                 (Query blk)
       , ShowProxy           (TxId (GenTx blk))
       ) => RunNode blk
+  -- This class is intentionally empty. It is not necessarily compositional - ie
+  -- the instance for 'HardForkBlock' might do more than merely delegate to the
+  -- instance for each era - but we want as many of its super classes as
+  -- possible to rely on compositional instances when possible. Not putting any
+  -- methods here helps encourage that.
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
@@ -89,6 +89,7 @@ class ( LedgerSupportsProtocol           blk
       , SerialiseNodeToClientConstraints blk
       , LedgerSupportsPeerSelection      blk
       , NodeInitStorage                  blk
+      , BlockSupportsMetrics             blk
       , Show                (CannotForge blk)
       , Show             (ForgeStateInfo blk)
       , Show      (ForgeStateUpdateError blk)
--- a/ouroboros-consensus-byron-test/src/Ouroboros/Consensus/ByronDual/Node.hs
+++ b/ouroboros-consensus-byron-test/src/Ouroboros/Consensus/ByronDual/Node.hs
@@ -251,4 +251,7 @@ instance NodeInitStorage DualByronBlock where
   RunNode instance
 -------------------------------------------------------------------------------}
 
+instance BlockSupportsMetrics DualByronBlock where
+  isSelfIssued = isSelfIssuedConstUnknown
+
 instance RunNode DualByronBlock
--- a/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node.hs
+++ b/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node.hs
@@ -281,4 +281,7 @@ instance NodeInitStorage ByronBlock where
   RunNode instance
 -------------------------------------------------------------------------------}
 
+instance BlockSupportsMetrics ByronBlock where
+  isSelfIssued = isSelfIssuedConstUnknown
+
 instance RunNode ByronBlock
--- a/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Node.hs
+++ b/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Node.hs
@@ -51,6 +51,9 @@ instance NodeInitStorage (SimpleBlock SimpleMockCrypto ext) where
 
   nodeCheckIntegrity _ _ = True
 
+instance BlockSupportsMetrics (SimpleBlock c ext) where
+  isSelfIssued = isSelfIssuedConstUnknown
+
 instance ( LedgerSupportsProtocol      (SimpleBlock SimpleMockCrypto ext)
          , Show (CannotForge           (SimpleBlock SimpleMockCrypto ext))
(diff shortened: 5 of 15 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
```diff
@@ -98,4 +111,16 @@ class ( LedgerSupportsProtocol           blk
...
-      ) => RunNode blk
+      ) => RunNode blk where
+  -- | A protocol may or may not be able to determine from the 'Header' whether
+  -- the header's block was self-issued.
+  --
+  -- TODO Only the @Shelley@ currently returns anything other than
+  -- 'MaybeSelfIssued'.
+  isSelfIssued :: TopLevelConfig blk -> Header blk -> WhetherSelfIssued
+
+  default isSelfIssued ::
```
nfrisby, line 122: I'm not keen on `default`, but it at least helps keep the diff small for this stage of the PR review.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
```diff
@@ -73,6 +76,16 @@ class ( ConvertRawHash blk
...
       ) => SerialiseNodeToClientConstraints blk
 
+-- | A bespoke 3-valued boolean: yes, no, or maybe.
+data WhetherSelfIssued =
+    -- | A block we produced ourself
+    IsSelfIssued
+    -- | A block produced by another node
+  | IsNotSelfIssued
+    -- | We do not have enough information to decide
+  | MaybeSelfIssued
```
nfrisby, line 86: Rename `Maybe` to `Unknown`; avoids some accidental connotations. Implies either `likely` or `we could figure out with some more effort` -- neither of which is what we want here.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
```diff
@@ -98,4 +111,16 @@ class ( LedgerSupportsProtocol           blk
...
       , ShowProxy                 (Query blk)
       , ShowProxy           (TxId (GenTx blk))
-      ) => RunNode blk
+      ) => RunNode blk where
+  -- | A protocol may or may not be able to determine from the 'Header' whether
+  -- the header's block was self-issued.
+  --
+  -- TODO Only the @Shelley@ currently returns anything other than
+  -- 'MaybeSelfIssued'.
+  isSelfIssued :: TopLevelConfig blk -> Header blk -> WhetherSelfIssued
```
nfrisby, line 120: Add a comment on `RunNode`; here's the idea. We want to keep `RunNode` empty. One reason for keeping it empty is so that delegating-HFC instances can have precise contexts: eg `LedgerSupportsPeerSelection` for HF block requires `LedgerSupportsPeerSelection` for each era. It may be right now, but in the past and in the future, `RunNode` and each of its superclasses need not necessarily be so compositional. Summary: we want the HFC instances to be explicitly compositional when possible.

nfrisby: Related: Pull this method out into its own `SupportsTracing` class; something that obviously exists because of business-level as opposed to correctness concerns (to the extent that those two are separable 👍).

ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
nfrisby: It actually confused us while we were reviewing this small PR ("wait, why does it build without an HFC instance?"). Definitely not going to be explicit!

edsko (approved):
Reviewed over meet.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
```diff
@@ -98,4 +111,16 @@ class ( LedgerSupportsProtocol           blk
       , ShowProxy                (Header blk)
       , ShowProxy                 (Query blk)
       , ShowProxy           (TxId (GenTx blk))
-      ) => RunNode blk
+      ) => RunNode blk where
+  -- | A protocol may or may not be able to determine from the 'Header' whether
+  -- the header's block was self-issued.
+  --
+  -- TODO Only the @Shelley@ currently returns anything other than
```
nfrisby, line 118: We'll open an Issue to support Byron too, since at least on mainnet, no one is forging Byron blocks.

nfrisby: I opened Issue 2936 (see link below).

nfrisby:
@EncodePanda Would you also review and Approve this? I touched more code than I expected while executing the TODOs we came up with with Edsko (who is unavailable this Friday). The overall direction didn't change, but I'd appreciate another careful eye for any typos etc. Thanks!

ouroboros-consensus/src/Ouroboros/Consensus/Block/SupportsMetrics.hs
```diff
@@ -0,0 +1,35 @@
...
+  ) where
+
+import           Ouroboros.Consensus.Block.Abstract
+
+{-------------------------------------------------------------------------------
+  Supported blocks
+-------------------------------------------------------------------------------}
+
+-- | A bespoke 3-valued boolean: yes, no, or unknown
+data WhetherSelfIssued =
```
EncodePanda, line 15: We use three different names for the same thing: issued, forged, produced. Unification on the name (picking one) would makes thing simpler and this amount of comments would not be needed at all, code would be crisp and clear.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
```diff
@@ -99,3 +100,6 @@ class ( LedgerSupportsProtocol           blk
       , ShowProxy                 (Query blk)
       , ShowProxy           (TxId (GenTx blk))
       ) => RunNode blk
+  -- This class is intentionally empty. It is not necessarily compositional, but
+  -- we want as many of its super classes as possible to rely on compositional
+  -- instances when possible. Not putting any methods here helps encourage that.
```
EncodePanda, line 105: I don't quite get what the comment is saying here. What do we understand by "the typeclass is not necessarily compositional" and "we want as many of its super classes as possible to rely on compositional"?

EncodePanda (approved):
Code added is nice and clear. 

I'm not a fun of comments that repeat 1:1 what code crisply says. But that is not a blocker, just my personal preference :)
https://img.devrant.com/devrant/rant/r_1795679_1JDDS.jpg

Few questions asked, but nothing that stops the PR from merging.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Run.hs
nfrisby: Yeah, "compositionality" is pretty high-level/vague. So I added mention of a slightly concrete example to the comment: the HFC `HardForkBlock`. Thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Block/SupportsMetrics.hs
nfrisby: Yeah, this is pervasive, some very nebulous tech debt regarding consistent terminology (eg the research papers mean one thing by "fork" and "hard fork" uses the word completely differently).

For whatever reason, the ledger (eg the definition of header fields) seems to favor `issue` whereas the consensus code uses `forge` and moreover we've recently been told to favor `mint` because `forge` reminds people of `forgery` :/ All of these words might have slightly different connotations (eg I wonder if `issue` is specifically related to public keys) in the blockchain domain that I have yet to come to appreciate.

I at least revised these comments to not juxtapose different terms. I'm glad you mentioned it.

nfrisby:
@newhoggy This PR is ready to merge, but there's no rush on our end.

Would you do some `-node` development against it and let me know if it's sufficient for CAD-2371? I'd rather modify this PR than have to follow-up it with another, causing more churn on `master`. Thanks!

newhoggy (approved):

newhoggy:
> @newhoggy This PR is ready to merge, but there's no rush on our end.
> 
> Would you do some `-node` development against it and let me know if it's sufficient for CAD-2371? I'd rather modify this PR than have to follow-up it with another, causing more churn on `master`. Thanks!

The additions to the API so far seem quite easy to use.

nfrisby:
Update: I'm not going to merge this until input-output-hk/cardano-node#2321 gets full approval from its stakeholders. There's a specific open question regarding which identity we should use. That affects this PR, which is fine as-is only if the cold key identity is appropriate.

newhoggy:
> Update: I'm not going to merge this until [input-output-hk/cardano-node#2321](https://github.com/input-output-hk/cardano-node/pull/2321) gets full approval from its stakeholders. There's a specific open question regarding which identity we should use. That affects this PR, which is fine as-is only if the cold key identity is appropriate.

Thanks for the update.  You mentioned we were using the cold keys, but @disassembler says we should be using the vrf keys.

nfrisby:
The team cleared it up in Slack: cold key is a sufficient and desirable notion of identity for this purpose, we're already using it for Shelley, and the dynamics only require the verification part, not the cold key itself. So this is good to go.

I just rebased without conflicts. Merging now.

Issue #2871, opened 2021-01-14
[CAD-1720] Integrate Alonzo ledger


nfrisby:
This Issue is to integrate the Alonzo ledger Shelley-based era into the Consensus layer.

The document added by PR #2847 should be an extremely useful guide about how to proceed. Just like Allegra and Mary, Alonzo is -- from the Consensus Layer's perspective -- a minor revision of Shelley. So it should follow the same pattern established by Allegra and Mary. That document walks through the necessary steps.

EncodePanda:
There is right now a branch where a bunch of appropriate changes were made https://github.com/input-output-hk/ouroboros-network/tree/EncodePanda/CAD-1720-intergrate-alonzo. We now wait for cardano-ledger-alonzo to provide missing bits.

PR #2932, opened 2021-02-10
Integrate Alonzo era

EncodePanda:
Closes https://github.com/input-output-hk/ouroboros-network/issues/2871


nfrisby:
Just a heads up: it's not immediately obvious to me why `cabal` in the CI builds is failing to satisfy `cardano-ledger-alonzo`.

EncodePanda:
I've fixed one issue that was failing CI BUT we still hit 

```
cabal: Could not resolve dependencies:
[__0] trying: cardano-ledger-alonzo-0.1.0.0 (user goal)
[__1] next goal: ouroboros-consensus-cardano (user goal)
```

for few builds. 

What is surprising is the fact that some builds manage to fetch the dependency just fine, where other struggle.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -79,6 +79,7 @@ import qualified Shelley.Spec.Ledger.API as SL
 
 import           Ouroboros.Consensus.Cardano.Block
 
+
```
nfrisby, line 82: Extra line

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -366,32 +368,42 @@ instance CardanoHardForkConstraints c => CanHardFork (CardanoEras c) where
...
     , translateLedgerView    =
           PCons translateLedgerViewByronToShelleyWrapper
         $ PCons translateLedgerViewAcrossShelley
         $ PCons translateLedgerViewAcrossShelley
+        $ PCons translateLedgerViewAcrossShelley
         $ PNil
     }
   hardForkChainSel =
         -- Byron <-> Shelley, ...
-        TCons (CompareBlockNo :* CompareBlockNo :* CompareBlockNo :* Nil)
```
nfrisby, line 383: Can we change this to `hcPure @(Proxy ____) CompareBlockNo`? If the `___` is too obfuscatory, we maybe should leave it as-is.

nfrisby: Maybe we could do similar for the Shelley eras too.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -398,6 +447,9 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
                     ProtocolParamsTransition {
                         transitionTrigger = triggerHardForkAllegraMary
                       }
+                    ProtocolParamsTransition {
+                        transitionTrigger = triggerHardForkMaryAlonzo
+                      }
```
nfrisby, line 452: We need to update `maxMajorProtVer` too, on old line 415.

ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
```diff
@@ -111,6 +111,9 @@ mkCardanoProtocolInfo genesisByron signatureThreshold genesisShelley initialNonc
       ProtocolParamsMary {
           maryProtVer    = ProtVer 4 0
         }
+      ProtocolParamsAlonzo {
+          alonzoProtVer    = ProtVer 5 0
```
nfrisby, line 115: This should eventually be 5, but I think we might want it to remain 4 for now. Duncan slated a change, hopefully for this sprint, that would let us handle this more smoothly (see #2938); we can discuss next time we chat.

nfrisby: Ahh, wait. I got a little ahead of myself. I was thinking of the `cardano-node` repo, but this is `ouroboros-consensus-cardano:exe:db-analyser`. I'm not sure what we should do here -- continuing the existing pattern is probably fine for an internal tool.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
```diff
@@ -351,6 +359,9 @@ exampleTxBodyAllegra = exampleTxBodyMA exampleCoin
 exampleTxBodyMary :: Core.TxBody StandardMary
 exampleTxBodyMary = exampleTxBodyMA exampleMultiAssetValue
 
+exampleTxBodyAlonzo :: Core.TxBody StandardAlonzo
```
nfrisby, line 362: FYI, this will be a type error once we have the real Alonzo era tag type: Alonzo uses some different types for its transactions (I cannot recall off the top of my head, unfortunately). Just a heads up.

nfrisby:
The translation from Byron to Shelley required "extra" information, as does Mary to Alonzo. It was piped through via the `shelleyLedgerGenesis` field of the Shelley ledger config.

https://github.com/input-output-hk/ouroboros-network/blob/2c63c74f1134ec2fe1a0a9dec6d84645b46ba3bf/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs#L434-L459

That route is what `hardForkEraTranslation`/`EraTranslation` already supports, and I think it would work for Mary-Alonzo too.

I wonder if the `SL.AdditionalGenesisConfig` would also suffice as a new argument to `ShelleyBasedLedgerConfig`.

https://github.com/input-output-hk/ouroboros-network/blob/77c4e0eebb99e3e02db704c3c1ab0e5c307ca4aa/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L101-L106

ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
```diff
@@ -64,6 +74,35 @@ import           Test.Consensus.Cardano.MockCrypto
...
+  arbitrary = undefined
+
+instance Arbitrary (TxBody (AlonzoEra c)) where
+  arbitrary = undefined
+
+instance Arbitrary (PParams (AlonzoEra c)) where
+  arbitrary = undefined
+
+instance Arbitrary (AuxiliaryData (AlonzoEra c)) where
+  arbitrary = undefined
```
JaredCorduan, line 92: @EncodePanda you can get these Arbitrary instances from `Test.Cardano.Ledger.Alonzo.Serialisation.Generators`.

PR #2947, opened 2021-02-23
Add srnEnableDevelopmentalVersions to StdRunNodeArgs

### LINKED ISSUE #2938: Add support for a `testing` configuration option in cardano-node
The following code _statically_ determines which eras, mini protocol versions, ledger queries, and serialization schemes a deployment of `cardano-node` will use. Furthermore, there is some code in `cardano-node` that does the same for other versioned features (eg the "max supported `ProtVer`" that announced via an otherwise-unused field in the header of each minted block).

This Issue is to make these configurations partially dynamic. In particular, we intend to be able to add support in the Network Layer and Consensus Layer for a new miniprotocol version and/or a new ledger era etc but to do so _without necessarily_ enabling that support in a defaultly configured `cardano-node` instance. Instead, a new `cardano-node` configuration option, such as `EnableTesting` or `EnableExperimental` etc would be enabled by the eg Benchmarking and QA Teams when vetting the _next_ release.

(There is an analogous `NodeToClient` that also needs the same treatment, but I've omitted it from all of these snippets.)

https://github.com/input-output-hk/ouroboros-network/blob/e142c7153b9ff0505a84ec68124f5bd80449eebd/ouroboros-consensus/src/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs#L38-L41

https://github.com/input-output-hk/ouroboros-network/blob/a9090675a7c63ae6369288faf91644fcb2c993fc/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L315-L323

https://github.com/input-output-hk/ouroboros-network/blob/1c4b3c7009abb9fd9390f7f61483f990c985cdb4/ouroboros-network/src/Ouroboros/Network/NodeToNode/Version.hs#L27-L51

https://github.com/input-output-hk/ouroboros-network/blob/a9090675a7c63ae6369288faf91644fcb2c993fc/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L216-L240

nfrisby:
Fixes #2938.

The interpretation of this new flag within `stdLowLevelRunNodeArgsIO` determines what the highest NTN and NTC version will be when the flag is disabled. When the flag is enabled, those bounds are ignored, and the relevant instance of `Ouroboros.Consensus.Node.NetworkProtocolVersion.SupportedNetworkProtocolVersion` is used unfiltered.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -614,15 +614,19 @@ stdRunDataDiffusion = runDataDiffusion
 --
 -- See 'stdLowLevelRunNodeArgsIO'.
 data StdRunNodeArgs m blk = StdRunNodeArgs
-  { srnBfcMaxConcurrencyBulkSync :: Maybe Word
-  , srnBfcMaxConcurrencyDeadline :: Maybe Word
-  , srnChainDbValidateOverride   :: Bool
+  { srnBfcMaxConcurrencyBulkSync   :: Maybe Word
+  , srnBfcMaxConcurrencyDeadline   :: Maybe Word
+  , srnChainDbValidateOverride     :: Bool
     -- ^ If @True@, validate the ChainDB on init no matter what
-  , srnDatabasePath              :: FilePath
+  , srnDatabasePath                :: FilePath
     -- ^ Location of the DBs
-  , srnDiffusionArguments        :: DiffusionArguments
-  , srnDiffusionTracers          :: DiffusionTracers
-  , srnTraceChainDB              :: Tracer m (ChainDB.TraceEvent blk)
+  , srnDiffusionArguments          :: DiffusionArguments
+  , srnDiffusionTracers            :: DiffusionTracers
+  , srnEnableInDevelopmentVersions :: Bool
+    -- ^ If @False@, then the node will limit the negotiated NTN and NTC
+    -- versions to the latest " official " release (as chosen by Network and
+    -- Consensus Team, with input from Node Team)
+  , srnTraceChainDB                :: Tracer m (ChainDB.TraceEvent blk)
     -- ^ ChainDB Tracer
   }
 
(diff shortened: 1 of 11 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -655,9 +658,11 @@ stdLowLevelRunNodeArgsIO RunNodeArgs{ rnProtocolInfo } StdRunNodeArgs{..} = do
             networkMagic
             (daDiffusionMode srnDiffusionArguments)
       , llrnNodeToNodeVersions =
-          supportedNodeToNodeVersions (Proxy @blk)
+            disableVersionsAbove NodeToNodeV_5
```
nfrisby, line 661: For now, I've hardcoded the correct values for `blk ~ CardanoBlock`.

But note that this code is polymorphic in `blk`. So we should do one of two things:

  1. Let the user (ie `cardano-node`) supply the upper bound as a `Maybe`.
  2. Add a method to the same class that provides `supportedNodeToNodeVersions` etc so the method defines what is the upper bound (ie what is the latest officially supported/released version).

nfrisby: The `Maybe` would be future-proof/flexible, but the instance would let `ouroboros-consensus` choose the default "stable" version per `blk` type, without `cardano-node` having to know what it is.

Though perhaps it should be `cardano-node` that owns that choice?

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -691,6 +696,11 @@ stdLowLevelRunNodeArgsIO RunNodeArgs{ rnProtocolInfo } StdRunNodeArgs{..} = do
             (\mc bfc -> bfc { bfcMaxConcurrencyBulkSync = mc })
             srnBfcMaxConcurrencyBulkSync
 
+    disableVersionsAbove :: forall k v. Ord k => k -> Map k v -> Map k v
+    disableVersionsAbove latestNonDevelopmentalVersion =
```
coot, line 700: I'd call this function `disableVersiosnStrictlyAbove`, at least for me above sometimes is confusing.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
coot: I'd like `NodeToNodeV_6` to not be masked in the next release (tagged this week and released next week).

coot (changes requested):
Is there a reason why we'd like to mask `NodeToNodeV_6` as developmental version?

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
nfrisby: I changed its semantics a bit, so the name is now more concrete.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
nfrisby: I Slacked Duncan and confirmed that we want `ouroboros-network` to determine the "latest released" versions. So I went with the new method in my most recent force push.

coot (approved):
LGTM

nfrisby:
Thanks, Marcin! I rebased (automatic, no conflicts) and will now bors.

PR #2953, opened 2021-02-25
Add hard fork ledger and consensus configuration queries

DavidEichmann:
Most of this PR is simply implementing serialization of said configurations.

This is prep work for a demo that I'm working on that wants access to the ledger and consensus configs without having to load configuration files.

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -1,14 +1,19 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE ConstraintKinds          #-}
 {-# LANGUAGE DataKinds                #-}
 {-# LANGUAGE DeriveAnyClass           #-}
 {-# LANGUAGE DeriveGeneric            #-}
+{-# LANGUAGE DerivingVia              #-}
 {-# LANGUAGE DisambiguateRecordFields #-}
 {-# LANGUAGE FlexibleContexts         #-}
 {-# LANGUAGE FlexibleInstances        #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase               #-}
 {-# LANGUAGE MultiParamTypeClasses    #-}
 {-# LANGUAGE NamedFieldPuns           #-}
 {-# LANGUAGE OverloadedStrings        #-}
+{-# LANGUAGE PackageImports     #-}
 {-# LANGUAGE RecordWildCards          #-}
 {-# LANGUAGE ScopedTypeVariables      #-}
 {-# LANGUAGE TupleSections            #-}
@@ -47,7 +52,7 @@ import qualified Cardano.Chain.Update as CC.Update
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Forecast
 import           Ouroboros.Consensus.HardFork.History (Bound (boundSlot),
-                     addSlots)
+                     addSlots, dummyEpochInfo)
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.TypeFamilyWrappers
 import           Ouroboros.Consensus.Util (eitherToMaybe)
@@ -78,6 +83,11 @@ import           Cardano.Ledger.Mary.Translation ()
 import qualified Shelley.Spec.Ledger.API as SL
 
 import           Ouroboros.Consensus.Cardano.Block
+
+
+import Codec.Serialise (Serialise(..))
+import Cardano.Ledger.Era (Era)
+import Ouroboros.Consensus.Cardano.Orphans ()
 
 {-------------------------------------------------------------------------------
   Figure out the transition point for Byron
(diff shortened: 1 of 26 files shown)


ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -311,6 +311,8 @@ library
                      , ouroboros-network-framework
                      , ouroboros-network
 
+                     , canonical-json
+                     , aeson
```
DavidEichmann, line 315: It's a bit annoying to add these deps. This is used in the following commit to serialize `ShelleyGenesis` and `GenesisData`. We could instead attempt to create orphan instances.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -331,6 +342,66 @@ data ShelleyPartialLedgerConfig era = ShelleyPartialLedgerConfig {
     }
   deriving (Generic, NoThunks)
 
+instance Era era => Serialise (ShelleyPartialLedgerConfig era) where
+  decode =
+    ShelleyPartialLedgerConfig
+      <$> (ShelleyLedgerConfig
+              <$> decode
+              <*> (SL.Globals
+                    dummyEpochInfo
```
DavidEichmann, line 351: perhaps I should add a comment referring to the relevant comment in `encode` below.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Compat.hs
```diff
@@ -115,6 +115,9 @@ singleEraCompatQuery epochSize slotLen f = go
     goHardFork :: QueryHardFork '[era] result -> m result
     goHardFork GetInterpreter = return $ Qry.mkInterpreter summary
     goHardFork GetCurrentEra  = return $ eraIndexZero
+    goHardFork GetLedgerCfg = undefined
+    goHardFork GetConsensusCfg = undefined
+    -- TODO ^^^ Pattern match(es) are non-exhaustive
```
DavidEichmann, line 120: TODO

DavidEichmann: I'm not entirely sure what to do here. If I understand correctly, we'll need to take the ledger and consensus config as new parameters to `singleEraCompatQuery`. Does that sound right?

DavidEichmann:
@nfrisby, if you're gonna have a look at this, do note that the final commit adds about 2.5MB of heap object residency. I figured this is negligible considering the node currently uses GBs of memory. It's also a very simple solution and gets rid of the pitfall of maintaining a `gdAvvmDistr` field with a cleared (i.e. incorrect) value. It's also seems simpler than the alternative: reloading configuration files every time the query is made. Still, there may have been a good reason to remove this data in the first place, so if you know of one then do say. All I found was #2533.

nfrisby:
Hi David. 

My plate is pretty full, so I might not be able to prioritize this soon :( I'll ask around to try to understand the relative priorities.

First thought: I missing some context.

> This is prep work for a demo that I'm working on that wants access to the ledger and consensus configs without having to load configuration files.

In particular, since your PR targets the `master` branch, I'd guess this demo is representative of future features that we know we want to support. Is that right? Can you point me towards the context for those features, like an Issue or a wiki page etc? Thanks.

The only comments I could make right now are:

  0) the AVVM thing was before my time, but it does look like maybe a premature optimization (I don't suppose there's a security reason to clear that memory?)
  1) there are some inconsistent styling issues in this diff (eg new import statements are not indented, some new lines are not alphabetically sorted/etc; cf eventually https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md eg CTRL-F for "import")
  2) no red flags; nice code, nice comments 👍 
  3) maybe I missed it: was there a comment motivating serializing via JSON?

DavidEichmann:
@nfrisby Thanks for your initial comments. I've spoken to @dcoutts about this, and I plan to make some changes, so you can hold off on reviewing for now. I may even split this into smaller PRs.

Some changes that I'll plan to make:

* Add a initial ledger state query
* Split the consensus config query into smaller queries for:
  * Security Parameter
  * Per era slot length
  * Per era epoch size
  * Per era partial consensus config

DavidEichmann:
I'm breaking this into smaller PRs

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
```diff
@@ -0,0 +1,44 @@
...
+deriving anyclass instance Serialise ActiveSlotCoeff
+deriving anyclass instance Serialise SL.Network
+deriving anyclass instance Serialise Shelley.Spec.Ledger.BaseTypes.UnitInterval
+deriving anyclass instance Serialise CC.Genesis.Config
+deriving via (SerialiseViaCanonicalJSON CC.Genesis.GenesisData) instance Serialise CC.Genesis.GenesisData
+deriving newtype instance Serialise CC.Genesis.GenesisHash
+deriving anyclass instance Serialise (Cardano.Crypto.Hashing.Hash Cardano.Binary.Raw)
+deriving anyclass instance Serialise Cardano.Crypto.ProtocolMagic.RequiresNetworkMagic
+deriving anyclass instance Serialise CC.CompactAddress
+deriving anyclass instance Serialise Cardano.Chain.UTxO.UTxOConfiguration
```
DavidEichmann, line 44: Can we make some of these non-orphan instances?

PR #2957, opened 2021-02-26
Add advice about how to add new argument in Ouroboros.Consensus.Node

nfrisby:
One commit. Only adds comments.


EncodePanda (approved):
Wow, this is ridiculously old PR that I forgot to review.
Comment still makes sense, thus 👍

Issue #2949, opened 2021-02-23
Add a "requested snapshotting interval" flag


EncodePanda:
# Current situation

Current OnDisk policy for snapshots is following:

```
    onDiskShouldTakeSnapshot (Just timeSinceLast) blocksSinceLast _slotNo =
           timeSinceLast >= secondsToDiffTime (fromIntegral (k * 2))
        || (   blocksSinceLast >= 50_000
            && timeSinceLast > 6 * secondsToDiffTime 60)
    onDiskShouldTakeSnapshot Nothing blocksSinceLast _slotNo =
           blocksSinceLast >= k
```

In the usual case, a new snapshot is taken if either of the following hold:
1. It has been at least k*2 seconds (called snapshotting period) since the last snapshot was taken
2. The last snapshotted ledger state was 50k blocks ago and it5was taken at least 6 minutes ago

# Problem

Benchmarking team has little control over when (or how often) snapshots are taken.

# Solution

Add a new non-mandatory flag called "requested snapshotting interval" that will override the behaviour of `onDiskShouldTakeSnapshot` in the following way:

In the usual case, a new snapshot is taken if either of the following hold:
1. It has been at least `requested snapshotting interval` (in seconds) since the last snapshot was taken. If `requested snapshotting interval` is not set, then it defaults to `k * 2` seconds
2. The last snapshotted ledger state was 50k blocks ago and it was taken at least 6 minutes ago

# Stakeholders:
* benchmarking team (@deepfire)

# Definition of Done
* flag is exposed via consensus
* value from the flag (if set) modifies the behaviour of DiskPolicy such that k*2 is overridden and value from the flag is being used
* a branch on cardano-node is created that integrates the changes from cardano-consensus
  * one can start the node with the newly added flag and see it in action (snapshots are created with a given frequency)

nfrisby:
LGTM.

@EncodePanda I edited the description; I think there was a typo in the logic. (You can use the GitHub UI to double-check my changes.)

One suggestion: the name "minimum snapshotting period" for a value X sounds like the interval between snapshots can be no smaller than X. But that's actually the opposite of what we're quasi-ensuring!5I suggest instead calling it "requested snapshotting interval" eg. And the semantics is that we make a best effort to ensure the interval between any two snapshots only exceeds that value as much as is necessary.

EncodePanda:
@nfrisby I've renamed the flag to `requested snapshotting interval`

PR #2966, opened 2021-03-02
Introduce Snapshot Interval

EncodePanda:
Fixes https://github.com/input-output-hk/ouroboros-network/issues/2949

### DIFF
--- a/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/DiskPolicy.hs
+++ b/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/DiskPolicy.hs
@@ -1,0 +1,267 @@
+{-# LANGUAGE LambdaCase         #-}
+{-# LANGUAGE NamedFieldPuns     #-}
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE TypeApplications   #-}
+
+module Test.Ouroboros.Storage.LedgerDB.DiskPolicy (tests) where
+
+import           Data.Time.Clock (DiffTime, diffTimeToPicoseconds,
+                     picosecondsToDiffTime, secondsToDiffTime)
+import           Data.Word
+
+import           Test.QuickCheck
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+
+import           Ouroboros.Consensus.Config.SecurityParam (SecurityParam (..))
+import           Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy
+                     (DiskPolicy (..), SnapshotInterval (..),
+                     TimeSinceLast (..), defaultDiskPolicy)
+
+tests :: TestTree
+tests =
+    testGroup "DiskPolicy" [
+        testGroup "defaultDiskPolicy" [
+            testProperty "onDiskNumSnapshots"       prop_onDiskNumSnapshots
+          , testProperty "onDiskShouldTakeSnapshot" prop_onDiskShouldTakeSnapshot
+          ]
+      ]
+
+{-------------------------------------------------------------------------------
+  Test inputs
+-------------------------------------------------------------------------------}
+
+-- | This contains sufficient inputs for each property in this module.
+data TestSetup = TestSetup {
+    -- | argument to 'onDiskShouldTakeSnapshot'
+    tsBlocksSince      :: Word64
+    -- | argument to 'defaultDiskPolicy'
+  , tsK                :: SecurityParam
+    -- | argument to 'defaultDiskPolicy'
+  , tsSnapshotInterval :: SnapshotInterval
+    -- | argument to 'onDiskShouldTakeSnapshot'
+  , tsTimeSince        :: TimeSinceLast DiffTime
+  }
+  deriving (Show)
+
+-- | The represented default 'DiskPolicy'
+toDiskPolicy :: TestSetup -> DiskPolicy
+toDiskPolicy ts = defaultDiskPolicy (tsK ts) (tsSnapshotInterval ts)
+
+-- | The result of the represented call to 'onDiskShouldTakeSnapshot'
+shouldTakeSnapshot :: TestSetup -> Bool
+shouldTakeSnapshot ts = onDiskShouldTakeSnapshot
+    (toDiskPolicy ts)
+    (tsTimeSince ts)
+    (tsBlocksSince ts)
+
+{-------------------------------------------------------------------------------
+  Generator and shrinker
+-------------------------------------------------------------------------------}
+
+instance Arbitrary TestSetup where
+  arbitrary = do
+      k <- frequency [
+          (9, choose (0, 3000))
+        , (1, choose (0, maxBound))
+        ]
+
+      -- values within usual expectations
+      let nominal =
+                (,)
+
+            -- 20 k is average number in a Shelley epoch
+            <$> choose (0, 20 * k)
+
+            -- a week is a defensible upper bound on the user input
+            <*> just95 (chooseSeconds 0 oneWeekInSeconds)
+
+      -- values near known cutoffs
+      let interesting =
+                (,)
+
+            <$> curry choose
+                  (minBlocksBeforeSnapshot `div` 2)
+                  (minBlocksBeforeSnapshot * 2)
+
+            <*> ( Just <$> chooseSeconds
+                    (minSecondsBeforeSnapshot `div` 2)
+                    (minSecondsBeforeSnapshot * 2)
+                )
+
+      -- all other conceivable values
+      let wild =
+                (,)
+            <$> choose (0, maxBound)
+            <*> just95 (chooseSeconds 0 oneCenturyInSeconds)
+
+      (b, t) <- frequency [
+          (80, nominal)
+        , (15, interesting)
+        , (5,  wild)
+        ]
+
+      -- this argument is provided from node via flag, we must anticipate values
+      -- to be completely arbitrary. However we still want to keep the distribution
+      -- of those values in such way that more probable values will be
+      -- more frequently test
+      tsSnapshotInterval <- frequency [
+          (45, pure DefaultSnapshotInterval)
+        , (45, RequestedSnapshotInterval <$> chooseSeconds 0      oneWeekInSeconds)
+        , ( 4, RequestedSnapshotInterval <$> chooseSeconds 0 (2 * oneWeekInSeconds))
+        , ( 4, RequestedSnapshotInterval <$> chooseSeconds 0 (3 * oneWeekInSeconds))
+        , ( 1, RequestedSnapshotInterval <$> chooseSeconds 0 (4 * oneWeekInSeconds))
+        , ( 1, RequestedSnapshotInterval <$> chooseSeconds 0 oneCenturyInSeconds)
+        ]
+
+      pure TestSetup {
+          tsBlocksSince = b
+        , tsK           = SecurityParam k
+        , tsSnapshotInterval
+        , tsTimeSince   = maybe NoSnapshotTakenYet TimeSinceLast t
+        }
+    where
+      -- 100 years seems a reasonable upper bound for consideration
+      oneCenturyInSeconds = 100 * 365 * oneDayInSeconds
+      -- one week seems a reasonable upper bound for relevance
+      oneWeekInSeconds = 7 * oneDayInSeconds
+      oneDayInSeconds  = 24 * 60 * 60
+
+      just95 :: Gen a -> Gen (Maybe a)
+      just95 m = frequency [(5, pure Nothing), (95, Just <$> m)]
+
+      -- both bounds are inclusive and in seconds
+      chooseSeconds :: Integer -> Integer -> Gen DiffTime
+      chooseSeconds lo hi = do
+          -- pick a second
+          s <- choose (lo, hi)
+          -- jitter within it
+          let nines = 10 ^ (12 :: Int) - 1
+          offset <- choose (negate nines, nines)
+          pure $ picosecondsToDiffTime $ max lo $ min hi $ s + offset
+
+  shrink (TestSetup x1 x2 x3 x4) = mconcat [
+        (\y -> TestSetup y x2 x3 x4) <$> shrink @Word64 x1
+      , (\y -> TestSetup x1 y x3 x4) <$> shrinkSecurityParam x2
+      , (\y -> TestSetup x1 x2 y x4) <$> shrinkSnapshotInterval x3
+      , (\y -> TestSetup x1 x2 x3 y) <$> shrinkTSL shrinkDiffTime x4
+      ]
+    where
+      shrinkSecurityParam =
+          fmap SecurityParam . shrink @Word64 . maxRollbacks
+
+      shrinkDiffTime =
+          fmap picosecondsToDiffTime
+        . shrink @Integer
+        . diffTimeToPicoseconds
+
+      shrinkTSL shnk = \case
+        NoSnapshotTakenYet -> []
+        TimeSinceLast    d -> NoSnapshotTakenYet : fmap TimeSinceLast (shnk d)
+
+      shrinkSnapshotInterval = \case
+        DefaultSnapshotInterval     -> []
+        RequestedSnapshotInterval d ->
+              DefaultSnapshotInterval
+            : (RequestedSnapshotInterval <$> shrinkDiffTime d)
+
+{-------------------------------------------------------------------------------
+  Properties
+-------------------------------------------------------------------------------}
+
+-- | Check 'onDiskNumSnapshots' of 'defaultDiskPolicy'
+prop_onDiskNumSnapshots :: TestSetup -> Property
+prop_onDiskNumSnapshots ts =
+    -- 'TestSetup' has more information than we need for this property
+      counterexample "should always be 2"
+    $ onDiskNumSnapshots (toDiskPolicy ts) === 2
+
+minBlocksBeforeSnapshot :: Word64
+minBlocksBeforeSnapshot = 50_000
+
+minSecondsBeforeSnapshot :: Integer
+minSecondsBeforeSnapshot = 6 * 60
+
+-- | Check 'onDiskShouldTakeSnapshot' of 'defaultDiskPolicy'
+prop_onDiskShouldTakeSnapshot :: TestSetup -> Property
+prop_onDiskShouldTakeSnapshot ts =
+    counterexample ("decided to take snapshot? " ++ show (shouldTakeSnapshot ts)) $
+    case t of
+      NoSnapshotTakenYet ->
+            counterexample "haven't taken a snapshot yet"
+          $ counterexample "should take snapshot if it processed at least k blocks"
+          $ shouldTakeSnapshot ts === (blocksSinceLast >= k)
+      TimeSinceLast    timeSinceLast ->
+            counterexample "have previously taken a snapshot"
+          $ isDisjunctionOf (shouldTakeSnapshot ts `named` "the decision")
+              [ systemChecksHowMuchTimeHasPassed timeSinceLast
+              , systemChecksHowManyBlocksWereProcessed timeSinceLast
+              ]
+  where
+    TestSetup {
+        tsBlocksSince      = blocksSinceLast
+      , tsK                = SecurityParam k
+      , tsSnapshotInterval = snapshotInterval
+      , tsTimeSince        = t
+      } = ts
+
+    kTimes2 :: DiffTime
+    kTimes2 = secondsToDiffTime $ fromIntegral $ k * 2
+
+    systemChecksHowMuchTimeHasPassed :: DiffTime -> NamedValue Bool
+    systemChecksHowMuchTimeHasPassed timeSinceLast =
+        case snapshotInterval of
+
+          DefaultSnapshotInterval ->
+              (timeSinceLast >= kTimes2) `named`
+                "time since last is greater then 2 * k seconds if snapshot interval is set to default"
+
+          RequestedSnapshotInterval interval ->
+              (timeSinceLast >= interval) `named`
+                "time since last is greater then explicitly requested interval"
+
+    systemChecksHowManyBlocksWereProcessed :: DiffTime -> NamedValue Bool
+    systemChecksHowManyBlocksWereProcessed timeSinceLast =
+        disjunct `named` msg
+      where
+        msg = unwords [
+            "we have processed", show minBlocksBeforeSnapshot
+          , "blocks and it's been more than", show minSecondsBeforeSnapshot
+          , "seconds since last snapshot was taken"
+          ]
+
+        disjunct =
+             blocksSinceLast >= minBlocksBeforeSnapshot
+          && timeSinceLast >= secondsToDiffTime minSecondsBeforeSnapshot
+
+{-------------------------------------------------------------------------------
+  Auxiliary   -- TODO relocate this somewhere more general
+-------------------------------------------------------------------------------}
+
+-- | A value with an associated user-friendly string
+data NamedValue a = NamedValue String a
+
+forgetName :: NamedValue a -> a
+forgetName (NamedValue _s a) = a
+
+infix 0 `named`
+
+named :: a -> String -> NamedValue a
+named = flip NamedValue
+
+-- | Use this instead of @x '===' 'or' ys@ to get a 'counterexample' message
+-- that explains which of the disjuncts were mismatched
+isDisjunctionOf :: NamedValue Bool -> [NamedValue Bool] -> Property
+isDisjunctionOf (NamedValue s b) ds =
+    counterexample msg $ b === any forgetName ds
+  where
+    msg =
+      unlines $
+          (    show b <> " for " <> s
+            <> ", but the " <> show (length ds) <> " disjuncts were: "
+          )
+        : [    "  "
+            <> "disjunct " <> show (i :: Int) <> ": "
+            <> show b' <> " for " <> s'
+          | (i, NamedValue s' b') <- zip [0..] ds
+          ]
--- a/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs
+++ b/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs
@@ -1,0 +1,16 @@
+{-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+module Test.Ouroboros.Storage.LedgerDB.OrphanArbitrary where
+
+import           Test.QuickCheck
+import           Data.Time.Clock (DiffTime, secondsToDiffTime, diffTimeToPicoseconds)
+import           Ouroboros.Consensus.Config.SecurityParam (SecurityParam(..))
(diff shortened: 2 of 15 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -79,16 +85,25 @@ data DiskPolicy = DiskPolicy {
...
-            && timeSinceLast > 6 * secondsToDiffTime 60)
-    onDiskShouldTakeSnapshot Nothing blocksSinceLast =
-           blocksSinceLast >= k
+      let snapshotIntervalSeconds      = fromIntegral $ interval requestedInterval
+          snapshotInterval             = secondsToDiffTime snapshotIntervalSeconds
+          itsBeen50kBlocks             = blocksSinceLast >= 50_000
+          itsBeen6MinSinceLastSnapshot = timeSinceLast > 6 * secondsToDiffTime 60
+      in
+           timeSinceLast >= snapshotInterval
+        || itsBeen50kBlocks && itsBeen6MinSinceLastSnapshot
```
nfrisby, line 106: 👏👏👏

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -17,6 +18,11 @@ import           Control.Monad.Class.MonadTime
 
 import           Ouroboros.Consensus.Config.SecurityParam
 
+
```
nfrisby, line 21: Extra newline.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -79,16 +85,25 @@ data DiskPolicy = DiskPolicy {
 -- take a snapshot roughly every @k@ blocks. It does mean the possibility of
 -- an extra unnecessary snapshot during syncing (if the node is restarted), but
 -- that is not a big deal.
-defaultDiskPolicy :: SecurityParam -> DiskPolicy
-defaultDiskPolicy (SecurityParam k) = DiskPolicy {..}
+defaultDiskPolicy :: SecurityParam -> RequestedInterval -> DiskPolicy
```
nfrisby, line 88: Would you update the Haddock on this to mention the disjunction in the logic of when we take snapshots? EG I'm not seeing anything that covers the `k*2` case. So I think adding an mention of the new5`RequestedInterval` would be an improvement.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -17,6 +18,11 @@ import           Control.Monad.Class.MonadTime
 
 import           Ouroboros.Consensus.Config.SecurityParam
 
+
+data RequestedInterval =
```
nfrisby, line 22: I think we should rename this to `RequestedDiskSnapshotInterval`. Especially when it occurs in eg `Node.hs` as `srnRequestedInterval`, "RequestedInterval" doesn't indicate the semantics as much as I'd like it too.

Also, some Haddock, especially indicating that the `Word64` is seconds.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -618,6 +614,7 @@ data StdRunNodeArgs m blk = StdRunNodeArgs
   , srnBfcMaxConcurrencyDeadline :: Maybe Word
   , srcChainDbValidateOverride   :: Bool
     -- ^ If @True@, validate the ChainDB on init no matter what
+  , srnRequestedInterval         :: RequestedInterval
```
nfrisby, line 617: When renaming `RequestedInterval`, also rename `srnRequestedInterval` accordingly

nfrisby (changes requested):
Looks good; the primary requested change is to make the name `RequestedInterval` more indicative (a pretty word I heard once but avoid using for this: _perspicuous_) and to add some comments.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
EncodePanda: renamed to `RequestedSnapshotInterval`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
EncodePanda: I've updated the comments but I must admit that I dislike them because they do the following:
1. state the obvious what can be clearly seen in the code
example: 
```
"- We want to take a snapshot every 50k blocks or roughly every requested interval."
```

2. comment the usage of the function (as if the function knew exactly how it is being used)
example
```
      -- * The distance in terms of blocks applied to the /oldest/ ledger
      --   snapshot in memory. During normal operation, this is the number of
      --   blocks written to the ImmutableDB since the last snapshot. On
      --   startup, it is computed by counting how many immutable blocks we had
      --   to reapply to get to the chain tip. This is useful, as it allows the
```

EncodePanda: Also the previous comment mentioned following

```
--- (...) when @k = 2160@ (for other values of @k@ we scale
--- proportionally), but not more rapidly than 10 per hour (which is important
--- for bulk sync).
```

No such thing was reflected in the code so I removed the comment. But I'm stating it here explicitly, maybe I'm missing something

EncodePanda: One last remark the comments: if we treat comments as "double bookkeeping" for the logic of our code, then those are not the best tool for the job as they can easily get out of sync and it requires the bookkeeper (human.closely looking at each PR) to keep them at sync. There is a better tool for the job and it is called specs/tests. Tests govern that the change in one breaks the other. And if it is the documentation that we are missing, then we can easily generate it from the spec. My #2 #cents :)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
nfrisby: > Also, some Haddock explaining the semantics and motivation, especially indicating that the Word64 is seconds.

I shouldn't have packed this bit into the same comment ^^^

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
nfrisby: We chatted on our call. We agreed that the comments especially on `defaultDiskPolicy` could have less implementation details. Something like "a disk policy suitable for use with `cardano-node`". The magic number explanations etc belong within the definition as inline comments.

nfrisby:
Status update: TODOs in my mind as follows

  * Update the comment on `deafultDiskPolicy` as we discussed on the call
  * Add Haddock to `RequestedSnapshotInterval` (especially "seconds")
  * The test case you're adding

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -96,15 +88,36 @@ defaultDiskPolicy (SecurityParam k) requestedInterval = DiskPolicy {..}
...
-          snapshotInterval             = secondsToDiffTime snapshotIntervalSeconds
-          itsBeen50kBlocks             = blocksSinceLast >= 50_000
-          itsBeen6MinSinceLastSnapshot = timeSinceLast > 6 * secondsToDiffTime 60
-      in
-           timeSinceLast >= snapshotInterval
-        || itsBeen50kBlocks && itsBeen6MinSinceLastSnapshot
+         timeSinceLast >= snapshotInterval
+      || substantialAmountOfBlocksWereProcessed blocksSinceLast timeSinceLast
+
+   -- | If users never leave their wallet running for long. This would mean
```
nfrisby, line 95: `long. This` -> `long, this`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -96,15 +88,36 @@ defaultDiskPolicy (SecurityParam k) requestedInterval = DiskPolicy {..}
...
 
-    interval (RequestedSnapshotInterval value) = value
-    interval DefaultSnapshotInterval    = k * 2
+    -- | Requested snapshot interval can be explicitly provided by the
+    -- caller (RequestedSnapshotInterval) or the caller might request the default
+    -- snapshot interval (DefaultSnapshotInterval). If the latter then the
+    -- snapshot interval is defaulted to k * 2 seconds - when @k = 2160@ the interval
+    -- defaults to 72 minutes.
+    snapshotInterval = case requestedInterval of
+      (RequestedSnapshotInterval value) -> value
```
nfrisby, line 122: Please remove the outer parentheses from the pattern.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -96,15 +88,36 @@ defaultDiskPolicy (SecurityParam k) requestedInterval = DiskPolicy {..}
...
+   -- that is not a big deal.
+    onDiskShouldTakeSnapshot Nothing blocksSinceLast = blocksSinceLast >= k
+
+    -- | We want to create a snapshot after a substantial amount of blocks were
+    -- processed (hard-coded to 50k blocks). Given the fact that during bootstrap
+    -- a fresh node will see a lot of blocks over a short period of time, we want
+    -- to limit this condition to happen not more often then a fixed amount of
+    -- time (here hard-coded to 6 minutes)
+    substantialAmountOfBlocksWereProcessed blocksSinceLast timeSinceLast =
+      let minBlocksBeforeSnapshot      = 50_000
```
nfrisby, line 111: Style: drop the extra spaces before `=`, unless you think they make it easier to read (I'm guessing they're just holdovers from the previous layout.)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -96,15 +88,36 @@ defaultDiskPolicy (SecurityParam k) requestedInterval = DiskPolicy {..}
...
+
+   -- | If users never leave their wallet running for long. This would mean
+   -- that under some circumstances we would never take a snapshot
+   -- So, on startup (when the 'time since the last snapshot' is `Nothing`),
+   -- we take a snapshot as soon as there are @k@ blocks replayed.
+   -- This means that even if users frequently shut down their wallet, we still
+   -- take a snapshot roughly every @k@ blocks. It does mean the possibility of
+   -- an extra unnecessary snapshot during syncing (if the node is restarted), but
+   -- that is not a big deal.
+    onDiskShouldTakeSnapshot Nothing blocksSinceLast = blocksSinceLast >= k
```
nfrisby, line 103: I would suggest moving the big comment into the body of the match. This is because, in my experience, it's easy to get confused when there are large comments between the various matches of a single function definition.

In fact, I typically favor using using one `case` instead of multiple matches for similar reasons. I think that could be a good fit here, now that you've nicely pulled out the local definitions.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -18,9 +18,15 @@ import           Control.Monad.Class.MonadTime
...
+-- | Length of time, requested by the user, that has to pass after which
+-- a snapshot is taken. It can be:
+--
+-- 1. either explicitly provided by user in seconds
+-- 2. or default value can be requested - DiskPolicy specify what that is
+-- exactly, see `defaultDiskPolicy` as an example
 data RequestedSnapshotInterval =
     DefaultSnapshotInterval
-  | RequestedSnapshotInterval Word64
+  | RequestedSnapshotInterval DiffTime
```
nfrisby, line 29: Great catch, changing this from `Word64 {- seconds -}` to `DiffTime` 👏

nfrisby:
I like that newest commit a lot. I added some relatively superficial comments.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -0,0 +1,67 @@
...
+    kTimes2 = secondsToDiffTime $ fromIntegral $ k * 2
+  in
+    shouldSnapshot === (timeSinceLast >= kTimes2)
+  where
+    diskPolicy = defaultDiskPolicy securityParam requestedInterval
+    shouldSnapshot = (onDiskShouldTakeSnapshot diskPolicy) (Just timeSinceLast) blocksSinceLast
+
+prop_onDiskNumSnapshots :: SecurityParam -> RequestedSnapshotInterval -> Property
+prop_onDiskNumSnapshots securityParam requestedInterval =
+  (diskPolicy & onDiskNumSnapshots) === 2
```
nfrisby, line 65: Please inline `&` here. That style hasn't been done consistently throughout the code base.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -0,0 +1,67 @@
...
+
+prop_shouldSnapshot_case1 :: Word64 -> SecurityParam -> RequestedSnapshotInterval -> Property
+prop_shouldSnapshot_case1 blocksSinceLast securityParam@(SecurityParam k) requestedInterval =
+  shouldSnapshot blocksSinceLast === (blocksSinceLast >= k)
+  where
+    diskPolicy = defaultDiskPolicy securityParam requestedInterval
+    shouldSnapshot = (onDiskShouldTakeSnapshot diskPolicy) Nothing
+
+prop_shouldSnapshot_case2 :: DiffTime -> Word64 -> SecurityParam -> RequestedSnapshotInterval -> Property
+prop_shouldSnapshot_case2 _ _ _ DefaultSnapshotInterval = discard
```
nfrisby, line 45: Instead of using `discard`, you can create a `newtype` in `.OrphanArbitrary` that is understood as the argument to `RequestedSnapshotInterval`. This property would take that new type (eg `ActualRequestedSnapshotInterval`, `RequestedSnapshotIntervalArg`, `RequestedSnapshotIntervalValue` etc) unwrap it, and then apply the `RequestedSnapshotInterval` constructor.

All told, that'd let you avoid having to write this "degenerate" case. In general, `discard` rarely passes "the smell test"; it usually indicates the generator can be tightened.

PS - The full `Arbitrary RequestedSnapshotInterval` instance could then reuse this new one.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs
```diff
@@ -0,0 +1,35 @@
...
+  Orphan Arbitrary instances
+-------------------------------------------------------------------------------}
+
+instance Arbitrary SecurityParam where
+  arbitrary = SecurityParam <$> choose (0, 6)
+  shrink (SecurityParam k) = SecurityParam <$> shrink k
+
+instance Arbitrary RequestedSnapshotInterval where
+  arbitrary =
+    oneof [ requestedSnapshotInterval
```
nfrisby, line 20: I suggest using `frequency` to make the `defaultSnapshotInterval` case much less likely. EG 9:1. We'd rather spend our "budget" on the case that has more parameters to explore.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs
```diff
@@ -0,0 +1,35 @@
...
+requestedSnapshotInterval :: Gen RequestedSnapshotInterval
+requestedSnapshotInterval =
+  RequestedSnapshotInterval <$> arbitrary
+
+defaultSnapshotInterval :: Gen RequestedSnapshotInterval
+defaultSnapshotInterval = pure DefaultSnapshotInterval
+
+instance Arbitrary DiffTime where
+  arbitrary = fromIntegral @Int <$> arbitrary
+  shrink dt = secondsToDiffTime <$> shrink (diffTimeToPicoseconds dt)
```
nfrisby, line 35: Hmm. Without double-checking the definitions, just reading the names here sounds like this `shrink` will be increasing the value by a factor of `Pico`. Can you check? If the names are merely misleading, would you add a comment assuring anyone else who's confused?

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs
```diff
@@ -0,0 +1,35 @@
...
+
+requestedSnapshotInterval :: Gen RequestedSnapshotInterval
+requestedSnapshotInterval =
+  RequestedSnapshotInterval <$> arbitrary
+
+defaultSnapshotInterval :: Gen RequestedSnapshotInterval
+defaultSnapshotInterval = pure DefaultSnapshotInterval
+
+instance Arbitrary DiffTime where
+  arbitrary = fromIntegral @Int <$> arbitrary
```
nfrisby, line 34: I have to check the details every time, but the `Arbitrary Int` instance is often _not_ what people would expect; it's a bit of a notorious wart. Therefore, I generally call `choose` explicitly.

nfrisby:
I reviewed the new test code. I ended up writing a lot of advice in the comments before I realized I was doing that. We can instead discuss it all on the next call we have that's not dominated by `AlonzoEra` integration if you'd rather not read all my words. 👍

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
EncodePanda: This means we have hlint not running for our project. I've added an issue https://github.com/input-output-hk/ouroboros-network/issues/3008

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/DiskPolicy.hs
EncodePanda: It's for readability 

```
    substantialAmountOfBlocksWereProcessed blocksSinceLast timeSinceLast =
      let minBlocksBeforeSnapshot      = 50_000
          minTimeBeforeSnapshot        = 6 * secondsToDiffTime 60
      in    blocksSinceLast >= minBlocksBeforeSnapshot
         && timeSinceLast   >= minTimeBeforeSnapshot
```

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/DiskPolicy.hs
EncodePanda: Nice, I was looking for something like this, normally (when using hedgehog) I would just have a bunch of functions returning generators and would never call arbitrary functions in test (just the generators + forAll).

This small little trick closes the gap. Thanks @nfrisby !

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs
```diff
@@ -0,0 +1,50 @@
...
+  shrink (RequestedSnapshotInterval v) = RequestedSnapshotInterval <$> shrink v
+
+requestedSnapshotInterval :: Gen SnapshotInterval
+requestedSnapshotInterval =
+  RequestedSnapshotInterval <$> arbitrary
+
+defaultSnapshotInterval :: Gen SnapshotInterval
+defaultSnapshotInterval = pure DefaultSnapshotInterval
+
+newtype AlwaysDefaultSnapshotInterval = AlwaysDefaultSnapshotInterval SnapshotInterval
```
nfrisby, line 34: This is what I had in mind. What do you think? An attractive advantage is that it avoids the need for `-fno-warn-incomplete-patterns`.

```
-- | The isolated 'DefaultSnapshotInterval' constructor
data TheDefaultSnapshotInterval = TheDefaultSnapshotInterval

-- | An isolated 'RequestedSnapshotInterval' constructor
newtype ARequestedSnapshotInterval = ARequestedSnapshotInterval DiffTime

fromTheDefaultSnapshotInterval :: TheDefaultSnapshotInterval -> RequestedSnapshotInterval
fromTheDefaultSnapshotInterval TheDefaultSnapshotInterval = DefaultSnapshotInterval

fromARequestedSnapshotInterval :: ARequestedSnapshotInterval -> RequestedSnapshotInterval
fromARequestedSnapshotInterval (ARequestedSnapshotInterval d) = RequestedSnapshotInterval d
```

And then the reuse in the `Arbitrary` instances is in the other direction.

```
instance Arbitrary TheDefaultSnapshotInterval where
  arbitrary = pure TheDefaultSnapshotInterval

instance Arbitrary ARequestedSnapshotInterval where
  arbitrary = ARequestedSnapshotInterval . fromIntegral <$> choose (1, 3600 :: Int)   -- seconds
  -- TODO shrink (see other comment thread about `Pico`)
  
instance Arbitrary SnapshotInterval where
  arbitrary =
    frequency [ (9, fromARequestedSnapshotInterval <$> arbitrary)
              , (1, fromTheDefaultSnapshotInterval <$> arbitrary)
              ]
  -- TODO shrink
```

nfrisby (changes requested):

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/DiskPolicy.hs
```diff
@@ -0,0 +1,90 @@
...
+                prop_shouldSnapshot_case3
+            , testProperty
+                "should take snapshot processed 50k and it's been more then 6 minutes since last snapshot was taken"
+                prop_shouldSnapshot_case4
+        ]
+      ]
+    ]
+
+prop_shouldSnapshot_case1 :: Word64 -> SecurityParam -> SnapshotInterval -> Property
+prop_shouldSnapshot_case1 blocksSinceLast securityParam@(SecurityParam k) snapshotInterval = do
```
nfrisby, line 44: We mentioned in the call that we should remove these `do` keywords throughout this module.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs
EncodePanda: After giving it some thought (and some coding, I have a commit ready but not pushed yet) I must say I dislike this approach as compared to what I wrote. For multiple reasons:

1. A general (arbitrary) instance is written using definitions of the "problem-specific" instances.

It seems more natural that it should be the other way around. 
In my code an arbitrary instance was implemented using two helper functions `requestedSnapshotInterval` and `defaultSnapshotInterval` and only then the two more "problem-specific" instances were defined using one function each.

2. This approach does not remove the need for `-fno-warn-incomplete-patterns`

The only reason that `-fno-warn-incomplete-patterns` was introduce was in a single test where not only we required a specific data constructor (`RequestedSnapshotInterval`) but we were also interested in the *value* of that requested interval.

```
(AlwaysRequestedSnapshotInterval snapshotInterval@(RequestedSnapshotInterval interval))
-----------------------------------------------------------------------  HERE  ---- ^
```

So after applying your idea, the tests looks following:

```
prop_shouldSnapshot_case2 :: DiffTime -> Word64 -> SecurityParam -> ARequestedSnapshotInterval -> Property
prop_shouldSnapshot_case2 timeSinceLast blocksSinceLast securityParam aRequestedSnapshotInterval = do
  -- given
  let
    snapshotInterval = fromARequestedSnapshotInterval aRequestedSnapshotInterval
    interval = undefined --  no what? :) pattern matching? failing? no clue really
    diskPolicy = defaultDiskPolicy securityParam snapshotInterval
  -- when
    shouldSnapshot = (onDiskShouldTakeSnapshot diskPolicy) (Just timeSinceLast) blocksSinceLast
  -- then
  shouldSnapshot === (timeSinceLast >= interval) -- used here
  ```

Please note that in the equivalent test - where we requested the second data constructor (`DefaultSnapthotInterval`) but were not interested in the values of it (because those do not exists) - we would not need the `-fno-warn-incomplete-patterns` at all. This is ONLY the problem when we want to get access to the values of the generated data.

3. This encode seems more boiler-plate with the additional helper functions that are needed instead of doing plain old pattern matching.

Thoughts?

File: ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs (hunk 14 above)
Comment (EncodePanda) [reply ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs:50]:
```
λ show <$> getCurrentTime 
"2021-03-23 13:25:27.497907 UTC"

λ show . utctDayTime <$> getCurrentTime 
"48331.201854s"

λ diffTimeToPicoseconds . utctDayTime <$> getCurrentTime 
48338194408000000

λ shrinkIntegral . diffTimeToPicoseconds . utctDayTime <$> getCurrentTime 
[0,24182675636500000,36274013454750000,42319682363875000,45342516818437500,46853934045718750,47609642659359375,47987496966179688,48176424119589844,48270887696294922,48318119484647461,48341735378823731,48353543325911866,48359447299455933,48362399286227967,48363875279613984,48364613276306992,48364982274653496,48365166773826748,48365259023413374,48365305148206687,48365328210603344,48365339741801672,48365345507400836,48365348390200418,48365349831600209,48365350552300105,48365350912650053,48365351092825027,48365351182912514,48365351227956257,48365351250478129,48365351261739065,48365351267369533,48365351270184767,48365351271592384,48365351272296192,48365351272648096,48365351272824048,48365351272912024,48365351272956012,48365351272978006,48365351272989003,48365351272994502,48365351272997251,48365351272998626,48365351272999313,48365351272999657,48365351272999829,48365351272999915,48365351272999958,48365351272999979,48365351272999990,48365351272999995,48365351272999998,48365351272999999]
```

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OrphanArbitrary.hs
nfrisby: 2. For the `undefined`: `let ARequestedSnapshotInterval interval = aRequestedSnapshotInterval`

We can discuss 1 and 3 on the call.

nfrisby (approved):
LGTM; the only remaining task is to squash/fixup the commits.

EncodePanda:
bors merge

Issue #2951, opened 2021-02-24
Updgrade stylish-haskell to v0.12.2.0


EncodePanda:
The latest release comes with some nice features that we would definitely use in consensus code.

EncodePanda:
Fixed in https://github.com/input-output-hk/ouroboros-network/pull/2981

Issue #2952, opened 2021-02-24
Configure stylish-haskell to format datatypes correctly


EncodePanda:
Blocked by: https://github.com/input-output-hk/ouroboros-network/issues/2951

It seems that all rules in the style guide regarding datatype formatting can be governed by stylish-haskell (version > 0.12.x.x)

EncodePanda:
As it turns out stylish (even in latest 0.12.2.2) still misses some critical components for us to use it. Mainly:

```
  #     equals - How to format equals sign between type constructor and data constructor.
  #     # Possible values:
  #     # - "same_line" -- leave "=" AND data constructor on the same line as the type constructor.
  #     # - "indent N" -- insert a new line and N spaces from the beginning of the next line.
```
We need to have ability to say "equals same line but NOT the data constructor

```   
  #     first field: How to format first field of each record constructor.
  #     # Possible values:
  #     # - "same_line" -- "{" and first field goes on the same line as the data constructor.
  #     # - "indent N" -- insert a new line and N spaces from the beginning of the data constructor
```
We need to say how we indent first field and other fields (because in our case this is different.

This means that this issue is blocked until necessary changes are done in stylish

Jasagredo:
We for now don't plan on having an auto-format on datatypes, but might eventually reconsider. Closing for now.

PR #2968, opened 2021-03-04
Upgrade stylish-haskell to 0.12.2.2

EncodePanda:
Fixes https://github.com/input-output-hk/ouroboros-network/issues/2951

Some of the important changes in stylish since 0.11.0.0

*  Use ghc-lib-parser rather than haskell-src-exts (will work with 8.10.x features)
* when wrapping import lists over multiple lines, you can repeat the module name
* a new Step was added, `module_header`, which formats the export list of a module
* a new option for aligning only adjacent items
* support for aligning MultiWayIf syntax
  
Opens up new possibilities like https://github.com/input-output-hk/ouroboros-network/issues/2952

```
$> nix-shell

$> stylish-haskell --version
stylish-haskell 0.12.2.0
```

### DIFF
--- a/nix/stylish-haskell.nix
+++ b/nix/stylish-haskell.nix
@@ -1,0 +1,16 @@
+{ pkgs
+}:
+let
+  hsPkgs = pkgs.haskell-nix.stackProject {
+    compiler-nix-name = "ghc8104";
+    modules = [ ];
+    src = pkgs.fetchFromGitHub {
+      owner = "jaspervdj";
+      repo = "stylish-haskell";
+      # 0.12.2.0 release
+      rev = "84770e33bb6286c163c3b2b10fa98d264f6672b8";
+      sha256 = "1jc844x8v93xgnl6fjrk31gb2zr96nxbqmgmnc4hdfhfyajh5y7w";
+    };
+  };
+in
+hsPkgs.stylish-haskell.components.exes.stylish-haskell
--- a/nix/stylish-haskell.nix
+++ b/nix/stylish-haskell.nix
@@ -1,8 +1,9 @@
 { pkgs
+, config
 }:
 let
   hsPkgs = pkgs.haskell-nix.stackProject {
-    compiler-nix-name = "ghc8104";
+    compiler-nix-name = config.compiler;
     modules = [ ];
     src = pkgs.fetchFromGitHub {
       owner = "jaspervdj";
--- a/ouroboros-consensus-byron-test/src/Ouroboros/Consensus/ByronDual/Node.hs
+++ b/ouroboros-consensus-byron-test/src/Ouroboros/Consensus/ByronDual/Node.hs
@@ -18,8 +18,8 @@ import           Data.Maybe (fromMaybe)
 
 import qualified Byron.Spec.Ledger.Core as Spec
 import qualified Byron.Spec.Ledger.Delegation as Spec
-import qualified Byron.Spec.Ledger.Update as Spec
 import qualified Byron.Spec.Ledger.UTxO as Spec
+import qualified Byron.Spec.Ledger.Update as Spec
 
 import qualified Test.Cardano.Chain.Elaboration.Block as Spec.Test
 import qualified Test.Cardano.Chain.Elaboration.Delegation as Spec.Test
@@ -29,9 +29,9 @@ import qualified Test.Cardano.Chain.UTxO.Model as Spec.Test
 
 import qualified Cardano.Chain.Block as Impl
 import qualified Cardano.Chain.Genesis as Impl
+import qualified Cardano.Chain.UTxO as Impl
 import qualified Cardano.Chain.Update as Impl
 import qualified Cardano.Chain.Update.Validation.Interface as Impl
-import qualified Cardano.Chain.UTxO as Impl
 
 import           Ouroboros.Consensus.HeaderValidation
 
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -31,7 +31,7 @@ import           Control.Monad.Except (Except, runExcept, throwError)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (listToMaybe, mapMaybe)
 import           Data.Proxy
-import           Data.SOP.Strict ((:.:) (..), NP (..), unComp)
+import           Data.SOP.Strict (NP (..), unComp, (:.:) (..))
 import           Data.Void (Void)
 import           Data.Word
 import           GHC.Generics (Generic)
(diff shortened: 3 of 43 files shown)


nix/stylish-haskell.nix
```diff
@@ -0,0 +1,16 @@
+{ pkgs
```
nfrisby, line 1: The current `stylish-haskell` is pulled in here

https://github.com/input-output-hk/ouroboros-network/blob/e142c7153b9ff0505a84ec68124f5bd80449eebd/shell.nix#L30

May we can make the change on that line? If not, we'd at least have to remove that, probably.

I don't know the details of this. Marcin might, or else the devops-crossteam-ci Slack channel eg.

nfrisby: Ah, I see now that you removed it below. I suppose setting that to `0.12.2.0` didn't work?

I wonder which part of `haskell.nix` upstream would have to update in order to support just changing that line.

EncodePanda: modifying it to 0.12.2.0 did not work, those are some "hard-coded" tools within haskell.nix.
it's just easier to point to an existing release

EncodePanda: Oh, and one more thing, with some things I'm working on atm in stylish we will most likely for a while work on an "unreleased" `0.12.3.0`

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -31,7 +31,7 @@ import           Control.Monad.Except (Except, runExcept, throwError)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (listToMaybe, mapMaybe)
 import           Data.Proxy
-import           Data.SOP.Strict ((:.:) (..), NP (..), unComp)
```
nfrisby, line 34: Punctuation vs alphabet seemed to have changed. (Not list in stylish-haskell CHANGELOG.)

ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
```diff
@@ -410,7 +410,7 @@ runThreadNetwork systemTime ThreadNetworkArgs
       forM vertexInfos0 $ \(coreNodeId, vertexStatusVar, readNodeInfo) -> do
         readTVar vertexStatusVar >>= \case
           VDown ch ldgr -> pure (coreNodeId, readNodeInfo, ch, ldgr)
-          _        -> retry
+          _             -> retry
```
nfrisby, line 413: I don't have any idea where this whitespace is coming from.

ouroboros-consensus/src/Ouroboros/Consensus/Block/Abstract.hs
```diff
@@ -74,11 +74,11 @@ import           Cardano.Slotting.Slot (EpochNo (..), EpochSize (..),
...
 
-import           Ouroboros.Network.Block (pattern BlockPoint, ChainHash (..),
-                     pattern GenesisPoint, HasHeader (..), HeaderFields (..),
-                     HeaderHash, Point, StandardHash, blockHash, blockNo,
-                     blockPoint, blockSlot, castHash, castHeaderFields,
-                     castPoint, pointHash, pointSlot)
+import           Ouroboros.Network.Block (ChainHash (..), HasHeader (..),
+                     HeaderFields (..), HeaderHash, Point, StandardHash,
+                     blockHash, blockNo, blockPoint, blockSlot, castHash,
+                     castHeaderFields, castPoint, pattern BlockPoint,
```
nfrisby, line 80: It's not clear to me what happened here with the pattern synonyms.

ouroboros-consensus-byron-test/src/Ouroboros/Consensus/ByronDual/Node.hs
```diff
@@ -18,8 +18,8 @@ import           Data.Maybe (fromMaybe)
 
 import qualified Byron.Spec.Ledger.Core as Spec
 import qualified Byron.Spec.Ledger.Delegation as Spec
-import qualified Byron.Spec.Ledger.Update as Spec
 import qualified Byron.Spec.Ledger.UTxO as Spec
+import qualified Byron.Spec.Ledger.Update as Spec
```
nfrisby, line 23: Looks like the old comparison ignored case. (Not list in stylish-haskell CHANGELOG.)

nix/stylish-haskell.nix
nfrisby: Looks like `tools` should pick up things that are on Hackage.

https://github.com/input-output-hk/haskell.nix/blob/3f9e7d746895aa9811867a85a3790ba1e3da2798/overlays/tools.nix#L96-L105

Jasper uploaded `0.12.2.0` for the 2020-10-08T12:49:46Z hackage timestamp. The `cabal` file on this branch specifies `index-state: 2020-12-11T00:00:00Z`, which is later. I wonder why that isn't enough for `tools` to find that `stylish-haskell`. :shrug:

Maybe that's a `tools` bug?

nfrisby: Actually, looks like it works on `master` as of today (maybe the index-state was increased recently? /shrug).

```
nfrisby@frisbycomp:~/ouroboros-network$ git lg -1
59e71cb54 - (HEAD, origin/master, origin/bors/staging, origin/HEAD) Merge #2965 (4 days ago) <iohk-bors[bot]>
nfrisby@frisbycomp:~/ouroboros-network$ git diff
diff --git a/shell.nix b/shell.nix
index b25df69a9..b70e01dd8 100644
--- a/shell.nix
+++ b/shell.nix
@@ -27,7 +27,7 @@ let
     ];
 
     tools = {
-      stylish-haskell = "0.11.0.0";
+      stylish-haskell = "0.12.2.0";
       ghcid = "0.8.7";
       cabal = "3.2.0.0";
       # todo: add back the build tools which are actually necessary
nfrisby@frisbycomp:~/ouroboros-network$ nix-shell --run 'stylish-haskell --version' 2>&1 | tail -n1
stylish-haskell 0.12.2.0
```

EncodePanda:
Regarding imports sorting that changed between 0.11.x.x and 0.12.x.x

 `lib/Language/Haskell/Stylish/Ordering.hs` is responsible for sorting elements within the single import list and
   is providing its own `Ord` so will most likely not change in the future (stylish controls the `Ord` instance).
   
   However the list of imports is sorted by ~Ord~ [provided by ghc-parser ](https://github.com/jaspervdj/stylish-haskell/blob/a996292f56810f96372c012ccef6adb52cdab487/lib/Language/Haskell/Stylish/Step/Imports.hs#L443). There will be a ticket in stylish-haskell to unify this so that stylish has control over how imports are sorted (and not delegate that to a dependency).

EncodePanda:
@nfrisby I've managed to unify the GHC version, so it is defined in a single place

EncodePanda:
Closing this PR. We've upgraded stylish haskell just by upgrading version number in 'tools' in https://github.com/input-output-hk/ouroboros-network/pull/2981.

Nevertheless two things came out from the work on this ticket:
1. A need to have a nix formatter (see https://github.com/input-output-hk/ouroboros-network/issues/2980)
2. A single place to define GHC version in Nix (see https://github.com/input-output-hk/ouroboros-network/issues/2982)

PR #2971, opened 2021-03-04
Updating cardano-ledger

JaredCorduan:
* The ledger state now stores an PulsingRewUpdate instead of a
  RewardUpdate. The PulsingRewUpdate allows us to compute the staking
  rewards incrementally instead of all at once. This involves a new
  serialization format for the reward updates.
* Annotated decoders are now used for PParams and PParamsDelta (the
  data structures now store their own serialization). Additionally, the
  LedgerState also now uses an annotated decoder. The serialization in
  consensus has been changed to use the cbor-in-cbor encoding for these
  types (including in the queries).
* A bug was fixed in the reward calculation which was introduced in
  commit 31083fc5cecbe17f3628393072094b130058edd9.
* The ledger repo now uses ghc-8.10.4 and cabal-3.4.0.0.

**Question:** Is this new cbor-in-cbor encoding of the ledger state and the protocol parameters okay?


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -484,27 +486,28 @@ encodeShelleyLedgerState
                        , shelleyLedgerTransition
                        } =
     encodeVersion serialisationFormatVersion2 $ mconcat [
-        CBOR.encodeListLen 2
+        CBOR.encodeListLen 3
```
JaredCorduan, line 489: I assume it is a bug to encode the length as 2, when in fact it contains three items.

nfrisby: 🤦

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -484,27 +486,28 @@ encodeShelleyLedgerState
...
       , encodeWithOrigin encodeShelleyTip shelleyLedgerTip
       , toCBOR shelleyLedgerState
       , encodeShelleyTransition shelleyLedgerTransition
       ]
 
 decodeShelleyLedgerState ::
      forall era s. ShelleyBasedEra era
-  => Decoder s (LedgerState (ShelleyBlock era))
-decodeShelleyLedgerState = decodeVersion [
+  => Decoder s (Lazy.ByteString -> LedgerState (ShelleyBlock era))
```
JaredCorduan, line 497: I am now considering if it would be better to do the cbor-in-cbor encoding in the ledger package itself, so that there could be a `FromCBOR` instance of `LedgerState` and thus break less downstream. I'll maybe tackle that on Monday if the others agree.

nfrisby: @JaredCorduan Edsko and I don't have a strong opinion on where it belongs, though it does seem more like a ledger implementation detail, especially based on how small this PR's diff is.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
nfrisby: Since you're changing the serialization scheme for Shelley ledgers, we already have to discard the snapshots, right? So this bugfix itself doesn't add any further breakage.

The encoder and decoder were wrong but compatible with each other, so the only problem here was that the CBOR itself was technically ill-formed. For example, general CBOR tooling would have failed to parse it.

cc: @dcoutts

nfrisby (approved):
I'm Approving, in case you need to close it. Though I do agree that the serialization changes may be a better fit upstream.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
JaredCorduan: I've talked with @nc6 about this, and we decided that not having these odd cbor-in-cbor `FromCBOR` instances is probably best. I think that storing ledger state to disk using the cbor-in-cbor encoding is less disruptive than other options (and I would probably need help with the other options, namely percolating the annotators up through the types), but others may have a different opinion. I'm not sure he consequences of using the cbor-in-cbor encoding in the queries.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
JaredCorduan: > Since you're changing the serialization scheme for Shelley ledgers, we already have to discard the snapshots, right? So this bugfix itself doesn't add any further breakage.

exactly! in fact, we are bundling several breaking changes to the serialization scheme: the incremental reward calculation, the unaggregated reward update, and the cbor-in-cbor serialization to handle the newly annotated ledger state.

nfrisby (changes requested):
I saw the conversation on Slack; I'm retracting my Approval because of the clarification there that we need to retain backwards compatibility.

PR #2973, opened 2021-03-04
Add a query for the ledger config

DavidEichmann:
See tracking issue for context: input-output-hk/cardano-node#2454

This new query will eventually be used by a cardano-api demo and then hopfuly db-sync to create the initial ledger state. This is extracted from #2953.

### DIFF
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -28,11 +28,12 @@ library
                        Ouroboros.Consensus.Cardano
                        Ouroboros.Consensus.Cardano.Block
                        Ouroboros.Consensus.Cardano.ByronHFC
-                       Ouroboros.Consensus.Cardano.ShelleyHFC
                        Ouroboros.Consensus.Cardano.Condense
                        Ouroboros.Consensus.Cardano.CanHardFork
                        Ouroboros.Consensus.Cardano.Node
+                       Ouroboros.Consensus.Cardano.Orphans
                        Ouroboros.Consensus.Cardano.ShelleyBased
+                       Ouroboros.Consensus.Cardano.ShelleyHFC
 
   build-depends:       base              >=4.9   && <4.15
                      , bytestring        >=0.10  && <0.11
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE DataKinds                #-}
 {-# LANGUAGE DeriveAnyClass           #-}
 {-# LANGUAGE DeriveGeneric            #-}
+{-# LANGUAGE DerivingStrategies       #-}
 {-# LANGUAGE DisambiguateRecordFields #-}
 {-# LANGUAGE FlexibleContexts         #-}
 {-# LANGUAGE FlexibleInstances        #-}
@@ -31,12 +32,15 @@ import           Control.Monad.Except (Except, runExcept, throwError)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (listToMaybe, mapMaybe)
 import           Data.Proxy
-import           Data.SOP.Strict ((:.:) (..), NP (..), unComp)
+import           Data.SOP.Strict (NP (..), unComp, (:.:) (..))
 import           Data.Void (Void)
 import           Data.Word
 import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks)
 
+import           Cardano.Prelude (Natural)
+
+import           Cardano.Binary
 import           Cardano.Crypto.DSIGN (Ed25519DSIGN)
 import           Cardano.Crypto.Hash.Blake2b (Blake2b_224, Blake2b_256)
 
@@ -47,7 +51,7 @@ import qualified Cardano.Chain.Update as CC.Update
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Forecast
 import           Ouroboros.Consensus.HardFork.History (Bound (boundSlot),
-                     addSlots)
+                     addSlots, dummyEpochInfo)
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.TypeFamilyWrappers
 import           Ouroboros.Consensus.Util (eitherToMaybe)
@@ -73,11 +77,14 @@ import           Ouroboros.Consensus.Shelley.Protocol
 
 import           Cardano.Ledger.Allegra.Translation ()
 import           Cardano.Ledger.Crypto (ADDRHASH, DSIGN, HASH)
+import           Cardano.Ledger.Era (Era)
 import qualified Cardano.Ledger.Era as SL
 import           Cardano.Ledger.Mary.Translation ()
 import qualified Shelley.Spec.Ledger.API as SL
+import           Shelley.Spec.Ledger.BaseTypes (ActiveSlotCoeff)
 
 import           Ouroboros.Consensus.Cardano.Block
+import           Ouroboros.Consensus.Cardano.Orphans ()
 
 {-------------------------------------------------------------------------------
   Figure out the transition point for Byron
@@ -277,6 +284,20 @@ data TriggerHardFork =
   | TriggerHardForkNever
   deriving (Show, Generic, NoThunks)
 
+instance ToCBOR TriggerHardFork where
+  toCBOR triggerHardFork = case triggerHardFork of
+    TriggerHardForkAtVersion v -> encodeTag 0 <> toCBOR v
+    TriggerHardForkAtEpoch e   -> encodeTag 1 <> toCBOR e
+    TriggerHardForkNever       -> encodeTag 2
+
+instance FromCBOR TriggerHardFork where
+  fromCBOR = do
+    decodeTag >>= \case
+      0   -> TriggerHardForkAtVersion <$> fromCBOR @Word16
+      1   -> TriggerHardForkAtEpoch <$> fromCBOR @EpochNo
+      2   -> return TriggerHardForkNever
+      tag -> fail $ "TriggerHardFork: unknown tag " ++ show tag
+
 -- | When Byron is part of the hard-fork combinator, we use the partial ledger
 -- config. Standalone Byron uses the regular ledger config. This means that
 -- the partial ledger config is the perfect place to store the trigger
@@ -286,7 +307,26 @@ data ByronPartialLedgerConfig = ByronPartialLedgerConfig {
       byronLedgerConfig    :: !(LedgerConfig ByronBlock)
     , byronTriggerHardFork :: !TriggerHardFork
     }
-  deriving (Generic, NoThunks)
+  deriving stock Generic
+  deriving anyclass (NoThunks)
+
+instance ToCBOR ByronPartialLedgerConfig where
+  toCBOR
+    (ByronPartialLedgerConfig
+      byronLedgerConfig
+      byronTriggerHardFork
+    ) = mconcat [
+            encodeListLen 2
+          , toCBOR @(LedgerConfig ByronBlock) byronLedgerConfig
+          , toCBOR @TriggerHardFork byronTriggerHardFork
+          ]
+
+instance FromCBOR ByronPartialLedgerConfig where
+  fromCBOR = do
+    enforceSize "ByronPartialLedgerConfig" 2
+    ByronPartialLedgerConfig
+      <$> fromCBOR @(LedgerConfig ByronBlock)
+      <*> fromCBOR @TriggerHardFork
 
 instance HasPartialLedgerConfig ByronBlock where
 
@@ -330,6 +370,72 @@ data ShelleyPartialLedgerConfig era = ShelleyPartialLedgerConfig {
     , shelleyTriggerHardFork :: !TriggerHardFork
     }
   deriving (Generic, NoThunks)
+
+instance Era era => FromCBOR (ShelleyPartialLedgerConfig era) where
+  fromCBOR =
+    ShelleyPartialLedgerConfig
+      <$> (ShelleyLedgerConfig
+        <$> fromCBOR @(CompactGenesis era)
+        <*> (SL.Globals
+              -- Globals
+              --
+              -- Note: we can't serialise EpochInfo as it contains
+              -- lambdas, but that's ok because it's value is known
+              -- staticaly. It is always just `dummyEpochInfo` when inside
+              -- a ShelleyPartialLedgerConfi
+              dummyEpochInfo
+              <$> fromCBOR @Word64
+              <*> fromCBOR @Word64
+              <*> fromCBOR @Word64
+              <*> fromCBOR @Word64
+              <*> fromCBOR @Word64
+              <*> fromCBOR @Word64
+              <*> fromCBOR @Natural
+              <*> fromCBOR @Word64
+              <*> fromCBOR @ActiveSlotCoeff
+              <*> fromCBOR @SL.Network
+            )
+      )
+      <*> fromCBOR @TriggerHardFork
+
+instance Era era => ToCBOR (ShelleyPartialLedgerConfig era) where
+  toCBOR
+    (ShelleyPartialLedgerConfig
+      (ShelleyLedgerConfig
+        myCompactGenesis
+        (SL.Globals
+          _epochInfo
+          slotsPerKESPeriod
+          stabilityWindow
+          randomnessStabilisationWindow
+          securityParameter
+          maxKESEvo
+          quorum
+          maxMajorPV
+          maxLovelaceSupply
+          activeSlotCoeff
+          networkId
+        )
+      )
+      triggerHardFork
+    )
+      -- CompactGenesis
+      = toCBOR @(CompactGenesis era) myCompactGenesis
+        -- Globals
+        --
+        -- Note: we don't serialise EpochInfo. See comment in `decode` above.
+        <> toCBOR @Word64 slotsPerKESPeriod
+        <> toCBOR @Word64 stabilityWindow
+        <> toCBOR @Word64 randomnessStabilisationWindow
+        <> toCBOR @Word64 securityParameter
+        <> toCBOR @Word64 maxKESEvo
+        <> toCBOR @Word64 quorum
+        <> toCBOR @Natural maxMajorPV
+        <> toCBOR @Word64 maxLovelaceSupply
+        <> toCBOR @ActiveSlotCoeff activeSlotCoeff
+        <> toCBOR @SL.Network networkId
+        -- TriggerHardFork
+        <> toCBOR @TriggerHardFork triggerHardFork
 
 instance ShelleyBasedEra era => HasPartialLedgerConfig (ShelleyBlock era) where
   type PartialLedgerConfig (ShelleyBlock era) = ShelleyPartialLedgerConfig era
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
@@ -1,0 +1,174 @@
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications  #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Ouroboros.Consensus.Cardano.Orphans () where
+
+import           Cardano.Binary
+import           Cardano.Chain.Common (Address, BlockCount, CompactAddress,
+                     KeyHash, Lovelace)
+import           Cardano.Chain.Delegation (Certificate)
+import           Cardano.Chain.Genesis
+import           Cardano.Chain.UTxO
+import           Cardano.Chain.Update (ProtocolParameters)
+import           Cardano.Crypto.ProtocolMagic
+import           Cardano.Crypto.Signing.Redeem
+import           Data.Map (Map)
+import           Data.Set (Set)
+import           Data.Word (Word64)
+import           Ouroboros.Consensus.Util.Orphans ()
+
+{-------------------------------------------------------------------------------
+  Orphan Instances
+
+  The following instances are for types in the `cardano-ledger-specs` repo.
+  These should eventually be moved to their respective packages.
+-------------------------------------------------------------------------------}
+
+{-------------------------------------------------------------------------------
+  From Package: cardano-ledger-byron
+-------------------------------------------------------------------------------}
+
+instance ToCBOR Config where
+  toCBOR
+    (Config
+      configGenesisData_
+      configGenesisHash_
+      configReqNetMagic_
+      configUTxOConfiguration_
+    ) = mconcat [
+            encodeListLen 4
+          , toCBOR @GenesisData configGenesisData_
+          , toCBOR @GenesisHash configGenesisHash_
+          , toCBOR @RequiresNetworkMagic configReqNetMagic_
+          , toCBOR @UTxOConfiguration configUTxOConfiguration_
+          ]
+
+instance FromCBOR Config where
+  fromCBOR = do
+    enforceSize "Config" 4
+    Config
+      <$> fromCBOR @GenesisData
+      <*> fromCBOR @GenesisHash
+      <*> fromCBOR @RequiresNetworkMagic
+      <*> fromCBOR @UTxOConfiguration
+
+
+instance ToCBOR GenesisData where
+  toCBOR
+    (GenesisData
+      gdGenesisKeyHashes_
+      gdHeavyDelegation_
+      gdStartTime_
+      gdNonAvvmBalances_
+      gdProtocolParameters_
+      gdK_
+      gdProtocolMagicId_
+      gdAvvmDistr_
+    ) = mconcat [
+            encodeListLen 8
+          , toCBOR @GenesisKeyHashes gdGenesisKeyHashes_
+          , toCBOR @GenesisDelegation gdHeavyDelegation_
+          , toCBOR {- @UTCTime -} gdStartTime_
+          , toCBOR @GenesisNonAvvmBalances gdNonAvvmBalances_
+          , toCBOR @ProtocolParameters gdProtocolParameters_
+          , toCBOR @BlockCount gdK_
+          , toCBOR @ProtocolMagicId gdProtocolMagicId_
+          , toCBOR @GenesisAvvmBalances gdAvvmDistr_
+          ]
+
+instance FromCBOR GenesisData where
+  fromCBOR = do
+    enforceSize "GenesisData" 8
+    GenesisData
+      <$> fromCBOR @GenesisKeyHashes
+      <*> fromCBOR @GenesisDelegation
+      <*> fromCBOR -- @UTCTime
+      <*> fromCBOR @GenesisNonAvvmBalances
+      <*> fromCBOR @ProtocolParameters
+      <*> fromCBOR @BlockCount
+      <*> fromCBOR @ProtocolMagicId
+      <*> fromCBOR @GenesisAvvmBalances
+
+instance ToCBOR GenesisKeyHashes where
+  toCBOR (GenesisKeyHashes gkh) = encodeListLen 1 <> toCBOR @(Set KeyHash) gkh
+
+instance FromCBOR GenesisKeyHashes where
+  fromCBOR = do
+    enforceSize "GenesisKeyHashes" 1
+    GenesisKeyHashes <$> fromCBOR @(Set KeyHash)
+
+instance ToCBOR GenesisDelegation where
+  toCBOR (UnsafeGenesisDelegation gd)
+    = encodeListLen 1
+      <> toCBOR @(Map KeyHash Certificate) gd
+
+instance FromCBOR GenesisDelegation where
+  fromCBOR = do
+    enforceSize "GenesisDelegation" 1
+    UnsafeGenesisDelegation <$> fromCBOR @(Map KeyHash Certificate)
+
+instance ToCBOR GenesisNonAvvmBalances where
+  toCBOR (GenesisNonAvvmBalances gnab)
+    = encodeListLen 1
+      <> toCBOR @(Map Address Lovelace) gnab
+
+instance FromCBOR GenesisNonAvvmBalances where
+  fromCBOR = do
+    enforceSize "GenesisNonAvvmBalances" 1
+    GenesisNonAvvmBalances <$> fromCBOR @(Map Address Lovelace)
+
+instance ToCBOR GenesisAvvmBalances where
+  toCBOR (GenesisAvvmBalances gab)
+    = encodeListLen 1
+      <> toCBOR @(Map CompactRedeemVerificationKey Lovelace) gab
+
+instance FromCBOR GenesisAvvmBalances where
+  fromCBOR = do
+    enforceSize "GenesisAvvmBalances" 1
+    GenesisAvvmBalances <$> fromCBOR @(Map CompactRedeemVerificationKey Lovelace)
+
+instance ToCBOR UTxOConfiguration where
+  toCBOR (UTxOConfiguration tcAssetLockedSrcAddrs_)
+    = encodeListLen 1
+    <> toCBOR @(Set CompactAddress) tcAssetLockedSrcAddrs_
+
+instance FromCBOR UTxOConfiguration where
+  fromCBOR = do
+    enforceSize "UTxOConfiguration" 1
+    UTxOConfiguration <$> fromCBOR @(Set CompactAddress)
+
+{-------------------------------------------------------------------------------
+  From Package: cardano-crypto-wrapper
+-------------------------------------------------------------------------------}
+
+instance ToCBOR CompactRedeemVerificationKey where
+  toCBOR (CompactRedeemVerificationKey a b c d)
+    = mconcat [
+        encodeListLen 4
+      , toCBOR @Word64 a
+      , toCBOR @Word64 b
+      , toCBOR @Word64 c
+      , toCBOR @Word64 d
+      ]
+
+instance FromCBOR CompactRedeemVerificationKey where
+  fromCBOR = do
+    enforceSize "CompactRedeemVerificationKey" 1
+    CompactRedeemVerificationKey
+      <$> fromCBOR @Word64
+      <*> fromCBOR @Word64
+      <*> fromCBOR @Word64
+      <*> fromCBOR @Word64
+
+instance ToCBOR RequiresNetworkMagic where
+  toCBOR = \case
+    RequiresNoMagic -> encodeTag 0
+    RequiresMagic   -> encodeTag 1
+
+instance FromCBOR RequiresNetworkMagic where
+  fromCBOR = decodeTag >>= \case
+    0   -> return RequiresNoMagic
+    1   -> return RequiresMagic
+    tag -> fail $ "RequiresNetworkMagic: unknown tag " ++ show tag
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
@@ -23,7 +23,7 @@ import           Data.Word (Word64)
 import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks (..))
 
-import           Cardano.Binary (FromCBOR, ToCBOR)
+import qualified Cardano.Binary
 
 import           Ouroboros.Network.Magic (NetworkMagic (..))
 
@@ -115,7 +115,7 @@ newtype CompactGenesis era = CompactGenesis {
       getCompactGenesis :: SL.ShelleyGenesis era
     }
   deriving stock (Eq, Show, Generic)
-  deriving newtype (FromCBOR, ToCBOR)
+  deriving newtype (Cardano.Binary.FromCBOR, Cardano.Binary.ToCBOR)
 
 deriving anyclass instance ShelleyBasedEra era => NoThunks (CompactGenesis era)
 
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -313,7 +313,6 @@ library
                      , ouroboros-network-framework
                      , ouroboros-network
 
-
   if os(windows)
      build-depends:    Win32 >= 2.6.2.0
   else
--- a/ouroboros-consensus/src/Data/SOP/Strict.hs
+++ b/ouroboros-consensus/src/Data/SOP/Strict.hs
@@ -39,6 +39,8 @@ import           Data.Coerce
 import           Data.Kind (Type)
 import           NoThunks.Class (NoThunks (..), allNoThunks)
 
+import           Cardano.Binary (FromCBOR (..), ToCBOR (..))
+import           Cardano.Prelude (Typeable)
 import           Data.SOP hiding (Injection, NP (..), NS (..), hd, injections,
                      shiftInjection, tl, unZ)
 import           Data.SOP.Classes (Same)
@@ -59,6 +61,14 @@ type instance AllN       NP c = All c
 type instance AllZipN    NP c = AllZip c
 type instance Prod       NP   = NP
 type instance SListIN    NP   = SListI
+
+instance (Typeable (NP f xs), All (Compose ToCBOR f) xs) => ToCBOR (NP f xs) where
+  toCBOR xsTop = hcfoldMap (Proxy @(Compose ToCBOR f)) toCBOR xsTop
+
+instance (Typeable (NP f xs), All (Compose FromCBOR f) xs) => FromCBOR (NP f xs) where
+  fromCBOR = hsequence' (hcpure
+    (Proxy @(Compose FromCBOR f))
+    (Comp fromCBOR))
 
 hd :: NP f (x ': xs) -> f x
 hd (x :* _) = x
--- a/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Types.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Types.hs
@@ -1,3 +1,4 @@
+
 {-# LANGUAGE DataKinds                  #-}
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE DerivingVia                #-}
@@ -25,6 +26,7 @@ module Ouroboros.Consensus.BlockchainTime.WallClock.Types (
   , slotLengthToMillisec
   ) where
 
+import           Cardano.Binary
 import           Codec.Serialise
 import           Control.Exception (assert)
 import           Data.Fixed
@@ -32,6 +34,7 @@ import           Data.Time (NominalDiffTime, UTCTime, addUTCTime, diffUTCTime)
 import           GHC.Generics (Generic)
 import           NoThunks.Class (InspectHeap (..), NoThunks,
                      OnlyCheckWhnfNamed (..))
+import           Ouroboros.Consensus.Util.Orphans ()
 import           Quiet
 
 {-------------------------------------------------------------------------------
@@ -42,7 +45,8 @@ import           Quiet
 --
 -- Slots are counted from the system start.
 newtype SystemStart = SystemStart { getSystemStart :: UTCTime }
-  deriving (Eq, Generic)
+  deriving stock (Eq, Generic)
+  deriving newtype (FromCBOR, ToCBOR)
   deriving NoThunks via InspectHeap SystemStart
   deriving Show via Quiet SystemStart
 
@@ -98,7 +102,8 @@ data SystemTime m = SystemTime {
 
 -- | Slot length
 newtype SlotLength = SlotLength { getSlotLength :: NominalDiffTime }
-  deriving (Eq, Generic, NoThunks)
+  deriving stock (Eq, Generic)
+  deriving newtype (NoThunks)
   deriving Show via Quiet SlotLength
 
 -- | Constructor for 'SlotLength'
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
@@ -6,12 +6,14 @@
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE KindSignatures             #-}
+{-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE RecordWildCards            #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeApplications           #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE UndecidableInstances       #-}
 
 module Ouroboros.Consensus.HardFork.Combinator.Basics (
     -- * Hard fork protocol, block, and ledger state
@@ -43,6 +45,7 @@ import           Data.Typeable
 import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks)
 
+import           Cardano.Binary
 import           Cardano.Slotting.EpochInfo
 
 import           Ouroboros.Consensus.Block.Abstract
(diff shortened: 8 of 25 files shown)


ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
```diff
@@ -0,0 +1,44 @@
...
+deriving anyclass instance Serialise ActiveSlotCoeff
+deriving anyclass instance Serialise SL.Network
+deriving anyclass instance Serialise Shelley.Spec.Ledger.BaseTypes.UnitInterval
+deriving anyclass instance Serialise CC.Genesis.Config
+deriving via (SerialiseViaCanonicalJSON CC.Genesis.GenesisData) instance Serialise CC.Genesis.GenesisData
+deriving newtype instance Serialise CC.Genesis.GenesisHash
+deriving anyclass instance Serialise (Cardano.Crypto.Hashing.Hash Cardano.Binary.Raw)
+deriving anyclass instance Serialise Cardano.Crypto.ProtocolMagic.RequiresNetworkMagic
+deriving anyclass instance Serialise CC.CompactAddress
+deriving anyclass instance Serialise Cardano.Chain.UTxO.UTxOConfiguration
```
DavidEichmann, line 44: TODO Can we make some of these non-orphan instances?

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
```diff
@@ -309,14 +309,17 @@ answerQueryAnytime HardForkLedgerConfig{..} =
...
 
 deriving instance Show (QueryHardFork xs result)
 
 instance All SingleEraBlock xs => ShowQuery (QueryHardFork xs) where
-  showResult GetInterpreter = show
-  showResult GetCurrentEra  = show
+  showResult query = case query of
+    GetInterpreter  -> show
+    GetCurrentEra   -> show
+    GetLedgerCfg    -> const "TODO show `HardForkLedgerConfig xs`"
```
DavidEichmann, line 322: Is this just for debugging? Is it ok to show a truncated version of the `HardForkLedgerConfig`?

DavidEichmann:
Update:
* Removed some commented out code
* Formatting
* Added comments on the addition of orphan instances.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
```diff
@@ -309,14 +309,19 @@ answerQueryAnytime HardForkLedgerConfig{..} =
...
 
 instance All SingleEraBlock xs => ShowQuery (QueryHardFork xs) where
-  showResult GetInterpreter = show
-  showResult GetCurrentEra  = show
+  showResult query = case query of
+    GetInterpreter -> show
+    GetCurrentEra  -> show
+    -- As the HardForkLedgerConfig is quite large and we want to avoid excessive
+    -- debug output, we truncate the contents of HardForkLedgerConfig here.
+    GetLedgerCfg   -> const "HardForkLedgerConfig{..}"
```
nfrisby, line 324: Would it be easy to use `Data.Data.showConstr` instead of hardcoding the current ctor name?

ouroboros-consensus/src/Ouroboros/Consensus/Util/Counting.hs
```diff
@@ -66,6 +68,13 @@ import           Ouroboros.Consensus.Util.SOP
 
 newtype Exactly xs a = Exactly { getExactly :: NP (K a) xs }
 
+-- TODO This could be a deriving newtype instance, but the constraints will be a bit more complicated.
```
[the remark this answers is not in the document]
nfrisby: I would prefer `NOTE`, yep, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
```diff
@@ -309,14 +309,19 @@ answerQueryAnytime HardForkLedgerConfig{..} =
 -------------------------------------------------------------------------------}
 
 data QueryHardFork xs result where
```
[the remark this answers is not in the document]
nfrisby: @DavidEichmann A key part of that comment: if you're changing the queries, we'll have to bump `NodeToClientVersion`. @EncodePanda has more experience with that than I do, though we're both pretty new to it.

nfrisby:
There are various https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md deviations, but we can defer those until @DavidEichmann and @EncodePanda navigate any other requisite changes 👍

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
DavidEichmann: Good idea!

DavidEichmann: Humm looks like I'll need to add a `Data (HardforkLedgerConfig xs)` constraint here and in various other places. GHC can't find an instance because we are polymorphic in `xs`. This doesn't quite seem worth it. Is there some other way to do this that I'm missing? Using `Typeable` to get the type name has a similar problem. A simpler solution could be to just add an explicit patter match so that if the constructor name changes, we'll get a compiler error any way:
```haskell
GetLedgerCfg   -> \HardForkLedgerConfig{} -> "HardForkLedgerConfig{..}"
```

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -313,6 +313,8 @@ library
                      , ouroboros-network-framework
                      , ouroboros-network
 
+                     , canonical-json
+                     , aeson
```
DavidEichmann, line 317: Can someone comment on if adding these deps is acceptable or not?

DavidEichmann (commented):
Thanks for the review @nfrisby. Regarding style, I did a pass over the PR and tweaked things as best I could. If I've missed something then let me know.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
nfrisby: Yeah, that's a good compromise 👍

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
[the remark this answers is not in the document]
DavidEichmann: Is there some test that would catch this stuff? E.g. something that looks at the query types / versions.
I actually don't see any versioning code for QueryHardFork :thinking: @EncodePanda, perhaps you know what's going on here?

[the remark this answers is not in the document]
nfrisby: (Pawel and I were preoccupied today with Alonzo stuff. I'm spending a brief time now trying to give you at least some feedback in the meantime.)

We have golden tests to check if existing serializations change (edit: maybe not for HFC queries?!), but nothing to check for eg new queries, AFAIK. Maybe we could somehow reformulate the generators in `ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs` so that the compiler will warn us if we're missing a query constructor (ie just put a dummy `case` expression in there, with explanatory comment).

Here's an example of where the queries are version-gated; hopefully that gives you enough to chase the existing patterns; let us know if not.

https://github.com/input-output-hk/ouroboros-network/blob/0fbc52ea24f5c111f5814d87a2dd9f10190be45d/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Serialisation/SerialiseNodeToClient.hs#L186

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Serialisation/SerialiseNodeToClient.hs
```diff
@@ -189,6 +188,10 @@ encodeQueryHardFork vHfc = \case
         Enc.encodeListLen 1
       , Enc.encodeWord8 1
       ]
+    Some GetLedgerCfg -> mconcat [
```
nfrisby, line 191: Above this you can see another example of the explicit versioning of these queries.

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
```diff
@@ -83,6 +91,7 @@ import qualified Data.Set as Set
 import           Data.Void
 import           Data.Word (Word64)
 import           GHC.Stack
+import qualified "canonical-json" Text.JSON.Canonical as CJSON
```
nfrisby, line 94: I was surprised to see `-XPackageImports`. Is that because of `hackage-security`? We probably don't need to defend against that conflict. Am I missing another risk?

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
```diff
@@ -386,3 +395,40 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+  ( Aeson.ToJSON a
+  , Aeson.FromJSON a
+  ) =>  Serialise (SerialiseViaAesonJSON a) where
+  encode (SerialiseViaAesonJSON a) = encode $ Aeson.encode a
+  decode = do
+    bs :: Lazy.ByteString <- decode
+    return
+      $ SerialiseViaAesonJSON
+      $ maybe (error "SerialiseViaAesonJSON: Failed to decode") id
+      $ Aeson.decode bs
```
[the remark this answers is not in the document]
nfrisby: `cardano-ledger-specs` provides CBOR serialization via the `ToCBOR` and `FromCBOR` classes, defined here https://github.com/input-output-hk/cardano-base/tree/master/binary/src/Cardano/Binary

https://github.com/input-output-hk/cardano-ledger-specs/blob/38bb3c95ee225445f6b7498e80765dde085665d5/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/Genesis.hs#L100-L109

I don't have the history on why `c-l-s` uses `cardano-binary` and `ouroboros-consensus` instead uses `Serialise`.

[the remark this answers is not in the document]
DavidEichmann: Ah great, so I can remove the need for `SerialiseViaAesonJSON` here. I'll update this soon.

DavidEichmann:
I need to update this PR to allow for backwards compatibility. This means:

* Version gating the new query
* Using explicit To/FromCBOR serialization instances instead of `Serialise` instances.

EncodePanda (changes requested):
I would love to have FromCBOR/ToCBOR being used here instead of Serialise. I don't know if this is not too much to ask though, because you've already spent a lot of time on this PR. Ideally I would just like have consensus using only serialization from cardano-base so we don't provide wrong encodings by accident. 
As mentioned on slack, if there is you need any help from us with this PR, let us know.

DavidEichmann:
As well as using FromCBOR/ToCBOR instead of Serialise, I can also use CBOR-in-CBOR for the query response, so as to ease backwards compatibility issues for that query.

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
```diff
@@ -386,3 +395,40 @@ fib n = round $ phi ** fromIntegral n / sq5
...
+  encode (SerialiseViaCanonicalJSON a) = runIdentity $ do
+    json <- CJSON.toJSON a
+    return . encode $ CJSON.renderCanonicalJSON json
+
+  decode = do
+    bs :: Lazy.ByteString <- decode
+    return
+      $ SerialiseViaCanonicalJSON
+      $ either error (either (error . show) id  . CJSON.fromJSON @(Either SchemaError) @a)
+      $ CJSON.parseCanonicalJSON bs
```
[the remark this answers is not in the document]
nfrisby: I don't see `ToCBOR` nor `FromCBOR` for it in the `Cardano.Chain.Genesis.Data` module that defines it, either.

I suggest we add those instances upstream, hoping we can avoid any JSON dependencies in Consensus altogether.

ouroboros-consensus/ouroboros-consensus.cabal
nfrisby: I just commented elsewhere that I think we can and should avoid them, in favor of `ToCBOR`/`FromCBOR`. The primary cost will be a hopefully small upstream PR to `cardano-ledger-specs`, if I have anticipated things correctly.

DavidEichmann:
I'm still working on this. Should have an update tomorrow.

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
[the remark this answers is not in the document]
DavidEichmann: I've Implemented the instances as orphans in this MR. Do we want move the instances to their respective places before or after merging? See changes to:
* ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
* ouroboros-consensus/src/Ouroboros/Consensus/Util/Orphans.hs

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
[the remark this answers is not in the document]
DavidEichmann: The latest commit adds such warnings and bumps the version number. Question! I plan on adding more hard fork queries after this PR. Will I need a new protocol version for each PR?!? Or do we only need to bump the version between releases? How does the release cycle even work?

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Serialisation/SerialiseNodeToClient.hs
DavidEichmann: updated

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
DavidEichmann: removed

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -1,21 +1,26 @@
...
+{-# LANGUAGE ConstraintKinds            #-}
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE DeriveAnyClass             #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE DerivingVia                #-}
+{-# LANGUAGE DisambiguateRecordFields   #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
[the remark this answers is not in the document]
DavidEichmann: removed

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
```diff
@@ -32,9 +32,13 @@ import           Ouroboros.Consensus.BlockchainTime
 import           Ouroboros.Consensus.Config
 
 import qualified Shelley.Spec.Ledger.API as SL
+import           Shelley.Spec.Ledger.Genesis ()
```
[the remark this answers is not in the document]
DavidEichmann: removed

DavidEichmann:
Updated. I'm waiting for comment now.

DavidEichmann:
Update: removed some unneeded imports to fixed a CI failure.

DavidEichmann:
Update: mu version check condition was inverted :disappointed: Fixed
I can't seem to run tests locally:

```
$ cabal test ouroboros-consensus-cardano-test           
Build profile: -w ghc-8.10.2 -O1
In order, the following will be built (use -v for more details):
 - ouroboros-consensus-cardano-test-0.1.0.0 (test:test) (ephemeral targets)
Preprocessing test suite 'test' for ouroboros-consensus-cardano-test-0.1.0.0..
Building test suite 'test' for ouroboros-consensus-cardano-test-0.1.0.0..
Running 1 test suites...
Test suite test: RUNNING...
cabal: <file descriptor: 11>: hGetContents: invalid argument (invalid byte sequence)
```

Then it just hangs. It seems to create some files in `ouroboros-consensus-cardano-test/test/golden/CardanoNodeToClientVersion7/`. Anyone know what's going on here?

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -33,6 +33,7 @@ library
                        Ouroboros.Consensus.Cardano.CanHardFork
                        Ouroboros.Consensus.Cardano.Node
                        Ouroboros.Consensus.Cardano.ShelleyBased
+                       Ouroboros.Consensus.Cardano.Orphans
```
nfrisby, line 36: This list should be in alphabetical order. Would you fixup the `ShellHFC` bit too? Thanks.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -286,7 +306,30 @@ data ByronPartialLedgerConfig = ByronPartialLedgerConfig {
       byronLedgerConfig    :: !(LedgerConfig ByronBlock)
     , byronTriggerHardFork :: !TriggerHardFork
     }
-  deriving (Generic, NoThunks)
+  deriving stock Generic
+  deriving anyclass (NoThunks)
+
+-- instance Serialise ByronPartialLedgerConfig where
```
nfrisby, line 312: Please remove this comment

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -331,6 +374,72 @@ data ShelleyPartialLedgerConfig era = ShelleyPartialLedgerConfig {
...
+              dummyEpochInfo
+              <$> fromCBOR
+              <*> fromCBOR
+              <*> fromCBOR
+              <*> fromCBOR
+              <*> fromCBOR
+              <*> fromCBOR
+              <*> fromCBOR
+              <*> fromCBOR
+              <*> fromCBOR @ActiveSlotCoeff
```
nfrisby, line 398: Are you applying a particular rule to determine which `fromCBOR`/`toCBOR` occurrences deserve type applications? It's not apparent to me.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
```diff
@@ -0,0 +1,187 @@
...
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE RankNTypes                 #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE TypeApplications           #-}
+{-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE TypeOperators              #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+{-# LANGUAGE NamedFieldPuns             #-}
```
nfrisby, line 18: Move this up to the upper group, please.

nfrisby: Hmm, actually, is it even being used?

I don't see any standalone deriving etc either, so maybe the whole set of LANGUAGE could use a trim.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
```diff
@@ -0,0 +1,187 @@
...
+import           Cardano.Chain.Update (ProtocolParameters)
+import           Cardano.Crypto.ProtocolMagic
+import           Cardano.Crypto.Signing.Redeem
+import           Data.Map (Map)
+import           Data.Set (Set)
+import           Data.Word (Word64)
+import           Ouroboros.Consensus.Util.Orphans ()
+
+{-------------------------------------------------------------------------------
+  Orphan Instances
```
nfrisby, line 36: These all look reasonable, and I like the type applications for avoiding silent changes.

I do wonder: maybe we should consult with the owners of these data types before (potentially) committing to a serialization. Just in case they have plans to eg change these types.

@nc6 @JaredCorduan --- What do you think? Are these types stable and these serializations appropriate? Thanks.

ouroboros-consensus/src/Data/SOP/Strict.hs
```diff
@@ -60,6 +62,19 @@ type instance AllZipN    NP c = AllZip c
 type instance Prod       NP   = NP
 type instance SListIN    NP   = SListI
 
+instance forall k xs . (Typeable (NP k xs), All (Compose ToCBOR k) xs) => ToCBOR (NP k xs) where
```
nfrisby, line 65: I have never seen a `forall` like this on an instance until last week. And now here it is again!

I do wonder: maybe we don't need it? `instance` already binds the free type variables. (I haven't double-checked it but I'm almost certain; else I would have known you could put `forall` here because I would have needed it.)

Not a major comment here.

ouroboros-consensus/src/Data/SOP/Strict.hs
```diff
@@ -60,6 +62,19 @@ type instance AllZipN    NP c = AllZip c
...
+  toCBOR xsTop = go xsTop
```
nfrisby, line 66: My `nix` is currently rebuilding the world :/, so I'll just ask what you think: can we use `hcfoldMap` here instead of this `go`?

ouroboros-consensus/src/Data/SOP/Strict.hs
nfrisby, line 65: I think all the existing definitions use `f` for the tyvar instead of `k`. Please follow suit.

ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Types.hs
```diff
@@ -2,7 +2,9 @@
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE DerivingVia                #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeApplications           #-}
```
nfrisby, line 5: Can we drop these LANGUAGE pragmas?

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
```diff
@@ -54,6 +56,7 @@ import           Ouroboros.Consensus.TypeFamilyWrappers
 import           Ouroboros.Consensus.Util (ShowProxy)
 import           Ouroboros.Consensus.Util.SOP (fn_5)
 
+import           Cardano.Binary
```
nfrisby, line 59: Would you move this up to the appropriate group of imports? There's a discussion in the https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md

It's preferred but not crucial for new modules, but I'd like to preserve the grouping for those modules that already have it. Thanks.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
```diff
@@ -112,6 +112,23 @@ data instance Query (HardForkBlock xs) :: Type -> Type where
     => QueryHardFork (x ': xs) result
     -> Query (HardForkBlock (x ': xs)) result
 
+  -- WARNING: please add new queries to the end of the list here and for nested
```
nfrisby, line 115: Great comment thanks.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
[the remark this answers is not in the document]
nfrisby: > Or do we only need to bump the version between releases? How does the release cycle even work?

Yeah, when merging a PR that touches this, we check to see if the latest version has been included in a release yet. I've only been involved for that for one such case. So my best advice is to consult others, esp Duncan, when the time comes.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
```diff
@@ -309,14 +326,19 @@ answerQueryAnytime HardForkLedgerConfig{..} =
 -------------------------------------------------------------------------------}
 
 data QueryHardFork xs result where
-  GetInterpreter :: QueryHardFork xs (History.Interpreter xs)
-  GetCurrentEra  :: QueryHardFork xs (EraIndex xs)
+  GetInterpreter  :: QueryHardFork xs (History.Interpreter xs)
+  GetCurrentEra   :: QueryHardFork xs (EraIndex xs)
+  GetLedgerCfg    :: QueryHardFork xs (HardForkLedgerConfig xs)
```
nfrisby, line 332: Please add a comment here that refers to the excellent comment at the end of `Query`.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
```diff
@@ -309,14 +326,19 @@ answerQueryAnytime HardForkLedgerConfig{..} =
...
```
nfrisby, line 331: I don't know how disruptive this would be, but why not `GetLedgerConfig`?

If you stick with `GetLedgerCfg`, please remove the extra space added before all the `::`s here.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
```diff
@@ -365,18 +392,20 @@ decodeQueryAnytimeResult :: QueryAnytime result -> forall s. Decoder s result
 decodeQueryAnytimeResult GetEraStart = decode
 
 encodeQueryHardForkResult ::
-     SListI xs
+     (ToCBOR (LedgerConfig (HardForkBlock xs)), SListI xs)
   => QueryHardFork xs result -> result -> Encoding
 encodeQueryHardForkResult = \case
     GetInterpreter -> encode
```
nfrisby, line 398: I opened Issue #3013 to come back and replace these uses of `encode`/`decode` with `toCBOR`/`fromCBOR`.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
nfrisby, line 331: It feels like I should have thought of this sooner... sorry.

The "give me the ledger config" query actually seems orthogonal to HFC. It just happens that HFC is the outermost layer of the blocks on `mainnet`. But that is an implementation detail, not necessarily true for other (future/anticipated) uses of this library.

Should we instead introduce an outer layer of query that would host the generic queries you're working to add? They seem focused on data that is just a consequence of the architecture, largely irrespective of the block type (other than the data varying with the block type).

Thoughts @DavidEichmann @dcoutts @EncodePanda ?

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/PartialConfig.hs
```diff
@@ -79,5 +82,8 @@ class ( UpdateLedger blk
 newtype WrapPartialLedgerConfig    blk = WrapPartialLedgerConfig    { unwrapPartialLedgerConfig    :: PartialLedgerConfig                   blk  }
 newtype WrapPartialConsensusConfig blk = WrapPartialConsensusConfig { unwrapPartialConsensusConfig :: PartialConsensusConfig (BlockProtocol blk) }
 
+deriving newtype instance (Typeable blk, ToCBOR   (PartialLedgerConfig blk))  => ToCBOR   (WrapPartialLedgerConfig    blk)
```
nfrisby, line 85: Please remove the extra spaces before the `=>`s and before the last occurrences of `blk`.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Serialisation/SerialiseNodeToClient.hs
```diff
@@ -173,22 +172,36 @@ instance SerialiseHFC xs
   Queries
 -------------------------------------------------------------------------------}
 
+querySupportedVersion :: Some (QueryHardFork xs)
```
nfrisby, line 175: This is clean, nice. I do wonder if (if only for consistency with other encode functions that I'm recalling) it'd be better to leave the guard near the encoder.

You could get the same re-use in the case expression by pulling the `| otherwise ->` alternative out as a default `_otherwise` alternative like you currently have as a guard. Does that suggestion make sense? What you think? I'm regretting the loss of the locality of putting the guard on the encoder definition itself.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -187,7 +187,7 @@ defaultCodecs ccfg version networkVersion = Codecs {
 
     , cStateQueryCodec =
         codecLocalStateQuery
-          (networkVersion >= NodeToClientV_8)
+          (networkVersion >= NodeToClientV_9)
```
nfrisby, line 190: Looking at the definition of `codecLocalStateQuery`, I'm not sure that this guard should change. What do you think?

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -236,7 +236,7 @@ clientCodecs ccfg version networkVersion = Codecs {
 
     , cStateQueryCodec =
         codecLocalStateQuery
-          (networkVersion >= NodeToClientV_8)
```
nfrisby, line 239: Same question as my previous comment just above.

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
```diff
@@ -1,11 +1,13 @@
...
+{-# LANGUAGE BangPatterns         #-}
+{-# LANGUAGE ConstraintKinds      #-}
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE GADTs                #-}
+{-# LANGUAGE KindSignatures       #-}
+{-# LANGUAGE LambdaCase           #-}
+{-# LANGUAGE PolyKinds            #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE UndecidableInstances #-}
```
nfrisby, line 10: Looks like TypeApplications and UndecdibleInstances are left overs from a previous iteration of this PR.

ouroboros-consensus/src/Ouroboros/Consensus/Util/Counting.hs
```diff
@@ -66,6 +69,19 @@ import           Ouroboros.Consensus.Util.SOP
 
 newtype Exactly xs a = Exactly { getExactly :: NP (K a) xs }
 
+-- NOTE: This could be a deriving newtype instance, but the constraints will be
+-- a bit more complicated.
+instance forall xs a . (Typeable xs, ToCBOR a, SListI xs) => ToCBOR (Exactly xs a) where
```
nfrisby, line 74: Similar to my question on the `NP` instance; maybe use `hfoldMap`?

ouroboros-consensus/src/Ouroboros/Consensus/Util/Counting.hs
```diff
@@ -66,6 +69,19 @@ import           Ouroboros.Consensus.Util.SOP
...
+  toCBOR (Exactly xsTop) = go xsTop
+    where
+      go :: NP (K a) xs' -> Encoding
+      go xs = case xs of
+        Nil        -> mempty
+        K x :* xs' -> toCBOR x <> go xs'
+
+instance (Typeable xs, FromCBOR a, SListI xs) => FromCBOR (Exactly xs a) where
+  fromCBOR = Exactly <$> (htraverse' (\_ -> K <$> fromCBOR) (hpure Proxy))
```
nfrisby, line 83: Please drop the undue parens. Consider also replacing `_` with `Proxy` or maybe `_Proxy` --- that just makes it slightly clearer what's going on. (I was expecting `core-sop` to provide a specialized function for this, but it's not so surprising it's not there.)

ouroboros-consensus/src/Ouroboros/Consensus/Util/Orphans.hs
```diff
@@ -68,6 +70,26 @@ instance Serialise (VerKeyDSIGN MockDSIGN) where
...
+      timeOfDayPico = diffTimeToPicoseconds timeOfDay
+
+instance FromCBOR UTCTime where
+  fromCBOR = do
+    year <- fromCBOR @Integer
+    dayOfYear <- fromCBOR @Int
+    timeOfDayPico <- fromCBOR @Integer
+    return (UTCTime (fromOrdinalDate year dayOfYear) (picosecondsToDiffTime timeOfDayPico))
+
+
```
nfrisby, line 92: Please remove the extra blank line.

ouroboros-consensus/src/Ouroboros/Consensus/Util/Orphans.hs
```diff
@@ -68,6 +70,26 @@ instance Serialise (VerKeyDSIGN MockDSIGN) where
   encode = encodeVerKeyDSIGN
   decode = decodeVerKeyDSIGN
 
+-- TODO put this next to From/ToCBOR class
```
nfrisby, line 73: I opened input-output-hk/cardano-base#208

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -26,12 +27,13 @@ module Ouroboros.Consensus.Cardano.CanHardFork (
   , translateChainDepStateAcrossShelley
   ) where
 
+
```
nfrisby, line 30: Please remove this extra blank line.

ouroboros-consensus/src/Ouroboros/Consensus/Util/Orphans.hs
```diff
@@ -38,6 +37,9 @@ import           Ouroboros.Network.AnchoredFragment (AnchoredFragment)
 import qualified Ouroboros.Network.AnchoredFragment as AF
 import           Ouroboros.Network.MockChain.Chain (Chain (..))
 
+import           Cardano.Binary
```
nfrisby, line 40: This is another place where the new imports are not grouped according to https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md

(Thank you for your patience 😬.)

nfrisby (changes requested):
I just posted a bunch of comments. They range the gamut from minor, to stylistic, to loose suggestions, to one that questions a relatively core choice.

Thanks again for your patience. I hope that I'll be able to respond more rapidly now that I've combed through the whole thing once.

DavidEichmann (commented):
I've fixed some issues locally but am still working on the larger comments on this PR.

ouroboros-consensus/src/Ouroboros/Consensus/Util/Counting.hs
nfrisby: Just FYI, I don't have a preference between `Proxy` and `K ()`. I personally think of both as https://hackage.haskell.org/package/base-4.14.1.0/docs/GHC-Generics.html#t:U1 FWIW, but that's probably a b/c of my personal chronology.

ouroboros-consensus/src/Ouroboros/Consensus/Util/Counting.hs
nfrisby: Coincidence that you named it `Top` --- there is https://hackage.haskell.org/package/sop-core-0.5.0.1/docs/Data-SOP-Constraint.html#t:Top you can use with `hcfoldMap` if you'd like.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
DavidEichmann: I'm a bit unsure what you are suggesting @nfrisby. It seems to me that the essence of the HFC is to abstract over a type level list of eras (here `xs`). I utilize this in all the queries I plan on adding, either in the return type or in the query parameters. For example, when querying the slot length, the query still requires a `EraIndex xs` even though the resulting `SlotLength` type is era agnostic (you mentioned this in your comment). Are you suggesting we can change the queries to *not* be parameterized over `xs`? I suppose we could use `Int` instead of `EraIndex xs`, but we'd lose some type safety and I don't see what we gain from that. In the case of this PR, the query result must be parameterized over the eras, else what would the query return? So it seems to me we need to parameterize of `xs` and that is the essence of the HFC, so it makes sense that these queries are associated with the HFC i.e. placed in `Query (HardForkBlock xs)`. Is there something I'm missing? Do you have a more concrete example of what you're thinking of?

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
DavidEichmann: Yes, good catch! I'll revert this change.

ouroboros-consensus/src/Ouroboros/Consensus/Util/Counting.hs
DavidEichmann: Ahhh interesting. Another SOP nugget of knowledge there. I'll make use of that here.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
nfrisby: > the essence of the HFC is to abstract over a type level list of eras (here xs). I utilize this in all the queries I plan on adding

But the result type of `HardForkLedgerConfig xs` is the same as `LedgerConfig blk`, under the assumption (true eg on mainnet) that `blk ~ HardForkBlock xs`, isn't' it?

> In the case of this PR, the query result must be parameterized over the eras, else what would the query return?

I think the "give me the ledger config" query could instead inhabit the type `Query blk (LedgerConfig blk)` if we were to reorganize things a bit. That query's core semantics make sense for any `blk` type, regardless of whether it's headed by `HardForkBlock`.

Or am I missing something?

I chatted with Edsko and he reminded me that Duncan and he were basically suggesting that every block type going forward should be an application of `HardForkBlock`. If we really commit to that , then my observation here is rather moot. Except... note the comment on `QueryHardFork`:

https://github.com/input-output-hk/ouroboros-network/blob/5916d85bd37246fbbea3f080559f00d015f33071/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs#L106-L113

Thus a chain running a block type `blk ~ HardForkBlock [TheOneBlock]` will not be able to use the `GetLedgerConfig` query that you're introducing in this PR. It seems to me like such a chain should be able to (see all the above reasoning).

Does that all clarify my concern here?

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
DavidEichmann: I made a WIP PR in the mean time. https://github.com/input-output-hk/cardano-ledger-specs/pull/2202

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
DavidEichmann: Thanks for your response, @nfrisby. This is making more sense to me now. I think you are suggesting a top level query type like this:

```haskell
data TopQuery blk result where
  QueryLedgerConfig :: TopQuery blk (LedgerConfg blk)
  InnerQuery :: Query blk result -> TopQuery blk result
```

Another interpretation of `TopQuery` is that we're specifying a set of common queries that all block types support. Then we'd have to change the `answerQuery` function in the `QueryLedger` class to use `TopQuery`:

```haskell
class (ShowQuery (Query blk), SameDepIndex (Query blk)) => QueryLedger blk where
  answerQuery :: ExtLedgerCfg blk -> Query blk result -> ExtLedgerState blk -> result
```

nfrisby: Yes, that's certainly the spirit of what I had in mind. Would you discuss this path -- specifically, whether it seems worthwhile -- with Duncan? (I'm guessing you're chatting with some regularity?) In particular, my observation about the constraint on `QueryHardFork` seems more important than I first had guessed.

nfrisby:
For my edification: how did this accidental inversion cause the test to hang?

> Update: my version check condition was inverted 😞 Fixed

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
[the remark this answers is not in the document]
nfrisby: > Do we want move the instances to their respective places before or after merging?

This question still applies to our orphans, even though we now avoid JSON, right?

My general preference would be to define them in the right place the first time around, and that would also involve approval on the instances themselves from the Ledger team. But the Ledger Team has their hands full with Alonzo. So it's a timeline question for you/them, I think.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
nfrisby: FYI, Edsko pointed out that this module exists, which lifts the "is multiple eras" constraint.

https://github.com/input-output-hk/ouroboros-network/blob/4404be28dc53948eac7f47f116d2b01c128016b1/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Compat.hs

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Orphans.hs
nfrisby: I see that it's merged upstream; great!

nfrisby:
@DavidEichmann How's it going? Anything we can do to help?

DavidEichmann:
@nfrisby of the 4 remaining PR discussions, 2 are related to serialization instances which are currently merged upstream, but require a version bump on the `ouroboros-network` side (https://github.com/input-output-hk/ouroboros-network/issues/3025). I'm also planning on discussing the other 2 issues with @dcoutts, though he seems particularly busy at the moment.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
dcoutts: Thoughts:
* Yes we do expect all uses of consensus to use the HFC, even if it's for a single era. If (as I assume it does) simplify things, then that's all to the better.
* The `IsNonEmpty xs => (x ': xs)` constraint is something I have never _really_ understood. My limited understanding is that it helped or helps with enforcing and testing that standalone eras and singleton-HFC-wrapped eras are equivalent. This was practically useful when we had the single Byron era actually in use. In practice now we don't need that, but do we still need it to help with the simplicity of test properties? I do not know. If we don't need the constraint for tests I'd be happy to drop it.
* Some of the queries that David is adding just depend on the block type (where for the HFC that is ~ [...] to some list of eras), but some also fit the pattern of the other HFC-query classes, of per-era, rather than era-independent. So given that, it's not clear to me that there's a big advantage to be had by pulling the era-independent ones out to a new top level wrapper.

Does that help? We can have a more detailed chat if needed.

DavidEichmann: Ok. So what I'm getting out of this is that really, instead of creating a new "top level query" type, `TopQuery blk result`,  we accept that HFC is hear to stay and hence such a top level query type already exists: `data instance Query (HardForkBlock xs) result`. We should perhaps move the `GetLedgerConfig` constructor from `data QueryHardFork` to `data instance Query (HardForkBlock xs)`:

```haskell
data instance Query (HardForkBlock xs) :: Type -> Type where
  -- ...
  GetLedgerConfig :: Query (HardForkBlock xs) (LedgerConfg (HardForkBlock xs))
```
In the case that `xs ~ [x]` (i.e. is a single era), and you just want `LedgerConfig x`, it's simply a matter of unwrapping it from the `HardForkLedgerConfig` and possibly converting from a `PartialLedgerConfig x` via `completeLedgerConfig`. Does this make sense @dcoutts @nfrisby ?

DavidEichmann:
@nfrisby 

> For my edification: how did this accidental inversion cause the test to hang?

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger/Query.hs
nfrisby: Yes, thanks both; that all makes sense.

I'll study a bit and also ask @edsko tomorrow for his thoughts on lifting the restriction. (Suppose he argues we should keep it; in that case, I imagine his objections would also apply to putting `GetLedgerConfig` in `Query` instead of in `QueryHardFork`, since that would also spoil the isomorphism.)

FWIW, it looks like `cardano-node` does indeed use the `CompatQuery` interface (which lifts the restriction). However, I don't yet know _why_ they do that. This is just another thread for one of us to eventually pull on.

```
~/cardano-node$ git grep -l -e CompatQuery origin/master
origin/master:cardano-node/src/Cardano/Node/Query.hs
```

dcoutts: Having cogitated on this a bit more, I'm more sympathetic to the general top level query idea from @nfrisby. I'd sumarrise the idea as saying that these are queries that consensus can _always_ support, for any protocol/ledger, irrespective of whether that protocol is composite (HFC) or not.

For the `QueryHardFork` queries, existing, or the new one, or other future ones:
* `GetInterpreter` this is indeed HFC-specific, the notion of the interpreter only makes sense for composite protocols
* `GetCurrentEra` this is clearly HFC-specific
* `GetLedgerConfig` this is indeed not HFC-specific. It would make sense for any block type, composite or not.
* A future "get initial ledger state". This also need not be HFC specific. And possibly the type would be simpler if it were not HFC specific. For the HFC case it'd return a ledger state in the union type, rather than specifically in the type of the first era.

So on reflection it's actually quite a reasonable idea to say we should hoist these to the top. In practice I'm not quite sure how best to do that. The `Query` type is a data family of course, so we don't currently have a convenient place to stick in some extra constructors. I suppose it'd need a distinction between the top level query type, with constructors for the generic queries, and the protocol-specific queries, and the data family for the types of those protocol-specific queries.

nfrisby: > I suppose it'd need a distinction between the top level query type, with constructors for the generic queries, and the protocol-specific queries, and the data family for the types of those protocol-specific queries.

That'd be my first choice, I think. We might have to do a bit of a dance regarding backward compatible serializers for that route. But it seems the cleanest in the end.

A looser alternative is to require that any `Query` instance satisfies a type class that ensures the general queries can be constructed. This would probably lead to each ledger having to reimplement the generic queries redundantly; but the generic queries are likely not very complicated, so maybe this is a reasonable at-least-temporary approach (which will probably have an easier time satisfying the backwards compatibility).

nfrisby: > `GetInterpreter` this is indeed HFC-specific,

This is just a reactionary thought at the moment: we could return a degenerate `Interpreter` for a non-HFC block, couldn't we? In particular, the `HardForkIndices blk :: [*]` type family is applicable to any block type. EG we already have `type instance HardForkIndices ByronBlock = '[ByronBlock]`

nfrisby: Update: We had a call. We're going to assess how disruptive it would be to introduce the new top-level query type that contains the black-independent queries. EG Several of the planned queries would fit naturally there, so it seems worth it. Unless we find unexpected prereqs that we don't want to pay down now.

PR #2975, opened 2021-03-05
WIP Add query for security parameter

DavidEichmann:
See tracking issue for context: https://github.com/input-output-hk/cardano-node/issues/2454

Depends on #2973

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
+++ b/ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
@@ -39,6 +39,8 @@ data NodeToClientVersion
     -- ^ enabled @CardanoNodeToClientVersion6@, adding a query
     | NodeToClientV_8
     -- ^ 'LocalStateQuery' protocol codec change, allows to acquire tip point.
+    | NodeToClientV_9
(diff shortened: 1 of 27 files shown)


ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
```diff
@@ -39,6 +39,8 @@ data NodeToClientVersion
     -- ^ enabled @CardanoNodeToClientVersion6@, adding a query
     | NodeToClientV_8
     -- ^ 'LocalStateQuery' protocol codec change, allows to acquire tip point.
+    | NodeToClientV_9
+    -- ^ enabled @CardanoNodeToClientVersion7@, adding a query
```
coot, line 43: Please rebase on top of current `master`; I expect that the `ouroboros-network:cddl` test to fail (if not that's a bug on our side).  You'll need to add next version to the [handshake-node-to-client cddl spec](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/test-cddl/specs/handshake-node-to-client.cddl) which is 
```
32777 = 9 `xor` 0x8000
```

coot (changes requested):

nfrisby:
Closing as stale.

PR #2979, opened 2021-03-09
Trace block propagation delay

karknu:
Trace the delay from when a block should have been forged to when we're
ready to adopt it.
The delay is not exact since it depends on the correctness of our's and
the block forges clock.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
+++ b/ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
@@ -20,15 +20,17 @@ module Ouroboros.Network.BlockFetch.Client (
     BlockFetchProtocolFailure,
   ) where
 
-import           Control.Monad (unless)
+import           Control.Monad (join, unless)
 import           Control.Monad.Class.MonadSTM.Strict
 import           Control.Monad.Class.MonadThrow
+import           Control.Monad.Class.MonadTime
 import           Control.Exception (assert)
 
 import qualified Data.Set as Set
 
 import           Control.Tracer (traceWith)
 
+import           Cardano.Slotting.Slot (WithOrigin (..))
 import           Ouroboros.Network.Block
 
 import           Ouroboros.Network.Mux (ControlMessageSTM)
@@ -77,7 +79,7 @@ type BlockFetchClient header block m a =
 -- work in conjunction with our fetch logic.
 --
 blockFetchClient :: forall header block m.
-                    (MonadSTM m, MonadThrow m,
+                    (MonadSTM m, MonadThrow m, MonadTime m,
                      HasHeader header, HasHeader block,
                      HeaderHash header ~ HeaderHash block)
                  => NodeToNodeVersion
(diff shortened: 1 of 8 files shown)


ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
```diff
@@ -290,6 +294,14 @@ blockFetchClient _version controlMessageSTM
             -- Add the block to the chain DB, notifying of any new chains.
             addFetchedBlock (castPoint (blockPoint header)) block
 
+            let slotNo = case pointSlot $ blockPoint header  of
+                               Origin -> SlotNo 0 -- We don't care
+                               (At s) -> s
+            forgeTime_m <- atomically $ slotToTime slotNo
```
coot, line 300: I'd prefer to use`slotNo_m  = Maybe SlotNo` rather than `SlotNo 0 -- We don't care`, then
`join <$> traverse (atomically . slotToTime) slotNo_m` or something similar.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
```diff
@@ -290,6 +294,14 @@ blockFetchClient _version controlMessageSTM
...
+            let blockDelay_m = case forgeTime_m of
+                                    Nothing -> Nothing
+                                    Just forgeTime -> Just (diffUTCTime now forgeTime)
```
coot, line 303:
```suggestion
            let blockDelay_m = diffUTCTime now <$> forgeTime_m
```

njd42:
No particular comments - LGTM

coot (commented):
LGTM

nfrisby:
@karknu This diff LGTM. However:

  * the PR targets my temporary branch, not master.
  * we very recently merged PR #2988, which introduced a different interface than the one a made available on the temporary branch -- that PR handled a couple of the fixups you did here (eg `.Examples` module etc) and also did not return a `Maybe` in the key function.
  
In light of that, do you still want to merge this PR?

karknu:
> @karknu This diff LGTM. However:
> 
>     * the PR targets my temporary branch, not master.
> 
>     * we very recently merged PR #2988, which introduced a different interface than the one a made available on the temporary branch -- that PR handled a couple of the fixups you did here (eg `.Examples` module etc) and also did not return a `Maybe` in the key function.
> 
> 
> In light of that, do you still want to merge this PR?

No, I will make a new version of this branch based on `master`.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
```diff
@@ -290,6 +294,12 @@ blockFetchClient _version controlMessageSTM
             -- Add the block to the chain DB, notifying of any new chains.
             addFetchedBlock (castPoint (blockPoint header)) block
 
+            let slotNo_m = case pointSlot $ blockPoint header  of
+                               Origin -> Nothing
+                               (At s) -> Just s
+            forgeTime_m <- join <$> traverse (atomically . slotToTime) slotNo_m
```
coot, line 300: [`WithOrigin`](https://github.com/input-output-hk/cardano-base/blob/master/slotting/src/Cardano/Slotting/Slot.hs#L56) has a `Traversable` instance, there's no need to use `Maybe`.

karknu: This is an outdated PR, `master` doesn't use `Maybe`. https://github.com/input-output-hk/ouroboros-network/blob/dea09273c87578a2047445e4434ba30df6570ac2/ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs#L297

PR #2984, opened 2021-03-10
Updating cardano-ledger

JaredCorduan:
* The ledger state now stores an PulsingRewUpdate instead of a
  RewardUpdate. The PulsingRewUpdate allows us to compute the staking
  rewards incrementally instead of all at once. This involves a new
  serialization format for the reward updates.
* Annotated decoders are now used for PParams and PParamsDelta (the
  data structures now store their own serialization). Additionally, the
  LedgerState also now uses an annotated decoder. The serialization in
  consensus has been changed to use the cbor-in-cbor encoding for these
  types (including in the queries).
* A bug was fixed in the reward calculation which was introduced in
  commit 31083fc5cecbe17f3628393072094b130058edd9.
* The MIR certificates now have additional functionality, but such
  functionality is prohibited until protocol major version 5. Binary
  compatibility is preserved for the original functionality. The new
  functionality is the ability to transfer lovelace between the reserves
  and the treasury.
* The ledger repo now uses ghc-8.10.4 and cabal-3.4.0.0.

Replaces #2971 (the branch named `jc/update-ledger-to-ba74779256` was overly confident :smile: )
Note that unlike #2971, binary compatibility of the protocol parameters is preserved (those golden tests are not touched in this PR).


ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
```diff
@@ -601,20 +602,14 @@ exampleNewEpochState value = SL.NewEpochState {
...
-                        (keyToCredential exampleStakeKey) $
-                        Set.singleton $ SL.Reward {
-                            SL.rewardType   = SL.MemberReward
-                          , SL.rewardPool   = (SL._poolId examplePoolParams)
-                          , SL.rewardAmount = SL.Coin 10
-                          }
-        , deltaF    = SL.DeltaCoin (- 3)
-        , nonMyopic = nonMyopic
-        }
+    rewardUpdate :: SL.PulsingRewUpdate (EraCrypto era)
```
nfrisby, line 605: @jared Do these values need to change?

note to self: Since they changed, is there a golden test that should also change?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -87,6 +88,11 @@ type EraCrypto era = Crypto era
 -- TODO Currently we include the constraint @SL.AdditionalGenesisConfig era ~
 -- ()@. When we fork to Alonzo we will need additional genesis config
 -- information.
+--
+-- TODO The Alonzo PParams and PParamsDelta will not have FromCBOR instances,
+-- since the protocal parameters need to be hashed in this era.
+-- We will need a plan to handle preserving the original bytestrings,
+-- sunch as using the Annotator (FromCBOR a) instances instead.
```
nfrisby, line 95: `Annotator (FromCBOR a)` -> `FromCBOR (Annotator a)`

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -484,27 +486,28 @@ encodeShelleyLedgerState
                        , shelleyLedgerTransition
                        } =
     encodeVersion serialisationFormatVersion2 $ mconcat [
-        CBOR.encodeListLen 2
```
nfrisby, line 487: This affects the NTC protocol, but only a `Debug*` constructor:

It also affects the on-disk snapshots, but those are already changing, so we're batching this in with that.

```
  -- | Only for debugging purposes, we don't guarantee binary compatibility.
  -- Moreover, it is huge.
  DebugNewEpochState
```

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -435,11 +439,11 @@ encodeShelleyResult query = case query of
     GetStakeDistribution                       -> toCBOR
     GetFilteredUTxO {}                         -> toCBOR
     GetUTxO                                    -> toCBOR
-    DebugEpochState                            -> toCBOR
+    DebugEpochState                            -> wrapCBORinCBOR toCBOR
     GetCBOR {}                                 -> encode
     GetFilteredDelegationsAndRewardAccounts {} -> toCBOR
     GetGenesisConfig                           -> toCBOR
-    DebugNewEpochState                         -> toCBOR
+    DebugNewEpochState                         -> wrapCBORinCBOR toCBOR
```
nfrisby, line 446: Duncan on Slack explained that the cost of introducing CBOR-in-CBOR for `Debug` is that it'll break all current clients that use the the `GetCBOR` approach. So we're all trying to figure out how to avoid that now.

The motivator for this change is that ledger does not want to expose a `Decoder s EpochState`, it only wants to expose `Decoder s (Annotator EpochState)`, which requires a CBOR-in-CBOR.

In particular, `Decoder` is not expressive enough to support a combinator that converts `Decoder s (Annotator EpochState)` (which assumes CBOR-in-CBOR) to a `Decoder s EpochState` (which does not).

This case expression is ultimately used to instantiate the Consensus Layer's `decodeResult` class, which requires a `Decoder`. If we generalized `decodeResult`, we could perhaps enable such a combinator.

JaredCorduan:
@nfrisby I've pushed the changes you and I made together today, namely using `forgetAnnotator` in order to avoid the cbor-in-cbor encoding.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -452,14 +457,18 @@ decodeShelleyResult query = case query of
...
     GetFilteredDelegationsAndRewardAccounts {} -> fromCBOR
     GetGenesisConfig                           -> fromCBOR
-    DebugNewEpochState                         -> fromCBOR
+    DebugNewEpochState                         -> forgetAnnotator fromCBOR
     DebugChainDepState                         -> fromCBOR
     GetRewardProvenance                        -> fromCBOR
+    where
+       forgetAnnotator ::
+         forall a r. (forall t. Decoder t (Annotator a)) -> Decoder r a
+       forgetAnnotator d = (\x -> runAnnotator x (Full mempty)) <$> d
```
nfrisby, line 474: This only works here because the `FromCBOR` instance for `Annotator NewEpochState` doesn't actually use the annotations. This is apparently recursively true for its fields too.

/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/LedgerState.hs#L675-L684

Here is an example of a `FromCBOR` that does use the annotation:

https://github.com/input-output-hk/cardano-ledger-specs/blob/5b6ab459841b59432400b5d0cb796c2a12344bf9/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/Tx.hs#L408

nfrisby: Ah. Shelley-based `NewEpochState` contains `EpochState` which contains `PParams` which, even for `Alonzo`, does not contain any annotations. That's why we can use our trick here. (Well, because _none_ of the fields in `NewEpochState` carry annotations. EG There are no `MemoBytes` anywhere in there.)

The annotations are only present in `LangDepView`, which is created by calling `getLanguageView` which ultimately (via `To`) depends on a `ToCBOR` instance.

nfrisby: Which is a long-winded way of emphasizing that `forgetAnnotator` relies on special assumptions: that the `FromCBOR (Annotator NewEpochState)` instance is rather degenerate: the `Annotator` layer is (at least currently) completely unnecessary.

JaredCorduan: gotcha! I hope to add back the relevant `FromCBOR` instances soon and then come clean this back up. Assuming we all agree that it is okay to have the node provide the serialization of the protocol parameters. Otherwise we will have to handle the annotation correctly. Thanks again for the help!

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -493,18 +494,19 @@ encodeShelleyLedgerState
...
     ]
   where
-    decodeShelleyLedgerState2 :: Decoder s' (LedgerState (ShelleyBlock era))
+    decodeShelleyLedgerState2 :: Decoder s' (Annotator (LedgerState (ShelleyBlock era)))
     decodeShelleyLedgerState2 = do
-      enforceSize "LedgerState ShelleyBlock" 2
-      shelleyLedgerTip        <- decodeWithOrigin decodeShelleyTip
+      enforceSize "LedgerState ShelleyBlock" 3
+      shelleyLedgerTip        <- pure <$> decodeWithOrigin decodeShelleyTip
       shelleyLedgerState      <- fromCBOR
```
nfrisby, line 507: I think if you apply the `(\x -> runAnnotator x (Full mempty))` lambda on this `fromCBOR`, then we can eliminate all of the other changes to this function. I'd like the smaller diff, especially since this is a stop-gap measure.

JaredCorduan: done! that's so much nicer!

JaredCorduan:
Hydra timed out:
```
[5 of 8] Compiling Ouroboros.Consensus.Cardano.Node ( src/Ouroboros/Consensus/Cardano/Node.hs, dist/build/Ouroboros/Consensus/Cardano/Node.o, dist/build/Ouroboros/Consensus/Cardano/Node.dyn_o )
building of '/nix/store/0kk4ishzh0awmiqk65nkgdba4v1qx09j-ouroboros-consensus-cardano-lib-ouroboros-consensus-cardano-0.1.0.0' timed out after 900 seconds of silence
```

It doesn't look like I have the permissions to re-run the tests.

nfrisby:
> building of '/nix/store/0kk4ishzh0awmiqk65nkgdba4v1qx09j-ouroboros-consensus-cardano-lib-ouroboros-consensus-cardano-0.1.0.0' timed out after 900 seconds of silence

900 seconds of silence! :grimace:

JaredCorduan:
The plan now is to wait for https://github.com/input-output-hk/cardano-ledger-specs/pull/2180 and then remove `forgetAnnotator` from this ouroboros-network PR.

nc6:
I think there's a strong reason to _not_ create these invalid `FromCBOR` instances, and rather to use the explicit `forgetAnnotator` here. Namely, I think we should merge this PR as is, because it identifies explicitly where we're doing this kind of unsafe behaviour.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
```diff
@@ -497,11 +498,12 @@ decodeShelleyLedgerState = decodeVersion [
...
     ]
   where
+    forgetAnnotator d = (\x -> runAnnotator x (Full mempty)) <$> d
     decodeShelleyLedgerState2 :: Decoder s' (LedgerState (ShelleyBlock era))
     decodeShelleyLedgerState2 = do
-      enforceSize "LedgerState ShelleyBlock" 2
+      enforceSize "LedgerState ShelleyBlock" 3
       shelleyLedgerTip        <- decodeWithOrigin decodeShelleyTip
-      shelleyLedgerState      <- fromCBOR
+      shelleyLedgerState      <- forgetAnnotator fromCBOR
```
nc6, line 506: I think this is the wrong thing to do here. This *would* explicitly break with Alonzo. The cbor-in-cbor solution is at least future proof in this case.

dcoutts (approved):
LGTM

JaredCorduan:
bors merge

PR #2993, opened 2021-03-13
Always use qualified import for Data.List

newhoggy:
This change will fix these potential warnings:

```
Error:     To ensure compatibility with future core libraries changes
    imports to Data.List should be
    either qualified or have an explicit import list.
   |
16 | import           Data.List
   |                  ^^^^^^^^^
```

The change is in anticipation of mitigating the effects of future core library changes.

Whilst it is possible to use explicit import lists, the use of qualified import by default will minimise future breakages, so that's the default option I'm taking.

Suggestions for exceptions to this rule welcome.

### DIFF
--- a/io-sim-classes/io-sim-classes.cabal
+++ b/io-sim-classes/io-sim-classes.cabal
@@ -59,7 +59,7 @@ library
                        bytestring,
                        mtl   >=2.2 && <2.3,
                        stm   >=2.5 && <2.6,
-                       time  >=1.6 && <1.11
+                       time  >=1.9.1 && <1.11
   ghc-options:         -Wall
                        -Wno-unticked-promoted-constructors
                        -Wcompat
--- a/io-sim/io-sim.cabal
+++ b/io-sim/io-sim.cabal
@@ -44,7 +44,7 @@ library
                        exceptions        >=0.10,
                        containers,
                        psqueues          >=0.2 && <0.3,
-                       time              >=1.6 && <1.11,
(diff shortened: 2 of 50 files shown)


newhoggy:
Also add lower bound to `time` to prevent these kinds of errors from popping up due to the whims of the solver:

```
src/Ouroboros/Network/Testing/ConcreteBlock.hs:52:61: error:
Error:     Module ‘Data.Time.Clock’ does not export ‘secondsToNominalDiffTime’
   |
52 | import           Data.Time.Clock (UTCTime (..), addUTCTime, secondsToNominalDiffTime)
   |                                                             ^^^^^^^^^^^^^^^^^^^^^^^^
```

nfrisby:
I like the diff and we don't like noisy warnings. Thanks, @newhoggy!

The Consensus parts look correct to me. My only hesitation is merge conflicts with existing branches (eg the WIP Alonzo integration branch). But I don't think they'll be severe. @EncodePanda What do you think, like easy enough to merge this?

So: I did not click Approve because 1) @coot @karknu etc own the Network parts, and 2) I want to give @EncodePanda a chance to object or explicitly accept to the merge conflicts.

karknu (changes requested):
Thanks, but I prefer these kinds of changes after `p2p-master` is merged.

EncodePanda:
I'm ok with fixing conflicts on my end on Alonzo branch. LGTM on my end :)

EncodePanda (approved):
LGTM

PR #2995, opened 2021-03-15
Trace block propagation delay

karknu:
Trace the delay from when a block should have been forged to when we're
ready to adopt it.
The delay is not exact since it depends on the correctness of our's and
the block forge's clock.


karknu:
#2979 version based on `master`.

njd42 (approved):
LGTM

PR #2999, opened 2021-03-16
cardano-ping: Unix socket support

karknu:


karknu:
Implements #2999 2989

network-mux/demo/cardano-ping.hs
```diff
@@ -126,14 +153,21 @@ main = do
...
     msgQueue <- newEmptyTMVarIO
 
-    when (isNothing $ host options) $ do
-        putStrLn "Specify host/ip with '-h <hostname>'"
+    when (isNothing (host options) && isNothing (unixSock options) ) $ do
+        putStrLn "Specify host/ip with '-h <hostname>' or a unix socket with -u <file name>"
         exitWith (ExitFailure 1)
 
-    addresses <- Socket.getAddrInfo (Just hints) (host options) (Just $ port options)
+    (addresses, versions) <- case unixSock options of
```
coot, line 160: The lines below seems to be over 80 lines, could you reduce the indentation.  Just two spaces is enough or break it further like this:
```
(...) <-
  case ... of
    ...
```

network-mux/demo/cardano-ping.hs
```diff
@@ -239,9 +289,14 @@ handshakeDec = do
                       return $ Left $ Refused vn msg
                   _ -> return $ Left $ UnknownTag tag
 
+         32773 -> decodeVersion
+         32774 -> decodeVersion
+         32775 -> decodeVersion
+         32776 -> decodeVersion
```
coot, line 295: Inlining these numbers is not clear, it's better to do the calculation that leads to them explicitly.

network-mux/demo/cardano-ping.hs
```diff
@@ -252,10 +307,14 @@ handshakeDec = do
              5 -> decodeWithMode NodeToNodeVersionV5
              6 -> decodeWithMode NodeToNodeVersionV6
              7 -> decodeWithMode NodeToNodeVersionV7
+             32773 -> Right . NodeToClientVersionV5 <$> CBOR.decodeWord32
+             32774 -> Right . NodeToClientVersionV6 <$> CBOR.decodeWord32
+             32775 -> Right . NodeToClientVersionV7 <$> CBOR.decodeWord32
+             32776 -> Right . NodeToClientVersionV8 <$> CBOR.decodeWord32
```
coot, line 313: The same applies here.

coot (changes requested):

network-mux/demo/cardano-ping.hs
karknu: I would if I could do any operations on constants in a `case` statement.

network-mux/demo/cardano-ping.hs
karknu: Same problem.

network-mux/demo/cardano-ping.hs
coot: I think one can do that this way:
```
case a of
  5 -> ...
  6 -> ...
  7 -> ...
  x | y <- f x -> 
```

network-mux/demo/cardano-ping.hs
```diff
@@ -381,13 +432,28 @@ pingClient tracer Options{quiet, magic, json, maxCount} peer = bracket
...
+
+    peerString :: IO String
+    peerString =
+        case Socket.addrFamily peer of
+             Socket.AF_UNIX -> return $ show $ Socket.addrAddress peer
+             _ -> do
+                  (Just host, Just port) <- Socket.getNameInfo [Socket.NI_NUMERICHOST
+                                                               , Socket.NI_NUMERICSERV]
+                                                  True True (Socket.addrAddress peer)
+                  return $ host ++ ":" ++ port
```
coot, line 455:
```suggestion
             _ -> do
                  (Just host, Just port) <- 
                    Socket.getNameInfo
                      [ Socket.NI_NUMERICHOST, Socket.NI_NUMERICSERV ]
                      True True (Socket.addrAddress peer)
                  return $ host ++ ":" ++ port
```

network-mux/demo/cardano-ping.hs
```diff
@@ -160,13 +194,17 @@ main = do
...
+                       | NodeToClientVersionV6 Word32
+                       | NodeToClientVersionV7 Word32
+                       | NodeToClientVersionV8 Word32
+                       | NodeToNodeVersionV1   Word32
+                       | NodeToNodeVersionV2   Word32
+                       | NodeToNodeVersionV3   Word32
+                       | NodeToNodeVersionV4   Word32 Bool
+                       | NodeToNodeVersionV5   Word32 Bool
+                       | NodeToNodeVersionV6   Word32 Bool
+                       | NodeToNodeVersionV7   Word32 Bool
```
coot, line 207: It is a bit strange to mix both versions together.  Is this because `cardano-ping` accepts a single option for address which might be either a network address of local socket?

coot: There's a distinction between `node-to-node` and `node-to-client` versions with `supportedNodeToNodeVersions` and `supportedNodeToClientVersions` which introduces the split between `NodeVersion`s.  I would suggest to do that differently:

* make the `pingClient` polymorphic over versions it accepts
* have two different types for `node-to-client` and `node-to-node` versions; if there are any operations needed on versions they could be provided with a type class.

karknu: That seems like an unnecessary abstraction.

coot (approved):

PR #3003, opened 2021-03-17
Add point to TraceBlockFetchServerSendBlock

karknu:
This will be used when implementing a new metric that gauges a node's importance to its downstream peers.


ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/Server.hs
```diff
@@ -132,5 +132,5 @@ blockFetchServer tracer chainDB _version registry = senderSide
 -- | Events traced by the Block Fetch Server.
 data TraceBlockFetchServerEvent blk =
     -- | The server sent a block to the peer.
-    TraceBlockFetchServerSendBlock
+    TraceBlockFetchServerSendBlock (Point blk)
```
coot, line 135:
```suggestion
    TraceBlockFetchServerSendBlock !(Point blk)
```

coot (approved):
LGTM

Issue #2969, opened 2021-03-04
Configure stylish-haskell to format module header


EncodePanda:
After https://github.com/input-output-hk/ouroboros-network/pull/2968 we can configure stylish-haskell to format module header. One less thing to worry about.

PR #3009, opened 2021-03-22
stylish-haskell formatting the module header

EncodePanda:
Fixes https://github.com/input-output-hk/ouroboros-network/issues/2969

Not all files were following the Style Guide. The first commit fixes that. Export lists sorted alphabetically in their own respective groups, opening bracket on the same line as the module name, each export on its own new line.

The second commit adds the automatic formating of module headers.
Please note that module_header is a feature on stylish-haskell
available since 0.12.1.0 BUT it is missing a crucial configuration
aspect: open_bracket, which was recently added to stylish-haskell but
not yet released (it should be shipped with 0.12.3.0 release).
Until 0.12.3.0 is not released we must use the custom version.


EncodePanda:
@nfrisby if we get this PR in we might consider updating Style Guide to say "stylish-haskell will format module header, including export list"

ouroboros-consensus-byronspec/src/Ouroboros/Consensus/ByronSpec/Ledger/Block.hs
```diff
@@ -5,12 +5,13 @@
 {-# LANGUAGE TypeFamilies      #-}
 
 module Ouroboros.Consensus.ByronSpec.Ledger.Block (
```
nfrisby, line 7: The way it moved the comment is a little confusing. Is there a clever way to resolve that?

This example of "type alias" is not crucial, but IIRC I'll see some below that are little more useful (eg I'm anticipating "opaque").

Worst-case: we can promote the important such comments from comments to relevant Haddock sections.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs
```diff
@@ -16,90 +16,37 @@ module Ouroboros.Consensus.Cardano.Block (
     -- checks from GHC. But GHC expects a data type, not a type family, that's
     -- why we sometimes mention the data type of the instance in these exports
     -- instead of the abstract type family.
-  , HardForkBlock (
```
nfrisby, line 19: @EncodePanda Is there an option for preserving the way this export-list was? It's not a dealbreaker, but I prefer the way it was.

I'm of two minds about alphabetizing these. For the most part, I would expect such lists to be alphabetized. But the era names are the one example where the non-alphabetical ordering is intentional. (Though I don't know if that's ultimately the wisest choice.)

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs
```diff
@@ -16,90 +16,37 @@ module Ouroboros.Consensus.Cardano.Block (
...
-        BlockByron
-      , BlockShelley
-      , BlockAllegra
-      , BlockMary
-      )
+  , HardForkBlock (BlockByron, BlockShelley, BlockAllegra, BlockMary)
     -- * Headers
   , CardanoHeader
-  , Header (
```
nfrisby, line 27: I'm of two minds about alphabetizing these. For the most part, I would expect such lists to be alphabetized. But the era names are the one example where the non-alphabetical ordering is intentional. (Though I don't know if that's ultimately the wisest choice.)

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -6,13 +6,13 @@
 {-# LANGUAGE UndecidableSuperClasses #-}
 module Ouroboros.Consensus.Shelley.Eras (
     -- * Eras based on the Shelley ledger
-    ShelleyEra
```
nfrisby, line 9: Ah, OK. Here's an example where it was intentionally out-of-order. What are our options, @EncodePanda? For example are there any easy options for special-casing this rule for the era names? They have a meaningful non-alphabetical chronological order which is what we have been using throughout.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
```diff
@@ -9,12 +9,13 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 module Ouroboros.Consensus.Shelley.Ledger.Config (
     BlockConfig (..)
-  , mkShelleyBlockConfig
   , CodecConfig (..)
+  , CompactGenesis
   , StorageConfig (..)
-  , CompactGenesis -- opaque
```
nfrisby, line 15: And here's one of the `-- opaque` comments I had expected.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs
EncodePanda: This is unfortunately not supported by stylish. It should be rather simple to add though. I've created a ticket for it  https://github.com/jaspervdj/stylish-haskell/issues/358

ouroboros-consensus-byronspec/src/Ouroboros/Consensus/ByronSpec/Ledger/Block.hs
EncodePanda: Yeah, this is definitely a bug. I've [created an issue in stylish-haskell](https://github.com/jaspervdj/stylish-haskell/issues/359). 
For now, do you want time to go through all occurrences of this issue and replace comments with haddock comments? I'm not really looking forward to this :) but then again.. I gave you a 350+ PR to review... :)

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
EncodePanda: I'm afraid there is no way to work around this at the moment. Possible options:

1. discard this file from being formatted completely OR format this file with a modified configuration
This would work but the approach does not scale. We would have to explicitly exclude files or keep a list of which files we want to format with a specific set of rules

2. turn off the `sort` option in the module header.
This means that the `module_header` step would stop completely sorting the export lists.  It would only check that: each export is in a separate line, open bracket starts on the same line as module name, indentation is correct (according to Style Guide).

3. Don't bother that the exported eras are not in order in which the `Cardano` combines them 
I honestly wouldn't mind. If anything I prefer consistency.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Config.hs
EncodePanda: Yeah, this is not great. As you mentioned, we could change to haddock and see how that goes.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs
EncodePanda: Responded here https://github.com/input-output-hk/ouroboros-network/pull/3009#discussion_r599467748

ouroboros-consensus-test/src/Test/Util/FS/Sim/MockFS.hs
```diff
@@ -48,16 +44,15 @@ module Test.Util.FS.Sim.MockFS (
...
-    -- opaque
-  , OpenHandleState
-    -- opaque
-  , ClosedHandleState
-    -- opaque
-  , FilePtr
-    -- opaque
   , Files
   , mockFiles
+    -- * opaque
```
nfrisby, line 49: Let's use `**` since these were all in the "Exported for the benefit of tests only" section

nfrisby: Actually, let's just remove `-- opaque` from the ones that are already in the testing section. We don't need the extra info about those, I don't think.

So just `MockFS` and `HandleMock` in the next section, I think.

ouroboros-consensus/src/Ouroboros/Consensus/BlockchainTime/WallClock/Types.hs
```diff
@@ -16,7 +16,6 @@ module Ouroboros.Consensus.BlockchainTime.WallClock.Types (
   , SystemTime (..)
     -- * Slot length
   , SlotLength
-    -- Opaque
```
nfrisby: You can put this on the `-- * Slot length` line, like `-- * Slot length (opaque)`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/Index/Primary.hs
```diff
@@ -17,7 +17,6 @@ module Ouroboros.Consensus.Storage.ImmutableDB.Impl.Index.Primary (
     SecondaryOffset
     -- * PrimaryIndex
   , PrimaryIndex (..)
-    -- Exported for the benefit of tests
```
nfrisby: I now see the

> For those types opaque comment was removed completely.

sentence in your commit message.

We should retain the `opaque` or `for testing` annotations in some way; I don't want to just drop them.

As you noticed -- I'm catching on now -- there are pretty often pre-existing sections where we could just relocate the `opaque` comment up into the section header. So: `-- * PrimaryIndex (opaque)`.

nfrisby (changes requested):
I'm Requesting Changes b/c of this request

https://github.com/input-output-hk/ouroboros-network/pull/3009/commits/b513a5720b39a907bba2fc64bee870eeb0baee36#r602980717

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/Index/Primary.hs
EncodePanda: I've addressed all opaque types in latest commit (https://github.com/input-output-hk/ouroboros-network/pull/3009/commits/e0d2510add8fc7a5c6f578fa38ac5f1e93bd47fc)

EncodePanda:
I've liked the 

```
-- ** opaque
```

idea if the type was already in its own group

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/BlockCache.hs
```diff
@@ -5,12 +5,13 @@
 --
 -- > import           Ouroboros.Consensus.Storage.ChainDB.Impl.BlockCache (BlockCache)
 -- > import qualified Ouroboros.Consensus.Storage.ChainDB.Impl.BlockCache as BlockCache
-module Ouroboros.Consensus.Storage.ChainDB.Impl.BlockCache
-  ( BlockCache -- opaque
-  , empty
-  , singleton
+module Ouroboros.Consensus.Storage.ChainDB.Impl.BlockCache (
+    BlockCache
+    -- opaque
```
nfrisby, line 10: We missed one here.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Chunks/Layout.hs
```diff
@@ -13,33 +13,34 @@
 -- relevant internally in the immutable DB.
 module Ouroboros.Consensus.Storage.ImmutableDB.Chunks.Layout (
     -- * Relative slots
-    RelativeSlot -- Opaque
-  , maxRelativeSlot
-  , relativeSlotIsEBB
-  , nthBlockOrEBB
+    RelativeSlot
+    -- Opaque
```
nfrisby, line 17: Another .... I'm guessing I didn't use case-insensitivity in my grep search.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -17,43 +17,44 @@
...
-   , ledgerDbMaxRollback
-   , ledgerDbIsSaturated
-     -- ** Pure API
-   , ledgerDbPush'
-   , ledgerDbPushMany'
-   , ledgerDbSwitch'
-   ) where
+    -- * LedgerDB proper
+    LedgerDB
+    -- opaque
```
nfrisby, line 22: Missing one here too

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl/FileInfo.hs
```diff
@@ -9,16 +9,17 @@
 --
 -- Intended for qualified import.
 module Ouroboros.Consensus.Storage.VolatileDB.Impl.FileInfo (
-    FileInfo      -- opaque
+    FileInfo
+    -- opaque
```
nfrisby, line 13: Missed one

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl/Index.hs
```diff
@@ -6,14 +6,15 @@
 -- Intended for qualified import
 -- > import qualified Ouroboros.Consensus.Storage.VolatileDB.Impl.Index as Index
 module Ouroboros.Consensus.Storage.VolatileDB.Impl.Index (
-    Index -- opaque
-  , empty
-  , lookup
-  , insert
+    Index
+    -- opaque
```
nfrisby, line 10: Missed one

ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl/Types.hs
```diff
@@ -3,22 +3,23 @@
 {-# LANGUAGE StandaloneDeriving #-}
 module Ouroboros.Consensus.Storage.VolatileDB.Impl.Types (
     -- * Blocks per file
-    BlocksPerFile -- opaque
-  , unBlocksPerFile
+    BlocksPerFile
+    -- opaque
```
nfrisby, line 7: Another

ouroboros-consensus/src/Ouroboros/Consensus/Util/CallStack.hs
```diff
@@ -4,7 +4,8 @@
 --
 -- Use of this module is intended to /replace/ import of @GHC.Stack@
 module Ouroboros.Consensus.Util.CallStack (
-    PrettyCallStack -- Opaque
+    PrettyCallStack
+    -- Opaque
```
nfrisby, line 8: Missed one

ouroboros-consensus/src/Ouroboros/Consensus/Util/EarlyExit.hs
```diff
@@ -10,10 +10,11 @@
 {-# LANGUAGE UndecidableInstances       #-}
 
 module Ouroboros.Consensus.Util.EarlyExit (
-    WithEarlyExit -- opaque
+    WithEarlyExit
+    -- opaque
```
nfrisby, line 14: Missed one

ouroboros-consensus/src/Ouroboros/Consensus/Util/MonadSTM/StrictMVar.hs
```diff
@@ -6,25 +6,26 @@
 {-# LANGUAGE TypeFamilies      #-}
 
 module Ouroboros.Consensus.Util.MonadSTM.StrictMVar (
-    StrictMVar(..) -- constructors exported for benefit of tests
+    StrictMVar (..)
+    -- constructors exported for benefit of tests
```
nfrisby, line 10: Missed another

nfrisby (commented):
GH forgot that I Viewed all the modules. I went through them again and caught a handful of "-- opaque" etc that we didn't yet fixup.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
nfrisby: We decided to sort them alphabetically, not chronologically. We think the extra automation is worth it, and moreover is consistent with how eg imports will be sorted.

If `stylish-haskell` were to offer an extensible ordering that the config could customize, then we might consider reintroducing the historical sequence of era names as part of the ordering.

nfrisby (approved):
We did it! Overcoming adversity! Epic file counts! This is one we'll remember :P

I think the fact that so many of the export lists were not sorted is in indication that -- even more so than normal terms/declarations -- this is a relatively neglected area: we programmers don't concern ourselves with it too much. That means automation is a welcome simplification here. And a dev that objects to the automation in a particular module can override it by introducing Haddock sections, which is a Better Practice anyway. 👍

A few commits could be squashed, but nothing is egregious. Approved.

nfrisby:
Explaining the above Approval and de-Approval: I dismissed my review. I had lost track of the TODOs in the GitHub UI.

ouroboros-consensus-byronspec/src/Ouroboros/Consensus/ByronSpec/Ledger/Block.hs
nfrisby: We rephrased these annotations via Haddock sections.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs
nfrisby: We're biting this bullet. Not a big deal.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Block.hs
nfrisby: We opened upstream jaspervdj/stylish-haskell#360

EncodePanda:
@nfrisby ok, this is now ready :)

nfrisby (approved):
:clap: Great stuff @EncodePanda! I think this is a good automation sweet spot that many devs will enjoy 👍

PR #3010, opened 2021-03-23
Documented GovernorMockEnvironment Generator

bolt12:
Added a check that local and public root peers are disjoint. Also tried very hard to extend some more governor properties with my new found insights from modelling the Governor with Alloy, and all that I found was this extra harmless check.

Trace properties already check for the validity of the `PeerSelectionStatee`, so we can be confident that at least the state is not corrupted after a certain amount of time. I also thought about adding consistency checks for each possible action (i.e. checking if there's no starvation of a particular action), but I am not sure if the `livelock` check covers that.


bolt12:
Seems like I found a slight bug on the generator, hopefully it is fixed now

bolt12:
Just for reference, the thing I "fixed" was actually intended to be that way, so I'll revert and document the reason for it!

PR #3014, opened 2021-03-24
Increase the backlog for local clients

karknu:
Increase the backlog of clients for unix sockets to 8. This match
the size of TCP sockets backlog.


coot (approved):
LGTM

Issue #3015, opened 2021-03-25
cardano-ping should not fail when trying to connect to ipv6 on machine without ipv6 if host has ipv4 address.


jbgi:
Given a relay node with a dns name that has both `A` and `AAAA` (ipv6) entries, `cardano-ping` will fail (error code 1) when trying to ping that relay from a machine without ipv6 connectivity.

Eg.:
```
$ dig -t A can1.ddnsgeek.com

;; ANSWER SECTION:
can1.ddnsgeek.com.      112     IN      A       174.112.109.149

$ dig -t AAAA can1.ddnsgeek.com

;; ANSWER SECTION:
can1.ddnsgeek.com.      120     IN      AAAA    2607:fea8:e3c0:7160:e86d:9b2f:1109:f81f
```

```
$ cardano-ping -h can1.ddnsgeek.com -p 6002 -c 1 --json
174.112.109.149:6002 TCP rtt: 0.103
174.112.109.149:6002 handshake rtt: 0.107056831s
174.112.109.149:6002 Negotiated version NodeToNodeVersionV5 764824073 False
{ "pongs": [ {"mean":0.1146601,"cookie":0,"sample":0.1146601,"median":0.1146601,"min":0.1146601,"max":0.1146601,"p90":0.1146601,"timestamp":"2021-03-25T13:22:45.490890509Z","host":"174.112.109.149:6002"}cardano-ping: Network.Socket.connec
t: <socket: 12>: does not exist (Network is unreachable)

$ echo $?
1
```

Expected: `cardano-ping` should just log a warning and proceed using ipv4 instead of ipv6, as `cardano-node` does:
```
[cardano.node.DnsSubscription:Warning:57] [2021-03-25 13:03:06.01 UTC] Domain: "can1.ddnsgeek.com" Unsupported remote target address [2607:fea8:e3c0:7160:e86d:9b2f:1109:f81f]:6002
[cardano.node.DnsSubscription:Notice:65] [2021-03-25 13:03:06.03 UTC] Domain: "can1.ddnsgeek.com" Connection Attempt Start, destination 174.112.109.149:6002
[cardano.node.DnsSubscription:Notice:65] [2021-03-25 13:03:06.13 UTC] Domain: "can1.ddnsgeek.com" Connection Attempt End, destination 174.112.109.149:6002 outcome: ConnectSuccessLast
```

karknu:
Fixed in #3016.

PR #3016, opened 2021-03-25
Catch and print exceptions from pingClient

karknu:
Catch and print errors from pingClient. If at least one pingClient
succeeded cardano-ping will exit with Success.


karknu:
Fixes #3015

jbgi (approved):

PR #3019, opened 2021-03-26
Add the eval-time GC roots to release.nix so they're cached.

jbgi:
Otherwise they'll get GCd very quickly, which means people will need to
 build a lot to even open the shell.


bolt12:
we use an updated nix config based on flakes, perhaps this can be closed

PR #3024, opened 2021-03-30
Added transition documentation to Connection Manager API

bolt12:


bolt12:
The implementation LGTM wrt to the specification. At least I tried to check all the transitions and they seemed fined. @coot says that the tests do not show any problems so far as well.

The Server and IPG modules could be structured in a more organized way. As it is now (`Server`  and `Server2`) it is not really intuitive where to look.

The code for the `ConnectionManager` is big and scary but after reading the documentation and understanding how the components are supposed to work together it  was easier to follow the code. Maybe it can be broken in several parts, `withConnectionManager` is over 1000 lines! I wonder if the state machine pattern matching can be refactored for each `impl` (probably not).

The tests are still in development. The only possible invariants that I could think of for the Connection Manager was for the relation between connected node to be injective (i.e. 1-to-1-to-1 relation between 2 nodes and a given state) and the `ConnectionManagerState` being a Map already gives us that. The other one was that in a given interval, assuming the 2 nodes are correct (do not fail), if node `A` puts `B` in `DuplexState`, then `B` puts `A` in `DuplexState` as well. The same for the other States (`OutboundDuplex` vs `InboundDuplex`, for example). There's already a test that checks if connections are being reused, which covers the `DuplexState` case. The other cases can be included in the tests for the IPG and Server; the Connection Manager simulation has only one side (Outbound).

coot (approved):
LGTM

bolt12:
bors merge

PR #3030, opened 2021-04-02
Added PeerSelectionCounters and respective tracers.

bolt12:
I've made this changes on top of `dcoutts/p2p-local-root-peers` since `p2p-master` does not have the new `test` structure. I then realized that this branch does not have the `PeerSelection` tracers in `DiffusionTracers` so that is still left TODO. Maybe after reviewing that the current additions make sense.


ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -268,6 +268,20 @@ data PeerSelectionState peeraddr peerconn = PeerSelectionState {
      }
   deriving (Show, Functor)
 
+data PeerSelectionCounters = PeerSelectionCounters {
+      coldPeers :: !Int,
+      warmPeers :: !Int,
+      hotPeers :: !Int
```
coot, line 274:
```suggestion
      hotPeers  :: !Int
```

PR #3040, opened 2021-04-07
Upgrade haskell.nix and bind nixpkgs to what haskell.nix exposes for better cache hits

EncodePanda:
We've upgraded haskell.nix to latest revision.

Also instead of explicitly pointing to nixpkgs (in sources.json) -
which requires us to manually updated it each time we upgrade
haskell.nix in order to get cache hits - we've modified the
default.nix so that it explicitly picks haskell.nix's nixpkgs.

In other words, next time some one wants to update haskell.nix it should be a matter
of just calling 'niv update haskell.nix' and commiting changes. Script
will automatically pick the right nixpkgs.

In this commit we did not modified iohk-nix entry. It might be the
case that it should also be updated.

Co-authored-by: Javier Sagredo <javier.sagredo@iohk.io>

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -7,22 +7,14 @@ let
   sources = import ./sources.nix { pkgs = import iohkNixMain.nixpkgs { }; }
     // sourcesOverride;
   iohkNixMain = import sources.iohk-nix { };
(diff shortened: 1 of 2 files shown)


Jasagredo (approved):
Looks good to me.

jbgi (approved):

EncodePanda:
bors merge

PR #3044, opened 2021-04-07
docs: Record how to establish bors persmissions

nfrisby:
We've had to track down our notes about how to complete this onboarding step more than once now, so I'm recording it in the place where at least the new person will be looking :)


dcoutts (approved):
Yeah. I always forget this. Thanks.

PR #3031, opened 2021-04-02
Added PeerSelectionCounters and respective tracers.

bolt12:
This PR deprecates #3030

PR #3047, opened 2021-04-08
Modified countConnections to use ConnectionManagerCounters instead.

bolt12:
Referencing @coot comment about this change: https://github.com/input-output-hk/ouroboros-network/pull/3031#discussion_r607983711


coot:
There's failure on hydra, could you force push to check if it persists?

coot:
Does not build, is this because `p2p-master`, fails to build?

bolt12:
> Does not build, is this because `p2p-master`, fails to build?

I think so, my branch is up-to-date with `p2p-master`

Issue #3025, opened 2021-03-30
Upgrade cardano-base and cardano-ledger-specs


EncodePanda:
@DavidEichmann let us known that in his recent PRs:

* https://github.com/input-output-hk/cardano-base/pull/209
* https://github.com/input-output-hk/cardano-ledger-specs/pull/2202

he added new CBOR instances. We want to try upgrade cardano-base and cardanon-ledger-spec, either on master on our alozno branch  https://github.com/input-output-hk/ouroboros-network/pull/2932

PR #3054, opened 2021-04-12
upgrade dependencies in ouroboros-network

EncodePanda:
- [x] carano-predule
- [x] cardano-base
- [x] cardano-ledger-specs

Fixes https://github.com/input-output-hk/ouroboros-network/issues/3025


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -49,6 +49,7 @@ import           Ouroboros.Consensus.Util.Condense
 import qualified Shelley.Spec.Ledger.API as SL
 import           Shelley.Spec.Ledger.BlockChain as SL (TxSeq (..))
 import qualified Shelley.Spec.Ledger.UTxO as SL (txid)
+import qualified Cardano.Ledger.Core as Core
```
Jasagredo, line 52: Isn't this import redundant?

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
```diff
@@ -80,7 +80,7 @@ import qualified Shelley.Spec.Ledger.API as SL
 import qualified Shelley.Spec.Ledger.BaseTypes as SL (Seed (..),
                      mkNonceFromNumber, textToUrl)
 import qualified Shelley.Spec.Ledger.BlockChain as SL (TxSeq (..))
-import qualified Shelley.Spec.Ledger.Coin as SL (DeltaCoin (..))
+import qualified Cardano.Ledger.Coin as SL (DeltaCoin (..))
```
Jasagredo, line 83: Maybe worth changing the qualified name to something that does not resemble Shelley as this seems to have been moved away from `Shelley.Spec`. And possibly moving the import 8 lines upward to have it sorted.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -38,6 +38,8 @@ import qualified Cardano.Ledger.Shelley.Constraints as SL
 import           Ouroboros.Consensus.Shelley.Protocol.Crypto (StandardCrypto)
 import qualified Shelley.Spec.Ledger.API as SL
 import qualified Shelley.Spec.Ledger.BaseTypes as SL
+import qualified Shelley.Spec.Ledger.Tx as SL
+import qualified Cardano.Ledger.Core as Core
```
Jasagredo, line 42: `Cardano.Ledger.Core` seems to already be in scope with qualified name `LC`.

Jasagredo (changes requested):
Apart from those import remarks, the changes look good to me.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
EncodePanda: I've forgot to format it :) Crap.
Regarding rename.. Hmm .. I think `SL` stands for "Spec Ledger" but I might be wrong

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
EncodePanda: Yeah, nice catch. I've removed `Tx`

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
EncodePanda: This got formatted

Jasagredo (approved):
LGTM :+1:

nfrisby (approved):
Nice diff!

I'd like a comment on the `Tx` instance if you have the patience for another iteration, but it's not crucial -- we'll revisit that soon enough.

Also, FYI, I renamed the name of the PR.

Approved. Thanks!

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -66,6 +67,8 @@ type EraCrypto era = Crypto era
   Era polymorphism
 -------------------------------------------------------------------------------}
 
+type instance Core.Tx era = SL.Tx era
```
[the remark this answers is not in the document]
EncodePanda: I've added a comment here and for `Witnesses` equivalence as well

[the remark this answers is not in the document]
DavidEichmann: I have a question: is the Tx type family only ever indexed on a sheley based eras? More specifically, is the Byron era never used in this type level machinery?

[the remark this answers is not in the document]
DavidEichmann: Ah ok the package description says so:
>   Cardano ledgers from the Shelley release onwards share a core basis rooted in
  the Shelley ledger specification. This package abstracts a number of components
  which we expect to be shared amongst all future ledgers implemented around this base.

DavidEichmann (approved):

EncodePanda:
bors merge

PR #3055, opened 2021-04-12
Don't rebuild github-pages on every push

coot:


nfrisby (approved):
The email GitHub sent me listed some unexpected changed files, but I'm guessing that was before you rebased your branch, or something like that.

The small 2 line diff LGTM. Approving 👍

coot:
Thanks @nfrisby; I initially pushed this forking from a wrong branch, now it only modifies the two files that are expected.

coot:
bors merge

PR #3056, opened 2021-04-13
Refactored localRootPeersProvider to accomodate the new LocalRootPeers (DNS task)

bolt12:


bolt12:
Still need to work on local config changes

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -101,12 +101,6 @@ instance Show TraceLedgerPeers where
     show FallingBackToBootstrapPeers = "Falling back to bootstrap peers"
     show DisabledLedgerPeers = "LedgerPeers is disabled"
 
--- | A relay can have either an IP address and a port number or
--- a domain with a port number
-data RelayAddress = RelayDomain !DomainAddress
-                  | RelayAddress !IP.IP !Socket.PortNumber
-                  deriving (Show, Eq, Ord)
-
```
bolt12, line 109: Had to remove it from `LedgerPeers` and define it in `RootPeersDNS` because of circular dependencies

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -265,34 +272,114 @@ newResolverResource resolvConf = go
...
+              ipsAddr
+
+  atomically $
+    writeTVar rootPeersGroupsVar placeholder'
+
+  -- Launch DomainAddress monitoring threads and
+  -- wait for threads to return or for local config changes
+  withAsyncAll (map (monitorDomain rr) domains) $ \as ->
+    atomically $ do
+      if null as
```
bolt12, line 333: Is this necessary? what happens if I wait on an empty list?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -265,34 +272,114 @@ newResolverResource resolvConf = go
...
+
+    updateList :: Int -> a -> [a] -> [a]
+    updateList at v l =
+      case splitAt at l of
+        (before, _:after) -> before ++ v:after
+        _ -> l
+
+    resolveDomain
+      :: DNS.Resolver
+      -> (DomainAddress, PeerAdvertise)
```
coot, line 358: why not two arguments?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
coot: looking at the implementation of `withAsyncAll` it will run the action with `as = []`; `waitAnySTM` will block on the empty list: https://hackage.haskell.org/package/async-2.2.3/docs/src/Control.Concurrent.Async.html#waitAny

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -265,34 +272,114 @@ newResolverResource resolvConf = go
 --
 
 data TraceLocalRootPeers =
-       TraceLocalRootDomains [(DomainAddress, PeerAdvertise)]
+       TraceLocalRootDomains [(Int, Map RelayAddress PeerAdvertise)]
```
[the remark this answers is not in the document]
bolt12: `Int` is to follow the same new structure of `LocalRootPeers`. I guess I can ditch the `Int` and just trace the Map, but I think knowing the groups targets is also useful, no?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: no actual reason, I can change to 2 arguments if prefered!

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: Does that mean I can simplify this code?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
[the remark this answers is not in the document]
karknu: Int is the configured valency for the local producer group. I think we should trace it too.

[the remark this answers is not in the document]
coot: Please add haddock to document this.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -278,52 +279,73 @@ data TraceLocalRootPeers =
...
+              placeholder
+              ipsAddr
+
+  atomically $
+    writeTVar rootPeersGroupsVar placeholder'
+
+  -- Launch DomainAddress monitoring threads and
+  -- wait for threads to return or for local config changes
+  withAsyncAll (map (monitorDomain rr) domains) $ \as ->
+    atomically $ do
```
coot, line 332: The `do` is not needed here.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -278,52 +279,73 @@ data TraceLocalRootPeers =
...
+      (void $ waitAnySTM as) <|> waitConfigChanged domainsGroups
```
coot: or `void (waitAnySTM as)`

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -288,29 +294,48 @@ localRootPeersProvider tracer timeout resolvConf rootPeersGroupsVar domainsGroup
...
-    -- be updated in the correct group.
+                  | (group, RelayDomain domain, pa) <- addresses ]
+        ipsAddr = [ (group, IP.toSockAddr (ip, port), pa)
+                  | (group, RelayAddress ip port, pa) <- addresses ]
+
+        -- Since we want to preserve the number of groups, the targets, and
+        -- the addresses within each group, we fill the TVar with
+        -- a placeholder list, in order for each monitored DomainAddress to
+        -- be updated in the correct group.
+        placeholder = map (\(t, _) -> (t, Map.empty)) domainsGroupsList
```
coot: please add type annotations for `addresses`, `domains` and `placeholder`.  I find it much easier to read the code when I don't need to infer the types.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -288,29 +294,48 @@ localRootPeersProvider tracer timeout resolvConf rootPeersGroupsVar domainsGroup
...
+        placeholder' =
+          foldr (\(group, addr, pa) ph
+                  -> let (target, entry) = ph !! group
+                     in updateList group (target, Map.insert addr pa entry) ph)
+                placeholder
```
coot: I would inline the `placeholder` here and renamed the `placeholder'`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -265,34 +272,113 @@ newResolverResource resolvConf = go
...
+      -- the addresses within each group, we fill the TVar with
+      -- a placeholder list, in order for each monitored DomainAddress to
+      -- be updated in the correct group.
+      placeholder = map (\(t, _) -> (t, Map.empty)) domainsGroups
+      placeholder' =
+        foldr (\(group, addr, pa) ph
+                -> let (target, entry) = ph !! group
+                   in updateList group (target, Map.insert addr pa entry) ph)
+              placeholder
+              ipsAddr
```
coot, line 324: I wonder if we could simplify this with something like [`mapKeysWith`](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Map-Internal.html#v:mapKeysWith).

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -265,34 +272,113 @@ newResolverResource resolvConf = go
...
+
   where
-    monitorDomain :: Resource DNSorIOError DNS.Resolver -> (DomainAddress, PeerAdvertise) -> IO Void
-    monitorDomain rr0 (domain@DomainAddress {daDomain, daPortNumber}, advertisePeer) =
+    waitConfigChanged :: [(Int, Map RelayAddress PeerAdvertise)] -> STM IO ()
+    waitConfigChanged dg = do
+      dg' <- readTVar domainsGroupsVar
+      check (dg /= dg')
+
+    updateList :: Int -> a -> [a] -> [a]
```
coot, line 348: I think, we should use an `IntMap` as an internal representation instead of encoding data as a list. This should eliminate `updateList`.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -665,6 +666,15 @@ runDataDiffusion tracers
         targetNumberOfActivePeers = min 2 (targetNumberOfActivePeers daPeerSelectionTargets)
       }
 
+    daLocalRootPeersVar <- newTVarIO $
+                            ([( 1
+                              , Map.fromList $
+                                  map (\(d,p) -> (RelayDomain d, p))
+                                      daLocalRootPeers)])
+    -- ^ TODO: This is just a simple transformation
+    daPublicRootPeersVar <- newTVarIO $ map RelayDomain daPublicRootPeers
```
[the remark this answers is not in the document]
bolt12: This PR only focuses on changing the providers to use TVars. This one #3079 refactors diffusion arguments and does this.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: The problem with `IntMap` is that the `Int` for the valency is not unique, and we lose information about each group by using an `IntMap`. We can use an `IntMap [(Int, Map RelayAddress PeerAdvertise)]` where the key is the indices of the list, that could work

bolt12: Should I use an IntMap internally and then export to `[(Int, Map RelayAddress PeerAdvertise)]` ?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: I'm not sure how

coot: Would this work: [this](https://github.com/input-output-hk/ouroboros-network/blob/coot/bolt12/master-local-root-providers/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs#L330).

coot: We don't need to use an `IntMap` if we don't modify the spine of the list.

bolt12: That would not work because the `Int` in `[(Int, Map RelayAddress PeerAdvertise)]` is not unique and 2 different groups might have the same valency

coot: Ach right the `Int` is the target; what about:
```
rootPeersGroups = map (\(target, m) -> (target, f m)) domainsGroups
  -- where `f` the same as in my branch
```

bolt12: Ah yes I think that would work! :D

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -305,28 +394,25 @@ localRootPeersProvider tracer timeout resolvConf rootPeersVar domains = do
...
-                                                , advertisePeer)
-                                              | (addr, _ttl) <- results ]
-                    rootPeers' :: Map DomainAddress (Map Socket.SockAddr PeerAdvertise)
-                    rootPeers' = Map.insert domain resultsMap rootPeers
+                rootPeersGroups <- readTVar rootPeersGroupsVar
+                let rootPeersGroups' = IntMap.fromList $ zip [0,1..] rootPeersGroups
+                    (target, entry) = rootPeersGroups' IntMap.! group
+                    resultsMap      = Map.fromList (map fst results)
+                    newEntry        = Map.union entry resultsMap
+                    newRootPeers    =
```
coot, line 407: It seems you wanted to indent all `=`.

coot (commented):
Do we have any evidence that this works as designed?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -265,34 +273,115 @@ newResolverResource resolvConf = go
...
+                     RelayAddress ip port ->
+                       IP.toSockAddr (ip, port)
+                     _ ->
+                       error "localRootPeersProvider: impossible happend"
+                   )
+               . Map.filterWithKey
+                   (\k _ -> case k of
+                     RelayAddress {} -> True
+                     RelayDomain {}  -> False
+                   )
```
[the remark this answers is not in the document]
bolt12: You're right, I forgot to remove them!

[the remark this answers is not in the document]
bolt12: actually they are needed for `monitorDomain`,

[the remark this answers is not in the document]
bolt12: > Do we have any evidence that this works as designed?

What do you mean by this @coot ?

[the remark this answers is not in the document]
coot: Do we have any tests? I think we don't, so how did you verified that it works?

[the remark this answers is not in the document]
bolt12: In the previous version (without `rootPeersGroup`) what I did was to look at the `TraceLocalRootDomains`. We should have tests, but as you said a couple days ago we should refactor `RootPeerDNS.hs` to be able to use `IOSim`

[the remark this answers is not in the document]
bolt12: Or should I write tests in `IO` for now?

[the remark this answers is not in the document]
coot: If you're confident that this works fine (e.g. you observed that node can resolve dns names), then we can merge it then work on `IOSim` / `IO` test in the simulation workstream.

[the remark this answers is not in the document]
bolt12: Yes, I am confident that his indeed resolves dns names!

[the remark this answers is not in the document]
coot: :+1:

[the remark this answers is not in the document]
coot: So let's just try to find out if we can make the `domains` thing more elegant, and we can merge it; I noticed that haddocks are not building, but it seems not related to this PR (`p2p-master` fails too). We'll need to fix this, but outside of this PR.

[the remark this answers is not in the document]
bolt12: okay, sounds good!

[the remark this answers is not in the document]
bolt12: What about merging the two list comprehensions:

```haskell
      domains :: [(Int, DomainAddress, PeerAdvertise)]
      domains = [ (group, domain, pa)
                  | (group, (_, m)) <- zip [0..] domainsGroups
                  , (RelayDomain domain, pa) <- Map.toList m ]
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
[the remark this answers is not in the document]
coot: These haddocks still needs to be added.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -305,28 +392,25 @@ localRootPeersProvider tracer timeout resolvConf rootPeersVar domains = do
...
-                let resultsMap :: Map Socket.SockAddr PeerAdvertise
-                    resultsMap = Map.fromList [ ( Socket.SockAddrInet
-                                                    daPortNumber
-                                                    (IP.toHostAddress addr)
-                                                , advertisePeer)
-                                              | (addr, _ttl) <- results ]
-                    rootPeers' :: Map DomainAddress (Map Socket.SockAddr PeerAdvertise)
-                    rootPeers' = Map.insert domain resultsMap rootPeers
+                rootPeersGroups <- readTVar rootPeersGroupsVar
+                let rootPeersGroups' = IntMap.fromList $ zip [0,1..] rootPeersGroups
```
coot, line 401: Nice progress :)! Two more remarks:

* it seems as we could store the `IntMap` inside `rootPeersGroupsVar`
* it would also be cleaner if we'd use a `Group` newtype wrapper rather than a bare `Int`

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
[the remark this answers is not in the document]
bolt12: https://github.com/input-output-hk/ouroboros-network/pull/3056/files#diff-d5dd266642bee6a6cf1ee9cc05dffd5aa0df7f7e2bdbe88d813593f55faa428eR277

I added it!

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -305,28 +392,25 @@ localRootPeersProvider tracer timeout resolvConf rootPeersVar domains = do
...
+                    (target, entry)  = rootPeersGroups' IntMap.! group
+                    resultsMap       = Map.fromList (map fst results)
+                    newEntry         = Map.union entry resultsMap
```
coot, line 404:
```suggestion
                    entry'         = entry
                                    <> Map.fromList (map fst results)
```

* `rootPeersGroups'` denote the new value, so let's be consistent and use `entry'` here as well.
* maybe it's just me, but using `<>` makes it more obvious that we're doing the right thing.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: > it seems as we could store the IntMap inside rootPeersGroupsVar

Hmmm I guess we could but the type of rootPeersGroupsVar is meant to be the same as the diffusion interface type

> it would also be cleaner if we'd use a Group newtype wrapper rather than a bare Int

We can also do this but I'd like to avoid changing this now since that would impact config parameters parsing (the same for the suggestion above)

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: Agree!

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
coot: In the long term it's better to merge clean things to `p2p-master` / `master`, otherwise we might end up with technical debt that we'll not clean for a long time. It shouldn't be too difficult to update the `Diffusion` interface.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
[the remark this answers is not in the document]
coot: Sorry, it seems I missed it.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: I understand that. In that case it might be better to do that here: https://github.com/input-output-hk/ouroboros-network/pull/3079 and then merge into this one and then, finally, merge into `p2p-master`. The scope of this PR wasn't to change anything in the Diffusion layer

coot: ok

coot (approved):
LGTM

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: See: https://github.com/input-output-hk/ouroboros-network/pull/3079/commits/16ad70aa569026a07d3a76bbbd51e4fa48ba1a1a

PR #3061, opened 2021-04-14
Add Haskell-Language-Server as a build input to the shell environment.

Jasagredo:
Adds the Haskell-Language-Server currently at version 1.1.0 (not yet released
in Hackage) through a niv dependency.

For a final user, adding this is fine because it is cached by Hydra so it will 
just download the whole shell.

Incidentally:
- Using the same set of nixpkgs as for the rest of the build when providing
  it to Niv for non-builtin fetchers.
- Replaced the sources.nix file with a newer generated version (probably this
  should actually be updated each time we update the niv version).
- Also rephrase how stylish-haskell was being built as we can just point
  to the repository at a specific branch with niv.
- Adding .envrc to .gitignore as not everybody uses lorri.

Co-authored-by: Pawel Szulc <paul.szulc@gmail.com>

### DIFF
--- a/nix/hls.nix
+++ b/nix/hls.nix
@@ -1,0 +1,25 @@
+################################################
+# Prepare Haskell-language-server to be added as a buildInput
+################################################
+{ pkgs
+, ghcVersion
+}:
+let
+  inherit (pkgs.commonLib) sources;
+
+  planConfig = {
+    compiler-nix-name = ghcVersion;
+    # get the sources from niv
+    src = sources.hls-released;
+    configureArgs = "--disable-benchmarks";
+    modules = [{ enableSeparateDataOutput = true; }];
+  };
+
+  # bring all exes into scope
+  hlsExes = (pkgs.haskell-nix.cabalProject planConfig).haskell-language-server.components.exes;
+in
+{
+  hls = hlsExes.haskell-language-server;
+  hls-wrapper = hlsExes.haskell-language-server-wrapper;
+  implicit-hie = pkgs.haskellPackages.implicit-hie;
+}
--- a/nix/stylish-haskell.nix
+++ b/nix/stylish-haskell.nix
@@ -1,17 +1,18 @@
+################################################
+# Prepare stylish-haskell to be added as a buildInput
+################################################
 { pkgs
+, ghcVersion
 }:
 let
-  hsPkgs = pkgs.haskell-nix.stackProject {
-    compiler-nix-name = "ghc8104";
+  inherit (pkgs.commonLib) sources;
+
+  planConfig = {
+    compiler-nix-name = ghcVersion;
+    # get the sources from niv
+    src = sources.stylish-haskell;
     modules = [ ];
-    src = pkgs.fetchFromGitHub {
-      owner = "jaspervdj";
-      repo = "stylish-haskell";
-      # 0.12.2.0 with custom feature that will be added to 0.12.3.0 release
-      # see https://github.com/jaspervdj/stylish-haskell/commit/3d5348041e2efbcae95f1c62ace5ac0ffd9629d2
-      rev = "3d5348041e2efbcae95f1c62ace5ac0ffd9629d2";
-      sha256 = "1rms68fll6w1rzdg9ppjs9g246mbgxi5589vxb8r8pcn5v1ncfj2";
-    };
   };
+
 in
-hsPkgs.stylish-haskell.components.exes.stylish-haskell
+(pkgs.haskell-nix.stackProject planConfig).stylish-haskell.components.exes.stylish-haskell
--- a/shell.nix
+++ b/shell.nix
@@ -9,8 +9,29 @@
 }:
 with pkgs;
 let
-  # This is temporally till we get a stylish 0.12.3.0 release available in hackage
-  stylish-haskell = import ./nix/stylish-haskell.nix { inherit pkgs; };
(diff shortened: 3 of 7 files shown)


nfrisby (approved):
Approved. Please remove that comment about opt-in, since we're taking @michaelpj's advice to rely on Hydra to cache the HLS build. That's not present in this PR yet, but we'll figure it out and merge it shortly.

nfrisby:
@dcoutts @coot @karknu 

This PR adds the `haskell-language-server` to our `nix-shell` environment. That is not a small dependency, but the recommendations we're received and others' experience has been that the Hydra cache mitigates the potential build-delays.

Are you three (Network Team) OK with this addition to the repositories `nix-shell`?

karknu:
> @dcoutts @coot @karknu
> 
> This PR adds the `haskell-language-server` to our `nix-shell` environment. That is not a small dependency, but the recommendations we're received and others' experience has been that the Hydra cache mitigates the potential build-delays.
> 
> Are you three (Network Team) OK with this addition to the repositories `nix-shell`?

I don't think any of us use nix-shell.

coot (approved):

nix/stylish-haskell.nix
```diff
@@ -1,17 +1,16 @@
...
+  planConfig = {
+    compiler-nix-name = ghcVersion;
+    # get the sources from niv
+    src = sources.stylish-haskell;
+    modules = [];
   };
+
 in
-hsPkgs.stylish-haskell.components.exes.stylish-haskell
+(pkgs.haskell-nix.stackProject planConfig)
```
coot, line 15: would `cabalProject` work as well?

Jasagredo: Most likely it would (I seem to recall that it actually worked) but we stayed with `stackProject` as it was how it was being built before the changes.

Jasagredo: We are sticking with `stackProject` because it is the same way `stylish-haskell` builds artifacts on its own CI.

coot: :+1:

EncodePanda (approved):

nix/hls.nix
```diff
@@ -0,0 +1,22 @@
+{ pkgs
+, ghcVersion
+}:
+let
+  sources = import ./sources.nix { };
```
jbgi, line 5:
```suggestion
  inherit (pkgs) sources;
```
this should allows to remove nixpkgs entry from sources.json

nix/stylish-haskell.nix
```diff
@@ -1,17 +1,15 @@
 { pkgs
+, ghcVersion
 }:
 let
-  hsPkgs = pkgs.haskell-nix.stackProject {
-    compiler-nix-name = "ghc8104";
+  sources = import ./sources.nix { };
```
jbgi, line 5:
```suggestion
  inherit (pkgs) sources;
```

jbgi (changes requested):

jbgi (approved):

nix/stylish-haskell.nix
Jasagredo: Suggestion added, thanks!

shell.nix
```diff
@@ -9,8 +9,13 @@
 }:
 with pkgs;
 let
+  ghcVersion = "ghc8104";
+
   # This is temporally till we get a stylish 0.12.3.0 release available in hackage
-  stylish-haskell = import ./nix/stylish-haskell.nix { inherit pkgs; };
+  stylish-haskell = import ./nix/stylish-haskell.nix { inherit pkgs ghcVersion; };
+
+  hls = import ./nix/hls.nix { inherit pkgs ghcVersion; };
```
nfrisby, line 17: As with `stylish-haskell` above, this is also temporary until eg `1.1.0.0` is on Hackage, right?

For both of these, once it's on Hackage, we can juse use the `tools` field, right? (And simultaneously remove them from the `sources.json`.)

Might be helpful to update these comments to say as much.

nfrisby (approved):
Approved.

I did make one request for some additional comments.

Thanks all!

shell.nix
Jasagredo: Reworded the comment, adding some explanation about why this is here, and what to do when the relevant versions arrive to Hackage. :+1:

Jasagredo:
After confirming locally that (almost) everything was copied from caches after the hydra evaluation of `native.shell` succeeded:
```
these derivations will be built:
  /nix/store/in3kpaw7kd7ns959fhl0j44dd6xb3nv1-builder.pl.drv
  /nix/store/4jp06k9xm6wv99mcxiwrs9xjziasdbad-cddl-0.8.9.drv
  /nix/store/29hl3ikkyzwcjdlipk7g6bxn6mdackic-cddl-0.8.9.drv
  /nix/store/7hlhy17pr1hn31dn2dvah6fz8k6b2a68-cbor-diag-0.5.6.drv
  /nix/store/i1nypbj1r3qslsz1xy87pgvgwj1hdnk4-ghc-shell-for-packages-ghc-8.10.4-env.drv
  /nix/store/pnbdbgjpi9z1vyhfyw5lzz9x6fngljxd-cbor-diag-0.5.6.drv
these paths will be fetched (443.06 MiB download, 5228.94 MiB unpacked):
...
```
I then proceed to merge this PR

Jasagredo:
bors merge

PR #3062, opened 2021-04-14
New KeepAlive codec

coot:
The new keep alive codec produces valid CBOR encoding.  This PR provides:

- KeepAlive: codec with valid CBOR encoding
- ouroboros-network-testing: added prop_valid_cbor
- Test that mini-protocol codecs produce valid CBOR
- NodeToNodeV_7: use new KeepAlive codec

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
@@ -228,9 +228,10 @@ defaultCodecs :: forall m blk. (IOLike m, SerialiseNodeToNodeConstraints blk,
                                 ShowProxy (GenTxId blk), ShowProxy (GenTx blk))
               => CodecConfig       blk
               -> BlockNodeToNodeVersion blk
+              -> NodeToNodeVersion
               -> Codecs blk DeserialiseFailure m
                    ByteString ByteString ByteString ByteString ByteString ByteString ByteString
-defaultCodecs ccfg version = Codecs {
+defaultCodecs ccfg version nodeToNodeVersion = Codecs {
       cChainSyncCodec =
         codecChainSync
           enc
@@ -278,7 +279,9 @@ defaultCodecs ccfg version = Codecs {
           dec
 
     , cKeepAliveCodec =
-        codecKeepAlive
+        if nodeToNodeVersion <= NodeToNodeV_6
+          then codecKeepAlive
+          else codecKeepAlive_v2
     }
   where
     p :: Proxy blk
@@ -443,11 +446,11 @@ mkApps
      )
   => NodeKernel m remotePeer localPeer blk -- ^ Needed for bracketing only
   -> Tracers m remotePeer blk e
-  -> Codecs blk e m bCS bCS bBF bBF bTX bTX2 bKA
+  -> (NodeToNodeVersion -> Codecs blk e m bCS bCS bBF bBF bTX bTX2 bKA)
   -> m ChainSyncTimeout
   -> Handlers m remotePeer blk
   -> Apps m remotePeer bCS bBF bTX bTX2 bKA ()
-mkApps kernel Tracers {..} Codecs {..} genChainSyncTimeout Handlers {..} =
+mkApps kernel Tracers {..} mkCodecs genChainSyncTimeout Handlers {..} =
     Apps {..}
   where
     aChainSyncClient
@@ -474,7 +477,7 @@ mkApps kernel Tracers {..} Codecs {..} genChainSyncTimeout Handlers {..} =
               (_, trailing) <-
                 runPipelinedPeerWithLimits
                   (contramap (TraceLabelPeer them) tChainSyncTracer)
-                  cChainSyncCodec
+                  (cChainSyncCodec (mkCodecs version))
                   (byteLimitsChainSync (const 0)) -- TODO: Real Bytelimits, see #1727
                   (timeLimitsChainSync chainSyncTimeout)
                   channel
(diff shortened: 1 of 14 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -443,11 +446,11 @@ mkApps
      )
   => NodeKernel m remotePeer localPeer blk -- ^ Needed for bracketing only
   -> Tracers m remotePeer blk e
-  -> Codecs blk e m bCS bCS bBF bBF bTX bTX2 bKA
+  -> (NodeToNodeVersion -> Codecs blk e m bCS bCS bBF bBF bTX bTX2 bKA)
   -> m ChainSyncTimeout
   -> Handlers m remotePeer blk
   -> Apps m remotePeer bCS bBF bTX bTX2 bKA ()
-mkApps kernel Tracers {..} Codecs {..} genChainSyncTimeout Handlers {..} =
```
nfrisby, line 450: Minor stylistic suggestion: what do you think about instead adding a where clause:

```
   Codecs {..} = mkCodecs version
```

EG you wouldn't have to change all the use sites.

nfrisby: Ah, nevermind: I was confused about scope.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs
```diff
@@ -228,9 +228,10 @@ defaultCodecs :: forall m blk. (IOLike m, SerialiseNodeToNodeConstraints blk,
                                 ShowProxy (GenTxId blk), ShowProxy (GenTx blk))
               => CodecConfig       blk
               -> BlockNodeToNodeVersion blk
+              -> NodeToNodeVersion
               -> Codecs blk DeserialiseFailure m
                    ByteString ByteString ByteString ByteString ByteString ByteString ByteString
-defaultCodecs ccfg version = Codecs {
+defaultCodecs ccfg version nodeToNodeVersion = Codecs {
```
nfrisby, line 234: I'm initially hesitant to add this parameter, because it's at least partially redundant with the existing `version` parameter.

  * We _could_ take only `nodeToNodeVersion` and map it to a `version` via `supportedNodeToNodeVersions`. However that lookup would be partial.
  
  * The way the dataflow currently works is that this pair of arguments comes from a traversal of `supportedNodeToNodeVersions`, so there is no such partiality to deal with.

  * Perhaps we should instead simply add an `assert` that `supportedNodeToNodeVersions` includes the `(version, nodeToNodeVersion)` pair.
  
What do you think @coot?

coot: What are the dependencies between block version and protocol version, some times a new block version might require a new protocol version, but never the other way around.  This means that from a network version we can derive the block version, rather than the other way around and that's quite nicely expressed by `supportedNodeToNodeVersions` type.  So it's actually `version` that's redundant rather than `nodeToNodeVersion`.

Let's consider this: we define a new version which for some reason the consensus will not use.  This is not a problem even if there's an independent implementation (a fork) which starts using the network version with some block version. That network version would never be negotiated between our node and the independent one.

It seems ok to have:
```
defaultCodecs :: HasNetworkProtocolVersion blk => CodecConfig blk -> NodeToNodeVersion -> Codecs ...
```

If I am right we don't even need an assertion, `error "impossible"` seems fine if I am not mistaken above; let's see if I'll still think the same in the morning :).

nfrisby: The negotiated `NodeToNodeVersion` determines the `BlockNodeToNodeVersion`. So yes, the redundancy could be resolved by dropping the `BlockNodeToNodeVersion` argument and a `Map.lookup` on `supportedNodeToNodeVersions`.

https://github.com/input-output-hk/ouroboros-network/blob/dea09273c87578a2047445e4434ba30df6570ac2/ouroboros-consensus/src/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs#L42-L43

Also, right now, the set of possible-to-negotatiate versions is also determined by `supportNodeToNodeVersions`.

https://github.com/input-output-hk/ouroboros-network/blob/ec79b87254589276e75f11ea95c069266b7b151d/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L362

https://github.com/input-output-hk/ouroboros-network/blob/ec79b87254589276e75f11ea95c069266b7b151d/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L664-L667

So `Map.findWithDefault (error _)` instead of `Map.lookup` could be justified right now by a simple (manual) data flow analysis.

I do worry though that that is sufficiently non-local reasoning to become a footgun in the future. Maybe we could refine the `supportedNodeToNodeVersions` interface (or perhaps even the `DiffusionApplications` callback interface) to better encode this functional relation invariant in the types.

coot: > I do worry though that that is sufficiently non-local reasoning to become a footgun in the future. 

Yes, that's a good point;

> Maybe we could refine the supportedNodeToNodeVersions interface (or perhaps even the DiffusionApplications callback interface) to better encode this functional relation invariant in the types.

I'd be glad to work on this, but I'd prefer to do this later, right now we have other priorities (p2p).

nfrisby: > I'd prefer to do this later, right now we have other priorities (p2p).

100% agree

karknu (approved):
LGTM

coot:
bors merge

PR #3063, opened 2021-04-14
consensus: update sketch implementation plan for Ouroboros Genesis

nfrisby:
Update the sketched plan after our discussion on 2021 April 13 and again on 2021 August 9 and 2021 August 12.

### DIFF
--- a/ouroboros-consensus/docs/GenesisDecomposition.md
+++ b/ouroboros-consensus/docs/GenesisDecomposition.md
@@ -187,3 +187,139 @@ I've also updated this file after a discussion with Duncan on 2021 Apr 13.
 
       * TODO testing etc - we'd _very much really_ like to use the ThreadNet
         rewrite for this
+
+-----
+
+Updated on 2021 August 9, after much additional thought and broader
+reconsiderations, kicked off by Javier Sagredo's observation of a stalling
+attack vector in the original sketch above.
+
+This new sketch updates much-but-not-all of the origial sketch above.
+
+- Execution begins in the _Syncing_ state.
+
+- While we are Syncing:
+
+    - If our valency falls below some threshold, then BlockFetch stops sending
+      new fetch requests until sufficient valency is recovered.
+
+    - BlockFetch can only download blocks from the headers that the density
+      rule approves.
+
+        - The density rule is: compare header chains based on the number of
+          headers in the relevant Genesis window (the 3k/f slots after the
+          intersection), though if the headers do not span the Genesis window
+          and the peer claims to have more headers we must wait for them
+          (because they might also be in the window).
+
+        - The Ouroboros Genesis paper proves -- excepting only disasterous
+          intervals -- that density rule will always strictly prefer the honest
+          chain over any possible alternative.
+
+    - Therefore, we require that each peer's highwater blockno is increasing
+      "fast enough on average" until we're at their tip, with the only
+      exceptional circumstance being when their latest header is beyond our
+      forecast range (since we don't even request a next header while that is
+      true).
+
+        - TODO Do we actually need that exception? Under what circumstances
+          would it be relevant, during Syncing?
+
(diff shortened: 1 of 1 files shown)


nfrisby:
That discussion didn't reach any conclusions, so we'll be having more. This PR just writes-up what came up during that first meeting.

nfrisby:
I've added a draft commit with another overall sketch -- this one reflects much additional thought about what exactly BlockFetch can and can't do, when we should respect the peers' claimed tips, and related timeouts.

ouroboros-consensus/docs/GenesisDecomposition.md
```diff
@@ -127,9 +136,190 @@ in 2021 March using chapter 21 of the Consensus Report.
...
+
+        - TODO I'm anticipating a token bucket for enforcing "fast enough on
+          average", but there remain plenty of details and thresholds to
+          consider.
+
+        - A possible refinement: if they can promise to send a specific k+1st
+          block (which the honest nodes would always do, up to their immutable
+          tip), then they're allowed to be somewhat slower, since we'll
+          disconnect from them if either they don't deliver that block or if
+          the eventual densest chain does not include that block.
```
Jasagredo, line 236: I assume this is the case but I want to confirm it. We are going to disconnect from adversarial peers and then wait until we reconnect to enough peers to continue?

ouroboros-consensus/docs/GenesisDecomposition.md
```diff
@@ -127,9 +136,190 @@ in 2021 March using chapter 21 of the Consensus Report.
...
+
+        - A possible refinment: each peer can offer _jump points_ that are
```
Jasagredo, line 238: This will happen in ChainSync, right?

ouroboros-consensus/docs/GenesisDecomposition.md
```diff
@@ -127,9 +136,190 @@ in 2021 March using chapter 21 of the Consensus Report.
...
+          transition? Even supposing validated, uninterruptible ChainSync
+          switches?
+
+        - TODO Perhaps we don't need it, since we assume we'll have at least
+          one honest peer. Their stream of headers should race ahead of the
+          corresponding stream of blocks until we're CaughtUp, and so that'll
+          hold back at least one of the other conjuncts. On the other hand, it
+          seems fine if we do need this, because of the timeout discussed
+          above.
+
```
Jasagredo, line 262: Wouldn't we also consider the case when the wallclock is closer than a genesis window length from the first intersection with out chain? In that case every chain tip is necessarily inside the genesis window, and we judge based on length. Or maybe this case is redundant with cases above.

Jasagredo: This probably would also imply the opposite situation, when the wallclock is further than a genesis window length from the [last?] intersection with our chain it means that if block production is not halted, all the forks will fill the genesis window and then we can decide on the only chain that is the honest one.

ouroboros-consensus/docs/GenesisDecomposition.md
```diff
@@ -127,9 +136,190 @@ in 2021 March using chapter 21 of the Consensus Report.
...
+              adversary can forge arbitrarily many blocks whenever it is
+              elected, just not on the same chain.
+
+- Transition from CaughtUp to Syncing whenever any of:
+
+    - The wallclock is "too far ahead" of the latest "meaningful" peer
+      interaction.
+
+        - TODO Sketch: we transition as soon N (?) of our peers' tips have a
+          time point that is more than LIM (?) behind our wallclock.
```
Jasagredo, line 307: As mentioned above, not sure if this would have to be something like "*the intersection* is a genesis window length behind our wallclock"

nfrisby:
Closing; migrated to https://github.com/input-output-hk/ouroboros-consensus/pull/487

PR #3069, opened 2021-04-19
toJSON/fromJSON instances for the new network config

bolt12:
WIP to add new network configuration. Just adds needed `ToJSON`/`FromJSON` instances to `ouroboros-network` defined types


bolt12:
Going to merge with a merge commit to keep things visible.

PR #3072, opened 2021-04-20
Make stake distribution vary through the MockPraos tests.

### LINKED ISSUE #2850: Vary the stake distribution in the MockPraos tests
The MockPraos tests currently use a fixed stake distribution. Though it has UTxO transactions, the active stake portion of the ledger is not updated accordingly -- it's simply a fixed parameter.

This would make the tests more realistic, certainly capable of finding additional classes of bugs.

Issue #2225 is somewhat related in that it's also about making the Mock tests more realistic. Mock ledgers can give simpler counterexamples (ie "smaller" in QC parlance). Pie in the sky culmination: perhaps shrinking could replace a real ledger with a mock ledger.

Jasagredo:
Closes #2850.

Previously, the stake distribution was static, it was stored in every block with
the same value and although it retrieved it from the last block of 2 epochs ago,
the value was the same. In this scenario, all the probabilities of forging were
fixed.

Now, a new parameter `praosEvolvingStake` was added to the `PraosConfig` (only
used in MockPraos), which will be used when determining the stake distribution at
a given slot.

Side changes can be seen in individual commits:
- Documentation
- Moving `Ouroboros.Consensus.Protocol.LeaderSchedule` to `ouroboros-consensus-mock`.
- Added hasktags and graphmod to shell.nix.

### DIFF
--- a/ouroboros-consensus-mock-test/test/Test/ThreadNet/Praos.hs
+++ b/ouroboros-consensus-mock-test/test/Test/ThreadNet/Praos.hs
@@ -3,6 +3,8 @@
 
 module Test.ThreadNet.Praos (tests) where
 
+import           Control.Monad (replicateM)
+import qualified Data.Map.Strict as Map
 import           Data.Word (Word64)
 import           Numeric.Natural (Natural)
 
@@ -30,21 +32,43 @@ import           Test.ThreadNet.Util.NodeRestarts
 import           Test.ThreadNet.Util.NodeToNodeVersion
 import           Test.ThreadNet.Util.SimpleBlock
 
+import           Ouroboros.Consensus.Node.ProtocolInfo
+                     (NumCoreNodes (NumCoreNodes), enumCoreNodes)
 import           Test.Util.HardFork.Future (singleEraFuture)
 import           Test.Util.Orphans.Arbitrary ()
+import           Test.Util.Slots (NumSlots (unNumSlots))
 
 data TestSetup = TestSetup
-  { setupEpochSize    :: EpochSize
-  , setupInitialNonce :: Natural
+  { setupEpochSize     :: EpochSize
+  , setupInitialNonce  :: Natural
     -- ^ the initial Shelley 'praosInitialEta'
     --
     -- This test varies it too ensure it explores different leader schedules.
-  , setupK            :: SecurityParam
-  , setupNodeJoinPlan :: NodeJoinPlan
-  , setupSlotLength   :: SlotLength
-  , setupTestConfig   :: TestConfig
+  , setupK             :: SecurityParam
+  , setupNodeJoinPlan  :: NodeJoinPlan
+  , setupSlotLength    :: SlotLength
+  , setupTestConfig    :: TestConfig
+  , setupEvolvingStake :: PraosEvolvingStake
   }
   deriving (Show)
+
+genEvolvingStake :: EpochSize -> TestConfig -> Gen PraosEvolvingStake
+genEvolvingStake epochSize TestConfig {numSlots, numCoreNodes} = do
+    chosenEpochs <- sublistOf [0..EpochNo $ max 1 maxEpochs - 1]
+    let l = fromIntegral maxEpochs
+    stakeDists <- replicateM l genStakeDist
+    return . PraosEvolvingStake . Map.fromList $ zip chosenEpochs stakeDists
+  where
+    maxEpochs = unNumSlots numSlots `div` unEpochSize epochSize
+    relativeStake ts nid stk = (nid, fromIntegral stk / ts)
+    genStakeDist = do
+      stakes <- vector (fromIntegral x) `suchThat` any (> 0) :: Gen [Amount]
+      let totalStake = fromIntegral $ sum stakes
+      return
+        . StakeDist
+        . Map.fromList
+        $ zipWith (relativeStake totalStake) (enumCoreNodes numCoreNodes) stakes
+    NumCoreNodes x = numCoreNodes
 
 instance Arbitrary TestSetup where
   arbitrary = do
@@ -59,6 +83,7 @@ instance Arbitrary TestSetup where
       let TestConfig{numCoreNodes, numSlots} = testConfig
 
       nodeJoinPlan   <- genNodeJoinPlan numCoreNodes numSlots
+      evolvingStake  <- genEvolvingStake epochSize testConfig
 
       pure $ TestSetup
         epochSize
@@ -67,6 +92,7 @@ instance Arbitrary TestSetup where
         nodeJoinPlan
         slotLength
         testConfig
+        evolvingStake
 
   -- TODO shrink
 
@@ -78,12 +104,13 @@ tests = testGroup "Praos"
 
 prop_simple_praos_convergence :: TestSetup -> Property
 prop_simple_praos_convergence TestSetup
-  { setupEpochSize    = epochSize
-  , setupK            = k
+  { setupEpochSize     = epochSize
+  , setupK             = k
   , setupInitialNonce
-  , setupNodeJoinPlan = nodeJoinPlan
-  , setupSlotLength   = slotLength
-  , setupTestConfig   = testConfig
+  , setupNodeJoinPlan  = nodeJoinPlan
+  , setupSlotLength    = slotLength
+  , setupTestConfig    = testConfig
+  , setupEvolvingStake = evolvingStake
   } =
     counterexample (tracesToDot testOutputNodes) $
     prop_general PropGeneralArgs
@@ -128,5 +155,6 @@ prop_simple_praos_convergence TestSetup
                                         k
                                         slotLength)
                                       setupInitialNonce
+                                      evolvingStake
             , mkRekeyM = Nothing
             }
--- a/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block/Praos.hs
+++ b/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block/Praos.hs
@@ -38,7 +38,6 @@ import           Ouroboros.Consensus.Ledger.SupportsProtocol
 import           Ouroboros.Consensus.Mock.Ledger.Address
 import           Ouroboros.Consensus.Mock.Ledger.Block
 import           Ouroboros.Consensus.Mock.Ledger.Forge
-import           Ouroboros.Consensus.Mock.Ledger.Stake
 import           Ouroboros.Consensus.Mock.Node.Abstract
 import           Ouroboros.Consensus.Mock.Protocol.Praos
 import           Ouroboros.Consensus.Protocol.Signed
@@ -120,24 +119,10 @@ instance ( SimpleCrypto c
          , PraosCrypto c'
          , Signable (PraosKES c') (SignedSimplePraos c c')
          ) => LedgerSupportsProtocol (SimplePraosBlock c c') where
-  protocolLedgerView   cfg _  = pretendTicked $ stakeDist cfg
-  ledgerViewForecastAt cfg st = constantForecastOf
-                                 (pretendTicked $ stakeDist cfg)
+  protocolLedgerView   _ _  = TickedTrivial
+  ledgerViewForecastAt _ st = constantForecastOf
+                                 TickedTrivial
                                  (getTipSlot st)
-
--- | Praos needs a ledger that can give it the "active stake distribution"
---
--- TODO: Currently our mock ledger does not do this, and just assumes that all
--- the leaders have equal stake at all times. In a way this is not wrong: it
--- is just a different instantiation of the same consensus algorithm (see
--- documentation of 'LedgerView'). Ideally we'd change this however, but it
--- may not be worth it; it would be a bit of work, and after we have integrated
--- the Shelley rules, we'll have a proper instance anyway.
-stakeDist :: LedgerConfig (SimplePraosBlock c c') -> StakeDist
-stakeDist = equalStakeDist . simpleMockLedgerConfig
-
-pretendTicked :: StakeDist -> Ticked StakeDist
-pretendTicked (StakeDist sd) = TickedStakeDist sd
 
 {-------------------------------------------------------------------------------
   Forging
--- a/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/UTxO.hs
+++ b/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/UTxO.hs
@@ -134,9 +134,13 @@ txOuts = Map.unions . map each . getMockTxs
         aux :: Ix -> TxOut -> (TxIn, TxOut)
         aux ix out = ((hashWithSerialiser toCBOR tx, ix), out)
 
+-- | @confirmed@ stands for all the transaction hashes present in the given
+-- collection.
 confirmed :: HasMockTxs a => a -> Set TxId
 confirmed = Set.fromList . map (hashWithSerialiser toCBOR) . getMockTxs
 
+-- |Update the Utxo with the transactions from the given @a@, by removing the
+-- inputs and adding the outputs.
 updateUtxo :: HasMockTxs a => a -> Utxo -> Except UtxoError Utxo
 updateUtxo = repeatedlyM each . getMockTxs
   where
--- a/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
+++ b/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
@@ -22,9 +22,11 @@ module Ouroboros.Consensus.Mock.Protocol.Praos (
   , HotKeyEvolutionError (..)
   , Praos
   , PraosChainDepState (..)
+  , PraosEvolvingStake (..)
   , PraosExtraFields (..)
   , PraosFields (..)
   , PraosParams (..)
+  , emptyPraosEvolvingStake
   , evolveKey
   , forgePraosFields
     -- * Tags
@@ -71,6 +73,7 @@ import           Cardano.Crypto.VRF.Mock (MockVRF)
 import           Cardano.Crypto.VRF.Simple (SimpleVRF)
 import           Cardano.Slotting.EpochInfo
 
+import           Data.Maybe (fromMaybe)
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Mock.Ledger.Stake
 import           Ouroboros.Consensus.NodeId (CoreNodeId (..))
@@ -90,7 +93,6 @@ data PraosFields c toSign = PraosFields {
   deriving (Generic)
 
 instance (PraosCrypto c, Typeable toSign) => NoThunks (PraosFields c toSign)
-  -- use generic instance
 
 -- | Fields that should be included in the signature
 data PraosExtraFields c = PraosExtraFields {
@@ -101,7 +103,6 @@ data PraosExtraFields c = PraosExtraFields {
   deriving (Generic)
 
 instance PraosCrypto c => NoThunks (PraosExtraFields c)
-  -- use generic instance
 
 data PraosValidateView c =
     forall signed. Cardano.Crypto.KES.Class.Signable (PraosKES c) signed
@@ -128,7 +129,6 @@ data HotKey c =
   deriving (Generic)
 
 instance PraosCrypto c => NoThunks (HotKey c)
-
 deriving instance PraosCrypto c => Show (HotKey c)
 
 -- | The 'HotKey' could not be evolved to the given 'Period'.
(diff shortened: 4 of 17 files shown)


Jasagredo:
Adresses #2850

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -78,31 +78,97 @@ import           Ouroboros.Consensus.Protocol.Abstract
...
+--
+-- A block as defined in Praos (section 3, definition 2, then extended in Fig 9)
+-- consist of a tuple:
+--
+-- > B = (slⱼ, st, d, B_πj, ρ, σⱼ)
+--
+-- where
+-- - slⱼ: the slot at which the block was generated.
+--
+-- - st: the current state.
```
nfrisby, line 91: Defn 2 from the paper also says `A state is a string st ∈ {0,1}^λ`. ... ah, reading farther along in that paragraph, I see that `st` is the hash of the previous block. That's probably a clarification to make here that'll avoid confusing future reader's confusion.

nfrisby: Also, let's call out the connections between these components and fields of `SimpleStdHeader` and/or `SimpleBody` and/or `SimplePraosExt`/`PraosFields` etc where we can. Ideally, these components and those fields will be one-to-one!

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -78,31 +78,97 @@ import           Ouroboros.Consensus.Protocol.Abstract
...
+--
+-- - d: data (transaction data in most cases).
+--
+-- - B_πj: block proof consisting of (Uᵢ, y, π).
+--      - y: a VRF output used to confirm tha Uᵢ was the slot leader.
```
nfrisby, line 96: tha -> that

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -78,31 +78,97 @@ import           Ouroboros.Consensus.Protocol.Abstract
...
+--      - π: the VRF proof of the above value.
+--
+--      > (y, π) ≜ VRF_evalProve(η, slⱼ, TEST) see Fig 9
+--
+-- - ρ: the block nonce consisting of (ρ_y, ρ_π), to capture entropy from the
+--      block forging process.
+--      - ρ_y: a VRF output used to confirm this block captured all the previous
```
nfrisby, line 103: I'm feeling clueless about rho_y; this is a part of the crypto details I've not attended to. Let's chat about it on a call.

I think this comment is useful, even without fully grokking it.

P.S. - Aha! I see (from your great comment below) in Fig. 10 that these rhos are ultimately used to build-up the eta. I'm familiar with eta, so now this suddenly makes a lot more sense to me.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -78,31 +78,97 @@ import           Ouroboros.Consensus.Protocol.Abstract
...
+-- - Tᵢ: the leader threshold for a specific stakeholder considering its
+--       relative stake (therefore depending on the slot). Defined in the Praos
+--       paper in Figure 4 using the definition for ϕ_f(αᵢ) from section 3.3.
+--
+--       > Tᵢ ≜ 2^(l_VRF) * pᵢ
+--       > pᵢ = ϕ_f(αᵢ) ≜ 1 - (1 - f)^(αᵢ)
+
 {-------------------------------------------------------------------------------
-  Fields required by Praos in the header
+  Fields required by Praos in the header of the block
```
nfrisby, line 134: Not a problem, but beware that "header" implicitly means "header of a block" throughout the code base, unless specified otherwise (eg in some lower-level Network module, likely).

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -78,31 +78,97 @@ import           Ouroboros.Consensus.Protocol.Abstract
...
 instance PraosCrypto c => NoThunks (PraosExtraFields c)
   -- use generic instance
 
+deriving instance PraosCrypto c => Show (PraosExtraFields c)
+deriving instance PraosCrypto c => Eq   (PraosExtraFields c)
+
+-- | A validate view is an association from the (@signed@) value to the
+-- @PraosFields@ that contains the signature that sign it.
+--
+-- In this case, PraosCrypto forces @signed@ to be @(Natural, SlotNo, VRFType)@ (why?)
```
nfrisby, line 171: Let's work through the `why?` question. I don't immediately know the answer -- we'll have to look some more.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -197,24 +275,26 @@ instance ToCBOR VRFType where
...
+-- | An error that can arise during validation
 data PraosValidationError c =
+      -- | The slot was invalid because it is not strictly increasing from the
+      -- previous block
       PraosInvalidSlot SlotNo SlotNo
-    | PraosUnknownCoreId CoreNodeId
-    | PraosInvalidSig String (VerKeyKES (PraosKES c)) Natural (SigKES (PraosKES c))
-    | PraosInvalidCert (VerKeyVRF (PraosVRF c)) (Natural, SlotNo, VRFType) Natural (CertVRF (PraosVRF c))
-    | PraosInsufficientStake Double Natural
+    | -- | The author of the block is unknown
```
nfrisby, line 290: Style: we put the `|` on the line with the constructor instead of on the line with the `-- |`. See https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md (CTRL-F for "-- | Baar".)

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -241,16 +321,22 @@ deriving instance PraosCrypto c => NoThunks (BlockInfo c)
   Protocol proper
 -------------------------------------------------------------------------------}
 
+-- | An uninhabited type for the Praos protocol.
```
nfrisby, line 324: Suggestion: change `for` to `representing` or to `identifying`.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -274,12 +360,63 @@ data instance ConsensusConfig (Praos c) = PraosConfig
   }
   deriving (Generic)
 
+instance PraosCrypto c => NoThunks (ConsensusConfig (Praos c))
+  -- use generic instance
```
nfrisby, line 364: I would prefer to not add comments like `-- use generic instance`. It's a slippery slope to try to explain Haskell semantics in the code's comments. I think `DeriveAnyClass` et al is something we should assume our audience understands.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -400,62 +547,15 @@ instance PraosCrypto c => ConsensusProtocol (Praos c) where
...
+-- | Probability for stakeholder Uᵢ to be elected in slot
+-- slⱼ considering its relative stake αᵢ.
 phi :: ConsensusConfig (Praos c) -> Rational -> Double
-phi PraosConfig{..} r = 1 - (1 - praosLeaderF) ** fromRational r
+phi PraosConfig{..} alpha = 1 - (1 - praosLeaderF) ** fromRational alpha
   where
     PraosParams{..} = praosParams
 
+-- | Compute Tᵢ for a given stakeholder @n@ at a @SlotNo@. Will be computed from
+-- @praosEvolvingStake@.
```
nfrisby, line 558: Let's change the `@` to `'`: `'praosEvolvingStake'`

nfrisby (changes requested):
Great job here!

These comments are wonderful and investing them into the mock Praos implementation in particular will be very helpful as a touchstone for understanding the consensus algorithm, as an approachable step in the learning.

I am Requesting Changes, mainly for some easy-to-misunderstand wordings I called out and because of my question about the exact semantics of the `Map EpochNo StakeDist`.

Maybe we can find a slot for a call tomorrow; or make some progress via Slack.

ouroboros-consensus-mock-test/test/Test/ThreadNet/Praos.hs
```diff
@@ -30,22 +31,43 @@ import           Test.ThreadNet.Util.NodeRestarts
...
-  , setupTestConfig   :: TestConfig
+  , setupK             :: SecurityParam
+  , setupNodeJoinPlan  :: NodeJoinPlan
+  , setupSlotLength    :: SlotLength
+  , setupTestConfig    :: TestConfig
+  , setupEvolvingStake :: Map EpochNo StakeDist
   }
   deriving (Show)
 
+-- TODO where should this go?
```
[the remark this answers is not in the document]
Jasagredo: Right, removed the TODO.

ouroboros-consensus-mock-test/test/Test/ThreadNet/Praos.hs
```diff
@@ -30,22 +31,43 @@ import           Test.ThreadNet.Util.NodeRestarts
...
+genEvolvingStake :: EpochSize -> TestConfig -> Gen (Map EpochNo StakeDist)
+genEvolvingStake epochSize TestConfig {numSlots, numCoreNodes = numCoreNodes@(NumCoreNodes x)} = do
```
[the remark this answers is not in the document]
Jasagredo: Good point. Applied.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/Block/Praos.hs
```diff
@@ -120,25 +119,11 @@ instance ( SimpleCrypto c
          , PraosCrypto c'
          , Signable (PraosKES c') (SignedSimplePraos c c')
          ) => LedgerSupportsProtocol (SimplePraosBlock c c') where
-  protocolLedgerView   cfg _  = pretendTicked $ stakeDist cfg
-  ledgerViewForecastAt cfg st = constantForecastOf
-                                 (pretendTicked $ stakeDist cfg)
+  protocolLedgerView   _ _  = TickedTrivial --pretendTicked $ stakeDist cfg
```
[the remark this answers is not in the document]
Jasagredo: Oops, a leftover.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -304,11 +322,13 @@ instance PraosCrypto c => ConsensusProtocol (Praos c) where
   tickChainDepState _ lv _ = TickedPraosChainDepState lv
 
   updateChainDepState cfg@PraosConfig{..}
-                      (PraosValidateView PraosFields{..} toSign)
+                      (PraosValidateView PraosFields{
+                          praosExtraFields = PraosExtraFields{..},
```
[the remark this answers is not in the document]
Jasagredo: Sounds right. I reverted it to how it was written before.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
```diff
@@ -443,10 +462,16 @@ leaderThreshold :: forall c. PraosCrypto c
                 -> SlotNo
                 -> CoreNodeId
                 -> Double
-leaderThreshold st xs s n =
-    let a = stakeWithDefault 0 n $ infosStake st xs (slotEpoch st s)
-    in  2 ^ (sizeHash (Proxy :: Proxy (PraosHash c)) * 8) * phi st a
-
+leaderThreshold config _blockInfos s n =
+    let
+      alpha = maybe 0 (stakeWithDefault 0 n)
```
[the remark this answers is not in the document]
Jasagredo: I created a new type and a function to query it. I think now it is better reflected the semantics in the comment.

ouroboros-consensus-test/ouroboros-consensus-test.cabal
```diff
@@ -116,6 +116,7 @@ library
                      , ouroboros-network
                      , ouroboros-network-framework
                      , ouroboros-consensus
+                     , ouroboros-consensus-mock
```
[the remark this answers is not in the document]
Jasagredo: I moved LeaderSchedule back into ouroboros-consensus and removed the dependency. I think it is quite reasonable to do this.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/UTxO.hs
```diff
@@ -106,6 +114,7 @@ data UtxoError
 instance Condense UtxoError where
   condense = show
 
+-- |Something that can generate transactions.
```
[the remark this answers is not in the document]
Jasagredo: True, changed.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/UTxO.hs
```diff
@@ -87,8 +92,11 @@ instance Condense Tx where
 type Ix     = Word
 type Amount = Word
 type TxId   = Hash MD5 Tx
+-- An input is a pair of the txhash and and index.
```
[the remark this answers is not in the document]
Jasagredo: I removed this one and some others in various places, I used these while going through the code to write assumptions and then check them.

Many comments were just stating the obvious but that helped me understand it better, although I agree that they should not remain there.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: I added lines specifying how they are named in the code, and clarified the `st` part. Good point! I actually had not map in my mind the state to the hash of the previous block.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: Removed, I used to deal with both Block Header and Transaction Header, that’s why I wrote that explicit description, but it seems redundant now.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: Right, I actually removed these comments. But thanks for the pointer to the style guide.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: Right, applied.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: This comment was already there, I just moved the instance, but I agree we can remove those.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: Applied.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: Fixed

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: I guess from this comment that it was okey then? :smile:

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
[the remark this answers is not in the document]
nfrisby: The new comment on `PraosEvolvingStake` doesn't make it obvious what the semantics of each association is.

I looked to these two code snippets to check it myself, and I think we could streamline them a bit.

```
stakeForNodeForEpoch :: PraosEvolvingStake -> CoreNodeId -> EpochNo -> Rational
stakeForNodeForEpoch (PraosEvolvingStake x) nid e =
  let thisStakeDist = fmap snd . Map.lookupMax . fst $ Map.split e x :: Maybe StakeDist
  in  fromMaybe 0 (thisStakeDist >>= (Map.lookup nid . stakeDistToMap))
```

```
      alpha = case slotEpoch config s of
        0 -> stakeWithDefault 0 n (praosInitialStake config)
        e -> stakeForNodeForEpoch (praosEvolvingStake config) n e
```

How about this?

```
...
--
-- Each pair stores the stakedist established by the end of the epoch.
-- See 'latestEvolvedStakeDistAsOfEpoch' for the intended interface.
newtype PraosEvolvingStake ...
```

```
-- | 'evolveStakeDistAsOfStartOfEpoch` provides that latest stakedist that had
-- evolved prior to the start of the given epoch
latestEvolvedStakeDistAsOfEpoch ::
  PraosEvolvingStake -> EpochNo -> Maybe StakeDist
latestEvolvedStakeDistAsOfEpoch (PraosEvolvingStake x) nid e =
    fmap snd . Map.lookupMax . fst $ Map.split e x
```

```
      alpha =
            stakeWithDefault 0 n
          $ fromMaybe (praosInitialStake config)
          $ latestEvolvedStakeDistAsOfEpoch (praosEvolvingStake config) e
```

Lastly, with the above semantics, it's OK for the `PraosEvolvingStake` mapping to include epoch 0 -- `fst . Map.split 0` returns a map that contains keys strictly less than `0`.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
nfrisby: There are lots and layers to the crypto types. I have trouble keeping them in my brain.

In this case, poking through it for the `n`-th time, I think this is the story:

  * In a value `PraosValidateView x y`,  `x` contains a KES signature and `y` is the data that was signed with the KES key to create that signature.

  * So `signed` here will be the parts of the block that are signed to create the signature in the block header -- that's what we use KES signatures for. See `Ouroboros.Consensus.Mock.Ledger.Block.Praos.SignedSimplePraos`. The consensus algorithm doesn't care what that is, so we use the existential here. For this mock implementation, we could probably have just arranged to import `SignedSimplePraos` here, but that'd be less flexible and wouldn't respect the divide we have between consensus protocol and block types (protocols don't depend on block details directly).
  
  * The `(Natural, SlotNo, VRFType)` type, on the other hand, is what is signed by the VRF key, not the KES key. The protocol itself does that, which is why that type's "VRF signability" constraint is baked into the `PraosCrypto` superclass.

nfrisby:
Thanks for Resolving all of my comment threads. I replied to the two remaining threads.

I'm Requesting Changes again because of the `PraosEvolvingStake` thread. I anticipate we can resolve it during the sync call tomorrow.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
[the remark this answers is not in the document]
Jasagredo: That seems a reasonable change, and quite orchestrated. I'm okey with it.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Protocol/Praos.hs
Jasagredo: I added a comment explaining what you just said above, that this could be simplified but from consensus point of view, it doesn't care what is the actual underlying type so we just use the existential. I think it is useful for someone who comes without prior knowledge in order to not puzzle around "there is only one type that fulfills this, why not simplify it?".

ouroboros-consensus-mock-test/test/Test/ThreadNet/Praos.hs
```diff
@@ -30,22 +32,44 @@ import           Test.ThreadNet.Util.NodeRestarts
...
+  , setupNodeJoinPlan  :: NodeJoinPlan
+  , setupSlotLength    :: SlotLength
+  , setupTestConfig    :: TestConfig
+  , setupEvolvingStake :: PraosEvolvingStake
   }
   deriving (Show)
 
+genEvolvingStake :: EpochSize -> TestConfig -> Gen PraosEvolvingStake
+genEvolvingStake epochSize TestConfig {numSlots, numCoreNodes} = do
+    chosenEpochs <- sublistOf [1..EpochNo $ max 1 maxEpochs - 1]
```
nfrisby, line 57: This list enumeration should start at `0` now, since `latestEvolvedStakeDistAsOfEpoch` (`Map.split` specifically) would drop a `(0, sd)` entry when applied at epoch `0`.

nfrisby (approved):
I made one more comment. I think it should be fixed, but I'm Approving anyway: we can fix it up later if you don't notice this.

Thanks for preparing this really nice PR!

ouroboros-consensus-mock-test/test/Test/ThreadNet/Praos.hs
Jasagredo: Indeed! fixed.

Jasagredo:
bors merge

PR #3079, opened 2021-04-21
Refactored DiffusionArguments for new root peers configuration

bolt12:


ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -444,19 +447,17 @@ localRootPeersProvider tracer
             Right results -> do
               atomically $ do
                 rootPeersGroups <- readTVar rootPeersGroupsVar
-                let rootPeersGroups' = IntMap.fromList $ zip [0,1..] rootPeersGroups
-                    (target, entry)  = rootPeersGroups' IntMap.! group
+                let (target, entry)  = rootPeersGroups `Seq.index` index
                     resultsMap       = Map.fromList (map fst results)
                     entry'           = entry <> resultsMap
```
coot, line 452: ^ shouldn't this be `resultsMap <> entry`, `Map.union` is left biased?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -319,8 +320,8 @@ newResolverResource resolvConf = go
 --
 
 data TraceLocalRootPeers =
-       TraceLocalRootDomains [(Int, Map RelayAddress PeerAdvertise)]
-       -- ^ 'Int' is the configured valency for the local producer groups
+     TraceLocalRootDomains [(Int, Map RelayAddress PeerAdvertise)]
+     -- ^ 'Int' is the configured valency for the local producer groups
```
coot, line 324: I think the original indentation was better.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: Ah! I think you're right, good catch! Thank you!

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -277,10 +275,9 @@ data DiffusionArguments = DiffusionArguments {
     , daPeerSelectionTargets :: PeerSelectionTargets
       -- ^ selection targets for the peer governor
 
-    , daStaticLocalRootPeers :: [(Socket.SockAddr, PeerAdvertise)]
-    , daLocalRootPeers       :: [(DomainAddress, PeerAdvertise)]
-    , daPublicRootPeers      :: [DomainAddress]
-    , daUseLedgerAfter       :: UseLedgerAfter
+    , daLocalRootPeersVar    :: StrictTVar m [(Int, Map RelayAddress PeerAdvertise)]
+    , daPublicRootPeersVar   :: StrictTVar m [RelayAddress]
+    , daUseLedgerAfterVar    :: StrictTVar m UseLedgerAfter
```
coot, line 280: Do we write to all three `TVar`s? Or all some of them read only as far as network is concerned?

coot: This patch is more than updating `ouroboros-consensus`, as indicated in the git message.

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/Instances.hs
```diff
@@ -81,6 +86,42 @@ instance Arbitrary PeerSelectionTargets where
...
+             <*> arbitrary
+
+instance Arbitrary RelayAddress where
+  arbitrary = do
+    isDomain <- arbitrary
+    if isDomain
+      then RelayDomain <$> arbitrary
+      else RelayAddress
+             <$> oneof [genIPv4, genIPv6]
+             <*> (fromIntegral <$> (arbitrary :: Gen Int))
```
coot, line 124: or
```
oneof [ RelayDomain <$> arbitrary
      , RelayAddresss <$> oneof [genIPv4, genIPv6]
                      <*> (fromIntegral @Int <$> arbitrary)
      ]
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
bolt12: We do write to all `TVar`s when one triggers the hot config update via SIGHUP

bolt12: I will update the commit and make it more descriptive!

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/Instances.hs
bolt12: Thanks!

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
coot: But isn't that done in `cardano-node`?

bolt12: Yes, and the Diffusion Layer provides the TVars to be updated by `cardano-node`

coot: it's the other way  around, 'DiffusionArguments` pass data from `cardano-node` to `ouroboros-network`.  The `TVar`'s must be created in `cardano-node` and in `ouroboros-network` we probably only need to read them.

bolt12: Ah, I see. So what happens is, only `daLocalRootPeersVar` is written by `ouroboro-network` the others are read-only I believe, but can change if over at `cardano-node` someone triggers local config update (i.e. `cardano-node` writes to them).

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/Instances.hs
```diff
@@ -81,6 +86,40 @@ instance Arbitrary PeerSelectionTargets where
...
+    DomainAddress . encodeUtf8
+      <$> elements domains
+      <*> (fromIntegral <$> (arbitrary :: Gen Int))
+    where
+      domains = [ "test1"
+                , "test2"
+                , "test3"
+                , "test4"
+                , "test5"
+                ]
```
coot, line 100: Are we using just five to make it more likely to draw the same name twice, i.e. when one is generating a list of `DomainAddress`-es?

coot (approved):
LGTM

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/Instances.hs
bolt12: I think that is a good reason, however I have to admit that I just followed what was done over at cardano-node (at `cardano-node/test/Test/Cardano/Node/Gen.hs`), but with less fun naming! 

 https://hackage.haskell.org/package/hedgehog-corpus-0.2.0/docs/src/Hedgehog.Corpus.html#cooking 😛

coot: Ha ha, that's a good one, I didn't know such a thing existed.

PR #3081, opened 2021-04-22
Add support for querying systemStart

newhoggy:


nfrisby:
@DavidEichmann What did we decide regarding "top-level" queries? This looks related to that aspect of your work. Would you comment on how to order your PR and this one relate? Thanks.

newhoggy:
I saw there was a PR to expose `ConfigLedger`, but I need `ConfigBlock` which is another field:

```
TopLevelConfig{
      topLevelConfigProtocol = coerce topLevelConfigProtocol
    , topLevelConfigLedger   = topLevelConfigLedger
    , topLevelConfigBlock    = coerce topLevelConfigBlock
    , topLevelConfigCodec    = coerce topLevelConfigCodec
    , topLevelConfigStorage  = coerce topLevelConfigStorage
    }
```

nfrisby:
>  I need `ConfigBlock` which is another field:

@newhoggy I pointed you at `ConfigBlock`, but (based on a some separate work I was doing) I see now that the `SystemStart` -- since it comes from the genesis block -- could be appropriate for other parts of the config too. (EG My other PR now has to add `SystemStart` to `ConsensusConfig` for Shelley.)

So I think your laser focus on `SystemStart` is for the best. But I think @DavidEichmann will have some insight/perspective on how to relate that to his plans for the query interface.

DavidEichmann:
It looks like you can derive the start time from the ledger config given that the block type is `ByronBlock` or a `HardForkBlock xs` where `xs` contains the byron are. If that's the case, then you could wait for me to add the new query for Ledger config.

To get the start time from the `LedgerConfig ByronBlock`:

```haskell
getSystemStart = SystemStart . gdStartTime . configGenesisData
```

newhoggy:
Linking to some relevant work:

* https://github.com/input-output-hk/ouroboros-network/tree/davide/query-ledger-config
* https://github.com/input-output-hk/ouroboros-network/tree/davide/top-level-query

newhoggy:
I've rebased this on top of `master`.

newhoggy:
The sync progress feature is actually quite useful.  Perhaps it's worth merging this PR so that this feature can be included earlier and then later on when the `LedgerConfig` stuff is merged we can switch to that implementation and rollback this API?

nfrisby:
(Copying what I had written in our Slack conversation.)

I agree that it's useful. But I'm hesitant to add something where it doesn't belong (ie to the HFC queries), even as a stop-gap. And especially as a stop-gap that would require special cases for backwards-compatible serialization routines once we eventually do fix up the implementation to use the more coherent solution (ie the top-level `blk`-independent layer of queries).

  * I think David's new PR #3106 is very close to mergeable. That'll unblock the more coherent solution. (It also unblocks the easier to maintain stop-gap of just having a relatively narrow `GetSystemStart` query before we add the full `GetLedgerConfig` query that supersedes it (ie David's work). Seems easy enough to carry both of those forward.)

  * This feature is a (very) nice-to-have, but not a necessity.

PR #3036, opened 2021-04-06
Provide slot-to-time conversion to ledger

nfrisby:
Fixes #3052.

Addresses CAD-2713. The ledger needs to be able to convert slots to UTC times, so as of this PR, the HFC now provides that to the ledger. See the commit message for more details.

PR #3087, opened 2021-04-23
Bump ledger to latest

nc6:


nfrisby:
Thanks, @nc6. I think PR #3036 supersedes this PR.

But it's good to know that the target commit of this PR wouldn't require any integration.

cc @EncodePanda

nc6:
We would like to issue a release of the node in the next few days, so if #3036 is not likely to merge in that time I would still like to merge this as an interim update.

nfrisby:
Closed in favor of PR #3036 (which just merged)

PR #3088, opened 2021-04-26
Refactor cddl specification

coot:
* per mini-protocol CDDL spec
* tx-submission2 CDDL spec
* keep-alive CDDL spec
* reimplement handshake spec: node-to-node & node-to-client
* update technical repor:
  - fixed a missplaced section
  - include CDDL mini-protocol encoding specs
  - standardise message and state indentifiers
  - link to haddock documentation

PR #3089, opened 2021-04-26
Fix some links to mini-protocol code in documentation

kantp:
A few of the links from the documentation to the code of mini protocols broke
during refactoring.


coot:
@kantp it's fixed in #3088. I also replaced links to the code with links to haddock.

Maybe the problem is that the link itself.  CI builds on every branch, and I suspect that the link in our `README`, points to the latest build, which might not be the latest version merged to master.

njd42 (approved):

coot (approved):
I resolved conflicts; one missing link found :dart:

coot:
bors cancel

PR #3091, opened 2021-04-26
Trace peer for chainsync clients

karknu:
This makes it possible to track which peer presented us which header at what time. Useful for debugging/performance analysis.


nfrisby (approved):
LGTM Happy to see this addition. Thanks!

nfrisby:
Here are some pointers about how to deal with the `stylish-haskell` objection in the buildkite CI.

https://github.com/input-output-hk/ouroboros-network/#formatting

PR #3093, opened 2021-04-27
new InboundGovernor metrics

bolt12:
This PR fixes `numberConns` metric name to a more precise one and adds `hotPeersRemote` metric


ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -100,13 +102,41 @@ inboundGovernor tracer serverControlChannel protocolIdleTimeout
...
+      chainSyncStatus <- atomically (getChainSyncStatus igsConnections)
+
+      let hotPeersRemote = length $ filter (== StatusRunning) chainSyncStatus
+
+      return $ InboundGovernorCounters { hotPeersRemote }
+      where
+        -- Due to circular dependencies it is not possible to import
+        -- Ouroboros.Network.NodeToNode (chainSyncMiniProtocolNum)
+        chainSyncMiniProtocolNum :: MiniProtocolNum
+        chainSyncMiniProtocolNum = MiniProtocolNum 2
```
karknu, line 118: Could you break out the protocol numbers into its own module, for example Ouroboros.Network.NodeToNode.MiniProtocolNums . ?

bolt12: The problem is that protocol numbers are defined at `ouroboros-network` repo and `ouroboros-network` depends on `ouroboros-network-framework` so I can not import that module

bolt12:
@karknu Optimized counter metrics tracing. Now counters are cached and whenever the state is updated for a given `connId` I compute the state that that remote peer has the local peer in. This way, instead of a big computation upfront, values are cached and compute on demand.

karknu (approved):
LGTM.

PR #3095, opened 2021-04-27
Update iohk-monitoring-framework to latest

newhoggy:


deepfire (approved):
The only dependency here is `contra-tracer`, which has seen no change -- so this is ~0 impact.

newhoggy:
bors merge

PR #3098, opened 2021-04-28
consensus: HFC now provides EpochInfo in Except monad

### LINKED ISSUE #2126: Replace `EpochInfo Identity` by `EpochInfo (Except PastHorizonException)`
The Shelley ledger as well as the `TPraos` implementation require an `EpochInfo Identity` in their config. Such a thing is really an impossibility: the `EpochInfo` changes as the ledger changes. The hard fork combinator alleviates this problem to some extent by introducing the concept of a partial config, and constructing the right `EpochInfo` on the fly. However, the `EpochInfo` that we pass to the ledger and to `TPraos` nonetheless has a limited range, and any attempt to use it beyond that range will result in a pure exception. It would be much better if the `EpochInfo` we construct as type `EpochInfo (Except PastHorizonException)`, so that an attempt to look past that horizon will result in an error we can more reliably catch. 

The changes consensus side are trivial (in fact, we have to do extra work to turn these exceptions into pure exceptions); the more laborious part is updating the `TPraos` and Shelley ledger implementations.

That said, this is mostly to aid debugging: any attempt to look past the horizon would be a bug.

nfrisby:
Fixes #2126. Also resolves the first draft of CAD-2713 on Jira.

This PR is a single commit; see the message.

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -28,7 +28,7 @@ module Ouroboros.Consensus.Cardano.CanHardFork (
   ) where
 
 import           Control.Monad
-import           Control.Monad.Except (Except, runExcept, throwError)
+import           Control.Monad.Except (runExcept, throwError)
 import qualified Data.Map.Strict as Map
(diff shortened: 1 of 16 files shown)


nfrisby:
@nc6 @polinavino -- FYI this PR will provide the `EpochInfo` needed for the Alonzo validity interval translation.

Jasagredo (approved):
This looks good to me.

PR #3106, opened 2021-04-29
Wrap all Node-To-Client queries in a top level query type

DavidEichmann:
This is just a refactoring in preparation for adding a new query: `GetPartialLedgerConfig :: Query blk (PartialLedgerConfig blk)` (I've got a WIP branch where I do that here: [davide/query-ledger-config](https://github.com/input-output-hk/ouroboros-network/tree/davide/query-ledger-config)). You may want to review this commit by commit. The first 2 commits are simple renamings

### DIFF
--- a/ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
+++ b/ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
@@ -378,7 +378,7 @@ instance CardanoHardForkConstraints c
 
 instance c ~ MockCryptoCompatByron
       => Arbitrary (WithVersion (HardForkNodeToClientVersion (CardanoEras c))
-                                (SomeSecond Query (CardanoBlock c))) where
+                                (SomeSecond BlockQuery (CardanoBlock c))) where
   arbitrary = frequency
       [ (1, arbitraryNodeToClient injByron injShelley injAllegra injMary)
       , (1, WithVersion
--- a/ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
+++ b/ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
@@ -105,9 +105,9 @@ instance Arbitrary (BlockNodeToNodeVersion blk)
 
 arbitraryNodeToNode
   :: ( Arbitrary (WithVersion ByronNodeToNodeVersion byron)
-     , Arbitrary shelley
-     , Arbitrary allegra
-     , Arbitrary mary
+     , Arbitrary (WithVersion ShelleyNodeToNodeVersion shelley)
+     , Arbitrary (WithVersion ShelleyNodeToNodeVersion allegra)
+     , Arbitrary (WithVersion ShelleyNodeToNodeVersion mary)
      )
   => (byron   -> cardano)
   -> (shelley -> cardano)
--- a/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/ByronCompatibility.hs
+++ b/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/ByronCompatibility.hs
@@ -136,11 +136,11 @@ toCardanoCodecConfig codecConfigByron =
 --
 -- Note that ledger state and all other types stored as part of the ledger
 -- snapshot are __not__ forwards compatible.
-newtype ByronToCardano                       = B2C        { unB2C        ::         ByronBlock   } deriving (Eq, Show)
-newtype instance Header ByronToCardano       = HeaderB2C  { unHeaderB2C  :: Header  ByronBlock   } deriving (Eq, Show)
-newtype instance GenTx ByronToCardano        = GenTxB2C   { unGenTxB2C   :: GenTx   ByronBlock   } deriving (Eq, Show)
-newtype instance TxId (GenTx ByronToCardano) = GenTxIdB2C { unGenTxIdB2C :: GenTxId ByronBlock   } deriving (Eq, Show)
-newtype instance Query ByronToCardano a      = QueryB2C   { unQueryB2C   :: Query   ByronBlock a } deriving (Eq, Show)
+newtype ByronToCardano                       = B2C        { unB2C        ::            ByronBlock   } deriving (Eq, Show)
+newtype instance Header ByronToCardano       = HeaderB2C  { unHeaderB2C  :: Header     ByronBlock   } deriving (Eq, Show)
+newtype instance GenTx ByronToCardano        = GenTxB2C   { unGenTxB2C   :: GenTx      ByronBlock   } deriving (Eq, Show)
+newtype instance TxId (GenTx ByronToCardano) = GenTxIdB2C { unGenTxIdB2C :: GenTxId    ByronBlock   } deriving (Eq, Show)
+newtype instance BlockQuery ByronToCardano a = QueryB2C   { unQueryB2C   :: BlockQuery ByronBlock a } deriving (Eq, Show)
 
 newtype instance NestedCtxt_ ByronToCardano f a where
   NestedCtxt_B2C :: NestedCtxt_ ByronBlock     f a
@@ -173,10 +173,10 @@ instance HasNestedContent Header ByronToCardano where
   nest (DepPair ctxt a) =
       HeaderB2C $ nest (DepPair (mapNestedCtxt unNestedCtxt_B2C ctxt) a)
 
-instance ShowQuery (Query ByronToCardano) where
+instance ShowQuery (BlockQuery ByronToCardano) where
   showResult (QueryB2C query) = showResult query
 
-instance SameDepIndex (Query ByronToCardano) where
+instance SameDepIndex (BlockQuery ByronToCardano) where
   sameDepIndex (QueryB2C q1) (QueryB2C q2) = sameDepIndex q1 q2
 
 {------------------------------------------------------------------------------
@@ -354,18 +354,18 @@ instance SerialiseNodeToClient ByronToCardano CC.ApplyMempoolPayloadErr where
   encodeNodeToClient = encodeNodeToClientB2C (Proxy @WrapApplyTxErr) id
   decodeNodeToClient = decodeNodeToClientB2C (Proxy @WrapApplyTxErr) (\(ApplyTxErrByron err) -> err)
 
-instance SerialiseNodeToClient ByronToCardano (SomeSecond Query ByronToCardano) where
+instance SerialiseNodeToClient ByronToCardano (SomeSecond BlockQuery ByronToCardano) where
   encodeNodeToClient = encodeNodeToClientB2C
-                         (Proxy @(SomeSecond Query))
+                         (Proxy @(SomeSecond BlockQuery))
                          (\(SomeSecond q) -> SomeSecond (unQueryB2C q))
   decodeNodeToClient = decodeNodeToClientB2C
-                         (Proxy @(SomeSecond Query))
+                         (Proxy @(SomeSecond BlockQuery))
                          (\(SomeSecond (QueryIfCurrentByron q)) -> SomeSecond (QueryB2C q))
 
-instance SerialiseResult ByronToCardano (Query ByronToCardano) where
+instance SerialiseResult ByronToCardano (BlockQuery ByronToCardano) where
   encodeResult (CodecConfigB2C ccfg) () (QueryB2C q) r =
       encodeResult ccfg byronNodeToClientVersion q r
-  decodeResult (CodecConfigB2C ccfg) () (QueryB2C (q :: Query ByronBlock result)) =
+  decodeResult (CodecConfigB2C ccfg) () (QueryB2C (q :: BlockQuery ByronBlock result)) =
       (\(QueryResultSuccess r) -> r) <$>
         decodeResult
           (toCardanoCodecConfig ccfg)
@@ -399,7 +399,7 @@ instance Arbitrary (GenTx ByronToCardano) where
 instance Arbitrary (GenTxId ByronToCardano) where
   arbitrary = GenTxIdB2C <$> arbitrary
 
-instance Arbitrary (SomeSecond Query ByronToCardano) where
+instance Arbitrary (SomeSecond BlockQuery ByronToCardano) where
   arbitrary = (\(SomeSecond q) -> SomeSecond (QueryB2C q)) <$> arbitrary
 
 instance Arbitrary (SomeResult ByronToCardano) where
@@ -417,11 +417,11 @@ instance Arbitrary (SomeResult ByronToCardano) where
 --
 -- Note that ledger state and all other types stored as part of the ledger
 -- snapshot are __not__ forwards compatible.
-newtype CardanoToByron                       = C2B        { unC2B        ::         ByronBlock   } deriving (Eq, Show)
-newtype instance Header CardanoToByron       = HeaderC2B  { unHeaderC2B  :: Header  ByronBlock   } deriving (Eq, Show)
-newtype instance GenTx CardanoToByron        = GenTxC2B   { unGenTxC2B   :: GenTx   ByronBlock   } deriving (Eq, Show)
-newtype instance TxId (GenTx CardanoToByron) = GenTxIdC2B { unGenTxIdC2B :: GenTxId ByronBlock   } deriving (Eq, Show)
-newtype instance Query CardanoToByron a      = QueryC2B   { unQueryC2B   :: Query   ByronBlock a } deriving (Eq, Show)
+newtype CardanoToByron                       = C2B        { unC2B        ::            ByronBlock   } deriving (Eq, Show)
+newtype instance Header CardanoToByron       = HeaderC2B  { unHeaderC2B  :: Header     ByronBlock   } deriving (Eq, Show)
+newtype instance GenTx CardanoToByron        = GenTxC2B   { unGenTxC2B   :: GenTx      ByronBlock   } deriving (Eq, Show)
+newtype instance TxId (GenTx CardanoToByron) = GenTxIdC2B { unGenTxIdC2B :: GenTxId    ByronBlock   } deriving (Eq, Show)
+newtype instance BlockQuery CardanoToByron a = QueryC2B   { unQueryC2B   :: BlockQuery ByronBlock a } deriving (Eq, Show)
 
 newtype instance NestedCtxt_ CardanoToByron f a where
   NestedCtxt_C2B :: NestedCtxt_ ByronBlock     f a
@@ -454,10 +454,10 @@ instance HasNestedContent Header CardanoToByron where
   nest (DepPair ctxt a) =
       HeaderC2B $ nest (DepPair (mapNestedCtxt unNestedCtxt_C2B ctxt) a)
 
-instance ShowQuery (Query CardanoToByron) where
+instance ShowQuery (BlockQuery CardanoToByron) where
   showResult (QueryC2B query) = showResult query
 
-instance SameDepIndex (Query CardanoToByron) where
+instance SameDepIndex (BlockQuery CardanoToByron) where
   sameDepIndex (QueryC2B q1) (QueryC2B q2) = sameDepIndex q1 q2
 
 {------------------------------------------------------------------------------
@@ -637,17 +637,17 @@ instance SerialiseNodeToClient CardanoToByron CC.ApplyMempoolPayloadErr where
   encodeNodeToClient = encodeNodeToClientC2B (Proxy @WrapApplyTxErr) ApplyTxErrByron
   decodeNodeToClient = decodeNodeToClientC2B (Proxy @WrapApplyTxErr) id
 
-instance SerialiseNodeToClient CardanoToByron (SomeSecond Query CardanoToByron) where
+instance SerialiseNodeToClient CardanoToByron (SomeSecond BlockQuery CardanoToByron) where
   encodeNodeToClient =
       encodeNodeToClientC2B
-        (Proxy @(SomeSecond Query))
+        (Proxy @(SomeSecond BlockQuery))
         (\(SomeSecond q) -> SomeSecond (QueryIfCurrentByron (unQueryC2B q)))
   decodeNodeToClient =
       decodeNodeToClientC2B
-        (Proxy @(SomeSecond Query))
+        (Proxy @(SomeSecond BlockQuery))
         (\(SomeSecond q) -> SomeSecond (QueryC2B q))
 
-instance SerialiseResult CardanoToByron (Query CardanoToByron) where
+instance SerialiseResult CardanoToByron (BlockQuery CardanoToByron) where
   encodeResult (CodecConfigC2B ccfg) () (QueryC2B q) (r :: result) =
       encodeResult
         (toCardanoCodecConfig ccfg)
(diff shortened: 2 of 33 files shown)


ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/ByronCompatibility.hs
```diff
@@ -421,7 +421,7 @@ newtype CardanoToByron                       = C2B        { unC2B        ::
 newtype instance Header CardanoToByron       = HeaderC2B  { unHeaderC2B  :: Header  ByronBlock   } deriving (Eq, Show)
 newtype instance GenTx CardanoToByron        = GenTxC2B   { unGenTxC2B   :: GenTx   ByronBlock   } deriving (Eq, Show)
 newtype instance TxId (GenTx CardanoToByron) = GenTxIdC2B { unGenTxIdC2B :: GenTxId ByronBlock   } deriving (Eq, Show)
-newtype instance Query CardanoToByron a      = QueryC2B   { unQueryC2B   :: Query   ByronBlock a } deriving (Eq, Show)
```
nfrisby, line 424: We'll have to check for alignment being busted like it is here.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Degenerate.hs
```diff
@@ -10,9 +10,10 @@
 
 {-# OPTIONS_GHC -Wno-orphans #-}
 
-module Ouroboros.Consensus.HardFork.Combinator.Degenerate (
```
nfrisby, line 13: Our `StyleGuide.md` is to leave the paren where it was.

(IIUC `stylish-haskell` on the CI will enforce this, FYI.)

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
```diff
@@ -39,6 +39,9 @@ data NodeToClientVersion
     -- ^ enabled @CardanoNodeToClientVersion6@, adding a query
     | NodeToClientV_8
     -- ^ 'LocalStateQuery' protocol codec change, allows to acquire tip point.
+    | NodeToClientV_9
+    -- ^ 'LocalStateQuery' protocol codec change, queries are now
```
nfrisby, line 43: Depending on where our merges fall between node releases, we may be able to add the new top-level queries to `NodeToClientV_9` as well.

nfrisby (changes requested):
Thanks so much for separating these commits nicely. Made my first pass a lot easier.

Most of my comments are style stuff.

There is at least one crucial comment: about backwards-compatible serialization.

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
coot: See also my comment [here](https://github.com/input-output-hk/ouroboros-network/pull/2975#discussion_r623812670).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -16,27 +22,111 @@ import           Data.Maybe (isJust)
...
+          | otherwise -> decodeResult codecConfig blockNodeToClientVersion blockQuery
+
+isQuerySerialisationSupported :: (SupportedNetworkProtocolVersion blk) => Proxy blk -> BlockNodeToClientVersion blk -> Bool
+isQuerySerialisationSupported blkP blockN2CV = any (>= NodeToClientV_9) supportedN2CVs
+  where
+    supportedN2CVs =
+      [ node2CV
+      | (node2CV, blockN2CV') <- Map.toList (supportedNodeToClientVersions blkP)
+      , blockN2CV == blockN2CV'
+      ]
```
DavidEichmann, line 104: @nfrisby I've added this version check, but things are a bit awkeward as the check depends on the `NodeToClientVersion` rather than the `BlockNodeToClientVersion blk`, so I've ended up having to add a `SupportedNetworkProtocolVersion blk` constraint in various places (this is all done in the latest commit). Any thoughts on this or if there is a better way of doing this?

DavidEichmann:
Latest update fixes a few test failures.

DavidEichmann:
Oooops, lost a change due to a local rebase. Latest update fixes a few test failures (for real this time :crossed_fingers:).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
DavidEichmann: This is resolved in the latest version.

DavidEichmann:
@nfrisby the version checks when serialising `Query` were a bit awkward because they required the `NodeToClientVersion` as opposed to just the `BlockNodeToClientVersion blk` available in `SerialiseNodeToClient`. My solution was just to use a custom function taking the `NodeToClientVersion` for decoding/encoding the `Query` instead of providing a `instance SerialiseNodeToClient blk (SomeSecond Query blk)`. You can see this in the diff linked below.

It's probably best you view the [diff since your last review](https://github.com/input-output-hk/ouroboros-network/compare/0bae6e25aefa79fb433c314356d554d369813244..dc24c93e5007245027638b10c49f0492a1d437ca). This will skip over some changes I've reverted regarding [this comment](https://github.com/input-output-hk/ouroboros-network/pull/3106#discussion_r625066873).

DavidEichmann:
I'm waiting for review. I don't have access to buildkite, so don't know why it's failing on CI.

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
```diff
@@ -111,8 +112,8 @@ roundtrip_all
      , SerialiseNodeToNodeConstraints   blk
      , SerialiseNodeToClientConstraints blk
 
-     , Show (BlockNodeToNodeVersion   blk)
-     , Show (BlockNodeToClientVersion blk)
+    , Show (BlockNodeToNodeVersion   blk)
```
nfrisby, line 115: Drop this undent please.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -190,10 +190,10 @@ defaultCodecs ccfg version networkVersion = Codecs {
           (networkVersion >= NodeToClientV_8)
```
nfrisby, line 190: Perhaps `codecLocalStateQuery` takes a boolean instead of the version for the same reasons you had mentioned about the awkwardness of importing the right types into the encoder's definition?

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
```diff
@@ -303,7 +303,7 @@ nullTracers = Tracers {
 showTracers :: ( Show peer
                , Show (GenTx blk)
                , Show (ApplyTxErr blk)
-               , ShowQuery (Query blk)
```
nfrisby, line 306: Hmm. This seems a bit overealous, since line 280 above still has:

```
tStateQueryTracer   :: f (TraceLabelPeer peer (TraceSendRecv (LocalStateQuery blk (Point blk) (Query blk))))
```

We probably want to leave the constraints on `Query` eg so that when we do add new top-level-only-queries this constraint need not change. Does that make sense to you?

I'm guessing this is the result of a find-and-replace etc, so there are probably other signatures and eg other classes that would require the same treatment.

nfrisby: IIUC, John Ky has written some of these instances already on his PR #3110

nfrisby:
The build kite failures on 072bdc40c6 are due to `stylish-haskell.`

https://github.com/input-output-hk/ouroboros-network/#formatting lists commands for you to run `stylish-haskell` localling -- be advised: it'll make the changes in place, IIUC.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
DavidEichmann: I think I did that originally but it results in a warning:

```
src/Ouroboros/Consensus/Network/NodeToClient.hs:303:16: warning: [-Wsimplifiable-class-constraints]
    • The constraint ‘ShowQuery (Query blk)’ matches
        instance ShowQuery (BlockQuery blk) => ShowQuery (Query blk)
          -- Defined at src/Ouroboros/Consensus/Ledger/Query.hs:50:10
      This makes type inference for inner bindings fragile;
        either use MonoLocalBinds, or simplify it using the instance
    • In the type signature:
        showTracers :: (Show peer,
                        Show (GenTx blk),
                        Show (ApplyTxErr blk),
                        ShowQuery (Query blk),
                        HasHeader blk) =>
                       Tracer m String -> Tracers m peer blk e
    |
303 | showTracers :: ( Show peer
    |                ^^^^^^^^^^^...
```

I could suppress the warning, but it seems simpler to just wait till the new top level query is added. At that point I think GHC will remind me to adjust these constraints. How does that sound?

nfrisby: Ah, I see! Harumph.

Yes, I think the extra git commit churn on these lines is probably the right choice instead of adding the warning suppressor to the whole module. (If pragmas could be specified on a per top-level declaration basis I might make the other choice :/)

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
nfrisby: I was suggesting you could change your encoder and decoder to take the `isTopLevelQueryIntroduced` boolean argument, comparable to the `canAcquireTip` flag that `codecLocalStateQuery` already takes.

Maybe that "cuts the knot" with respect to the types; avoiding the awkwardness.

nfrisby: I'll admit I haven't studied the details of the awkwardness you mentioned. I'll do that now.

DavidEichmann: I've been thinking a bit about this. It seems key that there are 2 separate layers of code here: the network layer and the consensus layer. `NodeToClientVersion` corresponds to the network layer version. `BlockNodeToClientVersion blk` corresponds to the consensus layer version.

What I've implicitly done in this PR is assume that the addition of a top level query, `Query blk`, is a change to the network layer. Hence I've added a new `NodeToClientVersion` (version 9). If that's correct, then I can continue on with this. I'd suggest just passing the `NodeToClientVersion` to `codecLocalStateQuery` and doing the version check for both `Query blk` and for  `canAcquireTip` based on that.

But is this assumption really correct? Perhaps the addition of `Query blk` is actually a consensus layer change. Indeed the network layer has always been polymorphic over the query type. If the change is part of the network layer, then that would mean I should add a new version to *all* `BlockNodeToClientVersion blk`s instead of just to `NodeToClientVersion`. Then I'd need to touch all thoughs types again every time I add a new constructor to `Query blk`.

Any thoughts on this would be helpful.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -16,27 +26,142 @@ import           Data.Maybe (isJust)
...
+  -> Query blk result
+  -> result
+  -> Encoding
+queryEncodeResult codecConfig version blockNodeToClientVersion query result
+  | not (version >= NodeToClientV_9)
+  = error $ "encode failure: Serializing Query result is not supported (on node to client version: " ++ show version ++ ")"
+  | otherwise
+  = case query of
+      BlockQuery blockQuery ->
+        encodeTag 0
```
nfrisby, line 120: My other comment suggests we put this logic off to the next PR, after merging this one. That said, I don't think the results will ever need an additional tag; the query is argument to `encodeResult` only so that we know which type we're serializing.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -16,27 +26,142 @@ import           Data.Maybe (isJust)
...
+  -> forall s. Decoder s result
+queryDecodeResult codecConfig version blockNodeToClientVersion query
+  | not (version >= NodeToClientV_9)
+  = error $ "encode failure: Deserializing Query result is not supported (on node to client version: " ++ show version ++ ")"
+  | otherwise
+  = do
+    tag <- decodeTag
+    case query of
+      BlockQuery blockQuery
+        | tag /= 0 -> fail $ "Query blk: Expected tag 0 but got " ++ show tag
```
nfrisby, line 138: Same as my comment on `queryEncodeResult`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -16,27 +26,142 @@ import           Data.Maybe (isJust)
...
+queryDecodeResult ::
+     forall blk result. SerialiseResult blk (BlockQuery blk)
+  => CodecConfig blk
+  -> NodeToClientVersion
+  -> BlockNodeToClientVersion blk
+  -> Query blk result
+  -> forall s. Decoder s result
+queryDecodeResult codecConfig version blockNodeToClientVersion query
+  | not (version >= NodeToClientV_9)
+  = error $ "encode failure: Deserializing Query result is not supported (on node to client version: " ++ show version ++ ")"
```
nfrisby, line 132: Same as my comment on `queryEncodeResult`; however, we will need backward compatibility, in general, not just `error` when the negotiated version is less than `V_9` (or whatever the threshold ends up being). We will still need to be able to communicate with "older" nodes.

ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
nfrisby: - See my comment on `queryEncodeResult`; we don't _need_ to worry about serialization at all in this PR. It's up to you whether we do or not, though I would prefer to defer it for this PR so we can work out the serialization details on a narrower PR.
- The semantics of `NodeToClientVersion` and `BlockNodeToClientVersion` accumulate in the opposite direction of the dependency: the `NodeToClientVersion` characterizes the behavior of the whole node (not just the network layer), whereas `BlockNodeToClientVersion` characterizes the behavior of a part of the Consensus part of the node. So if the Consensus part changes, then that means the whole node's behavior changes. (Somewhat tangential, but potentially clarifying remark: the network layer is responsible for negotiating the `NodeToClientVersion` with the peer).
- I agree that "the addition of Query blk is actually a consensus layer change", but because of the above point, that still means we'll need to bump `NodeToClientVersion`. We have to whenever we alter the set of possible queries and responses. Here's a historical example of that:

https://github.com/input-output-hk/ouroboros-network/blob/7c8fac89c8ce4c4f9e5d891c7fc1be6613af3750/ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs#L32-L33

- It helps me to think of `NodeToClientVersion` (and its counterpart `NodeToNodeVersion`) as merely the monotonically-increasing natural number that every pair of peers negotiates when connecting. That's its fundamental role and why it needs to reflect every breaking change.
- The evolution of `BlockNodeToClientVersion` versions then is in some sense just a subsequence (not substring) of the evolution of `NodeToClientVersion` versions: its counter corresponds to when the negotiated thing had to change in ways that (only?) `blk`-sensitive Consensus code needs to consider (eg for backwards-compatibility). (That's the number, like 6 in `CardanoNodeToClientVersion6` -- but the RHS of `CardanoNodeToClientVersion6` breaks it down even farther regarding which parts of Consensus are affected).

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
```diff
@@ -323,10 +326,13 @@ roundtrip_SerialiseNodeToClient
...
 roundtrip_SerialiseNodeToClient ccfg =
-    [ rt (Proxy @blk)                    "blk"
-    , rt (Proxy @(GenTx blk))            "GenTx"
-    , rt (Proxy @(ApplyTxErr blk))       "ApplyTxErr"
-    , rt (Proxy @(SomeSecond Query blk)) "Query"
+    [ rt (Proxy @blk)                         "blk"
+    , rt (Proxy @(GenTx blk))                 "GenTx"
+    , rt (Proxy @(ApplyTxErr blk))            "ApplyTxErr"
+    , rt (Proxy @(SomeSecond BlockQuery blk)) "BlockQuery"
+    , rtWith (Proxy @(SomeSecond Query blk))  "Query"
```
nfrisby, line 333: `rtWith` seems fine, but so does the pattern in the existing tests. Please make it uniform, either by inlining `rtWith` here and discarding the binding or else change the exists version-sensitive tests to also use `rtWith`.

Edit: The `roundtrip Result` test might not fit with the `rtWith` mold.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Node/Serialisation.hs
```diff
@@ -6,6 +6,7 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE StandaloneDeriving    #-}
 {-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE UndecidableInstances  #-}
```
nfrisby, line 9: I'm not anticipating why we'd need to add UndecidableInstances; new `BlockQuery` and old `Query` are the same thing. Is it really needed here?

Same question applies in `ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Serialisation/SerialiseNodeToClient.hs`

ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
```diff
@@ -406,6 +407,14 @@ instance c ~ MockCryptoCompatByron
       injAnytimeMary    (Some      query)  = SomeSecond (QueryAnytimeMary      query)
       injHardFork       (Some      query)  = SomeSecond (QueryHardFork         query)
 
+instance Arbitrary (WithVersion (HardForkNodeToClientVersion (CardanoEras c))
```
nfrisby, line 410: Can we have one `Arbitrary (WithVersion blk (SomeSecond BlockQuery blk)) => Arbitrary (WithVersion blk (SomeSecond Query blk))` instance? I anticipate it belongs somewhere in `ouroboros-consensus-test`, probably next to the `Arbitrary (SomeSecond BlockQuery blk)=> Arbitrary (SomeSecond Query blk)` instance.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
```diff
@@ -175,8 +176,14 @@ instance {-# OVERLAPPABLE #-} Arbitrary a
...
 -- 'ShelleyNodeToClientVersion1'.
 instance CanMock era
-      => Arbitrary (WithVersion ShelleyNodeToClientVersion (SomeSecond Query (ShelleyBlock era))) where
+      => Arbitrary (WithVersion ShelleyNodeToClientVersion (SomeSecond BlockQuery (ShelleyBlock era))) where
   arbitrary = do
       query@(SomeSecond q) <- arbitrary
       version <- arbitrary `suchThat` querySupportedVersion q
       return $ WithVersion version query
+
+instance Arbitrary (WithVersion ShelleyNodeToClientVersion (SomeSecond BlockQuery (ShelleyBlock era)))
```
nfrisby, line 185: Same question as in my comment on the corresponding instance in `ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs`

nfrisby (changes requested):
I'm Requesting Changes because I don't think the serializers as-is do backwards-compatibility correctly. I made some comments on that, but I also pointed out the higher-level option of simply deferring the serialization changes until the next PR.

I still have to think very carefully each time I consider the relationship between `NodeTo*Version` and `BlockNodeTo*Version`. But I wrote a big comment here that I hope can help clarify their semantics and roles for you.

I hope it makes more sense now, but if not (or even if so), please feel free to join the Consensus planning meeting tomorrow (see eg my calendar) and we can talk about it there.

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
nfrisby: > IIUC, nothing needs to change here yet.

I think I came to that same conclusion, eg here https://github.com/input-output-hk/ouroboros-network/pull/3106#discussion_r629762791

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -30,6 +37,39 @@ import           Ouroboros.Consensus.Util.DepPair
 -- by the result type.
 data Query blk result = BlockQuery (BlockQuery blk result)
 
+instance (ShowProxy (BlockQuery blk)) => ShowProxy (Query blk) where
+  showProxy (Proxy :: Proxy (Query blk)) = "Query (" ++ showProxy (Proxy @(BlockQuery blk)) ++ ")"
+
+instance (ShowQuery (BlockQuery blk)) => ShowQuery (Query blk) where
+  showResult (BlockQuery blockQuery) = showResult blockQuery
+
+instance SerialiseNodeToClient blk (SomeSecond BlockQuery blk) => SerialiseNodeToClient blk (SomeSecond Query blk) where
```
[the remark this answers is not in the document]
DavidEichmann: I assume this is no longer applicable.

ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/Generators.hs
nfrisby: Hmmm. I find overlapping instances tend to add more confusion than they're worth. In general, I prefer explicitness for the sake of transparency.

In this case, I'm hesitant to make the existing overlapping instance _more_ general. I'm guessing the earlier logic was that there are only two `ShelleyNodeToClientVersions` and the difference between them was the addition of one query. So most of the instances were trivial, hence the overlapping was used for that.

But with the new more general instance, this would apply to _any_ version type, even if it were rich. That's not a risk right now, since we already have instances for all the other types (since there was no such general instance), but this opens the door for us to accidentally rely on this general instance for a future version type.

However, that said, this is for `Arbitrary`. So the greatest risk is that we'd see test failures for mismatched versions and values. Once we debug the failure to see the cause, we can add the corresponding override instance. So I'm optimistic here. Let's go with what you have here.

nfrisby (approved):
LGTM! Looking forward to the new queries.

Please rebase on `origin/master` before telling `bors` to merge.

DavidEichmann:
bors merge

DavidEichmann:
@nfrisby Thanks for keeping up the review :-D I know it's been a long process.

nfrisby:
> Thanks for keeping up the review :-D I know it's been a long process.

Not at all! It's a non-trivial codebase, and we're spinning multiple plates :)

PR #3110, opened 2021-05-04
Functions to Map query

newhoggy:


newhoggy:
New the functions are needed to wrap and unwrap `BlockQuery`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -39,7 +39,9 @@ import qualified Data.Map.Strict as Map
 
 -- | Different queries supported by the ledger for all block types, indexed
 -- by the result type.
-data Query blk result = BlockQuery (BlockQuery blk result)
+newtype Query blk result = BlockQuery
+  { unQuery :: BlockQuery blk result
+  }
```
DavidEichmann, line 44: Changing to a newtype is not ideal since this will soon be changed to

```haskell
data Query blk result where
    BlockQuery :: BlockQuert blk result -> Query blk result
    GetPartialLedgerConfig :: Query blk (PartialLedgerConfig blk)
```

In which case `unQuery` would be less useful. It seems you're not using the `unQuery` projection function anywhere. Do you plan on using it?

DavidEichmann (commented):
I assume this is blocked on #3106. I've just looked at the final commit and besides one comment, all seems pretty cut and dry :+1:

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: I had expected to need it, but turns out not.  Will remove.

DavidEichmann:
Do these changes actually need to be based on my branch? Perhaps you can just rebase on master and get this merged earlier if you're keen.

nfrisby:
> Do these changes actually need to be based on my branch? Perhaps you can just rebase on master and get this merged earlier if you're keen.

IIUC, John's use case ultimately depends on your branch, so that wouldn't unblock him.

PR #3116, opened 2021-05-06
Make the mempool pure

### LINKED ISSUE #2591: Make the Mempool pure
We designed the Mempool with a monadic API, but it has now become clear that the Mempool could just be a pure value stored in a `TVar`. This will simplify its implementation and will make it easier to test.

Jasagredo:
Closes #2591

Splits the operations on the mempool into a separate module that implements them in a pure style.

The `Ouroboros.Consensus.Mempool.Impl` module was split into `Impl`, `Impl.Pure` and
`Impl.Types` where:
- `Impl`: contains the monadic side of the implementation, bridging between
  operations on an `InternalState` and the actual `Mempool` API.
- `Impl.Pure`: contains the pure side of the implementation, functions that
  update an `InternalState` according to the rules described on the `Mempool`
  API.
- `Impl.Types`: defines the `InternalState` used to fulfill the requirements of
  the implementation, and the `ValidationResult` used to extend the mempool. It
  is used by both `Impl` and `Impl.Pure`.

From outside of this module, the change should be transparent, as other modules
should only access the mempool via the API, and all relevant items are
re-exported through `Ouroboros.Consensus.Mempool`.

The tests are testing the public operations on the Mempool, so it was decided to keep
them as they are. A couple of tests were added to check that the mutable operations on
the mempool that take inputs (tryAddTxs and removeTxs) are "distributive" (if only one thread
performs them):

```
tryAddTxs ( x : y : ...) == tryAddTxs x >> tryAddTxs y >> ...
```

### DIFF
--- a/ouroboros-consensus-test/test-consensus/Test/Consensus/Mempool.hs
+++ b/ouroboros-consensus-test/test-consensus/Test/Consensus/Mempool.hs
@@ -65,7 +65,7 @@ tests = testGroup "Mempool"
       ]
   , testProperty "snapshotTxs == snapshotTxsAfter zeroIdx" prop_Mempool_snapshotTxs_snapshotTxsAfter
   , testProperty "valid added txs == getTxs"               prop_Mempool_addTxs_getTxs
-  , testProperty "addTxs txs == mapM (addTxs . pure) txs"  prop_Mempool_addTxs_one_vs_multiple
+  , testProperty "addTxs [..] == forM [..] addTxs"         prop_Mempool_semigroup_addTxs
   , testProperty "result of addTxs"                        prop_Mempool_addTxs_result
   , testProperty "Invalid transactions are never added"    prop_Mempool_InvalidTxsNeverAdded
   , testProperty "result of getCapacity"                   prop_Mempool_getCapacity
@@ -75,6 +75,7 @@ tests = testGroup "Mempool"
   , testProperty "Removed invalid txs are traced"          prop_Mempool_TraceRemovedTxs
   , testProperty "idx consistency"                         prop_Mempool_idx_consistency
   , testProperty "removeTxs"                               prop_Mempool_removeTxs
+  , testProperty "removeTxs [..] == forM [..] removeTxs"   prop_Mempool_semigroup_removeTxs
   ]
 
 {-------------------------------------------------------------------------------
@@ -99,15 +100,24 @@ prop_Mempool_addTxs_getTxs setup =
       return $ counterexample (ppTxs (txs setup)) $
         validTxs setup `isSuffixOf` map (txForgetValidated . fst) snapshotTxs
 
--- | Same as 'prop_Mempool_addTxs_getTxs', but add the transactions one-by-one
--- instead of all at once.
-prop_Mempool_addTxs_one_vs_multiple :: TestSetupWithTxs -> Property
-prop_Mempool_addTxs_one_vs_multiple setup =
-    withTestMempool (testSetup setup) $ \TestMempool { mempool } -> do
-      forM_ (allTxs setup) $ \tx -> addTxs mempool [tx]
-      MempoolSnapshot { snapshotTxs } <- atomically $ getSnapshot mempool
-      return $ counterexample (ppTxs (txs setup)) $
-        validTxs setup `isSuffixOf` map (txForgetValidated . fst) snapshotTxs
+-- | Test that both adding the transactions one by one and adding them in one go
+-- produce the same result.
+prop_Mempool_semigroup_addTxs :: TestSetupWithTxs -> Property
+prop_Mempool_semigroup_addTxs setup =
+  withTestMempool (testSetup setup) $ \TestMempool {mempool = mempool1} -> do
+  _ <- addTxs mempool1 (allTxs setup)
+  snapshot1 <- atomically $ getSnapshot mempool1
+
+  return $ withTestMempool (testSetup setup) $ \TestMempool {mempool = mempool2} -> do
+    forM_ (allTxs setup) $ \tx -> addTxs mempool2 [tx]
+    snapshot2 <- atomically $ getSnapshot mempool2
+
+    return $ counterexample
+      ("Transactions after adding in one go: " <> show (snapshotTxs snapshot1)
+       <> "\nTransactions after adding one by one: " <> show (snapshotTxs snapshot2)) $
+        snapshotTxs snapshot1 === snapshotTxs snapshot2 .&&.
+        snapshotMempoolSize snapshot1 === snapshotMempoolSize snapshot2 .&&.
+        snapshotSlotNo snapshot1 === snapshotSlotNo snapshot1
 
 -- | Test that the result of adding transaction to a 'Mempool' matches our
 -- expectation: invalid transactions have errors associated with them and
@@ -155,6 +165,25 @@ prop_Mempool_removeTxs (TestSetupWithTxInMempool testSetup txToRemove) =
         ("Transactions in the mempool after removing (" <>
          show txToRemove <> "): " <> show txsInMempoolAfter)
         (txToRemove `notElem` map txForgetValidated txsInMempoolAfter)
+
+-- | Test that both removing transactions one by one and removing them in one go
+-- produce the same result.
+prop_Mempool_semigroup_removeTxs :: TestSetupWithTxsInMempool -> Property
+prop_Mempool_semigroup_removeTxs (TestSetupWithTxsInMempool testSetup txsToRemove) =
+  withTestMempool testSetup $ \TestMempool {mempool = mempool1} -> do
+  removeTxs mempool1 $ map txId txsToRemove
+  snapshot1 <- atomically (getSnapshot mempool1)
+
+  return $ withTestMempool testSetup $ \TestMempool {mempool = mempool2} -> do
+    forM_ (map txId txsToRemove) (removeTxs mempool2 . (:[]))
+    snapshot2 <- atomically (getSnapshot mempool2)
+
+    return $ counterexample
+      ("Transactions after removing in one go: " <> show (snapshotTxs snapshot1)
+       <> "\nTransactions after removing one by one: " <> show (snapshotTxs snapshot2)) $
+        snapshotTxs snapshot1 === snapshotTxs snapshot2 .&&.
+        snapshotMempoolSize snapshot1 === snapshotMempoolSize snapshot2 .&&.
+        snapshotSlotNo snapshot1 === snapshotSlotNo snapshot1
 
 -- | Test that 'getCapacity' returns the 'MempoolCapacityBytes' value that the
 -- mempool was initialized with.
@@ -650,7 +679,7 @@ instance Arbitrary TestSetupWithTxInMempool where
     return $ TestSetupWithTxInMempool testSetup tx
 
   shrink (TestSetupWithTxInMempool testSetup _tx) =
-    [ TestSetupWithTxInMempool testSetup tx'
+    [ TestSetupWithTxInMempool testSetup' tx'
     | testSetup' <- shrink testSetup
     , not . null . testInitialTxs $ testSetup'
     , tx' <- testInitialTxs testSetup'
--- a/ouroboros-consensus-test/test-consensus/Test/Consensus/Mempool.hs
+++ b/ouroboros-consensus-test/test-consensus/Test/Consensus/Mempool.hs
@@ -656,6 +656,18 @@ instance Arbitrary TestSetupWithTxInMempool where
     , tx' <- testInitialTxs testSetup'
     ]
 
+data TestSetupWithTxsInMempool = TestSetupWithTxsInMempool TestSetup [TestTx]
+  deriving (Show)
+
+instance Arbitrary TestSetupWithTxsInMempool where
+  arbitrary = do
+    TestSetupWithTxs { testSetup } <-
+      arbitrary `suchThat` (not . null . testInitialTxs . testSetup)
+    txs <- sublistOf (testInitialTxs testSetup)
+    return $ TestSetupWithTxsInMempool testSetup txs
+
+  -- TODO shrink
+
 {-------------------------------------------------------------------------------
   TestMempool: a mempool with random contents
 -------------------------------------------------------------------------------}
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
@@ -1,25 +1,28 @@
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE OverloadedStrings          #-}
-{-# LANGUAGE RankNTypes                 #-}
+{-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
-{-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE UndecidableInstances       #-}
-{-# LANGUAGE UndecidableSuperClasses    #-}
 
 module Ouroboros.Consensus.Mempool.API (
-    ForgeLedgerState (..)
-  , Mempool (..)
+    -- * Mempool
+    Mempool (..)
   , MempoolAddTxResult (..)
-  , MempoolCapacityBytes (..)
-  , MempoolSize (..)
-  , MempoolSnapshot (..)
-  , TraceEventMempool (..)
-  , addTxs
   , isMempoolTxAdded
   , isMempoolTxRejected
   , mempoolTxAddedToMaybe
+    -- * Mempool Snapshot
+  , ForgeLedgerState (..)
+  , MempoolSnapshot (..)
+    -- * Mempool size and capacity
+  , MempoolCapacityBytes (..)
+  , MempoolCapacityBytesOverride (..)
+  , MempoolSize (..)
+  , computeMempoolCapacity
+    -- * Tracing support
+  , TraceEventMempool (..)
+  , addTxs
     -- * Re-exports
   , TxSizeInBytes
   ) where
@@ -28,17 +31,21 @@ import           Data.Word (Word32)
 
 import           Ouroboros.Network.Protocol.TxSubmission.Type (TxSizeInBytes)
 
-import           Ouroboros.Consensus.Block
+import           Ouroboros.Consensus.Block (SlotNo)
 import           Ouroboros.Consensus.Ledger.Abstract
 import           Ouroboros.Consensus.Ledger.SupportsMempool
 import           Ouroboros.Consensus.Util.IOLike
+
+{-------------------------------------------------------------------------------
+  Mempool API
+-------------------------------------------------------------------------------}
 
 -- | Mempool
 --
 -- The mempool is the set of transactions that should be included in the next
 -- block. In principle this is a /set/ of all the transactions that we receive
 -- from our peers. In order to avoid flooding the network with invalid
--- transactions,  however, we only want to keep /valid/ transactions in the
+-- transactions, however, we only want to keep /valid/ transactions in the
 -- mempool. That raises the question: valid with respect to which ledger state?
 --
 -- We opt for a very simple answer to this: the mempool will be interpreted
@@ -51,6 +58,17 @@ import           Ouroboros.Consensus.Util.IOLike
 --   of transactions that fits in a block.
 -- * It supports wallets that submit dependent transactions (where later
 --   transaction depends on outputs from earlier ones).
+--
+-- When only one thread is operating on the mempool, operations that mutate the
+-- state based on the input arguments (tryAddTxs and removeTxs) will produce the
+-- same result whether they process transactions one by one or all in one go, so
+-- this equality holds:
+--
+-- > void (tryAddTxs txs) === forM_ txs (tryAddTxs . (:[]))
+-- > void (trAddTxs [x,y]) === tryAddTxs x >> void (tryAddTxs y)
+--
+-- This shows that 'tryAddTxs' is an homomorphism from '++' and '>>', which
+-- informally makes these operations "distributive".
 data Mempool m blk idx = Mempool {
       -- | Add a bunch of transactions (oldest to newest)
       --
@@ -83,12 +101,11 @@ data Mempool m blk idx = Mempool {
       --      'MempoolTxRejected' value. These transactions are not in the
       --      Mempool.
       --
-      -- 2. A list containing the transactions that have not yet been added
-      --    yet, as the capacity of the Mempool has been reached. I.e., there
-      --    is no space in the Mempool to add the first transaction in this
-      --    list. Note that we won't try to add smaller transactions after
-      --    that first transaction because they might depend on the first
-      --    transaction.
+      -- 2. A list containing the transactions that have not yet been added, as
+      --    the capacity of the Mempool has been reached. I.e., there is no
+      --    space in the Mempool to add the first transaction in this list. Note
+      --    that we won't try to add smaller transactions after that first
+      --    transaction because they might depend on the first transaction.
       --
       -- POSTCONDITION:
       -- > let prj = \case
@@ -179,6 +196,10 @@ data Mempool m blk idx = Mempool {
     , zeroIdx        :: idx
     }
 
+{-------------------------------------------------------------------------------
+  Result of adding a transaction to the mempool
+-------------------------------------------------------------------------------}
+
 -- | The result of attempting to add a transaction to the mempool.
 data MempoolAddTxResult blk
   = MempoolTxAdded !(Validated (GenTx blk))
@@ -208,9 +229,7 @@ isMempoolTxRejected _                   = False
 -- This function does not sync the Mempool contents with the ledger state in
 -- case the latter changes, it relies on the background thread to do that.
 --
--- POSTCONDITON:
--- > processed <- addTxs mpEnv txs
--- > map fst processed == txs
+-- See the necessary invariants on the Haddock for 'API.tryAddTxs'.
 addTxs
   :: forall m blk idx. MonadSTM m
   => Mempool m blk idx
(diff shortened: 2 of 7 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -83,12 +95,11 @@ data Mempool m blk idx = Mempool {
...
       --      Mempool.
       --
-      -- 2. A list containing the transactions that have not yet been added
-      --    yet, as the capacity of the Mempool has been reached. I.e., there
-      --    is no space in the Mempool to add the first transaction in this
-      --    list. Note that we won't try to add smaller transactions after
-      --    that first transaction because they might depend on the first
-      --    transaction.
+      -- 2. A list containing the transactions that have not yet been added, as
+      --    the capacity of the Mempool has been reached. I.e., there is no
```
Jasagredo, line 99: This was just a repeated "yet".

ouroboros-consensus-test/test-consensus/Test/Consensus/Mempool.hs
```diff
@@ -650,12 +679,30 @@ instance Arbitrary TestSetupWithTxInMempool where
...
+  deriving (Show)
+
+instance Arbitrary TestSetupWithTxsInMempool where
+  arbitrary = do
+    TestSetupWithTxs { testSetup } <-
+      arbitrary `suchThat` (not . null . testInitialTxs . testSetup)
+    txs <- sublistOf (testInitialTxs testSetup)
+    return $ TestSetupWithTxsInMempool testSetup txs
+
+  -- shrink (TestSetupWithTxInMempool testSetup _tx) =
```
nfrisby, line 698: Let's go with either just `TODO shrink` or fix it up now (eg if you just forgot to).

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
```diff
@@ -58,6 +58,11 @@ import           Ouroboros.Consensus.Util.IOLike
 --   of transactions that fits in a block.
 -- * It supports wallets that submit dependent transactions (where later
 --   transaction depends on outputs from earlier ones).
+--
+-- When only one thread is operating on the mempool, operations that mutate the
+-- state based on the input arguments (tryAddTxs and removeTxs) will produce the
+-- same result whether they process transactions one by one or all in one go,
+-- i.e.  those operations are "distributive".
```
nfrisby, line 65: I don't think `distributive` is the right choice. Here's what we have:

`add [x,y] ~ add [x] >> add [y]`

That phrasing makes it rather clear that `add` is a homomorphism from `++` to `>>`. That wording is very precise, but is admittedly some heavyweight jargon. Hopefully including the code snippet I wrote would mitigate that.

Hmm. Actually, I take back what I said in the first sentence. Of the more common terms, maybe distributive is the closest after all. `x * (a + b) = (x * a) + (x * b)`; the left + is `++` and the right + is `>>` and the `x*` is `add $`. If we interpret `+` as `<>` and pretend there's a semigroup instance for `m a`, then that's actually pretty honest.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+
+
+{-------------------------------------------------------------------------------
+  Mempool Implementation
+-------------------------------------------------------------------------------}
+
+-- | Data constructor containing the control flow for 'tryAddTxs'. Can be
+-- interpreted in an IOLike monad with 'runTryAddTxs'.
+data TryAddTxs blk =
+  Done [GenTx blk]
```
nfrisby, line 49: Style: `Done` should be indented the same as `Trace`.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+           , [GenTx blk])
+runTryAddTxs stateVar trcr = go []
+  where go processed (Done remaining) =
+          return (reverse processed, remaining)
+        go processed (Trace tem tat') = do
+          return $ traceWith trcr tem
+          go processed tat'
+        go processed (WriteTx is tat') = do
+          writeTVar stateVar is
+          -- TODO: are intermediate states even needed? or could we stack them
```
nfrisby, line 89: Since the mini protocols for multiple peers are running concurrently, I anticipate that writing the intermediate states ASAP is in some sense the most "fair"? This might deserve some discussion on a sync call.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+      needsTrace =
+        if null txIds
+        then Nothing
+        else Just $
+               TraceMempoolManuallyRemovedTxs txIds (map fst (vrInvalid vr)) (isMempoolSize is')
+  in WriteRemoveTxs is' needsTrace
+
+-- | A datatype containing the new state produced by syncing with the Ledger, a
+-- snapshot of that mempool state and, if needed, a tracing message.
+data SyncWithLedger blk = NewSyncedState (InternalState blk) (MempoolSnapshot blk TicketNo) (Maybe (TraceEventMempool blk))
```
nfrisby, line 213: Style: wrap the long line

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+--
+-- See the documentation of 'runSyncWithLedger' for more context.
+pureSyncWithLedger :: (LedgerSupportsMempool blk, HasTxId (GenTx blk), ValidateEnvelope blk)
+                   => InternalState blk
+                   -> LedgerState blk
+                   -> LedgerConfig blk
+                   -> MempoolCapacityBytesOverride
+                   -> SyncWithLedger blk
+pureSyncWithLedger istate lstate lcfg capacityOverride =
+  let vr = validateIS istate lstate lcfg capacityOverride
```
nfrisby, line 245: Style: align these `=`s.

(There are some other let groups with the same ragged `=` layouts that should probably also be touched up.)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+  -> RemoveTxs blk
+  -> STM m (Maybe (TraceEventMempool blk))
+runRemoveTxs stateVar (WriteRemoveTxs is t) = do
+  writeTVar stateVar is
+  return t
+
+-- | Craft a 'RemoveTxs' that manually removes the given transactions from the
+-- mempool, returning inside it an updated InternalState.
+pureRemoveTxs
+  :: ( LedgerSupportsMempool blk
```
nfrisby, line 177: Style: the signatures in this module are inconsistent regarding `::` placement and indentation. Please make them uniform. (See `StyleGuide.md` for the preferred details, though "consistency within the module" is the only hard requirement.)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+                      , HasTxId (GenTx blk)
+                      , ValidateEnvelope blk
+                      )
+                   => LedgerConfig blk
+                   -> ForgeLedgerState blk
+                   -> MempoolCapacityBytesOverride
+                   -> InternalState blk
+                   -> MempoolSnapshot blk TicketNo
+pureGetSnapshotFor cfg blockLedgerState capacityOverride =
+    implSnapshotFromIS
```
nfrisby, line 269: Style: `implSnapshotFromIS` should be indented 2 more.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+    , snapshotTxsAfter    = implSnapshotGetTxsAfter    is
+    , snapshotTxsForSize  = implSnapshotGetTxsForSize  is
+    , snapshotLookupTx    = implSnapshotGetTx          is
+    , snapshotHasTx       = implSnapshotHasTx          is
+    , snapshotMempoolSize = implSnapshotGetMempoolSize is
+    , snapshotSlotNo      = isSlotNo                   is
+    , snapshotLedgerState = isLedgerState              is
+    }
+ where
+  implSnapshotGetTxs :: InternalState blk
```
nfrisby, line 292: Style: the `::` and `->` in these where clause are off.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+-- from the mempool, whereas valid transactions will remain.
+--
+-- n.b. in our current implementation, when one opens a mempool, we
+-- spawn a thread which performs this action whenever the 'ChainDB' tip
+-- point changes.
+runSyncWithLedger :: forall m blk. IOLike m
+  => StrictTVar m (InternalState blk)
+  -> SyncWithLedger blk
+  -> STM m (Maybe (TraceEventMempool blk), MempoolSnapshot blk TicketNo)
+runSyncWithLedger stateVar (NewSyncedState is msp mTrace) = do
```
nfrisby, line 230: Style: generally we indent by 4 even if there's no where clause, because there might be a where clause later.

But since you're consistent in the module, this is probably fine.

nfrisby (changes requested):
Most of my comments are style stuff. Feel free to ignore them if (a) it's just code you copied over or (b) you think it's significantly clearer as-is than the recommendation from `StyleGuide.md` would be.

I'm Requesting Changes because there is at least some dead commented code that we should fixup or remove.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: As this whole operation is done atomically, I don't think this is actually relevant and we can just write once.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
Jasagredo: I briefly chatted with Edsko about what name to use for this and although it is an homomorphism from `++` to `>>`, in terms of intuitive wording it is kind of "distributive" that's why I wrote it in quotes.

Let's discuss later what term to use.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: Indeed, fixed. (resolving)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: Fixed (resolving)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: I look forward to some kind of automation on this stuff, as I have to consult the guide each time as it is still not wired in my brain.

Fixed (resolving)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: Fixed (resolving)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: Indeed (I <TAB>ed the beginning of the type signature and forgot to move the rest). Fixed (resolving)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: More like my tooling is consistent in indenting with 2 despite I having set
```
(haskell-indent-offset 4)
```
Fixed this one (and some others). Resolving

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,321 @@
...
+                Processed (MempoolTxRejected firstTx err) $
+                go toAdd' is'
+
+-- | A datatype containing the state resulting after removing the requested
+-- transactions from the mempool and maybe a message to be traced while removing
+-- them.
+data RemoveTxs blk = WriteRemoveTxs (InternalState blk) (Maybe (TraceEventMempool blk))
+
+-- | Intepret a 'RemoveTxs' with the resulting values produced by manually
+-- removing the transactions given to 'pureRemovTxs' from the mempool.
```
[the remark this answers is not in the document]
Jasagredo: Fixed.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: For now I will leave it as is, probably will change it to be only one write based on previous discussions.

ouroboros-consensus-test/test-consensus/Test/Consensus/Mempool.hs
Jasagredo: Will put a TODO for now.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -104,16 +94,43 @@ mkMempool
...
-mkMempool env = Mempool
-    { tryAddTxs      = implTryAddTxs      env
-    , removeTxs      = implRemoveTxs      env
-    , syncWithLedger = implSyncWithLedger env
-    , getSnapshot    = implGetSnapshot    env
-    , getSnapshotFor = implGetSnapshotFor env
-    , getCapacity    = implGetCapacity    env
-    , getTxSize      = mpEnvTxSize        env
+mkMempool mpEnv = Mempool
+    { tryAddTxs      = \txs -> do
```
nfrisby, line 98: This `tryAddTxs` has a change in semantics that I hadn't noticed before.

Here's the telling observation: each call to the old `tryAddTxs` could induce multiple STM transactions, such as one per transaction in the given list. This new implementation will only ever create a single STM transaction.

I haven't carefully considered why that's an important behavior to preserve, but I think it is. The old transactions would each just attempt to add one tx to the mempool. The new transactions each try to add all of the txs to the mempool, or the longest prefix that would fit. EG If we consider multiple peers trying to add to our mempool, the one tx per transaction seems likely to exhibit better fairness.

But like I said, I haven't considered that very carefully. Perhaps Thomas wrote something relevant in the Consensus Report we could double-check.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -290,516 +207,24 @@ forkSyncStateOnTipPointChange registry menv =
...
+    MempoolEnv { mpEnvStateVar = istate
+               , mpEnvLedger = ldgr
+               , mpEnvTracer = trcr
+               , mpEnvLedgerCfg = cfg
+               , mpEnvCapacityOverride = co
+               } = menv
+
     action :: Point blk -> m ()
-    action _tipPoint = void $ implSyncWithLedger menv
+    action _tipPoint = void $ do
```
nfrisby, line 218: Style: This layout is a little unconventional. Please keep to the rule that a `do` introduces a 2-space indent. (Also, please use `do $` instead of `do (`.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
nfrisby: I commented elsewhere just a minute ago, and now I realized it was relevant to this thread.

Summary: the "whole operation is done atomically" is new behavior; it previous was not atomic: there was one `STM` transaction per tx. (Note that `atomically` was used once per call of the recursive function named `go` in the old code.)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,328 @@
...
+                                     -- additional write.
+                                     WriteTx is'
+
+            -- We only extended the ValidationResult with a single transaction
+            -- ('firstTx'). So if it's not in 'vrInvalid', it must be in
+            -- 'vrNewValid'.
+            getValidationOutcome = case eVtx of
+              Right vtx ->
+                assert (isJust (vrNewValid vr)) $
+                Trace (TraceMempoolAddedTx
```
nfrisby, line 139: Style: there are some inconsistencies in how you indent the arguments on a multi-line application. EG here the arguments are indented by 1 space (or 2 from the opening paren). In the next `Trace`, they're indented by negative 1 space (or 0 from the opening paren). On line 206 below (`Just $\n...`) it's indented negative.

Generally, negative indent is unconventional and we shouldn't do that in this code base. The `StyleGuide.md` recommends 2 spaces (from the function name; regardless of the opening paren). The most important rule is to be consistent.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Types.hs
```diff
@@ -0,0 +1,399 @@
...
+  -> [TxTicket (Validated (GenTx blk))]
+  -> ValidationResult (Validated (GenTx blk)) blk
+revalidateTxsFor capacityOverride cfg slot st lastTicketNo txTickets =
+    repeatedly
+      (extendVRPrevApplied cfg)
+      txTickets
+      (validationResultFromIS is)
+  where
+    is = initInternalState capacityOverride lastTicketNo slot st
+
```
nfrisby, line 322: Extra newline here

nfrisby (changes requested):
OK. Some more style stuff, again rather minor. The big idea being consistency within the module (but also nothing too unconventional).

I did catch a significant semantic change that I didn't notice the first time around. We can talk about the use of `go ... = ... join $ atomically (... <recursive call> ...)` if you're not familiar with that pattern -- I wasn't before working on this code base. It's a bit subtle.

Overall I like the new module organization and your comments are fantastically thorough. Thanks!

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: Indeed you are right, answered in the other thread. Closing this one.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
Jasagredo: This is indeed true. I re-implemented this operation to preserve the same execution flow as before, which should ensure fairness here.

This being said, I might still be missing some comments specifying why, but I think what I wrote should be enough.

TODO: test the fairness.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
Jasagredo: True, fixed. (Resolving)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
```diff
@@ -104,16 +94,43 @@ mkMempool
...
     , zeroIdx        = zeroTicketNo
     }
+   where MempoolEnv{ mpEnvStateVar = istate
+                   , mpEnvLedgerCfg = cfg
+                   , mpEnvTxSize = txSize
+                   , mpEnvTracer = trcr
+                   , mpEnvLedger = ldgr
+                   , mpEnvCapacityOverride = co
+                   } = mpEnv
+         mtrace = maybe (return ()) (traceWith trcr)
```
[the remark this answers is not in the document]
Jasagredo: Didn't know. I'm using it now, thanks! (Resolving)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Types.hs
Jasagredo: True, removed.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,328 @@
...
+import           Ouroboros.Consensus.Ledger.Abstract
+import           Ouroboros.Consensus.Ledger.SupportsMempool
+import           Ouroboros.Consensus.Mempool.API
+import           Ouroboros.Consensus.Mempool.Impl.Types
+import           Ouroboros.Consensus.Mempool.TxSeq (TicketNo, TxTicket (..),
+                     zeroTicketNo)
+import qualified Ouroboros.Consensus.Mempool.TxSeq as TxSeq
+import           Ouroboros.Consensus.Util.IOLike
+
+
```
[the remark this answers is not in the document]
Jasagredo: True, removed (Closing).

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,343 @@
...
+-- Mempool capacity was reached. See 'addTxs' for a function that blocks in
+-- case the Mempool capacity is reached.
+--
+-- Transactions are added one by one, updating the Mempool each time one was
+-- added successfully.
+--
+-- This function does not sync the Mempool contents with the ledger state in
+-- case the latter changes, it relies on the background thread to do that.
+--
+-- POSTCONDITON:
```
nfrisby, line 77: Javier noticed that this comment has been out-of-date since my Mempool changes for Alonzo. He's going to simplify this comment, saying something like `See the necessary invariants on the Haddock for 'API.tryAddTxs'.`

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl.hs
Jasagredo: As this was addressed in a follow-up issue #3127, I'm closing this comment.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
Jasagredo: Added a reference to the invariants in both comments that needed it.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,341 @@
...
+    -- | There are more transactions to process.
+  | More
+      (Maybe (InternalState blk))
+      -- ^ If the transaction was accepted, the new state that can be written to
+      -- the TVar.
+      (TraceEventMempool blk)
+      -- ^ The event emitted by the operation.
+      (MempoolAddTxResult blk)
+      -- ^ The result of trying to add the transaction to the mempool.
+      (InternalState blk -> TryAddTxs blk)
```
nfrisby, line 60: This last argument is a non-trivial maneuver. I'm wondering if we can simplify by being a bit more direct.

```
...
tryAddTxs      = implTryAddTxs istate
...

implTryAddTxs istate txs =
    go []
  where
    go acc = \case
      []     -> pure (reverse acc, [])
      tx:txs -> join $ atomically $ do
        is <- readTVar istate
        -- this case is essentially psuedo-code; missing eg tracer logic
        pureTryAddTx {- NOTE one tx at a time -} ... is tx >>= \case
          Nothing            -> pure $ pure (reverse acc, tx:txs)
          Just (is', result) -> do writeTVar istate is'; pure $ go (result:acc) txs
```

The pure code handles a single tx at a time, while the monadic wrapper handles one tx at a time, only maintaining the list of results and the list of unattempted txs.

This way `TryAddTxs` doesn't even have to be recursive, let alone a coroutine sort of thing.

nfrisby (changes requested):
I'm sorry to Request Changes of you again. I hadn't noticed the last argument of `More` while we were on the call this morning; wish we could have chatted through it before the end of your day.

It seems correct as-is to me, but I think we can get by with something simpler. Please see my comment.

nfrisby (approved):
LGTM, thanks for iterating through it with me!

Jasagredo:
Re-organized the commits and rebased.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Pure.hs
```diff
@@ -0,0 +1,342 @@
...
+pureSyncWithLedger
+  :: (LedgerSupportsMempool blk, HasTxId (GenTx blk), ValidateEnvelope blk)
+  => InternalState blk
+  -> LedgerState blk
+  -> LedgerConfig blk
+  -> MempoolCapacityBytesOverride
+  -> SyncWithLedger blk
+pureSyncWithLedger istate lstate lcfg capacityOverride =
+    let vr          = validateIS istate lstate lcfg capacityOverride
+        mempoolSize = isMempoolSize istate
```
nfrisby, line 264: (Caught many days later.)

I think some of these `istate` arguments should be `istate'`. See Issue #3181

PR #3119, opened 2021-05-07
Add Alonzo Era

EncodePanda:
In this commit we are adding a new era: Alonzo Era. This is
still a "shelley based era" we were able to reuse a lot of code and
functionality. It is however noticeable that this commit is rather
substantial. This is due to the new development and improvements on
the ledger-spec side, that required modifications in the consensus code.

Most noticeable changes were:
1. introduction of a TranslationContext
2. introduction of different type families, replacing concrete types
that were unified for all previous eras.

Co-authored-by: nfrisby <nick.frisby@iohk.io>


nfrisby:
Pawel indicated that the `Init CardanoEras` route was going to involve a lot of duplication.

So we decided to go the other route: leave `AlonzoEra` in `CardanoEras`, and use that for the tests etc, but never create `AlonzoEra` summands in the generators. It will be follow-up work to do so, but is not necessary eg for the current `testnet` "release" target.

Thus: Closing this, in favor of PR 3130, PR 3131, and so on.

PR #3121, opened 2021-05-07
Network Mux Close

coot:


njd42 (approved):
This is all looking good. The logic of the "group" didn't jump out at me, but thats fine.

network-mux/src/Network/Mux.hs
```diff
@@ -406,8 +410,17 @@ monitor tracer timeout jobpool egressQueue cmdQueue muxStatus =
...
+            size <- JobPool.readGroupSize jobpool MiniProtocolJob
+            case size of
+              0  | Just (MuxError MuxBearerClosed _) <- fromException e
+                -> writeTVar muxStatus MuxStopped
+                >> return True
+              _ -> writeTVar muxStatus (MuxFailed e)
+                >> return False
+          if r
+            then return ()
+            else throwIO e
```
karknu, line 423: :+1: no more 'MuxError BearerClosed' questions!

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
```diff
@@ -0,0 +1,333 @@
...
+--
+data CloseError = CloseTimeoutError
+  deriving (Show, Typeable)
+
+instance Exception CloseError
+
+
+data SuccessOrFailure = Success | Failure
+
+type Size = Int64
```
karknu, line 145: Why is this `Int64` and not `Word64`?

karknu (approved):
LGTM.

network-mux/src/Network/Mux.hs
coot: but only in the `p2p` branch. Clean close needs to await for all mini-protocols to terminate and we don't have such a component in the current system.  It wouldn't be difficult to implement though.

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
coot: The only reason is that lazy bytestring is using `Int64` for [`length`](https://hackage.haskell.org/package/bytestring-0.11.1.0/docs/Data-ByteString-Lazy.html#v:length).

coot:
bors merge

PR #3122, opened 2021-05-08
pick policies context

coot:
Expose 'PeerSource', and fail count in pick policies.

- peer-selection: make PeerSource available to PickPolicy
- peer-selection: update PickPolicy
- peer-selection: refactor Diffusion.Policies
- peer-selection: expose number of failed connection attempts


bolt12:
LGTM

dcoutts (changes requested):
I think we should follow a slightly different style for extending the pick policy with extra attributes.

I've pushed an example of what I mean to branch `dcoutts/pick-policy`. My suggestion is that we re-do this PR on that style. It's pretty similar so the actual policy function changes should be the same. It's just about how we pass the info.

PR #3124, opened 2021-05-09
Clean PeerSelectionCounters

coot:


bolt12:
LGTM! That is _definitely_ a better naming 😄

PR #3128, opened 2021-05-11
Fix InboundGovernorCounters

coot:


ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -101,100 +103,98 @@ inboundGovernor tracer serverControlChannel protocolIdleTimeout
...
+          :: ConnectionState muxMode peerAddr m a b
+          -> STM m (Maybe ProtocolTemperature)
+        inboundTemperature ConnectionState { csMux, csMiniProtocolMap } =
+              fmap go
+            . sequence
+            . Mux.miniProtocolStateMap
+            $ csMux
+          where
+            go :: Map (MiniProtocolNum, MiniProtocolDir) MiniProtocolStatus
+              -> Maybe ProtocolTemperature
```
bolt12, line 135: Just a nitpick: align to the `::`

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -329,14 +235,18 @@ inboundGovernor tracer serverControlChannel protocolIdleTimeout
             Just err -> traceWith tracer (TrMuxErrored connId err)
 
           -- the connection manager does should realise this on itself.
-          newState <- updateState (unregisterConnection connId state) connId
-          inboundGovernorLoop newState
+          let state' = unregisterConnection connId state
+          traceWith tracer ( TrInboundGovernorCounters
```
bolt12, line 239: I'm not sure if it is worth to trace before the recursive call because that is the first thing the recursive call is going to do!

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -348,8 +258,10 @@ inboundGovernor tracer serverControlChannel protocolIdleTimeout
               traceWith tracer $
                 TrResponderErrored tConnId num e
 
-              newState <- updateState (unregisterConnection tConnId state) tConnId
-              inboundGovernorLoop newState
+              let state' = unregisterConnection tConnId state
+              traceWith tracer ( TrInboundGovernorCounters
```
bolt12, line 262: Same here and on all other places

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -358,9 +270,18 @@ inboundGovernor tracer serverControlChannel protocolIdleTimeout
                 Right completionAction -> do
                   traceWith tracer (TrResponderRestarted tConnId num)
 
-                  newState <- updateState (updateMiniProtocol tConnId num completionAction state)
-                                          tConnId
-                  inboundGovernorLoop newState
+                  let isHot = mpdMiniProtocolTemp == Hot
```
bolt12, line 273: We can get rid of this if we leave the tracing for the recursive call

bolt12:
LGTM! Super cool, I learned a lot :)

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
coot: Good point, I'll remove it from all places but the first one.

coot:
I think on top of that we could add incremental computation of counters, since the transitions `Warm -> Hot` and `Hot -> Warm` are explicit; another refactoring that the `inbound-governor` is really asking for are two monoids:
* `FirstToFinish m a`
* `LastToFinish m a`

Issue #3105, opened 2021-04-29
Clean-up our `qualified as` nicknames for Ledger imports


nfrisby:
SL, E, CL, Era, etc -- we're confusing ourselves.

This Issue is for @EncodePanda and I to come up with a good set of qualified nicknames and apply them consistently in the code base.

nfrisby:
How about `Core`? That's what ledger uses internally.

PR #3130, opened 2021-05-11
Update cardano-ledger-specs dependency

nfrisby:
This is a single commit. See the message.

### DIFF
--- a/ouroboros-consensus-mock-test/src/Test/Consensus/Ledger/Mock/Generators.hs
+++ b/ouroboros-consensus-mock-test/src/Test/Consensus/Ledger/Mock/Generators.hs
@@ -61,11 +61,22 @@ instance Arbitrary (HeaderHash blk) => Arbitrary (Point blk) where
   Generators
 -------------------------------------------------------------------------------}
 
--- These generators blindly create random values, so the block will not be
--- valid, but this does not matter for serialisation tests.
-
+-- | This blindly creates random values, so the block will not be valid, but
+-- this does not matter for serialisation tests.
 instance (SimpleCrypto c, Arbitrary ext, Serialise ext)
       => Arbitrary (SimpleBlock c ext) where
+  arbitrary = do
+    simpleStdHeader <- arbitrary
+    body            <- arbitrary
+    ext             <- arbitrary
+    let hdr = mkSimpleHeader encode simpleStdHeader ext
+    return $ SimpleBlock hdr body
+
+-- | This blindly creates random values, so the block will not be valid, but
+-- this does not matter for serialisation tests. Except we do touch-up the
+-- 'simpleBodySize'; hence 'Coherent'.
+instance (SimpleCrypto c, Arbitrary ext, Serialise ext)
+      => Arbitrary (Coherent (SimpleBlock c ext)) where
   arbitrary = do
     simpleStdHeader <- arbitrary
     body            <- arbitrary
@@ -76,7 +87,7 @@ instance (SimpleCrypto c, Arbitrary ext, Serialise ext)
             simpleBodySize = fromIntegral $ Lazy.length $ serialise body
           }
         hdr = mkSimpleHeader encode simpleStdHeader' ext
-    return $ SimpleBlock hdr body
+    return $ Coherent $ SimpleBlock hdr body
 
 instance (SimpleCrypto c, Arbitrary ext, Serialise ext, Typeable ext)
       => Arbitrary (Header (SimpleBlock c ext)) where
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
@@ -30,24 +30,23 @@ import           GHC.Records
 import           NoThunks.Class (NoThunks)
 import           Numeric.Natural (Natural)
 
+import           Cardano.Binary (FromCBOR, ToCBOR)
+
 import           Cardano.Ledger.Allegra (AllegraEra)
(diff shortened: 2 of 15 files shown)


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -30,25 +30,24 @@ import           GHC.Records
...
 import           Numeric.Natural (Natural)
 
+import           Cardano.Binary (FromCBOR, ToCBOR)
+
 import           Cardano.Ledger.Allegra (AllegraEra)
-import qualified Cardano.Ledger.Core as LC
+import           Cardano.Ledger.Allegra.Translation ()
+import qualified Cardano.Ledger.Core as Core
 import           Cardano.Ledger.Era (Crypto, SupportsSegWit (..))
+import qualified Cardano.Ledger.Era as SL (TranslateEra (..), TxInBlock)
```
Jasagredo, line 39: `Cardano.Ledger.Era` was qualified above as `Core`. Maybe we should keep the same consistent naming, but also it could be the case that this module specifically deals with stuff very specific to `Cardano.Ledger.Era`. In that case I would maybe qualify it as `Era`?

Jasagredo (approved):
Except from that optional comment, the rest looks fine. However I did not look into the non-edited files to search for more occurrences of similar stuff. Maybe it would be worth modifying other files to make the qualification more uniform across our code.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
nfrisby: Ah; I have commit on the next PR that does that. I'll move it to this one.

I went with `.Era` because generally what we're getting from there is the same conceptual thing as we get from `.Core` --- even the module header on `.Era` says it exports "core" stuff.

nfrisby: > Maybe it would be worth modifying other files to make the qualification more uniform across our code.

That is for Issue #3105, which I' don't want to address yet. I needed to rename them in the modules in this PR to avoid import conflicts, so I did it where necessary. There are several other modules with out-of-date imports.

We'll need to do a broader review and pick a Pareto-optimal scheme -- there are unfortunately multiple dimensions that come to mind.

nfrisby:
I pushed up a commit that disables the failing tests. See Issue #3134.

However, we might be able to fix up ledger in the next couple hours. So I'll wait until then -- maybe we can merge without this workaround.

nfrisby:
@Jasagredo I re-requested review, because I added the `Coherent` newtype and adjusted some instances.

The new `Arbitrary (Coherent (ShelleyBlock era))` instance in `Test.Consensus.Shelley.Generators` is the key change. Jared rapidly added a new generator to `cardano-ledger-specs` that is more careful than the (recently changed) `Arbitrary (SL.Block era)` instance that has supported our existing `Arbitrary (ShelleyBlock era)` instance. So now the properties we have that actually expect some integrity to the arbitrary Shelley blocks explicitly requires it via the `Coherent` wrapper in their arguments. (A `suchThat` or QC implication would have been hopelessly improbable.)

Jasagredo (approved):
Looks reasonable. `Coherent` looks right and the uses seem correct. I'm fine with merging this.

Jasagredo:
The Hydra build errors seem to be due to the long compiling times on `Ouroboros.Consensus.Cardano.Node`. Perhaps we should turn off those optimizations while this bug is present.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
Jasagredo: Perfect! resolving this thread then.

EncodePanda (approved):
Thank you for well described commit message. It was straight-forward to do the review (even after 3 days-off :) ).
In general I under the need for the `Coherent` wrapper. I get why some instances were wrapped, though I'm not sure why some other instances were wrapped. But I assume that was because GHC required it :)

ouroboros-consensus-mock-test/src/Test/Consensus/Ledger/Mock/Generators.hs
```diff
@@ -61,11 +61,20 @@ instance Arbitrary (HeaderHash blk) => Arbitrary (Point blk) where
...
 instance (SimpleCrypto c, Arbitrary ext, Serialise ext)
       => Arbitrary (SimpleBlock c ext) where
+  arbitrary = do
+    simpleStdHeader <- arbitrary
+    body            <- arbitrary
+    ext             <- arbitrary
+    let hdr = mkSimpleHeader encode simpleStdHeader ext
+    return $ SimpleBlock hdr body
+
+-- | This blindly creates random values, so the block will not be valid, but
```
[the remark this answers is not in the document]
nfrisby: I updated the comment a bit and added a corresponding one on the other instance. Hope that helps.

nfrisby:
b117726b4 is b29f6e190 with comments/stylish-haskell changes only.

nfrisby:
I created a local variant that removes `Coherent` et al and only uses `genCoherentBlock` for `Arbitrary (ShelleyBlock era)`. I built that and am running the tests suites again locally. If that has roughly the same times as b117726 , then I'll merge this PR as-is. If not, I'll force-push the one without `Coherent` and merge that.

nfrisby:
The timings seem the same. Merging.

nfrisby:
1d5b0c7a5 is b117726 rebased onto `origin/master`

PR #3135, opened 2021-05-13
Monoidal synchronisations via FirstToFinish and LastToFinish

coot:


njd42 (approved):
To me, this is a great simplification (at least in the cognitive load to understand what is going on). 

I wonder if we can reframe some of the other things (like issues of timeout) inside this framing?

coot:
> To me, this is a great simplification (at least in the cognitive load to understand what is going on).

For me too. When using `foldr (<|>) retry` vs `foldr (\ma mb -> (<>) <$> ma <*> mb) (pure mempty)` it's quite easy to plug the wrong identity and introduce a subtle synchornisation bug, which aren't easy to debug.

> I wonder if we can reframe some of the other things (like issues of timeout) inside this framing?

Timeouts using `STM` are possible, see [this](https://github.com/input-output-hk/ouroboros-network/pull/3135/commits/c455cd79d54ec3989156e19e252c64e163ecefbf#diff-5103d29aa29793fb7e60105eab8ac00b809c237370b47570992ab941b7722801R336).

bolt12:
Great addition!

dcoutts (commented):
Nice.

PR #3140, opened 2021-05-14
System start 4

newhoggy:


DavidEichmann:
I'm a little confused about what this PR is doing. The title says  "System start 4", so I assumed it would be using `GetPartialLedgerConfig` to derive the start time, but I don't see `GetPartialLedgerConfig` getting used anywhere. There is one commit adding some mapping functions that seems useful.

coot:
I removed me and Karl from reviewers, I think we were added only because of the changes in `cabal.project` file.

PR #3142, opened 2021-05-14
Removed assertions in PeerStateActions

coot:
If a demotion failed half the way, peer state actions could modify some
of the tvars.  Next time the governor would try to demote a peer it
these assertions would trigger.


bolt12 (approved):
LGTM

PR #3152, opened 2021-05-17
Make dummyEpochInfo generic

newhoggy:


nfrisby:
Hi John. The signature change seems reasonable to me. But the `cabal.project` change looks unnecessary. Can you update that or explain that? Thanks.

PR #3161, opened 2021-05-18
handshake codec

coot:
The following two changes make handshake codec behave the same way as
all the other codecs:

* make handshake decoder more strict: also check length of encoded data;
* formatting of decoder errors


coot:
Closing as it is included in #3165

Issue #3145, opened 2021-05-14
Figure out exactDeps, cabal 3.2 vs 3.4, source-repository-package stanzas, etc, etc


nfrisby:
TODO figure out a better Issue name

As of PR #3138, we're seeing build errors in _some_ of our build environments.

That PR will unblock downstream teams, so we may find a "whatever works for most people" option and go with that.

This Issue is to actually figure out the right way to handle the configuration.

Jasagredo:
`cabalWrapped` doesn't present the issues that plain `cabal` did. In this case it was not being able to find `happy:happy` coming from `plutus-core`, although `nix` was able to resolve the dependency graph.

`cabalWrapped` works fine.

nfrisby:
As of commit 82527e9, `exactDeps` does not affect the success of entering the shell or running `cabal build all` inside of it. We tried with letting it default (which we think is `false`) and with explicitly setting it to either value. All worked on that commit.

We think, in general, we want `exactDeps = true`. That would likely incur the fewest surprises during our normal work.

If you want a more capable/unrestrained package search, you can opt-in to that with `nix-shell --arg { exactDeps = false; }`, eg.

Also, we peeked at the `haskell.nix` source and it looks like `exactDeps` only affects the `nix-shell`.

PR #3169, opened 2021-05-20
Replace cabal with iohk-nix's cabalWrapped

EncodePanda:
This addresses issue #3145. In a nutshell, cabalWrapped will remove
source-repository-package stanzas while invoking cabal - this avoids
unnecessary rebuilds of dependencies that were already built by nix.

Fixes almost all issue from #3145 except for HLS

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -147,6 +147,12 @@ allow-newer:
   monoidal-containers:aeson,
   size-based:template-haskell
 
+
+-- ---------------------------------------------------------
+-- cabalWrapped - TODO: comment this
+-- --------------------------- 8< --------------------------
+-- Please do not put any `source-repository-package` clause above this line.
+
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/Win32-network
--- a/cabal.project
+++ b/cabal.project
@@ -143,6 +143,10 @@ package plutus-tx
 package prettyprinter-configurable
   tests: False
 
+allow-newer:
+  monoidal-containers:aeson,
+  size-based:template-haskell
+
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/Win32-network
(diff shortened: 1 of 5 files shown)


cabal.project
```diff
@@ -143,6 +143,16 @@ package plutus-tx
 package prettyprinter-configurable
   tests: False
 
+allow-newer:
+  monoidal-containers:aeson,
+  size-based:template-haskell
+
+
+-- ---------------------------------------------------------
+-- cabalWrapped - TODO: comment this
```
jbgi, line 152:
```suggestion
-- The "cabal" wrapper script provided by nix-shell will cut off / restore the remainder of this file
-- in order to force usage of nix provided dependencies for `source-repository-package`s.
```

jbgi (approved):

Jasagredo (approved):
Nice!

EncodePanda:
bors merge

Jasagredo:
HLS is also working as expected. :+1:

rvl:
This looks like a good workaround for a long-standing Haskell.nix and/or Cabal issue.
I'm also pleasantly surprised that HLS is also working as expected.

I will probably apply something similar to `cardano-wallet`. But first:
1. When using this, have you ever observed "concurrency" issues when starting nix-shell while a cabalWrapped is running, or when trying to edit `cabal.project`?
2. Has anyone added extra things to `cabal.project` below the `--- 8< ---` line yet? :grin:

nfrisby:
1. We're just a few days in, so that scenario has yet to arise, I think. I don't know how I'd do this, "when starting nix-shell while a cabalWrapped is running", since we only run the `cabalWrapped` exe from inside the `nix-shell`. And for "when trying to edit cabal.project", I don't think there'd be serious harm; I'm guessing the `cabalWrapped` exe only ever reads `cabal.project` (maybe I'm missing something?)
2. We're just a few days in, but it does seem inevitable. I can't remember which repo, but someone else has been doing this for much longer, if you want a real answer :)

Issue #3139, opened 2021-05-14
Shore up Alonzo era tech debt


nfrisby:
PR #3138 introduces some incompleteness:

  - Some of the `Cardano` generators used by serialization tests never generate the Alonzo summand
     - https://github.com/input-output-hk/ouroboros-network/pull/3138/files#diff-b1254cce14c6ba79bdba35bd1e5b3b4cfafde262aed1f2c734626804b6ee912bR81
     - https://github.com/input-output-hk/ouroboros-network/pull/3138/files#diff-b1254cce14c6ba79bdba35bd1e5b3b4cfafde262aed1f2c734626804b6ee912bR103
     - https://github.com/input-output-hk/ouroboros-network/pull/3138/files#diff-b1254cce14c6ba79bdba35bd1e5b3b4cfafde262aed1f2c734626804b6ee912bR226
     - https://github.com/input-output-hk/ouroboros-network/pull/3138/files#diff-b1254cce14c6ba79bdba35bd1e5b3b4cfafde262aed1f2c734626804b6ee912bR388
  - `examplesAlonzo` is `mempty`
    - https://github.com/input-output-hk/ouroboros-network/pull/3138/files?file-filters%5B%5D=.cabal&file-filters%5B%5D=.hs&file-filters%5B%5D=.json&file-filters%5B%5D=.project&file-filters%5B%5D=.yml#diff-fcfcb4e320429ae495bdbeea318dbe1d7c6f54de2ae40adbe7d6993cfb443ef2R320
   

This Issue is to track the tech debt taken on by that PR.

nfrisby:
We've closed this Issue. Issue #3179 is a continuation. It's more tech debt, but far less important. This Issue was the more urgent/important stuff.

PR #3170, opened 2021-05-20
Define ShelleyLedgerExamples and rework the shelley examples around it.

Jasagredo:
`ShelleyLedgerExamples` contains examples of the types that come entirely from the
ledger side and are required to create `Examples` which we were using on our
side for the Golden tests.

The actual values that are tested are unchanged.

Addresses #3139.

We have `Examples` which we want to keep, but those are mainly generated from ledger
types. Eventhough this PR looks like it changes a lot things, it is mostly just moving code around.
Aside from moving things around, the changes are basically defining this new
datatype (`ShelleyLedgerExamples`) and grouping all the ledger-specific functions at 
the top of the module, sometimes adapting them (e.g. `exampleHashHeader -> exampleHeaderHash`).

In a follow up PR, the section between `-- To Ledger START` and `-- To Ledger END` will 
be moved to cardano-ledger-specs.

The main functions that will need to be provided by ledger in the end are now named 
`ledgerExamplesX` where `X ∈ {Shelley,Allegra,Mary,Alonzo}`.

### DIFF
--- a/ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
+++ b/ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE FlexibleContexts         #-}
 {-# LANGUAGE NamedFieldPuns           #-}
 {-# LANGUAGE OverloadedStrings        #-}
+{-# LANGUAGE RecordWildCards          #-}
 {-# LANGUAGE ScopedTypeVariables      #-}
 {-# LANGUAGE TypeApplications         #-}
 {-# LANGUAGE TypeFamilies             #-}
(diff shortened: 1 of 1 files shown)


EncodePanda:
@nc6 to explain the motivation behind this PR, everything between `-- To Ledger START` and `-- To Ledger END` will be moved to a `ledger-spec` once this PR gets into master

EncodePanda (approved):
LGTM

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
```diff
@@ -3,38 +3,20 @@
 {-# LANGUAGE FlexibleContexts         #-}
 {-# LANGUAGE NamedFieldPuns           #-}
 {-# LANGUAGE OverloadedStrings        #-}
+{-# LANGUAGE RecordWildCards          #-}
```
nc6, line 6: We tend to avoid this one in the ledger, but maybe you guys are fine with it?

nc6 (approved):
Looks fine to me, and the stuff to move to the ledger looks straightforward enough

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Examples.hs
EncodePanda: greping for `RecordWildCards` on ouroboros gives 169 results :)
but i promise we will not move it to ledger :)

EncodePanda:
bors merge

Issue #3156, opened 2021-05-17
Update ledger-spec and bring back buildkite dependency check


EncodePanda:
Blocked by https://github.com/input-output-hk/cardano-ledger-specs/issues/2291

Once ledger-spec update their plutus dependency and bring back buildkite check that was removed here https://github.com/input-output-hk/ouroboros-network/pull/3138/commits/cd196cf8f3e6302923e55b22ccb1d1b12122c614

PR #3171, opened 2021-05-20
Update to the latest ledger-specs (and associated plutus)

EncodePanda:
Also bring back the dependency check on buildkite, which was removed
previously when we were using plutus that was not on master

Fixes https://github.com/input-output-hk/ouroboros-network/issues/3156

Co-authored-by: Javier Sagredo <javier.sagredo@iohk.io>


Jasagredo (approved):
LGTM

EncodePanda:
bors merge

PR #3172, opened 2021-05-21
Fix signature of `Control.Monad.Class.MonadSTM.Strict.stateTVar`

coot:
This PR requires a careful review.  There are various places where
```
stateTVar :: MonadSTM m => StrictTVar m s -> (s -> (s, s)) -> STM m s
```
is used, and thus the change made is not visible in types.

`IOLike` module in consensus exports `Control.Monad.Class.MonadSTM.Strict` so
this inlucdes all the places in consensus which use it.

- io-sim-classes: fix type signature of `Strict.stateTVar`
- Updated call sites of `Strict.stateTVar`

### DIFF
--- a/io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
+++ b/io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
@@ -138,10 +138,10 @@ writeTVar StrictTVar { tvar, invariant } !a =
 modifyTVar :: MonadSTM m => StrictTVar m a -> (a -> a) -> STM m ()
 modifyTVar v f = readTVar v >>= writeTVar v . f
 
-stateTVar :: MonadSTM m => StrictTVar m a -> (a -> (a, b)) -> STM m b
+stateTVar :: MonadSTM m => StrictTVar m s -> (s -> (a, s)) -> STM m a
 stateTVar v f = do
     a <- readTVar v
-    let (a', b) = f a
+    let (b, a') = f a
     writeTVar v a'
     return b
 
@@ -152,7 +152,7 @@ swapTVar v a' = do
     return a
 
 
-updateTVar :: MonadSTM m => StrictTVar m a -> (a -> (a, b)) -> STM m b
+updateTVar :: MonadSTM m => StrictTVar m s -> (s -> (a, s)) -> STM m a
 updateTVar = stateTVar
 {-# DEPRECATED updateTVar "Use stateTVar" #-}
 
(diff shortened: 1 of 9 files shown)


io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
```diff
@@ -131,14 +131,14 @@ writeTVar StrictTVar { tvar, invariant } !a =
 modifyTVar :: MonadSTM m => StrictTVar m a -> (a -> a) -> STM m ()
 modifyTVar v f = readTVar v >>= writeTVar v . f
 
-stateTVar :: MonadSTM m => StrictTVar m a -> (a -> (a, b)) -> STM m b
+stateTVar :: MonadSTM m => StrictTVar m a -> (a -> (b, a)) -> STM m b
```
nfrisby, line 134: Maybe we should use the same tyvar names as the `stm` package, `s -> (a, s)` while we're at it?

http://hackage.haskell.org/package/stm-2.5.0.1/docs/Control-Concurrent-STM-TVar.html#v:stateTVar

nfrisby (commented):
I went through the consensus bits, comparing this diff against the results of `git grep -C3 -e stateTVar -e updateTVar origin/master`. LGTM. Also, while inspecting that diff, I didn't see any combinators defined via `stateTVar`/`updateTVar` that has symmetric types; so that grep should cover the at-risk cases (eg updating a counter while also returning the old value)

I'm not Approving, only because I didn't review the Network side of things.

nfrisby:
We might also consider introducing and using a combinator dedicated to the usual symmetric case; something like:

```
newtype Old a = Old a

getAndUpdateTVar :: MonadSTM m => StrictTVar m s -> (s -> s) -> STM m (Old s)
```

That -- with or without the newtype -- might be overkill, but might also prevent some headaches.

nfrisby:
The CI failure is a timeout. I suspect that if you rebase onto master, you'll no longer time out. We recently disabled some warnings in an extremely slow module to avoid just this.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
coot: I think we should.

coot:
@nfrisby I am all for `getAndUpdateTVar` or something similar, but this is mostly in consensus code, so I'll leave it too you guys.

karknu (approved):
LGTM

coot:
bors merge

PR #3175, opened 2021-05-21
Top Level Query Versioning

DavidEichmann:
## Motivation

This is an attempt to figure out how to properly version the use of the new top level `Query blk` type. In the original PR #3106, `Query blk` was added, but serialization did not change and so no version bump was necessary. This PR ispreparation for adding a `GetPartialLedgerConfig :: Query blk` constructor, which will require correct versioning of top level `Query blk` queries.

## Approach

I've split this into sensible commits, but the key changes are:

  1. Add `QueryVersion` type [code](https://github.com/input-output-hk/ouroboros-network/pull/3175/files#diff-fc3cd7a4e00ee64cb67d5161db6722f250329f502dce3c97d2f10cda24729a41)
  2. `supportedNodeToClientVersions` now map to `QueryVersion` as well as
     `BlockNodeToClientVersion blk` [code](https://github.com/input-output-hk/ouroboros-network/pull/3175/files#diff-300c6e56bba1985bcb3335523928444c719cff653f05db0079f31f0692756c2bR48).
  3. `class SerialiseNodeToClient`'s functions now take `QueryVersion` as
     well as `BlockNodeToClientVersion blk` [code](https://github.com/input-output-hk/ouroboros-network/pull/3175/files#diff-3a76315ea776b906995ea9a2ed0caf423a42d2a9b2791ee9b54ca231d5cbcba4L36-R76).
  4. The serialization of `Query blk` now looks at `QueryVersion`
     and either de/encodes `BlockQuery blk` directly (equivalent to the current serialization) or uses a tagged union
     de/encoding for `Query blk` [code](https://github.com/input-output-hk/ouroboros-network/pull/3175/files#diff-7a9bca2632645104c60e31ddb8bfdb1b4ce4acb413fed51fdf7057ec9d27b1fcL26-R68).
  5. Add a new `NodeToClientV_10 :: NodeToClientVersion` for the addition of the top level query [code](https://github.com/input-output-hk/ouroboros-network/pull/3175/files#diff-234887b587438e0586caa2bada40d28527a92659cd9afec9208cb38d83386d8cR44).
  6. Update the `supportedNodeToClientVersions` map for `CardanoBlock c` with support for `QueryV_1` [code](https://github.com/input-output-hk/ouroboros-network/pull/3175/files#diff-9dfcc4a2e75e87f02faf4beb2f3442a380208661db085f2b010f7764ff16bd46R387).


DavidEichmann:
@nfrisby This PR is really just aimed at the versioning of top level query stuff. We discussed this in a previous meeting and this PR reflects my understanding of that meeting. I wasn't entirely sure if my understand was correct, so I'd appreciate some feedback on the general approach here. Have a great weekend :smile:

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
```diff
@@ -41,6 +41,8 @@ data NodeToClientVersion
     -- ^ 'LocalStateQuery' protocol codec change, allows to acquire tip point.
     | NodeToClientV_9
     -- ^ enabled @CardanoNodeToClientVersion7@, i.e., Alonzo
+    | NodeToClientV_10
+    -- ^ enable @QueryV_1@, i.e. top level queries.
```
nfrisby, line 45: I know this PR is currently supposed to be minimal, and so it doesn't actually add a top-level query. But I think we should prefer actually adding a query before `V_10` gets released. It'd be confusing to genuinely allocate a version that provides no actual downstream enrichment.

The actionable thing I'm suggesting probably requires some cooperative planning, eg involving the following events a) merging this PR b) merging the next PR that actually justifies a version bump c) updating downstream libraries' dependencies to involve those PRs we will have merged.

I had originally thought that (c) would be about making a proper release, but that's a little narrower than truth, unless we're willing to get a little clever here https://github.com/input-output-hk/ouroboros-network/blob/eb02c2d627b728a48449d314e56c347f1798503b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L717-L718

(This is probably hard to follow; consider it a mixture of pointing out a concern and making a note to my future self.)

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query/Version.hs
```diff
@@ -0,0 +1,8 @@
+
+module Ouroboros.Consensus.Ledger.Query.Version (QueryVersion (..)) where
+
+-- | Version of the `Query blk` type.
+data QueryVersion
+  = TopLevelQueryDisabled
+  | QueryV_1
```
nfrisby, line 7: Please add some comments to each constructor. (My other comment on this PR reveals my hope that we'll need to update the comment on `QueryV_1` when we subsequently enrich it to actually provide a top-level query.)

ouroboros-consensus/src/Ouroboros/Consensus/Node/NetworkProtocolVersion.hs
```diff
@@ -44,7 +45,7 @@ class HasNetworkProtocolVersion blk => SupportedNetworkProtocolVersion blk where
 
   -- | Enumerate all supported node-to-client versions
   supportedNodeToClientVersions
-    :: Proxy blk -> Map NodeToClientVersion (BlockNodeToClientVersion blk)
+    :: Proxy blk -> Map NodeToClientVersion (QueryVersion, BlockNodeToClientVersion blk)
```
nfrisby, line 48: Just as with my other comment on `SerialiseNodeToClient`, I'm also thinking that this class shouldn't be related to `QueryVersion`.

This one is a bit murkier, I think, because of the kind of inconsistent way we're reusing the single `NodeToClientVersion` value for multiple `blk` types. We're very much evolving `NodeToClientVersion` itself as we add features on `mainnet`, ie `blk ~ CardanoBlock`. But we have instances for older block types too, eg `ByronBlock`. This is a mishmash, which is likely only fine so far because we're only ever really running this code with non-trivial versioning scenarios when `blk` is in fact `CardanoBlock`.

Anyhow: since, as I argued for `SerialiseNodeToClient`, no NTC serializations will depend on the `QueryVersion` except the serialization of `Query` itself, I think we can just have the `Query` serialization be the only place that `QueryVersion` is used, and moreover it can internally calculate the `QueryVersion` value as a direct `blk`-independent function of `NodeToClientVersion`.

Does this sound worth it? I can try to further sell both of these changes to you by pointing out that so many of the source files in this diff would no longer need any changes if we don't change either of these classes :P

nfrisby (changes requested):
I made a couple high-level comments that I'm sad to say would lead to a fair amount of undoing, if you find them persuasive. At least they would lead to a smaller diff, if I'm right :)

I'm Requesting Changes only to make it clear that I want you and I to reach an agreement regarding those suggestions before we proceed. In particular, other than those suggestions, I think the content here is (at least  very nearly) mergeable as-is. Thanks in particular for the careful style adherence 👍

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query/Version.hs
DavidEichmann: Good idea!

ouroboros-consensus/src/Ouroboros/Consensus/Node/Serialisation.hs
```diff
@@ -71,21 +72,21 @@ class SerialiseNodeToNode blk a where
 -- | Serialise a type @a@ so that it can be sent across the network via
 -- node-to-client protocol.
 class SerialiseNodeToClient blk a where
-  encodeNodeToClient :: CodecConfig blk -> BlockNodeToClientVersion blk -> a -> Encoding
-  decodeNodeToClient :: CodecConfig blk -> BlockNodeToClientVersion blk -> forall s. Decoder s a
+  encodeNodeToClient :: CodecConfig blk -> QueryVersion -> BlockNodeToClientVersion blk -> a -> Encoding
+  decodeNodeToClient :: CodecConfig blk -> QueryVersion -> BlockNodeToClientVersion blk -> forall s. Decoder s a
```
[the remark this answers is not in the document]
nfrisby: Hmmm. A scenario occurred to me that is more concrete than what I had been thinking about so far.

Suppose we add the `GetPartialLedgerConfig` query in `QueryV_1`. Today, that query would return:

```
data ShelleyPartialLedgerConfig era = ShelleyPartialLedgerConfig {
      shelleyLedgerConfig    :: !(ShelleyLedgerConfig era)
    , shelleyTriggerHardFork :: !TriggerHardFork
    }

data ShelleyLedgerConfig era = ShelleyLedgerConfig {
      shelleyLedgerCompactGenesis     :: !(CompactGenesis era)
    , shelleyLedgerGlobals            :: !SL.Globals
    , shelleyLedgerTranslationContext :: !(Core.TranslationContext era)
    }
```

But note that we added the `shelleyLedgerTranslationContext` field as part of Alonzo. So that serialization would have had to change different if this top-level query had existed before we integrated Alonzo.

Does the `BlockNodeToClientVersion blk` argument to `encodeNodeToClient` provide enough information for us to have done that? Or would `QueryVersion` be relevant somehow?

More generally, what exactly does `QueryVersion` constrain/reflect? And how does its evolution relate to the evolution of `BlockNodeToClientVersion`?

I'm hoping you already have more clarity on this than I do and can share it with me, since it feels too nebulous right now.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/ByronCompatibility.hs
```diff
@@ -12,6 +12,8 @@
 {-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
 module Test.Consensus.Cardano.ByronCompatibility (tests) where
 
+import qualified Data.Map.Strict as Map
+
```
nfrisby, line 16: Style: this import should be part of the group that is right now just below it.

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
```diff
@@ -212,6 +216,27 @@ data WithVersion v a = WithVersion v a
 -- @('WithVersion' v a)@.
 type ArbitraryWithVersion v a = (Arbitrary (WithVersion v a), Eq a, Show a)
 
+instance ( Arbitrary (WithVersion (BlockNodeToClientVersion blk) (SomeSecond BlockQuery blk))
+         , SupportedNetworkProtocolVersion blk
```
nfrisby, line 220: The code seems to compile without `SupportedNetworkProtocolVersion blk`. Do you think there's a reason to require it anyway?

nfrisby (changes requested):
I asked for some "simplifications by removal" :)

I Requesting Changes again partly for that, and partly also because we're still having the ongoing discussion on Slack about what exactly `QueryVersion` means/is it necessary.

nfrisby:
I have some ideas/proposals.

Context

  - Part of the purpose of `BlockNodeToClientVersion` is to provide more structure than just the enumeration that is `NodeToClientVersion`. EG The HFC aspects of it. This EG lets the serializers' sensitivity use clearer pattern matches **_and_** it serves as a mediator between the serialization logic and the underlying enumeration.
  
  - Ultimately, that mediation is chosen by `cardano-node`, since they're the ones who invoke `Ouroboros.Consensus.Node.run`. (Note the "`cardano`" of the name; they are thus choosing how the Cardano chain interprets the `NodeToClientVersion` number.) 
  
  - For more details, see the `llrnNodeToClientVersions` field in `Ouroboros.Consensus.Node`, which _defaults_ to whatever is in the relevant `SupportedNetworkVersions` instance -- but can be overridden. https://github.com/input-output-hk/ouroboros-network/blob/ba062897f3ef1859df7dcfb4b284722b08acdd0c/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs#L195-L199
  
I think those observations are important: the Consensus code is general (eg `BlockNodeToClientVersion` is basically several (mostly?) independent flags), but the `cardano-node` code is specific to a particular interpretation of the `NodeToClientVersion` enumeration.

Proposal

Now that we're adding `blk`-independent queries, `QueryVersion` serves the same purpose as a mediator. Thus I think we should keep it, as introduced in this PR.

Once it has a non-trivial value, we should also add a `llrnQueryVersions` field, a la `llrnNodeToClientVersions`. (I'm not sure how we should express its default; a `blk` indexed class does seem wrong for this one.)

Qualifiers

Depending on the response to https://github.com/input-output-hk/ouroboros-network/pull/3175#discussion_r643574103 we might not ultimately need any versioning for NTC queries.

Also, this still doesn't resolve the non-obvious balance of responsibilities between `QueryVersion` and `BlockNodeToClientVersion`. I'm hoping that will become clearer as we start adding concrete top-level queries.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Serialisation.hs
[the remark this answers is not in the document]
DavidEichmann: My understanding is that we DO aim to support clients that are on an older `NodeToClientVersion` than the node. Hence the existence of `NodeToClientVersion` does make sense.

In practice the client supports a single version and hopes that the node is backwards compatible with it. Often the node *is* backwards compatible. For example, going from version 3 to 4 we've added some new queries. In a way, this seems incompatible because the client can't decode responses to the newer queries. Actually, this *is* compatible because the older client will never use any of the new queries and so will never need to decode responses to them. The fact that these things are compatible requires understanding of how the protocol works. We don't get static checks for this compatibility i.e. via some sort of "migration" code. The lack of static checks is what makes things confusing. At least we have some golden tests to check that our serialization code matches serializations from older versions though I doubt this implies much confidence that our versioning code is correct.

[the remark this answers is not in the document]
DavidEichmann: On that note, we already, in a way, allow backwards compatibility in the other direction: "newer" clients are compatible  with "older" nodes, but only if the newer client uses an older `NodeToClientVersion`. In that case it is that client's responsibility to avoid all functionality of newer versions even if the code (e.g. new query constructors) is available.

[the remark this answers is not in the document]
nfrisby: Suppose a year from now we introduce a new era.

  - Release X: we have the `GetLedgerConfig :: Query (LedgerConfig blk) ` query  and the last era is `Foo`.
  - Release X+1: we still have the `GetLedgerConfig :: Query (LedgerConfig blk)` query  and the last era is `Bar`, the successor of `Foo`.

If the client is running X+1 and the server is running X, then the server will send the serialization of a `LedgerConfig blk` that excludes `Bar` but the client will try to deserialize that into a `LedgerConfig blk` that includes `Bar`.

I don't see how the newer client in that case could (easily & cleanly) avoid the difference other than not sending the `GetLedgerConfig` query, which seems too severe.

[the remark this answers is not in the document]
nfrisby: I just looked over the types of the existing Shelley queries, wondering "how do they avoid the product semantics challenge?"

https://github.com/input-output-hk/ouroboros-network/blob/e50613562d6d4a0f933741fcf590b0f69a1eda67/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs#L82

All of the result types involve a _single_ `era`. So I think they all follow the sum semantics, not the product semantics. ... Maybe for `GetGenesisConfig` we're still going to have to figure out what to do for the Alonzo bit? I'm not quite sure what `GetGenesisConfig` is used for, and whether it needs to include the new Alonzo "genesis data".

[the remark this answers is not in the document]
nfrisby: We had a call that included Duncan. In particular, he clarified that we do intend for newer nodes to support older clients, for the convenience of users doing incremental upgrades (updating the node first, since that's more important).

We three also concluded that as long as we avoid product semantics in the query arguments, this should be fine; it's a continuation of our practices so far: eg the node would truncate its HFC era list when serializing query results according to the negotiated `NodeToClientVersion`.

We do not anticipate _needing_ query with HFC-product arguments. But if we're wrong about that, we do have three main options in mind.

  - Option 1: decompose the hypothetical query into queries with a sum argument and therefore force the client to send multiple queries and assemble their results as it sees fit
  
  - Option 2: have the node deserializer fill-in the missing product components. By a parametricity-like argument, the fill value shouldn't matter. (We're not excited by this option.)
  
  - Option 3: disallow the query until the client catches up to the node, enforced via the negotiated version. (We're also not excited about this, since we generally don't want a node update to cause feature degradation in a client --- it's unlikely to be graceful degradation.)

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
DavidEichmann: Ah yes. This is a vestigial change. I'll remove it.

DavidEichmann:
Update

* Style fixes
* Documentation on `Query` regarding what corresponds to what version type
* Removed some added `SupportedNetworkProtocolVersion` constraints and instances
* Rename `fromNodeToClientVersion` to `nodeToClientVersionToQueryVersion`. This is a bit awkward with the word "To" appearing twice, but I think it's less awkward than trying to work around that.

DavidEichmann:
@nfrisby I'll let you trigger the merge in case you want to have a final look over this  / my latest changes.

nfrisby (approved):
I rebased onto master. It was automatic. But that's why GH lists me on the commits too now.

Approving. Thanks again for working this all out with me.

I also confirmed the tests files are all just renames:

```
$ git diff origin/master origin/davide/top-level-query-versioning  --name-status | grep -Ev -e '^R'
M	ouroboros-consensus-byron-test/ouroboros-consensus-byron-test.cabal
M	ouroboros-consensus-byron-test/test/Test/Consensus/Byron/Golden.hs
M	ouroboros-consensus-cardano-test/ouroboros-consensus-cardano-test.cabal
M	ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Golden.hs
M	ouroboros-consensus-shelley-test/ouroboros-consensus-shelley-test.cabal
M	ouroboros-consensus-shelley-test/test/Test/Consensus/Shelley/Golden.hs
M	ouroboros-consensus-test/src/Test/Util/Orphans/Arbitrary.hs
M	ouroboros-consensus-test/src/Test/Util/Serialisation/Golden.hs
M	ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
M	ouroboros-consensus/ouroboros-consensus.cabal
M	ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
A	ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query/Version.hs
M	ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToClient.hs
$
```

PR #3177, opened 2021-05-24
Remove ShelleyLedgerExamples and update cardano-ledger-specs

Jasagredo:
As by [cardano-ledger-specs' PR #2298](https://github.com/input-output-hk/cardano-ledger-specs/pull/2298), the code that was isolated previously is
being moved to the ledger side. This PR updates the ledger-spec dependency
commit to the head of that branch and removes the relevant code.

As a side change, `StandardCrypto` was defined in `Ouroboros.Consensus.Shelley.Protocol.Crypto` as
well as in `Cardano.Ledger.Crypto`, with the same verbatim declaration. Now that
we generate data on ledger's side we are using their `StandardCrypto` definition
which is incompatible with ours, so in order to fix type errors, our definition
was removed in favor of theirs.

This shouldn't mean any real change because as stated in their code (see the
comment removed in their [PR#2298](https://github.com/input-output-hk/cardano-ledger-specs/pull/2298), both definitions MUST have been (and remain)
equal at all times.

Note that the two first commits are related to this PR and the last one updates 
ledger-specs and plutus to the correct commits, and fixes imports.

This addresses #3139


nfrisby (approved):
LGTM, nice 👍

PR #3178, opened 2021-05-24
Incorporate Alonzo summands to CardanoBlock generators

Jasagredo:
This PR builds on top of #3177. It will be set to merge into master once that one is already in.

Adresses the first bullet point in #3139.

Also `CanMockPreAlonzo` is now redundant so it was removed.

### DIFF
--- a/ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
+++ b/ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
@@ -39,7 +39,7 @@ import           Test.Cardano.Ledger.AllegraEraGen ()
 import           Test.Cardano.Ledger.Alonzo.AlonzoEraGen ()
 import           Test.Cardano.Ledger.MaryEraGen ()
 import           Test.Cardano.Ledger.ShelleyMA.Serialisation.Generators ()
-import           Test.Consensus.Shelley.MockCrypto (CanMock, CanMockPreAlonzo)
+import           Test.Consensus.Shelley.MockCrypto (CanMock)
 import           Test.Shelley.Spec.Ledger.ConcreteCryptoTypes as SL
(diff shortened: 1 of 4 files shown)


EncodePanda (approved):
LGTM

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
```diff
@@ -55,21 +56,22 @@ import           Test.Shelley.Spec.Ledger.Serialisation.Generators ()
...
 
-instance CanMockPreAlonzo era => Arbitrary (Header (ShelleyBlock era)) where
+instance CanMock era => Arbitrary (Header (ShelleyBlock era)) where
   arbitrary = getHeader <$> arbitrary
 
 instance SL.Mock c => Arbitrary (ShelleyHash c) where
   arbitrary = ShelleyHash <$> arbitrary
 
-instance CanMockPreAlonzo era => Arbitrary (GenTx (ShelleyBlock era)) where
+instance (CanMock era , Arbitrary (Core.Witnesses era))
```
nfrisby, line 73: Is there anything preventing us from also moving `Arbitrary (Core.Witnesses era)` to `CanMock`? IIUC, it should hold for every Shelley-based era.

nfrisby (approved):
LGTM. Glad to see these simplifications falling into place, as eg the ledger instances a fully fleshed out.

One comment, but it's rather minor.

ouroboros-consensus-shelley-test/src/Test/Consensus/Shelley/Generators.hs
Jasagredo: Moved to `CanMock` and everything builds fine. Nice suggestion.

Jasagredo:
Rebased on top of `master`

PR #3184, opened 2021-05-25
Mempool: Fix pureSyncWithLedger.

### LINKED ISSUE #3181: Fixup mempool ledger sync function
When PR #3116 merged, we mixed up the order of operations.

- `TraceMempoolRemoveTxs` is reporting the size from before the removal
- `pureSyncWithLedger` is returning the snapshot from before the removal

Those values should be constructed from the state _after_ the removal.

Jasagredo:
Closes #3181 

The returned snapshot and the traced event were referring to the old state
instead of the new state.


nfrisby (approved):
LGTM, thanks! @Jasagredo said on Slack that he did look for similar accidental rearrangements in the other methods, but they all looked right.

PR #3185, opened 2021-05-25
CAD-3027 Merging p2p-master with master (Diffusion layer factorization)

bolt12:
This PR factors the `Diffusion` Layer in `P2P` and `NonP2P` and makes the necessary changes in both `ouroboros-network` and `ouroboros-consensus` repos, in order to enable a switch between starting a node as `p2p` or `non-p2p`

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -134,6 +134,7 @@ library
                        Ouroboros.Consensus.Node
                        Ouroboros.Consensus.Node.DbLock
                        Ouroboros.Consensus.Node.DbMarker
+                       Ouroboros.Consensus.Node.ErrorPolicy
                        Ouroboros.Consensus.Node.RethrowPolicy
                        Ouroboros.Consensus.Node.Exit
                        Ouroboros.Consensus.NodeId
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -23,13 +23,12 @@ module Ouroboros.Consensus.Node (
   , stdVersionDataNTC
   , stdVersionDataNTN
   , stdWithCheckedDB
+    -- ** P2P Switch
+  , NetworkP2PMode (..)
     -- * Exposed by 'run' et al
   , ChainDB.RelativeMountPoint (..)
   , ChainDB.TraceEvent (..)
   , ChainDbArgs (..)
-  , ConnectionId (..)
-  , DiffusionArguments (..)
-  , DiffusionTracers (..)
   , HardForkBlockchainTimeArgs (..)
   , LastShutDownWasClean (..)
   , LowLevelRunNodeArgs (..)
@@ -57,25 +56,30 @@ import           Data.Hashable (Hashable)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Typeable (Typeable)
-import           Data.Void (Void)
 import           System.FilePath ((</>))
 import           System.Random (StdGen, newStdGen, randomIO, randomRIO)
 
 import           Control.Monad.Class.MonadTime (MonadTime)
 import           Control.Monad.Class.MonadTimer (MonadTimer)
-
 import           Ouroboros.Network.BlockFetch (BlockFetchConfiguration (..))
-import           Ouroboros.Network.Diffusion
+import           Ouroboros.Network.Diffusion (DiffusionApplications,
+                     DiffusionArguments, DiffusionTracers, daDiffusionMode,
+                     getDiffusionArguments, mkDiffusionApplicationsNonP2P,
+                     mkDiffusionApplicationsP2P, runDataDiffusion)
 import           Ouroboros.Network.Magic
-import           Ouroboros.Network.NodeToClient (LocalAddress,
-                     NodeToClientVersionData (..))
+import           Ouroboros.Network.NodeToClient (ConnectionId, LocalAddress,
+                     NodeToClientVersionData (..), combineVersions,
+                     simpleSingletonVersions)
 import           Ouroboros.Network.NodeToNode (DiffusionMode,
-                     MiniProtocolParameters (..), NodeToNodeVersionData (..),
-                     RemoteAddress, combineVersions,
+                     MiniProtocolParameters, NodeToNodeVersionData (..),
+                     RemoteAddress, blockFetchPipeliningMax,
                      defaultMiniProtocolParameters)
+import           Ouroboros.Network.PeerSelection.LedgerPeers
+                     (LedgerPeersConsensusInterface (..))
 import           Ouroboros.Network.PeerSelection.PeerMetric (PeerMetrics (..),
                      newPeerMetric, reportMetric)
 import           Ouroboros.Network.Protocol.Limits (shortWait)
+import           Ouroboros.Network.RethrowPolicy
 
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.BlockchainTime hiding (getSystemStart)
@@ -89,6 +93,7 @@ import qualified Ouroboros.Consensus.Network.NodeToClient as NTC
 import qualified Ouroboros.Consensus.Network.NodeToNode as NTN
 import           Ouroboros.Consensus.Node.DbLock
 import           Ouroboros.Consensus.Node.DbMarker
+import           Ouroboros.Consensus.Node.ErrorPolicy
 import           Ouroboros.Consensus.Node.InitStorage
 import           Ouroboros.Consensus.Node.NetworkProtocolVersion
 import           Ouroboros.Consensus.Node.ProtocolInfo
@@ -138,6 +143,8 @@ data RunNodeArgs m addrNTN addrNTC blk = RunNodeArgs {
                        -> NodeKernel m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
                        -> m ()
 
+      -- | Network P2P Mode switch
+    , rnEnableP2P :: NetworkP2PMode
     }
 
 -- | Arguments that usually only tests /directly/ specify.
@@ -184,10 +191,12 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk = L
     , llrnRunDataDiffusion ::
            ResourceRegistry m
         -> DiffusionApplications
-             addrNTN        addrNTC
-             versionDataNTN versionDataNTC
+             addrNTN
+             addrNTC
+             versionDataNTN
+             versionDataNTC
              m
-        -> m Void
+        -> m ()
 
     , llrnVersionDataNTC :: versionDataNTC
 
@@ -203,12 +212,18 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk = L
     , llrnMaxClockSkew :: ClockSkew
     }
 
+-- | P2P Switch
+--
+data NetworkP2PMode = EnabledP2PMode
+                    | DisabledP2PMode
+  deriving (Eq, Show)
+
 -- | Combination of 'runWith' and 'stdLowLevelRunArgsIO'
 run :: forall blk.
      RunNode blk
   => RunNodeArgs IO RemoteAddress LocalAddress blk
   -> StdRunNodeArgs IO blk
-  -> IO Void
+  -> IO ()
 run args stdArgs = stdLowLevelRunNodeArgsIO args stdArgs >>= runWith args
 
 -- | Start a node.
@@ -224,7 +239,7 @@ runWith :: forall m addrNTN addrNTC versionDataNTN versionDataNTC blk.
      )
   => RunNodeArgs m addrNTN addrNTC blk
   -> LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
-  -> m Void
+  -> m ()
 runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
 
     llrnWithCheckedDB $ \(LastShutDownWasClean lastShutDownWasClean) ->
@@ -298,6 +313,7 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
       let ntnApps = mkNodeToNodeApps   nodeKernelArgs nodeKernel peerMetrics
           ntcApps = mkNodeToClientApps nodeKernelArgs nodeKernel
           diffusionApplications = mkDiffusionApplications
+                                    rnEnableP2P
                                     (miniProtocolParameters nodeKernelArgs)
                                     ntnApps
                                     ntcApps
@@ -321,7 +337,14 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
       -> NodeKernel     m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
       -> PeerMetrics m addrNTN
       -> BlockNodeToNodeVersion blk
-      -> NTN.Apps m (ConnectionId addrNTN) ByteString ByteString ByteString ByteString ByteString ()
+      -> NTN.Apps m
+          (ConnectionId addrNTN)
+          ByteString
+          ByteString
+          ByteString
+          ByteString
+          ByteString
+          ()
     mkNodeToNodeApps nodeKernelArgs nodeKernel peerMetrics version =
         NTN.mkApps
           nodeKernel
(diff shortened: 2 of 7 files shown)


coot (commented):
I see that one part of complexity that is pushed onto `ouroboros-consensus` comes from the difference between `DiffusionApplications` for non p2p and p2p versions, there are two parts that contribute to this:

1. different types in application fields
2. different auxiliaty members, e.g. `daErrorPolicies` vs `daRethrowPolicy`, and various additional fields in `p2p`: e.g. `daMiniProtocolParameters`, `daPeerMetrics`, `daBlockFetchMode`

To address 1. it could help if we made the non-p2p to use the same types as the p2p version
To address 2. we could either split the `DiffusionArguments` or provide a polymorphic member which would be instantiated to different types for p2p and non p2p. Whatever would make the use of the api simpler.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -226,7 +226,7 @@ run :: forall blk.
      RunNode blk
   => RunNodeArgs IO RemoteAddress LocalAddress blk
   -> StdRunNodeArgs IO blk
-  -> IO Void
+  -> IO (Either () Void)
```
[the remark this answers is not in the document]
bolt12: I tried to keep it as `Void` but it is not possible... so, in order to make the least decisions possible I just kept everything as `Either (nonp2p type) (p2p type)`

bolt12:
> I see that one part of complexity that is pushed onto `ouroboros-consensus` comes from the difference between `DiffusionApplications` for non p2p and p2p versions, there are two parts that contribute to this:
> 
>     1. different types in application fields
> 
>     2. different auxiliaty members, e.g. `daErrorPolicies` vs `daRethrowPolicy`, and various additional fields in `p2p`: e.g. `daMiniProtocolParameters`, `daPeerMetrics`, `daBlockFetchMode`
> 
> 
> To address 1. it could help if we made the non-p2p to use the same types as the p2p version
> To address 2. we could either split the `DiffusionArguments` or provide a polymorphic member which would be instantiated to different types for p2p and non p2p. Whatever would make the use of the api simpler.

I tried to make the least decisions possible and to uniform everything between p2p and non-p2p. However, I figured the difference in the types would be too big and the needed modifications would be too complex. I like the way I have things because I believe it will make it easier to revert back once we decide to only keep the P2P part!

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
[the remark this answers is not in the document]
coot: I'd suggest to use `()` instead, we'll switch to `Void` when it will be possible.

coot:
> I tried to make the least decisions possible and to uniform everything between p2p and non-p2p. However, I figured the difference in the types would be too big and the needed modifications would be too complex. I like the way I have things because I believe it will make it easier to revert back once we decide to only keep the P2P part!

The most difficult part of this would be to modify the `non-p2p` part.  I am not 100% if this would work, but I'd like to experiment with it.  If it does work, we should be able to have PR to `master` which prepares that branch for merging `p2p-master` (it would include some other commits from `p2p-master` branch like  364a18ce6 and cb54ad7c2).  The modifications of `p2p`, i.e. the split of `DiffusionApplications` could stay with us for longer.

bolt12:
Hmm, I guess I can try open a PR against master that tries to achieve that, it might end up being simpler than I am imagining! 😄

coot:
> Hmm, I guess I can try open a PR against master that tries to achieve that, it might end up being simpler than I am imagining! smile

To make sure we don't have a complex conflict, make  a PR against `p2p-master` first, later we'll try to rewrite the history of `p2p-master` so this lands at the bottom of it (this is where a conflict might surface).

bolt12:
@coot `P2P` and `Non-P2P` `runDataDiffusion` functions are very different, I was able to refactor `DiffusionTracers` and make everything run on `IO ()`, but `DiffusionApplications` and `DiffusionArguments` are very difficult to refactor, there are a lot of implementation details in each one, so changing the `non-p2p` part would have to introduce quite some of the `p2p` features. Do you have any suggestion or guidance that you can provide? I still think that making a clear distinction between the `p2p` and `non-p2p` for now is the simplest and cleanest approach, whenever we decide to make the full switch to `p2p`, we just need to remove all the `NonP2P.x` things

bolt12:
**NOTE**: Before merging make sure that we are using the most recent versions of `P2P` ( `p2p-master` branch ) and `NonP2P` ( `master` branch ), and if not update the code.

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
```diff
@@ -0,0 +1,1260 @@
...
+
+    -- | /node-to-client/ rethrow policy
+    --
+    , daLocalRethrowPolicy :: RethrowPolicy
+
+    , daPeerMetrics    :: PeerMetrics m ntnAddr
+
+      -- | Used by churn-governor
+      --
+    , daBlockFetchMode :: STM m FetchMode
```
coot, line 335: Let's align the types (`::`).

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
```diff
@@ -0,0 +1,1260 @@
...
+--
+-- Governor type aliases
+--
+
+type NodeToNodePeerConnectionHandle (mode :: MuxMode) a =
+    PeerConnectionHandle
+      mode
+      SockAddr
+      ByteString
+      IO () a
```
coot, line 466: This belongs to the previous section, but it's a mistake in the `p2p-master`, so let's leave it for now.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -1,1320 +1,406 @@
...
+        (dargs { daP2P = da })
+        (dapps { dapP2P = dapp })
+    (Right t, Right da, Right dapp) ->
+      void
+        $ P2P.runDataDiffusion
+            (tr { dtP2P = t})
+            (dargs { daP2P = da })
+            (dapps { dapP2P = dapp })
+    _                               ->
+      error "Non-matching arguments, every argument should be on the same side!"
```
coot, line 406: We could eliminate it with a type level tag
```
data P2PMode = P2P | NonP2P
```
And then index all types with either `P2P` or `NonP2P`.  But maybe it is not worth it, I'll leave it up to. 

Hydra guys are also using data diffusion interface as well, we might need to help them to make a transition.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -1,1320 +1,406 @@
...
+  , mkDiffusionTracersP2P
+  , DiffusionArguments
+  , getDiffusionArguments
+  , daDiffusionMode
+  , mkDiffusionArgumentsNonP2P
+  , mkDiffusionArgumentsP2P
+  , DiffusionApplications
+  , mkDiffusionApplicationsP2P
+  , mkDiffusionApplicationsNonP2P
   , runDataDiffusion
```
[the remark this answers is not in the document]
bolt12: good idea!

coot (changes requested):
Very nice, some cosmetic remarks.

ouroboros-network/src/Ouroboros/Network/Diffusion/Common.hs
```diff
@@ -0,0 +1,205 @@
...
+      -- | Diffusion initialisation tracer
+    , dtDiffusionInitializationTracer
+        :: Tracer IO DiffusionInitializationTracer
+
+      -- | Ledger Peers tracer
+    , dtLedgerPeersTracer
+        :: Tracer IO TraceLedgerPeers
+
+      -- | P2P or NonP2P DiffusionTracers
+    , dtP2P :: p2p
```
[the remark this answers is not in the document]
bolt12: sounds good to me! :D

ouroboros-network/src/Ouroboros/Network/Diffusion/NonP2P.hs
```diff
@@ -0,0 +1,500 @@
...
+                  ( DiffusionTracers(..)
+                  , DiffusionArguments(..)
+                  , DiffusionApplications(..)
+                  , DiffusionInitializationTracer(..)
+                  , DiffusionFailure(..)
+                  )
+
+-- | NonP2P DiffusionTracers Extras
+--
+data DTracersExtra = DTracersExtra {
```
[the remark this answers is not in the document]
bolt12: Hm, between `DiffusionTracersExt` and `DiffusionTracersExtra`, I prefer the latter, if it is okay having long names

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
bolt12: Got it!

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
bolt12: I don't think its worth it, the smart constructors will take care of everything and if someone uses the wrong smart constructors, I think the compiler can not type check

bolt12:
Thanks @coot , will address your reviews!

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -345,51 +384,87 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
...
+           (consensusErrorPolicy (Proxy @blk))
+      where
+        initiator =
+          combineVersions
+           [ simpleSingletonVersions
+               version
+               llrnVersionDataNTN
+               (NTN.initiator miniProtocolParams version $ ntnApps blockVersion)
+           | (version, blockVersion) <- Map.toList llrnNodeToNodeVersions
+           ]
```
coot, line 447: I think consensus is using 2 space for indentation, rather than one.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -345,51 +384,87 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
...
+        responder =
+          combineVersions
+           [ simpleSingletonVersions
+              version
+              llrnVersionDataNTC
+              (NTC.responder version $ ntcApps blockVersion version)
+           | (version, blockVersion) <- Map.toList llrnNodeToClientVersions
+           ]
```
coot, line 464: The same here.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -102,6 +96,17 @@ import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.Orphans ()
 import           Ouroboros.Consensus.Util.ResourceRegistry
 import           Ouroboros.Consensus.Util.Time (secondsToNominalDiffTime)
+import           Ouroboros.Network.Diffusion (DiffusionApplications,
```
nfrisby, line 99: Please move these up with the other `Ouroboros.Network.*` imports.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -222,10 +237,11 @@ runWith :: forall m addrNTN addrNTC versionDataNTN versionDataNTC blk.
...
      , Hashable addrNTN, Ord addrNTN, Typeable addrNTN
      )
-  => RunNodeArgs m addrNTN addrNTC blk
+  => NetworkP2PMode
+  -> RunNodeArgs m addrNTN addrNTC blk
   -> LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
-  -> m Void
-runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
+  -> m ()
+runWith enableP2P RunNodeArgs{..} LowLevelRunNodeArgs{..} =
```
nfrisby, line 244: Please add `enableP2P` as a field in `RunNodeArgs` instead of adding it as a new separate parameter to `run` and `runWith`.

nfrisby: As a field, you can give it a polarized name, so it can just be a `Bool` if you wish. eg `rnEnableP2P :: Bool`.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -338,58 +362,96 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
...
+      case enP2P of
+        EnabledP2PMode ->
+          mkDiffusionApplicationsP2P
+            initiator
+            initiatorAndResponder
+            responder
+            ledgerPeersConsensusInterface
+            miniProtocolParams
+            (consensusRethrowPolicy (Proxy @blk))
+            mempty
```
nfrisby, line 414: Would you add a comment here, explain what `mempty` is?

Or better yet, create a `where` binding

```
localRethrowPolicy :: RethrowPolicy
localRethrowPolicy = mempty
```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -321,15 +338,22 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
...
+      -> NTN.Apps m
+          (ConnectionId addrNTN)
+          ByteString
+          ByteString
+          ByteString
+          ByteString
+          ByteString
+          ()
     mkNodeToNodeApps nodeKernelArgs nodeKernel peerMetrics version =
-        NTN.mkApps
```
nfrisby, line 326: Please undo these white-space changes.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -338,58 +362,96 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
       -> NodeToClientVersion
       -> NTC.Apps m (ConnectionId addrNTC) ByteString ByteString ByteString ()
     mkNodeToClientApps nodeKernelArgs nodeKernel blockVersion networkVersion =
-        NTC.mkApps
```
nfrisby, line 341: Please undo these white-space changes.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -338,58 +362,96 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
...
 
     mkDiffusionApplications
-      :: MiniProtocolParameters
-      -> (   BlockNodeToNodeVersion blk
-          -> NTN.Apps m (ConnectionId addrNTN) ByteString ByteString ByteString ByteString ByteString ()
-         )
-      -> (   BlockNodeToClientVersion blk
+      :: NetworkP2PMode
+      -> MiniProtocolParameters
+      -> (BlockNodeToNodeVersion blk
```
nfrisby, line 374: Our style (link for reference https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md ) is

```
     Foo
  -> Bar
  -> Baz
```

so please re-introduce the leading space on the first argument.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -338,58 +362,96 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
...
+          -> NTN.Apps
+               m
+               (ConnectionId ntnAddr)
+               ByteString
+               ByteString
+               ByteString
+               ByteString
+               ByteString
+               ())
+      -> (BlockNodeToClientVersion blk
```
nfrisby, line 384: Our style (link for reference https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md ) is

```
     Foo
  -> Bar
  -> Baz
```

so please re-introduce the leading space on the first argument.

nfrisby (changes requested):
Just a few minor alternations requested to the Consensus files. Thanks!

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -238,6 +238,8 @@ library
                        -- Strict wrapper around SOP
                        Data.SOP.Strict
 
+  other-modules:       Ouroboros.Consensus.Node.ErrorPolicy
```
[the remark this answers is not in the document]
bolt12: Will do! Thanks!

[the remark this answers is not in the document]
bolt12: Got it! Thanks :)

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
bolt12: I think I prefer having the `NetworkP2PMode` type, if that is okay

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -114,6 +113,11 @@ import           Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy
                      (SnapshotInterval (..), defaultDiskPolicy)
 import           Ouroboros.Consensus.Storage.VolatileDB
                      (BlockValidationPolicy (..))
+import           Ouroboros.Consensus.Util.Args
```
nfrisby, line 116: Please put these back where they were. I know the reasoning is not obvious; please bear with me.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -203,13 +211,20 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk = L
...
+data NetworkP2PMode = EnabledP2PMode
+                    | DisabledP2PMode
+  deriving (Eq, Show)
+
 -- | Combination of 'runWith' and 'stdLowLevelRunArgsIO'
 run :: forall blk.
      RunNode blk
   => RunNodeArgs IO RemoteAddress LocalAddress blk
   -> StdRunNodeArgs IO blk
-  -> IO Void
```
nfrisby, line 211: Please undo this line-break.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
bolt12: Builkite was failing because of this, I believe, that's why I did this

bolt12: Should I ignore buildkite?

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
bolt12: Got it!

bolt12: hmm this seems to not be related to the newest state of the file

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
nfrisby: Hmm. Please do respect BuildKite, yeah.

I'm guessing you removed the blank line (old line 105). I could see how that would invite a BuildKite error.

My request includes re-introducing that blank line too. To elaborate a bit the `Ouroboros.Consensus.Storage.*` imports have their own group, and it is separated by blank lines from the above group `Ouroboros.Consensus.*` imports.

bolt12: Ah I see! Thanks for clarifying! Will do!

nfrisby (commented):
Thanks Armando! There are a couple conversations still open from my first pass (about indenting the first argument type) and I made a couple more. But you already moved the new argument into the record type, and that was the only really important one in my opinion.

So I'll change my Review to "LGTM"! Thanks for coping with our style restrictions etc. Generally: if you make the least possible formatting changes, it'll be easier to get them right :)

The only reason I'm not Approving, is to ensure that others have the opportunity to review. 👍  Thanks again, and thank you for your patience re my review.

bolt12:
Thank you @nfrisby ! Yes, thank you for pointing out the styling issues, I am more than happy to please the code owners, since I am manipulating their code base, so thank you as well for insisting on them!

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
nfrisby: Please change `(BlockNodeToNodeVersion blk` to `(   BlockNodeToNodeVersion blk`

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
nfrisby: Please change `(BlockNodeToNodeVersion blk` to `(   BlockNodeToNodeVersion blk`

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
```diff
@@ -0,0 +1,1260 @@
...
+                 sn = Snocket.localSnocket iocp addr
+             traceWith tracer $ CreateSystemdSocketForSnocketPath addr
+             sd <- Snocket.open sn Snocket.LocalFamily
+             traceWith tracer $ CreatedLocalSocket addr
+             return (Right (sn, sd, addr))
+    )
+    -- We close the socket here, even if it was provided to us.
+    (\case
+      Left  (sn, sd)    -> Snocket.close sn sd
+      Right (sn, sd, _) -> Snocket.close sn sd)
```
coot, line 1245:
```suggestion
      Right (sn, sd, _) -> Snocket.close sn sd
    )
````

File: ouroboros-consensus/src/Ouroboros/Consensus/Node.hs [hunk 21]
```diff
@@ -203,12 +212,18 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk = L
     , llrnMaxClockSkew :: ClockSkew
     }
 
+-- | P2P Switch
+--
+data NetworkP2PMode = EnabledP2PMode
+                    | DisabledP2PMode
+  deriving (Eq, Show)
```
coot, line 219:
```suggestion
data NetworkMode = P2PMode
                 | FederatedMode
  derivng (Eq, Ord)
```
Maybe let's use on slack for other ideas.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -316,12 +332,19 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
     codecConfig :: CodecConfig blk
     codecConfig = configCodec cfg
 
-    mkNodeToNodeApps
-      :: NodeKernelArgs m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
+    mkNodeToNodeApps ::
+         NodeKernelArgs m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
```
coot, line 336:
```suggestion
    mkNodeToNodeApps
      :: NodeKernelArgs m (ConnectionId addrNTN) (ConnectionId addrNTC) blk

```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -331,8 +354,8 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
           (reportMetric peerMetrics)
           (NTN.mkHandlers nodeKernelArgs nodeKernel)
 
-    mkNodeToClientApps
-      :: NodeKernelArgs m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
+    mkNodeToClientApps ::
+         NodeKernelArgs m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
```
coot, line 358: The same here, let's not change the consensus style.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -344,52 +367,92 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
           (NTC.defaultCodecs codecConfig blockVersion networkVersion)
           (NTC.mkHandlers nodeKernelArgs nodeKernel)
 
-    mkDiffusionApplications
-      :: MiniProtocolParameters
+    mkDiffusionApplications ::
+         NetworkP2PMode
```
coot, line 371: and here.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
coot: Maybe `Federated` and `P2P` are also nicer suffixes than `NonP2P` and `P2P` (but it would be a pervasive change).

coot (changes requested):

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
bolt12: Yes, thank you!

nfrisby (commented):
I marked two last style-y comments; quite minor.

Thanks much for restyling. I leave it to a Network Team member to proper Approve.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -345,51 +368,91 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
...
-         )
+          -> NTN.Apps
+               m
+               (ConnectionId ntnAddr)
+               ByteString
+               ByteString
+               ByteString
+               ByteString
+               ByteString
+               ())
```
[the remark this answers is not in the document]
bolt12: Got it! :)

coot (approved):
LGTM

PR #3190, opened 2021-05-27
CAD-2907 blockfetch client:  add SendFetchRequest msg to TraceFetchClientState

deepfire:
A message to mark the point when a fragment request is actually sent to the peer.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
+++ b/ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
@@ -190,6 +190,7 @@ blockFetchClient _version controlMessageSTM
                 Right lower = AF.last fragment
                 Right upper = AF.head fragment
 
+        traceWith tracer (SendFetchRequest fragment)
         return $
           SenderPipeline
             (ClientAgency TokIdle)
--- a/ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientState.hs
+++ b/ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientState.hs
@@ -370,6 +370,11 @@ data TraceFetchClientState header =
      | AcknowledgedFetchRequest
          (FetchRequest header)
 
+       -- | Mark the point when fetch request for a fragment is actually sent
+       -- over the wire.
+     | SendFetchRequest
+         (AnchoredFragment header)
+
        -- | Mark the start of receiving a streaming batch of blocks. This will
        -- be followed by one or more 'CompletedBlockFetch' and a final
        -- 'CompletedFetchBatch'.


ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
```diff
@@ -190,6 +190,7 @@ blockFetchClient _version controlMessageSTM
                 Right lower = AF.last fragment
                 Right upper = AF.head fragment
 
+        traceWith tracer (SendFetchRequest (FetchRequest [fragment]))
```
nfrisby, line 193: Maybe use `AnchoredFragment` as the tracer argument?

nfrisby (commented):
LGTM. I'm not Approving only because this isn't my jurisdiction. I made a comment, very minor.

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientState.hs
```diff
@@ -370,6 +370,11 @@ data TraceFetchClientState header =
      | AcknowledgedFetchRequest
          (FetchRequest header)
 
+       -- | Mark the point when fetch request for a fragment is actually sent
+       -- over the wire.
+     | SendFetchRequest
```
nfrisby, line 375: Actually; I don't know what level of abstraction you're focused on, but I suppose you could get by via one of these tracers, matching on the `MsgRequestRange` constructor, instead of adding a constructor to `TraceFetchClientState`.

https://github.com/input-output-hk/ouroboros-network/blob/eb02c2d627b728a48449d314e56c347f1798503b/ouroboros-consensus/src/Ouroboros/Consensus/Network/NodeToNode.hs#L330-L331

I'm only pointing out what superficially seems to me like an option that doesn't involve changing the code.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
deepfire: Great point, by the way!

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientState.hs
deepfire: I would rather avoid enabling low-level protocol tracers, because they _seem_ to be more expensive to have enabled.

Maybe my intuition is off, though?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Client.hs
deepfire: Done!

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientState.hs
nfrisby: These tracers are applied per message.

My anticipation is that a per-message tracer that does nothing for most messages `Tracer $ \case Foo -> ...; _ -> pure ()` wouldn't be distinguishable from a per-message tracer that does nothing for all messages (`nullTracer`). Specifically, I'd be surprised if enough inlining is happening for GHC to really generate significantly tighter code when the supplied tracer is `nullTracer`.

But, I haven't measured anything, so that's all just guess work.

nfrisby: If a per-message tracer is doing something non-trivial for many messages, than that would seem likely to incur noticeable overhead.

deepfire: The problem with high-frequency tracers is that once you enable them, then all messages they emit are processed by the switchboard (which is the definition of "tracer enabledness" in the node), which is _slow_ -- and that's the primary reason why I tend to avoid them.

nfrisby: Ah OK. I'm not really familiar with what you mean by switchboard, but I think I've seen something in the node code that makes the difference you're referring to.

👍

coot (approved):
LGTM

PR #3191, opened 2021-05-27
Update cardano-ledger-specs dependency for ledger bugfix

nfrisby:
Single commit, see message.

Specifically, this brings in Merged PR input-output-hk/cardano-ledger-specs#2305


nfrisby:
I double-checked that our `source-repository-stanza` commits tags are still in agreement. I didn't check `sources.json`; relying on CI to catch any potential problems there.

nc6 (approved):

PR #3192, opened 2021-05-28
Relax Win32 lowerbound

newhoggy:


newhoggy:
So `Win32-2.6.2.0` is definitely needed.  I'm trying to figure out what is causing to `Win32-2.6.2.0` to be rejected in the constraint solver.

newhoggy:
Closing.

PR #3202, opened 2021-06-07
Sync progress partial config

newhoggy:


ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
```diff
@@ -4,8 +4,8 @@
 module Ouroboros.Network.NodeToClient.Version
   ( NodeToClientVersion (..)
   , NodeToClientVersionData (..)
-  , nodeToClientVersionCodec
   , nodeToClientCodecCBORTerm
+  , nodeToClientVersionCodec
```
coot, line 8: I don't think this change is necessary.

coot (changes requested):

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -87,15 +87,16 @@ queryEncodeNodeToClient ::
...
-    TopLevelQueryDisabled -> case query of
-      BlockQuery blockQuery -> encodeNodeToClient
-        @blk
-        @(SomeSecond BlockQuery blk)
-        codecConfig
-        blockVersion
-        (SomeSecond blockQuery)
-      GetPartialLedgerConfig ->
-        encodeTag 1
+      TopLevelQueryDisabled ->
```
nfrisby, line 90: When `ToplevelQueryDisabled`, there are no top-level queries. That's why the previous code was correct and the new code here is contradictory.

Before we have any activated top-level queries, we'll have to introduce a `TopLevelQueryV_1` or some such in that enumeration type.

nfrisby (changes requested):
I looked at the last three commits. The first two were non-functional changes and the third seems confused.

@newhoggy Does the `cardano-node` progress bar need something more than `GetPartialLedgerConfig`? I suspect that's what @DavidEichmann is working towards right now. I didn't see anything within the commits in this PR that you solo authored that seemed to go beyond what I think David is currently doing.

If there is, would you reorganize the commits a bit so that it's more obvious what content you're proposing is this PR?

Also, if the PR is not ready for merge, please put it in the Draft PR state. Thanks!

karknu:
Could this PR get a description?
 From a network perspective it doesn't seem to have much to do with "sync progress partial config".

PR #3204, opened 2021-06-10
coot/cardano ping

coot:
- Add support for V7 keepalive
- cardano-ping: adhare to keep alive protocol

### DIFF
--- a/network-mux/demo/cardano-ping.hs
+++ b/network-mux/demo/cardano-ping.hs
@@ -211,7 +211,7 @@ data NodeVersion =       NodeToClientVersionV5 Word32
                        deriving (Eq, Ord, Show)
 
 keepAliveReqEnc :: NodeVersion -> Word16 -> CBOR.Encoding
-keepAliveReqEnc (NodeToNodeVersionV7 _ _) cookie =
+keepAliveReqEnc v cookie | v >= NodeToNodeVersionV7 minBound minBound =
        CBOR.encodeListLen 2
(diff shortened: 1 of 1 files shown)


network-mux/demo/cardano-ping.hs
```diff
@@ -222,6 +222,12 @@ keepAliveReqEnc _ cookie =
 keepAliveReq :: NodeVersion -> Word16 -> ByteString
 keepAliveReq v c = CBOR.toLazyByteString $ keepAliveReqEnc v c
 
+keepAliveDone :: ByteString
+keepAliveDone =
+    CBOR.toLazyByteString $
+         CBOR.encodeListLen 1
+      <> CBOR.encodeWord 2
```
karknu, line 229: You should cover the pre v7 version of the protocol too.

karknu (changes requested):

network-mux/demo/cardano-ping.hs
coot: Good point

coot: Note the shortcoming of the approach of integration of version and version data: we need to pattern match on the version number: when we will add `v8` for `node-to-node` the `cardano-ping` will break if one forgets to update the encoder.

coot: I pushed one more commit to fix this.

karknu (approved):
LGTM

coot:
bors merge

PR #3207, opened 2021-06-11
Use newtype for StrictTVar if checktvarinvariant cabal flag is not set

coot:
- renamed cabal.project.ci.windows file
- io-sim-classes: StrictTVar representation
- Updated cabal.project.local.ci


io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
```diff
@@ -76,11 +76,17 @@ type LazyTMVar m = Lazy.TMVar m
   Strict TVar
 -------------------------------------------------------------------------------}
 
+#if CHECK_TVAR_INVARIANT
 data StrictTVar m a = StrictTVar
    { invariant :: !(a -> Maybe String)
      -- ^ Invariant checked whenever updating the 'StrictTVar'.
    , tvar      :: !(LazyTVar m a)
    }
+#else
```
bolt12, line 85: Why the emphasizes on using a newtype instead of data in this particular case?

coot: `newtype`s do not change runtime representation; By moving from `data` with two fields to a `newtype` we save 3 machine words and one indirection (pointer) to the actual data that is interesting: the `tvar`.

bolt12: ah TIL! But is this saving worth the complexity? Should this be done whenever possible?

coot: Actually the complexity is quite low because the constructor is not exported making the impact isolated to a single module. Building both versions of `io-sim-classes` on `ci` shouldn't be too difficult too.

bolt12: Yeah, that makes sense! Nice, thanks :D

bolt12 (approved):
LGTM!

karknu (approved):
LGTM.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
```diff
@@ -113,9 +127,14 @@ newTVarWithInvariantIO :: (MonadSTM m, HasCallStack)
                        => (a -> Maybe String) -- ^ Invariant (expect 'Nothing')
                        -> a
                        -> m (StrictTVar m a)
-newTVarWithInvariantIO invariant !a =
+#if CHECK_TVAR_INVARIANT
```
nfrisby, line 130: This seems unnecessary, since the definition of `checkInvariant` already uses CPP to make it a no op without the flag.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
```diff
@@ -131,9 +150,11 @@ readTVarIO :: MonadSTM m => StrictTVar m a -> m a
 readTVarIO StrictTVar { tvar } = Lazy.readTVarIO tvar
 
 writeTVar :: (MonadSTM m, HasCallStack) => StrictTVar m a -> a -> STM m ()
-writeTVar StrictTVar { tvar, invariant } !a =
-    checkInvariant (invariant a) $
-    Lazy.writeTVar tvar a
+writeTVar v !a =
+#if CHECK_TVAR_INVARIANT
```
nfrisby, line 154: This seems unnecessary, since the definition of `checkInvariant` already uses CPP to make it a no op without the flag.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
nfrisby: Without knowing any details, I would guess this won't make an observable difference in any performance metrics. So Armando's question about complexity is the first that came to my mind, too. But I was worried about "the less CPP the better" more so than encapsulation concerns etc.

I'm not objecting, just sharing my opinion. The extra CPP here is not overwhelming.

We could get by entirely without CPP by using a conditional in the cabal file to choose between two very short source files that both define `StrictTVar` and its fundamental interface. That appeals to me, but also seems a bit extreme in the other direction.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
```diff
@@ -100,7 +110,11 @@ toLazyTVar :: StrictTVar m a -> LazyTVar m a
 toLazyTVar StrictTVar { tvar } = tvar
 
 newTVar :: MonadSTM m => a -> STM m (StrictTVar m a)
+#if CHECK_TVAR_INVARIANT
```
nfrisby, line 113: How about `newTVar a = newTVarWithInvariantIO  (const Nothing)`?

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
nfrisby: Ah, nevermind. I misread the difference. It's about `invariant` being in scope.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
nfrisby: Ah, I misread the diff here too. It's about `invariant` being in scope or not.

This is fine as is. But I'm thinking we could consolidate the CPP into a single definition of 

```
getInvariant :: StrictTVar -> a -> Maybe String
```

And then use `getInvariant` to define both `castStrictTVar` and `writeTVar` without CPP.

nfrisby (commented):
I'm not objecting to the PR -- others have already approved it.

But I did offer some suggestions to minimize the amount of CPP. Take them or leave them. (Although the fact that I got confused even during my review about how confusing it is possibly evidence that the less CPP the better :).)

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
coot: That would not compile the lhs is in `STM m` while the right hand side is in monad `m`.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
coot: We need `getInvariant` too and `setInvariant` to consolidate the CPP pragrams. Check 11dc98633.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
nfrisby: Ah, right. And I suppose there's no `newTVarWithInvariant` lying around.

I wonder if a little more refactoring could achieve something similar and thereby avoid another `#if`. But I'll leave it there; I'm pretty sure you get the point of the question and I know you can judge it 👍

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
nfrisby: It might be nicer to use `mkStrictTVar :: (a -> Maybe String) -> LazyTVar m a -> StrictTVar m a`. Unlike `setInvariant`, that wouldn't require partial record expressions.

At this point, with `getInvariant` and `mkStrictTVar`, we've created an abstract interface that handles both values of the CPP -D variable. The interface looks like invariants are always present, but we can ignore them and/or use `const Nothing` as a filler when their disabled/irrelevant.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
coot: The performance impact will not be visible, we'd need to have 100s if not 1000s of `StrictTVar` for it to be visible in some way; but I think if we want to put it on hackage we should make the change.

> We could get by entirely without CPP by using a conditional in the cabal file to choose between two very short source files that both define StrictTVar and its fundamental interface. That appeals to me, but also seems a bit extreme in the other direction.

Yes that's another option, though it would not make a difference: one needs to compile the package twice to type check all the code.  We could go around this with two modules and a CPP declaration used to import one of the representations, but it feels like we'll trying to be to clever for something that's quite simple.

io-sim-classes/src/Control/Monad/Class/MonadSTM/Strict.hs
coot: > It might be nicer to use mkStrictTVar :: (a -> Maybe String) -> LazyTVar m a -> StrictTVar m a. Unlike setInvariant, that wouldn't require partial record expressions.

Yes I realised that `setInvariant` is partial, while `mkStirctTVar` is not.

coot:
bors merge

PR #3209, opened 2021-06-11
Use a counter to provide a unique remote addr

karknu:
The inbound governor requires all clients to have a unique address. UNIX
sockets and windows named pipes lack this so we use a counter to
generate remote addresses for local clients.

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE CPP                 #-}
 {-# LANGUAGE DeriveGeneric       #-}
 {-# LANGUAGE DerivingVia         #-}
@@ -38,6 +39,7 @@ import           Data.Bifunctor (Bifunctor (..))
 import           Data.Bifoldable (Bifoldable (..))
 import           Data.Hashable
 import           Data.Typeable (Typeable)
+import           Data.Word
 import           GHC.Generics (Generic)
 import           Quiet (Quiet (..))
 #if !defined(mingw32_HOST_OS)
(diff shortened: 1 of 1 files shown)


karknu:
Fixes https://github.com/input-output-hk/cardano-node/issues/2726

ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
```diff
@@ -147,19 +150,32 @@ berkeleyAccept ioManager sock = go
...
+
+            -- UNIX sockets don't provide a unique endpoint for the remote
+            -- side, but the InboundGovernor/Server requires one in order to
+            -- track connections.
+            -- So to differentiate clients we use a simple counter as the
+            -- remote end's address.
+            --
+            addr'' <- case addr' of
+                           Socket.SockAddrUnix _ ->
+                               return $ Socket.SockAddrUnix (show cnt)
```
coot, line 162: That's a hack, but I don't see a better solution.  We certainly don't want to end up with a sum type as a `LocalAddress`.

What about adding some a prefix (`ephemeral-` or `temp-`) to make it self explanatory in log files.

karknu: Yes, it is a hack.
I'll add `temp-` to the remote end.

ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
```diff
@@ -147,19 +150,33 @@ berkeleyAccept ioManager sock = go
...
+            -- side, but the InboundGovernor/Server requires one in order to
+            -- track connections.
+            -- So to differentiate clients we use a simple counter as the
+            -- remote end's address.
+            --
+            addr'' <- case addr' of
+                           Socket.SockAddrUnix _ ->
+                               return $ Socket.SockAddrUnix $
+                                   "temp-" ++ show cnt
+                           _                     -> return addr'
```
coot, line 164: It looks strange when it's hanging that much when the previous line is so much to the left; up to you if you want to include it or not.
```suggestion
                           _  -> return addr'
```

coot (approved):
LGTM

ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
karknu: I prefer to have the `->` aligned.

PR #3210, opened 2021-06-11
Integrate alonzo examples

Jasagredo:
Integrates the Alonzo examples provided in https://github.com/input-output-hk/cardano-ledger-specs/pull/2323.


Jasagredo:
The CI is failing every time regarding Win32 version mismatch. Cabal seems to choose the Win32 version that comes with GHC 8.10.4, which is 2.6.1.0 and ouroboros-consensus explicitly requires 2.6.2.0. The lower bound has been there since Oct 2019 and only now it is complaining.

Removing the lower bound to see what Cabal comes up with, and comparing it to a run on current master shows that the build plan changes this way:

| packages built only in master | packages built only in this PR |
| ---- | ---- |
| Cabal-3.4.0.0 (lib) | cassava-0.5.2.0 (lib) |
| Win32-2.8.5.0 (lib) | plutus-tx-plugin-0.1.0.0 (lib) |
| directory-1.3.6.1 | text-short-0.1.3 (lib) |
| process-1.6.11.0 | |
| time-1.9.3 | |

`text-short` is a dependency of `cassava` which was introduced in the `plutus-core` library ([see](https://github.com/input-output-hk/plutus/blob/c89e13e66c5bef628a1f7ea3a3732951be8be8ba/plutus-core/plutus-core.cabal#L250)). `plutus-tx-plugin` has also been introduced by plutus and by updating the ledger to the [latest master](https://github.com/input-output-hk/cardano-ledger-specs/blob/master/cabal.project#L69).

ouroboros-consensus-shelley-test/src/Test/ThreadNet/TxGen/Shelley.hs
```diff
@@ -170,3 +172,8 @@ mkGenEnv whetherPPUs coreNodes = Gen.GenEnv keySpace constants
             ksStakePools
           } =
             Gen.Presets.keySpace @(MockShelley h) constants
+
+    scriptSpace :: Gen.ScriptSpace (MockShelley h)
```
nfrisby, line 176: Is this PR still in Draft because we're waiting for `cardano-ledger-specs` to export this so that we don't have to write it here?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
```diff
@@ -69,8 +69,7 @@ import           Ouroboros.Consensus.Util.Versioned
 import qualified Cardano.Ledger.BaseTypes as SL (ActiveSlotCoeff, Seed)
 import           Cardano.Ledger.Crypto (StandardCrypto, VRF)
 import qualified Shelley.Spec.Ledger.API as SL
-import qualified Shelley.Spec.Ledger.BlockChain as SL (checkLeaderValue, mkSeed,
-                     seedEta, seedL)
+import qualified Shelley.Spec.Ledger.BlockChain as SL (mkSeed, seedEta, seedL)
```
nfrisby, line 72: So `checkLeaderValue` was unused? I wonder why that hadn't triggered a wanring.

nfrisby:
Have you looked through which binary files changed? Would you summarize what changed and why that wasn't unexpected, preferably in a commit message. Thanks!

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs
Jasagredo: It is used [here](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol.hs#L507) but:
- `Shelley.Spec.Ledger.BlockChain` exports it.
- `Shelley.Spec.Ledger.API.Protocol` re-exports it.
- `Shelley.Spec.Ledger.API` re-exports all from `Shelley.Spec.Ledger.API.Protocol`.
- `Shelley.Spec.Ledger.API` is already imported qualified as `SL`.

So GHC claims it is a redundant import.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/TxGen/Shelley.hs
Jasagredo: No, we were waiting for the function two lines below that one, `Gen.Presets.scriptSpace`. This is just a local binding.

Jasagredo:
> Have you looked through which binary files changed? Would you summarize what changed and why that wasn't unexpected, preferably in a commit message. Thanks!

Some of the *Alonzo* golden artifacts changed between:
- the first version of the cardano-ledger-specs Alonzo examples branch
- later versions of the Alonzo examples branch and commits in master after that branch was merged.

But they were *only* Alonzo artifacts that changed. 

Unfortunately I cannot reproduce it anymore because the branch was force-pushed and those previous commits are no longer in Github.

The old Alonzo examples branch was branching from [here](https://github.com/input-output-hk/cardano-ledger-specs/commit/66df483d167072fc488efb9e0e4ca912863c16ec) and now it branched from [here](https://github.com/input-output-hk/cardano-ledger-specs/commit/d6b3d6b8ed16b2c69684fa44218f4e7060f166d3). The only change in between that could have affected this is [this commit](https://github.com/input-output-hk/cardano-ledger-specs/commit/30a8caff606524827e896bec8247cdcd81787417) that changes serialization from `Flat` to `CBOR` and it even mentions: "Nothing changes in the ledger proper since it already treated scripts as opaque bytestrings, but some tests needed fixing." So probably we don't need to worry about this deviation I saw.

nfrisby (approved):
LGTM The first commit makes the change, the second commit updates the examples. The git diff for the second shows that all the changes it makes are adding new files. 👍

PR #3163, opened 2021-05-18
IOSim versions of Server2 tests

UlfNorell:

PR #3216, opened 2021-06-15
server2 iosim test

coot:
- get ShowProxy instance from Orphans module
- Add NonFailingBearerInfoScript
- server2 test: fail property if protocol does not complete
- server2 test: shrink ClientAndServerData
- server2 test: IOSim versions of existing IO properties


coot:
Updated version of #3163

Issue #3211, opened 2021-06-11
Mempool: Bound ex units of chosen transactions when forging new block


EncodePanda:
## Origin of the issue

Recently @dcoutts pointed out that

>Something I was worried about last night: have we done anything yet for block forging to select txs up to but not exceeding our limit for script execution units? My guess was we have not, and we'll need to do this to complete things for Alonzo.
I think we can start with a naive algorithm, rather than any box packing algorithms. (...). i.e. select them in order, up to any of the 3 limits being hit: block size, script mem units, script step units

### Exectution units
Each transaction holds the exact number of ex units (execution units) that it will require for its evaluation. Thus the overall block execution cost is the sum of all ex units from each of the transactions within that block.

Ledger has implemented a rule validating that the block execution does not exceed the maximum value (defined by the protocol).

### The Mempool

Consensus maintains an abstraction called Mempool. It aggregates transactions that will become part of the next block that will be forged by a given node.

## The problem

Upon forging a block, a selected prefix of transactions will be taken to become the content of the next forged block. It will be a prefix because the node checks if the transactions don't build up to an overweight (in terms of bytes) block (after the block gets serialized). However, this procedure does *not* take into account the overall block execution cost (in ex units). This means that our node will forge a block that will be invalid (too "expensive").

## Solution
As pointed out by @dcoutts we can have a simple extension to the "prefixing" algorithm that will take also into account the execution cost of created block.

## Definition of done

We are done when:

- [ ] We have a PASSING test that demonstrates the problematic behavior (this test should fail before any changes are done to the production code).
- [ ] We have extended that "prefixing" feature so that it makes the above-mentioned test pass

PR #3224, opened 2021-06-23
Mempool: Bound ex units of chosen transactions when forging new block

EncodePanda:
Fixes https://github.com/input-output-hk/ouroboros-network/issues/3211

1. Add maxTxExUnitCapacity and txExUnitsSize to LedgerSupportsMempool

2. Move MaxTxCapacityOverride to Ourboros.Consensus.Block.Forging

Previously NodeKernel was filtering upfront the list of transactions
that will be included in the block. This is sole responsibility of the
ledger, consensus should not really have any logic that decides which
transactions become a member of a block and which don't.

This commit is a first step towards that scenario. We move away any
logic that filters the list of transactions and we make forgeBlock
function responsible to make appropriate decisions on which
transactions to include. Following commit (in the future) should consider
moving that logic completely back to ledger, so that there is a zero
chance that the consensus could create a block that by definition will
be rejected ledger

3. Modify forgeBlock so that it takes all txs from the mempool and decides which to include in a block

4. Extend the prefixing mechanism in Shelley forgeBlock to include the ExUnits max size


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -128,6 +128,10 @@ instance ShelleyBasedEra era
 
   txForgetValidated (ShelleyValidatedTx txid tx) = ShelleyTx txid (SL.extractTx tx)
```
EncodePanda, line 130: Here we do have a problem and I would love to discuss it quickly.

Instance for `LedgerSupportsMempool` was written for all blocks which are "shelley based". However information about execution units is only held in Alonzo types (e.g fields like `totExunits`). We can not add constraints to `ShelleyBasedEra` like `HasField "totExunits"`.
It seems that we have to just create a dedicated instance for `LedgerSupportsMempool` that is crafted just for Alonzo, and potentially more general one for the rest of the shelley eras. Or maybe there is some other, smarter way to do things here

ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Forge.hs
```diff
@@ -156,22 +166,34 @@ forgeRegularBlock cfg bno sno st txs isLeader =
...
-        ByronUpdateProposal _ prop -> bp { bpUpProposal = Just (void prop) }
-        ByronUpdateVote     _ vote -> bp { bpUpVotes    = void vote : bpUpVotes }
+    extendBlockPayloads maxBlockSize genTx bp@BlockPayloads{bpTxs, bpDlgCerts, bpUpVotes, bpSize}
+      -- do not include transaction if it will exceed maximum block size
+      | bpSize + txInBlockSize genTx > maxBlockSize = bp
+      | otherwise                                   =
+        -- TODO: We should try to use 'recoverProof' (and other variants of
+        -- 'recoverBytes') here as opposed to throwing away the serializations
+        -- (the 'ByteString' annotations) with 'void' as we're currently doing.
+        case genTx of
```
nfrisby, line 180: It would be more terse to introduce a `bp' = bp { bpSize = bpSize + txInBlockSize genTx }` and then update `bp'` in each branch. That makes it clear that regardless of the `genTx` constructor, we update the `bpSize` field the same way.

What do you think?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -61,7 +73,50 @@ forgeShelleyBlock hotKey canBeLeader cfg curNo curSlot tickedLedger txs isLeader
...
+    appendTx maxBlockSize tx bp@(BlockPayloads bpTxs bpSize bpExUnits)
+      -- do not include transaction if it will exceed maximum block size
+      | bpSize + txInBlockSize genTx > maxBlockSize        = bp
+      -- do not include transaction if it will exceed maximum ex unit size
+      | exUnitsExceedsLimit (bpExUnits <> txExUnitsSize genTx) = bp
+      | otherwise = bp { bpTxs = tx : bpTxs
+                       , bpSize = txInBlockSize (txForgetValidated tx) + bpSize
+                       , bpExUnits = bpExUnits <> txExUnitsSize genTx
+                       }
+      where
```
nfrisby, line 109: Similar to my comment on the Byron function, my instinct is to define `bp'` in the `where` clauses and use it in the guard conditions and as the guarded return value.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
nfrisby: I haven't worked through it, but my suggestion would be to add the new methods to the `ShelleyBasedEra` class instead of to the `LedgerSupportsMempool` class. Two benefits in light of this current diff:

  1) We don't need to define these methods for Byron, ByronDual, HardForkBlock, etc.
  
  2) The `ShelleyBasedEra (AlonzoEra c)` instance can have the `HasField "totExunits"` constraint and the other eras can just use `mempty`, with no constraint.
  
Give that a try and let me know what you think.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -93,6 +95,14 @@ class ( UpdateLedger blk
   -- annotation.
   txInBlockSize :: GenTx blk -> Word32
 
+  -- | The maximum number of ExUnits worth of transactions that can be put into
```
nfrisby, line 98: See my response in this thread: https://github.com/input-output-hk/ouroboros-network/pull/3224/files#r657368550

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -284,6 +284,7 @@ library
                      , binary            >=0.8   && <0.11
                      , bytestring        >=0.10  && <0.11
                      , cardano-binary
+                     , cardano-ledger-alonzo
```
nfrisby, line 287: I'm hoping that with my suggestion in this thread https://github.com/input-output-hk/ouroboros-network/pull/3224/files#r657368550 , we wouldn't need this dep here (since we'd no longer need `ExUnits` in the `LedgerSupportsMempool` declaration).

nfrisby (changes requested):
Looks real nice. I made a couple suggestions regarding shortening some of the new code and moving the new methods from `LedgerSupportsMempool` to `ShelleyBasedEra`; I think that'll address the comment you made.

There are some style stuff (`=` alignments, import statements orders etc) but we can rip through those on the call.

Nice!

ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Forge.hs
nfrisby: (You can also do `(\bp' -> bp' { bpSize = bpSize + txInBlockSize genTx }) $ case genTx of ...` or the other way around, if you like. That approach makes it impossible to accidentally use `bp` instead of `bp'` in one of the case alternatives.

ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Ledger/Forge.hs
```diff
@@ -139,10 +147,12 @@ forgeRegularBlock cfg bno sno st txs isLeader =
     epochSlots :: CC.Slot.EpochSlots
     epochSlots = byronEpochSlots cfg
 
+    computedMaxTxCapacity = computeMaxTxCapacity st maxTxCapacityOverride
+
     blockPayloads :: BlockPayloads
     blockPayloads =
         foldr
```
nfrisby, line 154: It occurs to me that `foldr` was fine when we were preserving the total length of the list, but `foldr` does not work now that we're dropping some of the list. Specifically, the current logic (5a1d48e) takes the longest _suffix_ of the input list that contains less than some total number of bytes.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Forge.hs
```diff
@@ -61,7 +73,50 @@ forgeShelleyBlock hotKey canBeLeader cfg curNo curSlot tickedLedger txs isLeader
...
+    computedMaxTxExUnitCapacity = maxTxExUnitCapacity tickedLedger
+
+    body =
+        SL.toTxSeq @era
+      . Seq.fromList
+      . toTxInBlock
+      $ toPayloads txs
+
+    toPayloads :: [Validated (GenTx (ShelleyBlock era))] -> BlockPayloads era
+    toPayloads = foldr (appendTx computedMaxTxCapacity) initBlockPayloads
```
nfrisby, line 86: Same issue here, with `foldr` as in the Byron ledger: this is taking suffixes.

My advice for both is to right a `go` list, not use `foldr` or `foldl` etc.

nfrisby (commented):
Had an insight while waking up this morning: we need to apply the aggregation and cut-off in the other order.

nfrisby (approved):
Approving. We reviewed over Meet.

Issue #3225 and Issue #3226 capture the follow-up work.

PR #3229, opened 2021-06-28
Section on suboptimal chain selection

edsko:
Documents both the implementation and its justification.

### DIFF
--- a/ouroboros-consensus/docs/report/chapters/future/genesis.tex
+++ b/ouroboros-consensus/docs/report/chapters/future/genesis.tex
@@ -1328,6 +1328,7 @@ date (because our tip is too far from the wallclock). However, this scenario
 needs special case anyway; we will come back to it in \cref{low-density}.
 
 \subsection{Block production}
+\label{sec:block-production-when-behind}
 
 In principle, block production can happen even when the node is not up to date.
 There is however not much point: any blocks that the node will produce while it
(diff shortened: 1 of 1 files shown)


EncodePanda (approved):
LGTM

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
+disconnect from the peer that served us the offending chain, at which point a
+different chain might now be the preferred chain. This introduces a delay in
+selecting the optimal valid chain, but we accept this risk as sufficiently small
+that we do not pre-emptively download blocks from all chains. After all, there
+is little incentive for an attacker with the right to produce blocks to produce
+\emph{invalid} blocks.}. Under normal circumstances this is the \emph{only}
+chain we will download the blocks of and adopt, but occassionally we might want
+to download blocks from chains \emph{other} than the best chain, perhaps because
+downloading that best chain is taking a very long time (it might even be the
+case that the remote node is intentionally trying to stall us, as a form of
```
dnadales, line 1532: There's something I don't understand in this narrative: it is clear that we might want to chose to download a sub-optimal chain because the best chain is taking a long time to download. But what I don't get is what will we do with this sub-optimal chain. Could it be that the sub-optimal chain ends up being the optimal chain after some time? 

Also I'm assuming that here we already know that a chain is optimal (or sub-optimal) through its headers. It this correct?

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
 
 \end{itemize}
+
+\section{Suboptimal chain selection}
+
+The Ouroboros papers assume, by way of simplification, that nodes broadcast
+entire chains to each other. Every node collects all of these chains, and
+every slot selects the best chain they saw. In addition, the Ouroboros analysis
+assumes a constant $\Delta$, such that chains are submitted to all nodes in
+the network within $\Delta$ slots. A consequence of $\Delta > 1$ is that
```
dnadales, line 1510: Why clarify that `\Delta > 1`? I guess `=0` makes no sense, but still.

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
+
+Chainsync can enforce this by simply truncating suboptimal chains after $k$
+blocks after the intersection point before presenting them to blockfetch.
+\end{itemize}
+\end{definition}
+
+There is one subtlety here: we are checking the $k$ limit when we instruct
+blockfetch to download blocks, but this does not automatically mean that when
+the chain database is presented with those blocks by blockfetch, the property
+still holds. Fortunately, we can argue that it does.
```
dnadales, line 1613: Sorry, which property? :grimacing:

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
+
+Under exceptional circumstances it is possible that if we \emph{ourselves} forge
+a block on top of a suboptimal chain, the distance from that suboptimal chain to
+the intersecion with the optimal one might now exceed $k$. However, the
+``network respects $\Delta$'' principle saves us from having to treat this as a
+special case.
+
+\subsection{When the node is behind}
+
+To a certain extend, no special provisions need to be made when the node is
```
dnadales, line 1660:
```suggestion
To a certain extent, no special provisions need to be made when the node is
```

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
+behind. The rule that prohibits adopting more than $k$ blocks from suboptimal
+chains still guarantees that we can continue to switch to the optimal chain
+once we get its blocks; the only difference is that we might hit that limit
+much more quickly.
```
dnadales, line 1664: >  the only difference is that we might hit that limit much more quickly.

I did not get this part.

ouroboros-consensus/docs/report/chapters/future/genesis.tex
dnadales: I can see our urges to extend chains :D

ouroboros-consensus/docs/report/chapters/future/genesis.tex
dnadales: >  Could it be that the sub-optimal chain ends up being the optimal chain after some time?

After reading the report, it seems the sub-optimal chain could become the optimal chain if we forge a block on top of it. But what about peers that do not forge a node. Do they have to keep on waiting till all the blocks from the optimal chain are provided?

ouroboros-consensus/docs/report/chapters/future/genesis.tex
EncodePanda: Delta could be something between 0 and 1. But since we know it must be bigger then one slot, and slot is 1 second, we can assume > 1 second

ouroboros-consensus/docs/report/chapters/future/genesis.tex
edsko: The common case is that the suboptimal chain is a short fork, which due to the way the fork is resolved elsewhere (other nodes forging blocks) ultimately becomes the optimal chain.

ouroboros-consensus/docs/report/chapters/future/genesis.tex
edsko: If `Delta = 1`, every block makes it in one slot, so it always arrives at the next slot leader in time.

ouroboros-consensus/docs/report/chapters/future/genesis.tex
edsko: When we are up to date, we expect one block every 20 seconds. When we are bulk syncing, we expect many of blocks per second.

ouroboros-consensus/docs/report/chapters/future/genesis.tex
EncodePanda: Not necessarily, the example that we've discussed couple of days ago was following:
1. imagine two nodes A & B  and two chains C1 & C2
2. A follows B
3. A is fetching optimal C1 to adopt it, but also (in parallel) fetching suboptimal chain C2
4. time passes and A is going to be selected as a leader for the next block. But before it's his turn he learns that B have build on top of C2 and now C2 is optimal
5. A is starting to switch to C2 (request fetch to block fetcher)
6. because most blocks were pre-fetched already, block fetcher only has to download the recently forged blocks and thus A has higher probability to forge on C2 

If suboptimal chain was not happening then A would learn that C2 is optimal, instruct block fetcher to get its blocks, but because that operation is time consuming, there is a higher probability that A would forge on C1 (not on C2) and essentially wasting resources, forging a block that will be wasted.

Now disclaimers: 
1. this is how I remembered the example that said to be a motivation behind the suboptimal chain selection. This is for optimalization purposes only. It would be cool if @nfrisby or @edsko  could confirm it
2. What I don't seem to understand just yet was why this example was not part of the chapter in the report and instead the example there (motivation) was about optimal chain being not delivered on time (slowed done, presumably as adversarial behaviour). I would be more then happy to revist this during our sync.

Jasagredo: This example seems like a nice justification for a specific situation in which suboptimal chain selection is beneficial, but ultimately it boils down to the same problem:
> downloading that best chain is taking a very long time
In this example, either B doesn't know at all about C1 or it is also not being able to download it, so we are in the same situation. (Or B itself forged the tip of C2 and therefore adopted it even though for A it is not the optimal one, this is an extreme case)

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
+sound. Note that a simple argument that we might anyway not yet have seen the
+optimal chain does not hold water, due to the $\Delta$ guarantee.
+
+TODO\todo{TODO}: Should we think harder about pre-emptively download blocks?
+After all, a node giving us invalid blocks is \emph{also} introducing delays,
+which is precisely the kind of thing we are worried about here.
+One difference is that with an invalid block, we will disconnect; however,
+nothing is stopping a node from doing this again and again.
+I see two possible solutions: pre-emptively download blocks after all, or
+(more appealing) associate invalid blocks with signer keys?
```
Jasagredo, line 1544: This is an interesting approach, but nothing is preventing an adversary of transferring the stake to a new fresh account and then bypass the banned signer key.

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
+the \emph{adoption} of the best chain must not be unnecessarily delayed.
+
+\item[The network respects $\Delta$]
+It is not in fact very clear how the value of $\Delta$ as it is used in the
+Ouroboros paper relates to the implementation, which differs from the paper
+rather significantly. However, what we \emph{can} say is that when we get to
+forge a block, \emph{either} we received all blocks we should have received
+``within $\Delta$'', in which case forging that block on top of the best chain
+we have available matches the paper; \emph{or} some blocks got delayed ``more
+than $\Delta$'', and we end up forging on the ``wrong'' chain. It can not,
```
Jasagredo, line 1563: I don't think using `wrong` here is the right word. If we didn't receive the optimal one within delta, then the "virtual optimal" chain is the one that we previously considered suboptimal, so it is not "the wrong chain", at least not for us.

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
+however, be the task of chain selection to detect such violations of $\Delta$
+(much less try to fix them); indeed, we have no guarantee of establishing
+consensus when $\Delta$ is respected even if we do \emph{not} do suboptional
```
Jasagredo, line 1566:
```suggestion
consensus when $\Delta$ is respected even if we do \emph{not} do suboptimal
```

ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1499,3 +1500,178 @@ \subsection{Possible optimisations}
...
+property that if we have the right to produce a block in the current slot then
+we must have seen the relevant predecessor to use. We might be able to solve
+this by not producing blocks when we are behind, but this leads to complications
+elsewhere (for example, support for low density chains, \cref{low-density}).
+TODO\todo{TODO}: Unlike what the report says, we in fact decided we \emph{do}
+want to produce blocks when behind, to avoid attackers being able to trick us
+into skipping our slot; we should update that section
+(\cref{sec:block-production-when-behind}) and then refer to it here.
+
+It is therefore probably safer to disable suboptimal chain selection when we
```
Jasagredo, line 1676: So would we disable it when being up to date and then falling behind? i.e. could a node adopt at most k blocks of a suboptimal chain, then fall behind (the wallclock goes over the end of the genesis window) and then disable suboptimal chain selection and waiting for the optimal chain to arrive?

ouroboros-consensus/docs/report/chapters/future/genesis.tex
EncodePanda: You are correct @Jasagredo Also as pointed out in the meeting yesterday, this example is merely a optimalization. But without suboptimal chain selection we can also get adversarial attacks on the whole chain. I will add those examples to the chapter (learning tex by copy and paste :) )

ouroboros-consensus/docs/report/chapters/future/genesis.tex
nfrisby, line 1566: Should this be `when $\Delta$ is +not+ respected`

Jasagredo:
I think we decided to not implemente suboptimal chain selection, (please correct me if I'm wrong), so maybe we can close this?
cc @dnadales @nfrisby

dnadales:
Closing as per @Jasagredo 's comment. Feel free to re-open.

Issue #3205, opened 2021-06-10
Don't punish local client for invalid scripts


nfrisby:
If a remote peer sends us an Alonzo transaction with a failing script, we include that transaction on the chain despite its ~invalidity~ phase 2 validation failure -- it's explicitly marked on the chain as ~invalid~ phase 2 invalid so that can be properly handled (eg it doesn't necessary consume its declared input UTxOs). We're only including it on the chain so that the party that created it must pay the cost of its (failed) execution.

On the other hand, if a _local_ peer (ie the wallet) sends us such a transaction, it would be more friendly of us to simply reject the transaction and inform them that something has gone wrong between their assumed intentions and our dynamic checks.

This Issue is there to enrich the `Mempool` interface and the `LocalTxSubmission` mini protocol server definition as necessary so that we can be lenient with a "local" peer that's providing transactions. Recall that we assume the Node-to-Client protocols (such as `LocalTxSubmission`) are local and imply some trust/good will in both directions.

tdiesler:
Related to: https://github.com/input-output-hk/cardano-node/issues/2942

PR #3230, opened 2021-06-28
Reject invalid scripts from our wallet

nfrisby:
Fixes Issue #3205.

This PR introduces a flag to `applyTx` that lets us use different logic when the transaction was submitted by our local neighbor, the wallet.


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -197,21 +197,27 @@ instance Show (GenTxId (ShelleyBlock era)) where
...
     (mempoolState', vtx) <-
        SL.applyTx
          (shelleyLedgerGlobals cfg)
          (SL.mkMempoolEnv   innerSt slot)
          (SL.mkMempoolState innerSt)
          tx
 
+    case wtf of
+      DoForgive | not (scriptsWereOK @era Proxy vtx) ->
+        throwError $ SL.ApplyTxError []   -- TODO what to put in this list?
```
nfrisby, line 218: @dcoutts @JaredCorduan @nc6 Do you have any clever ideas about what to put in this list?

Or should we introduce a new constructor specifically for this message? If so, I'm not sure if it should be defined in Ledger or in Consensus. Thoughts? Thanks!

P.S. - For context: if the local wallet sends us a transaction with an invalid script, we want to reject it instead of let it onto the chain (ie into the mempool). That's why the "forgiving" thing to do is to reject the valid transaction. But what error message should we send to the wallet in that case?

P.P.S. - There might also be an initial fix here and then a later richer fix for providing eg the exact Plutus error message.

nfrisby:
@EncodePanda @Jasagredo What do you think about the new flag? Should we instead have introduced a new method? Is the naming reasonable? Should `scriptsWereOK` be its own class? Etc; please share your opinions -- the current decisions are just me following my initial instincts.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
nc6: Yeah, there's not much to say here right now, since the ledger only has error types for when there's a mismatch between the tag and the result. I suggest in the present having a specific constructor defined in consensus for this, and in the future we might have more detailed information from Plutus about why a script fails and can report that.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -137,18 +138,30 @@ class ( SL.ShelleyBasedEra era
   -- etc.
   shelleyBasedEraName :: proxy era -> Text
 
+  scriptsWereOK :: proxy era -> Core.TxInBlock era -> Bool
```
nfrisby, line 141: @Jasagredo and I suspect this should live in `LedgerSupportsMempool`.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
nfrisby: @nc6 I grepped input-output-hk/cardano-node and input-output-hk/cardano-wallet for uses of `SubmitFail` -- that's what LocalTxSubmission sends to reject a transaction. It looks like "the reason" attached to the rejection is only ever logged as a string/JSON, HTTP non-success response, etc or else converted to a type that is then only used in those ways.

If we were to introduce a new Consensus-level error constructor, then some of the downstream processing would need to be expanded to account for it. On the other hand, I'm thinking there would be no need for downstream integration in order to account for a new Ledger-level error constructor; that's the dimension in which the downstream uses are already extensible (by merely appealing to some kind of pretty-printer).

EG https://github.com/input-output-hk/cardano-wallet/blob/83f9af2da68a6e678e57f3b1633aa39acf0f26d3/lib/shelley/src/Cardano/Wallet/Shelley/Network.hs#L462-L487

EG https://github.com/input-output-hk/cardano-wallet/blob/83f9af2da68a6e678e57f3b1633aa39acf0f26d3/lib/core/src/Cardano/Wallet/Api/Server.hs#L3186-L3209

So, if we add a Ledger `PredicateFailure` constructor (even just a nullary `NeighborlyScriptSomehowInvalid` stub), then I think there would be no downstream integration work. That's seems appealing since downstream teams (eg Node and Wallet) are already quite busy with more crucial Alonzo concerns.

-----

I see two options for how to proceed:

  - Option C = Add a Consensus-level error constructor (ie in the `ApplyTxErr` instance for Alonzo)

  - Option L = Add a Ledger-level error constructor (ie somewhere within Alonzo's `PredicateFailure "LEDGER"`)

Option C would be appealing if we need this kind of error to be handled differently, since this option would require downstream code to change.

Option L would be appealing if we don't need this kind of error to be handled differently, since this option would not require downstream code to change (the pretty-printers already being invoked would seamlessly handle the new constructor).

In my opinion, it comes down to "Do we want Node/Wallet/etc to handle this tx rejection any differently than traditional tx rejections?" and "If so, do they have resources/time to integrate it right now?" -- I don't know who can answer those questions.

Another concern is: should the Ledger be aware of this new "neighborly" transaction validation mode? Right now, this PR makes Consensus aware but uses a small shortcut to workaround the fact that Ledger is not aware. If Ledger should be aware, then what I called Option C above is more like "Create a new rule and only add this kind of rejection there instead of adding it to `PredicateFailure "LEDGER"`".

nc6: Ultimately, I don't think option `L` makes sense - errors arising here are _not_ predicate failures of the `LEDGER` rule, but something else. Even if Plutus were to add the facility to get errors out of the script interpreter (i.e. we know how the script fails) then it would need to be reported through a side channel rather than as an error.

I guess the above is mostly saying: no, I don't think ledger should be aware of the "neighbourly" mode. From the ledger PoV, a transaction failing phase-2 validation is perfectly valid. The ledger may expose more details to downstream consumers (as in here, and e.g. the event log) to allow them to make different decisions.

If we're mostly concerned about the short term, I would suggest that the current solution - that of an empty list of errors - isn't the worst solution. Basically says "we're rejecting this transaction even though it's perfectly fine", which is true.

ouroboros-consensus-byronspec/src/Ouroboros/Consensus/ByronSpec/Ledger/Mempool.hs
```diff
@@ -41,7 +41,7 @@ newtype instance Validated (GenTx ByronSpecBlock) = ValidatedByronSpecGenTx {
 type instance ApplyTxErr ByronSpecBlock = ByronSpecGenTxErr
 
 instance LedgerSupportsMempool ByronSpecBlock where
-  applyTx cfg _slot tx (TickedByronSpecLedgerState tip st) =
+  applyTx cfg _wtf _slot tx (TickedByronSpecLedgerState tip st) =
```
EncodePanda, line 44: a bit unfortunate abbreviation :)

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/SupportsMempool.hs
```diff
@@ -32,6 +33,19 @@ data family GenTx blk :: Type
 -- error type as when updating it with a block
 type family ApplyTxErr blk :: Type
 
+data WhetherToForgive
+  = DoForgive
```
EncodePanda, line 37: Why not just `Forgive`? My eyes will have better way of differentiating between those two values.

EncodePanda (approved):
LGTM, minor suggestions, pls squash commits before merger or reword them

ouroboros-consensus-byronspec/src/Ouroboros/Consensus/ByronSpec/Ledger/Mempool.hs
nfrisby: 😏  It'll change 😁

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/SupportsMempool.hs
nfrisby: I like the strict duality between `Do` and `DoNot`. But I agree with you that it's easier on the eyes, so I'm dropping it 👍

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
nfrisby: I think I see the appropriate way forward. I'll be pushing some new commits tomorrow.

Main idea: when the ledger returns a valid tx, it can also include "warnings". The analogy is that `LocalTxSubmission`'s "neighbor mode" is akin to the `-Werror` flag.

So I'll design around that, but this PR will go with the `SL.ApplyTxErr []` shortcut for now, so that we can defer the downstream integration of these new warnings. (Of which there is only the one, at the moment: "this tx though valid contains and invalid script".)

nfrisby:
Once PR #3241 merges, we can rebase this and it'll change far fewer files.

It currently includes that other PR's commit because this PR will change _one_ of the Golden test files that that PR belatedly adds.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -132,23 +147,162 @@ class ( SL.ShelleyBasedEra era
 
       , ToCBOR (Core.Witnesses era)
 
+      , Eq (ShelleyApplyTxError era)
+      , FromCBOR (ShelleyApplyTxError era)
+      , Show (ShelleyApplyTxError era)
+      , ShowProxy (ShelleyApplyTxError era)
+      , ToCBOR (ShelleyApplyTxError era)
       ) => ShelleyBasedEra era where
```
nfrisby, line 155: @nc6 Do the additions to this PR seem sane? The key idea is that -- at the Consensus level -- Alonzo will differ from the preceding eras in that its `applyTx` failures can additionally be due to "warnings promoted to failures". Think of it like the local node has set `-Werror`, where, in this case, the warnings would (eventually) be the Plutus error messages.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -132,23 +147,162 @@ class ( SL.ShelleyBasedEra era
...
+  -- | The transaction is valid, but it has warnings that were promoted to
+  -- errors as part of an intervention
+  --
+  -- See 'WhetherToIntervene'.
+  | AlonzoApplyTxErrorPromotedWarning !AlonzoApplyTxWarning
+  deriving (Eq, Show)
+
+instance Typeable c => ShowProxy (AlonzoApplyTxError c) where
+
+instance SL.PraosCrypto c => ToCBOR (AlonzoApplyTxError c) where
```
nfrisby, line 263: This serialization is not backwards compatible. However, this is only used for the new `AlonzoEra` so it need not be backwards compatible.

I do wonder if this is too clever/too subtle. Something like: future eras will probably have this same structure, so should the Consensus API reflect it (... maybe adding `WhetherToIntervene` is enough?), or should it remain "hidden" structure inside of integration code like this?

ouroboros-consensus/src/Ouroboros/Consensus/Node/Serialisation.hs
```diff
@@ -149,10 +148,3 @@ instance SerialiseNodeToNode blk (GenTxId     blk)
       encodeNodeToNode cfg version h
   decodeNodeToNode cfg version =
       WrapGenTxId <$> decodeNodeToNode cfg version
-
-instance SerialiseNodeToClient blk (ApplyTxErr     blk)
```
nfrisby, line 153: I'm not exactly sure why I need to remove this. But it seemed like the cleanest option. I'd like to discuss it on a call with someone. The motivations are some combination of:

  - `SerialiseNodeToClient` being a multi-parameter class
  - Certain types (eg the new `AlonzoApplyTxError`) being defined "above" the scope of `ShelleyBlock`
  - `Ouroboros.Consensus.Node.Run.SerialiseNodeToClientConstraints blk` requiring `SerialiseNodeToClient blk (ApplyTxErr blk)`
  - Wanting a single `era`-polymorphic instance of `SerialiseNodeToClient (ShelleyBlock era)` and hence `SerialiseNodeToClientConstraints (ShelleyBlock era)`
  - No longer having an `era`-polymorphic instance of `ApplyTxErr (ShelleyBlock era)`; we sort of have that, but it just defers to another type family for whom we do not have a single `era`-polymorphic instance.

nfrisby: Removing this instance is why I have to adjust other `... ApplyTxErr blk ...` constraints/instances to explicitly include the `WrapApplyTxErr` throughout the rest of this PR.

nfrisby:
The current commits are still reflective of my trailblazing; hence still Draft PR.

Also, this latest commit introduces some workarounds that will soon be properly provided by the ledger. The current tip is merely a proof of concept demonstrating that we can set `type instance GenTx (ShelleyBlock era) = TxInBlock era`.

Some Alonzo serialization golden tests fail, as expected. The Mary-to-Alonzo ThreadNet does pass; there are no Alonzo transactions in that, though, so it's not really proving much.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
nfrisby: We took a different path than the above out-dated conversation once we realized that the best solution is to have Tx-submitters send not only the Alonzo tx but also the flag claiming whether its scripts are invalid.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
nfrisby: Outdated -- PR input-output-hk/cardano-ledger-specs#2379 supplants this

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
nfrisby: Outdated -- PR input-output-hk/cardano-ledger-specs#2379 supplants this

ouroboros-consensus/src/Ouroboros/Consensus/Node/Serialisation.hs
nfrisby: Outdated -- PR input-output-hk/cardano-ledger-specs#2379 supplants this

nfrisby:
@nc6 @EncodePanda would you review again please. Context/reasons for Draft:

  - This PR's `cabal.project` targets input-output-hk/cardano-ledger-specs#2379 which I think will merge soon.
  
  - This PR currently includes the commits from Alexey's PR #3268, which integrates some `cardano-ledger-specs` changes that Clarke's upstream PR is branched off of. It will also merge soon.
  
  - I have a couple stray diffs (eg the `inotify-tools`) that I still need to tidy up.

nfrisby:
@Jasagredo @nc6 I now think this is ready for merge; please do a final review. Thanks!

@EncodePanda you're welcome to review it if you need a short mixup from your Genesis focus etc, but don't feel obligated.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
nfrisby: Outdated comment.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -124,36 +140,134 @@ class ( SL.ShelleyBasedEra era
...
+              wti
+              tx{Alonzo.isValidating = Alonzo.IsValidating (not flag)}
+        _ -> throwError e
+               -- reject the transaction, protecting the local wallet
+
+-- not exported
+--
+-- The ledger failure we see when the transaction's claimed 'IsValidating'
+-- flag was incorrect
+pattern IncorrectClaimedFlag ::
```
nc6, line 246: Nice use of a pattern here!

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -124,36 +140,134 @@ class ( SL.ShelleyBasedEra era
...
+
+      handler e = case (wti, e) of
+        (DoNotIntervene, SL.ApplyTxError errs)
+          | flag:flags <- nubOrd [b | IncorrectClaimedFlag b <- errs] ->
+            if not (null flags)
+            then throw $ UnexpectedAlonzoLedgerErrors (flag:flags)
+            else
+            -- rectify the flag and include the transaction
+            --
+            -- This either let's the ledger punish the script author for
```
nc6, line 226:
```suggestion
            -- This either lets the ledger punish the script author for
```

nc6 (approved):
One grammar fix, otherwise looks good.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
nfrisby: Fixed, thanks!

nfrisby:
I addressed Clarke's comments and opened the #3276 follow-up. Rebased onto `origin/master`, which was automatic with no conflicts. Thanks, reviewers!

PR #3234, opened 2021-07-01
Delay reporting of failed hot to warm demotions

karknu:
When hot to warm demotion fails because the peer is already cold
delay the throwing of `ColdDeactivationException` so that
`Monitor.connections` has a chance to notice the cold peer.


karknu:
deprecated

PR #3242, opened 2021-07-05
Add CardanoNodeToClientVersion7 golden test artifacts

DavidEichmann:
It seems running `cabal test all` creates these test artifacts. I'm curious why these artifacts aren't already added, is there some reason not to add them?


DavidEichmann:
Duplicate of #3242

PR #3244, opened 2021-07-05
Mempool tracer without bang patterns

jutaro:
For new tracing it is advisable/necessary  to remove the strictness annotations of the tracer type.

1. These objects are guaranteed to live only very short (with the new tracing framework).
2. Their is a great probability, that these messages will be filtered out, in which case laziness is better.
3. This is of all the dozens of tracer types the only one with such strictness annotations.
4. The new tracing infrastructure works prototype based. Since it is difficult to construct these objects, we use 'undefined' to make it work, which is not possible with these strictness annotations.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Mempool/API.hs
@@ -419,38 +419,38 @@ instance Monoid MempoolSize where
 -- | Events traced by the Mempool.
 data TraceEventMempool blk
   = TraceMempoolAddedTx
-      !(Validated (GenTx blk))
+      (Validated (GenTx blk))
       -- ^ New, valid transaction that was added to the Mempool.
-      !MempoolSize
+      MempoolSize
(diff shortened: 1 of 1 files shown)


nfrisby (approved):
Thanks for the nice PR description! LGTM.

nfrisby:
It looks like you won the lottery with that Nix failure 👏👏👏

I have opened https://github.com/input-output-hk/ouroboros-network/issues/3250 to look into it.

I anticipate that if we ran this PR's Hydra evals again, it wouldn't happen; I think you just got "lucky". So I'm going to try to force Hydra to do that now.

EncodePanda:
This PR is now green, should we merge it? @nfrisby @jutaro

jutaro:
Yes, please merge

nfrisby:
@jutaro Would you rebase it? I expect that `git fetch origin; git rebase -i origin/master` will work without conflicts. Then we can invoke `bors` to merge it. Thanks!

jutaro:
I have done the rebasing

Issue #3235, opened 2021-07-01
Consider moving `maxTxCapacityOverride` to `StdRunNodeArgs`


EncodePanda:
Right now Node can modify `MaxTxCapacityOverride` by calling `llrnCustomiseNodeKernelArgs`, but they never did :)
Moving this configuration option to `StdRunNodeArgs` will make it more visible to the Node team.

Also make it possible to override *some* of the components of the `Measure` in a specific era.

PR #3246, opened 2021-07-06
Expose srnMaxTxCapacityOverride allowing Node to modify MaxTxCapacityOverride

EncodePanda:
Fixes https://github.com/input-output-hk/ouroboros-network/issues/3235

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
@@ -190,14 +190,22 @@ data MaxTxCapacityOverride blk
 
 type family Overrides blk
 
+-- | Computes maximum capacity for a given block type.
+-- If node operator chose not to override this value, we choose the default value
+-- from ledger (ledgerLimit).
+-- If node operator chose to override this value, they will provide an
+-- override function Measure blk -> Measure blk, which we apply with the ledger
+-- default. Result of calling this override is compared (pointwiseMin) with the
+-- ledger default, and the result of that is returned to the caller of the
+-- computeMaxTxCapacity
 computeMaxTxCapacity ::
-     forall blk. (TxLimits blk, Overrides blk ~ Measure blk)
+     forall blk. (TxLimits blk, Overrides blk ~ (Measure blk -> Measure blk))
   => TickedLedgerState blk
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
```diff
@@ -190,14 +190,22 @@ data MaxTxCapacityOverride blk
...
+-- If node operator chose not to override this value, we choose the default value
+-- from ledger (ledgerLimit).
+-- If node operator chose to override this value, they will provide an
+-- override function Measure blk -> Measure blk, which we apply with the ledger
+-- default. Result of calling this override is compared (pointwiseMin) with the
+-- ledger default, and the result of that is returned to the caller of the
+-- computeMaxTxCapacity
 computeMaxTxCapacity ::
-     forall blk. (TxLimits blk, Overrides blk ~ Measure blk)
+     forall blk. (TxLimits blk, Overrides blk ~ (Measure blk -> Measure blk))
```
nfrisby, line 202: I don't know if we want to commit to this clever representation more so than you have in this PR as-is, but if we're effectively requiring that `Overrides` is a product of zero or more endomorphisms (eg one for "leaf" blocks and an `NP` for the HFC), then we can drop the `MaxTxCapacityOverride` data declaration. We don't need the `NoMaxTxCapacityOverride` case anymore, since, if they don't want to override anything, they can supply `id`, ie `mempty` for https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Monoid.html#t:Endo

nfrisby (approved):
What a nice PR, so direct :)  Functions are so neat.

I had one suggestion. Let me know what you think. I think it would simplify the types on both our end and on the `cardano-node` end, but it might be less future-proof.

ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
EncodePanda: I understand the idea. The only issue I have with it is the fact that we would still have to introduce some functions like `noOvverride = id` for readability. Hmm... I will do the following: I will merge this PR as is and will open new one implementing your comment. We will then see which approach works better

EncodePanda:
bors merge

PR #3252, opened 2021-07-08
Add --no-create flag to consensus tests under nix

### LINKED ISSUE #3251: Make CI fail when golden test files are missing
PR UnkindPartition/tasty-golden#23 added the `--no-create` flag to `tasty-golden` three years ago.

This Issue is to enable that flag on CI, so we can't forget again to add the new test golden expectation files when adding golden tests.

rvl:
Here's one way how to change test suite CLI args in the Haskell.nix build. Sorry for stepping on your lawn with a PR, but I thought I may as well since it's a one-line change more or less.

- nix: Add `--no-create` flag to consensus tests.
- github actions: consensus tests are not executed.

Resolves #3251

### Comments

It's never simple, is it. CI was supposed to fail because of c4d9cae2c00bc1d01e72c40a0ae6fd94174ef844, but it hasn't. Perhaps I got the component/test-suite name wrong?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -41,7 +41,7 @@ let
 
     checks = recurseIntoAttrs {
       # `checks.tests` collect results of executing the tests:
-      tests = collectChecks haskellPackages;
+      tests = collectChecks' haskellPackages;
     };
 
     # These are not run on hydra, but will be built separately in a nightly
--- a/nix/ouroboros-network.nix
+++ b/nix/ouroboros-network.nix
@@ -29,14 +29,20 @@ let
     inherit compiler-nix-name src;
     modules = [
 
-      # Compile all local packages with -Werror:
       {
+        # Compile all local packages with -Werror:
         packages = lib.genAttrs projectPackages
           (name: { configureFlags = [ "--ghc-option=-Werror" ]; });
       }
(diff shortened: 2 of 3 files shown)


nfrisby:
Thanks for looking into it; I appreciate the guidance. No lawn concerns when an expert swoops in :)

I also would expect removing that file to cause a failure with `--no-create`; I'll try reproing locally.

nfrisby:
Ha! You've just got the touch @rvl. The file you removed was one of two that are actually vestigial! I've pushed up another commit that removes the _live_ version of the same file you removed. It should now fail, if `--no-create` is working.

nfrisby:
It did fail 🙌 , with `Golden file does not exist; --no-create flag specified`.

@rvl Do you want to continue ushering this PR and then I'll Approve? Or we could just take it from here.

Thanks again for your help.

rvl:
Hi @nfrisby please do take the PR from here, thanks!

nfrisby (approved):
I'm Approving. The only thing I did was rebase Rodney's work onto `origin/master`. It was automatic, with no conflicts.

Thanks again, Rodney.

PR #3259, opened 2021-07-09
Handle failed demotions directly.

karknu:
Instead of depending on connection monitoring to process cold peers
handle it directly.


karknu:
Alternative fix for #3234 .

karknu:
deprecated

PR #3261, opened 2021-07-12
Move override support for max tx capacities into the ProtocolInfo constructors

### LINKED ISSUE #3235: Consider moving `maxTxCapacityOverride` to `StdRunNodeArgs`

nfrisby:
Fixes #3235 (again).

See the individual commit messages, some of which are extensive.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
@@ -27,7 +27,7 @@ module Ouroboros.Consensus.Shelley.Ledger.Mempool (
   , mkShelleyTx
   , mkShelleyValidatedTx
   , perTxOverhead
-  -- * Exported for tests
+    -- * Exported for tests
   , AlonzoMeasure (..)
   ) where
 
@@ -41,6 +41,8 @@ import           NoThunks.Class (NoThunks (..))
 
 import           Cardano.Binary (Annotator (..), FromCBOR (..),
                      FullByteString (..), ToCBOR (..))
+import           Data.DerivingVia (InstantiatedAt (..))
+import           Data.Measure (BoundedMeasure, Measure)
 
 import           Ouroboros.Network.Block (unwrapCBORinCBOR, wrapCBORinCBOR)
 
@@ -276,39 +278,24 @@ theLedgerLens f x =
 -------------------------------------------------------------------------------}
 
 instance (SL.PraosCrypto c) => TxLimits (ShelleyBlock (ShelleyEra c)) where
-  type Measure (ShelleyBlock (ShelleyEra c)) = ByteSize
-  txMeasure    = ByteSize . txInBlockSize . txForgetValidated
-  maxCapacity  = ByteSize . txsMaxBytes
-  pointwiseMin = min
+  type TxMeasure (ShelleyBlock (ShelleyEra c)) = ByteSize
+  txMeasure        = ByteSize . txInBlockSize . txForgetValidated
+  txsBlockCapacity = ByteSize . txsMaxBytes
 
 instance (SL.PraosCrypto c) => TxLimits (ShelleyBlock (AllegraEra c)) where
-  type Measure (ShelleyBlock (AllegraEra c)) = ByteSize
-  txMeasure    = ByteSize . txInBlockSize . txForgetValidated
-  maxCapacity  = ByteSize . txsMaxBytes
-  pointwiseMin = min
+  type TxMeasure (ShelleyBlock (AllegraEra c)) = ByteSize
+  txMeasure        = ByteSize . txInBlockSize . txForgetValidated
+  txsBlockCapacity = ByteSize . txsMaxBytes
 
 instance (SL.PraosCrypto c) => TxLimits (ShelleyBlock (MaryEra c)) where
-  type Measure (ShelleyBlock (MaryEra c)) = ByteSize
-  txMeasure    = ByteSize . txInBlockSize . txForgetValidated
-  maxCapacity  = ByteSize . txsMaxBytes
-  pointwiseMin = min
-
-data AlonzoMeasure = AlonzoMeasure {
-    byteSize :: !ByteSize
-  , exUnits  :: !ExUnits
-  } deriving stock (Show, Eq)
-
-instance Semigroup AlonzoMeasure where
-  (AlonzoMeasure bs1 exu1) <> (AlonzoMeasure bs2 exu2) =
-    AlonzoMeasure (bs1 <> bs2) (exu1 <> exu2)
-
-instance Monoid AlonzoMeasure where
-  mempty = AlonzoMeasure mempty mempty
+  type TxMeasure (ShelleyBlock (MaryEra c)) = ByteSize
+  txMeasure        = ByteSize . txInBlockSize . txForgetValidated
+  txsBlockCapacity = ByteSize . txsMaxBytes
 
 instance ( SL.PraosCrypto c
          ) => TxLimits (ShelleyBlock (AlonzoEra c)) where
 
-  type Measure (ShelleyBlock (AlonzoEra c)) = AlonzoMeasure
+  type TxMeasure (ShelleyBlock (AlonzoEra c)) = AlonzoMeasure
 
   txMeasure (ShelleyValidatedTx _txid vtx) =
     AlonzoMeasure {
@@ -316,12 +303,17 @@ instance ( SL.PraosCrypto c
       , exUnits  = totExUnits (SL.extractTx vtx)
       }
 
-  maxCapacity ledgerState =
-    let pparams  = getPParams $ tickedShelleyLedgerState ledgerState
-    in AlonzoMeasure {
-        byteSize = ByteSize $ txsMaxBytes ledgerState
-      , exUnits  = getField @"_maxBlockExUnits" pparams
-      }
+  txsBlockCapacity ledgerState =
+      AlonzoMeasure {
+          byteSize = ByteSize $ txsMaxBytes ledgerState
+        , exUnits  = getField @"_maxBlockExUnits" pparams
+        }
+    where
+      pparams = getPParams $ tickedShelleyLedgerState ledgerState
 
-  pointwiseMin (AlonzoMeasure bs1 (ExUnits mem1 steps1)) (AlonzoMeasure bs2 (ExUnits mem2 steps2)) =
-    AlonzoMeasure (bs1 `min` bs2) (ExUnits (mem1 `min` mem2) (steps1 `min` steps2))
+data AlonzoMeasure = AlonzoMeasure {
+    byteSize :: !ByteSize
+  , exUnits  :: !ExUnits
+  } deriving stock (Eq, Generic, Show)
+    deriving (BoundedMeasure, Measure)
+         via (InstantiatedAt Generic AlonzoMeasure)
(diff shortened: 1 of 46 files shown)


nfrisby:
@Jasagredo This is ready for review now. The first 5 commits are as before, the remaining commits are a bit of a departure. It's something Pawel and I had discussed before, and once I got my hands in here, it seemed worthwhile. The commit message on `consensus: implement TxLimits.pointwiseMin via a meet-semilattice` explains my reasoning.

I would understand if this feels like too much. The reason I've left it in one PR is because the downstream interface is different as of the last commit. I don't like merging work that is known to be incomplete, so I'd like to review these commits together.

On that note, I've left it as a Draft PR after all, since I'll want to squish these commits down once we're happy with them; the first commit goes one way (introduces `Overrides = Measure -> Measure`1), but we end up somewhere else by the end of the PR (`Overrides = Measure`).

Also, if the `Generics` stuff is unfamiliar, let me know and I can ping Clarke/Edsko/etc for review there 👍

nfrisby:
We discussed this in the Planning Meeting; Consensus Team + @nc6

  - The lattice based approach is fine, has enough minor benefits that they add up.
  - Same for the generic definitions -- they're very simple and allow for deriving everything for `AlonzoMeasure` (and future such types).
  - We should define the lattice stuff in `cardano-base`, however. (Spoiler: I'm going to send `FromGeneric` with it in my PR there.)

We also spent a couple hours (by now) discussing what the right interface is for these lattice concepts we're invoking. This question remains unanswered. But here are some remarks.

  - We know we have a ("full") lattice: both meet and join and both bottom and top. EG in the concrete cases (like `AlonzoMeasure`), we want https://en.wikipedia.org/wiki/Product_order , which gives a full lattice.
  - However, we _do not_ want a join operation. Whereas meet lets the user's override tighten the bound (ie restrict it), join would loosen it (ie relax it). Which would create a block that other nodes _not_ overriding the ledger limit would then consider invalid. On the other hand, meet can't create invalid blocks, just smaller-than-necessary blocks.
  - So what we want is the full lattice structure, but also to explicitly prevent any (direct?) use of join.

Then we questioned if we definitely want to disallow relaxing the ledger limit -- eg maybe benchmarking team would do that to every node on a testnet. But in that case they should just change the limit in the ledger state so that the largest block they'd want to create is valid on _every_ node. They could still use the restriction (ie meet) in order to have some nodes create smaller-than-necessary blocks, if they're going for a mixture.

Ultimately, though, we're wondering who really needs this --- maybe we should rip it out, the ultimate simplification :D

nfrisby:
> Ultimately, though, we're wondering who really needs this --- maybe we should rip it out, the ultimate simplification :D I've asked on the Node Team slack channel for clarification regarding this, included pinging Benchmarking Team lead.

@njd42 and I discussed it. He pointed out that honest SPOs being able to quickly restrict the size of blocks could help us avoid potential disaster scenarios. And some honest SPOs may (currently?) need this to make-up for limited CPU/Mem/etc resources. So we should include it. (Even if we excluded it, it'd be easy enough for some one dedicated to patch into the code.)

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/TxLimits.hs
```diff
@@ -1,14 +1,30 @@
...
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE DerivingVia                #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE TypeApplications           #-}
+{-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE TypeSynonymInstances       #-}
```
Jasagredo, line 11: This can be reduced much further as we aren't using some of these. This combination still works:
```
{-# LANGUAGE DerivingVia                #-}
{-# LANGUAGE FlexibleContexts           #-}
{-# LANGUAGE StandaloneDeriving         #-}
```
(provided you also do this at the bottom of the file):
```
 -- | Apply the override function and then take the pointwise minimum
-applyOverrides :: forall blk.
+applyOverrides ::
      TxLimits blk
```

Jasagredo (approved):
In general, and provided that the Semilattice machinery is moved to `cardano-base`, this PR looks really good.

scripts/buildkite/check-stylish.sh
```diff
@@ -2,6 +2,6 @@
 
 set -euo pipefail
 
-stylish-haskell -i `git ls-files -- 'ouroboros-consensus*/*.hs' | grep -v Setup.hs`
+stylish-haskell -i $(git ls-files -- 'ouroboros-consensus*/*.hs' | grep -v -e 'Setup.hs' -e 'Ouroboros/Consensus/Mempool/TxLimits\.hs')
```
Jasagredo, line 5: Also if this is not already reported, I would report it in the stylish-haskell repository.

nfrisby:
@EncodePanda -- as has become par for the course, I'm surprised that I didn't get this merged during your vacation week. I'll Slack you some details, but please focus on Genesis. I expect we can review this on a quick phone call sometime early this week.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -268,62 +273,53 @@ theLedgerLens f x =
...
-  pointwiseMin (AlonzoMeasure bs1 (ExUnits mem1 steps1)) (AlonzoMeasure bs2 (ExUnits mem2 steps2)) =
-    AlonzoMeasure (bs1 `min` bs2) (ExUnits (mem1 `min` mem2) (steps1 `min` steps2))
+data AlonzoMeasure = AlonzoMeasure {
+    byteSize :: !ByteSize
+  , exUnits  :: !AlonzoExUnits
+  } deriving stock (Eq, Generic, Show)
+    deriving (BoundedMeasure, Measure)
+         via (InstantiatedAt Generic AlonzoMeasure)
+
+-- | Used only to avoid orphan instances
```
nfrisby, line 318: See PR input-output-hk/cardano-ledger-specs#2391

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/TxLimits.hs
nfrisby: Done, thanks.

scripts/buildkite/check-stylish.sh
nfrisby: I opened Issue haskell/stylish-haskell#374

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/TxLimits.hs
```diff
@@ -43,46 +47,58 @@ import           Ouroboros.Consensus.Ticked (Ticked (..))
...
-noOverrides :: Overrides blk
-noOverrides = Overrides id
+-- | @'applyOverrides' 'noOverrides' m = m@
+noOverrides :: TxLimits blk => Overrides blk
+noOverrides = mkOverrides id
 
-mkOverrides :: (Measure blk -> Measure blk) -> Overrides blk
-mkOverrides = Overrides
+-- | Define an override as some update of @'getOverrides' 'noOverrides'@
+mkOverrides :: TxLimits blk => (TxMeasure blk -> TxMeasure blk) -> Overrides blk
```
nfrisby, line 95: Remove the function, change `noOverrides` to `noOverridesMeasure :: BoundedMeasure a => a`.

ouroboros-consensus/src/Ouroboros/Consensus/Mempool/TxLimits.hs
```diff
@@ -31,20 +47,58 @@ import           Ouroboros.Consensus.Ticked (Ticked (..))
...
+{-------------------------------------------------------------------------------
+  Overrides
+-------------------------------------------------------------------------------}
+
+-- | How to override the limits set by the ledger state
+--
+-- The forge logic must use the @Measure.min@ of the limits from ledger state
+-- and this override. More concretely: for each component of @'TxMeasure' blk@
+-- it will use the minimum of the ledger's limits and this override.
+newtype Overrides blk = Overrides {getOverrides :: TxMeasure blk}
```
nfrisby, line 82: Elaborate here: motivate the newtype

  - specifically what kind of `TxMeasure` is this?
  
  - allow a hook for the `Semigroup` and `Monoid` instances to mean conjunction (ie `meet`) instead of addition.
  
Call out the 'Measure' semantics and specifically the lattice-i-ness of it.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
nfrisby: That merged, so I updated the code to remove the orphan-avoiding newtype; yay!

nfrisby:
@EncodePanda I'll resolve the two open Comments today (comments on `Overrides` and the change in type on `noOverrides` and `mkOverrides` to make it simpler for users to understand and use).

Would you review the rest of the PR around those? Those two will be straight forward for me, so please feel free to Approve before I upload fixes for those.

ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
```diff
@@ -148,82 +147,40 @@ data BlockForging m blk = BlockForging {
...
-  -> MaxTxCapacityOverride blk
-  -> Measure blk
-computeMaxTxCapacity ledger maxTxCapacityOverride = case maxTxCapacityOverride of
-      NoMaxTxCapacityOverride              -> ledgerLimit
-      MaxTxCapacityOverride modifyCapacity -> pointwiseMin @blk ledgerLimit (modifyCapacity ledgerLimit)
-  where
-    ledgerLimit = maxCapacity ledger
+  -> TxLimits.Overrides blk
+  -> TxLimits.Measure blk
+computeMaxTxCapacity ledger overrides =
```
EncodePanda: One could argue if the `computeMaxTxCapacity` function is needed anymore, we could might as well call `applyOverrides`.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs
```diff
@@ -294,8 +294,8 @@ instance (SL.PraosCrypto c) => TxLimits (ShelleyBlock (MaryEra c)) where
   pointwiseMin = min
 
 data AlonzoMeasure = AlonzoMeasure {
-    byteSize :: ByteSize
-  , exUnits  :: ExUnits
+    byteSize :: !ByteSize
```
EncodePanda: good catch

EncodePanda (approved):
LGTM, minor comments

ouroboros-consensus/src/Ouroboros/Consensus/Block/Forging.hs
nfrisby: Yeah, 54c9328db5e removed it, applying the overrides in `takeLargestPrefixThatFits`.

nfrisby:
Thank you @EncodePanda.

I rebased just now; automatic, no conflicts 👍

PR #3262, opened 2021-07-12
verifyAbstracTransition: transitions from TerminatedSt

coot:
Allow transitions from `TermiantedSt` to `ReservedOutboundSt` or
`UnnegotiatedSt Inbound` state.

### DIFF
--- a/ouroboros-network-framework/test/Test/Ouroboros/Network/ConnectionManager.hs
+++ b/ouroboros-network-framework/test/Test/Ouroboros/Network/ConnectionManager.hs
@@ -1721,6 +1721,7 @@ verifyAbstractTransition Transition { fromState, toState } =
       --
 
       -- @Reserve@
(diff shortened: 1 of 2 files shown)


bolt12:
Should this be possible in the Non-abstract state as well?

coot:
I am not sure if I follow.  The map from non-abstract to abstract transition is injective, so it is possible in the non-abstract state.

coot:
The `outbound` transition is executed [here](https://github.com/input-output-hk/ouroboros-network/blob/p2p-master/ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs#L1111).

bolt12:
Nevermind, makes sense, the scope is testing so that is why it is on `AbstractState` type

bolt12 (approved):
LGTM

PR #3266, opened 2021-07-14
shelley: fix bug that used maxTxExUnits instead of maxBlockExUnits

### LINKED ISSUE #3265: TxLimits mix up 'maxTxExecutionUnits' and 'maxBlockExecutionUnits'
I think the implementation mixes up the per-transaction execution limits and the per-block execution limits.

https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Mempool.hs#L325

When running the tx-generator benchmarks I noticed that all blocks that are forged only included a very small number of transactions with Plutus scripts.

nfrisby:
Fixes #3265.


MarcFontaine (approved):
LGTM

PR #3270, opened 2021-07-19
Cleanup unused dependencies in cabal files

Jasagredo:
I happened to find that several dependencies were listed in cabal files but were not used at all. Removing them results in no harm.

Feel free to point to specific dependencies that are removed in this PR but you want to be kept as they will be used shortly.

### DIFF
--- a/ouroboros-network-framework/ouroboros-network-framework.cabal
+++ b/ouroboros-network-framework/ouroboros-network-framework.cabal
@@ -83,8 +83,6 @@ library
                      , typed-protocols >=0.1   && < 0.2
                      , Win32-network   >=0.1   && < 0.2
 
-                       -- Used by the handshake protocol for now
-                     , serialise       >=0.2   && <0.3
                        -- Remove once codec is defined locally
                      , typed-protocols-examples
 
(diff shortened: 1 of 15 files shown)


ouroboros-network-framework/ouroboros-network-framework.cabal
```diff
@@ -76,15 +76,11 @@ library
                      , contra-tracer
 
                      , io-classes      >=0.1   && < 0.3
-                     , monoidal-synchronisation
-                                       >=0.1   && < 0.2
```
coot, line 80: Could you leave it, it's used in `p2p-master` branch which sooner or later will be merged to `master`.

Jasagredo: Brought back, thanks!

nfrisby (approved):
LGTM.

As a follow-up, let's consider configuring `weeder` in our CI.

Jasagredo:
> LGTM.
> 
> As a follow-up, let's consider configuring `weeder` in our CI.

I agree we should add weeder as a step on the CI. We just need to identify the "top" modules (or roots as weeder names them). But regarding unused dependencies, weeder can't help us there, it is only for finding dead code:
> Weeder is an application to perform whole-program dead-code analysis.

dcoutts (approved):
:tada:

PR #3272, opened 2021-07-19
Update Haskell.nix for updating HLS to latest version

Jasagredo:
Pulling Haskell-language-server 1.3.0.0 in and updating Haskell.nix meanwhile. 

- `cardano-crypto-class` is removed from `shell.nix` because of this change in Haskell.nix https://github.com/input-output-hk/haskell.nix/commit/d7e5cface8d3b60eb60bd7b1a693587560023c0a
- `Win32` was fixed to 2.6.1.0 by  #3286.


cabal.project
```diff
@@ -1,4 +1,4 @@
-index-state: 2021-02-15T00:00:00Z
```
nfrisby, line 1: I don't know if this should be a blocker, but I have been assuming we get eg better caching by using the same `index-state` that our upstream dependency `cardano-ledger-specs` is using.

Do you have any clarity on what the pros/cons of using a different `index-state` are?

shell.nix
```diff
@@ -19,7 +19,7 @@ let
   #
   # > tools = {
   # >   ...
-  # >   haskell-language-server = "1.1.0";
+  # >   haskell-language-server = "1.2.0";
```
nfrisby, line 22: 1.2.0 or latest?

shell.nix
```diff
@@ -48,16 +46,14 @@ let
...
-      hls.hls-wrapper
-      hls.implicit-hie
     ];
 
     tools = {
       ghcid = "0.8.7";
       hasktags = "0.71.2";
       # https://hackage.haskell.org/package/graphmod
       graphmod = "1.4.4";
+      haskell-language-server = "latest";
```
nfrisby, line 56: 1.2.0 or latest?

nfrisby (commented):
Looks good to me; mostly just removals 👍 

I'll let someone from DevOps actually Approve this PR.

shell.nix
Jasagredo: Should have been "1.2.0.0" but it was complaining :smile:

shell.nix
Jasagredo: Same as above

cabal.project
Jasagredo: I'm quite sure that this only influences the first time Hydra builds the stuff. Once that is done, the whole shell is cached and every dependency is also cached, so it shouldn't impact performance except in the first run.

It is true though that if we use the same as an upstream dependency they would have already built derivations that we will get for free, but it should benefit *only the first compilation*

shell.nix
Jasagredo: Fixed

cabal.project
nfrisby: > it should benefit only the first compilation

I'm not sure when a build result gets added to Hydra. Is it only once it's merged? Or is it anytime a PR is built by CI? Etc.

For example: just me entering the `nix-shell` does not cause my resulting Nix closure to be uploaded to Hydra.

cabal.project
```diff
@@ -30,6 +30,7 @@ constraints:
   , bimap >= 0.4.0
   , ListLike >= 4.7.3
   , witherable >= 0.4
+  , Win32 == 2.6.2.0
```
nfrisby, line 33: This seems suspect to me, but that's mostly because I'm ignorant re windows deps.

@coot, what do you think?

coot: It does not sounds right, why `Win32 >= 2.6.2.0` (which is already declared in `ouroboros-consensus.cabal`), does not work?

cabal.project
Jasagredo: I think Hydra has to build it on CI, and probably it only keeps artifacts that are created when merging into master.

cabal.project
Jasagredo: If I use `Win32 >= 2.6.2.0`, it automatically picks `Win32-2.10.1.1` (not sure why it doesn't jump all the way up to `Win32-2.12.0.1`), but with `2.10.1.1` haskell.nix seems to have some problems to build it:
```
❯ nix build -f release.nix x86_64-w64-mingw32.tests.ouroboros-consensus-cardano-test.test.x86_64-linux
...
[1/0/325 built] building Win32-lib-Win32-2.10.1.1-x86_64-w64-mingw32 (buildPhase): Preprocessing library for Win32-2.10.1.1..
```
It seems to deadlock at the preprocessing step for some reason that escapes my knowledge of Nix. Master (as seen [here](https://hydra.iohk.io/build/7046919#tabs-build-deps)) uses Win32-2.8.5.0, which also makes haskell.nix deadlock in the preprocessing step:
```
❯ nix build -f release.nix x86_64-w64-mingw32.tests.ouroboros-consensus-cardano-test.test.x86_64-linux
...
[1/5/330 built, 1 copied (0.1/0.1 MiB), 0.2 MiB DL] building Win32-lib-Win32-2.8.5.0-x86_64-w64-mingw32 (buildPhase): Preprocessing library for Win32-2.8.5.0..
```

I understand this is not acceptable and I'm not planning on merging this until we have a solution for Win32.

Jasagredo: The compilation doesn't deadlock, just takes a lot of time. When doing this in Hydra, it progresses further and then fails when compiling io-sim library:
```
Preprocessing library for io-sim-0.2.0.0..
Building library for io-sim-0.2.0.0..
[1 of 2] Compiling Control.Monad.IOSim.Internal ( src/Control/Monad/IOSim/Internal.hs, dist/build/Control/Monad/IOSim/Internal.o )
---> Starting remote-iserv on port 7515
---| remote-iserv should have started on 7515
Could not load wine-gecko. HTML rendering will be disabled.
wine: configuration in '/build' has been updated.
Listening on port 7515
remote-iserv.exe:  | /nix/store/fyqacymzm7d7py7n8vw9y51s6zj9l2zk-Win32-lib-Win32-2.10.1.1-x86_64-w64-mingw32/lib/x86_64-windows-ghc-8.10.4/Win32-2.10.1.1-KzpEiQBEAFE8GpgOOc9GS6/HSWin32-2.10.1.1-KzpEiQBEAFE8GpgOOc9GS6.o: unknown symbol `GetNLSVersionEx'
remote-iserv.exe: Could not load Object Code /nix/store/fyqacymzm7d7py7n8vw9y51s6zj9l2zk-Win32-lib-Win32-2.10.1.1-x86_64-w64-mingw32/lib/x86_64-windows-ghc-8.10.4/Win32-2.10.1.1-KzpEiQBEAFE8GpgOOc9GS6/HSWin32-2.10.1.1-KzpEiQBEAFE8GpgOOc9GS6.o.


<no location info>: error:
    ghc: unable to load package `Win32-2.10.1.1'
---> killing remote-iserv...
```

Jasagredo: If #3286 gets merged, we will not have this problem anymore as the version of Win32 will be fixed to 2.6.1.0. The problem of building later versions in nix will still be present but it will be non-urgent then.

Jasagredo: #3286 is merged so we are now using 2.6.1.0.

shell.nix
```diff
@@ -19,7 +19,7 @@ let
   #
   # > tools = {
   # >   ...
-  # >   haskell-language-server = "1.1.0";
+  # >   haskell-language-server = "1.3.0.0";
```
dnadales, line 22: Why do we maintain these comments?

dnadales (approved):

cabal.project
Jasagredo: When Hydra evals and builds the project, it will create artifacts/derivations which will be cached for future builds (until eventual GC). 

It is true that if we don't have the same version of hackage index-state than our upstream dependencies, their derivations will most likely not work for us as they might use a different version of some dependencies, and we will build all of cardano-ledger-specs on our side with our index-state.

I'm running a build at their index-state to see if everything works as expected, and we can build HLS. If so, I will revert the change on the index-state.

PR #3274, opened 2021-07-20
diffusion using io-classes

coot:
This PR allows to run diffusion in io-sim.  It also refactors the p2p
/ non-p2p: using records instead of functions with many arguments and enables
to have simpler `cardano-node` integration.

- diffusion-policies: removed dead code
- ouroboros-network: various stylystic changes
- snocket: support getLocalName for named pipes (Windows)
- netenv: removed nsBoundAddresses
- netenv: more descriptive error messages
- handshake: remove versioned application from HandshakeArguments
- peer-selection: polymorphic address type
- diffusion: generalise to an abstract peer address
- diffusion: run diffusion in an abstract monad
- diffusion: simplify data diffusion
- node-to-client: export LocalSocket constructor

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -1,9 +1,13 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE GADTs               #-}
+{-# LANGUAGE KindSignatures      #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE MonadComprehensions #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE Rank2Types          #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving  #-}
 {-# LANGUAGE TypeApplications    #-}
 -- | Run the whole Node
 --
@@ -62,17 +66,16 @@ import           System.Random (StdGen, newStdGen, randomIO, randomRIO)
 import           Control.Monad.Class.MonadTime (MonadTime)
 import           Control.Monad.Class.MonadTimer (MonadTimer)
 import           Ouroboros.Network.BlockFetch (BlockFetchConfiguration (..))
-import           Ouroboros.Network.Diffusion (DiffusionApplications,
-                     DiffusionArguments, DiffusionTracers, daDiffusionMode,
-                     getDiffusionArguments, mkDiffusionApplicationsNonP2P,
-                     mkDiffusionApplicationsP2P, runDataDiffusion)
+import qualified Ouroboros.Network.Diffusion as Diffusion
+import qualified Ouroboros.Network.Diffusion.NonP2P as NonP2P
+import qualified Ouroboros.Network.Diffusion.P2P as P2P
 import           Ouroboros.Network.Magic
 import           Ouroboros.Network.NodeToClient (ConnectionId, LocalAddress,
-                     NodeToClientVersionData (..), combineVersions,
+                     LocalSocket, NodeToClientVersionData (..), combineVersions,
                      simpleSingletonVersions)
 import           Ouroboros.Network.NodeToNode (DiffusionMode,
                      MiniProtocolParameters, NodeToNodeVersionData (..),
-                     RemoteAddress, blockFetchPipeliningMax,
+                     RemoteAddress, Socket, blockFetchPipeliningMax,
                      defaultMiniProtocolParameters)
 import           Ouroboros.Network.PeerSelection.LedgerPeers
                      (LedgerPeersConsensusInterface (..))
@@ -122,7 +125,7 @@ import           Ouroboros.Consensus.Storage.VolatileDB
 
 -- | Arguments expected from any invocation of 'runWith', whether by deployed
 -- code, tests, etc.
-data RunNodeArgs m addrNTN addrNTC blk = RunNodeArgs {
+data RunNodeArgs m addrNTN addrNTC blk (p2p :: Diffusion.P2P) = RunNodeArgs {
       -- | Consensus tracers
       rnTraceConsensus :: Tracers m (ConnectionId addrNTN) (ConnectionId addrNTC) blk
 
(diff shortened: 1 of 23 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -62,17 +66,17 @@ import           System.Random (StdGen, newStdGen, randomIO, randomRIO)
 import           Control.Monad.Class.MonadTime (MonadTime)
 import           Control.Monad.Class.MonadTimer (MonadTimer)
 import           Ouroboros.Network.BlockFetch (BlockFetchConfiguration (..))
-import           Ouroboros.Network.Diffusion (DiffusionApplications,
-                     DiffusionArguments, DiffusionTracers, daDiffusionMode,
-                     getDiffusionArguments, mkDiffusionApplicationsNonP2P,
-                     mkDiffusionApplicationsP2P, runDataDiffusion)
+import qualified Ouroboros.Network.Diffusion as Diffusion
+-- import qualified Ouroboros.Network.Diffusion.Common as Diffusion
```
bolt12, line 70: Is this comment necessary?

bolt12 (approved):
LGTM! This should also have a corresponding PR in `cardano-node` since you changed the Diffusion Layer

bolt12:
Also it seems CI is not passing

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
coot: removed

coot:
> LGTM! This should also have a corresponding PR in cardano-node since you changed the Diffusion Layer

I almost have it, just need to test if the node can run in both p2p and non-p2p mode.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/Index/Primary.hs
```diff
@@ -641,7 +641,7 @@ backfillChunk chunkInfo chunk (NextRelativeSlot nextExpected) offset =
 ------------------------------------------------------------------------------}
 
 (!) :: (HasCallStack, V.Unbox a) => Vector a -> Int -> a
-v ! i
```
nfrisby, line 644: Is this for `hasktags` or some such? If it's just an aesthetics change, please remove it.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -673,31 +692,44 @@ stdVersionDataNTC networkMagic = NodeToClientVersionData
...
   -> IO ()
-stdRunDataDiffusion = runDataDiffusion
+stdRunDataDiffusion = Diffusion.run
 
 -- | Higher-level arguments that can determine the 'LowLevelRunNodeArgs' under
 -- some usual assumptions for realistic use cases such as in @cardano-node@.
 --
 -- See 'stdLowLevelRunNodeArgsIO'.
-data StdRunNodeArgs m blk = StdRunNodeArgs
+data StdRunNodeArgs m blk (p2p :: Diffusion.P2P) = StdRunNodeArgs
```
nfrisby, line 716: Changing this data type changes the way that `cardano-node` invokes the Consensus layer. So: be advised they might need guidance about how to use these two new arguments etc.

nfrisby (dismissed):
I reviewed the Consensus changes. Thanks for integrating it 👍 

I Requested Changes only because I think that's an unnecessary change and it's the only change in that file; so the git history would be cleaner if we avoid that diff.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/Index/Primary.hs
coot: `ghcid` tripped over it

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
coot: I have a corresponding pr here: https://github.com/input-output-hk/cardano-node/pull/2950

coot: btw, both are to `p2p-master` branch

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/Index/Primary.hs
coot: I can put this in a separate commit; I didn't intended to commit this, but it would be quite useful for the networking team.  Working with network integration is hard because building `lib:ouroboros-consensus` after a simple change in `ouroboros-network` requires re-building > 200 modules which are independent of `ouroboros-netowrk`. `ghcid` allows to mitigate this.  We have an issue #2614 to split the consensus into smaller logical units which I would like to resurrect once we merge `p2p-master` to `master`.

coot: Other than that, I agree that `(!)` notation is ugly.

coot: The error I am getting is:
```
src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/Index/Primary.hs:643:1-3: error:
    The type signature for ‘!’ lacks an accompanying binding
      (The type signature must be given where ‘!’ is declared)
    |
643 | (!) :: (HasCallStack, V.Unbox a) => Vector a -> Int -> a
    | ^^^
src/Ouroboros/Consensus/Storage/ImmutableDB/Impl/Index/Primary.hs:650:12-14: error:
    The INLINE pragma for ‘!’ lacks an accompanying binding
      (The INLINE pragma must be given where ‘!’ is declared)
    |
650 | {-# INLINE (!) #-}
    |            ^^^
```

I think that's caused by `:set -XBangPatterns` in my `~/.ghci`; I'll remove it, and clean the commit.

nfrisby: Huh! I've never come across that. Seems like a GHC bug to me, on first thought -- but many GHC things require more than the first thought...

Thanks for adjusting your setup.

nfrisby (commented):
Thanks for addressing my concern. Dismissing my Request Changes review so that Armando's Approval is unmasked.

PR #3278, opened 2021-07-26
Update Win32-network dependency

### LINKED ISSUE #3271: Include argument value in Windows error "CreateNamedPipe: invalid argument"
I'm trying to run `cardano-node` on Windows, but it fails with the error below:

```
cardano-node.exe: CreateNamedPipe: invalid argument (The filename, directory name, or volume label syntax is incorrect.)
```

The `--socket-path` switch is set to `\\.\pipe\node-bft1`, which appears correct to me.

It would be very helpful if the above error included the actual value of the "invalid argument", so it would be possible to deduce why it's invalid.  
It would also be very helpful if the value of the argument that is printed _does not_ escape backslashes, hereby avoiding ambiguity regarding whether there are too many backslashes in the argument value or whether these backslashes have been escaped for printing.

coot:
Fixes #3271

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -173,8 +173,8 @@ constraints:
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/Win32-network
-  tag: 5b3d08c454f425da5cf045fe7865950d7c806691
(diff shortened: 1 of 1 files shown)


bolt12 (approved):
LGTM

coot:
bors merge

PR #3279, opened 2021-07-28
consensus/docs: draft internal-CHANGELOG.md

nfrisby:
This PR adds a single new file, which contains its own motivation.

I had promised this file to @erikd a while ago, and @rvl just asked me for it. Rodney's request nicely motivated it by explaining that wallet, `db-sync`, etc may be integrating our changes especially long after we merge our PRs. So when I merge a PR and send a helpful Slack message warning them of breaking changes and how to integrate them, it's very easy for them to forget those notes event exist before they actually have to deal with the integration challenges.

### DIFF
--- a/ouroboros-consensus/docs/interface-CHANGELOG.md
+++ b/ouroboros-consensus/docs/interface-CHANGELOG.md
@@ -1,0 +1,117 @@
+# Interface Change Log
+
+- This change log is primarily intended to benefit downstream Cardano
+  development teams. We therefore are not committing to discuss changes that
+  aren't relevant to that audience. If such changes belong anywhere, it would
+  most likely be the `cardano-node` release notes.
+
+- So far, versioning of "internal" Cardano packages is based only on commit
+  hashes, except for the `cardano-node` repo.
+
+For those reasons, we organize this log chronologically: with most recent
+changes near the top.
+
+We recommend three ways to query this log.
+
+- `git diff revOld revNew -- ouroboros-consensus/docs/interface-CHANGELOG.md`
+
+  That command identifies the relevant log entries, those that discuss changes
+  between `revOld` and `revNew`. This is probably what downstream maintainers
+  most often need.
+
+- `git blame [--first-parent] --
+  ouroboros-consensus/docs/interface-CHANGELOG.md`
+
+  That command identifies the commit that introduced each log entry, in case
+  you'd like to learn more about the changes discussed by some entry.
+
+  We make a best effort to add a log entry in the same PR that makes the changes
+  that need to be discussed in this file. Also, the commits on our mainline
+  branch -- which is the only branch we assume downstream teams are using -- are
+  all PR merge commits. We don't do fast-forward merges. Each such merge commit,
+  thanks to `bors`, lists the corresponding PR number. Thus the `--first-parent`
+  flag will tell you which PR to investigate if you're interested in more
+  details/discussion around a particular log entry.
+
+- Manually inspect the dates in the (chronologically sorted) entries. We make a
+  best effort to ensure the entry's date indicates when the corresponding PR was
+  merged. We write "Circa YYYY-MM-DD" to emphasize that, for various reasons,
+  the date might not be exact (time zones, pathologically slow CI runs, etc). If
+  you do need precise timestamps, see instead the `git blame` output.
+
+Inevitably, we will eventually need to add errata to a pre-existing log entry.
+In order for the recommended `diff` command to be minimal and the recommended
+`blame` to be most legible, we make a best effort to avoid editing the lines
+within old entries when updating this file. Those lines are ideally immutable,
+just like their corresponding commits. If we find a particularly problematic
+mistake in an old entry, we address it by adding new lines at the top of that
+log entry. The new list item within the old entry should follow this template `-
+Errata Circa 2022-12-30: The original discussion, still found below,
+oversimplifies the ...`. Such errata only correct mistakes in the original log
+entry; crucially, the code relevant to that entry is not changing. Such code
+changes would justify their own new log entry instead of adding errata to an old
+one. Because of this scheme, the errata dates, unlike the top-level entry dates,
+may appear out of chronological order.
+
+The internals of each entry are organized similar to
+https://keepachangelog.com/en/1.1.0/, adapted to our plan explained above.
+
+## Circa 2021-07-29
+
+### Added
+
+- `ouroboros-consensus/docs/interface-CHANGELOG.md`. Now that this file exists,
+  entries added to it by later PRs will be narrower than this entry. This entry
+  should be the only one that is bootstrapping the file by discussing multiple
+  recent changes.
+
+- The `*MaxTxCapacityOverrides` fields to the `ProtocolParams*` record types,
+  one per ledger era (eg `byronMaxTxCapacityOverrides`,
+  `allegraMaxTxCapacityOverrides`, etc). In particular, these record types are
+  used to organized the lost list of arguments to
+  `Ouroboros.Consensus.Cardano.Node.protocolInfoCardano`. Downstream libraries
+  invoke that function.
+
+  Integration advice: These fields have type
+  `Ouroboros.Consensus.Mempool.TxLimits.Overrides` and therefore can be
+  defaulted with `mempty` or more explicitly with `mkOverrides
+  noOverridesMeasure` (exported by that same module).
+
+  Eventually, the SPOs should be able to set options in configuration files that
+  influence the value of these new override fields in order to restrict the
+  various notions of total transaction size allowed in a block they forge. These
+  limits are already provided by the ledger state, but the SPO may have reason
+  to reduce that limit via this override. The fields are per-era because 1) some
+  eras introduce new dimensions/components of the "transaction size" concept and
+  2) even for a single such component eg script complexity, it can be reasonable
+  for the SPO to set different limits for different eras.
+
+### Changed
+
+- On-the-wire content of Alonzo transactions. Previously they were sent across
+  the wire without the flag indicating whether or not their carried scripts were
+  expected to pass. Now that flag is included on-the-wire.
+
+  The original motivation is so that the local node can reject mistakenly
+  invalid scripts from the local client when the client sent the positive flag
+  but the scripts failed. This is a courtesy to our local client, catching
+  mistakes that otherwise would have incurred on-chain penalties.
+
+  Beyond the original motivation, it is also useful to receive this flag from
+  untrusted peers. If they send the wrong flag, we correct it before propagating
+  it (implemented) and then disconnect (not yet implemented), because the sender
+  is either buggy or adversarial.
+
+  Integration advice: the simplest option is to include the positive flag when
+  creating a transaction to send on the wire. That's likely the intention for
+  our nascent Alonzo functionalities.
+
+  The new ubiquity of this flag also led to simplifications in the interface to
+  `cardano-ledger-specs`. As a result Consensus removed all use of `TxInBlock`,
+  removed the `Validated` outermost layer from the codomain of `extractTxs`, etc
+  -- I don't know to what extent you'll face those transitive changes, but
+  beware of them.
+
+  A new feature could later be added eg to the Cardano CLI so that the client
+  can submit txs with the flag set in the negative, eg for QA to use when
+  testing non-nominal scenarios etc.


ouroboros-consensus/docs/internal-CHANGELOG.md
```diff
@@ -0,0 +1,83 @@
...
+  - So far, versioning of "internal" Cardano libraries is based only on commit
+    hashes, except for the `cardano-node` repo.
+
+For those reasons, we primarily organize this log chronologically: most recent
+changes first.
+
+In order to identify changes by commit hash, Pull Request number, precise
+timestamp, etc, invoke `git blame --first-parent` or `git blame` on this file.
+
+Relatedly, we will NEVER alter entries in this log (until we shift to an
```
rvl, line 17: Is it a problem to fix up or correct old changelog entries? It's always nice to get it right the first time, in the relevant PR, but I don't think anyone would mind previous entries being mutated.

I plan to read the changelog by running `git diff revA revB --  ouroboros-consensus/docs/internal-CHANGELOG.md`.

rvl (approved):
Thanks @nfrisby!

ouroboros-consensus/docs/internal-CHANGELOG.md
nc6: I guess the intention is to keep the alignment between `git blame` and the original commit?

nc6 (approved):
Think this is a great idea. It would be nice if we could include the PR number, but I appreciate that's particularly tricky!

ouroboros-consensus/docs/internal-CHANGELOG.md
```diff
@@ -0,0 +1,83 @@
...
+immutable once their PR is merged into the mainline branch. If there is an
+extremely problematic typo eg, then I suggest pseudo-violating the immutability
+rule only by _adding_ a new corrected line nearby. Ideally such new lines are
+surrounded by lines from the original commit -- this will make the `git blame`
+output easiest to parse.
+
+Each entry is organized similar to https://keepachangelog.com/en/1.1.0/, adapted
+to our plan explained above.
+
+## 2021 Jul 28
```
dnadales, line 28: So in this case we won't be able to retrieve the commits related to the changes below based on this changelog blame. Is this an exception worth mentioning?

ouroboros-consensus/docs/internal-CHANGELOG.md
```diff
@@ -0,0 +1,83 @@
...
+
+  Integration advice: These fields have type
+  `Ouroboros.Consensus.Mempool.TxLimits.Overrides` and therefore can be
+  defaulted with `mempty` or more explicitly with `mkOverrides
+  noOverridesMeasure` (exported by that same module).
+
+  Eventually, the SPOs should be able to set options in configuration files that
+  influence the value of these new override fields in order to restrict the
+  various notions of total transaction size allowed in a block they forge. These
+  limits are already provided by the ledger state, but the SPO may have reason
```
DavidEichmann, line 51: reason *to* reduce

ouroboros-consensus/docs/internal-CHANGELOG.md
DavidEichmann, line 28: It's unclear to me how we should use dates. This entry is large as it incorporates a bunch of recent changes. In the future are we intending to simple add a data and only 1 (or few) changes corresponding to e.g. a single PR? In that case we need need to worry too much about keeping everything immutable; we can just keep the date line immutable and look at the commit for the date line via git blame.

DavidEichmann (changes requested):

ouroboros-consensus/docs/internal-CHANGELOG.md
DavidEichmann: @nc6 that's my understanding, but as i mention [below](https://github.com/input-output-hk/ouroboros-network/pull/3279#discussion_r679113587) I think we can weaken this requirement to simply the date lines being immutable.

Jasagredo (commented):
I am okey with the idea of the ChangeLog although I would suggest a couple of improvements:
- Keep links to code/PRs as changes can span several PRs.
- Allow appending information to an entry to further refine it. If something is fundamentally wrong, probably just strikethrough it and add a reference to some explanation below, instead of deleting it.

nfrisby:
Thanks for the comments everyone.

I think the immutability would be nice, but I likely overstated how important it is.

I'll update the draft based on the following notes, prompted by your comments. Thanks!

  - Downstream users deal with commit hashes. They have at least: the latest hash they've already integrated as well as the new hash they're considering how to integrate. This enables Rodney's clever plan of `git diff revA revB -- ouroboros-consensus/docs/internal-CHANGELOG.md`. I'll include that recommended command in the doc itself. If the user does that, then those of us writing the document can't really cause too much trouble unless we do something silly like change every line (eg DOS -> WIN line endings).
  
  - Moreover, downstream users deal with commit hashes from our mainline branch. That's where to find our "releases".
  
  - As of some point in our development history, all mainline commits are PR merge commits that list the PR number (ie we only use `bors` and never fast-forwards etc). Thus, indirectly, downstream users are picking some prefix of all of our merged PRs. But to emphasize: they don't ultimately care about PR numbers; they just want to know what changed between two hashes. The PR numbers would only help them (or someone else) do archaeology to figure out more details about a particular change. The hope is that the entries in this file will be complete enough that they won't have to do that.

  - I want the PR that makes the change to also add an entry to this log file. If we forget, and have to add the entry in a later PR, so be it -- we're human. But the goal is for the PR that makes the change to also update the log file. That would yield the most natural results.

  - We could put PR numbers in the doc itself: that number will already be known before the commit is finalized. But PR numbers are generally not of interest to our downstream users. And can be recovered more directly via a `git log` command.
  
    ```
    $ git log --first-parent --oneline 6cef814ca^..999f25035
    999f25035 Merge #3261
    f026e71e8 Merge #3278
    681351c06 Merge #3270
    36cb0b970 Merge #3230
    6cef814ca Merge #3275
    ```

  - Unlike PR numbers, we cannot predict the merge commit hash (that'd be self-referential). The idea to rely on `git blame` lets us pretend that the correct hash is automatically present in the file itself, since `git blame --first-parent` or `git blame` will display the relevant PR merge commit hash right next to each log entry.
  
  - I proposed including the date just so there is something in the file itself that indicates the time spans between entries. As David mentioned, it's not clear exactly which date this is/should be. Ideally, it's the timestamp of when the PR was merged; but that's self-referential. We could make the last step before merging the PR be to update the timestamp in the log, but that'll be easy to forget. So I think a practical and least-surprising/most-honest semantics is that the date in the file is "some time near" the PR merge timestamp.

ouroboros-consensus/docs/internal-CHANGELOG.md
nfrisby: I think the new text I just wrote does a better job of that than the old "bootstrapping" word I was hoping would cover it all by itself.

Please let me know if that's clearer now.

ouroboros-consensus/docs/internal-CHANGELOG.md
nfrisby: Thanks for the sharing the explicit command, Rodney. I found that very helpful.

Clarke: yes.

David: I'd say you're ultimately right about the dateline being the only necessarily immutable one, but I still think there is value in treating the entries as append-only.

ouroboros-consensus/docs/internal-CHANGELOG.md
nfrisby: I think my new introduction/overview section addresses this. Please let me know if not.

nfrisby:
I've updated the PR, and I think the introduction/overview is much clearer now. Please let me know whether it addresses all your discussion points.

ouroboros-consensus/docs/interface-CHANGELOG.md
```diff
@@ -0,0 +1,117 @@
+# Interface Change Log
```
nfrisby, line 1: @Jasagredo said: 

> I am okey with the idea of the ChangeLog although I would suggest a couple of improvements:
>
> - Keep links to code/PRs as changes can span several PRs.
> - Allow appending information to an entry to further refine it. If something is fundamentally wrong, probably just strikethrough it and add a reference to some explanation below, instead of deleting it.

- Hmm. I think we should have one PR per log entry. We don't merge a PR until it builds, so at the very least some coherent portion of the ultimate change will already be in place. The log entry for each preparatory PR can at least list what changed and then defer more details by explicitly saying "a subsequent log entry will say more". How disruptive/challenging do you think that scheme would be for us?
- If we add the `<s>` and `</s>` tags on their own line, then we can strikethrough the surrounded lines without altering them, thereby preserving the `git blame` output 👍. I tested this a bit and if even works for surrounded headers as long as you add blanks lines around the strikethrough tags.

ouroboros-consensus/docs/interface-CHANGELOG.md
```diff
@@ -0,0 +1,117 @@
...
+  merged. We write "Circa YYYY-MM-DD" to emphasize that, for various reasons,
+  the date might not be exact (time zones, pathologically slow CI runs, etc). If
+  you do need precise timestamps, see instead the `git blame` output.
+
+Inevitably, we will eventually need to add errata to a pre-existing log entry.
+In order for the recommended `diff` command to be minimal and the recommended
+`blame` to be most legible, we make a best effort to avoid editing the lines
+within old entries when updating this file. Those lines are ideally immutable,
+just like their corresponding commits. If we find a particularly problematic
+mistake in an old entry, we address it by adding new lines at the top of the log
```
dnadales, line 47: at the top of the *old* log entry I assume. I wonder if this could be interpreted as adding new lines at the top of a new log entry.

ouroboros-consensus/docs/interface-CHANGELOG.md
```diff
@@ -0,0 +1,117 @@
...
+entry. The new list item within the old entry should follow this template `-
+Errata Circa 2022-12-30: The original discussion, still found below,
+oversimplifies the ...`. Such errata only correct mistakes in the original log
```
dnadales, line 50:
```suggestion
oversimplifies the ...`. Such errata only corrects mistakes in the original log
```

ouroboros-consensus/docs/internal-CHANGELOG.md
dnadales: If you refer to this line:

> Now that this file exists,
>  entries added to it by later PRs will be narrower than this entry. This entry
>  should be the only one that is bootstrapping the file by discussing multiple
>  recent changes.

then yeah, it is clearer now. Thanks.

ouroboros-consensus/docs/interface-CHANGELOG.md
nfrisby: `errata` is plural, so `correct` is ... correct. Yeah?

ouroboros-consensus/docs/interface-CHANGELOG.md
nfrisby: Fixed, thanks. I changed `top of the log entry` to `top of that log entry`.

nfrisby:
I think I've now responded to everyone's comments. Please have a look again if you recall unfinished business :)

ouroboros-consensus/docs/interface-CHANGELOG.md
dnadales: Oh, I see. Sorry about that.

ouroboros-consensus/docs/interface-CHANGELOG.md
```diff
@@ -0,0 +1,117 @@
...
+  the wire without the flag indicating whether or not their carried scripts were
+  expected to pass. Now that flag is included on-the-wire.
+
+  The original motivation is so that the local node can reject mistakenly
+  invalid scripts from the local client when the client sent the positive flag
+  but the scripts failed. This is a courtesy to our local client, catching
+  mistakes that otherwise would have incurred on-chain penalties.
+
+  Beyond the original motivation, it is also useful to receive this flag
+  untrusted peers. If they send the wrong flag, we correct it before propagating
```
DavidEichmann, line 101: I feel like part of this sentence was deleted. Do you mean "... useful to receive this flag in order to detect untrusted peers"?

DavidEichmann: ... or a missing "from"

ouroboros-consensus/docs/internal-CHANGELOG.md
DavidEichmann: @nfrisby absolutely agree

ouroboros-consensus/docs/internal-CHANGELOG.md
DavidEichmann: Perfect! Thanks :-)

ouroboros-consensus/docs/interface-CHANGELOG.md
```diff
@@ -0,0 +1,117 @@
...
+Inevitably, we will eventually need to add errata to a pre-existing log entry.
+In order for the recommended `diff` command to be minimal and the recommended
+`blame` to be most legible, we make a best effort to avoid editing the lines
+within old entries when updating this file. Those lines are ideally immutable,
+just like their corresponding commits. If we find a particularly problematic
+mistake in an old entry, we address it by adding new lines at the top of that
+log entry. The new list item within the old entry should follow this template `-
+Errata Circa 2022-12-30: The original discussion, still found below,
+oversimplifies the ...`. Such errata only correct mistakes in the original log
+entry; crucially, the the code relevant to that entry is not changing. Such code
```
Jasagredo, line 51:
```suggestion
entry; crucially, the code relevant to that entry is not changing. Such code
```

Jasagredo (approved):
Looks good to me

ouroboros-consensus/docs/interface-CHANGELOG.md
nfrisby: Fixed, thanks!

ouroboros-consensus/docs/interface-CHANGELOG.md
nfrisby: Fixed, thanks!

nfrisby:
Hi @DavidEichmann. I think we're good to go here. Would you dismiss your Request Changes review? And then invoke `bors r+`? Thanks!

DavidEichmann (approved):

PR #3281, opened 2021-07-29
Bolt12/p2p master attenuation

bolt12:
Adds Attenuation to tests and fixes a couple of bugs found in the way.

NOTE: These changes break some of the tests, there's still WIP

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
@@ -661,11 +661,10 @@ withConnectionManager ConnectionManagerArguments {
                   return ( Map.delete peerAddr state
                          , Left (Known (TerminatedState Nothing))
                          )
-                Here (connVar, transition) -> do
-                  traceWith trTracer transition
+                Here connVarAndTransition -> do
                   close cmSnocket socket
                   return ( state
-                         , Right connVar
+                         , Right connVarAndTransition
                          )
 
             case mConnVar of
@@ -680,7 +679,7 @@ withConnectionManager ConnectionManagerArguments {
                                          , toState   = Unknown
                                          })
               Left _ -> error "Impossible happened"
-              Right connVar ->
+              Right (connVar, transition) ->
                 do traceWith tracer (TrConnectionTimeWait connId)
                    when (cmTimeWaitTimeout > 0) $
                      unmask (threadDelay cmTimeWaitTimeout)
@@ -690,6 +689,8 @@ withConnectionManager ConnectionManagerArguments {
                   -- - handshake negotiation; or
                   -- - `Terminate: TerminatingState → TerminatedState` transition.
                   traceWith tracer (TrConnectionTimeWaitDone connId)
+                  traceWith trTracer transition
+                  -- ^ TerminatingState -> TerminatedState transition
                   trs <- atomically $ do
                     --  We have to be careful when deleting it from
                     --  'ConnectionManagerState'.
--- a/ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
@@ -566,168 +566,161 @@ withConnectionManager ConnectionManagerArguments {
         . connectionManagerStateToCounters
       <$> traverse readTVar st
 
-    -- Start connection thread and run connection handler on it.
+
+    -- Fork connection thread.
     --
     -- TODO: We don't have 'MonadFix' instance for 'IOSim', so we cannot
     -- directly pass 'connVar' (which requires @Async ()@ returned by this
     -- function.  If we had 'MonadFix' at hand We could then also elegantly
     -- eliminate 'PromiseWriter'.
-    runConnectionHandler :: StrictTMVar m (ConnectionManagerState peerAddr handle handleError version m)
-                         -> ConnectionHandlerFn handlerTrace socket peerAddr handle handleError version m
-                         -> socket
-                         -> peerAddr
-                         -> PromiseWriter m (Either handleError (handle, version))
-                         -> m (ConnectionId peerAddr, Async m ())
-    runConnectionHandler stateVar handler socket peerAddr writer = do
-      localAddress <- getLocalAddr cmSnocket socket
-      let connId = ConnectionId { remoteAddress = peerAddr
-                                , localAddress
-                                }
+    forkConnectionHandler
+      :: StrictTMVar m (ConnectionManagerState peerAddr handle handleError version m)
+      -> socket
+      -> ConnectionId peerAddr
+      -> PromiseWriter m (Either handleError (handle, version))
+      -> ConnectionHandlerFn handlerTrace socket peerAddr handle handleError version m
+      -> m (Async m ())
+    forkConnectionHandler stateVar
+                          socket
+                          connId@ConnectionId { remoteAddress = peerAddr }
+                          writer
+                          handler =
+        asyncWithUnmask $ \unmask ->
+          runWithUnmask
+            (handler socket writer
+                     (TrConnectionHandler connId `contramap` tracer)
+                     connId
+                     (\bearerTimeout ->
+                       toBearer
+                         cmSnocket
+                         bearerTimeout
+                         (WithMuxBearer connId `contramap` muxTracer)))
+            unmask
+          `finally` cleanup
+      where
+        cleanup :: m ()
+        cleanup =
+          -- We must ensure that we update 'connVar',
+          -- `requestOutboundConnection` might be blocked on it awaiting for:
+          -- - handshake negotiation; or
+          -- - `Terminate: TerminatingState → TerminatedState` transition.
+          -- That's why we use 'uninterruptibleMask'. Note that this cleanup
+          -- function after all is interruptible, because we unmask async
+          -- exceptions around 'threadDelay', but even if an async exception
+          -- hits there we will update `connVar`.
+          uninterruptibleMask $ \unmask -> do
+            traceWith tracer (TrConnectionCleanup connId)
+            mConnVar <- modifyTMVar stateVar $ \state -> do
+              wConnVar <- uninterruptibleMask_ $ atomically $ do
+                case Map.lookup peerAddr state of
+                  Nothing      -> return Nowhere
+                  Just connVar -> do
+                      connState <- readTVar connVar
+                      case connState of
+                        ReservedOutboundState -> do
+                          writeTVar connVar (TerminatedState Nothing)
+                          return $ There connState
+                        UnnegotiatedState {} -> do
+                          writeTVar connVar (TerminatedState Nothing)
+                          return $ There connState
+                        OutboundUniState {} -> do
+                          writeTVar connVar (TerminatedState Nothing)
+                          return $ There connState
+                        OutboundDupState {} -> do
+                          writeTVar connVar (TerminatedState Nothing)
+                          return $ There connState
+                        OutboundIdleState {} -> do
+                          writeTVar connVar (TerminatedState Nothing)
+                          return $ There connState
+                        InboundIdleState {} -> do
+                          writeTVar connVar (TerminatedState Nothing)
+                          return $ There connState
+                        InboundState {} -> do
+                          writeTVar connVar (TerminatedState Nothing)
+                          return $ There connState
+                        DuplexState {} -> do
+                          writeTVar connVar (TerminatedState Nothing)
+                          return $ There connState
+                        TerminatingState {} -> do
+                          return $ Here connVar
+                        TerminatedState {} ->
+                          return $ There connState
+              case wConnVar of
+                Nowhere -> do
+                  close cmSnocket socket
+                  return ( state
+                         , Left Unknown
+                         )
+                There connState -> do
+                  close cmSnocket socket
+                  return ( Map.delete peerAddr state
+                         , Left (Known connState)
+                         )
+                Here connVar -> do
+                  close cmSnocket socket
+                  return ( state
+                         , Right connVar
+                         )
 
-      let cleanup :: m ()
-          cleanup =
-            -- We must ensure that we update 'connVar',
-            -- `requestOutboundConnection` might be blocked on it awaiting for:
-            -- - handshake negotiation; or
-            -- - `Terminate: TerminatingState → TerminatedState` transition.
-            -- That's why we use 'uninterruptibleMask'. Note that this cleanup
-            -- function after all is interruptible, because we unmask async
-            -- exceptions around 'threadDelay', but even if an async exception
-            -- hits there we will update `connVar`.
-            uninterruptibleMask $ \unmask -> do
-              traceWith tracer (TrConnectionCleanup connId)
-              mConnVar <- modifyTMVar stateVar $ \state -> do
-                wConnVar <- uninterruptibleMask_ $ atomically $ do
-                  case Map.lookup peerAddr state of
-                    Nothing      -> return Nowhere
-                    Just connVar -> do
-                        connState <- readTVar connVar
-                        case connState of
-                          ReservedOutboundState -> do
-                            writeTVar connVar (TerminatedState Nothing)
-                            return $ There connState
-                          UnnegotiatedState {} -> do
-                            writeTVar connVar (TerminatedState Nothing)
-                            return $ There connState
-                          OutboundUniState {} -> do
-                            writeTVar connVar (TerminatedState Nothing)
-                            return $ There connState
-                          OutboundDupState {} -> do
-                            writeTVar connVar (TerminatedState Nothing)
-                            return $ There connState
-                          OutboundIdleState {} -> do
-                            writeTVar connVar (TerminatedState Nothing)
-                            return $ There connState
-                          InboundIdleState {} -> do
-                            writeTVar connVar (TerminatedState Nothing)
-                            return $ There connState
-                          InboundState {} -> do
-                            writeTVar connVar (TerminatedState Nothing)
-                            return $ There connState
-                          DuplexState {} -> do
-                            writeTVar connVar (TerminatedState Nothing)
-                            return $ There connState
-                          TerminatingState {} -> do
-                            return $ Here connVar
-                          TerminatedState {} ->
-                            return $ There connState
-                case wConnVar of
-                  Nowhere -> do
-                    close cmSnocket socket
-                    return ( state
-                           , Left Unknown
-                           )
-                  There connState -> do
-                    close cmSnocket socket
-                    return ( Map.delete peerAddr state
-                           , Left (Known connState)
-                           )
-                  Here connVar -> do
-                    close cmSnocket socket
-                    return ( state
-                           , Right connVar
-                           )
+            case mConnVar of
+              Left !connState -> do
+                traceCounters stateVar
+                traceWith trTracer (TransitionTrace peerAddr
+                                      Transition
+                                         { fromState = connState
+                                         , toState   = Unknown
+                                         })
+              Right connVar ->
+                do traceWith tracer (TrConnectionTimeWait connId)
+                   when (cmTimeWaitTimeout > 0) $
+                     unmask (threadDelay cmTimeWaitTimeout)
+                `finally` do
+                  -- We must ensure that we update 'connVar',
+                  -- `requestOutboundConnection` might be blocked on it awaiting for:
+                  -- - handshake negotiation; or
+                  -- - `Terminate: TerminatingState → TerminatedState` transition.
+                  traceWith tracer (TrConnectionTimeWaitDone connId)
+                  trs <- atomically $ do
+                    mConnState <- readTMVar stateVar
+                              >>= traverse readTVar . Map.lookup peerAddr
+                    -- We can always write to `connVar`, since a new
+                    -- connection will use a new 'TVar', but we have to be
+                    -- careful when deleting it from 'ConnectionManagerState'.
+                    let connState'  = TerminatedState Nothing
+                    writeTVar connVar connState'
+                    updated <-
+                      modifyTMVarPure
+                        stateVar
+                        ( swap
+                        . Map.updateLookupWithKey
+                            (\_ v ->
+                              -- only delete if it wasn't replaced
+                              if eqTVar (Proxy :: Proxy m) connVar v
+                                then Nothing
+                                else Just v
+                            )
+                            peerAddr
+                        )
+                    let connState   = maybe Unknown Known mConnState
+                        kConnState' = Known connState'
 
-              case mConnVar of
-                Left !connState -> do
+                    case updated of
+                      Nothing ->
+                        return [ Transition { fromState = connState
+                                            , toState   = kConnState'
+                                            }
+                               , Transition { fromState = kConnState'
+                                            , toState   = Unknown
+                                            }
+                               ]
+                      Just _ ->
+                        return [ Transition { fromState = connState
+                                            , toState   = kConnState'
+                                            }
+                               ]
+
+                  traverse_ (traceWith trTracer . TransitionTrace peerAddr) trs
                   traceCounters stateVar
-                  traceWith trTracer (TransitionTrace peerAddr
-                                        Transition
-                                           { fromState = connState
-                                           , toState   = Unknown
-                                           })
-                Right connVar ->
-                  do traceWith tracer (TrConnectionTimeWait connId)
-                     when (cmTimeWaitTimeout > 0) $
-                       unmask (threadDelay cmTimeWaitTimeout)
-                  `finally` do
-                    -- We must ensure that we update 'connVar',
-                    -- `requestOutboundConnection` might be blocked on it awaiting for:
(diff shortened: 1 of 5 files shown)


coot:
Could you rebase the branch so that `fixup! server-test: use DataFlowProtocolData` is applied.

ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Unversioned.hs
```diff
@@ -66,6 +68,34 @@ unversionedProtocol :: app
 unversionedProtocol = simpleSingletonVersions UnversionedProtocol UnversionedProtocolData
 
 
+-- | Alternative for 'UnversionedProtocolData' which contains 'DataFlow'.
+--
+newtype DataFlowProtocolData = DataFlowProtocolData { getProtocolDataFlow :: DataFlow }
```
coot, line 73:
```suggest
newtype DataFlowProtocolData =
    DataFlowProtocolData { getProtocolDataFlow :: DataFlow }

```

ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Unversioned.hs
```diff
@@ -66,6 +68,34 @@ unversionedProtocol :: app
...
+      decodeTerm (CBOR.TBool False) = Right (DataFlowProtocolData Unidirectional)
+      decodeTerm (CBOR.TBool True)  = Right (DataFlowProtocolData Duplex)
+      decodeTerm t                  = Left $ T.pack $ "unexpected term: " ++ show t
+
+dataFlowProtocol :: DataFlow
+                 -> app
+                 -> Versions UnversionedProtocol
+                             DataFlowProtocolData
+                             app
+dataFlowProtocol dataFlow = simpleSingletonVersions UnversionedProtocol (DataFlowProtocolData dataFlow)
```
coot, line 97:
```suggest
dataFlowProtocol dataFlow =
    simpleSingletonVersions UnversionedProtocol (DataFlowProtocolData dataFlow)
```

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -1722,6 +1723,33 @@ prop_multinode_Sim serverAcc (ArbDataFlow dataFlow) absBi script =
         Nothing -> IdleConn
         Just {} -> ActiveConn
 
+
+-- | Connection terminated while negotiating it.
+--
+unit_1 :: Property
+unit_1 = prop_multinode_Sim
```
coot, line 1772:
```suggest
unit_connection_terminated_when_negotiatin :: Property
unit_connection_terminated_when_negotiating =
    prop_multinode_Sim
```

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -637,6 +651,24 @@ withBidirectionalConnectionManager name timeouts trTracer snocket socket localAd
             }
 
 
+reqRespSizeLimits :: forall req resp. ProtocolSizeLimits (ReqResp req resp) ByteString
```
coot, line 657:
```suggest
reqRespSizeLimits :: forall req resp. ProtocolSizeLimits (ReqResp req resp)
                                                          ByteString
```

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1488,8 +1488,8 @@ withConnectionManager ConnectionManagerArguments {
                 return (DemoteToColdLocalNoop Nothing
                                               (abstractState $ Known connState))
 
-              InboundIdleState _connId _connThread _handle dataFlow ->
-                assert (dataFlow == Duplex) $
+              InboundIdleState _connId _connThread _handle _dataFlow ->
+                -- assert (dataFlow == Duplex) $
```
coot, line 1482: Why this assertion is disabled?

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -689,7 +721,18 @@ mkSnocket state tr = Snocket { getLocalAddr
                                    (Just remoteAddress)
                                    (STBearerInfo bearerInfo))
             -- connection delay
-            threadDelay (biConnectionDelay bearerInfo)
+            unmask (threadDelay (biConnectionDelay bearerInfo `min` connectTimeout))
+              `catch` \(_ :: SomeException) ->
+                atomically $ modifyTVar (nsConnections state)
+                                        (Map.delete (normaliseId connId))
```
coot, line 727: Is this right thing to do? `SomeException` most likely mean an asynchronous exception in the application.  What we do here is automatic resource cleanup by on the OS side.  I am not sure if cancelling `connect` would behave like that on `*nix` platforms.

Let's ask Neil (@njd42) and @karknu.

ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
```diff
@@ -111,6 +111,7 @@ run :: forall muxMode socket peerAddr versionNumber m a b.
        , MonadCatch    m
        , MonadEvaluate m
        , MonadLabelledSTM  m
+       , MonadMask     m
```
coot, line 114: I think I was involved in writing this [commit](https://github.com/input-output-hk/ouroboros-network/pull/3281/commits/dc87fe9844e9d8f80aa6abcea7788616543cc8ec) too, isn't it; if that's the case one can add `Authors: ...` in the commit message (git does not support multi-authored commits.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1110,16 +1106,10 @@ withConnectionManager ConnectionManagerArguments {
...
-                  let connState' = ReservedOutboundState
-                  writeTVar connVar connState'
-                  return ( Just (Left (TransitionTrace
-                                        peerAddr
-                                        (mkTransition connState connState')))
-                         , connVar
-                         , Right Nowhere
-                         )
+                  -- the connection terminated; we can not reset 'connVar' and
+                  -- start afresh. We should wait for the removal of the connection from
```
coot, line 1110: line too long

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -875,8 +865,14 @@ withConnectionManager ConnectionManagerArguments {
             -- idle, it will call 'unregisterInboundConnection' which will
             -- perform the aforementioned @Commit@ transition.
 
-            traceWith trTracer (TransitionTrace peerAddr transition)
-            return (Connected connId dataFlow handle)
+            traverse_ (traceWith trTracer . TransitionTrace peerAddr) mbTransition
+
+            -- If mbTransition is Nothing, it means that the connVar was read either in
+            -- Terminating or TerminatedState. Either case we should return Disconnected
+            -- instead of Connected.
```
coot, line 872: lines too long

coot (changes requested):
This is quite large PR, we should avoid this in the future (reviewing is quite hard of such large changes).  If needed an issue can always be broken into multiple PRs.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -1552,8 +1552,8 @@ data EffectiveDataFlow
     deriving (Eq, Show)
 
 -- | Property wrapping `multinodeExperiment`.
-prop_multinode_Sim :: Int -> ArbDataFlow -> MultiNodeScript Int TestAddr -> Property
-prop_multinode_Sim serverAcc (ArbDataFlow dataFlow) script =
+prop_multinode_Sim :: Int -> ArbDataFlow -> AbsBearerInfo -> MultiNodeScript Int TestAddr -> Property
```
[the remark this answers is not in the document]
bolt12: If I generate a list than the test fails calling the partial function `last` and `unidirectional_Sim` also fails. I'm not sure if a non-empty list is the best as for the tests to pass I believe we should have the last `BearerInfo` used to be `noAttenuation`

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1157,8 +1163,8 @@ withConnectionManager ConnectionManagerArguments {
               -- Both the 'bracketOnError''s resource finaliser and the
               -- connection handler cleanup function are responsible for:
               --
-              -- * closing the socket
-              -- * freeing the slot in connection manager state map
+              --    * closing the socket
+              --    * freeing the slot in connection manager state map
```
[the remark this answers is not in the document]
bolt12: It does not, I ought to fix this!

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -821,6 +821,10 @@ data ConnectionManagerTrace peerAddr handlerTrace
   | TrConnectionManagerCounters  !ConnectionManagerCounters
   | TrState                      !(Map peerAddr AbstractState)
   -- ^ traced on SIGUSR1 signal, installed in 'runDataDiffusion'
+  | TrUnexpectedlyMissingConnectionState !(ConnectionId peerAddr)
+  -- ^ This is triggered by a failed lookup on the 'ConnectionManagerState'
+  -- using the ConnectionId's peerAddr, during connection handler cleanup
+  -- function.
```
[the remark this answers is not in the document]
bolt12: From what I can remember from our session together this shouldn't really be possible to happen. `TrUnexpectedlyMissingConnectionState` is here mostly for us to be able to be total in our case expression over at connection handler cleanup. I'll dig more into this and try to make a more explanatory comment!

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
bolt12: Yes, it means an asynchronous exception. Well I am not really sure if cancelling connect would behave like this but this is what I thought needs to be done to fix the test environment. Without this catch and cleanup thee test would fail with a `NotReleasedConnections` error from an `Half-Close`d peer.

ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
bolt12: Yes you were! I will add that!

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
bolt12: This is written in the commit message, by you:

>  server-test: track when to call unregisterOutboundConnection

> This is not perfect, so we also remove an assertion from
connection-manager.  This assertion is benign and also hit rarely (once
per 100_000 simulations).

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
coot: Why this  test failure is not a resource leak?

ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
coot: Thanks, then a log reader might know where to ask.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
bolt12: I'm not sure I understand your question. The test was failing with a `NotReleasedConnections`. And it was due an `AsyncCancelled` exception thrown that would make everything die without cleaning up the state.

coot: `NotReleasedConnections` means that there is a resource leak in the test.  If throwing somewhere in the application an asynchronous exceptions leads to `NotReleasedConnections` this should rather be fixed in the application not the simulation environment.

coot: This could be as simple as using `bracket` somewhere at the top level (where the connection is created) or maybe it's an actual bug in the server.

Even if this is not a bug in connection-manager or server, such a fix would hide a true resource leak bug in diffusion (which should not leak any connections).

I bet this is an error somewhere inside the test itself (`prop_multinode_Sim`).  It doesn't seem very careful about resources.

bolt12: Okay I got confused, this patch appears here but shoudn't, this is probably due to some mistake of mine while rebasing. 

I fetched the original commit from you: https://github.com/input-output-hk/ouroboros-network/commit/d5026a1694bc6c3ea934a400293ae35c0a68fae2

coot: yes, that's my mistake.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
coot: ok; let's just add a `TODO` there.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
bolt12: And the thing I was making confusion was this one: https://github.com/input-output-hk/ouroboros-network/pull/3309/commits/885b39aee1f1257ea6f016901a44ef677f4362bd#diff-9dc60e7be09dd9b4174122cea073cec55fb005ee452d64dafab6a12cfce6ed8bR728

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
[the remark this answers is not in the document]
coot: > If I generate a list than the test fails calling the partial function last and unidirectional_Sim also fails.

Let's try to fix this.

> I'm not sure if a non-empty list is the best as for the tests to pass I believe we should have the last BearerInfo used to be noAttenuation

That's right. It's common to have such invariants inside generators, i.e. generate non empty lists which last element is `noAttenuation`.  One needs to make sure that shrinking preserves the invariant (which in this case is easy to do).

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Types.hs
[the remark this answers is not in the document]
coot: Then we can add a test that verifies that this does not happen or even include an assertion. Note that assertions are disabled in the production system, we only use them in our tests.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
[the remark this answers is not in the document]
bolt12: What I meant by last used was exactly the last one used, not necessarily the last one in the list, does this make sense?

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
coot: I take my word back.  We can catch an exceptions inside `connect` and update the simulation state. We're not closing the connection there (which we cannot do), we just remove it from the map of existing connections which is fine.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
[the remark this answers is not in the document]
coot: Ah, I see. That's ok then as it would need to redesign the test.   The diffusion simulation will not need to relay on such aninvariant; the outbound governor is the component which makes sure that a progress is made.  Maybe just leave a note why we do this this way, mention the limitation of the test (there's nothing that will guarantee progress like outbound governor).

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -818,8 +818,8 @@ unidirectionalExperiment timeouts snocket socket clientAndServerData = do
                 (property True)
                 $ zip rs (expectedResult clientAndServerData clientAndServerData)
 
-prop_unidirectional_Sim :: AbsBearerInfo -> ClientAndServerData Int -> Property
-prop_unidirectional_Sim absBi clientAndServerData =
+prop_unidirectional_Sim :: NonFailingBearerInfoScript -> ClientAndServerData Int -> Property
```
coot, line 821: the line is too long:
```suggestion
prop_unidirectional_Sim :: NonFailingBearerInfoScript
                        -> ClientAndServerData Int
                        -> Property
```

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -852,19 +854,9 @@ withConnectionManager ConnectionManagerArguments {
                 DuplexState {} ->
                   throwSTM (withCallStack (ImpossibleState peerAddr))
 
-                TerminatingState {} -> do
-                  let connState' = InboundIdleState
-                                     connId connThread handle
-                                     (connectionDataFlow version)
-                  writeTVar connVar connState'
-                  return (mkTransition connState connState')
+                TerminatingState {} -> return Nothing
```
coot, line 857: It's not clear to me if this is a problem with the testing environment or the connection manager itself.  

Hmm, I think I see now: when we are in `TerminatingState` or `TerminatedState` the socket might be already closed so we should not try to use it.  Would you agree with that?

This is a nice example bug caught by the testing framework, please create a jira ticket for it and link it back here.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -1730,9 +1730,12 @@ verifyRemoteTransition Transition {fromState, toState} =
...
       -- This might happen if starting any of the responders errored.
       (Nothing,           Nothing)           -> True
-      -- @RemoteWarmSt → RemoteWarmSt@ trnasition is observed if a hot or warm
-      -- protocol terminates (which triggers @RemoteEstSt -> RemoteWarmSt@)
+      -- @RemoteWarmSt → RemoteWarmSt@, @RemoteIdleSt → RemoteIdleSt@ and
+      -- @RemoteColdSt → RemoteColdSt@ transition are observed if a hot or
+      -- warm protocol terminates (which triggers @RemoteEstSt -> RemoteWarmSt@)
       (Just RemoteWarmSt, Just RemoteWarmSt) -> True
+      (Just RemoteIdleSt, Just RemoteIdleSt) -> True
+      (Just RemoteColdSt, Just RemoteColdSt) -> True
```
coot, line 1738: Add a jira ticket for this bug. Describe there why these identity transitions are needed, link back to this thread and suggest that we need to update the specification.

ouroboros-network/test/Test/Ouroboros/Network/Diffusion/Node.hs
```diff
@@ -201,7 +200,7 @@ run blockGeneratorArgs limits ni na =
                      muxErrorRethrowPolicy
                   <> ioErrorRethrowPolicy
                   <> mkRethrowPolicy
-                      ( \ _ (_ :: CloseError) -> ShutdownNode)
+                      ( \ _ (_ :: SomeException) -> ShutdownNode)
```
coot, line 203: This should be removed rather than to use `SomeException`.  Even in testing we don't want to shutdown the node if some exception happens (only very specific exceptions).

Note that such a change could be disastrous if committed to the node itself. Why? Because one could try to trigger some erroneous condition in a thread that would result in stopping the whole node.  If one would find a way to propagate this through network, whole system could stop.

coot (changes requested):

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -681,43 +693,39 @@ withConnectionManager ConnectionManagerArguments {
...
+                          case Map.lookup peerAddr state of
+                            Nothing -> (state                   , Nothing)
+                            Just v  ->
                               if eqTVar (Proxy :: Proxy m) connVar v
-                                then Nothing
-                                else Just v
-                            )
-                            peerAddr
+                                then (Map.delete peerAddr state , Just Nothing)
+                                else (state                     , Just (Just v))
```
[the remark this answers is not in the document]
bolt12: Good point!

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -821,6 +821,12 @@ data ConnectionManagerTrace peerAddr handlerTrace
   | TrConnectionManagerCounters  !ConnectionManagerCounters
   | TrState                      !(Map peerAddr AbstractState)
   -- ^ traced on SIGUSR1 signal, installed in 'runDataDiffusion'
+  | TrUnexpectedlyMissingConnectionState !(ConnectionId peerAddr)
+  -- ^ This is triggered by a failed lookup on the 'ConnectionManagerState'
+  -- using the 'ConnectionId'\'s peerAddr, during connection handler cleanup
+  -- function. This shouldn't really be possible to achieve and its use is
+  -- mostly for us to be able to be total in our case expression over at
+  -- connection handler cleanup.
```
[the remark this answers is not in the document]
bolt12: We added the trace because we needed to pattern match on a value that is theoretically impossible to reach (theoretically because it is not enforced by the code). Since the case is a nonsense to reach and we want to be total in the pattern matching the best thing to do is to trace that we somehow reached this unexpectedly.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
bolt12: Exactly so testing found two problems to be related to weird race conditions and the cause was a wrong assumption that the `connVar` could be reused if in the `Terminating` or `TerminatedState`.

I am not creating jira tickets but GitHub issues and linking them to Jira, is that alright?

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -491,19 +504,28 @@ mkSockType FDConnecting {}    = ConnectionSock
...
+    = STConnecting   (FD_ m addr) addr
     | STConnected    (FD_ m addr) OpenType
     | STBearerInfo   BearerInfo
     | STConnectError (FD_ m addr) addr IOError
+    | STConnectTimeout TimeoutDetail
     | STBindError    (FD_ m addr) addr IOError
-    | STClosing      SockType Bool
-    | STClosed       SockType
+    | STClosing      SockType (Wedge (ConnectionId addr) [addr])
+    | STClosed       SockType (Maybe (Maybe OpenState))
```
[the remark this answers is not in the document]
bolt12: I'm not really sure what the meaning is, I think this change carried out from a rebase.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
bolt12: Basically this happens because in the code we are not making an effort to not log unspecified identity transitions. I do not see this as a bug, and only enabled these two transitions because you enabled `RemoteWarmSt -> RemoteWarmSt`. If you feel this shouldn't be allowed I can fix the code to ignore these transition cases.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Types.hs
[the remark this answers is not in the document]
coot: This is scheduled in #3321

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Types.hs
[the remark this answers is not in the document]
coot: It will be cleaner to define special type for that, than to use a tracer for that.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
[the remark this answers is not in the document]
coot: ok, then could you add a TODO.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
coot: > I am not creating jira tickets but GitHub issues and linking them to Jira, is that alright?

That's alright, maybe add some label so we can easily find them, e.g. `discovered-by-quickcheck` or something like that.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
coot: I added an issue for this, it's just a matter of specifying them explicitly in the documentation: #3402.

coot (changes requested):
Almost there, there are just two comments left to be addressed.

PR #3283, opened 2021-08-03
Added section about especially long suboptimal candidate

Jasagredo:


ouroboros-consensus/docs/report/chapters/future/genesis.tex
```diff
@@ -1655,6 +1655,61 @@ \subsection{Implementation}
...
+provided us the $C_1$ candidate). Our work on forging the block will be
+discarded.
+
+However, the act of creating this block made the distance between the
+intersection with the pending optimal chain and our current chain grow to $k +
+1$ long, so the ChainSync protocol will disconnect from the peer(s) that were
+announcing the candidate $C_0$ and we would then be able to progress adopting
+more blocks from $C_1$. Eventhough the work we made on forging will be
+discarded, it is a fair tradeoff because it allows us to unblock from the
+suboptimal ``dilemma''.
```
Jasagredo, line 1704: Probably `"deadlock"` is better than `"dilemma"`

Jasagredo:
I think this section is no longer relevant. May I close this? @nfrisby @amesgen

nfrisby:
As of the latest Genesis design, I don't think this concern is pertinent. Though it might be until we take the (generally agreed upon) action to simplify BlockFetch to fetch however many chains are being offered instead of budgeting out a threshold on how many different chains it'll request.

PR #3285, opened 2021-08-04
Update to latest cardano-ledger-specs.

nc6:


cabal.project
```diff
@@ -234,8 +234,8 @@ source-repository-package
 source-repository-package
   type: git
   location: https://github.com/input-output-hk/plutus
-  tag: 826c2514a40e962c2e4d56ce912803a434cc28fe
-  --sha256: 04jryrgi2wggkf4a994smbcp8j39clshbzfnvfvvk7lq0sm8kssz
+  tag: 985bd37845cd0996fc38eaccd1db1a9ad3edab3a
```
nfrisby, line 237: 3091 of `cls` uses `95e5d7488451e43062ca84d5376b3adcc465f1cd` for `plutus` -- should we match that here?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Eras.hs
```diff
@@ -253,7 +253,7 @@ pattern IncorrectClaimedFlag claimedFlag <-
           (Alonzo.UtxosFailure
             (Alonzo.ValidationTagMismatch
               (Alonzo.IsValid claimedFlag
-      )))))
+      ) _))))
```
nfrisby, line 256: Minor: this feels easy to overlook to me. A couple ideas:

- rename it `_<foo>` if only so it's easier to notice

- change the layout (perhaps of the whole pattern) so the `_` is on its own line

nfrisby (commented):
I wrote a couple minor comments, please consider those before merging.

Also: I just want to confirm before Approving: we're changing the serialization of all Shelley-based ledger states?

cabal.project
nc6: Erm... it uses the same version: https://github.com/input-output-hk/cardano-ledger-specs/blob/309178369eebe9ce02171987f6d23c7778f7db26/cabal.project#L72

nfrisby: Aha! I glanced too quickly. `95e5d7488451e43062ca84d5376b3adcc465f1cd` is for `https://github.com/Quid2/flat.git`, not for `plutus`.

nc6:
Yes, serialisation is changed. I think there's nothing we can do about this, we require additional information in the pulsing calculation. I'm hoping to bundle this together with other changes we make to the serialisation.

nfrisby (approved):
Approved. Thanks much.

nfrisby:
I rebased onto now's `master`; it was automatic with no conflicts.

PR #3294, opened 2021-08-11
Simulated P2P Diffusion

### LINKED ISSUE #3291: Ephermeral ports in diffusion simulations
Currently we use an `Ord` instance with a maximal address for ephemeral
addresses (all addresses smaller than the bound are ephemeral).  This does not
work for `(IP, PortNumber)` (which we would like to use in diffusion
simulations).  We need a better abstraction for ephermeral addresses.

Ephermeral addresses in network simulation correspond to OS ephemeral ports.
The simulation has no notion of a port number, hence the name ephemeral
address.

coot:
Changes which enable running p2p diffusion in `io-sim`.

- handshake: more convenient type of unversionedProtocolDataCodec
- RootPeersDNS test: use io-sim-classes rather
- ouroboros-network: code style
- chain-sync: controlledClient
- ConcreteBlock.Block - use ByteString as a payload
- simulated P2P diffusion
- net-sim: sim-net: provide GlobalAddressScheme type class

Fixes #3291


bolt12 (approved):
LGTM! Very nice! It seems CI had some issues (some tests did not pass)

coot:
@bolt12 I pushed one more commit, so I re-requested your review.

bolt12 (approved):
👍🏼

PR #3295, opened 2021-08-12
consensus-cardano: enable the Alonzo era by default

nfrisby:
A node built against this commit will no longer need to set `TestEnableDevelopmentHardForkEras` in its configuration file in order to enable Alonzo for Cardano.


nfrisby:
@coot @karknu We're ready to do this, yeah? There's nothing in the Network Layer that is unblocked by this version bump that shouldn't be running on `mainnet`?

coot (approved):

dnadales (approved):

PR #3298, opened 2021-08-16
Check git dependencies

newhoggy:
This helps flag any dependencies who's hash is not on `master`.

### DIFF
--- a/.github/bin/check-git-dependencies
+++ b/.github/bin/check-git-dependencies
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+
+set -e
+
+case "$(uname -s)" in
+  Darwin) GREP=ggrep  ;;
+  *)      GREP=grep   ;;
+esac
+
+
+RED='\033[0;31m'
+YELLOW='\033[0;33m'
+NC='\033[0m' # No Color
+
+mkdir -p tmp
+
+rm -rf tmp/dep-repo-result
+
+# Convert source-repository-package stanzas to lines 'location: $url tag: $hash'
+cat cabal.project \
+    | "$GREP" -v -E -e '^ *--' \
+    | "$GREP" -E -w -e '^source-repository-package' -e '^ *location:' -e '^ *tag:' \
+    | tr '\n\r' ' ' \
+    | sed 's/source-repository-package/\n/g' \
+    > tmp/repositories.txt
+
+basedir="$(pwd)"
+
+# Check each repository
+cat tmp/repositories.txt | while read -r line; do
+  # re-read the line into an array of words
+  read -r -a words <<< "$line"
+
+  # skip whitespace-only lines
+  [ "${#words[@]}" -eq 0 ] && continue
+
+  # parse the line as "location: $url tag: $hash"
+  [ "${#words[@]}" -eq 4 ] && [ "${words[0]}" = "location:" ] && [ "${words[2]}" = "tag:" ] \
+      || { echo "Could not parse as '^ *location: +\$url +tag: +\$hash *$': $line"
+	   false
+         }
+  url="${words[1]}"
+  hash="${words[3]}"
+
+  rm -f tmp/tmp-dep-repo-result
+  rm -rf tmp/dep-repo
+  echo "Checking $url"
+
+  # clone, but don't download any files (ie no trees, so no blobs either)
+  git clone --filter=tree:0 --no-checkout "$url" tmp/dep-repo
+  ( cd tmp/dep-repo
+
+    { git branch -r --contains "$hash"
+      git tag --contains "$hash"
+    } \
+        | "$GREP" -q -E -w \
+               -e '^ *origin/master' \
+               -e '^ *origin/develop' \
+               -e '^ *origin/main' \
+               -e '^ *origin/[^ ]*release' \
+               -e '*v[0-9]+' \
+        || { echo "Commit $hash from $url is not on the main branch or any release branches" \
+                 | tee -a "$basedir/tmp/dep-repo-result"
+             false
+           }
+  )
+
+  rm -rf tmp/dep-repo
+done
+
+if [ -s tmp/dep-repo-result ]; then
+  printf "${RED}Commits not on the main branch or any release branches detected in dependencies${NC}\n"
+  printf "${YELLOW}"
+  cat tmp/dep-repo-result
+  printf "${NC}"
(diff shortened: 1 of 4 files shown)


newhoggy:
For example: https://github.com/input-output-hk/cardano-node/pull/3083

nfrisby:
We have this check already, via `buildkite`. But IIUC we're retiring `buildkite`. So thanks for opening this!

nfrisby:
That `check-stylish` CI failure is suspicious for two reasons

- This PR doesn't touch Haskell code.
- It looks like Nix is failing to enter the shell and/or build `stylish-haskell`.

nc6:
The utility of this check is questionable - Plutus at least is moving towards a model of having release branches, which we will rely on downstream.

newhoggy:
This PR has been updated to accomodate release branches.  A branch is considered a `release` branch if it contains the word `release`.

newhoggy:
Also modified this to use recognise version tags that contain `v[9-9]+`.

Also forked the `flat` repo into IOHK Github org with a version tag, which I think is good practise.

nfrisby:
Hi John. I like the idea, and I think the high-level logic you wrote is a good place to start.

The algorithm seemed a bit round-about to me, but I didn't know enough to give a simple suggestion. So I just worked out an alternative, see here https://github.com/input-output-hk/ouroboros-network/tree/nfrisby/check-git-dependencies

Please consider integrating some of these changes I made into your commit on this PR:

- I renamed the variables to `$url` and `$hash` instead of `$location` and `$tag`, because calling a git commit hash (which these tings often are) a git tag is confusing, especially since we're also checking actual git tags. After digging a bit into the `cabal-install` source code, it looks like the `tag: ___` field in a `source-repository-package` is named that because of either GNU Bazaar           https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/cabal-install/src/Distribution/Client/VCS.hs#L259 or Darcs https://github.com/haskell/cabal/blob/00a2351789a460700a2567eb5ecc42cca0af913f/cabal-install/src/Distribution/Client/VCS.hs#L304 I'm bummed that this uses `tag` differently than what `git` uses it for 😞 especially since in this project we only ever populate the stanza's `tag:` with a commit hash.

- Only download the commit objects, no files etc. https://github.com/input-output-hk/ouroboros-network/blob/d3c68aadca731c33530773d12d00f26e3398eeaf/.github/bin/check-git-dependencies#L43-L44

- Use `git branch` and `git tag` to query which (remote) branches/tags include the commit, instead of listing all commits on each branch/tag and checking if the commit occurs as one. https://github.com/input-output-hk/ouroboros-network/blob/d3c68aadca731c33530773d12d00f26e3398eeaf/.github/bin/check-git-dependencies#L47-L53

- Use bash to parse the `source-repository-package` stanzas. Your `yq` trick was nice once I understood it, but I anticipated two problems:

  - If the `cabal.project` has an unexpected shape, the `yq` error might be unhelpful. (This is minor, since such a `cabal.project` probably wouldn't work elsewhere.)
  
  - `yq` was failing for me, when I ran the script locally. Googling for `yq` gives two main results: a wrapper around `jq` or separate tool implemented in Go. GitHub's `ubuntu-latest` uses the separate tool in Go, so I tried to add `yq-go` to my `shell.nix` to get it. Unfortunately, the nix package is v3, whereas `ubuntu-latest` is LTS 20.04 which uses v4, so the command still failed for me locally, since v3->v4 was a bit interface change for `yq-go`.
      - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md
      - https://mikefarah.gitbook.io/yq/v/v4.x/#v3-users
  
  - At that point, I figured a pure `bash` solution would be simple enough to avoid this kind of future confusion, so I wrote https://github.com/input-output-hk/ouroboros-network/blob/d3c68aadca731c33530773d12d00f26e3398eeaf/.github/bin/check-git-dependencies#L13-L37

nfrisby (dismissed):
I'm Requesting Changes mainly because I think it should be trivial to run the script locally. I use `nix-shell` for that kind of thing.

See my previous top-level PR comment for more details.

newhoggy:
Your version is quite nice.  I'm happy to go with what you have except that the `grep` needs special handling on a Mac.

This is because Macs use the BSD version of `grep` which behaves differently.  For example:

```
grep -q -E -w -e '^ *origin/master' -e '^ *origin/[^ ]*release' -e '*v[0-9]+'
```

produces this error:

```
grep: repetition-operator operand invalid
```

On Macs, the GNU version of `grep` is called `ggrep`, so we either need to use only flags that are common to both versions of `grep` or define a `GREP` variable that points to `grep` or `ggrep` depending on OS.

nfrisby:
> except that the `grep` needs special handling on a Mac.

Bah. This is why I know too much `bash` for my own good, but not enough :P

Would you resolve that?

kk-hainq:
I'd like to add that a few repositories use `develop` as the "main" branch. For example:
- https://github.com/input-output-hk/cardano-crypto lists `develop` as the default branch.
- Plutus specifically tracks the +6 commits `develop` branch of https://github.com/input-output-hk/iohk-monitoring-framework too.

nfrisby:
The leading `*` in `'*v[0-9]+'` looks like a typo (by me) to me. Maybe the BSD grep is happy if we fix that, since the error message mentions the repetition operator (ie `*`). ... Should it be `.*v[0-9]+'`?

And add a `-e '^ *origin/develop' \` line for @kk-hainq 's observation.

Then we're good to go?

This PR seems like a nice-to-have for us. As long as `cardano-node` does a comprehensive check, ours doesn't ultimately matter, yeah?

tmp/repositories.json
```diff
@@ -0,0 +1,38 @@
+[
```
nfrisby, line 1: Your `.gitignore` diff suggests this file should not be added.

nfrisby (changes requested):
Looking good. I Requested Changes because of the unwanted `tmp` file and some of the commits need to be squashed.

tmp/repositories.json
newhoggy: I have removed the `tmp` file and squashed the commits.

nfrisby (approved):
LGTM. Thanks John!

Please rebase and push before merging.

newhoggy:
bors merge

PR #3299, opened 2021-08-17
Monitor acceptLoopDo

karknu:
Previusly exceptions thrown by acceptLoopDo was ignored. Change it so
that if either accept loop or main loop throws an exception they are
both killed and the exception rethrown.


dcoutts (approved):
LGTM

PR #3300, opened 2021-08-17
Update to latest haskell.nix

EncodePanda:
This also restores ability to enter nix shell be Mac users. In a
nutshell when entering a shell one would get an error stating that
PlutusCore.Evaluation.Result module does not exists. As
it turns out Mac is (most likely) case insensitive system. In
plutus-core they've added to gitignore 'results*'.

Change 20297c67255b19edf23cc2f9507981f3459d466f in haskell.nix fixed
it. This commit updated haskell.nix to the latest version that also
includes 20297c67255b19edf23cc2f9507981f3459d466f

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -5,10 +5,10 @@
         "homepage": "https://input-output-hk.github.io/haskell.nix",
         "owner": "input-output-hk",
         "repo": "haskell.nix",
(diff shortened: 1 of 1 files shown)


nfrisby (approved):
😵‍💫  oof

PR #3301, opened 2021-08-18
Update to latest Plutus.

nc6:
This change entails some serialisation changes:

- Plutus no longer restricts the bytestring inside Data to 64 bytes, so
  our tests and CDDL have been adapted accordingly.
- The Plutus Data Constructor 102 is now serialized as
	`#6.102([uint, [* a]])` instead of `#6.102([uint, * a])`.
- The serialized size of the Plutus cost model was problematically
  large. We no longer transmit the keys, only the values (and assume
  they are sent in ascending order by key name).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,9 +8,3 @@ steps:
     command: 'nix-shell --run scripts/buildkite/check-nixpkgs-fmt.sh'
     agents:
       system: x86_64-linux
-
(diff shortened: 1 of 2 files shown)


nc6:
The check "commits must be on master" is unhelpful. The Plutus commit referenced is on an Alonzo release branch, and will stay there.

nfrisby (approved):
CI will check anything I could think of here. Thanks, Clarke!

Approving

PR #3305, opened 2021-08-20
Handle exceptions thrown by accept

karknu:
Exceptions thrown by accept can either trigger an exit or be ignored.
When the exception is ignored the acceptLoop will resume waiting for new
connections.


bolt12 (approved):
LGTM

dcoutts (commented):
LGTM

PR #3312, opened 2021-09-02

nc6:
This PR pulls in changes to the stake aggregation computation to allow it to be computed incrementally. This will result in the ledger state needing to be rebuilt.


dnadales (approved):

nc6:
bors merge

EncodePanda (approved):
I think I'm too late because bors is already running. I was just about to add that those two commits should have been squashed before merging, because we want to limit number of failing commits. Other then that LGTM

nc6:
I can squash

nc6:
bors merge

PR #3313, opened 2021-09-02
New GetCurrentBlockNo query

newhoggy:


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -195,6 +224,9 @@ answerQuery ::
 answerQuery cfg query st = case query of
   BlockQuery blockQuery -> answerBlockQuery cfg blockQuery st
   GetSystemStart -> getSystemStart (topLevelConfigBlock (getExtLedgerCfg cfg))
+  GetCurrentBlockNo -> case headerStateTip (headerState st) of
```
nfrisby, line 227: `0` isn't necessarily the right answer here. (I wish it were.) It's probably fine for all blocks we have.

But the more general/future-proof thing would be for the query to return `WithOrigin BlockNo` and have this "what is the block no of `Origin`?" question deferred to the caller -- ie to the logic downstream of the Consensus layer.

nfrisby: Yep, `headerStateTip` is the right function to use.

The `st` here is the result of the latest `Acquire` call in the `LocalStateQuery` communication. If the `Acquire` request provided `Just p`, then the block no is the block no of that requested point. If the `Acquire` left it as `Maybe`, then your subsequent `GetCurrentBlockNo` will be the block no of the tip as of the `Acquire` call.

Therefore, I think `GetAcquiredBlockNo` might be a better name. Although maybe the fact that `Current` means "what I last `Acquire`d" is already obvious-enough from the `LocalStateQuery` context?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -65,12 +71,19 @@ data Query blk result where
   -- Supported by 'QueryVersion' >= 'QueryVersion1'.
   GetSystemStart :: Query blk SystemStart
 
+  -- | Get the 'GetCurrentBlockNo' time.
+  --
+  -- Supported by 'QueryVersion' >= 'QueryVersionX'. -- TODO jky what version?
+  GetCurrentBlockNo :: Query blk BlockNo
+
+
```
nfrisby, line 79: Extra newline.

nfrisby (commented):
I'm only not Approving just because of the TODO about "which version", and that'll take some communications, so I haven't done it yet. But I think you're very much on the right path.

nfrisby:
I just noticed there are a few "competing" PRs that do similar things. If I understand correctly, PR 3346 is the active one. Can we close the rest of these? Thanks.

PR #3314, opened 2021-09-02
Accept handing

karknu:
Differentiate between temporary errors (`ECONNABORTED`) and exceptions that signal resource exhaustion or internal bugs. Only temporary errors should be ignored.

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE DataKinds           #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE GADTs               #-}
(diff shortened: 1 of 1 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
```diff
@@ -180,7 +194,30 @@ run ServerArguments {
...
+                      -- On Windows the network packet classifies all errors
+                      -- as OtherError. This means that we're forced to match
+                      -- on the error string. The text string comes from
+                      -- the network package's winSockErr.c, and if it ever
+                      -- changes we must update our text string too.
+                      if ioeGetErrorString ioErr /=
+                          "Software caused connection abort (WSAECONNABORTED)"
+                         then throwIO ioErr
+                         else threadDelay 0.5 >>
+                             acceptLoop localAddress acceptNext
```
coot, line 214: Let's define `iseCONNABORTED` for `mingw32_HOST_OS`.  This way we won't have to use `CPP`.

coot (changes requested):

ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
karknu: I'll define `iseCONNABORTED` for windows but it we still need CPP. It just means that the string comparison will get moved to that function.

coot: Yes, we will still need CPP, but it will be just in one place.

coot (approved):
LGTM

bolt12 (approved):
LGTM

PR #3316, opened 2021-09-02
Some comments on Diffs

TimSheard:
1) Change the order of (<>) for Diff
2) Use strict left folds
3) Think of (Map k (Diff v)) as a logical extension to (Diff v)
4) Is  (Diff v) a functor? Does it meet the functor laws, or meet some other similar laws, or not at all?

### DIFF
--- a/ledger-ondisk/src/LedgerOnDisk/Diff.hs
+++ b/ledger-ondisk/src/LedgerOnDisk/Diff.hs
@@ -38,6 +38,13 @@ data Diff v where
 deriving stock instance Eq v => Eq (Diff v)
 deriving stock instance Show v => Show (Diff v)
 
+{-  (<>) on Diff is not commutative, so I think we should change the order of arguments here for 2 reasons
+ 1) I like to think of (d1 <> d2) as composition of changes. So in accordance with Haskell's composition
+    we should think of addpying the d2 first.  E.g. in Haskell (f . g) x == f (g x). So apply (x <> y) x == apply x (apply y x)
+ 2) When we lift (Diff v) to Map k (Diff v), we will want to use Map.insertWith (<>) k v x, and Map.unionWith (<>) m1 m2
+    and these require to us apply the change on the right first.
+So it is best to change the order.
+
 instance Semigroup (Diff v) where
   x <> DNoChange = x
   DMappend x <> DMappend y = DMappend $ x <> y
@@ -45,22 +52,58 @@ instance Semigroup (Diff v) where
   DRemove <> DMappend y = DChangeTo y
   _ <> y = y
 
+Another thing, is that DRemove (applied first) cancels a subsequent (DMappend v) so the 4th line
+when appying things in the eft first should be
+   DRemove <> DMappend y = DRemove.
+We can also make the function clearer by making it more verbose, so it is easy to see all the cases are covered.
+So I suggest this.
+-}
+
+-- (<>) is composition of changes. So apply (x <> y) x == apply x (apply y x)
+-- Note we apply the change on the right first.
+instance Semigroup v => Semigroup (Diff v) where
+  DNoChange <> x          = x
+  DRemove <> x            = DRemove
+  DChangeTo v <> x        = DChangeTo v
+  DMappend v <> DNoChange = DMappend v
+  DMappend v <> DRemove   = DRemove
+  DMappend v <> DChangeTo u = DChangeTo (v <> u)
+  DMappend v <> DMappend u = DMappend (v <> u)  
+
 instance Monoid (Diff v) where
   mempty = DNoChange
 
+
+-- Not sure this is a very useful function, it is just the 'fmap' method of functor.
 mapD :: Semigroup b => (a -> b) -> Diff a -> Diff b
 mapD f = \case
   DChangeTo x -> DChangeTo $ f x
-  DMappend x -> DChangeTo $ f x
+  -- DMappend x -> DChangeTo $ f x  -- I am also worried about this clause, but it is hard to articulate why.
+  --  mapD f (Mappend x) <> mapD f (Mappend y) ---> DChangeTo (f x) <> DChangeTo (f y)
+  --  I think I would like this to be Mappend (f(x<>y)), but that is not the case. Of course my intuition might be wrong.
+  --  Do you get the same feeing that something is fishy here?
   DRemove -> DRemove
   DNoChange -> DNoChange
 
+{- WHY is the x a maybe type? Every (Diff v) can be appied to any 'v'
 applyD :: Semigroup v => Maybe v -> Diff v -> Maybe v
 applyD x = \case
   DChangeTo v -> Just v
   DMappend v -> Just $ maybe v (<> v) x
   DRemove -> Nothing
   DNoChange -> x
+-}
+
+applyD :: Semigroup v => v -> Diff v -> Maybe v
+applyD x = \case
+  DChangeTo v -> Just v
+  DMappend v -> Just (v <> x)
+  DRemove -> Nothing
+  DNoChange -> Just x
+
+-- Not sure applyDforK is a usefull function. Once we lift (Diff v) to (Map k (Diff v)), we think of
+-- applying the whole d::(Map k (Diff v)) to m::(Map k v). The idea is that any keys in 'm' not in 'd'
+-- are not changed. applyDfork is really applyDtoHashMap with a singleton (Map k (Diff v))
 
 applyDforK :: (Eq k, Hashable k, Semigroup v) => k -> Diff v -> HashMap k v -> HashMap k v
 applyDforK k = \case
@@ -73,6 +116,8 @@ applyDtoHashMap :: (Semigroup v, Eq k, Hashable k) => HashMap k (Diff v) -> Hash
 applyDtoHashMap d = appEndo (HashMap.foldMapWithKey go d) where
   go k = Endo . applyDforK k
 
+
+{- We probably want to use a strict left fold, to avoid creating thunks. Maybe the use of Endo is overkill?
 applyDifftoMap :: Semigroup v => Ord k => Map k (Diff v) -> Map k v -> Map k v
 applyDifftoMap d = appEndo (Map.foldMapWithKey go d) where
   go k = Endo . \case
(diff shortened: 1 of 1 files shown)


ledger-ondisk/src/LedgerOnDisk/Diff.hs
```diff
@@ -38,29 +38,72 @@ data Diff v where
...
   DMappend x <> DMappend y = DMappend $ x <> y
   DChangeTo x <> DMappend y = DChangeTo $ x <> y
   DRemove <> DMappend y = DChangeTo y
   _ <> y = y
 
+Another thing, is that DRemove (applied first) cancels a subsequent (DMappend v) so the 4th line
+when appying things in the eft first should be
+   DRemove <> DMappend y = DRemove.
+We can also make the function clearer by making it more verbose, so it is easy to see all the cases are covered.
+So I suggest this.
```
duog, line 59: I think either of
```
DRemove <> DMappend _ = DRemove
```
or
```
DRremove <> DMappend x = DChangeTo x
```
are valid instances. Am I wrong? Otherwise can you elaborate on why you prefer the first?

ledger-ondisk/src/LedgerOnDisk/Diff.hs
```diff
@@ -38,29 +38,72 @@ data Diff v where
...
+  DMappend v <> DNoChange = DMappend v
+  DMappend v <> DRemove   = DRemove
+  DMappend v <> DChangeTo u = DChangeTo (v <> u)
+  DMappend v <> DMappend u = DMappend (v <> u)  
+
 instance Monoid (Diff v) where
   mempty = DNoChange
 
+
+-- Not sure this is a very useful function, it is just the 'fmap' method of functor.
```
duog, line 77: Agree it's not so useful. But it's the best version of fmap available , due to the DMappend constructor.
Undocumented precondition is that f is a Semigroup homomorphism.

duog (commented):
Sorry I've not finished replying, will finish later on.

Diff cannot be a functor becasue of the DMappend constructor.

ledger-ondisk/src/LedgerOnDisk/Diff.hs
```diff
@@ -38,29 +38,72 @@ data Diff v where
...
 mapD :: Semigroup b => (a -> b) -> Diff a -> Diff b
 mapD f = \case
   DChangeTo x -> DChangeTo $ f x
-  DMappend x -> DChangeTo $ f x
+  -- DMappend x -> DChangeTo $ f x  -- I am also worried about this clause, but it is hard to articulate why.
+  --  mapD f (Mappend x) <> mapD f (Mappend y) ---> DChangeTo (f x) <> DChangeTo (f y)
+  --  I think I would like this to be Mappend (f(x<>y)), but that is not the case. Of course my intuition might be wrong.
+  --  Do you get the same feeing that something is fishy here?
```
duog, line 84: I completely agree with you.

ledger-ondisk/src/LedgerOnDisk/Diff.hs
```diff
@@ -73,13 +116,34 @@ applyDtoHashMap :: (Semigroup v, Eq k, Hashable k) => HashMap k (Diff v) -> Hash
...
+
+-- So if (Diff v) represent changesto 'v'
+-- then (Map k (Diff v)) represents changes to (Map k v)
+-- Just like we want to compose changes to 'v', we want to compose changes to (Map k v)
+-- We have (<>) :: Diff v -> Diff v -> Diff v
+-- So we want (<>) :: Map k (Diff v) -> Map k (Diff v) -> Map k (Diff v)
+-- We would like to write this
+-- instance (Ord k,Semigroup v) => Semigroup (Map k (Diff v)) where
+--    (<>) x y  = (Map.unionWith <> x y)
+-- But this is already instance, Because (Diff v) has a Semigroup instance!
```
duog, line 146: I would guess that using `Data.Strict.Map.foldMapWithKey` and `Endo` is roughly equivalent to `Data.Strict.Map.foldlWithKey'`, if you prefer the latter that's good enough for me.

You've also implemented the change to `DRemove <> DMappend _` here.

Not sure if I'm misunderstanding you, but the `Semigroup` instance for `Map k v` does not use `v`'s `Semigroup` instance. It uses `union`, which uses the left value if both values are present.

If you look at `DiffMap` you can see it's `Semigroup` instance uses `MonoidalMap` (from https://hackage.haskell.org/package/monoidal-containers).

duog: Actually, for some reason it isn't :/ but, the Semigroup instance for Diffmap should (IMO) be written
```
deriving via (MonoidalMap k (Diff v)) instance Ord k => Semigroup (DiffMap k v)
```

ledger-ondisk/src/LedgerOnDisk/Diff.hs
TimSheard: You don't really see the difference till you lift the Diff to Maps.
I want  (apply (d1 <> d2) m)  to be  apply d1 (apply d2 m)
so at the map level suppose d1 was { 'a' -> Dremove} and d2 was {'a' -> DMappend 4}  and m = {'a' -> 3, 'b' -> 5}
So     apply d1 (apply d2 m)  goes in two steps {'a' -> 3, 'b' -> 5}  apply d2 {'a' -> 7, 'b' -> 5} apply d2  {'b' -> 5}

Now  under the second interpretation of <> for Diff, we have at the map level  {'a' -> Dremove} <> {'a' -> Mppend 4} goes to
{'a' -> DChangeTo '4'}   and now       (apply (d1 <> d2) m)  goes to   apply {'a' -> DChangeTo '4'} {'a' -> 3, 'b' -> 5} goes to
{'a' -> 7, 'b' -> 5},  under the first interpretation we get {'b' -> 5}

So the law     (apply (d1 <> d2) m)  == apply d1 (apply d2 m)  No longer holds. In the end, we deveop all his to work at the map level, so the first interpretation is what we want.

ledger-ondisk/src/LedgerOnDisk/Diff.hs
TimSheard: I want above all  that he law    (apply (d1 <> d2) m) == apply d1 (apply d2 m)      holds.   And I want there to be nice properties for fmap as well.   If fmap is the function part of the Functor, then the identity and composition laws hold. But they seem to be at odds with the apply laws in some strange way.  So I ask, do we ever use this function?   If not lets remove it. Its type suggests it s a functor, but it doesn't really behave as one.  If we do use it, lets be sure we are using it the correctly.

This reaches back to the comment you made about the use of functions in my Message type, and my reply that we could make first order representaitons of functions.  So lets see how that works out. Suppose we had the choice of using either

data Diff v =
  DMappend v | DFun (v -> v)

The we could represent (DMappend 3) as (DFun (3 <>))  and we would get the same answer.   We could also represent DNoChange  with (DFun id),  So this looks promissing, But we can't make a functor instance of a type with DFun here as well. given

data Diff v = DChangeTo v | DFun (v -> v)

fmap f (DChangeTo v) = DChangeTo (f v)
fmap f (DFun g) = DFun( f . g . h)
   where h is the inverse of f,  but how do we get that?

So a Functor instance seems impossible.  One last thought. Suppose f and h form an isomorphism, then we can write something approaching a Functor instance

isomap :: (Iso a b) -> Diff a -> Diff b
isomap (Iso f h)  DChangeTo (f v)
isomap (Iso f h) (DFun g) = DFun( f . g . h)

ledger-ondisk/src/LedgerOnDisk/Diff.hs
TimSheard: What I am complaining about is that we get overlapping instances. My instance is the one I want, the one supplied by haskell does not have the prpperties I want:   apply (d1 <> d2) m) to be apply d1 (apply d2 m)

ledger-ondisk/src/LedgerOnDisk/Diff.hs
duog: I agree that it is very important that `(apply (d1 <> d2) m) == apply d1 (apply d2 m)` holds, up to I'm used to thinking of `d1 <> d2` applying d1 then d2. In this note I'm going to use `<>` for the semigroup operator that applies `d2` first and `++` for the semigroup operator that applies `d1` first to try to remove ambiguity.

I think there are two adjustments we're talking about:
* The change so that `<>` is the semigroup operator rather than the current `++`
* The change so that `DMappend x <> DRemove == DRemove` rather than the current `DMappend x <> DRemove == DChangeTo x`

Let's asssume that we have applied change 1 but not change 2. Now, setting 
```haskell
d1 = {'a' -> DRemove}
d2 = {'a' -> DMappend 4}
m = {'a' -> 3, 'b' -> 5}
```
we have
```
d1 <> d2 == {'a' -> DRemove }
apply d2 m == {'a' -> 7, 'b' -> 5 }
apply d1 (apply d2 m) == {'b' -> 5 }
apply (d1 <> d2) == {'b' -> 5}
```
So this is not a counteraxample.

If we try
````haskell
d2 <> d1 == { 'a' -> DChangeTo 4}
apply d1 m == {'b' -> 5 }
apply d2 (apply d1 m) == {'a' -> 4, 'b' -> 5}
apply (d2 <> d1) m == {'a' -> 4, 'b' -> 5}
```
So the law holds here too. I think either interpretation of `DMappend x <> DRemove` satisfies our law.

ledger-ondisk/src/LedgerOnDisk/Diff.hs
duog: Yeah it's an annoying aspect of `containers`. We need to use a newtype for the Map for this reason
```haskell
newtype DiffMap k v = DiffMap (Map k (Diff v))
  deriving stock (Show, Eq)
  deriving (Semigroup, Monoid) via MonoidalMap k (Diff v)
```

ledger-ondisk/src/LedgerOnDisk/Diff.hs
duog: I'm not attached to this function at all, but I do think it's fine if `f` is a semigroup homomorphism (i.e. `f (x <> y) == f x <> f y`). I think your `isomap` is also fine.

Jasagredo:
I think this can be closed? It will remain in the list of closed PRs, but I think it is not actionable anymore? cc @TimSheard

dnadales:
@TimSheard, I'm closing this lest we'll forget and leave this hanging another year. Feel free to re-open.

PR #3317, opened 2021-09-03
Update Win32 versions shipped with ghc to account for 8.10.6+

jbgi:
cf. https://www.snoyman.com/base/

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -326,13 +326,16 @@ library
 
 
   if os(windows)
-    if impl(ghc >= 8.10.2) && impl(ghc < 8.11)
-      -- Fix the Win32 version to be whatever is shipped with GHC.  This is important
-      -- because we have other code that uses the GHC API which requires this specific
-      -- version of the Win32 or constraint solving will fail.
+    -- Fix the Win32 version to be whatever is shipped with GHC.  This is important
+    -- because we have other code that uses the GHC API which requires this specific
+    -- version of the Win32 or constraint solving will fail.
+    if impl(ghc < 8.10.6)
       build-depends:   Win32 == 2.6.1.0
     else
-      build-depends:   Win32
+      if impl(ghc < 8.11)
+        build-depends:   Win32 == 2.6.2.1
+      else
+        build-depends:   Win32 == 2.10.0.0
   else
      build-depends:    unix
                      , unix-bytestring


ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -326,13 +326,16 @@ library
...
+    -- because we have other code that uses the GHC API which requires this specific
+    -- version of the Win32 or constraint solving will fail.
+    if impl(ghc < 8.10.6)
       build-depends:   Win32 == 2.6.1.0
     else
-      build-depends:   Win32
+      if impl(ghc < 8.11)
+        build-depends:   Win32 == 2.6.2.1
+      else
+        build-depends:   Win32 == 2.10.0.0
```
coot, line 338: I think we can be more relaxed than that and specify `Win32 >= 2.6.2.1` regardless of `ghc` version, then the upstream dependency that needs to have exact version will be able to set it.

@nfrisby what do you think?

nfrisby: I don't think I have an opinion. It was @Jasagredo that had concerns here due to the HLS build. (He's on vacation this week.)

coot: I prefer to make bounds as relaxed as possible, this way the upstream dependency can decide what version it needs. But on the other side, we only test with whatever we choose and not what upstream might end up using - though this could introduce a subtle bug, it's rather unlikely.

Jasagredo: There are some problems with later versions of `Win32-network` in Nix, as mentioned in this thread https://github.com/input-output-hk/ouroboros-network/pull/3272#discussion_r678144101

If no problems are found on compilation then I'm fine with setting the bounds as relaxed as possible.

nfrisby:
Does this PR need any attention? (I still don't think I'm the best reviewer -- just pinging those involved.)

Jasagredo:
> Does this PR need any attention? (I still don't think I'm the best reviewer -- just pinging those involved.)

Ping again cc @coot @jbgi :smile:

coot:
Right now we use:
```
  if os(windows)
     build-depends:    Win32            >= 2.6.1.0
  else
     build-depends:    unix
                     , unix-bytestring
```
And it works for `ghc-8.10` and `ghc-9.2`; I think we can close this issue. @jbgi please re-open if you disagree.

PR #3319, opened 2021-09-04
New get header state tip query

newhoggy:
I was able to query the local state tip and observe that it matches the chain sync result.

```
CARDANO_NODE_SOCKET_PATH=example/node-bft1/node.sock  $(./scripts/bin-path.sh cardano-cli) query tip --testnet-magic 42 --cardano-mode
headerStateTip At HeaderStateTip {queriedTipSlotNo = SlotNo 57, queriedTipBlockNo = BlockNo 4, queriedTipInfo = e7013470b0d6f915005e27a735c5abc6ad99416801819c1c9ef72fdf29784bc7}
{
    "epoch": 0,
    "hash": "e7013470b0d6f915005e27a735c5abc6ad99416801819c1c9ef72fdf29784bc7",
    "slot": 57,
    "block": 4,
    "era": "Alonzo",
    "syncProgress": "100.00"
}
```

### DIFF
--- a/cardano-client/src/Cardano/Client/Subscription.hs
+++ b/cardano-client/src/Cardano/Client/Subscription.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE TypeFamilies        #-}
@@ -38,10 +39,10 @@ import           Ouroboros.Network.NodeToClient (ClientSubscriptionParams (..),
                      ConnectionId, LocalAddress,
                      NetworkClientSubcriptionTracers,
                      NodeToClientProtocols (..),
+                     NodeToClientVersion,
                      NodeToClientVersionData (NodeToClientVersionData),
                      ncSubscriptionWorker, newNetworkMutableState,
                      versionedNodeToClientProtocols)
-import           Ouroboros.Network.NodeToClient (NodeToClientVersion)
 import           Ouroboros.Network.Protocol.Handshake.Version (Versions,
                    foldMapVersions)
 import qualified Ouroboros.Network.Snocket as Snocket
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
@@ -81,7 +81,7 @@ data Query blk result where
 
   -- | Get the 'GetHeaderStateTip' time.
   --
-  -- Supported by 'QueryVersion' >= 'QueryVersionX'. -- TODO jky what version?
+  -- Supported by 'QueryVersion' >= 'QueryVersion2'.
   GetHeaderStateTip :: Query blk (WithOrigin (HeaderStateTip blk))
 
 
@@ -120,6 +120,7 @@ data QueryEncoderException blk =
 deriving instance Show (SomeSecond BlockQuery blk) => Show (QueryEncoderException blk)
 instance (Typeable blk, Show (SomeSecond BlockQuery blk)) => Exception (QueryEncoderException blk)
 
+
 queryEncodeNodeToClient ::
      forall blk.
      Typeable blk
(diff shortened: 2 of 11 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -231,7 +236,7 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk = L
 
 -- | Combination of 'runWith' and 'stdLowLevelRunArgsIO'
 run :: forall blk.
-     RunNode blk
+     (RunNode blk, Serialise (TipInfo blk))
```
newhoggy, line 239: I'd like to not need to introduce extra constraints fo functions like this.

cardano-client/src/Cardano/Client/Subscription.hs
```diff
@@ -54,8 +57,10 @@ import           Ouroboros.Consensus.Node.NetworkProtocolVersion
                      (BlockNodeToClientVersion, supportedNodeToClientVersions)
 import           Ouroboros.Consensus.Node.Run (RunNode)
 
+import           Ouroboros.Consensus.HeaderValidation (TipInfo (..))
+
 subscribe ::
-     RunNode blk
+     (RunNode blk, Serialise (TipInfo blk))
```
newhoggy, line 63: I'd like to not need to introduce extra constraints fo functions like this.

cardano-client/src/Cardano/Client/Subscription.hs
```diff
@@ -78,7 +83,7 @@ subscribe snocket codecConfig networkMagic tracers subscriptionParams protocols
             protocols version codecs connectionId))
 
 versionedProtocols ::
-     forall blk m appType bytes a b. (MonadST m, RunNode blk)
+     forall blk m appType bytes a b. (MonadST m, RunNode blk, Serialise (TipInfo blk))
```
newhoggy, line 86: I'd like to not need to introduce extra constraints fo functions like this.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -182,9 +214,22 @@ instance SameDepIndex (BlockQuery blk) => SameDepIndex (Query blk) where
...
+    = Just Refl
+  sameDepIndex GetHeaderStateTip _
+    = Nothing
 
 deriving instance Show (BlockQuery blk result) => Show (Query blk result)
 
+data HeaderStateTip blk = HeaderStateTip {
+      queriedTipSlotNo  :: !SlotNo
+    , queriedTipBlockNo :: !BlockNo
+    , queriedTipInfo    :: !(TipInfo blk)
```
nfrisby, line 227: Would your use case let you simply use `HeaderHash` instead of `TipInfo` here? You can use the `tipInfoHash` method to get it.

That way, you wouldn't need the `Serialise (TipInfo blk)` constraints anywhere. (I haven't double-checked, but I'm assuming `RunNode blk` implies `Serialise (HeaderHash blk)` or else something as effective.)

If you do need `TipInfo`, then would adding the serialization constraint to the superclasses of `HasAnnTip` suffice?

```
class ( StandardHash blk
      , Show     (TipInfo blk)
      , Eq       (TipInfo blk)
      , NoThunks (TipInfo blk)
      , Serialise (TipInfo blk)
      ) => HasAnnTip blk where
```

Or perhaps adding it instead to `RunNode`?

(I'll need to better consider the context and revisit where similar constraints are pooled up in order to review such a change.)

newhoggy: Your suggestion to use `tipInfoHash` works great!

newhoggy:
This PR obsoletes https://github.com/input-output-hk/ouroboros-network/pull/3313

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -65,12 +79,19 @@ data Query blk result where
   -- Supported by 'QueryVersion' >= 'QueryVersion1'.
   GetSystemStart :: Query blk SystemStart
 
+  -- | Get the 'GetHeaderStateTip' time.
+  --
+  -- Supported by 'QueryVersion' >= 'QueryVersionX'. -- TODO jky what version?
+  GetHeaderStateTip :: Query blk (WithOrigin (HeaderStateTip blk))
```
newhoggy, line 85: @DavidEichmann Do you know if I can use `QueryVersion1` or do I need to use `QueryVersion2`?

DavidEichmann: Hey @newhoggy. You are indeed augmenting the protocol with a new query, so the version must be bumped :+1: .

DavidEichmann: ... @newhoggy  I perhaps spoke too quickly. There is an exception. You don't need to bump the version number if the previous version number has not yet been released, so you should confirm that. Specifically, figure out what commit was the latest release of `ouroboros-network` (I assume this corresponded to the release of Alonzo). If `QueryVersion1` was added after that release, then you actually don't need to bump the version number.

DavidEichmann: @nfrisby Is there some procedure for easily knowing what the latest release commit hash is? I admit I'm not familiar with the release process.

nfrisby: @DavidEichmann I go to the "Releases" page of `cardano-node` to find the commit of the latest release, and look at the `ouroboros-network` commit hash listed in the corresponding `cabal.project` file. I think that's the source of truth.

PR #3323, opened 2021-09-09
update cardano-ledger-specs

JaredCorduan:
This bring two changes:

* The Alonzo `TxOut` type is now a sum type (to save memory).
* The MIR event now filters out the  extant reward accounts.


erikd (approved):

nc6:
The MIR event fix was already included.

erikd:
> The MIR event fix was already included.

This also includes the `TxOutCompactDH` constructor which `db-sync` needs

nc6:
bors merge

PR #3333, opened 2021-09-09
Add block size to CompletedBlockFetch trace event

karknu:
Block size has an impact on block delay, which is already traced in this
event. It should be tracked here.


bolt12 (approved):
LGTM

coot (approved):
LGTM

PR #3335, opened 2021-09-10
Enforce sleeping at least acceptedConnectionsDelay at hard limit

karknu:
Previous logic was switched which could cause the accept thread to sleep
too long.

Thanks to @gufmar for helping with tracking down this issue.


coot (approved):
LGTM!

PR #3346, opened 2021-09-13
New GetChainBlockNo and GetChainPoint queries

newhoggy:


ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -182,19 +259,33 @@ instance SameDepIndex (BlockQuery blk) => SameDepIndex (Query blk) where
...
   -> result
 answerQuery cfg query st = case query of
   BlockQuery blockQuery -> answerBlockQuery cfg blockQuery st
   GetSystemStart -> getSystemStart (topLevelConfigBlock (getExtLedgerCfg cfg))
+  GetChainBlockNo -> case headerStateTip (headerState st) of
+    Origin -> Origin
+    At hst -> At (annTipBlockNo hst)
+  GetChainPoint -> case headerStateTip (headerState st) of
+    Origin -> Point Origin
+    At hst -> Point (At (Point.Block (annTipSlotNo hst) (tipInfoHash (Proxy @blk) (annTipInfo hst))))
```
dcoutts, line 288: Isn't it as simple as:
```suggestion
  GetChainPoint -> headerStatePoint (headerState st)
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -182,19 +259,33 @@ instance SameDepIndex (BlockQuery blk) => SameDepIndex (Query blk) where
...
   => ExtLedgerCfg blk
   -> Query blk result
   -> ExtLedgerState blk
   -> result
 answerQuery cfg query st = case query of
   BlockQuery blockQuery -> answerBlockQuery cfg blockQuery st
   GetSystemStart -> getSystemStart (topLevelConfigBlock (getExtLedgerCfg cfg))
+  GetChainBlockNo -> case headerStateTip (headerState st) of
+    Origin -> Origin
+    At hst -> At (annTipBlockNo hst)
```
dcoutts, line 285: Can we make this a helper function, e.g. `headerStateBlock`? Put it next to the existing `headerStatePoint` function.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -1,10 +1,12 @@
+{-# LANGUAGE FlexibleContexts    #-}
```
dcoutts, line 1: There are no actual changes in this module, I think all the changes in this module can be dropped from the PR.

ouroboros-network/src/Ouroboros/Network/Block.hs
```diff
@@ -421,6 +421,12 @@ instance Serialise (HeaderHash block) => Serialise (Point block) where
   encode = encodePoint encode
   decode = decodePoint decode
 
+instance (Serialise (HeaderHash block), Typeable block) => ToCBOR (Point block) where
+  toCBOR = encodePoint encode
+
+instance (Serialise (HeaderHash block), Typeable block) => FromCBOR (Point block) where
+  fromCBOR = decodePoint decode
+
```
dcoutts, line 429: It's uncelar to me if this is the "Right Thing":tm:

ouroboros-network/src/Ouroboros/Network/Point.hs
```diff
@@ -27,5 +30,20 @@ data Block slot hash = Block
...
+
+instance (ToCBOR slot, ToCBOR hash) => ToCBOR (Block slot hash) where
+  toCBOR blk = mconcat
+    [ encodeListLen 2
+    , toCBOR (blockPointSlot blk)
+    , toCBOR (blockPointHash blk)
+    ]
+
+deriving instance (Serialise slot, Serialise hash) => Serialise (Block slot hash)
+
```
dcoutts, line 47: What is this used for?

dcoutts (dismissed):
So if we're adding `GetChainBlockNo` and `GetChainPoint` queries then we do not need one for the tip. So we can get rid of the patch that adds `GetHeaderStateTip` and we don't need the corresponding tip type that's added there.

I think we can also get rid of the dependency on the `serialise` package.

I think we ought to be able to shrink this PR somewhat. Check we really need all the new code we're adding.

I agree that the right way to structure this PR is as two patches, the first that only adds support for another protocol version, and then the second that adds the two new queries.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -22,6 +22,9 @@ module Ouroboros.Consensus.Ledger.Query (
   , queryEncodeNodeToClient
   ) where
 
+import           Cardano.Slotting.Block (BlockNo (..))
+import           Cardano.Slotting.Slot (WithOrigin (..))
+
```
Jasagredo, line 27: I would move these imports next to `Cardano.Binary`. I think usually it is like:
1. 3rd party libraries
2. Cardano-base
3. Ledger imports
4. Ouroboros-network 
5. Ouroboros-consensus

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -47,6 +51,12 @@ import           Ouroboros.Consensus.Node.Serialisation
 import           Ouroboros.Consensus.Util (ShowProxy (..), SomeSecond (..))
 import           Ouroboros.Consensus.Util.DepPair
 
+import           Ouroboros.Consensus.HeaderValidation (AnnTip (..),
+                     HasAnnTip (..), HeaderState (..))
+
+import           Ouroboros.Network.Block (HeaderHash, Point (..), StandardHash)
+import qualified Ouroboros.Network.Point as Point
```
Jasagredo, line 58: Same, move this up.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -80,9 +103,17 @@ instance Eq (SomeSecond BlockQuery blk) => Eq (SomeSecond Query blk) where
   SomeSecond GetSystemStart == SomeSecond GetSystemStart = True
   SomeSecond GetSystemStart == _                         = False
 
+  SomeSecond GetChainBlockNo == SomeSecond GetChainBlockNo = True
+  SomeSecond GetChainBlockNo == _                        = False
+
+  SomeSecond GetChainPoint == SomeSecond GetChainPoint = True
+  SomeSecond GetChainPoint == _                      = False
```
Jasagredo, line 110: Alignment
```suggestion
  SomeSecond GetChainBlockNo == _                          = False

  SomeSecond GetChainPoint == SomeSecond GetChainPoint = True
  SomeSecond GetChainPoint == _                        = False
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -146,14 +202,27 @@ queryDecodeNodeToClient ::
...
+    handleTopLevelQuery = do
         size <- decodeListLen
         tag  <- decodeWord8
         case (size, tag) of
-          (2, 0) -> decodeBlockQuery
-          (1, 1) -> return (SomeSecond GetSystemStart)
+          (2, 0) -> requireVersion "BlockQuery"        QueryVersion1   decodeBlockQuery
+          (1, 1) -> requireVersion "GetSystemStart"    QueryVersion1 $ return (SomeSecond GetSystemStart)
+          (1, 2) -> requireVersion "GetChainBlockNo"     QueryVersion2 $ return (SomeSecond GetChainBlockNo)
+          (1, 3) -> requireVersion "GetChainPoint"       QueryVersion2 $ return (SomeSecond GetChainPoint)
```
Jasagredo, line 216: I would also align this
```suggestion
          (2, 0) -> requireVersion "BlockQuery"          QueryVersion1   decodeBlockQuery
          (1, 1) -> requireVersion "GetSystemStart"      QueryVersion1 $ return (SomeSecond GetSystemStart)
          (1, 2) -> requireVersion "GetChainBlockNo"     QueryVersion2 $ return (SomeSecond GetChainBlockNo)
          (1, 3) -> requireVersion "GetChainPoint"       QueryVersion2 $ return (SomeSecond GetChainPoint)
```

Jasagredo (commented):
It looks good to me, though I concede with Duncan's comments. I just made some minor comments on style and alignment.

ouroboros-network/src/Ouroboros/Network/Point.hs
newhoggy: The `Block` type is a substructure of `Point`, so `Point` isn't serialisable without the above instances.

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
```diff
@@ -231,14 +231,29 @@ instance ( blockVersion ~ BlockNodeToClientVersion blk
       Query.TopLevelQueryDisabled ->
         arbitraryBlockQuery queryVersion
 
-      Query.QueryVersion1 ->
+      Query.QueryVersion1 -> genTopLevelQuery queryVersion
+      Query.QueryVersion2 -> genTopLevelQuery queryVersion
+    where
+      genTopLevelQuery queryVersion =
```
dnadales, line 237: This will change the distribution of what QC generates, won't it? So we'll go from a `15` to `1` to a `15` to `3` block query to "with version" generated data. Is this desired?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -80,9 +103,17 @@ instance Eq (SomeSecond BlockQuery blk) => Eq (SomeSecond Query blk) where
...
+  SomeSecond GetChainBlockNo == _                        = False
+
+  SomeSecond GetChainPoint == SomeSecond GetChainPoint = True
+  SomeSecond GetChainPoint == _                      = False
+
 instance Show (SomeSecond BlockQuery blk) => Show (SomeSecond Query blk) where
   show (SomeSecond (BlockQuery blockQueryA)) = "Query " ++ show (SomeSecond blockQueryA)
   show (SomeSecond GetSystemStart) = "Query GetSystemStart"
+  show (SomeSecond GetChainBlockNo) = "Query GetChainBlockNo"
+  show (SomeSecond GetChainPoint) = "Query GetChainPoint"
```
dnadales, line 116: @Jasagredo, does your previous remark implies that we would want to align the definitions here as well?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -162,16 +231,24 @@ queryDecodeNodeToClient codecConfig queryVersion blockVersion
         blockVersion
       return (SomeSecond (BlockQuery blockQuery))
 
-instance SerialiseResult blk (BlockQuery blk) => SerialiseResult blk (Query blk) where
+instance (SerialiseResult blk (BlockQuery blk), Serialise (HeaderHash blk), Typeable blk) => SerialiseResult blk (Query blk) where
```
dnadales, line 234: Would it make sense to [separate this into several lines](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md#L430)?

dnadales: Ugh, links to lines in Markdown files do not seem to work. I was referring to this

image

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Fixed.  Thanks!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Done!

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
newhoggy: Done!

ouroboros-network/src/Ouroboros/Network/Block.hs
newhoggy: What does the "Right Thing" look like?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Done!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Done!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Done!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Done!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Aligned!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Done!

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
newhoggy: What principle should be followed for this kind of decision making?

ouroboros-network/src/Ouroboros/Network/Point.hs
newhoggy: Not sure what it is, but I removed the `ToCBOR` and `FromCBOR` instances and it compiles fine now.  :/

newhoggy: I think it's because some compiler errors suggested I needed to add one when adding a `Serialise`, but that I eventually added an `Serialise` instance with an implementation that doesn't need it.

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
dnadales: Ideally, we'd have coverage testing which would safeguard that we still cover a sufficient percentage of the cases we consider important. I'll ping you on slack to discuss this further.

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
```diff
@@ -231,14 +231,29 @@ instance ( blockVersion ~ BlockNodeToClientVersion blk
...
         frequency
-          [ (15, arbitraryBlockQuery queryVersion)
-          , (1,  do blockV <- arbitrary
-                    return (WithVersion (queryVersion, blockV)
-                                        (SomeSecond GetSystemStart)))
+          [ queryFrequencyFor Query.TopLevelQueryDisabled 15 $
```
dnadales, line 239: One thing that I'd like to discuss in a call is why not just hardcoding the frequencies as it used to be

dnadales: So, in particular, I was wondering if it'd be possible to write something like:

```haskell
    case queryVersion of
      Query.TopLevelQueryDisabled -> arbitraryBlockQuery queryVersion
      Query.QueryVersion1         -> arbitraryBlockVersion1Query
      Query.QueryVersion2         -> arbitraryBlockVersion1Query

  where
    query item = do blockV <- arbitrary
                     return (WithVersion (queryVersion, blockV)
                                         (SomeSecond item)))
    arbitraryBlockVersion1Query =
      frequency
        [ (15, arbitraryBlockQuery queryVersion)
        , (1,  query GetSystemStart)))
        ]
    arbitraryBlockVersion2Query =
      oneof
        [ arbitraryBlockVersion1Query
        , frequency [ (1, query GetChainBlockNo)
                    , (1, query GetChainPoint)
                    ]
        ]
```

dnadales: Or, repeating code, but removing one level of indirection for the reader so that it is easier for them to understand what the frequency distributions are:

```haskell
    arbitraryBlockVersion2Query =
      frequency
        [ (15, arbitraryBlockQuery queryVersion)
        , (1,  query GetSystemStart)))
        , (1, query GetChainBlockNo)
        , (1, query GetChainPoint)
        ]
```

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
```diff
@@ -31,6 +31,7 @@ module Ouroboros.Consensus.HeaderValidation (
   , HeaderState (..)
   , castHeaderState
   , genesisHeaderState
+  , headerStateBlockNo
```
dnadales, line 34: Why is this needed? Maybe it's worth adding a comment in the function.

ouroboros-network/src/Ouroboros/Network/Block.hs
newhoggy: Setting aside whether it's the "Right Thing", I've tested and it works.  😁

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -182,19 +260,29 @@ instance SameDepIndex (BlockQuery blk) => SameDepIndex (Query blk) where
...
   sameDepIndex GetSystemStart _
     = Nothing
+  sameDepIndex GetChainBlockNo GetChainBlockNo
+    = Just Refl
+  sameDepIndex GetChainBlockNo _
+    = Nothing
+  sameDepIndex GetChainPoint GetChainPoint
+    = Just Refl
+  sameDepIndex GetChainPoint _
+    = Nothing
```
dnadales, line 270: Would this also work?

```haskell
  sameDepIndex (BlockQuery blockQueryA) (BlockQuery blockQueryB) = Just Refl
  sameDepIndex GetSystemStart                   GetSystemStart                 = Just Refl
  sameDepIndex GetChainBlockNo                GetChainBlockNo             = Just Refl
  sameDepIndex GetChainPoint                     GetChainPoint                   = Just Refl
  sameDepIndex _                                          _                                         = Nothing 
```
(I don't know what GH does with the alignment, sorry)

My guess is that it would, but we want to prevent a situation in which we add another constructor and it gets caught by the wildcard case.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -182,19 +260,29 @@ instance SameDepIndex (BlockQuery blk) => SameDepIndex (Query blk) where
...
+answerQuery :: forall blk result.
+     (QueryLedger blk, ConfigSupportsNode blk, HasAnnTip blk)
   => ExtLedgerCfg blk
   -> Query blk result
   -> ExtLedgerState blk
   -> result
 answerQuery cfg query st = case query of
   BlockQuery blockQuery -> answerBlockQuery cfg blockQuery st
   GetSystemStart -> getSystemStart (topLevelConfigBlock (getExtLedgerCfg cfg))
+  GetChainBlockNo -> headerStateBlockNo (headerState st)
```
dnadales, line 284: About the names, can you remind me why did we chose `GetChainBlockNo` instead of `GetBlockNo` and why did we choose `GetChainPoint` instead of `GetPoint`? 

I have a similar question regarding the `headerState` prefix in the name of the functions used here.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: I suspect being able to get a compile error is more important.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: @dcoutts suggested the chain prefix which I presume means the block number of the tip of the chain as it is currently rather than any other block.  This has precedence in `cardano-api` we had a Shelley specific query called `QueryChainPoint` and this PR enables us to replace that with an era-independent query of the same name.

The `headerState` prefix I had was simply saying where I got the data from, so provenance rather than description.

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
newhoggy: `headerStateBlockNo` is a convenience accessor much like `headerStatePoint` just below.

It was requested by @dcoutts to simplify the implementation of `answerQuery` https://github.com/input-output-hk/ouroboros-network/blob/0a341b18f414bf21e426bc0090dd432c0121aa97/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs#L284

Originally I had used a case expression so looking at it now the its `fmap` form, inlining the function into `answerQuery` is not so bad.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
dnadales: Yeah, definitely!

ouroboros-consensus/src/Ouroboros/Consensus/HeaderValidation.hs
dnadales: I see. Thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
dnadales: Ok, that makes sense. Thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -146,14 +198,27 @@ queryDecodeNodeToClient ::
...
-          (1, 1) -> return (SomeSecond GetSystemStart)
+          (2, 0) -> requireVersion "BlockQuery"           QueryVersion1   decodeBlockQuery
+          (1, 1) -> requireVersion "GetSystemStart"       QueryVersion1 $ return (SomeSecond GetSystemStart)
+          (1, 2) -> requireVersion "GetChainBlockNo"      QueryVersion2 $ return (SomeSecond GetChainBlockNo)
+          (1, 3) -> requireVersion "GetChainPoint"        QueryVersion2 $ return (SomeSecond GetChainPoint)
           _      -> fail $ "Query: invalid size and tag" <> show (size, tag)
-  where
+
+    requireVersion :: String -> QueryVersion -> Decoder s (SomeSecond Query blk) -> Decoder s (SomeSecond Query blk)
+    requireVersion name expectedVersion f =
```
nfrisby, line 216: I prefer if `name` were somehow derived from `f` (which would requires rejigging `f`).

Or instead, replace `name` with the `(size, tag)`. My concern is that human error will lead to `name` not matching `f`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -118,16 +146,40 @@ queryEncodeNodeToClient codecConfig queryVersion blockVersion (SomeSecond query)
...
+
+        GetChainPoint ->
+          requireVersion QueryVersion2 queryVersion $ mconcat
+            [ encodeListLen 1
+            , encodeWord8 3
+            ]
+
   where
+    requireVersion :: QueryVersion -> QueryVersion -> a -> a
+    requireVersion expectedVersion actualVersion a =
```
nfrisby, line 178: `queryVersion` is in scope here, so you can `let actualVersion = queryVersion` instead of repeatedly passing it as a parameter.

ouroboros-network/src/Ouroboros/Network/Block.hs
nfrisby: This instance also makes me feel uneasy. I _think_ I'm concerned about perhaps needing to version the serialization of a `Point`.

@newhoggy can you point out where this instance is being used? Thanks.

nfrisby (commented):
I shared some minor-to-medium observations, mostly around avoiding code duplication.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Fixed!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -162,16 +228,26 @@ queryDecodeNodeToClient codecConfig queryVersion blockVersion
...
+    = encodePoint encode result
 
   decodeResult codecConfig blockVersion (BlockQuery query)
     = decodeResult codecConfig blockVersion query
   decodeResult _ _ GetSystemStart
     = fromCBOR
+  decodeResult _ _ GetChainBlockNo
+    = fromCBOR
+  decodeResult _ _ GetChainPoint
+    = decodePoint decode
```
newhoggy, line 250: I've inlined code that was previously in the typeclass.

newhoggy: @nfrisby I've inlined the type class instance here.  Although it's unclear if that fixes the problem.

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
newhoggy: How does it look now?

ouroboros-network/src/Ouroboros/Network/Block.hs
newhoggy: See https://github.com/input-output-hk/ouroboros-network/pull/3346#discussion_r710947500

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -374,18 +374,19 @@ instance CardanoHardForkConstraints c
...
+      , (NodeToClientV_5 , CardanoNodeToClientVersion4)
+      , (NodeToClientV_6 , CardanoNodeToClientVersion5)
+      , (NodeToClientV_7 , CardanoNodeToClientVersion6)
+      , (NodeToClientV_8 , CardanoNodeToClientVersion6)
+      , (NodeToClientV_9 , CardanoNodeToClientVersion7)
+      , (NodeToClientV_10, CardanoNodeToClientVersion7)
       ]
 
-  latestReleasedNodeVersion _prx = (Just NodeToNodeV_7, Just NodeToClientV_9)
+  latestReleasedNodeVersion _prx = (Just NodeToNodeV_7, Just NodeToClientV_10)
```
nfrisby, line 389: I'm thinking out loud here, in case someone sees something objectionable.

It's fine to bump this, since it's inconsequential whether the next release/release candidate tries to negotiate this new version. Esp: if they talk to a node that doesn't support it, they'll simply negotiate the eg previous version. Thus, as long as the `cardano-node` dependence on these new queries is well-guarded wrt to the negotiated version, then it's fine to "enable" them here.

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
```diff
@@ -228,17 +228,32 @@ instance ( blockVersion ~ BlockNodeToClientVersion blk
...
+    where
+      mkEntry :: QueryVersion
+        -> Query blk query
+        -> Gen
+            (WithVersion (QueryVersion, blockVersion) (SomeSecond Query blk))
+      mkEntry qv q = do
+        blockV <- arbitrary
+        return (WithVersion (qv, blockV) (SomeSecond q))
+
+      genTopLevelQuery1 = let version = Query.QueryVersion1 in
```
nfrisby, line 243: Hmm. The `StyleGuide.md` is not clear on this, which surprises me.

So I'll just phrase this a style _request_ :D

```
genTopLevelQuery1 =
  let version = Query.QueryVersion1
  in frequency
       [ (15, arbitraryBlockQuery version    )
       , (1 , mkEntry version GetSystemStart )
       ]
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
nfrisby: I think it does, at least for me. Because now, if we need version-dependent serialization of `Point` in the future, then we can scrutinize the in-scope `version` argument here (currently named `_`).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -95,6 +123,7 @@ data QueryEncoderException blk =
 deriving instance Show (SomeSecond BlockQuery blk) => Show (QueryEncoderException blk)
 instance (Typeable blk, Show (SomeSecond BlockQuery blk)) => Exception (QueryEncoderException blk)
 
+
```
nfrisby, line 126: Please remove this extra newline.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -162,16 +228,26 @@ queryDecodeNodeToClient codecConfig queryVersion blockVersion
         blockVersion
       return (SomeSecond (BlockQuery blockQuery))
 
-instance SerialiseResult blk (BlockQuery blk) => SerialiseResult blk (Query blk) where
+instance  ( SerialiseResult blk (BlockQuery blk)
+          , Serialise (HeaderHash blk)
```
nfrisby, line 232: We have a general preference for `ToCBOR` and `FromCBOR` instead of `Serialise`. But using `Serialise` here is not without precedent (eg `BlockQuery ShelleyBlock` uses `Serialise` for a `Point`'s `HeaderHash`).

Ultimately, the `Serialise` vs `ToCBOR`/`FromCBOR` vs `SerialiseNodeToClient`/`SerialiseNodeToNode` vs `SerialiseResult` is in a confused and confusing state, in my opinion. The only stark contrast is that many "low-level" `Serialise` instances come from 3rd party libraries, and so we shouldn't be using them: changes in our codecs should never be accidental, and so shouldn't run the risk of accidentally depending on 3rd party serializations.

So: if it's not too much trouble, I would prefer at least `ToCBOR` and `FromCBOR` instead of `Serialise`. But really, this is already in the scope of existing tech debt :(

nfrisby (dismissed):
I pointed out enough small-to-medium things that I'm Requesting Changes.

Also: apologies for not grouping all my comments into this review; I made a few before realizing there'd be more than a couple.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: I've eliminated the `name` argument.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: If it's okay I'd like to do this in a separate PR.

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
```diff
@@ -41,6 +41,8 @@ data NodeToClientVersion
     -- ^ 'LocalStateQuery' protocol codec change, allows to acquire tip point.
     | NodeToClientV_9
     -- ^ enabled @CardanoNodeToClientVersion7@, i.e., Alonzo
+    | NodeToClientV_10
+    -- ^ enabled @CardanoNodeToClientVersion7@, i.e., Alonzo
```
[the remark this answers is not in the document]
newhoggy: Done!

ouroboros-consensus-test/src/Test/Util/Serialisation/Roundtrip.hs
newhoggy: Done!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
newhoggy: Done!

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -162,16 +228,26 @@ queryDecodeNodeToClient codecConfig queryVersion blockVersion
         blockVersion
       return (SomeSecond (BlockQuery blockQuery))
 
-instance SerialiseResult blk (BlockQuery blk) => SerialiseResult blk (Query blk) where
+instance  ( SerialiseResult blk (BlockQuery blk)
```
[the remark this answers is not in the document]
newhoggy: Done!

ouroboros-network/src/Ouroboros/Network/Block.hs
```diff
@@ -92,7 +92,6 @@ import           GHC.Generics (Generic)
 import           NoThunks.Class (NoThunks)
 
 import           Cardano.Binary (Case (..), Size, szCases, szGreedy)
-
```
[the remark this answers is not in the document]
newhoggy: Done!

nfrisby:
@dnadales pointed out on Slack that this PR is enriching our interface, and so deserves an entry in https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/interface-CHANGELOG.md 🙌 

I'm sorry I didn't realize that sooner.

@newhoggy Would you add a section at the top there. I'm guess it'll be similarly formatted to the latest one (ie Circa 2021-08-31). If the top of the `interface-CHANGELOG.md` file doesn't give you enough guidance, please let me know 👍

dnadales (approved):

PR #3354, opened 2021-09-16
CAD-3444 pull changes from p2p-master

coot:
- Snocket.Accept
- ouroboros-network-testing: generate delay with a given precision
- handshake codec
- typed-protocols: added TokPeerRole
- handshake: simplify types of client / server
- handshake: resiliant for simultanous open
- handshake: refactor the codec
- handshake: added test for a simultaneous open connections
- handshake: updated network-spec
- handshake: remove versioned application from HandshakeArguments
- Use a counter to provide a unique remote addr
- snockets - monadic toBearer
- snocket: support getLocalName for named pipes (Windows)
- snocket: store local and remote path     
- snocket: make Window's LocalSocket strict 
- network-mux: label mini-protocol thread  
- network-mux: improve attenuated channel errors 
- network-mux: SDUSize show via Quiet      
- network-mux: mux runtime exceptions
- io-sim: label Async's TVar
- io-sim: log EventMask                    
- io-sim: added pretty printers            
- io-sim: Octopus                          
- io-sim: selectTraceEvents' and friends

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
@@ -345,7 +345,13 @@ type LocalHandle = Socket
 --
 #if defined(mingw32_HOST_OS)
 data LocalSocket = LocalSocket { getLocalHandle :: LocalHandle
+                                 -- ^ underlying windows 'HANDLE'
                                , getLocalPath   :: LocalAddress
+                                 -- ^ original path, used when creating the handle
+                               , getRemotePath  :: LocalAddress
+                                 -- ^ unique identifier (not a real path).  It
+                                 -- makes the pair of local and remote
+                                 -- addresses unique.
                                }
     deriving (Eq, Generic)
     deriving Show via Quiet LocalSocket
@@ -369,7 +375,7 @@ localSnocket :: IOManager -> LocalSnocket
 #if defined(mingw32_HOST_OS)
 localSnocket ioManager = Snocket {
       getLocalAddr  = return . getLocalPath
-    , getRemoteAddr = return . getLocalPath
+    , getRemoteAddr = return . getRemotePath
     , addrFamily    = LocalFamily
 
     , open = \(LocalFamily addr) -> do
@@ -389,7 +395,7 @@ localSnocket ioManager = Snocket {
           `catch` \(SomeAsyncException _) -> do
             Win32.closeHandle hpipe
             throwIO e
-        pure (LocalSocket hpipe addr)
+        pure (LocalSocket hpipe addr (LocalAddress ""))
 
     -- To connect, simply create a file whose name is the named pipe name.
     , openToConnect  = \(LocalAddress pipeName) -> do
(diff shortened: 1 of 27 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
```diff
@@ -342,25 +342,41 @@ type LocalHandle = Socket
...
+--
+localSnocket :: IOManager -> LocalSnocket
 #if defined(mingw32_HOST_OS)
-localSnocket ioManager path = Snocket {
-      getLocalAddr  = \_ -> return localAddress
-    , getRemoteAddr = \_ -> return localAddress
-    , addrFamily  = \_ -> LocalFamily
+localSnocket ioManager = Snocket {
+      getLocalAddr  = return . getLocalPath
+    , getRemoteAddr = return . getLocalPath
```
karknu, line 372: Reusing the local path as the remote path doesn't make for a unique connection id. I think we need to store the counter path we use for `accept` here too. Otherwise the address return from `accept` won't match subsequent calls to getRemoteAddr.

bolt12:
Why are these commits being pulled to master?

ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
coot: Then we need to store the counter inside the `LocalSnocket`, let's see if this will work.

coot:
> Why are these commits being pulled to master?

Smaller `p2p-master` branch is easier to maintain.

bolt12:
I see, but won't this make rebasing harder?

bolt12:
You're also experiencing the same issues as I am with Windows CI

coot:
> I see, but won't this make rebasing harder?

No, because I first rebased my local `p2p-master`, and then took the interesting commits, and rebased that on top of current `master` (there was a small conflict, but it will be easy to fix once we rebase `p2p-master` over the merged commits).

bolt12:
I see, thank you!

ouroboros-network-framework/src/Ouroboros/Network/Snocket.hs
coot: I think it will be unique: check [this](https://github.com/input-output-hk/ouroboros-network/pull/3354/files/f755ea5eabf0ef76b2ea3b82ce063d7fd14b3fa1?file-filters%5B%5D=.hs#diff-b89417778eab9a61bc587228ba77a40615f2a501aae28864cbf631139769f7d7R464-R465).

But it does not seem right that we use `getLocalPath` for both local and remote.  We could store both: unmodified path (to be used by `getLocalAddr) and modified one (for `getRemoteAddr`).

karknu: >But it does not seem right that we use getLocalPath for both local and remote. We could store both: unmodified path (to be used by getLocalAddr) and modified one (for getRemoteAddr`).
Yes.

coot:
Note, the following two commits are new:

* snocket: make Window's LocalSocket strict 
* snocket: store local and remote path

coot:
I will divide this into smaller PRs.

PR #3356, opened 2021-09-16
Update ledger.

nc6:
This reverts the change to the incremental stake computation, which was
causing performance regressions. As such, the on-disk state has again
changed.


nfrisby (approved):
Nothing objectionable in this ~5 line diff.

nc6:
bors merge

PR #3368, opened 2021-09-20
Workaround for OSX accept bug

karknu:
There is a bug in accept for IPv6 sockets. Instead of returning -1
and setting errno to ECONNABORTED an invalid (>= 0) file descriptor
is returned, with the client address left unchanged. The uninitialized
client address causes the network package to throw a user error.


coot (approved):
LGTM, thanks!

PR #3379, opened 2021-09-22
[CAD-3383] Add --num-blocks-to-process option to db-analyser

EncodePanda:
Add --limit option so that user can control number of blocks they want to process

```
λ cabal run ouroboros-consensus-cardano:db-analyser -- \
  --db db_mainnet_with_alonzo_40217333 cardano \
  --configByron dbacfg/mainnet-byron-genesis.json \
  --configShelley dbacfg/mainnet-shelley-genesis.json \
  --nonce 1a3be38bcbb7911969283716ad7aa550250226b76a61fc51cc9a9a35d9276d81 \
  --configAlonzo dbacfg/mainnet-alonzo-genesis.json \
  --show-slot-block-no \
  --limit 10

BlockNo 0	SlotNo 0
BlockNo 1	SlotNo 0
BlockNo 2	SlotNo 1
BlockNo 3	SlotNo 2
BlockNo 4	SlotNo 3
BlockNo 5	SlotNo 4
BlockNo 6	SlotNo 5
BlockNo 7	SlotNo 6
BlockNo 8	SlotNo 7
BlockNo 9	SlotNo 8
ChainDB tip: At (Block {blockPointSlot = SlotNo 40217333, blockPointHash = fd4487549a0d2c6dcba32255c2a4f7e1419cb1c88a07b8872a32db19e4b02a41})
```

### DIFF
--- a/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
+++ b/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
@@ -7,6 +7,7 @@
 module Analysis (
     AnalysisEnv (..)
   , AnalysisName (..)
+  , Limit (..)
   , runAnalysis
   ) where
 
@@ -79,6 +80,7 @@ data AnalysisEnv blk = AnalysisEnv {
     , db         :: Either (ImmutableDB IO blk) (ChainDB IO blk)
     , registry   :: ResourceRegistry IO
     , ledgerDbFS :: SomeHasFS IO
+    , limit      :: Limit
     }
 
 {-------------------------------------------------------------------------------
@@ -86,8 +88,8 @@ data AnalysisEnv blk = AnalysisEnv {
 -------------------------------------------------------------------------------}
 
 showSlotBlockNo :: forall blk. HasAnalysis blk => Analysis blk
-showSlotBlockNo AnalysisEnv { db, registry } =
-    processAll_ db registry GetHeader process
+showSlotBlockNo AnalysisEnv { db, registry, limit } =
+    processAll_ db registry GetHeader limit process
   where
     process :: Header blk -> IO ()
     process hdr = putStrLn $ intercalate "\t" [
@@ -100,8 +102,8 @@ showSlotBlockNo AnalysisEnv { db, registry } =
 -------------------------------------------------------------------------------}
 
 countTxOutputs :: forall blk. HasAnalysis blk => Analysis blk
-countTxOutputs AnalysisEnv { db, registry } = do
-    void $ processAll db registry GetBlock 0 process
+countTxOutputs AnalysisEnv { db, registry, limit } = do
+    void $ processAll db registry GetBlock limit 0 process
   where
     process :: Int -> blk -> IO Int
     process cumulative blk = do
@@ -121,9 +123,9 @@ countTxOutputs AnalysisEnv { db, registry } = do
 -------------------------------------------------------------------------------}
 
 showHeaderSize :: forall blk. HasAnalysis blk => Analysis blk
-showHeaderSize AnalysisEnv { db, registry } = do
+showHeaderSize AnalysisEnv { db, registry, limit } = do
     maxHeaderSize <-
-      processAll db registry ((,) <$> GetSlot <*> GetHeaderSize) 0 process
+      processAll db registry ((,) <$> GetSlot <*> GetHeaderSize) limit 0 process
     putStrLn ("Maximum encountered header size = " <> show maxHeaderSize)
   where
     process :: Word16 -> (SlotNo, Word16) -> IO Word16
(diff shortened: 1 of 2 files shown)


ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -124,6 +126,16 @@ storeLedgerParser = (StoreLedgerStateAt . SlotNo . read) <$> strOption
   <> metavar "SLOT NUMBER"
   <> help "Store ledger state at specific slot number" )
 
+parseLimit :: Parser Limit
+parseLimit = asum [
```
nfrisby, line 130: I was expecting to see https://hackage.haskell.org/package/optparse-applicative-0.16.1.0/docs/Options-Applicative.html#v:value (which would require `option` instead of `strOPtion`, I think) used here to specify defaults. But I see that what you've written here at least matches the rest of the option definitions in this module.

I don't recall if there's a difference here. I'm guessing `value` _might_ lead to the default value being displayed in the `--help` message?

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -235,11 +237,19 @@ storeLedgerStateAt slotNo (AnalysisEnv { db, registry, initLedger, cfg, ledgerDb
   Auxiliary: processing all blocks in the DB
 -------------------------------------------------------------------------------}
 
+data Limit = Limit Int | Unlimited
+
+decreaseLimit :: Limit -> Limit
+decreaseLimit Unlimited = Unlimited
+decreaseLimit (Limit 0) = Limit 0
```
nfrisby, line 244: `0 -> 0` is a little unexpected. One way to avoid it:

```
decreaseLimit :: Limit -> Maybe Limit
...
decreaseLimit (Limit 0) = Nothing
```

And then use `case decreaseLimit lt of ...` instead of `case lt of ...` in the `go` loops.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -79,15 +80,16 @@ data AnalysisEnv blk = AnalysisEnv {
     , db         :: Either (ImmutableDB IO blk) (ChainDB IO blk)
     , registry   :: ResourceRegistry IO
     , ledgerDbFS :: SomeHasFS IO
+    , limit      :: Limit
```
nfrisby, line 83: `limit` is a bit terse, for the name in the code and for the name of the command-line option.

Maybe `NumBlocksToProcess` and `--num-blocks-to-process`?

nfrisby (commented):
👍 A few minor suggestions.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
EncodePanda: when you run the tool you get meta info what you are limiting

```
Usage: db-analyser --db PATH [--verbose]
                   (....)
                   [--limit NUMBER OF BLOCKS]
 ```
 
TBH I much prefer using a tool where I have to type `--limit` then `--num-blocks-to-process`. It's shorter and easy to remember. `--limit` I can type, `--num-blocks-to-process` I will most likely copy-paste from help message

File: ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs (hunk 2 above)
Comment (EncodePanda) [reply ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs:244]: I don't like this idea. The whole point to have a function was to call it without pattern matching on the Limit. Now I will just change the Limit to Maybe, and will have to pattern match anyway. If that's the case then I would rather prefer pattern matching on the Limit directly and remove the `decreaseLimit` function.

I still think `decreaseLimit` makes sense, but maybe it should have been a private function, defined only in that one single place where it is being used.

File: ouroboros-consensus-cardano/tools/db-analyser/Main.hs (hunk 1 above)
Comment (EncodePanda) [reply ouroboros-consensus-cardano/tools/db-analyser/Main.hs:130]: we are going to revisit the whole interface before we deliver the tool to ledger team and benchmarking team https://input-output.atlassian.net/browse/CAD-3477

File: ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs (hunk 2 above)
Comment (EncodePanda) [reply ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs:244]: Ok, after a call, I get what you meant.

**Comment by EncodePanda** [comment]
@nfrisby this is now ready for 2nd review

**Comment by EncodePanda** [comment]
```
Usage: db-analyser --db PATH [--verbose] [--onlyImmutableDB]
                   [--validate-all-blocks | --minimum-block-validation] COMMAND
                   [--show-slot-block-no | --count-tx-outputs |
                     --show-block-header-size | --show-block-txs-size |
                     --show-ebbs | --store-ledger SLOT NUMBER]
                   [--num-blocks-to-process INT]
```

File: ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs [hunk 4]
```diff
@@ -250,46 +260,53 @@ processAll_ ::
...
-    go :: ChainDB.Iterator IO blk b -> st -> IO st
-    go itr !st = do
-      itrResult <- ChainDB.iteratorNext itr
-      case itrResult of
-        ChainDB.IteratorExhausted    -> return st
-        ChainDB.IteratorResult b     -> callback st b >>= go itr
-        ChainDB.IteratorBlockGCed pt -> error $ "block GC'ed " <> show pt
+    go :: ChainDB.Iterator IO blk b -> Limit -> st -> IO st
+    go itr lt !st = case decreaseLimit lt of
+      Nothing -> return st
```
nfrisby, line 284: Please align these `->` and remove the parens around the `Just` pattern.

EncodePanda:
bors merge

PR #3380, opened 2021-09-22
cardano-ping: Make sure that the connection is reset by using SO_LINGER

karknu:
The way the connection is torn down isn't ideal. Unless the 5s timeout
triggers the connection termination is initiated by the server which
can leave connections in TIME_WAIT.
In the future `ouroboros-network` will use SO_LINGER to trigger a reset
of the connection, but for now at least cardano-node can set it.


coot (approved):
LGTM

PR #3388, opened 2021-09-23
CAD-3474 p2p version of osx accept workaround

karknu:


ouroboros-network-framework/test/Test/Ouroboros/Network/Subscription.hs
```diff
@@ -32,6 +32,11 @@ import           Data.Void (Void)
 import           Data.Word
 import qualified Network.DNS as DNS
 import qualified Network.Socket as Socket
+#if !defined(mingw32_HOST_OS)
+import qualified Network.Socket.ByteString.Lazy as Socket (recv, sendAll)
+#else
+import qualified System.Win32.Async.Socket.ByteString.Lazy as Win32.Async
+#endif
```
coot: If we use the same qualification, we can avoid using `CPP` below.  `win32-network` modules and api where structured after the `network` package for this purpose.

```suggestion
#if !defined(mingw32_HOST_OS)
import qualified Network.Socket.ByteString.Lazy as Socket (recv, sendAll)
#else
import qualified System.Win32.Async.Socket.ByteString.Lazy as Socket (recv, sendAll)
#endif
```

ouroboros-network-framework/test/Test/Ouroboros/Network/Subscription.hs
```diff
@@ -493,13 +491,10 @@ prop_sub_io lr = ioProperty $ withIOManager $ \iocp -> do
...
-#if defined(mingw32_HOST_OS)
-        Win32.Async.sendAll sd $ BL.singleton 42
-        _ <- Win32.Async.recv sd 1
-#else
-        Socket.sendAll sd $ BL.singleton 42
-        _ <- Socket.recv sd 1
+    initiatorCallback clientCountVar _sd = do
+#if !defined(mingw32_HOST_OS)
+        Socket.sendAll _sd $ BL.singleton 42
+        _ <- Socket.recv _sd 1
```
coot, line 497: We don't need to do that on windows?

ouroboros-network-framework/test/Test/Ouroboros/Network/Subscription.hs
```diff
@@ -518,13 +513,10 @@ prop_sub_io lr = ioProperty $ withIOManager $ \iocp -> do
...
-#if defined(mingw32_HOST_OS)
-                        buf <- Win32.Async.recv sd' 1
-                        Win32.Async.sendAll sd' buf
-#else
-                        buf <- Socket.recv sd' 1
-                        Socket.sendAll sd' buf
+                    (\(_sd',_) -> do
+#if !defined(mingw32_HOST_OS)
+                        buf <- Socket.recv _sd' 1
+                        Socket.sendAll _sd' buf
```
coot, line 519: We don't need to support windows?

coot (changes requested):

ouroboros-network-framework/test/Test/Ouroboros/Network/Subscription.hs
karknu: We only need to do this on OSX. I couldn't get it to work on windows and got tired of pushing guesses to get the CI to pass.

ouroboros-network-framework/test/Test/Ouroboros/Network/Subscription.hs
karknu: See previous comment.

PR #3397, opened 2021-09-28
[WIP][CAD-3383] Init db-analyser from stored ledger state snapshot

EncodePanda:


ouroboros-consensus-cardano/tools/db-analyser/HasAnalysis.hs
```diff
@@ -9,14 +9,14 @@ import           Data.Map.Strict (Map)
...
+import           Ouroboros.Consensus.HeaderValidation (HasAnnTip (..))
 import           Ouroboros.Consensus.Node.ProtocolInfo
 import           Ouroboros.Consensus.Storage.Serialisation (SizeInBytes)
 
 {-------------------------------------------------------------------------------
   HasAnalysis
 -------------------------------------------------------------------------------}
-
-class GetPrevHash blk => HasAnalysis blk where
+class (HasAnnTip blk, GetPrevHash blk) => HasAnalysis blk where
```
nfrisby, line 19: Style: put the blank line above the class, like it was before.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -245,38 +249,50 @@ decreaseLimit (Limit 0) = Nothing
...
   -> ResourceRegistry IO
   -> BlockComponent blk b
+  -> ExtLedgerState blk
   -> Limit
   -> st
   -> (st -> b -> IO st)
   -> IO st
-processAllChainDB chainDB rr blockComponent limit initState callback = do
-    itr <- ChainDB.streamAll chainDB rr blockComponent
+processAllChainDB chainDB registery blockComponent (ExtLedgerState {..}) limit initState callback = do
```
nfrisby, line 285: Typo: `registery` should be `registry`

(This is elsewhere in the module too.)

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -171,6 +187,7 @@ getCmdLine = execParser opts
 -------------------------------------------------------------------------------}
 
 analyse ::
```
nfrisby, line 189: We may need to be clearer about what it means for the tool to "start from the given snapshot".

Specifically, in the current PR, if the command is to open the ChainDB and not just the ImmDB, then the ChainDB will initialize its LgrDB in the same way it always does, via the internal call to `initLedgerDB`.

https://github.com/input-output-hk/ouroboros-network/blob/d6826269569f239f262d8e72f2fd6f280334e4f7/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs#L180-L237

Thus, if "the given snapshot" is not the one that `initLedgerDB` will choose, then there is room for the user to be surprised/confused.

Specifically, if the given snapshot is not the latest snapshot, then ChainDB will initialize from a later snapshot than "the given snapshot".

Since we're considering using this tool for benchmarking, for example, we need to be really clear about what phases it goes through. Moreover, it's possible the user will want more control that this PR currently offers. Notably, the user may want to specify which ledger snapshot the ChainDB's LedgerDB is initialized from. (At the moment, they'd have to do that by removing any later snapshots from the snapshot directory.)

-----

In fact, re-reading my comment above, I'm thinking what we probably want is a mode in which this tool opens the ImmDB and the VolDB but never opens the LgrDB. That'd be the easiest way to ensure that the "startup" phase of the tool involves no replay whatsoever, which I anticipate would be desirable for benchmarking uses.

nfrisby: In particular the current `--help` of `"Initialize from stored ledger state stored at specific slot number"` is likely misleading for our current envisioned users/uses. I'd suggest instead

>Begin the analysis phase from the ledger state stored at the given slot number

nfrisby (changes requested):
I'm Requesting Changes if only to ensure the `--help` text becomes more specific. But we may need to have a higher-level conversation (at our Wednesday sync, tomorrow) about what it means for the tool "to start from a snapshot".

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -1,9 +1,14 @@
...
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 -- | Database analyse tool.
 module Main (main) where
 
+import           Codec.CBOR.Decoding (Decoder)
+import           Codec.Serialise (Serialise (decode))
```
nfrisby, line 10: Is it easy enough to use `fromCBOR` instead of `decode`? That's generally preferable and would avoid having to add the `serialise` package to `build-depends`.

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
nfrisby: One short term option: we only allow `--initialize-from` if `--onlyImmutableDB` is also present. That avoids the potential for confusion without complicating this PR.

EncodePanda: Discussion I had with @nfrisby on this particular topic is below:

## Example

Imagine a scenario in which
your local blockchain has 1000
k  paramter is  200
immutable has 800 blocks, and there are 200 blocks which are still in volatile

in the db/ledger there are two snapshots:
* 700
* 500_db-analyser

User starts db-analyser with following parameters: --initialize-from 500 --num-blocks-to-process 400


What will happen:

1. ChainDB will find the latest snapshot (that is 700)
2. it will then start replying blocks from that snapshot up until tip (block 1000), in other words from 701 - 1000
3. That means that 300 blocks were processed while replay happened
4. After that we have:
* 200 blocks (volatile DB)
* we have the tip and its ExtLedgerState in memory
* we have intermediate ExtLedgerState in memory as well (for all 200 blocks)
5. This concludes ChainDB initialization


And now our db-analyser starts doing its thing:
1. it loads snapshot 500 (which holds both the ExtLedgerState and the slot no where it was taken)
2. it asks ChainDB to give it 400 blocks (300 from immutable and 100 from volatile - both loaded from disk)
3. and then for some commands (like --store-ledger) it will start applying those blocks to the ledger state - the one that was selected by --initialize-from (that will be 500)

## Proposed changes:

0. rename --onlyImmutableDB to --only-immutable-db
1. rename --intinialize-from to --analyse-from
2. make --analyse-from only an option if user calls the tool with --only-immutable-db

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
EncodePanda: Not sure, we don't have `FromCBOR` defined for `ExtLedgerState` 

```
    • Could not deduce (cardano-binary-1.5.0:Cardano.Binary.FromCBOR.FromCBOR
                          (ExtLedgerState blk))
        arising from a use of ‘fromCBOR’
```

and this is now `LederDB` reads those snapshots

```
initFromDisk LgrDbArgs { lgrHasFS = hasFS, .. }
             replayTracer
             immutableDB = wrapFailure (Proxy @blk) $ do
    (_initLog, db, replayed) <-
      LedgerDB.initLedgerDB
        replayTracer
        lgrTracer
        hasFS
        decodeExtLedgerState'
        decode
        (configLedgerDb lgrTopLevelConfig)
        lgrGenesis
        (streamAPI immutableDB)
    return (db, replayed)
  where
    ccfg = configCodec lgrTopLevelConfig

    decodeExtLedgerState' :: forall s. Decoder s (ExtLedgerState blk)
    decodeExtLedgerState' = decodeExtLedgerState
                              (decodeDisk ccfg)
                              (decodeDisk ccfg)
                              (decodeDisk ccfg)
```

We could potentially provide `FromCBOR` instance. But that would be a bit out of scope for this ticket (we can create new ticket though).

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -73,15 +86,30 @@ parseCmdLine = CmdLine
...
-            long "onlyImmutableDB"
-          , help "Validate only the Immutable DB (e.g. do not do ledger validation)"
-          ])
+    <*> parseSelectDB
     <*> parseValidationPolicy
     <*> blockTypeParser
     <*> parseAnalysis
     <*> parseLimit
 
+parseSelectDB :: Parser SelectDB
```
nfrisby, line 95: What does the `--help` look like with this? Is there any indication that they cannot pass `analyse-from` unless they pass `only-immutable-db`?

I anticipate it's necessary to encode that requirement explicitly, as an ad-hoc pass that happens immediately after we parse the command line.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -245,38 +249,50 @@ decreaseLimit (Limit 0) = Nothing
...
   -> ResourceRegistry IO
   -> BlockComponent blk b
+  -> ExtLedgerState blk
   -> Limit
   -> st
   -> (st -> b -> IO st)
   -> IO st
-processAllChainDB chainDB rr blockComponent limit initState callback = do
-    itr <- ChainDB.streamAll chainDB rr blockComponent
+processAllChainDB chainDB registry blockComponent (ExtLedgerState {..}) limit initState callback = do
```
nfrisby, line 285: This function no longer needs this, right? We decided (at least for now) in that other comment thread, that the ChainDB option does not support an `--analyse-from`.

nfrisby: EG Move the new `ExtLedgerState blk` to `processAll` and `processAll_` into the `Left` side of the `Either`, paired with the `ImmutableDB`.

nfrisby: Better yet: leave `processAll` alone, and introduce a new `process***` combinator for analyses that need an initial ledger state. That combinator would eg immediately always fail if `db` is a `Right`.

nfrisby (changes requested):
I made some suggestions for simplifying based on the fact that a ChainDB-based analysis always starts from `Origin` (since we had decided that `--analyse-from` should require `--only-immutable-db`).

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
nfrisby: If this refactor is too intensive, feel free to open a tech-debt follow-up issue to unblock this PR. My request here is an implementation detail. (Unlike the other thread where I requested an explicit check on the command-line arguments' coherency.)

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
EncodePanda: I've spent some time on this so that it looks exactly as you would hope it to look :)

If you run the tool with no arguments you get

```
cabal run ouroboros-consensus-cardano:db-analyser
Temporary modify `cabal.project` to use nix builds of `source-repository-package`s (haskell/cabal#5444 and haskell/cabal#6249).
Up to date
Missing: --db PATH COMMAND

Usage: db-analyser --db PATH [--verbose]
                   [--only-immutable-db [--analyse-from SLOT_NUMBER]]
                   [--validate-all-blocks | --minimum-block-validation] COMMAND
                   [--show-slot-block-no | --count-tx-outputs |
                     --show-block-header-size | --show-block-txs-size |
                     --show-ebbs | --store-ledger SLOT_NUMBER]
                   [--num-blocks-to-process INT]
```

If you provide the `--analyse-from` but forget to mention to run on immutable db only you get a nice error


```
cabal run ouroboros-consensus-cardano:db-analyser -- --analyse-from 1
Temporary modify `cabal.project` to use nix builds of `source-repository-package`s (haskell/cabal#5444 and haskell/cabal#6249).
Up to date
Missing: --db PATH --only-immutable-db COMMAND

Usage: db-analyser --db PATH [--verbose]
                   [--only-immutable-db [--analyse-from SLOT_NUMBER]]
                   [--validate-all-blocks | --minimum-block-validation] COMMAND
                   [--show-slot-block-no | --count-tx-outputs |
                     --show-block-header-size | --show-block-txs-size |
                     --show-ebbs | --store-ledger SLOT_NUMBER]
                   [--num-blocks-to-process INT]
```

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
EncodePanda: I'm ok implementing it now. I'm only hesitant because I have a gut feeling that the ledger and/or benchmarking team will explicitly ask for `--analyse-from` for ChainDB. But we can always revisit it I guess (even reverting the single commit if needed)

EncodePanda: After thinking about it, how about we create a tech-debt ticket but we will implement it in a week or two from now, once the tool is used by other teams and they will confirm that they like what they have?

EncodePanda: I've created two tickets:
1. https://input-output.atlassian.net/browse/CAD-3477
2. https://input-output.atlassian.net/browse/CAD-3516

ouroboros-consensus-cardano/tools/db-analyser/Main.hs
nfrisby: I think these are all done. Approving shortly!

nfrisby (approved):
Approved!

- Please clean-up commits before merging (looks like the last two have the same message?)

- Similarly, somethings (like `--onlyImmutableDb` having old capitalization) are out of date in the PR description, which `bors` puts as the merge commit's message.

EncodePanda:
bors merge

PR #3398, opened 2021-09-29
Fix race condition in the thread tracking set

karknu:
It was possible for threads to get "removed" before spawnOne had added
the threads. This caused those threads to never get removed which would
trigger ratelimiting and eventually the node would refuse to accept new
connections.

Change it so that if the thread isn't inserted yet we will wait for
spawnOne to do it before we remove the thread.

Thanks to @psychomb for helping troubleshoot this problem.


coot (approved):
LGTM

dcoutts (approved):

PR #3400, opened 2021-09-29
[CAD-3383] Stop processing when snapshot created

EncodePanda:
Improvement to the `--store-ledger` functionality. It stops the blockchain processing once the snapshot is created.

### DIFF
--- a/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
+++ b/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
@@ -1,8 +1,10 @@
 {-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE BlockArguments      #-}
 {-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections       #-}
 {-# LANGUAGE TypeApplications    #-}
 module Analysis (
     AnalysisEnv (..)
@@ -105,7 +107,7 @@ countTxOutputs :: forall blk. HasAnalysis blk => Analysis blk
 countTxOutputs AnalysisEnv { db, registry, limit } = do
     void $ processAll db registry GetBlock limit 0 process
   where
-    process :: Int -> blk -> IO Int
+    process :: Int -> blk -> IO (NextStep, Int)
     process cumulative blk = do
         let cumulative' = cumulative + count
         putStrLn $ intercalate "\t" [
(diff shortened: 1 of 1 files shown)


dnadales:
I wonder if we could leave `processAll` unchanged, and define it in terms of a more general function:

```haskell
-- | Process all blocks with the given callback until either:
--
-- - The termination condition evaluates to True
-- - There are no more blocks left.
--
processAllUntil ::
     forall blk b st. HasHeader blk
  => Either (ImmutableDB IO blk) (ChainDB IO blk)
  -> ResourceRegistry IO
  -> BlockComponent blk b
  -> st
  -> (st -> b -> IO st)
  -- ^ State transforming function (callback).
  -> (st -> b -> st -> Bool)
  -- ^ Early-termination condition.
  --
  -- The first parameter is the termination condition is the state prior to
  -- applying the given state transforming function. The second parameter is
  -- the block to which this state transforming function is applied, and the
  -- third parameter is its result.
  -> IO st

processAll db rr blockComponent st callback =
  processAllUntil db rr blockComponent st callback (const . const . const False)
```

In this way:

- we offer more flexibility to define other stopping conditions (we don't know with which use cases people are going to come with)
- we minimize the changes in the existing functions and present and future duplication (no need to alter the `process` functions and repeat the same `return (Continue, _)` pattern)

I also wonder, should the termination condition be an IO action as it is now, or do we want to keep it pure?

EncodePanda:
@dnadales I've went with different approach. Introduced `processAllUntil` with a general callback function. Then `processAll` is build on top of it, and `processAll_` is build on top of `processAll`. Let me know what you think.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -200,15 +202,20 @@ storeLedgerStateAt slotNo (AnalysisEnv { db, registry, initLedger, cfg, limit, l
...
   where
-    process :: ExtLedgerState blk -> blk -> IO (ExtLedgerState blk)
+    process :: ExtLedgerState blk -> blk -> IO (NextStep, ExtLedgerState blk)
     process oldLedger blk = do
       let ledgerCfg     = ExtLedgerCfg cfg
           appliedResult = tickThenApplyLedgerResult ledgerCfg blk oldLedger
           newLedger     = either (error . show) lrResult $ runExcept $ appliedResult
-      when (slotNo == blockSlot blk) $ storeLedgerState blk newLedger
-      return newLedger
+      when (blockSlot blk == slotNo) $ storeLedgerState blk newLedger
```
nfrisby, line 212: Consider calling `fail` if we reach `Stop` without ever having seen `blockSlot blk == slotNo`.

This will minimize surprise for a user with a typo in their desired slot number.

nfrisby (approved):
One last UX suggestion (which we chatted about before). Otherwise LGTM. Approved.

dnadales (approved):

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
nfrisby: Are even writing the ledger snapshot for the next available slot and then issuing a WARNING that they specified an empty slot.

EncodePanda:
bors merge

PR #3409, opened 2021-10-05
Use real byte size limits

### LINKED ISSUE #1727: Add per state size limits in Ouroboros-Network

coot:
Fixes #1727.


coot:
> Is that right? That seems fine for Cardano mainnet. But is it a fit for other (potential) users of this library? If not, then we ultimately need to make those numbers depend on eg the block type blk.

Block size is a parameter of the network and yes these limits should ultimately be configurable rather than hard coded.  But block size is independent of block type (though tx size, probably does depend on tx type).  So ultimately, these limits should be provided by a `cardano-node` configuration file (or hard coded there).

I think this makes it clear why we should keep the interface proposed in this PR.  Note also timeouts are different, since they are much less conservative than size limits, but in long run we will also need to make them configurable.

PR #3413, opened 2021-10-06
Expose ability to configure mempoolCapacityOverride

newhoggy:


ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -677,7 +680,9 @@ stdLowLevelRunNodeArgsIO RunNodeArgs{ rnProtocolInfo } StdRunNodeArgs{..} = do
       , llrnChainDbArgsDefaults =
           updateChainDbDefaults $ ChainDB.defaultArgs mkHasFS diskPolicy
       , llrnCustomiseChainDbArgs = id
-      , llrnCustomiseNodeKernelArgs
+      , llrnCustomiseNodeKernelArgs = case rnMaybeMempoolCapacityOverride of
```
nfrisby, line 683: Would you move this logic down to lines 730? And follow the pattern there with `maybe id`? This code itself is currently wrong since a `Just` value here will override the existing logic on line 730.

PR #3417, opened 2021-10-07
Connection Manager/Inbound Governor Counters tests

bolt12:


ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2123,6 +2131,90 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
                        maxAcceptedConnectionsLimit l
 
+-- | Property wrapping `multinodeExperiment`.
+--
+-- Note: this test validates connection manager counters.
+--
+prop_connection_manager_counters :: Int
+                               -> ArbDataFlow
+                               -> MultiNodeScript Int TestAddr
+                               -> Property
```
coot, line 2141: Please align the type signature.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2123,6 +2131,90 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+prop_connection_manager_counters serverAcc (ArbDataFlow dataFlow)
+                               script@(MultiNodeScript l) =
```
coot, line 2143: aligning the arguments would help here as well

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2123,6 +2131,90 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+              Map.adjust (ConnectionManagerCounters 0 0 0 0 1 <>) ta st
+            InboundMiniprotocols _ ta _ ->
+              Map.adjust (ConnectionManagerCounters 0 0 0 0 0 <>) ta st
+            OutboundMiniprotocols _ ta _ ->
+              Map.adjust (ConnectionManagerCounters 0 0 0 0 0 <>) ta st
+            CloseOutboundConnection _ ta ->
+              Map.adjust (ConnectionManagerCounters 0 0 0 1 0 <>) ta st
+            _ -> st
+          )
+          Map.empty
```
coot, line 2211: What's your thought about using information available at `Simulation.Network.Snocket.NetworkState`?  I think it should be posilbe to improve this test: rather than checking for an uperbound, we should be able to compute what's the current number of inbound / outbound connections.   Then we could use a special `Tracer` which would verify that the counters agree with the `NetworkState`.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2123,6 +2131,90 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+       ( \ trs
+        -> case trs of
+          TrConnectionManagerCounters cmc ->
+            AllProperty
+              $ counterexample
+                  ("Upper bound is: " ++ show upperBound
+                  ++ "\n But got: " ++ show cmc)
+                  (property $ cmc <= upperBound)
+          _                               ->
+            mempty
```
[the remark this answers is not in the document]
bolt12: I like having symmetries but do not have a strong opinion on how it should be formatted, if you make enough pressure I'll change otherwise I'll keep it as is 😛

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
coot: ok, ping me when you'll be ready with this.

ouroboros-network-testing/src/Ouroboros/Network/Testing/Data/Script.hs
```diff
@@ -60,22 +61,44 @@ arbitraryScriptOf :: Int -> Gen a -> Gen (Script a)
...
+    return x
+
+stepScriptSTM :: MonadSTMTx stm
+               => TVar_ stm (Script (stm b)) -> stm b
+stepScriptSTM scriptVar = do
+  Script (script :| scripts) <- LazySTM.readTVar scriptVar
+  case scripts of
+    []     -> return ()
+    x':xs' -> LazySTM.writeTVar scriptVar (Script (x' :| xs'))
+  script
```
coot, line 100:
```suggestion
stepScriptSTM = stepScriptSTM >=> id
```

I think I would avoid it after all and retain the original api

coot: It doesn't seem used in this commit (*Factor Signal, Script, Utils modules*) as well.

bolt12: Hmm good point, I'll look into it!

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -207,6 +209,15 @@ data NetworkState m addr = NetworkState {
 
     }
 
+-- | Simulation accessible network environment consumed by 'simSnocket'.
+--
+newtype UniverseState addr = UniverseState {
```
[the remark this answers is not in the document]
bolt12: Yeah, that is better!

ouroboros-network-testing/src/Ouroboros/Network/Testing/Data/Script.hs
coot: What was your conclusion?

bolt12: My conclusion was that it was not used in this PR but used in another one ( I added it under a different name, in the respective PR)

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2151,6 +2159,173 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+    serverAddress :: SimAddr
+    serverAddress = TestAddress 0
+
+    -- We count all connections as prunable because we do not have a better way
+    -- to know what transitions will a given connection go through. We also
+    -- count every client connection as unidirectional because they do not
+    -- negotiate duplex data flow.
+    --
+    -- We do not count Incoming/Outgoing/Closing of connections since the
+    -- UniverseState is a much more reliable source of information.
```
coot, line 2231: `UniverseState` -> `ObservableNetworkState`.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2151,6 +2159,173 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+      evsCMT :: Trace (SimResult ())
+                        (ConnectionManagerTrace
+                          SimAddr
+                          (ConnectionHandlerTrace
+                            UnversionedProtocol
+                            DataFlowProtocolData))
+      evsCMT = traceWithNameTraceEvents trace
+
+      -- Needed for calculating a more accurate upper bound
+      evsUS :: [ObservableNetworkState SimAddr]
```
coot, line 2191: `evsUS` -> `networkEvents`

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2151,6 +2159,173 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+--
+prop_connection_manager_counters :: Int
+                                 -> ArbDataFlow
+                                 -> MultiNodeScript Int TestAddr
+                                 -> Property
+prop_connection_manager_counters serverAcc (ArbDataFlow dataFlow)
+                                 script@(MultiNodeScript l) =
+  let trace = runSimTrace sim
+
+      evsCMT :: Trace (SimResult ())
```
coot, line 2182: I'd prefer use a full worlds than short names, e.g. `connectionManagerEvents`.  We tried to be quite consistent with this.  Yes, this is a bit more annoying to write, but pays off when reading the code and that's done more often than writing.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2151,6 +2159,173 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+                            UnversionedProtocol
+                            DataFlowProtocolData))
+      evsCMT = traceWithNameTraceEvents trace
+
+      -- Needed for calculating a more accurate upper bound
+      evsUS :: [ObservableNetworkState SimAddr]
+      evsUS = selectTraceEventsDynamic trace
+
+      upperBound =
+        collapseCounters True
```
coot, line 2195: I would do that where `upperBound` is used, this would make it clearer that we are collapsing on both sides of the `<=`.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2151,6 +2159,173 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+  where
+    serverAddress :: SimAddr
+    serverAddress = TestAddress 0
+
+    -- We count all connections as prunable because we do not have a better way
+    -- to know what transitions will a given connection go through. We also
+    -- count every client connection as unidirectional because they do not
+    -- negotiate duplex data flow.
+    --
+    -- We do not count Incoming/Outgoing/Closing of connections since the
```
coot, line 2230: just for the consistency sake, we call them inbound / outbound rather than incoming / outgoing.  I'd like to change this in counters too at some stage.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2151,6 +2159,173 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+    -- UniverseState is a much more reliable source of information.
+    multiNodeScriptToCounters :: DataFlow
+                              -> [ConnectionEvent Int TestAddr]
+                              -> [ObservableNetworkState SimAddr]
+                              -> ConnectionManagerCounters
+    multiNodeScriptToCounters df ces uss =
+      let ifDuplex = bool 0 1 (df == Duplex)
+          ifUni = bool 0 1 (df == Unidirectional)
+          ceCounters = foldMap' id
```
coot, line 2239: `ceCounters` -> `connectionManagerCounters`

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2151,6 +2159,173 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+                        (\ st ce -> case ce of
+                          StartClient _ ta ->
+                            Map.insert ta (ConnectionManagerCounters 1 0 1 0 0) st
+                          StartServer _ ta _ ->
+                            Map.insert ta (ConnectionManagerCounters 1 ifDuplex ifUni 0 0) st
+                          _ -> st
+                        )
+                        Map.empty
+                     $ ces
+          usCounters = foldl'
```
coot, line 2250: `usCounters` -> `networkStateCounters`

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2151,6 +2159,173 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+                          maxCounters cmc $
+                          Map.foldl'
+                           (\cmc' provenance ->
+                             cmc' <>
+                             if provenance == serverAddress
+                                then ConnectionManagerCounters 0 0 0 0 1
+                                else ConnectionManagerCounters 0 0 0 1 0
+                           )
+                           mempty
+                           conns
```
coot, line 2261: It would be good to add a note that this calculation is right for the main node, because the simulation never attempts to make other connections that to or from the main node.

coot: This also means that we can simplify the `ceCounters`, we don't need to account counters for other nodes than the main node, isn't it?

coot (changes requested):

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -207,6 +209,15 @@ data NetworkState m addr = NetworkState {
 
     }
 
+-- | Simulation accessible network environment consumed by 'simSnocket'.
+--
+newtype ObservableNetworkState addr = ObservableNetworkState {
```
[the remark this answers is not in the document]
bolt12: Ah thanks for noticing!

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
bolt12: Yes I will add a note explaining, thanks! :D

> This also means that we can simplify the ceCounters, we don't need to account counters for other nodes than the main node, isn't it?

Every `StartClient` and `StartServer` will be associated to  an Inbound or Outbound connection request to/from the main node. Since we do not know if it will perform both (making a duplex connection), `ceCounters` look at the `StartClient` and `StartServer` events and count the number of unidirectional/duplex/prunable connections assuming it reaches the last state of the state machine (`DuplexSt`). Then the `ObservableNetworkState` is used to make a more accurate guess of the actual upper bound of inbound/outbound connections.

I'll also leave a comment explaining this.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2318,13 +2514,114 @@ prop_inbound_governor_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+        . foldl'
+          (\ st ce -> case ce of
+            StartClient _ ta ->
+              Map.insertWith (<>) ta Map.empty st
+            StartServer _ ta _ ->
+              Map.insertWith (<>) ta Map.empty st
+            InboundConnection _ ta ->
+              Map.update (Just . Map.insertWith (<>) taServerAcc mempty)
+                         ta
+                         st
```
coot, line 2604: it's fine to merge them into one line

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2318,13 +2514,114 @@ prop_inbound_governor_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+                  (property $ igc <= upperBound)
+          _                               ->
+            mempty
+       )
+    $ inboundGovernorEvents
+  where
+    bundleToCounters :: Bundle [Int] -> InboundGovernorCounters
+    bundleToCounters (Bundle hot warm _) =
+      let warmRemote = bool 1 0 (null warm)
+          hotRemote  = bool 1 0 (null hot)
```
coot, line 2569: It's worth to mention that this is an estimate if there is no attenuation.

An idea: would this test pass if we know that there's no attenuation and we changed the upper bound into exact equality?

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2318,13 +2514,114 @@ prop_inbound_governor_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+       in
+        (\x ->
+          let serverAccEntry = x Map.! taServerAcc
+              mapWithoutServerAcc = Map.delete taServerAcc x
+           in Map.foldlWithKey'
+                (\igt ta entry ->
+                  case Map.lookup ta serverAccEntry of
+                    Nothing  -> igt <> foldMap' bundleToCounters entry
+                    Just bun -> igt <> foldMap' bundleToCounters entry
+                                   <> bundleToCounters bun
```
coot, line 2589: Please align the line

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2144,13 +2152,199 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+-- approach since there's no reliable way to reconstruct the value that the
+-- counters should have at a given point in time. It's not quite possible to
+-- have the god's view of the system consistent with the information that's
+-- traced because there can be timing issues where a connection is in
+-- TerminatingSt (hence counted as 0) but in the God's view the connection is
+-- still being deleted (hence counted as 1). This is all due to the not having
+-- a better way of injecting god's view traces in a way that the timing issues
+-- aren't an issue.
+--
+prop_connection_manager_counters :: Int
```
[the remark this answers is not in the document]
bolt12: Right, thanks for noticing!

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2164,12 +2358,13 @@ prop_inbound_governor_valid_transitions serverAcc (ArbDataFlow dataFlow)
                                         absBi script@(MultiNodeScript l) =
```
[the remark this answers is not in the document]
bolt12: I will!

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2318,13 +2514,114 @@ prop_inbound_governor_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+-- | Property wrapping `multinodeExperiment`.
+--
+-- Note: this test validates inbound governor counters.
+--
+prop_inbound_governor_counters :: Int
+                               -> ArbDataFlow
+                               -> MultiNodeScript Int TestAddr
+                               -> Property
+prop_inbound_governor_counters serverAcc (ArbDataFlow dataFlow)
+                               script@(MultiNodeScript l) =
```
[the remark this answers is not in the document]
bolt12: I wrote l because it is a list, I am bad with names :sweat_smile:

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
bolt12: Good question, the answer will be no, because there's a lot of action going on in the state machine and the main reason why it is hard to test the counters with exact equality is because we can not predict when a given connection is going to transition, because that logic is hidden inside the `ConnectionManager`.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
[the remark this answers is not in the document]
bolt12: I'll change this to `events`

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
coot: We could do that if we had better scheduling of events, but this is outside of this pr.

coot (approved):
Thanks @bolt12, approved.

bolt12:
bors merge

PR #3423, opened 2021-10-08
Add `GetRewardInfoPools` local state query

HeinrichApfelmus:
### Issue number

[CAD-3479](https://input-output.atlassian.net/browse/CAD-3479)

### Summary

Introduces a new local state query `GetRewardInfoPools` which provides up-to-date information the stake pool rewards and ranking.


HeinrichApfelmus:
> Isn't this one a bit redundant with `GetRewardProvenance` 🤔 ?

Unfortunately, it's not redundant enough. 😅 Specifically,

* The `GetRewardProvenance` query returns the stake distribution and pool parameters from the snapshot labeled "go". At the time of query execution, this snapshot is two epochs old.
* The new `GetRewardInfoPools` query returns the current stake distribution and pool parameters\*. To make an informed delegation choice, this information is required, as only this information pertains to the current epoch.

\* "current" pool parameters (cost, margin) include any parameter changes from the previous epoch, but not from the current epoch.

nfrisby:
Lastly (I'm sorry for the comment barrage), we just merged a PR that updates the `cardano-ledger-specs` dependency. So you'll likely need to rebase onto `master` and then you'll be able to drop at least one of your commits (the want that handles a deprecation warning from `c-l-s`).

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -361,6 +362,21 @@ pattern CardanoNodeToClientVersion7 =
       :* Nil
       )
 
+-- | The hard fork enabled, and the Shelley, Allegra, Mary and Alonzo eras enabled
+-- Using 'ShelleyNodeToClientVersion5' for the Shelley-based eras , which
+-- enables new queries.
```
dnadales, line 367: Would it make sense to refer to what these new queries are?

dnadales (approved):
LGTM!

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
HeinrichApfelmus: I figured that since comments are particularly susceptible to bitrot, it would be best if I mention the specific queries in a single place only, in this case the definition of `ShelleyNodeToClientVersion5`.

But I you prefer, I'm happy to the specific query (`GetRewardInfoPools`) at the definition of `CardanoNodeToClientVersion8` as well.

dnadales: > I figured that since comments are particularly susceptible to bitrot, it would be best if I mention the specific queries in a single place only, in this case the definition of ShelleyNodeToClientVersion5.

Sure, that makes sense. I'm ok with the comment as is.

nfrisby (approved):
@HeinrichApfelmus Please issue `bors merge` whenever you're content with the commits (eg after one final rebase, etc). Thanks!

HeinrichApfelmus:
bors merge

PR #3425, opened 2021-10-12
Diable nix-fmt for time being

EncodePanda:
All PRs are broken now because buildkite is failing on
nixpkgs-fmt. Till this gets fixed, we disabled nix-fmt.


bolt12:
I get somewhat unpredictable results in my PRs, today I got it working after a couple of force pushes, but yesterday I couldn't getting it to pass. Locally it works without any error too.

newhoggy:
Thanks @EncodePanda !

Issue #3378, opened 2021-09-22
Network simulator should allow to specify BearerInfo scripts per connection


coot:
Currently `BearerInfo`'s are specified as a list which is stepped over for all
connections.  If we do that per connection we will get better shrinking
behaviour for network simulations.

PR #3429, opened 2021-10-13
Use adequate attenuation newtypes in Server2 tests

bolt12:


ouroboros-network-framework/test/Test/Ouroboros/Network/Orphans.hs
```diff
@@ -1,14 +1,46 @@
...
+
+    shrink (Script (x :| [])) = [ Script (x' :| []) | x' <- shrink x ]
+    shrink (Script (x :| xs)) =
+        Script (x :| [])                          -- drop whole tail
+      : Script (x :| take (length xs `div` 2) xs) -- drop half the tail
+      : Script (x :| init xs)                     -- drop only last
+
+        -- drop none, shrink only elements
+      : [ Script (x' :| xs) | x'  <- shrink x ]
+     ++ [ Script (x :| xs') | xs' <- shrinkListElems shrink xs ]
```
coot, line 35: What's the advantage over `shrinkList`?

bolt12: I think I just copied this definition from the one in `ouroboros-network` didn't gave it much thought.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -1984,11 +1984,12 @@ verifyRemoteTransitionOrder (h:t) = go t h
 --
 prop_connection_manager_valid_transitions :: Int
                                           -> ArbDataFlow
-                                          -> AbsBearerInfo
+                                          -> BearerInfoScript
```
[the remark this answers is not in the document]
bolt12: Yes, good point!

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/PeerGraph.hs
```diff
@@ -33,11 +34,13 @@ import qualified Data.Tree as Tree
 
 import           Control.Monad.Class.MonadTime
 
+import           Ouroboros.Network.Testing.Data.Script
+                   (Script(..), TimedScript, ScriptDelay(NoDelay), arbitraryScriptOf)
```
coot, line 38: This line seems to long.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -204,6 +206,15 @@ data NetworkState m addr = NetworkState {
 
     }
 
+-- | Simulation accessible network environment consumed by 'simSnocket'.
+--
+data UniverseState addr = UniverseState {
+      -- | Registry of active connections and respective provider
+      --
+      usConnections       :: Map (NormalisedId addr) addr
```
coot, line 225: Can we use a newtype for the exposed connection state?  It will be easier to extend in the future.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2124,6 +2132,171 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
                        (Script (toBearerInfo absBi :| [noAttenuation]))
                        maxAcceptedConnectionsLimit l
 
+-- | Property wrapping `multinodeExperiment`.
```
coot, line 2139:
```suggestion
-- | Check connection manager counters in `multinodeExperimen`.
```

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2124,6 +2132,171 @@ prop_connection_manager_valid_transition_order serverAcc (ArbDataFlow dataFlow)
...
+--
+-- Note: this test validates connection manager counters using an upper bound
+-- approach since there's no reliable way to reconstruct the value that the
+-- counters should have at a given point in time. It's not quite possible to
+-- have the god's view of the system consistent with the information that's
+-- traced because there can be timing issues where a connection is in
+-- TerminatingSt (hence counted as 0) but in the God's view the connection is
+-- still being deleted (hence counted as 1). This is all due to the not having
```
coot, line 2147: And what about using something like: [stable](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/test/Data/Signal.hs#L196), note also that we know the `TimeWaitTimeout` value so if the connection was closed at time `t` we know that it will be forgotten at `t + timeWaitTimeout`.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2252,6 +2425,107 @@ prop_inbound_governor_valid_transition_order serverAcc (ArbDataFlow dataFlow)
                        (Script (toBearerInfo absBi :| [noAttenuation]))
                        maxAcceptedConnectionsLimit l
 
+-- | Property wrapping `multinodeExperiment`.
```
coot, line 2437: This should be more descriptive, check the other suggestion.

ouroboros-network-testing/src/Ouroboros/Network/Testing/Utils.hs
```diff
@@ -45,6 +47,67 @@ instance Arbitrary SmallDelay where
...
+
+
+-- | Like 'shrinkList' but only shrink the elems, don't drop elements.
+--
+-- Useful when you want a custom strategy for dropping elements.
+--
+shrinkListElems :: (a -> [a]) -> [a] -> [[a]]
+shrinkListElems _   []     = []
+shrinkListElems shr (x:xs) = [ x':xs | x'  <- shr x ]
+                          ++ [ x:xs' | xs' <- shrinkListElems shr xs ]
```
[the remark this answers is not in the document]
coot: I realised this comes from another place, so it's outside of this commit / PR.  Anyway, I'd be keen to rename it.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
bolt12: Ah, that could work, I'd have to try it! But I prefer to do it in a different PR.

coot: Do you already know if that was feasable or not?

bolt12: I'd have to give it a try, I wouldn't be able to retrieve all counters information from the `ObservableNetworkState` but I could probably at least test the number of incoming/outgoing connections. In either case it is preferable to keep this upper bound test too.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2447,11 +2454,12 @@ prop_inbound_governor_no_unsupported_state serverAcc (ArbDataFlow dataFlow)
...
 prop_inbound_governor_no_invalid_traces :: Int
                                         -> ArbDataFlow
-                                        -> AbsBearerInfo
+                                        -> BearerInfoScript
                                         -> MultiNodeScript Int TestAddr
                                         -> Property
 prop_inbound_governor_no_invalid_traces serverAcc (ArbDataFlow dataFlow)
-                                          absBi (MultiNodeScript events) =
+                                        (BearerInfoScript absBi)
+                                        (MultiNodeScript events) =
```
coot: Please align the arguments

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -199,13 +201,23 @@ data NetworkState m addr = NetworkState {
...
+                              (Map (NormalisedId addr) (Connection m addr)),
 
       -- | Get an unused ephemeral address.
       --
       nsNextEphemeralAddr :: AddressType -> STM m addr,
 
-      nsBearerInfo        :: LazySTM.TVar m (Script BearerInfo)
+      nsBearerInfo        :: LazySTM.TVar
+                              m
+                              (Script (LazySTM.STM m (LazySTM.TVar m (Script BearerInfo)))),
```
coot, line 214: Could you explain why we need script of scripts?

Is this #3378?  This will also not shrink well.  A better approach is to use a map `Map (ConnectionId peerAddr) (Script BearerInfo)`, this will allow to control how each connection is shrinking.  This map would be passed to `withSnocket`.  When we lookup the map and there's no script we should use `noAttenuation`.

Still I would like to hear your original idea, I am sure it is worth to consider.

bolt12: So `nsBearerInfo` is an infinite source (infinite because when it reaches the end of the script it returns always the same one) of `BearerInfo` TVars. Everytime a new connection arrives we check if an entry for it exists in the `nsAttenuationMap` if it doesn't, we reach `nsBearerInfo` for one and add an entry to the map, otherwise we just fetch the `TVar` and step the script.

bolt12: This is where everything happens.

https://github.com/input-output-hk/ouroboros-network/pull/3429/files/6bf811c01626b113f17fe2d23e24168ef72f5d0a..3d2352d22570002fca6c0da7ae13ac6a76668702#diff-9dc60e7be09dd9b4174122cea073cec55fb005ee452d64dafab6a12cfce6ed8bR746-R756

coot: I see.  This will not shrink well because if the simulation shrinks the connections it starts will start using different bearer infos anyway, but this wasn't the aim of this change, so ok.   I would suggest doing this differently though:

* keep `nsBearerInfo :: LazySTM.STM m (Script BearerInfo)`
* add `nsLastBearerInfo :: Script BearerInfo`.

Also this will be easier to translate into what I am eventually looking for: `Map ConnectionId (LazySTM.STM m (Script BearerInfo))` and a default `Script BearerInfo` for things that either don't exist in this map or reached the end of the script.

coot: btw, the link is stale, probably you forced pushed the branch after posting the link - that's a bit annoying from github.

coot: hmm, `nsAttenuationMap` is introduced below, so I am not sure why we need `nsBearerInfo` at all, just the default value for `BearerInfo` would suffice.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1038,22 +1038,29 @@ withConnectionManager ConnectionManagerArguments {
...
+                                -- 'handleError' might be either a handshake negotiation
+                                -- a protocol failure (an IO exception, a timeout or
+                                -- codec failure).  In the first case we should not reset
+                                -- the connection as this is not a protocol error.
+                                --
+                                -- If we are deleting the connState from the the state
+                                -- then connState' can be TerminatingSt in which case we
+                                -- are going to transition TerminatingSt -> TerminatedSt.
+                                -- Otherwise, Connection Manager cleanup will take care of
+                                -- tracing accordingly.
```
coot, line 1059: I am a bit confused by this explanation.  Do you actually mean that `connState` can be `TerminatingSt`?  Certainly if `connState'` is `TerminatingSt` then we're doing `someState -> TerminationSt` transition, not `TerminatingSt -> TerminatedSt`.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1038,22 +1038,29 @@ withConnectionManager ConnectionManagerArguments {
...
+                                writeTVar connVar connState'
+
+                                return (Map.delete peerAddr state , True)
+                              else return (state                     , False)
```
coot, line 1063: I think you wanted to align the `,`.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
coot: Could you create an issue for this fixed simulation bug (use the right labels, check the other such issues), please describe the test failure there.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -331,7 +344,10 @@ newNetworkState bearerInfoScript = atomically $ do
     -- nsNextEphemeralAddr
     <*> pure nextEphemeralAddr
     -- nsBearerInfo
-    <*> initScript bearerInfoScript
+    <*> LazySTM.newTVar (initScript <$> script)
+    -- attenuationMap
+    <*> newTVar Map.empty
```
coot, line 349: This means we're actually never use the map.  Instead of passing `Script (Scipt BearerInfo)` to `newNetworkState` (and `withSnocket` as well), we should pass:

* `Map (ConnectionId peerAddr) (Script BearerInfo)`
* the default value for `Script BearerInfo`

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -721,13 +737,23 @@ mkSnocket state tr = Snocket { getLocalAddr
...
               let connId = ConnectionId { localAddress, remoteAddress }
 
+              attenuationMap <- readTVar (nsAttenuationMap state)
+
+              bearerInfo <- case Map.lookup connId attenuationMap of
+                Nothing     -> do
+                  script <- stepScriptSTMTx (nsBearerInfo state)
+                  writeTVar (nsAttenuationMap state)
+                            (Map.insert connId script attenuationMap)
+                  stepScriptSTM script
```
coot, line 753: If we have an explicit default script, we can just step over it.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -721,13 +737,23 @@ mkSnocket state tr = Snocket { getLocalAddr
...
                   Nothing   -> nsNextEphemeralAddr state (getAddressType remoteAddress)
               let connId = ConnectionId { localAddress, remoteAddress }
 
+              attenuationMap <- readTVar (nsAttenuationMap state)
+
+              bearerInfo <- case Map.lookup connId attenuationMap of
+                Nothing     -> do
+                  script <- stepScriptSTMTx (nsBearerInfo state)
+                  writeTVar (nsAttenuationMap state)
+                            (Map.insert connId script attenuationMap)
```
coot, line 752: This defies the purpose of having an explicit `Map ConnectionId BearerInfo`, drawing things from a list will not improve shrinking (we're already doing that).

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
bolt12: Yes, it can be `TerminatingSt`, so we have to move the `writeTVar` inside this block to avoid tracing an out of place transition. I will add an issue for this fixed simulation bug

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
bolt12: fixed

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
bolt12: fixed

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
bolt12: fixed!

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
bolt12: fixed!

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
bolt12: fixed!

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
bolt12: fixed!

ouroboros-network-testing/src/Ouroboros/Network/Testing/Data/AbsBearerInfo.hs
```diff
@@ -241,62 +242,57 @@ instance Arbitrary AbsBearerInfo where
...
-    NonFailingBearerInfoScript (Script AbsBearerInfo)
+newtype NonFailingAbsBearerInfoScript =
+  NonFailingAbsBearerInfoScript { unNFBIScript :: Script AbsBearerInfo }
   deriving       Show via (Script AbsBearerInfo)
   deriving stock Eq
 
-toNonFailingBearerInfoScript :: BearerInfoScript -> NonFailingBearerInfoScript
-toNonFailingBearerInfoScript (BearerInfoScript script) =
-    NonFailingBearerInfoScript $ fmap unfail script
+toNonFailingAbsBearerInfoScript :: AbsBearerInfoScript -> NonFailingAbsBearerInfoScript
```
coot, line 279:
```suggestion
toNonFailingAbsBearerInfoScript :: AbsBearerInfoScript
                                -> NonFailingAbsBearerInfoScript
```

coot (commented):
Mostly syntactic changes.   I would only try to slightly simplify the `withSnocket` type signature, check the comments there.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -387,13 +400,15 @@ withSnocket
        )
     => Tracer m (WithAddr (TestAddress peerAddr)
                           (SnocketTrace m (TestAddress peerAddr)))
-    -> Script BearerInfo
+    -> BearerInfo
+    -> Map (NormalisedId (TestAddress peerAddr))
+          (LazySTM.TVar m (Script BearerInfo))
```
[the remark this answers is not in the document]
bolt12: I thought there was but there isn't, I can just update the entry. I'll update that!

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2980,13 +3116,13 @@ unit_connection_terminated_when_negotiating =
...
-          , abiOutboundWriteFailure = Just 3
-          , abiSDUSize = LargeSDU
-          }
+           { abiConnectionDelay = SmallDelay
+           , abiInboundAttenuation = NoAttenuation FastSpeed
+           , abiOutboundAttenuation = NoAttenuation FastSpeed
+           , abiInboundWriteFailure = Nothing
+           , abiOutboundWriteFailure = Just 3
+           , abiSDUSize = LargeSDU
+           }
```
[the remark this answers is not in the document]
bolt12: This is needed in order for this to type check, but this is simplified later.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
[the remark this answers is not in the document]
bolt12: Actually it is easier to keep things as they are because the alternative would require to make the whole map a TVar. Unless you prefer it

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -1358,14 +1412,17 @@ instance Arbitrary req =>
...
+            events
+            attenuationMap) =
     MultiNodePruningScript
         <$> (AcceptedConnectionsLimit
               <$> shrink hardLimit
               <*> shrink softLimit
               <*> pure delay)
         <*> (makeValid
             <$> shrinkList shrinkEvent events)
+        <*> shrink attenuationMap
```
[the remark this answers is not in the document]
bolt12: done!

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2642,7 +2747,7 @@ prop_connection_manager_pruning serverAcc
       connectionManagerEvents = withNameTraceEvents trace
 
   in tabulate "ConnectionEvents" (map showConnectionEvents events)
-    . counterexample (ppScript (MultiNodeScript events))
+    . counterexample (ppScript (MultiNodeScript events attenuationMap))
```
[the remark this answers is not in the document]
bolt12: Here I am going from a `MultiNodePruningScript` to a `MultiNodeScript` so I need the rewraps

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
bolt12: done!

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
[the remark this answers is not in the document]
coot: Are you sure:
```
f :: Map a b -> IO (Map a (TVar b))
f = traverse newTVarIO
```

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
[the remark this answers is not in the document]
coot: The white space?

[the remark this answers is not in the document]
bolt12: I thought you were referring to the whole code block, yeah the spacing is not necessary!

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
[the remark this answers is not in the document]
bolt12: Yes because `Snocket` gets passed the network state and the actions (`accept`, `connect`, etc..) all reach for the state, so we need thing in `TVar` so we can update some of the structures in place. 

I use the pure version in the generators and` traverse newTVarIO` is what I use when initializing the state: https://github.com/input-output-hk/ouroboros-network/pull/3429/files#diff-c14d3f02791c85daaf52c0549d7323e19d21e6055d7d921276460786771d9131R3104

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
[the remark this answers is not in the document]
bolt12: fixed!

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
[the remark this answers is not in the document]
coot: You just need to pass the `Map ... (TVar )`. to `accept`, `connect` etc as an argument, that's fine.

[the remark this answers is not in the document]
bolt12: I will!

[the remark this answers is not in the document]
bolt12: @coot Actually, I thought you wanted to remove the TVars from the NetworkState definition but now I realize it might be in `newNetworkState`, if so that makes more sense. The pure version already comes from the outside. In `newNetworkState` we just bundle everything together so I think it is better to just switch the type signature from that function to everything consistent.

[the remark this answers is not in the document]
bolt12: Fixed!

coot (approved):
LGTM

bolt12:
bors merge

PR #3430, opened 2021-10-13
Revert "Diable nix-fmt for time being"

EncodePanda:
This reverts commit 125f144cd5d5077421379ccb644b8b2c0c712e76.
As it turns out buildkite failure had nothing to do with
nixpkgs-fmt. After removing this script, stylish-haskell one started
failing.


EncodePanda:
Current workaround is to rerun the buildkite script few times till it works 🤷

PR #3432, opened 2021-10-14
[CAD-3383] Replace putStrLn with tracer

EncodePanda:
We have better control over where trace events are going to (console, file etc). We can also introduce a --silent option


EncodePanda:
bors merge

Issue #3436, opened 2021-10-18
Extend MonadMaskingState


coot:
'MonadMaskingState' should aslo provide:

* [x] [`interruptible`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Exception.html#v:interruptible)
* [x] [`allowInterrupt`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Exception.html#v:allowInterrupt)

PR #3437, opened 2021-10-18
MonadMaskingState

coot:
- io-sim: do not arbitrary set masking state in `Throw` or `ThrowTo`
- io-classes: `MonadMaskingState`
- io-sim: introduce various test for the masking state

Related to #3436


coot:
bors merge

Issue #3460, opened 2021-10-27
Re-order outbound governor actions


coot:
Asynchronous demotions should be done before other events.

PR #3468, opened 2021-11-01
Check for connection errors before job results.

karknu:
Leave error handling to Monitor.hs and do connection monitoring before processing job results.


karknu:
Implements #3460.

PR #3471, opened 2021-11-03
Heap profiling tools

nc6:
Some additions to db-analyser to deal with heap profiling

### DIFF
--- a/ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
+++ b/ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
@@ -1,12 +1,12 @@
-{-# LANGUAGE FlexibleInstances   #-}
-{-# LANGUAGE MultiParamTypeClasses   #-}
-{-# LANGUAGE OverloadedStrings   #-}
-{-# LANGUAGE RankNTypes          #-}
-{-# LANGUAGE RecordWildCards     #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeApplications    #-}
-{-# LANGUAGE TypeOperators       #-}
-{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE RecordWildCards       #-}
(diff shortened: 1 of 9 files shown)


ouroboros-consensus-cardano/tools/db-analyser/Main.hs
```diff
@@ -149,6 +150,11 @@ parseAnalysis = asum [
           long "count-blocks"
         , help "Count number of blocks processed"
         ]
+    , checkNoThunksParser
+    , flag' TraceLedgerProcessing $ mconcat [
+          long "trace-lgr"
```
nfrisby, line 155: Please rename to `trace-ledger`. Please also enrich the `help` message so that it indicate the following two things to some degree.

- These phases are determined by logic specific to this tool.
- They are emitted via the GHC `eventlog`.

ouroboros-consensus-cardano/tools/db-analyser/HasAnalysis.hs
```diff
@@ -3,25 +3,38 @@ module HasAnalysis (
...
+
 class (HasAnnTip blk, GetPrevHash blk) => HasAnalysis blk where
+
   countTxOutputs :: blk -> Int
   blockTxSizes   :: blk -> [SizeInBytes]
   knownEBBs      :: proxy blk -> Map (HeaderHash blk) (ChainHash blk)
 
+  -- | Emit trace markers at points in processing.
+  emitTraces     :: WithLedgerState blk -> [String]
+  emitTraces _ = []
```
nfrisby, line 36: We try to avoid defaults. I think that it just requires adding the `const []` method to the Byron instance in this case, yeah?

ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
```diff
@@ -56,6 +64,37 @@ analyseBlock f =
     p :: Proxy HasAnalysis
     p = Proxy
 
+analyseWithLedgerState ::
```
nfrisby, line 67: Please add a comment: the goal is just to apply the given analysis to the current ledger state, right?

nfrisby (approved):
Looks good to me! I made a few minor requests based on the fact that we'll probably need to, at some point, maintain/explain this code you wrote :)

Thanks! Especially happy to see the nothunks check in here.

nfrisby:
FYI @EncodePanda I approved this. Review soon if you have any objections :D

nc6:
bors merge

Issue #3466, opened 2021-10-29
Relax Win32 dependency


coot:
`ouroboros-consensus` has `Win32==2.6.1.0` constraint [set](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/ouroboros-consensus.cabal#L333).  This should be
relaxed, which will be necessary once we switch to `ghc-8.10.7`, which comes
with `Win32-2.6.2.1`.

PR #3481, opened 2021-11-05
Avoid needing to specify Win32 constraint

newhoggy:
This removes the restriction `Win32 == 2.6.1.0` when compiling for `ghc-8.10.4`.

The restriction was previously in place to prevent `Ouroboros.Consensus.Storage.Seek` from being removed.  

`Ouroboros.Consensus.Storage.Seek` implements a function that exists in `Win32-2.6.2.0` onward.  The module previously removed because of that which caused issues for downstream projects.  `ouroboros-network` must not use those functions introduced in `Win32-2.6.2.0` because that would break downstream projects that rely on `ghc` package.

This is because projects that used the `ghc` package must use the same version of `Win32` that was shipped with GHC, and for `ghc-8.10.4`, that was `Win32-2.6.1.0`.

Because upgrading to `ghc-8.10.7` is expected to be soon, I'm satisfied that replacing this constraint with a comment that says remove `Ouroboros.Consensus.Storage.Seek` after we have upgraded to `ghc-8.10.7` but not before is sufficient protection against repeating this mistake.

Doing so simplifies the `cabal` file and avoids having to carefully craft complex `if/then/else`s to allow downstream projects to upgrade to `ghc-8.10.7` on their own schedule.

After upgrading to `ghc-8.10.7` in all projects, there will no longer be any gotchas to be worried about.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.4"]
+        ghc: ["8.10.4", "8.10.7"]
         os: [ubuntu-20.04, windows-latest]
 
     steps:
@@ -22,7 +22,7 @@ jobs:
       id: setup-haskell
       with:
         ghc-version: ${{ matrix.ghc }}
-        cabal-version: 3.4.0.0
+        cabal-version: 3.6.2.0
 
     - name: Show Haskell tool versions
       run: |
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -243,7 +243,10 @@ library
                        -- Strict wrapper around SOP
                        Data.SOP.Strict
   if os(windows)
-     exposed-modules:  Ouroboros.Consensus.Storage.Seek
+     exposed-modules:
+                       -- TODO Delete this module once all our projects have upgraded to at least
+                       -- ghc-8.10.7, but not before
+                       Ouroboros.Consensus.Storage.Seek
 
   default-language:    Haskell2010
   other-extensions:
@@ -326,13 +329,7 @@ library
 
 
   if os(windows)
-    if impl(ghc >= 8.10.2) && impl(ghc < 8.11)
-      -- Fix the Win32 version to be whatever is shipped with GHC.  This is important
-      -- because we have other code that uses the GHC API which requires this specific
-      -- version of the Win32 or constraint solving will fail.
-      build-depends:   Win32 == 2.6.1.0
-    else
-      build-depends:   Win32
+     build-depends:    Win32            >= 2.6.1.0
   else
      build-depends:    unix
                      , unix-bytestring


ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -326,13 +329,7 @@ library
 
 
   if os(windows)
-    if impl(ghc >= 8.10.2) && impl(ghc < 8.11)
-      -- Fix the Win32 version to be whatever is shipped with GHC.  This is important
-      -- because we have other code that uses the GHC API which requires this specific
-      -- version of the Win32 or constraint solving will fail.
-      build-depends:   Win32 == 2.6.1.0
-    else
-      build-depends:   Win32
```
newhoggy, line 335: We no longer have these constraints because the hope is we will transition to `ghc-8.10.7` soon.

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -243,7 +243,10 @@ library
                        -- Strict wrapper around SOP
                        Data.SOP.Strict
   if os(windows)
-     exposed-modules:  Ouroboros.Consensus.Storage.Seek
+     exposed-modules:
+                       -- Delete this module once all our projects have upgraded to at least
+                       -- ghc-8.10.7
+                       Ouroboros.Consensus.Storage.Seek
```
newhoggy, line 249: Documenting when this can be removed so it doesn't accidentally get removed too early.

.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.4"]
+        ghc: ["8.10.4", "8.10.7"]
```
newhoggy, line 16: Ensure we work across these two versions until we have upgraded.

newhoggy:
I think the test failures are unrelated.

ouroboros-consensus/ouroboros-consensus.cabal
coot: This fixes #3466 :tada:

coot (commented):
LGTM, but you need to get review from the consenus team.

ouroboros-consensus/ouroboros-consensus.cabal
tmcgilchrist: Minor suggestion to tag this with `TODO` to make it more visible. My workflow is to periodically scan for TODO items in code.

newhoggy: Done!

ouroboros-consensus/ouroboros-consensus.cabal
nfrisby: > We no longer have these constraints because the hope is we will transition to ghc-8.10.7 soon.

Can you elaborate here (or better yet: in the PR description) about why this restriction was here in the first place and why it's safe to remove it now. In particular, since we were building with 8.10.4 and now we're building with 8.10.4 and 8.10.7, why is it safe to remove this now?

The existing comment mentions constraint solving and I see a cabal bump here -- did cabal just get better? Etc.

My concern is that we remove this restriction and comment and then someone experiences a constraint solving failure again and doesn't know why.

nfrisby (dismissed):
I'm only Requesting Changes because I'd like more documentation about what is happening here and why. Thanks!

ouroboros-consensus/ouroboros-consensus.cabal
newhoggy: Done!

coot: > because we have other code that uses the GHC API which requires this specific version

@nfrisby could it be because your Win32 specific stuff would fail because it's importing `Win32` unqualified and introduces something that becomes available in a later version?

nfrisby (approved):
Thanks for adding the detailed PR description. Approving.

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

newhoggy:
bors merge

PR #3482, opened 2021-11-05
Factor Signal, Script, Utils modules to ouroboros-network-testing

bolt12:


bolt12:
Should also probably move `IPv4` and `IPv6` orphan `Arbitrary` instances to `ouroboros-network-testing`

ouroboros-network-testing/src/Ouroboros/Network/Testing/Data/AbsBearerInfo.hs
```diff
@@ -241,6 +241,7 @@ instance Arbitrary AbsBearerInfo where
       | a <- shrink (abiSDUSize abi)
       ]
 
+
```
coot, line 244: This change seems unrelated to this commit (introduced in `Added github actions/nix testing`).

coot (approved):
LGTM

ouroboros-network-testing/src/Ouroboros/Network/Testing/Data/AbsBearerInfo.hs
bolt12: Yeah, I'll remove it, I added it just to trigger hydra to build and not use cache.

PR #3490, opened 2021-11-16
Make compatible with ghc-9

locallycompact:


coot:
Which version of `ghc-9` are you targeting?  `ghc-9.0` won't be adopted, and I think we will wait for a more stable `ghc-9.2` branch.

locallycompact:
I'm building for 9.2, but this change is for 9.0.

amesgen:
Superseded by #4144

PR #3495, opened 2021-11-19
Review of pruning

### LINKED ISSUE #3487: prop_never_above_hardlimit failure
The `prop_never_above_hardlimit` property fails.  A counterexample is pushed to
`coot/never-above-hardlimit-failure`.

The counter example has two outbound connections one created through
simultaneous open the other normal.  Prunning policy chooses the kill the
outbound only connection which does not make a progress twards the goal of
inbound connections, hence the failure.

coot:
Fixes #3487.

- pruning: present only inbound connections to the pruning policy
- pruning: fix typos
- pruning: include current connection in the choice map
- pruning: improved logging
- pruning: set connection state to TerminatedState
- pruning: factor out pruning
- pruning: do not prune in Duplex → InboundState transition
- pruning: improved a comment

### DIFF
--- a/ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
+++ b/ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
@@ -55,6 +55,7 @@ import qualified Data.Map.Strict as Map
 import           Data.Maybe (fromMaybe, fromJust, isJust)
 import           Data.Monoid (Sum (..))
 import           Data.Monoid.Synchronisation (FirstToFinish (..))
+import qualified Data.Set as Set
 import           Data.Typeable (Typeable)
 import           Data.Void (Void)
 import           Foreign.C.Error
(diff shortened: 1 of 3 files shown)


bolt12 (approved):
LGTM Very nice! :D

coot:
bors merge

coot:
bors cancel

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2508,6 +2513,19 @@ prop_never_above_hardlimit serverAcc
                                      )
                     . property
                     $ incomingConns cmc <= fromIntegral hardlimit
+                (TrPruneConnections prunnedSet numberToPrune choiceSet) ->
```
bolt12, line 2516: I'd prefer a separate test for this. It will make it easier to change the counters test to use the Signal API and the pruning test doesn't require it.

bolt12 (changes requested):

bolt12:
Also isn't it worth it to update the docs in this PR?

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
coot: I thought about making a separate test but I decided against it because we have to be careful for the test suite to not take too much time to run.  This test runs in about 30sec, if we duplicate it will double the time.  Still I like to keep properties that test different thing separate as much as possible, whether this will be possible is not yet clear.   Maybe, we'll need to optimise `io-sim` at some point.   As for signal api, it might not be worth it in this test, the condition is precise and clear as it stands.

coot:
> Also isn't it worth it to update the docs in this PR?

Neil is reviewing the documentation right now, so it's better not to interfere with his work.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
bolt12: Fair enough!

bolt12 (approved):

coot:
bors merge

PR #3497, opened 2021-11-19
Multinode Diffusion Simulation

bolt12:


ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,346 @@
...
+                     vectorOf,
+                     suchThat, oneof
+                   )
+
+-- | Diffusion Simulator Arguments
+--
+-- Contains all necessary randomly generated values needed to run diffusion in
+-- simulation.
+--
+type SimArgs =
```
coot, line 89: I think it it deserves a data type with proper haddocks.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,346 @@
...
+                         , MonadThrow  (STM m)
+                         , Eq (Async m Void)
+                         , forall a. Semigroup a => Semigroup (m a)
+                         )
+                      => BearerInfo
+                      -> MultiNodeScript
+                      -> m Void
+multinodeSim
+  defaultBearerInfo
+  (MultiNodeScript args) = withAsyncAll (map runNode args) $ \nodes -> do
```
coot, line 228: * It would be useful to vary when each node joins the system.
* It would be nice to verify what happens if a node is shutdown with an async exception, will it be able to rejoin the network again?

This suggests we actually need a script with a few instructions:
* `JoinNetwork`
* `Kill`
anything else that you think would be useful?

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,346 @@
...
+            , dMap
+            , rap
+            , lrp
+            , peerSelectionTargets
+            , dnsTimeout
+            , dnsLookupDelay
+            ) =
+      withSnocket nullTracer defaultBearerInfo Map.empty
+        $ \ntnSnocket _ -> withSnocket nullTracer defaultBearerInfo Map.empty
+        $ \ntcSnocket _ ->
```
coot, line 248: I wonder if this will work:
```suggestion
      withSnocket nullTracer defaultBearerInfo Map.empty
        $ \ntnSnocket _ ->
          withSnocket nullTracer defaultBearerInfo Map.empty
        $ \ntcSnocket _ ->
```
anyway splitting the line is a good idea, even if the later one needs more indentation.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,346 @@
...
+                    (byteLimitsChainSync (const 0))
+                    (timeLimitsChainSync stdChainSyncTimeout)
+                    defaultMiniProtocolsLimit
+                    (byteLimitsKeepAlive (const 0))
+                    timeLimitsKeepAlive
+                    defaultMiniProtocolsLimit
+                    (ProtocolSizeLimits (const smallByteLimit) (const 0))
+                    (ProtocolTimeLimits (const (Just 60)))
+                    defaultMiniProtocolsLimit
+                    (ProtocolSizeLimits (const (4 * 1440))
```
coot, line 288: It's good to define constants like these at the top level.  This allows to add haddocks, and easily reconfigure the nodes (easier to find), even if it is used just once.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,346 @@
...
+                               0
+                               0
+                               peerSelectionTargets
+                               (return lrp)
+                               readPRP
+                               readULA
+                               5
+                               30
+                               dnsTimeout
+                               dnsLookupDelay
```
coot, line 317: Let's use named fields, easier to figure out what is what.

Thinks like, `aKeepAliveInterval`, `aPingPongInterval`, `aProtocolTimeout`, `aTimeWaitTimeout` should be defined at the top level. 

I would rename `dnsTimeout -> dnsTimeoutScript` and `dnsLookupDelay` to `dnsLookupDelayScript`.  It's not that much more to type, and it caries the meaning what it actually is.

ouroboros-network/test/Test/Ouroboros/Network/Diffusion/Node.hs
```diff
@@ -245,8 +257,8 @@ run blockGeneratorArgs limits ni na =
 
     args :: Diff.Arguments (NtNFD m) NtNAddr (NtCFD m) NtCAddr
     args = Diff.Arguments
-      { Diff.daIPv4Address   = Just . Right . aIPv4Address $ na
-      , Diff.daIPv6Address   = Just . Right . aIPv6Address $ na
+      { Diff.daIPv4Address   = Right <$> (ntnToIPv4 . aIPAddress $ na)
```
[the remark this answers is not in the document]
bolt12: Good catch :D

ouroboros-network/test/Test/Ouroboros/Network/Diffusion/Node/NodeKernel.hs
```diff
@@ -72,6 +75,18 @@ data NtNAddr_
   | IPAddr IP.IP PortNumber
   deriving (Eq, Ord)
 
+instance Arbitrary NtNAddr_ where
+  arbitrary = do
+    -- TODO: Move this IP generator to ouroboros-network-testing
```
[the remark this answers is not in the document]
bolt12: Yes, we should!

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,346 @@
...
+  , PeerSelectionTargets
+  , Script DNSTimeout
+  , Script DNSLookupDelay
+  )
+
+-- | Multinode Diffusion Simulator Script
+--
+-- List of 'SimArgs'. Each element of the list represents one running node.
+--
+newtype MultiNodeScript = MultiNodeScript
```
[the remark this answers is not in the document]
bolt12: Sounds good!

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Will do!

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: > It would be useful to vary when each node joins the system.

You mean a delay ?

> It would be nice to verify what happens if a node is shutdown with an async exception, will it be able to rejoin the network again?

I agree I just decided to come up with simplest setting possible.

> This suggests we actually need a script with a few instructions:
>
>  -  JoinNetwork
>  -  Kill
>
>    anything else that you think would be useful?

I don't think there's much more that can happen on a real scenario, right?

coot: Yes a startup delay.  It seems we can only  came up with the two :).  But one could be able to chain them `[JoinNetwork, Kill, ... ]`.  The generators written by Ulf are a good example hot to generate stateful instructions.  Also note that we don't need many of them, a list of 2 / 3 on average, 5-8 at max should be ok.  At the first iteration we rather want to explore the case you were thinking about: all nodes are up and running, and maybe some times some of them go down and re-join.  But later this will allow us to extend the generator to include situations like: all nodes going down at the same time, and stay down for some time, rejoin simultaneously.

There's another aspect of this, and needs explanation.  We can simulate network errors through attenuation.  I argue that network attenuation never bring a node down, so there's no overlap, making this quite a bit simpler.  Well, unless we inject an `accept` error some of them can bring the node down, I think that's the only case (do you agree?).  If so we could ignore this (but it should be described, that the tests are designed with that in mind).   The problem with this is that we'd have two ways of bringing the node down: `Kill` and an accept exception.  This means we cannot relay on the interval between a node being killed and re-joining the network, as it could be in that state for longer.  This seems like a minor caveat, since probably none of ours properties will relay on this information (am I right?).

bolt12: So if I understood correctly, we want that each node have a script with instructions + some delay (in the same way `Server2` `MultiNodeScript` does). 

And regarding your point in the second paragraph, you are saying to ignore `accept` errors since we do not/won't inject any? And that we can not write tests relying on just `Kill` command alone because it could have been killed for longer due to a possible `accept` exception?

coot: > So if I understood correctly, we want that each node have a script with instructions + some delay (in the same way Server2 MultiNodeScript does).
Yes, that's right.

> And regarding your point in the second paragraph, you are saying to ignore accept errors since we do not/won't inject any? And that we can not write tests relying on just Kill command alone because it could have been killed for longer due to a possible accept exception?

My original thought was that we at least we don't inject errors that would kill the node, because this makes it harder to make a script which `Kill` will actually kill a node.  But, giving it another thought, I think we don't need to worry about it at all, `Kill` if a  nodes does not work would be no-op, and it would not affect `Start` (for which we need to track if the node is not running, but if it stopped earlier due to some exception, that's ok).

One additional useful command would be `Reconfigure` which reconfigures the local / public root peers.

bolt12: Addressed

ouroboros-network/test/Test/Ouroboros/Network/Diffusion/Node/NodeKernel.hs
```diff
@@ -63,6 +63,9 @@ import qualified Ouroboros.Network.Testing.ConcreteBlock as ConcreteBlock
 import           Simulation.Network.Snocket (AddressType (..),
                      GlobalAddressScheme (..))
 
+import           Test.QuickCheck (Arbitrary(..), oneof, chooseInt, choose)
+import Data.IP (IP(..), toIPv4, toIPv6)
```
coot, line 67: Could you run `scripts/ci/check-stylish-network.sh` locally (it's added back to ci in #3558).

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,446 @@
...
+      -- ^ 'Arguments' 'aDNSTimeoutScript' value
+    , saDNSLookupDelayScript :: Script DNSLookupDelay
+      -- ^ 'Arguments' 'aDNSLookupDelayScript' value
+    }
+    deriving (Show)
+
+data Commands = JoinNetwork DiffTime
+               | Kill DiffTime
+               | Reconfigure DiffTime
+                             [(Int, Map RelayAccessPoint PeerAdvertise)]
```
coot, line 127:
```suggestion
data Commands = JoinNetwork DiffTime
              | Kill DiffTime
              | Reconfigure DiffTime
                            [(Int, Map RelayAccessPoint PeerAdvertise)]
```

coot: I would rename the type to `Command`; then `[Command]` looks like a list of commands, rather than `[Commands]` (a list of list of commands).

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,446 @@
...
+
+  where
+    genIP :: Gen IP
+    genIP =
+      let genIPv4 = IPv4 . toIPv4 <$> replicateM 4 (choose (0,255))
+          genIPv6 = IPv6 . toIPv6 <$> replicateM 8 (choose (0,0xffff))
+       in oneof [genIPv4, genIPv6]
+
+genCommands :: [(Int, Map RelayAccessPoint PeerAdvertise)] -> Gen [Commands]
+genCommands raps = sized $ \size -> do
```
coot, line 149: This will only generate a list which starts with `JoinNetwork` and continues with one of `Reconfigure` or `Kill`, and a kill can be followed by a single `JoinNetwork`.  It would be good to cover `JoinNetwork` after being `Kill`-ed in a more general way.

If you want to look at inspiration how this can be done more generally, take a look at
 [Test.Ouroboros.Network.Server2](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs?plain=1#L1235) - which is a statefull generator.  Here we'd need to know if a node is running or not. (I don't meant it must be done this way).
 
We don't include here attenuations which kill the node at specific times, e.g. an accept error, but in the future we could extend this generator to also generate attenuation as well.  Right now, with decoupled node state and its attenuation, we will need to be lucky to have `Kill` followed by `JoinNetwork` to restart after an attenuation error.  For this reason it's better to have slightly more nodes.

It would be nice to have some parameter which allows us to configure how often a node is restarted and have nodes which are restarted often and some others that are restarted more rarely (and thus are more stable).

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,446 @@
...
+  -> m Void
+diffusion_simulation
+  defaultBearerInfo
+  (DiffusionScript args) = withAsyncAll (map (uncurry (runCommands Nothing)) args) $ \nodes -> do
+    (_, x) <- waitAny nodes
+    return x
+  where
+    runCommands
+      :: Maybe (Async m Void
+               , StrictTVar m [(Int, Map RelayAccessPoint PeerAdvertise)])
```
coot, line 285:
```suggestion
      :: Maybe ( Async m Void
               , StrictTVar m [(Int, Map RelayAccessPoint PeerAdvertise)] )
```

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,446 @@
...
+      -> SimArgs
+      -> [Commands]
+      -> m Void
+    runCommands Nothing simArgs [] = do
+      threadDelay 3600
+      runCommands Nothing simArgs []
```
coot, line 291: The outbound governor simulations are written in never terminating style.  The drawback of it is that one has to be careful when printing simulation trace.  I don't object, it's just good to remember.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,446 @@
...
+    runCommands (Just (async, _)) _ [] =
+      wait async
```
coot, line 293: So the simulation can actually terminate.  I think we should either continue, or also return in the other case.  So we're consistent.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,446 @@
...
+  => BearerInfo
+  -> DiffusionScript
+  -> m Void
+diffusion_simulation
+  defaultBearerInfo
+  (DiffusionScript args) = withAsyncAll (map (uncurry (runCommands Nothing)) args) $ \nodes -> do
+    (_, x) <- waitAny nodes
+    return x
+  where
+    runCommands
```
coot, line 283: Add haddock; it would be good to explain that this function runs a single node (am I right?).

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,446 @@
...
+      runCommands Nothing simArgs []
+    runCommands (Just (async, _)) _ [] =
+      wait async
+    runCommands Nothing simArgs (JoinNetwork delay:cs) = do
+      threadDelay delay
+      lrpVar <- newTVarIO $ saLocalRootPeers simArgs
+      withAsync (runNode simArgs lrpVar) $ \nodeAsync ->
+        runCommands (Just (nodeAsync, lrpVar)) simArgs cs
+    runCommands _ _ (JoinNetwork _:_) =
+      error "Impossible happened"
```
coot, line 300:
```suggestion
      error "runCommands: Impossible happened"
```
otherwise it will be impossible to find which `Impossible happened` is it :grin:.  This also applies to other impossible `error`s below.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,446 @@
...
+    runCommands (Just (async, _)) simArgs (Kill delay:cs) = do
+      threadDelay delay
+      cancel async
```
coot, line 303: Yes we want `cancel` so we're sure that when we resume recursive call that the node isn't running.

coot (commented):
Good step towards PR being merged, some remarks follow.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Good point!

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: This comment runs the Script for a given node, yes. So it can either run a single node, or kill it or change its configuration.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Of course :D

bolt12: Addressed!

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Addressed

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Addressed

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: There's a different way I could do this that is by generating an arbitrary list of commands and then fixing it up. We shouldn't have 2 consecutive `Kill` or 2 consecutive `JoinNetwork`, and after a `Kill` the only that makes sense to follow it is `JoinNetwork`. So something like this:

```haskell
fixupCommands :: [Command] -> [Command]
fixupCommands [] = []
fixupCommands (h:t) = go h t
  where
    go :: Command -> [Command] -> [Command]
    go _ [] = []
    go prev (cmd:cmds) =
      case (prev, cmd) of
        (JoinNetwork _, JoinNetwork dt) ->
          JoinNetwork dt : go cmd cmds
        (Kill _, Kill dt) ->
          Kill dt : go cmd cmds
        (Kill dt, Reconfigure _ _) ->
          Kill dt : go prev cmds
        (Reconfigure dt lrp, JoinNetwork _) ->
          Reconfigure dt lrp : go prev cmds
        _ -> prev : go cmd cmds

```

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Addressed

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
coot: ^ it does not guarantee that `Kill` or `Reconfigure` runs when a node is running.

io-classes/src/Control/Monad/Class/MonadAsync.hs
```diff
@@ -203,9 +203,7 @@ newtype Concurrently m a = Concurrently { runConcurrently :: m a }
 instance Functor m => Functor (Concurrently m) where
     fmap f (Concurrently ma) = Concurrently (fmap f ma)
 
-instance ( Applicative m
-         , MonadAsync m
-         ) => Applicative (Concurrently m) where
+instance (MonadAsync m) => Applicative (Concurrently m) where
```
coot, line 206:
```suggestion
instance MonadAsync m => Applicative (Concurrently m) where
```

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,443 @@
...
+          case (prev, cmd) of
+            (JoinNetwork _, JoinNetwork dt) ->
+              JoinNetwork dt : go cmd cmds
+            (Kill _, Kill dt) ->
+              Kill dt : go cmd cmds
+            (Kill dt, Reconfigure _ _) ->
+              Kill dt : go prev cmds
+            (Reconfigure dt lrp, JoinNetwork _) ->
+              Reconfigure dt lrp : go prev cmds
+            _ -> prev : go cmd cmds
```
coot, line 161: The drawback of this approach is that the generated list of commands usually will be much shorter than `size`; this can be addressed by rescaling the generator to something larger than the default `100`.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Doesn't it? I'd say otherwise but might be overlooking something, can you point me to it?

coot: e.g. `[Kill]` or `[Kill, Reconfigure, Kill, Reconfigure]` are both fixpoints of `fixupCommands`; I think you need to track state if a node is running or not. There might be an arbitrary long strides of `Reconfigure` until a node is `Kill`-ed when it must be restarted.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: That's true! Should I rescale it in the generator?

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Oh right! I think singleton lists are the only thing I missed, I should add a clause:

```haskell
    go _ [] = []
    go (JoinNetwork dt) [] = [JoinNetwork dt]
    ...
```

I think this ought to fix the problem. For other singleton lists I will ignore them and return `[]`. I dont think I need to keep track of state for `fixupCommands`, I can just assume the node is running at the current recursion point and decide what to do with whatever 2 consecutive commands are there next. So for example I can not have 2 `Kills` consecutively and also a `Kill` after a `Reconfigure` doesn't make much sense to. 

> There might be an arbitrary long strides of Reconfigure until a node is Kill-ed when it must be restarted.

Are you saying after a `Reconfigure` I need to restart the node? Because if that wouldn't really be necessary, the node handles that by itself.

coot: I am convinced. 

> Are you saying after a Reconfigure I need to restart the node? Because if that wouldn't really be necessary, the node handles that by itself

no.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
coot: The question is how much...

bolt12: It is going to depend on how much `fixupCommands` shrinks the initial list, I'll investigate this.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Addressed

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: From what I have seen it does not reduce the list that much, might be proven wrong when deploying more property tests and checking coverage. But my idea is that we do not want much variability in the commands, just enough to test all scenarios.

coot: ok, let's see how it goes.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
coot: This is not addressed.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -124,12 +124,10 @@ import           Ouroboros.Network.Testing.Data.AbsBearerInfo
 import           Ouroboros.Network.Testing.Utils (genDelayWithPrecision)
 
 import           Test.Ouroboros.Network.ConnectionManager
-                     (verifyAbstractTransition)
+                     (allValidTransitionsNames, validTransitionMap,
+                     verifyAbstractTransition)
```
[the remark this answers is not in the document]
bolt12: It doesn't I ran stylish-haskell before force pushing, actually it was stylish-haskell that formatted this like this.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -0,0 +1,501 @@
...
+                         <$> zip (dMap Map.! d) (repeat p)
+                         )
+                       | DomainAccessPoint d p <- domainsAP \\ daps
+                       , Map.member d dMap
+                       ]
+      Map.fromList mapDomains
+
+
+ntnToPeerAddr :: IP -> PortNumber -> NtNAddr
+ntnToPeerAddr a b = TestAddress (Node.IPAddr a b)
```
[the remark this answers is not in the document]
bolt12: Yeah, although I don't find the n-ary compose operators very readable.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Sorry I thought I had addressed it.

bolt12:
bors merge

PR #3500, opened 2021-11-22
Use block download when ranking peers

karknu:
Use both who ever gave us a header first, and who ever provided us with
the block first when ranking peers in deadline mode. For larger blocks
forged far away it is likely that we learned of the header
from a node on another continent, but the node from which we first
downloaded the block is much closer. This change gives both nodes
credit.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerMetric.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerMetric.hs
@@ -145,11 +145,11 @@ upstreamyness = Pq.fold' count Map.empty
 
 -- Returns a Map which counts the number of bytes downloaded
 -- for a given peer.
-fetchyness
+fetchynessBytes
     :: forall p.  ( Ord p )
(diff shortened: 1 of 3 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerMetric.hs
```diff
@@ -163,5 +163,24 @@ fetchyness = Pq.fold' count Map.empty
         fn Nothing = Just $ fromIntegral bytes
         fn (Just oldBytes) = Just $! oldBytes + fromIntegral bytes
 
+-- Returns a Map which counts the number of times a given peer
+-- was the first we downloaded a block from.
+fetchyness'
```
coot, line 168: Could you rename  `fetchyness'` maybe as `fetchynessBlockCount` and also rename `fetchyness` as `fetchynessBytes`.  Or something like that.

coot (approved):
LGTM, just a minor comment about better names.

PR #3502, opened 2021-11-24
Lazy traces

coot:
Traces can be lazy, they are either thrown away or transform into some
representation (text, json, etc).

- lazy traces
- Removed Counter type


coot:
I'll resubmit the pr, to check jira integration.

PR #3503, opened 2021-11-24
CAD-3722 Lazy traces

coot:
Trace types should be lazy, they are either discarded or converted to
some representation (text, json, etc) and never stored.

- lazy traces
- Removed Counter type


karknu (approved):
LGTM

coot:
bors merge

PR #3513, opened 2021-11-30
Move TPraos protocol into its own package.

nc6:
Intention will be to implement the new Praos protocol also in this package.

This PR simply moves things around. I didn't move the PBFT protocol into this package, since it seemed reasonable to just isolate it in Byron.

### DIFF
--- a/ouroboros-consensus-protocol/ouroboros-consensus-protocol.cabal
+++ b/ouroboros-consensus-protocol/ouroboros-consensus-protocol.cabal
@@ -1,0 +1,51 @@
+cabal-version:      3.0
+name:               ouroboros-consensus-protocol
+version:            0.1.0.0
+synopsis:           Cardano consensus protocols.
+author:             IOHK Formal methods team
+maintainer:         nick@topos.org.uk
+
+source-repository-head
+    type: git
+    location: https://github.com/input-output-hk/ouroboros-network
+    subdir: ouroboros-consensus-protocol
+
+flag asserts
+  description: Enable assertions
+  manual:      False
+  default:     False
+
+library
+  hs-source-dirs: src
+  default-language:    Haskell2010
+  exposed-modules:
+    Ouroboros.Consensus.Protocol.Ledger.HotKey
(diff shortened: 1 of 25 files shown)


ouroboros-consensus-protocol/ouroboros-consensus-protocol.cabal
```diff
@@ -0,0 +1,34 @@
...
+    cardano-ledger-shelley,
+    cardano-protocol-tpraos,
+    cardano-slotting,
+    cborg,
+    containers,
+    mtl,
+    nothunks,
+    ouroboros-consensus,
+    serialise,
+    text
```
nfrisby, line 34: I suspect we also want to copy over the rest of the litany, especially `ghc-options`.

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs
```diff
@@ -8,7 +8,7 @@
 -- | Hot key
 --
 -- Intended for qualified import
-module Ouroboros.Consensus.Shelley.Protocol.HotKey (
+module Ouroboros.Consensus.Protocol.Ledger.HotKey (
```
nfrisby, line 11: Thinking out loud: is the hotkey more a `Ledger` concern or more a `BlockForging` concern?

Actually, I'm less sure how to distinguish which parts of this are unique to `BlockForging` than I had thought I was. EG only a block-producing node would ever use `evolve`, `sign`, and `mkHotKey`, if I understand correctly.

I suppose it's natural to leave this module unbroken, and so `Ledger` really is the most appropriate point in the conceptual hierarchy for this bag of definitions.

nfrisby (commented):
Approved. Thanks!

The only key request is to copy over the `ghc-options` block.

nfrisby (approved):
Approved! Sorry -- I meant to Approve instead of just Comment last time.

nc6:
bors merge

nc6:
bors retry

PR #3514, opened 2021-11-30
Fix clean state initialization

Jasagredo:
Due to the unmasking of exceptions during ChainDB initialization, it is the case right now that the exit code is different from `DatabaseCorruption` and therefore the `withCheckedDb` function considers that it has to create the clean file. Because of this, a Ctrl+C during initialization would skip the validation on the next startup.

This PR ensures that the clean file is only removed after the chainDB is initialized (and therefore potentially modified) and from that point on it can be created again.

| initial status | killed when           | on `SIGKILL`     | on `DatabaseCorruption` | on other exception (e.g. `^C`) |
|----------------|-----------------------|------------------|------------------------|---------------------------|
| clean          | during initialization | unclean → clean | clean → unclean    | clean                     |
| clean          | after initialization  | unclean          | unclean                | clean                     |
| unclean        | during initialization | unclean          | unclean                | clean → unclean          |
| unclean        | after initialization  | unclean          | unclean                | clean                     |

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -52,7 +52,6 @@ module Ouroboros.Consensus.Node (
   ) where
 
 import           Codec.Serialise (DeserialiseFailure)
-import           Control.Monad (when)
 import           Control.Tracer (Tracer, contramap)
 import           Data.ByteString.Lazy (ByteString)
 import           Data.Hashable (Hashable)
@@ -183,8 +182,14 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
                          (p2p :: Diffusion.P2P) =
    LowLevelRunNodeArgs {
 
-      -- | How to manage the clean-shutdown marker on disk
-      llrnWithCheckedDB :: forall a. (LastShutDownWasClean -> m a) -> m a
+      -- | An action that will receive a marker indicating whether the previous
+      -- shutdown was considered clean and a wrapper for installing a handler to
+      -- create a clean file on exit if needed. See
+      -- 'Ouroboros.Consensus.Node.Recovery.runWithCheckedDB'.
+      llrnWithCheckedDB :: forall a. (  LastShutDownWasClean
+                                     -> (ChainDB m blk -> m a -> m a)
+                                     -> m a)
+                        -> m a
 
       -- | The " static " ChainDB arguments
     , llrnChainDbArgsDefaults :: ChainDbArgs Defaults m blk
@@ -277,9 +282,8 @@ runWith :: forall m addrNTN addrNTC versionDataNTN versionDataNTC blk p2p.
   -> m ()
 runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
 
-    llrnWithCheckedDB $ \(LastShutDownWasClean lastShutDownWasClean) ->
+    llrnWithCheckedDB $ \(LastShutDownWasClean lastShutDownWasClean) continueWithCleanChainDB ->
     withRegistry $ \registry -> do
-
       let systemStart :: SystemStart
           systemStart = getSystemStart (configBlock cfg)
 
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
```diff
@@ -37,11 +37,14 @@ createMarkerOnCleanShutdown ::
      (IOLike m, StandardHash blk, Typeable blk)
   => Proxy blk
   -> HasFS m h
+  -> StrictTVar m Bool
   -> m a  -- ^ Action to run
   -> m a
-createMarkerOnCleanShutdown pb mp = onExceptionIf
+createMarkerOnCleanShutdown pb mp shouldCreate = onExceptionIf
     (not . exceptionRequiresRecovery pb)
```
dnadales, line 44: So the problem that we had is that the `SIGINT` signal resulted in an exception that did not require recovery, is that correct? 

I wonder if one could translate this exception into one that requires recovery when we're in the initialization phase still and re-validating the chunk files.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -352,6 +355,9 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
                                     nodeKernel
                                     peerMetrics
 
+      atomically $ modifyTVar shouldCreateCleanMarker (const True)
+      cleanRemover
```
dnadales, line 359: What is the `cleanRemover` for, and how does it interact with `shouldCreateCleanMarker`?

dnadales: Why was this not needed before? I mean, why didn't we need to create the clean marker after successfully opening the chain DB?

dnadales (commented):
LGTM, but it'd like to go through this over a call.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -184,7 +184,10 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
    LowLevelRunNodeArgs {
 
       -- | How to manage the clean-shutdown marker on disk
-      llrnWithCheckedDB :: forall a. (LastShutDownWasClean -> m a) -> m a
+      llrnWithCheckedDB :: forall a. (   LastShutDownWasClean
```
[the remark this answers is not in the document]
Jasagredo: Yes probably this deserves a datatype. The TVar means whether it should create a new `clean` file in presence of a non-corrupting exception. The function is the one to remove the old `clean` file.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
Jasagredo: The TVar means whether it should create a new clean file in presence of a non-corrupting exception. The function is the one to remove the old clean file.

It is still created on exit by `withCheckedDb`. We are now delaying the deletion of the previous one until we opened the DB. Notice that I captured the IO action for deleting it into the `cleanRemover`

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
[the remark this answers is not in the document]
nfrisby: Yeah, this comment was terse. We could elaborate it to, eg:

```
-- | A bracket function that manages the clean-shutdown marker on disk.
--
-- o If the marker is missing on startup, then ChainDB initialization will revalidate the database contents.
--
-- o If the OS kills the nodes, then we don't even have the opportunity to write out the marker file, which is appropriate, since we want the next startup to do revalidation. (TODO But need we revalidate _everything_ in this case?)
--
-- o If initialization was cleanly interrupted (eg SIGINT), then we leave the marker how we found it upon this starting of the node.
--
-- o At the end of a successful initialization, we remove the marker and install a shutdown handler that writes the marker except for certain exceptions (see 'exceptionRequiresRecovery') that suggest corruption, for which we want the next startup to do revalidation.
```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
nfrisby: In both the old code and the new, the marker is created by the `bracket`-like closing part of `llrnWithCheckedDB`. In Javier's PR here, the new bit is the reference which might prevent the marker from being created on a clean shutdown.

nfrisby (changes requested):
I agree that this does what we want, but I think there's a more suitable implementation; one that doesn't use a mutvar "unnecessarily".

I'm thinking something like:

```
llrnWithCheckedDB :: forall a. (LastShutDownWasClean -> m (a, ThisShutDownWasClean)) -> m a
```

Or maybe something like

```
llrnWithCheckedDB :: forall a. (LastShutDownWasClean -> (forall b. m b -> m b) -> m a) -> m a
```

Where the passes wrapper is applied to the continuation of initialization, and it's the part that removes the marker (if present) and installs the handler that writes it.

Let's chat about it.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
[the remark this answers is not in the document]
Jasagredo: Added this documentation

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -183,8 +182,26 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
...
-      -- | How to manage the clean-shutdown marker on disk
-      llrnWithCheckedDB :: forall a. (LastShutDownWasClean -> m a) -> m a
+      -- | A bracket function that manages the clean-shutdown marker on disk.
+      --
+      -- - If the marker is missing on startup, then ChainDB initialization will
+      --   revalidate the database contents.
+      --
+      -- - If the OS kills the nodes, then we don't even have the opportunity to
+      --   write out the marker file, which is appropriate, since we want the
+      --   next startup to do revalidation.
```
dnadales, line 192: Is "even" necessary? Also, I don't know if "appropriate" is the right word here. I don't have a good alternative, I'd just say:

```
which is fine, since ...
```

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -183,8 +182,26 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
...
+      --
+      -- - If initialization was cleanly interrupted (eg SIGINT), then we leave
+      --   the marker how we found it upon this starting of the node.
```
dnadales, line 195:
```suggestion
      --   the marker in the same state as it was at the beginning of said initialization.
```

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
```diff
@@ -83,6 +71,67 @@ exceptionRequiresRecovery pb e = case toExitReason pb e of
     DatabaseCorruption -> True
     _                  -> False
 
+-- | A bracket function that manages the clean-shutdown marker on disk.
```
dnadales, line 74: Oh, wait, we have duplicated comments. Why not keeping this, and explaining what are the 

```
LastShutDownWasClean
```
and 

```
ChainDB m blk -> m a -> m a
```

parameters in the `llrnWithCheckedDB` constructor?

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
Jasagredo: Done (in the other comment as suggested) and reworded this one.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
```diff
@@ -183,8 +182,26 @@ data LowLevelRunNodeArgs m addrNTN addrNTC versionDataNTN versionDataNTC blk
...
+      --
+      -- - At the end of a successful initialization, we remove the marker and
+      --   install a shutdown handler that writes the marker except for certain
+      --   exceptions (see 'exceptionRequiresRecovery') that suggest corruption,
```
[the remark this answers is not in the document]
Jasagredo: Done (in the other comment as suggested) and reworded this one.

ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
Jasagredo: Done (in the other comment as suggested) and reworded this one.

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
Jasagredo: Done

ouroboros-consensus/src/Ouroboros/Consensus/Node/Recovery.hs
Jasagredo: I think from the discussion that this is not needed in the end with the new mechanism

dnadales (approved):

nfrisby:
I amended your commit with a tiny change, just renaming `wrapper`:

```
$ git show
commit ec01f675b0fe647b04e55d6c246d7bfe60bbc18b (HEAD -> jasagredo/fix-clean-state-initialization)
Author: Nicolas Frisby <nick.frisby@iohk.io>
Date:   Thu Dec 2 10:13:48 2021 -0800

    TOSQUASH: rename wrapper

diff --git a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
index f530ebe37..7c655a3bb 100644
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node.hs
@@ -282,7 +282,7 @@ runWith :: forall m addrNTN addrNTC versionDataNTN versionDataNTC blk p2p.
   -> m ()
 runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
 
-    llrnWithCheckedDB $ \(LastShutDownWasClean lastShutDownWasClean) wrapper ->
+    llrnWithCheckedDB $ \(LastShutDownWasClean lastShutDownWasClean) continueWithCleanChainDB ->
     withRegistry $ \registry -> do
       let systemStart :: SystemStart
           systemStart = getSystemStart (configBlock cfg)
@@ -315,7 +315,7 @@ runWith RunNodeArgs{..} LowLevelRunNodeArgs{..} =
       chainDB <- openChainDB registry inFuture cfg initLedger
                 llrnChainDbArgsDefaults customiseChainDbArgs'
 
-      wrapper chainDB $ do
+      continueWithCleanChainDB chainDB $ do
         btime <-
           hardForkBlockchainTime $
           llrnCustomiseHardForkBlockchainTimeArgs $
```

I squashed that into your commit and force-pushed it back up.

nfrisby (approved):
Looks great!

nfrisby:
bors merge

PR #3518, opened 2021-12-01
[CAD-3723]  Add progress to initial chain selection

EncodePanda:


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -516,21 +526,22 @@ instance ApplyBlock l blk => ApplyBlock (LedgerDB l) blk where
 -------------------------------------------------------------------------------}
 
 pureBlock :: blk -> Ap m l blk ()
-pureBlock = ReapplyVal
+pureBlock blk = ReapplyVal blk
```
Jasagredo, line 529: This seems to be a leftover

Jasagredo (approved):
Looks super to me. :ok_hand:

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -429,26 +431,34 @@ ledgerDbPush cfg ap db =
...
 ledgerDbPushMany :: (ApplyBlock l blk, Monad m, c)
-                 => LedgerDbCfg l
+                 => (UpdateLedgerDbTraceEvent blk -> m ())
+                 -> LedgerDbCfg l
                  -> [Ap m l blk c] -> LedgerDB l -> m (LedgerDB l)
-ledgerDbPushMany = repeatedlyM . ledgerDbPush
+ledgerDbPushMany trace = repeatedlyM . pushAndTrace
+  where
+    pushAndTrace cfg ap db = do
+      trace StartedPushingBlockToTheLedgerDb
```
dnadales, line 440: Do we trace somewhere else some details about the block we're validating? If not we might want to add these details here. 

Also, I don't know if we should state we're pushing a block to the ledger DB, since that depends upon the block being valid.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -429,26 +431,34 @@ ledgerDbPush cfg ap db =
...
+      res <- ledgerDbPush cfg ap db
+      trace PushedBlockToTheLedgerDb
```
dnadales, line 442: I wonder if we need this event: if we don't see the next `StartedPushingBlockToTheLedgerDb` we can assume the node is still busy validating the block logged in the current `StartedPushingBlockToTheLedgerDb` event.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -429,26 +431,34 @@ ledgerDbPush cfg ap db =
...
+      return res
 
 -- | Switch to a fork
 ledgerDbSwitch :: (ApplyBlock l blk, Monad m, c)
                => LedgerDbCfg l
                -> Word64          -- ^ How many blocks to roll back
+               -> (UpdateLedgerDbTraceEvent blk -> m ())
```
dnadales, line 449: In principle, this could be any action. I wonder if we want to be more specific and declare that this should actually be a tracer.

dnadales (commented):
I left some comments. We can have a call to discuss them.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda: > Do we trace somewhere else some details about the block we're validating? If not we might want to add these details here.

What we have here is just `ap`, to get more details we would have to call `resolveBlock r` for some of the `ap`s. This means we would have to do it twice (once for tracing and once in `applyBlock`). 
Alternative would be to push the tracing into `Ap` (we've went that path in this PR as you recall). Or pushing the `trace` function down to `applyBlock`. But that means that `trace` would change the signature of `ledgerDbPush`, a function that is used in other places

EncodePanda: > Also, I don't know if we should state we're pushing a block to the ledger DB, since that depends upon the block being valid.

Function is called `ledgerDBPush` (not `tryLedgerDBPush`) which got me confused, but I see that we throw an error for some data constructors of `Ap`. We can:
a) rename it, though I can not think of a name that will be a right one
b) push the trace to `applyBlock` and fire events accordingly

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda: I thought that as well, but the output on logs got user confusing in the past. For example in different place there was an event `OpenedImmutableDB`, only we knew that it means that immediately we start opening `VolatileDB` after that. And if there was no feedback, they would get confused or anxious. I think tat `Starting/Done` pair of events was requested recently by Serge, right?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda: Good question, but at least debatable. Even though I follow your line of thought and you are right, in principle this could be any hook but:
a) this could be a slipper slope to messy code in a future, because it is so easy to just "add one more thing" to a hook.
b) I've followed existing convention in our code were we pass `trace` functions as arguments

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda: thx! fixed

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda: Decisons during PP:
1. we will add `RealPoint blk` to the `StartedPushingBlockToTheLedgerDb` event
2. we will remove the `PushedBlockToTheLedgerDb` event
3. we will add progress to the `UpdateLedgerDbTraceEvent` so that we can track progress in cardano-node

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda: answered in previous comment

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda:
```
data ChainSelEnv m blk = ChainSelEnv
    { lgrDB             :: LgrDB m blk
    , trace             :: TraceValidationEvent blk -> m ()
 ```
`ChainSelEnv` is using `trace` and that pollutes our design. I think I will change that to provide a `tracer` instead but in a follow up PR.

Comment (nfrisby) [reply ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs:455]: I think it should be `Tracer`. Tracer is isomorphic to that type, and transparent, so there's no restriction gained by using either type. But `Tracer` is the tell-tale type throughout our codebase for this kind of concern.

Comment (EncodePanda) [reply ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs:461]: @nfrisby are you ok with doing this in separate PR so this one does not get polluted with to many concerns?

File: ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs [hunk 5]
```diff
@@ -437,16 +437,22 @@ ledgerDbPush cfg ap db =
...
-                 => (UpdateLedgerDbTraceEvent blk -> m ())
-                 -> LedgerDbCfg l
-                 -> [Ap m l blk c] -> LedgerDB l -> m (LedgerDB l)
-ledgerDbPushMany trace = repeatedlyM . pushAndTrace
+ledgerDbPushMany ::
+     forall m c l blk . (ApplyBlock l blk, Monad m, c)
+  => (UpdateLedgerDbTraceEvent blk -> m ())
+  -> LedgerDbCfg l
+  -> [Ap m l blk c] -> LedgerDB l -> m (LedgerDB l)
+ledgerDbPushMany trace cfg aps initDb = (repeatedlyM pushAndTrace) aps initDb
```
dnadales, line 445:
```suggestion
ledgerDbPushMany trace cfg aps initDb = repeatedlyM pushAndTrace $ aps initDb
```

EncodePanda:
@nfrisby last commit (https://github.com/input-output-hk/ouroboros-network/pull/3518/commits/5c59ed7590da2a2f7df54f108a1eb0fe25f64132), is that what you were thinking of?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/Types.hs
```diff
@@ -8,7 +12,19 @@ import           Ouroboros.Consensus.Block.RealPoint (RealPoint)
...
+
+newtype Pushing blk = Pushing { unPushing :: RealPoint blk }
+  deriving (Show, Eq)
+
 data UpdateLedgerDbTraceEvent blk =
     -- | Event fired when we are about to push a block to the LedgerDB
-      StartedPushingBlockToTheLedgerDb !(RealPoint blk)
+      StartedPushingBlockToTheLedgerDb
+        !(Pushing blk)
+        -- ^ point which block we are about to push
```
dnadales, line 25: I think we need to capitalize these.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -437,16 +437,22 @@ ledgerDbPush cfg ap db =
...
   where
-    pushAndTrace cfg ap db = do
-      trace $ StartedPushingBlockToTheLedgerDb $ toRealPoint ap
-      res <- ledgerDbPush cfg ap db
-      return res
+    pushAndTrace ap db = do
+      traceStep $ StartedPushingBlockToTheLedgerDb (Pushing $ toRealPoint ap)
```
dnadales, line 448: What about something more direct (with the use of auxiliary `mk` functions)? 

```haskell
trace $ tartedPushingBlockToTheLedgerDb $ (mkPushing ap ) (mkPushGoal aps)
```

ouroboros-consensus-test/test-consensus/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
```diff
@@ -208,6 +208,9 @@ initLgrDB k chain = do
 
     genesisLedgerDB = LgrDB.ledgerDbWithAnchor testInitExtLedger
 
+    noopTrace :: blk -> m ()
```
dnadales, line 211: We have other uses of `const $ pure ()`. Don't we want to replace these with `noopTrace`? Or just use `const $ pure ()`

dnadales (approved):
I left some comments. LGTM.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda: I like it

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/Types.hs
EncodePanda: 👍

ouroboros-consensus-test/test-consensus/Test/Consensus/MiniProtocol/LocalStateQuery/Server.hs
EncodePanda: hmm.. good question, i think it's better if we replace with `noop` helper function, otherwise someone who will read the code will have to check each time to understand "wtf is const pure unit"

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
EncodePanda: this does not compile

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/Types.hs
```diff
@@ -0,0 +1,30 @@
...
+
+newtype Pushing blk = Pushing { unPushing :: RealPoint blk }
+  deriving (Show, Eq)
+
+data UpdateLedgerDbTraceEvent blk =
+    -- | Event fired when we are about to push a block to the LedgerDB
+      StartedPushingBlockToTheLedgerDb
+        !(Pushing blk)
+        -- ^ Point which block we are about to push
+        (PushGoal blk)
```
dnadales, line 26: Is there a `!` missing here?

EncodePanda:
bors merge

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
nfrisby: I think for the the new field it should just be `Tracer`, regardless of new code, but yeah no need to change existing code in this PR.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -323,6 +325,7 @@ initStartingWith tracer cfg streamAPI initDb = do
                        (ledgerState (ledgerDbCurrent db))
                        (ledgerState (ledgerDbCurrent db'))
 
+
```
nfrisby, line 328: Style: undue newline

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -533,4 +535,5 @@ data TraceReplayEvent blk replayTo
     -- The @blockInfo@ parameter corresponds replayed block and the @replayTo@
     -- parameter corresponds to the block at the tip of the ImmutableDB, i.e.,
     -- the last block to replay.
+  | UpdateLedgerDbTraceEvent (UpdateLedgerDbTraceEvent blk)
```
nfrisby, line 538: Please add a Haddock comment on this one, describing the general meaning of the `UpdateLedgerDbTraceEvent` type/category.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -428,27 +437,36 @@ ledgerDbPush cfg ap db =
...
+ledgerDbPushMany ::
+     forall m c l blk . (ApplyBlock l blk, Monad m, c)
+  => (UpdateLedgerDbTraceEvent blk -> m ())
+  -> LedgerDbCfg l
+  -> [Ap m l blk c] -> LedgerDB l -> m (LedgerDB l)
+ledgerDbPushMany trace cfg aps initDb = (repeatedlyM pushAndTrace) aps initDb
+  where
+    pushAndTrace ap db = do
+      let pushing = Pushing . toRealPoint $ ap
+          goal    = PushGoal . toRealPoint . last $ aps
```
nfrisby, line 449: I'd prefer to not call partial functions correctly. So write a case expression with a nice `error` message instead of calling `last`.

nfrisby:
Bah --- I just realized this was already Merged :)

PR #3525, opened 2021-12-03
Register owners and reward accounts of pools in Shelley Genesis

kderme:
Shouldn't this work like this? Currently the reward account of pools is not registered. So it doesn't receive leader rewards, unless it is also in the `sgsStake` or manually registered later. Same for pool owners.

An additional argument can be made that pool owners should be included in the `SL._delegations` field by default. This is not done in this pr.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
@@ -10,6 +10,7 @@
 {-# LANGUAGE OverloadedStrings        #-}
 {-# LANGUAGE RecordWildCards          #-}
 {-# LANGUAGE ScopedTypeVariables      #-}
+{-# LANGUAGE TupleSections            #-}
 {-# LANGUAGE TypeApplications         #-}
 {-# LANGUAGE TypeFamilies             #-}
 {-# LANGUAGE TypeOperators            #-}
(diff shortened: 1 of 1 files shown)


nc6:
So, this function was originally intended to support getting stake pools up and running, so that we could test decentralisation without waiting a long time for delegations. Rewards are not part of that; in the vein of adding only the minimal logic to a "here be dragons" function that mucks around with internal state, they are therefore not part of this. It's completely legitimate to have an unregistered reward account.

That having been said, if there's a compelling argument *for* having this, I don't see any reason why we shouldn't register the reward accounts. Is there testing you want to do that would require this?

Registering the owner keys is a different matter. Owners do not get rewards in the classical manner, and in any case should have been included in the delegation map (since they are expected to contribute to the pledge). So I see no logic in automatically registering the accounts here, especially since anything registered here mucks up accounting by having no deposit.

So, before approving this, I would like to see the following:

- Update the PR description with the motivation for needing this. If there's no compelling reason, I would prefer to avoid adding more logic into this function.
- Remove the automatic owner registration.

nc6 (changes requested):
As per comment

kderme:
Yes I'm using this function to start tests from an initial state. I thought it makes more sence to have registered reward accounts as a default. But I also see the point of just doing the minimum here to have pools up and running and bootstrap block production for d=0.

This pr is not strictly necessary for my tests, since owners and reward accounts can be registered separately. So closing this.

PR #3526, opened 2021-12-06
make unregister functions non-blocking

coot:
- connection-manager: cancel threads using throwTo
- ~connection-manager: added transition tracer~
- connection-manager-test: removed prop_connectionManagerSimulation
- connection-manager: enforce WaitTime timeout
- connection-manager: change how masking of the cleanup function is done.
- server-test: comment why 'splitConns' is correct
- connection-manager: demotedToColdRemote


ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -675,7 +677,10 @@ withConnectionManager ConnectionManagerArguments {
 
             when shouldTrace $
               traceWith trTracer tr
-            traverse_ cancel (getConnThread connState)
+            -- using 'cancel' here, since we want to block until connection
+            -- handler thread terminates.
+            traverse_ (\thread -> cancel thread)
```
bolt12, line 684: is this diff necessary?

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -822,7 +824,19 @@ withConnectionManager ConnectionManagerArguments {
...
+                         forceThreadDelay delay = do
+                           t <- getMonotonicTime
+                           unmask (threadDelay delay)
+                             `catch` \e -> 
+                                case fromException e
+                                of Just (AsyncCancelled) -> do
+                                     t' <- getMonotonicTime
+                                     forceThreadDelay (delay - t' `diffTime` t)
+                                   _ -> throwIO e
+                     in forceThreadDelay cmTimeWaitTimeout
```
bolt12, line 839: Maybe move the definition elsewhere?

This function ignores `AsyncCancelled` exceptions but does not re-throws them; should it?

bolt12 (changes requested):
Looks good! just a few comments!

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
coot: no, it isn't. I'll fix this.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
coot: Good question.  The answer is yes, it ignores them to make sure it will delay for the given time.  We `throwTo ... AsyncCancelled` and we don't care how much time it will take for the thread to continue.  The exception is when the connection manager shuts down, there we use `cancel`, which will wait until the thread terminates (and thus `WaitTime` passes).  We could just kill the threads with another exception and not wait for the the final transitions (or rethrow `AsyncCancelled`), but this would break tests, since the final transition is done after `WaitTime` passes.

bolt12: Got it! Thanks!

coot:
I removed the _connection-manager: added transition tracer_, to avoid modification of traces, this currently causes problems for merging the new tracing system.

bolt12:
Maybe this will fix my timeout enforced tests

coot:
bors merge

PR #3532, opened 2021-12-07
Added prop_timeouts_enforced

bolt12:


ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2366,6 +2370,112 @@ prop_connection_manager_counters serverAcc (ArbDataFlow dataFlow)
...
+               counterexample
+                 (errorMsg tt t' t (tTimeWaitTimeout simTimeouts))
+               $ diffTime t t' <= tTimeWaitTimeout simTimeouts
+              _                -> property True
+         in case to of
+           InboundIdleSt _       -> valid .&&. verifyTimeouts newStateMap' xs
+           OutboundDupSt Ticking -> valid .&&. verifyTimeouts newStateMap' xs
+           OutboundIdleSt _      -> valid .&&. verifyTimeouts newStateMap' xs
+           TerminatingSt         -> valid .&&. verifyTimeouts newStateMap' xs
+           _                     -> verifyTimeouts newStateMap xs
```
coot, line 2441: I think this is not right.  If we have a transtion `OutboundDupSt Ticking -> OutboundDupSt Expired` we will not do the check.  In all cases of the `to` state we can check either of the two:
* timeout expired
* the transition was triggered by some other event, sooner than the timeout
These two cases would be covered by:
```
case to of
  _ -> valid .&&. verifyTimeouts newStateMap' xs
```
but this will not work for the `InboundIdleState^\tau -> OutboundState^\tau -> OutboundState` sequence.  The first transition would be fine, but for the second we need the time when we transitioned into `InboundIdleState` and not `OutboundState` (which is not `t` when we check the other transition).

coot (changes requested):

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2366,6 +2370,112 @@ prop_connection_manager_counters serverAcc (ArbDataFlow dataFlow)
...
+    verifyTimeouts stateMap [] =
+      counterexample
+        ("These states didn't timeout:\n"
+        ++ show stateMap
+        )
+      $ Map.null stateMap
+    verifyTimeouts stateMap ((t, TransitionTrace _ tt@(Transition from to)):xs)
+      | Map.member from stateMap =
+        let newStateMap  = Map.delete from stateMap
+            newStateMap' = Map.insert to t newStateMap
```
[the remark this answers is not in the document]
bolt12: We need the intermediate state because depending on a condition we either want one state or the other.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2366,6 +2370,112 @@ prop_connection_manager_counters serverAcc (ArbDataFlow dataFlow)
...
+    verifyAllTimeouts (t:tts) =
+      counterexample ("\nConnection transition trace:\n"
+                     ++ intercalate "\n" (map show t)
+                     )
+      $ verifyTimeouts Map.empty t
+      .&&. verifyAllTimeouts tts
+
+    verifyTimeouts :: Map AbstractState Time
+                   -> [(Time , AbstractTransitionTrace SimAddr)]
+                   -> Property
```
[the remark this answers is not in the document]
bolt12: Right, will do!

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
bolt12: I think you are right when you say I need to check timeout expired on the catch-all case, thanks!

I didn't think about that transition, I think it is easy to accommodate for that case, we just have to keep the time from the first transition into InboundIdleState as you said :)

bolt12: Addressed!

bolt12:
@coot Refactored the code as in our call. It turns out I ended up needing to be extra explicit on all the states but I think the code turned out better than the one I had!

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -682,8 +687,7 @@ withConnectionManager ConnectionManagerArguments {
             -- using 'cancel' here, since we want to block until connection
             -- handler thread terminates.
             traverse_ cancel (getConnThread connState)
-          )
-          state
+          ))
```
coot, line 690: would it work if we `throwTo` and then wait for all the threads to terminate; this would avoid calling `cancel` from a separate thread.  The drawback is that if a connection is executing the `cleanup` handler, it will block for the `TIME_WAIT` interval.

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
```diff
@@ -92,6 +89,8 @@ writeQueueChannel QueueChannel { qcWrite } msg =
         Nothing -> return False
         Just q  -> writeTQueue q msg
                 >> case msg of
+                      -- Match SO_LINGER set with 0 interval
+                      -- and close both ends without any handshake
```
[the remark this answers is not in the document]
bolt12: I'd say in a way we are already simulating this case by writing the message and not waiting on receiving some sort of Ack. Of course since this is a simulation environment we do not actually need to do much because the message doesn't go through the wire.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -112,9 +112,10 @@ inboundGovernor trTracer tracer serverControlChannel inboundIdleTimeout
        (\(e :: SomeAsyncException) -> do
          state <- atomically $ readTVar st
          _ <- Map.traverseWithKey
-               (\connId _ ->
+               (\connId _ -> do
+                 let state' = unregisterConnection connId state
```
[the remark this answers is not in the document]
bolt12: Oh this is just a better way to trace the `RemoteTransitionTrace` without the `emptyState` hack. Basically we want to remove the connection from the state and then `mkRemoteTransitionTrace` will look into the difference and trace the correct transition

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
bolt12: `throwTo` is also possibly blocking right? waiting on the thread to actually receive the exception. So I do not think we would gain much from using `throwTo`.

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2450,6 +2453,233 @@ prop_connection_manager_counters serverAcc (ArbDataFlow dataFlow)
...
+               .&&. verifyTimeouts Nothing xs
+
+             TerminatedSt ->
+               counterexample (errorMsg tt t' t idleTimeout)
+               $ diffTime t t' <= timeWaitTimeout
+               .&&. verifyTimeouts Nothing xs
+
+             _ -> error ("Unexpected invalid transition: " ++ show (st, tt))
+
+           _ -> error ("Should be a \tau state: " ++ show st)
```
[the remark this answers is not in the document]
bolt12: > We actually have one more timeout which is not reflected on the state machine: handshake one. It starts in UnnegotiatedState _ and terminates on its first transition (to Outbound Unidirectional, InboundIdleState^\tau Duplex, InboundIdleState\tau Unidirectional or TerminatedState).

Oh yeah I could add this ! Thanks :) 

> Have you thought how to show some statistics which timeouts are we actually testing?

Like showing some sort of coverage? I know we have the state coverage for the outbound governor so one can infer some statistics from there, since we test a timeout everytime we reach a \tau state. If that coverage test shows good coverage for the \tau states then I think we are good

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
bolt12: @coot

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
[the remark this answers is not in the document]
bolt12: Addressed!

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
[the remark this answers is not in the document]
bolt12: Addressed

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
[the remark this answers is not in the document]
bolt12: Addressed

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
coot: * `cancel` blocks until thread terminates
* `throwTo` blocks until an exception is raised

so `cancel` can block for much longer than `throwTo`, since during `TIME_WAIT` interval, we will accept asynchronous exceptions (though they will be ignored).  This will unblock `throwTo`, but not `cancel`.

bolt12: Addressed

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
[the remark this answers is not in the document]
coot: If we want to do that properly, we should thread the state and use `Map.foldrWithKey` instead, otherwise ignoring `state'` is just confusing.

bolt12:
We are not ignoring state' we are using it to build the trace. That's all
that matters here because the inbound governor is going to die. What we had
previously was a hack and I was building the trace using the empty state,
with this patch it is a cleaner way to do it.

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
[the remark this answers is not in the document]
coot: > I'd say in a way we are already simulating this case by writing the message and not waiting on receiving some sort of Ack.

That's not right; resting a connection does not make it wait on an ack, the kernel will just forget the state of the connection when it sends the packet with `RST` option set.   If the receiver does not get the packet it will get an error when next time it will try to send a packet (a reset will be send as a reply).

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -679,11 +686,20 @@ withConnectionManager ConnectionManagerArguments {
...
+            traverse
+              (\thread -> do
+                throwTo (asyncThreadId thread) AsyncCancelled
+                pure thread
+              )
+              (getConnThread connState)
+          ) state
+
+        atomically $ runLastToFinishM
+                   $ foldMap (LastToFinishM . void <$> waitCatchSTM) asyncs
```
coot, line 701: Nice use case for `LastToFinishM`.

bolt12:
Hmmm right, is not very faithful.

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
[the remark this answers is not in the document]
bolt12: I will experiment with simulating a lost message like you suggested and see if that still passes the test. The problem before was that we were waiting for an Ack and that could take very long. Hopefully just closing and losing the MsgClose message will not delay it much

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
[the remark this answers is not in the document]
bolt12: Just noticed replies by email do not go to the correct thread :/

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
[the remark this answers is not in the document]
bolt12: Tried it and the tests are passing 😄 I think now it simulates the LINGER option more faithfully.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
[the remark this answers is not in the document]
bolt12: > If we want to do that properly, we should thread the state and use Map.foldrWithKey instead, otherwise ignoring state' is just confusing.

> We are not ignoring state' we are using it to build the trace. That's all
that matters here because the inbound governor is going to die. What we had
previously was a hack and I was building the trace using the empty state,
with this patch it is a cleaner way to do it.

All we need is to remove that particular connection, I understand now what you are saying about threading state, but it is not really necessary, here.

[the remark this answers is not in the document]
coot: right, also every next `unregisterConnection` will look through smaller map (though performance is not critical here as well).

ouroboros-network-framework/test/Test/Ouroboros/Network/Server2.hs
```diff
@@ -2450,6 +2456,269 @@ prop_connection_manager_counters serverAcc (ArbDataFlow dataFlow)
...
+         in case to of
+           InboundIdleSt _       -> verifyTimeouts newState xs
+           OutboundDupSt Ticking -> verifyTimeouts newState xs
+           OutboundIdleSt _      -> verifyTimeouts newState xs
+           TerminatingSt         -> verifyTimeouts newState xs
+           _                     -> verifyTimeouts Nothing xs
+
+    getTraceEvents :: forall a b. Typeable b
+                      => SimTrace a
+                      -> [(Time, b)]
```
coot, line 2702: if we write this function to return a `Trace` not a list of events, we will not need a separate `splitConns` (which is unfortunate, because this means we duplicate critical code for some of the tests).

There's `traceSelectEventsDynamic` which filters `Dynamic` values checking their type.
All together `transitionSignal` (a few lines above) is probably slightly more complex than:
```haskell
      transitionSignal :: [[(Time, AbstractTransitionTrace SimAddr)]]
      transitionSignal = Trace.toList                      
                       . splitConns
                       . traceSelectTraceEventsDynamic
                       $ trace
```

bolt12: The `splitConns` function that works on `Trace` does not get us the Time of the event which is needed for this particular test. That's why I duplicated the code. I gave your suggestion a go and it doesn't seem to have that type (it has `[[AbstractTransition]]` )

bolt12: I think we lose the `Time` info when we go from `SimTrace` to `Trace` via `traceSelectTraceEventsDynamic`, that's why I duplicated the code to write a version that would give me the `Time` as well.

coot: There's a fix for that by using contra-variance of the tracer.  One needs to define

```
data WithTime a = WithTime Time a
```
And then use
```
tracerWithTime :: Tracer m (WithTime a) -> Tracer m a
tracerWithTime (Tracer f) = Tracer $ \a -> WithTime <$> getMonotonicTime <*> f a
```

bolt12: Interesting, but I am failing to see how that is going to help, `multinodeExperiment` requires `Tracer m (WithName (Name peerAddr) (AbstractTransitionTrace peerAddr))`. And we need to extract the time from `SimEvent`, at least it seems cleaner to extract it from there than from `getMonotonicTime`

coot: So you need to construct:
```haskell
Tracer m (WithTime (WithName (Name peerAddr) (AbstractTransition peerAddr)))
```
pass it through `tracerWithTime` and you have what you need.

How to construct it? The same:
```haskell
tr :: Tracer m (WithTime (WithName (Name peerAddr) (AbstractTransition peerAddr)))
tr = Tracer IOSim.traceM
```

bolt12: What about adding 

```hs
traceSelectTraceEventsTime
    :: (SimEventType -> Maybe b)
    -> SimTrace a
    -> Trace (SimResult a) (Time, b)
traceSelectTraceEventsTime fn = bifoldr ( \ v _acc -> Nil v )
                                        ( \ eventCtx acc
                                         -> case fn (seType eventCtx) of
                                              Nothing -> acc
                                              Just b  -> Cons (seTime eventCtx, b) acc
                                        )
                                        undefined -- it is ignored
```

to `IOSim.hs` ?

This would avoid making a convoluted change to the code and avoid duplication in this case, and I think it is a nice, justified addition to IOSim module

bolt12: Note that [here](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/MockEnvironment.hs#L491) something similar is done.

coot: I'd prefer to keep `io-sim` api as small as possible, the other place should also use contravariant tracer to get the time; another thing is that this also works in `IO`, not just `io-sim`.

bolt12: I am afraid your suggestion does not typecheck :/ `Tracer` performs an `m ()` action, so all we really can do is to perform `getMonotonicTime` in order to print it, not add it to `WithTime`

bolt12:
```hs
λ data WithTime a = WithTime Time a
λ :t contramapM $ \a -> flip WithTime a <$> getMonotonicTime
contramapM $ \a -> WithTime <$> getMonotonicTime <*> pure a
  :: MonadMonotonicTime m => Tracer m (WithTime a) -> Tracer m a
```

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
```diff
@@ -85,16 +82,22 @@ readQueueChannel QueueChannel { qcRead } =
...
-                      MsgClose -> writeTVar qcWrite Nothing
-                      _        -> return ()
-                >> return True
+        Nothing -> writeTVar qcRead Nothing
+                >> return False
+        Just q  ->
+          case msg of
+            MsgClose -> writeTVar qcWrite Nothing
+            _        -> writeTQueue q msg
+          >> return True
```
coot, line 100: `do` notation is probably a better choice. `>>` nicely inlines with `->`, I like `$>` though some don't (e.g. Duncan).

coot (approved):
LGTM, thanks @bolt12!

network-mux/src/Network/Mux/Bearer/AttenuatedChannel.hs
bolt12: Addressed!

bolt12:
bors merge

bolt12:
bors merge

PR #3530, opened 2021-12-07
Update cardano-ledger dependency

newhoggy:

PR #3537, opened 2021-12-10
Update nix environment

Jasagredo:
Updated:
- Index-state to 23-11-2021
- haskell.nix to c8faecc486239050221d4546779aec9f45f3c404
- HLS to 1.5.1
- stylish-haskell to 0.13.0.0
- hasktags to 0.72.0

Added:
- profiteur
- eventlog2html
- hp2pretty


cabal.project
```diff
@@ -165,6 +165,8 @@ constraints:
   -- ones they reuse the one from 'some', but there isn't e.g. a proper version
   -- constraint from dependent-sum-template (which is the library we actually use).
   , dependent-sum > 0.6.2.0
+  -- needed until we update past cardano-ledger@43f7c7318e38c501c2d2a2c680251c7c1f78d0fd
+  , hashable < 1.3.4.0
```
Jasagredo, line 169: If this PR goes after https://github.com/input-output-hk/ouroboros-network/pull/3530 we can drop this change.

dnadales (approved):

coot (approved):

bolt12 (approved):

nix/hls.nix
```diff
@@ -12,11 +12,6 @@ let
     src = sources.hls-released;
     configureArgs = "--disable-benchmarks";
     modules = [{ enableSeparateDataOutput = true; }];
-    sha256map =
```
EncodePanda, line 15: why this gets removed? reason is not mentioned in the commit message. why was it here in the first place?

EncodePanda (approved):
LTGM. Consider motivating that `sha256map` change because it is not clear why it is being removed.]

nix/hls.nix
Jasagredo: Some sha256 hashes of dependencies were wrong upstream, and it was needed to patch it like that for the version we were using. It is not needed anymore.

Jasagredo:
ThreadScope had to be dropped because GTK fails to buiild on MacOS

nfrisby:
> ThreadScope had to be dropped because GTK fails to buiild on MacOS

Please update the PR description accordingly.

cabal.project
```diff
@@ -1,4 +1,4 @@
-index-state: 2021-02-15T00:00:00Z
+index-state: 2021-11-23T00:00:00Z
```
nfrisby, line 1: The `master` branch of `cardano-ledger-specs` is currently at 12-03:06:23Z

https://github.com/input-output-hk/cardano-ledger/blob/80d2624174b3fc598c95ab9bf0a21c9bc7e045d3/cabal.project#L1

My intuition is that we'll benefit from having the same index-state as our primary upstream library. Would that work here?

(I made the same comment on PR #3530.)

nfrisby (approved):
Approved. Thanks!

Please update the PR description as my previous comment requested. And consider matching the `cardano-ledger-specs` index-state, as my other comment requested.

nfrisby:
bors merge

PR #3540, opened 2021-12-14
CAD-3783 DoneAddingBlock event tracer for block delay metrics

### LINKED ISSUE #3382: Add new tracer event that reports a block's propagation delay within our node
@karknu and @njd42 pointed out that a a helpful diagnostic metric on running nodes would be (statistics about) the delay between receiving a block and when we began offering it to peers. In some sense, this is our node's internal contribution to the that block's propagation delay (at least for p2p connections whose shortest path involves us).

This Issue is to create a tracer event that includes that specific delay along with the block's metadata and likely additional information about the corresponding chain selection event (eg if it was a roll back, eg sizes of other blocks).

There are some open questions, such as: Would it be ok to only emit this event for some block when that block becomes our _tip_? Recall that a block can be on our chain without having ever been our tip (eg if it's in the prefix of short fork that later wins out). Only emitting the event for our tip would be quite a bit easier to implement, eg (boils down to adding an ingress timestamp to the add-block queue, I think).

nfrisby:
Fixes #3382.

The commits are well-separated; likely best reviewed individually.

Issue 3382 (aka CAD-3783) is resolved by this PR by the addition of the new `DoneAddingBlock` event, specifically the `UTCTime`s it contains.

This PR introduces a few lightweight types that help preserve the separation of concerns within the code itself, despite needing to calculate those times. In particular, the correct input-output behavior of the code that emits this new event does not otherwise require any of the times necessary to construct the event. We accordingly use the `OnlyForTracer` newtype to limit what the the new monadic constraint and new ChainDB logic that compute these times can influence (ie only `Tracer`s). Similarly, the `ChainSel` code that emits the event doesn't otherwise need access to the constraints that are required for the principal intended use of the event; hence the `EventConstrainedData` newtype in the tracer event itself.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
@@ -67,11 +67,16 @@ import           Control.Monad (void)
 import           Data.Typeable (Typeable)
 import           GHC.Generics (Generic)
 
+import           Cardano.Slotting.Time (RelativeTime)
+
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment)
 import qualified Ouroboros.Network.AnchoredFragment as AF
 import           Ouroboros.Network.Block (ChainUpdate, MaxSlotNo,
                      Serialised (..))
 import qualified Ouroboros.Network.Block as Network
+import           Ouroboros.Network.BlockFetch (FromConsensus)
+
+import           Ouroboros.Network.Tracers.OnlyForTracer (OnlyForTracer)
 
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.HeaderStateHistory
@@ -316,6 +321,13 @@ data ChainDB m blk = ChainDB {
       -- longer detected by this function, the 'Fingerprint' doesn't have to
       -- change, since the function will not detect new invalid blocks.
     , getIsInvalidBlock :: STM m (WithFingerprint (HeaderHash blk -> Maybe (InvalidBlockReason blk)))
+
+      -- | For a header that satisfies the 'FromConsensus' invariants, we can
+      -- always compute the corresponding slot's 'RelativeTime'. In the absence
+      -- of those invariants, it's possible we cannot translate the header's
+      -- slot to a 'RelativeTime' or can only (silently!) translate it
+      -- _incorrectly_.
+    , calculateHeaderSlotTime :: FromConsensus (Header blk) -> STM m (OnlyForTracer RelativeTime)
 
       -- | Close the ChainDB
       --
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
@@ -26,6 +26,7 @@ module Ouroboros.Consensus.Storage.ChainDB.Impl.LgrDB (
   , currentPoint
   , getCurrent
   , getDiskPolicy
+  , getRunWithCachedSummary
   , setCurrent
   , takeSnapshot
   , trimSnapshots
@@ -70,6 +71,10 @@ import           Ouroboros.Consensus.Protocol.Abstract
 import           Ouroboros.Consensus.Util.Args
 import           Ouroboros.Consensus.Util.IOLike
 import           Ouroboros.Consensus.Util.ResourceRegistry
+
+import           Ouroboros.Consensus.HardFork.Abstract (HardForkIndices,
+                     HasHardForkHistory, hardForkSummary)
+import qualified Ouroboros.Consensus.HardFork.History as History
 
 import           Ouroboros.Consensus.Storage.Common
 import           Ouroboros.Consensus.Storage.FS.API (SomeHasFS (..),
@@ -119,6 +124,8 @@ data LgrDB m blk = LgrDB {
     , diskPolicy     :: !DiskPolicy
     , hasFS          :: !(SomeHasFS m)
     , tracer         :: !(Tracer m (TraceEvent blk))
+    , hfSummaryCache :: !(History.RunWithCachedSummary (HardForkIndices blk) m)
+      -- ^ INVARIANT: this uses 'cfg' and re-reads 'varDB' on a cache miss
     } deriving (Generic)
 
 deriving instance (IOLike m, LedgerSupportsProtocol blk)
@@ -173,6 +180,7 @@ openDB :: forall m blk.
           , LedgerSupportsProtocol blk
           , LgrDbSerialiseConstraints blk
           , InspectLedger blk
+          , HasHardForkHistory blk
           , HasCallStack
           )
        => LgrDbArgs Identity m blk
@@ -216,6 +224,10 @@ openDB args@LgrDbArgs { lgrHasFS = lgrHasFS@(SomeHasFS hasFS), .. } replayTracer
     let dbPrunedToImmDBTip = LedgerDB.ledgerDbPrune (SecurityParam 0) db
     (varDB, varPrevApplied) <-
       (,) <$> newTVarIO dbPrunedToImmDBTip <*> newTVarIO Set.empty
+
+    cache <- History.runWithCachedSummary
+               (toSummary lgrTopLevelConfig <$> readTVar varDB)
+
     return (
         LgrDB {
             varDB          = varDB
@@ -225,9 +237,22 @@ openDB args@LgrDbArgs { lgrHasFS = lgrHasFS@(SomeHasFS hasFS), .. } replayTracer
           , diskPolicy     = lgrDiskPolicy
           , hasFS          = lgrHasFS
           , tracer         = lgrTracer
+          , hfSummaryCache = cache
           }
       , replayed
       )
+
+toSummary ::
+     ( HasHardForkHistory blk
+     , IsLedger (LedgerState blk)
+     )
+  => TopLevelConfig blk
+  -> LedgerDB' blk
+  -> History.Summary (HardForkIndices blk)
+toSummary cfg  =
+    hardForkSummary (configLedger cfg)
+  . ledgerState
+  . LedgerDB.ledgerDbCurrent
 
 initFromDisk
   :: forall blk m.
@@ -265,12 +290,20 @@ initFromDisk LgrDbArgs { lgrHasFS = hasFS, .. }
                               (decodeDisk ccfg)
 
 -- | For testing purposes
-mkLgrDB :: StrictTVar m (LedgerDB' blk)
-        -> StrictTVar m (Set (RealPoint blk))
+mkLgrDB :: ( HasHardForkHistory blk
+           , MonadSTM m
+           , LedgerSupportsProtocol blk
+           )
+        => LedgerDB' blk
         -> (RealPoint blk -> m blk)
         -> LgrDbArgs Identity m blk
-        -> LgrDB m blk
-mkLgrDB varDB varPrevApplied resolveBlock args = LgrDB {..}
+        -> m (LgrDB m blk)
+mkLgrDB genesisLedgerDB resolveBlock args = do
+    varDB          <- newTVarIO genesisLedgerDB
+    varPrevApplied <- newTVarIO mempty
+    hfSummaryCache <-
+      History.runWithCachedSummary (toSummary cfg <$> readTVar varDB)
+    pure LgrDB {..}
   where
     LgrDbArgs {
         lgrTopLevelConfig = cfg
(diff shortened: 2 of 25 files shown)


nfrisby:
@coot If the `OnlyForTracer` type is worthwhile, I think we should move it to the upstream `iohk-monitoring-framework` repo. But for now, for convenience, I put it in `ouroboros-network`. If that's a blocker for you, please Request Changes and I'll mark this as a  Draft PR.

nfrisby:
@karknu Because of the lightweight types I added in this PR, we can use time in tracers while simultaneously preventing it from "polluting" the code that emits those tracers.

As a result, the event data constructor itself can have essentially the exact fields you originally wanted. If I understand/recall correctly. Thus: I'm specifically hoping you like what you see in the definition of the `TraceDoneAddingBlockEvent` record data type on this PR.

ouroboros-network/src/Ouroboros/Network/Tracers/OnlyForTracer.hs
```diff
@@ -0,0 +1,99 @@
...
+    -- | This is only safe to use _inside_ of another 'OnlyForTracer'
+    -- computation
+    unsafeRunOnlyForTracer :: a
+  }
+
+instance Functor OnlyForTracer where
+  fmap f (UnsafeOnlyForTracer a) = UnsafeOnlyForTracer (f a)
+
+instance Prelude.Applicative OnlyForTracer where
+  pure = pure
```
coot, line 45: I see `pure` is defined at the top level; why not just 'UnsafeOnlyForTracer' here?

ouroboros-network/src/Ouroboros/Network/Tracers/OnlyForTracer.hs
```diff
@@ -0,0 +1,99 @@
...
+-- | The only safe use of 'UnsafeOnlyForTracer'
+pure :: a -> OnlyForTracer a
+pure = UnsafeOnlyForTracer
+
+-- | The only way to eliminate the 'OnlyForTracer' structure is to let it
+-- influence a 'Tracer'
+contramap :: OnlyForTracer (a -> b) -> Tracer m b -> Tracer m a
+contramap = Tracer.contramap . unsafeRunOnlyForTracer
+
+-- | A exported wrapper only used to avoid impredicative polymorphism
```
coot, line 57: Add a `todo` note that we should be able to get rid of it in `ghc-9.2`.

ouroboros-network/src/Ouroboros/Network/Tracers/OnlyForTracer.hs
```diff
@@ -0,0 +1,99 @@
...
+lift = contramap mkRun
+
+contramapM :: Monad m
+           => OnlyForTracer (a -> m b)
+           -> Tracer m b
+           -> Tracer m a
+contramapM kOft bTr =
+    contramap
+      (fmap (\kRun a -> kRun kOft a) mkRun)
+      (Tracer.contramapM id bTr)
```
coot, line 87: why not just `bTr` (`Tracer.contramapM id = id`).

coot: Forget my comment.  I `Tracer.contramapM id :: Tracer m a -> Tracer m (m a)`.

coot: Hmm, isn't this a fancy way of writing:
```
contramapM (UnsafeOnlyForTracer f) = Tracer.contramapM f
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -627,6 +643,23 @@ data TraceAddBlockEvent blk =
     -- This is done for all blocks from the future each time a new block is
     -- added.
   | ChainSelectionForFutureBlock (RealPoint blk)
+
+    -- | The ChainDB has finished processing an instruction to add a block.
+    --
+    -- If the block was relevant, valid, new, and so on, then this instruction
```
dnadales, line 649: Do we need the "and so on" part?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -642,6 +675,55 @@ deriving instance
...
+    TraceDoneAddingBlockEvent {
+        -- | The block itself
+        addedBlock             :: EventAuxData "addedBlock" blk
+        -- | The possibly-new ChainDB tip that resulted from this block having
+        -- been added
+      , addedBlockNewSelection :: Point blk
+        -- | The point of the added block
+      , addedBlockPoint        :: RealPoint blk
+
+        -- These fields are all times. They're ordered by expected value instead
```
dnadales, line 689: These fields "are present all times"?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -642,6 +675,55 @@ deriving instance
...
+        -- of by name.
+
```
dnadales, line 691: I notice we use `Time`, `RelativeTime`, and `UTCTime`. Do we elaborate somewhere on why do we use different times?

dnadales (approved):
Looks great.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
nfrisby: "Each of these fields is a time." 👍

dnadales: Ohhhh, now it makes sense to me :)

ouroboros-network/src/Ouroboros/Network/Tracers/OnlyForTracer.hs
```diff
@@ -0,0 +1,99 @@
...
+-- leak the contained data outside of tracer invocations, but that should be
+-- caught and avoided during code review.
+newtype OnlyForTracer a = UnsafeOnlyForTracer {
+    -- | This is only safe to use _inside_ of another 'OnlyForTracer'
+    -- computation
+    unsafeRunOnlyForTracer :: a
+  }
+
+instance Functor OnlyForTracer where
+  fmap f (UnsafeOnlyForTracer a) = UnsafeOnlyForTracer (f a)
```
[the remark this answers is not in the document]
nfrisby: Done. I added a note explaining that `Foldable` and `Traversable` should *not* be derived.

ouroboros-network/src/Ouroboros/Network/Tracers/OnlyForTracer.hs
nfrisby: I switched to the following, which avoids the awkward `pure = pure` line.

```
instance Prelude.Applicative OnlyForTracer where
  pure = UnsafeOnlyForTracer
  UnsafeOnlyForTracer f <*> UnsafeOnlyForTracer a = UnsafeOnlyForTracer (f a)

-- | The only safe use of 'UnsafeOnlyForTracer'
pure :: a -> OnlyForTracer a
pure = Prelude.pure
```

ouroboros-network/src/Ouroboros/Network/Tracers/OnlyForTracer.hs
nfrisby: Yep. My goal for these "convenience" synonyms is to define them all via the core `contramap` interface. This ensures that no "convenience" function accidentally adds additional expressivity to the safe interface.

I added a comment saying as much.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -265,12 +290,19 @@ initFromDisk LgrDbArgs { lgrHasFS = hasFS, .. }
...
+           , IsLedger (LedgerState blk)
+           , MonadSTM m
+           )
+        => StrictTVar m (LedgerDB' blk)
         -> StrictTVar m (Set (RealPoint blk))
         -> (RealPoint blk -> m blk)
         -> LgrDbArgs Identity m blk
-        -> LgrDB m blk
-mkLgrDB varDB varPrevApplied resolveBlock args = LgrDB {..}
+        -> m (LgrDB m blk)
```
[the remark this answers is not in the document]
nfrisby: Yep, it was. I folded those arguments in to this function.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -186,6 +195,52 @@ getMaxSlotNo CDB{..} = do
...
+--
+-- By the PRECONDITIONs documented in the 'FromConsensus', we can assume that
+-- the given header was validated by the ChainSync client (or else forged by the
+-- local node). This means its slot was, at some point, within the ledger view
+-- forecast range of the ledger state of our contemporary intersection with the
+-- header itself (and that intersection extended our contemporary immutable
+-- tip). A few additional facts ensure that we will always be able to thereafter
+-- correctly convert that header's slot using our current chain's ledger state.
+--
+--   o For under-developed reasons, the ledger view forecast range is equivalent
```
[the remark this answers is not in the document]
nfrisby: I don't think the guide mentions it. Edsko used this in PRs I reviewed, so I've been doing it to.

ouroboros-consensus/src/Ouroboros/Consensus/Util/EventAuxData.hs
```diff
@@ -0,0 +1,58 @@
...
+import           GHC.TypeLits (KnownSymbol, Symbol, symbolVal)
+
+{-------------------------------------------------------------------------------
+  Auxiliary data
+-------------------------------------------------------------------------------}
+
+-- | A wrapper for which '==' always returns 'True' and 'show' is a placeholder.
+--
+-- It is convenient for tracer event types to be particularly self-contained.
+-- Which means they may contain more fields that absolutely necessary. In
```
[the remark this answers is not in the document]
nfrisby: Fixed, thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
nfrisby: Thanks for calling out a wishy-washy word. I simplified the comment.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
nfrisby: I added a sentence in the `TraceDoneAddingBlockEvent` data type's Haddock.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
```diff
@@ -317,6 +322,10 @@ data ChainDB m blk = ChainDB {
       -- change, since the function will not detect new invalid blocks.
     , getIsInvalidBlock :: STM m (WithFingerprint (HeaderHash blk -> Maybe (InvalidBlockReason blk)))
 
+      -- | For a header that satisfies the 'FromConsensus' invariants, we can
```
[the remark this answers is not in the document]
nfrisby: Yep. I added a sentence to this comment.

ouroboros-network/src/Ouroboros/Network/Tracers/OnlyForTracer.hs
nfrisby: Done, thanks!

nfrisby:
@coot @dnadales Thanks for your comments. I replied to them and Resolved the threads that were clear. I left a couple threads unresolved in case you wanted to reply to my reply. If my reply was sufficient, please Resolve those threads. Thanks!

nfrisby:
FYI, after some chats on Slack, I opened https://github.com/avieth/contra-tracer/issues/3 upstream.

karknu:
> @karknu Because of the lightweight types I added in this PR, we can use time in tracers while simultaneously preventing it from "polluting" the code that emits those tracers.
> 
> As a result, the event data constructor itself can have essentially the exact fields you originally wanted. If I understand/recall correctly. Thus: I'm specifically hoping you like what you see in the definition of the `TraceDoneAddingBlockEvent` record data type on this PR.

I'm really happy with the new `TraceDoneAddingBlockEvent`, and the way you solved it.

dnadales:
> > @karknu Because of the lightweight types I added in this PR, we can use time in tracers while simultaneously preventing it from "polluting" the code that emits those tracers.
> > As a result, the event data constructor itself can have essentially the exact fields you originally wanted. If I understand/recall correctly. Thus: I'm specifically hoping you like what you see in the definition of the `TraceDoneAddingBlockEvent` record data type on this PR.
> 
> I'm really happy with the new `TraceDoneAddingBlockEvent`, and the way you solved it.

@karknu did you want the field of the new event to be strict?

nfrisby:
In a call just now, we came up with a plan with @karknu .

1) Karl will integrate the tip of this PR into his `cardano-node` branch.

2) We'll create and advertise a commit off of the `cardano-node` `master` branch that integrates this PR in the simplest possible way 👍

3) Then we can merge this PR!

nfrisby:
Closing as stale.

PR #3431, opened 2021-10-13
[SRE-202] Fix stylish-haskell and nixfmt

craigem:
Additionally remove all references to buildkite.

PR #3544, opened 2021-12-15
stylish haskell changes

coot:
- stylish-haskell: let the script run over ouroboros-network packages
- stylish-haskell changes in the network layer


coot:
This will conflict with #3431 which moves the `check-stylish.sh` script.

coot:
bors merge

PR #3551, opened 2021-12-16
Initialise incremental stake when registering funds.

nc6:
This is needed in order to create the correct snapshot at the end of the
first era.


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
```diff
@@ -567,9 +567,11 @@ registerInitialFunds initialFunds nes = nes {
           SL._reserves = reserves <-> coin (SL.balance initialFundsUtxo)
         }
 
+    initIncrStake = SL._stakeDistro utxoState
     ledgerState' = ledgerState {
           SL._utxoState = utxoState {
-              SL._utxo = utxo'
+              SL._utxo = utxo',
+              SL._stakeDistro = SL.updateStakeDistribution initIncrStake (SL.UTxO mempty) utxo'
```
nfrisby, line 574: Would you add a note (or use a local binding) that makes it clear the `SL.UTxO mempty` is the set of UTxO to delete?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
```diff
@@ -567,9 +567,11 @@ registerInitialFunds initialFunds nes = nes {
           SL._reserves = reserves <-> coin (SL.balance initialFundsUtxo)
         }
```
nfrisby, line 569: Style: some of these new `=` and adjacent old `=` ought to align, I think.

nfrisby (approved):
Approved, with some minor requests.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Node.hs
nc6: Aligned!

nfrisby: Line 576 and 577 too please :) `SL._utxo = ` and `SL.stakeDistro =`.

nfrisby: (And the I script has opinions on your long `import` line.)

nfrisby:
Looks like the on-disk ledger state serialization changed? So that's either accidental or you'll need to run the test suite(s?) passing in the `--accept` flag (I think) to update the expected golden result files tracked in the repo.

nc6:
bors merge

nc6:
bors retry

PR #3557, opened 2021-12-23
Unify code of forkSyncStateOnTipPointChange and syncWithLedger

EncodePanda:
forkSyncStateOnTipPointChange’s action is duplicating code defined in syncWithLedger

syncWithLedger is NOT used in production code (only in tests that disable forkSyncStateOnTipPointChange) so originally I thoight about removing syncWithLedger from the Mempool API, but turnout out to be quite tricky and cumbersome.

Eventually I’ve decided it’s just simpler to unify code from both functions.


Jasagredo (approved):
Looks good

EncodePanda:
bors merge

PR #3559, opened 2021-12-24
[CAD-3718] Document how the current Mempool is working in the current design

EncodePanda:
PDF generated here: [Mempool.pdf](https://github.com/input-output-hk/ouroboros-network/files/7774136/Mempool.pdf)

Though it's just better to read this in Emacs where you can navigate between code and this text

.This document helped me to stay sane while constantly switching context between Mempool and other tasks that were poping out in the last 3 weeks. It explains how Mempool works and spots 3 places in code that will need to change with the new HD interface.

Next step will be quickly drafting changes that need to be done in the mempool and then jumping to actuall hacking those changes (the fun part :) ).

### DIFF
--- a/ouroboros-consensus/docs/Mempool.org
+++ b/ouroboros-consensus/docs/Mempool.org
@@ -115,12 +115,25 @@
 
 *** initMempoolEnv
 
-   ~initMempoolEnv~ creates ~MempoolEnv m blk~ whose values will be later used to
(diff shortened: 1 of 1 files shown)


EncodePanda:
After internal discussion, I'm closing this PR. I have this document offline if anyone needs it. Though I doubt anyone will beside me :)

PR #3561, opened 2022-01-06
Remove files and directories from CODEOWNERS that no longer exist

newhoggy:


.github/CODEOWNERS
```diff
@@ -9,13 +9,11 @@
 docs                               @coot @karknu @bolt12 @njd42 @peterwthomson @input-output-hk/network-superowners
 
 io-sim                             @coot @input-output-hk/network-superowners
-io-sim-classes                     @coot @input-output-hk/network-superowners
```
coot, line 12: this should be changed to `io-classes`

coot (changes requested):

PR #3563, opened 2022-01-10
UTxO-HD coupled-ledger implementation

Jasagredo:


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -434,12 +495,14 @@ defaultReadKeySets f dbReader = runReaderT (runDbReader dbReader) f
 -------------------------------------------------------------------------------}
 
 -- | The ledger state at the tip of the chain
-ledgerDbCurrent :: (forall mk. GetTip (l mk)) => LedgerDB l -> l EmptyMK
-ledgerDbCurrent = either unCheckpoint unCheckpoint . AS.head . ledgerDbCheckpoints
+ledgerDbCurrent :: LedgerDB l -> l EmptyMK
+ledgerDbCurrent = seqLast . dbChangelogStates . ledgerDbChangelog
```
Jasagredo, line 499: @nfrisby Do you think that we should actually check for agreement among the implementations when we are performing a query like this? It would require a `OldVsNew l` constraint that might propagate a bit too much, but otherwise it doesn't sound crazy.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -654,27 +733,24 @@ deriving instance NoThunks (LedgerCfg l) => NoThunks (LedgerDbCfg l)
 
 type instance HeaderHash (LedgerDB l) = HeaderHash l
 
-instance IsLedger l => GetTip (LedgerDB l) where
-  getTip = castPoint . getTip . ledgerDbCurrent
-
```
Jasagredo, line 659: Note to self: It's here!

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -329,11 +329,9 @@ getCurrent LgrDB{..} = readTVar varDB
 setCurrent :: IOLike m => LgrDB m blk -> LedgerDB' blk -> STM m ()
 setCurrent LgrDB{..} = writeTVar $! varDB
 
-currentPoint :: forall blk. UpdateLedger blk => LedgerDB' blk -> Point blk
+currentPoint :: LedgerSupportsProtocol blk => LedgerDB' blk -> Point blk
 currentPoint = castPoint
-             . ledgerTipPoint (Proxy @blk)
-             . ledgerState
-             . LedgerDB.ledgerDbCurrent
+             . LedgerDB.ledgerDbTip
```
Jasagredo, line 334: Check again :suspect:

Jasagredo (commented):
Ready to review

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -378,6 +381,9 @@ type LedgerConfig      blk    = LedgerCfg (LedgerState blk)
 type LedgerError       blk    = LedgerErr (LedgerState blk)
 type TickedLedgerState blk mk = Ticked1   (LedgerState blk) mk
 
+class OldVsNew a b where
```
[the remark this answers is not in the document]
Jasagredo: No longer applicable

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -378,6 +381,9 @@ type LedgerConfig      blk    = LedgerCfg (LedgerState blk)
...
+  oldVsNew :: a -> b -> Bool
```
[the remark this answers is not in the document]
Jasagredo: Might for now go with it and modify it later on?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -148,32 +149,33 @@ data LedgerDB (l :: LedgerStateKind) = LedgerDB {
       -- | Ledger states
       ledgerDbCheckpoints :: AnchoredSeq
                                (WithOrigin SlotNo)
-                               (Checkpoint l)
-                               (Checkpoint l)
+                               (Checkpoint (l ValuesMK))
+                               (Checkpoint (l ValuesMK))
     , ledgerDbChangelog   :: DbChangelog l
+    , runDual             :: Bool
```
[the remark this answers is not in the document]
Jasagredo: This is a flag that will control whether to run both implementations or only the new one. In principle this is just a temporary workaround to avoid adding a new flag to the node, etc as it is just for development period.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: Fixed!

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
dnadales: It seems we're casting a point twice in a row. Is this needed? 

```haskell
ledgerDbTip :: IsLedger l => LedgerDB l -> Point l
ledgerDbTip = castPoint . getTip . ledgerDbCurrent
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
dnadales: It might not be a problem if we temporary alter our `TableStuff` to include the `oldVsNew` method, or even better, we could add these two methods:

```haskell
update :: l ValuesMK -> l TrackingMK -> l ValuesMK
cmp :: l mk -> l mk -> Ordering
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
Jasagredo: Not sure where that excerpt comes from, I have the following:
```
-- | Reference to the block at the tip of the chain
ledgerDbTip :: IsLedger l => LedgerDB l -> Point (l EmptyMK)
ledgerDbTip = getTip . ledgerDbCurrent
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -305,30 +308,73 @@ applyBlock :: forall m c l blk
...
+
+      ApplyRef r -> do
+        b <- resolveBlock r
+
+        forgetLedgerStateTracking <$>
+             ( either (throwLedgerError db (blockRealPoint b)) return
+             $ runExcept
+             $ tickThenApply cfg b oldLedgerDbCurrent)
+
+      Weaken _ -> error "unreachable"
```
dnadales, line 349: Maybe we could say that we don't expect `ap` to contain nested `Weaken`, however, I worry that we're depending on an implementation detail from somewhere else. Why not simply use the old pattern in which `Weaken` would cause the function to recurse?

dnadales (commented):
LGTM, I have just a couple of questions that we could check over a video call.

dnadales (approved):

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl.hs
```diff
@@ -135,12 +137,13 @@ openDBInternal args launchBgTasks = runWithTempRegistry $ do
               immutableDbTipPoint
               (contramap TraceLedgerReplayEvent tracer)
       traceWith tracer $ TraceOpenEvent StartedOpeningLgrDB
+      let runDual = maybe False (== "1") $ unsafePerformIO (lookupEnv "DUAL_LEDGER")
```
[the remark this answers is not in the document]
Jasagredo: Done

PR #3569, opened 2022-01-13
consensus/docs: FastLaneSketch.md

nfrisby:
Closes CAD-3813.

### DIFF
--- a/ouroboros-consensus/docs/FastLaneSketch.md
+++ b/ouroboros-consensus/docs/FastLaneSketch.md
@@ -115,8 +115,7 @@ That's the big picture. Some more details follow.
     * The `MsgFast` timeout is comparable to the ChainSync timeout after
       receiving `MsgAwaitReply`.
     * Every `MsgFast` should be strictly preferable to the previous `MsgFast`
-      that peer sent (and from that peer's latest ChainSync, but I haven't
-      convinced myself that's worth checking).
+      that peer sent.
 * Like ChainSync, FastLane is pull-based, so we can't get overwhelmed.
 * Like ChainSync, FastLane is also pipelined, so that we can hide some latency.
 * If a peer offered a (valid) header for an invalid block via ChainSync we
@@ -132,3 +131,22 @@ That's the big picture. Some more details follow.
 * Am I overlooking some fundamental difference when I consider FastLane as just
   a variant of ChainSync? IE is it more than just something that makes headers
   available to the BlockFetch decision logic?
+
+* Should FastLane also require that each header is strictly preferable to the
+  best header the same peer has has ever sent via ChainSync? The whole ChainSync
+  mini protocol ensures that the peer is on a recent enough chain (eg
+  `ForksTooDeep`). So if we require FastLane is ahead of the same peer's
+  ChainSync, then it can only be used for what we intended it for and
+  specifically not abused to send historical headers, which could waste our CPU
+  validating the header they sent. (TODO I'm not totally convinced this is
+  necessary: a FastLane client could simply immediately ignore such historical
+  headers even without validating since they don't extend our chain, them. But
+  perhaps the wasted bandwidth is sufficient reason to introduce this
+  cross-mini-protocol invariant.)
+
+* I don't think we can rate-limit a peer for sending too many invalid blocks via
+  FastLane. First, we can't tell if they're just forwarding the or if they're
+  minting them. Second, because the header must fit on our chain in order for us
+  to fetch it and it must also be better than any header they've previously
+  sent, this can't happen very often, can it? (TODO maybe OpCert ratcheting
+  would let them firehose the net this way each time they lead?)


johnalanwoods (approved):
@nfrisby and I had a call to discuss, I like this approach keeping `chainSync` intact and adding the new `fastLane` protocol.

This design looks good to me.

We discussed minor changes such as the potential of disconnecting clients that repeatedly fastLane headers that result in bad blocks, and the potential value of restricting headers sent over fastLane, so that headers fastLaned must be "better" blocks than those fastLaned before && those offered via chainSync.

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+* Should FastLane just send the block? Is the extra RTT worthwhile? If it just
+  sends the block, then I worry it's a DoS vector for them to consume our
+  bandwidth.
+
+* Am I overlooking some fundamental difference when I consider FastLane as just
+  a variant of ChainSync? IE is it more than just something that makes headers
+  available to the BlockFetch decision logic?
+
+* Should FastLane also require that each header is strictly preferable to the
+  best header the same peer has has ever sent via ChainSync? The whole ChainSync
```
amesgen, line 136:
```suggestion
  best header the same peer has ever sent via ChainSync? The whole ChainSync
```

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+  ChainSync, then it can only be used for what we intended it for and
+  specifically not abused to send historical headers, which could waste our CPU
+  validating the header they sent. (TODO I'm not totally convinced this is
+  necessary: a FastLane client could simply immediately ignore such historical
+  headers even without validating since they don't extend our chain, them. But
+  perhaps the wasted bandwidth is sufficient reason to introduce this
+  cross-mini-protocol invariant.)
+
+* I don't think we can rate-limit a peer for sending too many invalid blocks via
+  FastLane. First, we can't tell if they're just forwarding the or if they're
```
amesgen, line 148:
```suggestion
  FastLane. First, we can't tell if they're just forwarding them or if they're
```

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+
+* Should FastLane also require that each header is strictly preferable to the
+  best header the same peer has has ever sent via ChainSync? The whole ChainSync
+  mini protocol ensures that the peer is on a recent enough chain (eg
+  `ForksTooDeep`). So if we require FastLane is ahead of the same peer's
+  ChainSync, then it can only be used for what we intended it for and
+  specifically not abused to send historical headers, which could waste our CPU
+  validating the header they sent. (TODO I'm not totally convinced this is
+  necessary: a FastLane client could simply immediately ignore such historical
+  headers even without validating since they don't extend our chain, them. But
```
amesgen, line 143:
```suggestion
  headers even without validating since they don't extend our chain, then. But
```
?

amesgen (approved):
Looks very reasonable, thanks! I think I now have a sufficiently accurate mental model of how this should work. 

It is my understanding that all of this new functionality is to be protocol version-gated, such that nodes not supporting this diffusion pipelining can still participate without any changes (but if too many nodes don't support it, there won't be any performance benefit ofc). Shall we add this to the document?

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+    block will be written to the VolDB.
+
+  * Whenever that background thread processes a block-to-add, it performs chain
+    selection on the entirety of the VolDB (including "future blocks", but
+    that's a story for another time).
+
+  * If the block we just added is valid, the best we've ever seen etc, then the
+    ChainDB will switch its selected chain to the chain that ends in that new
+    block. If the block is invalid, we should disconnect from the peers who
+    sent its header via ChainSync and any that do so later: ChainSync conveys
```
dnadales, line 33: Would this make the sentence easier to parse? 

```suggestion
    sent its header via ChainSync and any peer that do so later: ChainSync conveys
```

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+* relevant background thread https://github.com/input-output-hk/ouroboros-network/blob/b0d95f33314e0562b6eea55455f1847efb85f8a4/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs#L523-L536
+* chain selection logic https://github.com/input-output-hk/ouroboros-network/blob/b0d95f33314e0562b6eea55455f1847efb85f8a4/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
+
+# Initial implementation plan
+
+I think I see a lightweight means of extending the above system to "rush"
+through the latest block and to do so only when it would matter.
+
+* Extend the ChainDB API with a part that notifies subscribed threads whenever
+  a new most-desirable-block-ever-seen is added.
```
dnadales, line 88: What is a new most-desirable-block-ever-seen? 

- How do we determine the desirability between two blocks? By their block numbers? 
- Most desirable among the blocks in the volatile DB?

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+* Define a new mini protocol named FastLane as a pared down ChainSync.
+    * It has two messages: `MsgRequestNextFast`, `MsgFast`
+    * The intended semantics is that the server sends `MsgFast hdr` to the
+      client as soon as the server downloads a new most-desirable block.
+    * Notably, the server sends `MsgFast` _before_ attempting to validate the
+      new block.
+* Expand the input to the block fetch decision logic so that it includes the
+  latest header received via FastLane.
```
dnadales, line 96: So does this mean that the block fetch decision logic will now consider downloading blocks from two sources? 

0. The chain sync client headers
1. The fast lane headers


Which, I guess, means that BF would have to distribute its bandwidth somehow.

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+* Crucially, peers are to be forgiven for sending us an invalid block this way.
```
dnadales, line 97: Can adversaries ruin the pipelining by injecting garbage blocks into it? I guess they can make it once per each slot they lead ...

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+    * I'm tempted to make it more liberal, but that would involve extra
+      complexity, along the line of checking whether all the blocks between the
+      new header and its intersection with our selection are already in the
+      VolDB.
+    * Notably, if pipelining has been working, then the simpler check should
+      suffice.
+* Like ChainSync, if the header is valid, then the FastLane client makes its
+  available to the BlockFetch decision logic.
+    * The FastLane client disconnects from the peer if the header extends the
+      node's selection chain but is invalid.
```
dnadales, line 112: What is invalid? The header of the block? :thinking:

ouroboros-consensus/docs/FastLaneSketch.md
amesgen: Yes, in this case a single invalid block diffuses. My understanding is that this is not too bad, as otherwise the diffusion would stop at the adversary's peers, so the network would have been idle, so we are only wasting network resources that could not be used either way.

ouroboros-consensus/docs/FastLaneSketch.md
johnalanwoods: Fastlane headers are always preferred. In fact, Fastlane occurs first for all blocks, then the blocks will transit the standard ChainSync route too.

ouroboros-consensus/docs/FastLaneSketch.md
johnalanwoods: Notably they won't be added to the chain in this implementation (unlike aggressive pipelining)

ouroboros-consensus/docs/FastLaneSketch.md
```diff
@@ -0,0 +1,152 @@
...
+      that peer sent.
+* Like ChainSync, FastLane is pull-based, so we can't get overwhelmed.
+* Like ChainSync, FastLane is also pipelined, so that we can hide some latency.
+* If a peer offered a (valid) header for an invalid block via ChainSync we
+  disconnect. If the peer only offered instead that header via FastLane, then
+  we don't disconnect.
+
+# Doubts
+
+* Should FastLane just send the block? Is the extra RTT worthwhile? If it just
```
johnalanwoods, line 127: I don't think so. I think we should mirror the existing primitive which is predicated on headers.

nfrisby:
We've already merged the design based on Duncan's fundamental tentative header idea.

Closing this as WontDo/Superseded.

Issue #3570, opened 2022-01-13
KeepAlive restart assertion failure.


karknu:
{"thread":"941544","sev":"Notice","data":{"kind":"MuxTrace","event":"Miniprotocol MiniProtocolNum 8 InitiatorDir terminated with exception Assertion failed\nCallStack (from HasCallStack):\n  assert, called at src/Ouroboros/Network/BlockFetch/ClientRegistry.hs:199:11 in ouroboros-network-0.1.0.0-af9948a3716d6519db861fe72f2421e66372a0206f06012c6493302e1e92c70b:Ouroboros.Network.BlockFetch.ClientRegistry","bearer":{"remote":{"addr":"1.1.1.1","port":"7"},"local":{"addr":"1.1.1.2","port":"6"}}},"loc":null,"env":"1.31.0:0ff76","msg":"","app":[],"host":"ip-172-3","at":"2021-12-31T21:49:54.55Z","ns":["cardano.node.Mux"],"pid":"26846"}

It is possible for the node to end up in a state where it can no longer re-establish connections to a specific peer (could be a local root peer).  Promotion from cold to warm will keep failing with the log message above.

karknu:
Done

PR #3575, opened 2022-01-17
Stop mux incase termination of miniprotocols timeout

karknu:
When demoting a peer to cold there is a 300s timeout for the
miniprotocols to terminate. In case of timeout the connection will be
left hanging, it will only be unregistered. The next time the
peer is promoted to warm there may still be lingering miniprotocols in
an unknown state.

By calling stopMux we ensure that the connection will be
reset.


ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -903,6 +904,7 @@ withPeerStateActions PeerStateActionsArguments {
       case res of
         Nothing -> do
           -- timeout fired
+          Mux.stopMux pchMux
           _ <- unregisterOutboundConnection spsConnectionManager (remoteAddress pchConnectionId)
```
coot, line 908: You can leave remove this call. There's no need to unregister a connection that died.

coot (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
karknu: I left it there to match https://github.com/input-output-hk/ouroboros-network/blob/f2425ffba98a60c253d6ba136b5af7006b40bd85/ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs#L834-L837 .

But will change both places to not call `unregisterOutboundConnection` after `stopMux`.

coot: You can remove both in a separate commit.  Both are complex `no-op`: https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs?plain=1#L2035

coot (approved):
LGTM

coot:
bors merge

karknu:
Fixes #3570

PR #3579, opened 2022-01-19
Johnalanwoods/block diffusion pipelining

johnalanwoods:
Completely refactored the architectural document on diffusion pipelining (non-heretic), and extended it to include a new implementation approach.


johnalanwoods:
The changes added to the doc are all in this commit: https://github.com/input-output-hk/ouroboros-network/pull/3579/commits/756fdd6d6c592d95066af9512d7dbba6362cc169

coot:
@johnalanwoods I think a better place for this document is `ouroboros-consensus/docs` folder, since this is likely more about consensus related changes that the pipelining made available at the networking layer.

johnalanwoods:
> @johnalanwoods I think a better place for this document is `ouroboros-consensus/docs` folder, since this is likely more about consensus related changes that the pipelining made available at the networking layer.

Moved! :)

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+}
+
+\maketitle
+
+\section{Purpose}
+The purpose of this document is to present a proposed change to the algorithm
+for Ouroboros chain diffusion that, if implemented, could facilitate higher block validation times
+for the same choice of the $\Delta$~diffusion time.
+Ergo, we would enjoy a faster $\Delta$~diffusion time in the context of our current block validation requirements.
+This additional budget may be spent on increasing block-size, and/or increasing CPU \& memory limits for Plutus scripts.
```
nfrisby, line 71: Minor: I think no comma in "block-size, and/or"

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+This document will attempt to analyse the algorithm, and review its implications. The document also discusses which
```
nfrisby, line 72: Minor: I think no comma in "algorithm, and review"

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+\end{tikzpicture}
+\end{center}
+
+\pagebreak
+
+\section{The proposed block relaying scheme}
+By propagating forward a block as soon as it has arrived locally (and its header is validated), the block 
+validation time is overlapped with the propagation time and the overall timeline becomes materially shorter.
+
+\begin{center}
```
nfrisby, line 228: I think these pictures are hard to understand. In particular, the pictures in this chapter do not show that a node is simultaneously validating the block while also transmitting the header and body. I _think_ we're only seeing the validation time and the part of the block transmission that ends after block validation completes (even though transmission might end before validation, yeah?). Relatedly: I think header transmission is not even shown (there are only 5 rectangles instead of 6 from the previous chatper's pictures).

Ideally, the picture could be enriched (eg with a thatched section implying an overlaid grey and white rectangles) (eg dividing each node's horizontal bar into three thinner lanes, one for header transmission, one for block validation, and one for block transmission -- the only rule is that block transmission comes after block validation; header transmission and block validation could and would likely begin simultaneously).

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+network graph. The goal is to prevent honest nodes from being required to do an
+unbounded amount of work, and indeed more stringently, to prevent honest nodes
+from having to do more work than the dishonest nodes engaged in an attack. Thus,
+if we can relax or defer validation without increasing the work compared with the
+normal valid case, then the risk may be acceptable. In particular, we need to avoid any
+situations whereby previously we executed a single expensive step, and now we would do many.
+
+The combination of checks 1.~and 2.~above should mean that the possibly-invalid
+block is one which was signed by the block producer (KES validation), who (in-turn) was
+authorised to create a block in this slot (VRF verification). This should mean that
```
nfrisby, line 367: > This should mean that there cannot be a greater number of invalid blocks, in total, than in the case of valid blocks.

I'm not sure I follow this sentence. I'm guessing it's the "number of" and/or "in total" part I'm not grokking.

The proposal doesn't create new opportunities for minting invalid blocks. But it does -- for the first time ever -- allow honest nodes to transmit invalid blocks.

So, what is being counted in the words "greater number"?

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+there cannot be a greater number of invalid blocks, in total, than in the case of valid blocks. In
+particular intermediate nodes cannot forge any extra or altered block(s) without
+falling afoul of these checks. Thus, it should be the case that only the
```
nfrisby, line 370: > Thus, it should be the case that only the authorised block producer can create many valid blocks, or many valid block headers with invalid block bodies.

Right, this is also something that was true before the proposal and remains true after the proposal.

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+In principle the certificate issue numbers are not bounded, but the use case
+of recovering from compromised keys does not require more than two issue
+numbers. So 'progress' in the certificate issue number can be bounded by
+allowing at most two different blocks with the same issuer in the same slot.
+There is also a 'no regression' rule with regard to the issue numbers used on the chain, so
+even if an attacker has several compromised certificates, as soon as they
+use a later one, then the earlier ones can no longer be used again.
+
+\subsection{Many valid block bodies}
+The case for many block bodies is straightforward: they make no difference to
```
nfrisby, line 437: I think this section should explicitly emphasize that different blocks have different headers, so the previous subsection already addresses every pertinent concern re DoS.

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+employ some state to temporarily store the block issuer details, then we can use that reference
+to not consider any more blocks signed by this issuer in this slot. They
+have demonstrated themselves to be adversarial by signing a bad block, and we
+shouldn't spend any further resources validating blocks from them in the short term. 
+Note that for this purpose we must identify the block issuer by the hot key certificate and 
+not the `cold' permanent identity. This is due to the fact that a compromised hot key could 
+be used to sign bad blocks, and that ought to still be a situation from which a block producer can 
+recover from by signing a new certificate with a higher issue number.
+
+It appears that it would be sufficient to hold a state of bad block issuers
```
nfrisby, line 478: > only until the current chain is successfully extended

I _think_ that rule addresses a specific nuance. And I think that nuance should be explained more explicitly.

Suppose A, B, and C are all connected to each other, all on the same chain, the adversary A is sole leader of the next slot, and the adversary A mints and sends an invalid block ab to B and mints and send a valid block ac to C. B will ban A's hot cert and then C will offer to send the (valid) block ac it got from A to B, who will refuse to download it, since the hot cert is the same as that of the ab it already discovered is invalid. This steady state is alarming, since two honest nodes B and C are now in a steady-state but on different chains (specifically, in this example, C is one block (=ac) ahead of B).

The upshot is that B may need to fetch and download a block (eg the block ac) from a hot cert it has banned, but only if a not-yet-banned issuer has extended such block (eg if C were to mint a block that extends ac).

It's not terribly clear in the current text that "until .... successfully extended" in particular means "until I receive a valid header that extends a block I had so far been refusing because of a hot cert ban".

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+the tracking set) is bounded by the leadership schedule and the honest stake
+assumption. The certificate counters are bounded (or can be bounded) as
+described in \cref{many-valid-block-headers}.
+
+\pagebreak
+
+\section{Implementation changes}
+The implementation changes required to deliver on the approach discussed in this document
+are all to be made within the Consensus layer. The discussion in this section pertains specifically 
+to the "non-aggressive" diffusion pipelining variant, the specific novel behaviour here is, nodes now
```
nfrisby, line 491: Minor: I think the comma here "variant, the specific novel" should instead be a full stop

nfrisby, line 491: Minor: I think the comma here "here is, nodes" should be replaced by the word "that"

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+forward \emph{one unvalidated block}, so that multiple nodes can download and validate the latest 
+block \emph{concurrently}. To begin, let us remind ourselves how Cardano currently diffuses data:
+
+\begin{itemize}
+  \item With Ouroboros Praos, we want the block minted in a given slot to diffuse to the leader of the next slot early enough that they in-turn can extend the chain on-top of it.
+  \item With Praos, a node will adopt the chain which is most dense and whose intersection with the previous "best" known chain is less than $k$ blocks, where $k$ is currently set to 2160.
+  \item Cognisant of the above, the process continues as follows:
+  \begin{enumerate}
+    \item A node will notify interested peers that a new block is available by posting the new header to \lstinline{chainSync}, after block validation.
```
nfrisby, line 500: Minor: we tend to write ChainSync, camel case and without any font adjustments.

For broader audiences, I would tend to write "the ChainSync mini protocol".

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+block \emph{concurrently}. To begin, let us remind ourselves how Cardano currently diffuses data:
+
+\begin{itemize}
+  \item With Ouroboros Praos, we want the block minted in a given slot to diffuse to the leader of the next slot early enough that they in-turn can extend the chain on-top of it.
+  \item With Praos, a node will adopt the chain which is most dense and whose intersection with the previous "best" known chain is less than $k$ blocks, where $k$ is currently set to 2160.
+  \item Cognisant of the above, the process continues as follows:
+  \begin{enumerate}
+    \item A node will notify interested peers that a new block is available by posting the new header to \lstinline{chainSync}, after block validation. 
+    \item After validation of the header locally, the downstream peer will request/pull the corresponding block body.
+    \item The block is validated, and if it satisfies checks resulting in a new longer/better chain, it will be appended as the head of our "new" chain.
```
nfrisby, line 502: I would say "selected" instead of "appended", since selecting it might require a switch (ie some rollback), not just extension.

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+\label{conclusion-and-considerations}
+It appears that diffusion pipelining is an exciting and potentially highly valuable 
+modification to the block propagation approach for Cardano core. There are some considerations
+though which one should bear in mind during implementation:
+
+\begin{itemize}
+  \item We must of course, ensure faster than real-time syncing of the chain. This is not an issue for the proposed implementation herein, however it's something we must be cognisant of in the future.
+  \item The changes proposed herein don’t affect the structural properties of the chain.
+  \item Intuitively one may be inclined to suggest a change to minting policy, given that with pipelining an SPO now \emph{knows} that a potentially better chain exists to extend. However, knowledge of this reality (\emph{which existed prior to pipelining}), should not alter minting behaviour. 
+  \item We should not expedite/pipeline a subsequent block until the previously pipelined block is validated, i.e. do not pipeline multiple blocks concurrently.
```
nfrisby, line 545: I think it'd help to have a second sentence here explaining why we don't think this would be worth the extra complexity.

nfrisby (approved):
I suggested some minor and medium revisions. But it's fit for purpose as-is, so also Approving! Thanks much.

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+scaling linearly with the number of hops a block travels during its diffusion across the critical path.
+Let us now consider the following implementation changes to data diffusion in the Consensus layer to achieve
+a faster block propagation and adoption time with pipelining, the goal is to provide a \emph{mechanism} by which we 
+can expedite a block's propagation prior to its validation whilst not invalidating existing protocol primitives:
+
+\begin{itemize}
+  \item The aim is to expedite data propagation where appropriate, and to do so without destructive change.
+  \begin{enumerate}
+    \item Extend the current API to pre-notify subscribed threads when a new desirable block is seen, prior to block body validation.
+    \item Implement a new mini-protocol which can function as an "expedited" version of \lstinline{chainSync}.
```
coot, line 517: There's an aspect of the argument (_not invalidating existing protocol primitives_) to keep this protocol separate from chain-sync that's worth an emphasis.  `chain-sync` will keep receiving headers which bodies are already validated. @nfrisby if this is true, will `chain-sync` need to be modified?  The `chain-sync` server will need to have access to the already validated chain of blocks, not a tentative one.

During bulk-sync phase we should use `chain-sync` stream of headers, rather than headers from the new protocol.  This means that the `FetchMode` (bluk-sync vs deadline mode) will need to be accessed by the decision logic, but that should be easy since this is already part of `NodeKernel`.

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
johnalanwoods: agreed!

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
johnalanwoods: yep I will try and add to this later

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
johnalanwoods: yep fair point

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
johnalanwoods: agreed!

johnalanwoods: agree

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
johnalanwoods: agreed!

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
johnalanwoods: agreed!

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
johnalanwoods: agreed!

johnalanwoods:
@nfrisby - all comments addressed, and code now rebased on latest master branch.

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+                              {\small \texttt{duncan.coutts@iohk.io}} \\
+   \and Matthias Fitzi     \\ {\small \texttt{matthias.fitzi@iohk.io}} \\
+   \and John Woods         \\ {\small \texttt{john.woods@iohk.io}} \\
+                              {\small \texttt{john@postquantum.dev}} \\
+}
+
+\maketitle
+
+\section{Purpose}
+The purpose of this document is to present a proposed change to the algorithm
```
dnadales, line 67:
```suggestion
The purpose of this document is to propose a change to the algorithm
```

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+assumption. The certificate counters are bounded (or can be bounded) as
+described in \cref{many-valid-block-headers}.
+
+\pagebreak
+
+\section{Implementation changes}
+The implementation changes required to deliver on the approach discussed in this document
+are all to be made within the Consensus layer. The discussion in this section pertains specifically 
+to the "non-aggressive" diffusion pipelining variant. The specific novel behaviour is that nodes now
+forward \emph{one unvalidated block}, so that multiple nodes can download and validate the latest
```
dnadales, line 492:
```suggestion
forward \emph{at most one unvalidated block}, so that multiple nodes can download and validate the latest 
```

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+\section{Implementation changes}
+The implementation changes required to deliver on the approach discussed in this document
+are all to be made within the Consensus layer. The discussion in this section pertains specifically 
+to the "non-aggressive" diffusion pipelining variant. The specific novel behaviour is that nodes now
+forward \emph{one unvalidated block}, so that multiple nodes can download and validate the latest 
+block \emph{concurrently}. To begin, let us remind ourselves how Cardano currently diffuses data:
+
+\begin{itemize}
+  \item With Ouroboros Praos, we want the block minted in a given slot to diffuse to the leader of the next slot early enough that they in-turn can extend the chain on-top of it.
+  \item With Praos, a node will adopt the chain which is most dense and whose intersection with the previous "best" known chain is less than $k$ blocks, where $k$ is currently set to 2160.
```
dnadales, line 497: Aren't we going to consider density only when we move to Genesis? I thought at the moment we only considered chains length.

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+  \item Cognisant of the above, the process continues as follows:
+  \begin{enumerate}
+    \item A node will notify interested peers that a new block is available by posting the new header to \lstinline{ChainSync} mini-protocol, after block validation. 
+    \item After validation of the header locally, the downstream peer will request/pull the corresponding block body.
+    \item The block is validated, and if it satisfies checks resulting in a new longer/better chain, it will be selected as the head of our "new" chain.
```
dnadales, line 502: The blocks might arrive out of order, in which case a successful block validation might change the current chain, or extend it, but it does not necessarily mean that the newly validated block is the "head" of our new chain. 

Also about the notation, I think we might prefer "tip" as opposed to "head" to refer to the newest element in a chain.

ouroboros-consensus/docs/diffusion-pipelining-design/block-diffusion-pipelining.tex
```diff
@@ -0,0 +1,554 @@
...
+It's clear from the above description that without pipelining we incur the block validation overhead,
+scaling linearly with the number of hops a block travels during its diffusion across the critical path.
+Let us now consider the following implementation changes to data diffusion in the Consensus layer to achieve
+a faster block propagation and adoption time with pipelining, the goal is to provide a \emph{mechanism} by which we 
+can expedite a block's propagation prior to its validation whilst not invalidating existing protocol primitives:
+
+\begin{itemize}
+  \item The aim is to expedite data propagation where appropriate, and to do so without destructive change.
+  \begin{enumerate}
+    \item Extend the current API to pre-notify subscribed threads when a new desirable block is seen, prior to block body validation.
```
dnadales, line 516: Would it make sense to clarify which API are we referring to here?

amesgen:
Block diffusion pipelining (first version) has been implemented in #3688

karknu:
@amesgen This branch contains the specification for pipelining. IMHO it would be better to have the doc updated and merged instead of having it forgotten.
I don't think there is any other documentation of pipelining.

johnalanwoods:
I agree!

amesgen:
Thanks for pointing that out, just created #4020 to track that effort :+1:

PR #3581, opened 2022-01-19
Fix: Stylish-haskell script now finds every file

Jasagredo:
As @amesgen mentioned, the glob was messing up the directories `fd` would search in.

### DIFF
--- a/scripts/ci/check-stylish.sh
+++ b/scripts/ci/check-stylish.sh
@@ -3,6 +3,6 @@
 set -euo pipefail
 
 # TODO the export of the <= operator TxLimits crashes stylish-haskell
-fd -p ouroboros-consensus* -e hs -E Setup.hs -E Ouroboros/Consensus/Mempool/TxLimits.hs -X stylish-haskell -i
+fd -p ouroboros-consensus -e hs -E Setup.hs -E ouroboros-consensus/src/Ouroboros/Consensus/Mempool/TxLimits.hs -X stylish-haskell -i
 
 git diff --exit-code


scripts/ci/check-stylish.sh
```diff
@@ -3,6 +3,6 @@
 set -euo pipefail
 
 # TODO the export of the <= operator TxLimits crashes stylish-haskell
-fd -p ouroboros-consensus* -e hs -E Setup.hs -E Ouroboros/Consensus/Mempool/TxLimits.hs -X stylish-haskell -i
+find ouroboros-consensus* -type f -name "*.hs" -not -name "Setup.hs" -not -path ouroboros-consensus/src/Ouroboros/Consensus/Mempool/TxLimits.hs -print -exec stylish-haskell -i {} \;
```
amesgen, line 6: For reference, this is how the `fd` version could be fixed:
```suggestion
fd -p ouroboros-consensus -e hs -E Setup.hs -E '**/Ouroboros/Consensus/Mempool/TxLimits.hs' -X stylish-haskell -i
```
 - The trailing * is interpreted by the shell, and expands to all directories starting with `ouroboros-consensus` *except* `ouroboros-consensus`
 - The input to `-E` are glob patterns, so as soon as the contain a `/`, they have to match the entire path, so a `**/` has to be prepended to the `TxLimits.hs` exclusion.

🤡

Jasagredo: Using this I changed `**/` for `ouroboros-consensus/src/` and it works perfectly. Thanks!

amesgen (approved):

PR #3588, opened 2022-01-26
Ignore monitoring of cold hot cons beeing demoted

karknu:
Let the con monitor ignore currently cold connections that where hot and
is in the process of beeing demoted. This means will leave it up to the
demotion job to detect the error and fail.


coot (approved):
LGTM

PR #3599, opened 2022-02-04
update to the latest version of cardano-ledger

JaredCorduan:
This update mostly includes performance improvements to the ledger and new features for the upcoming ledger era, both of which are unnoticeable to consensus and networking.

This update also includes a change to the transaction index type, `TxIx`, which is now backed by `Word16` instead of `Int`.

This update also includes new ledger events which alert on the reward calculation as the calculation is in progress, as opposed to alerting only at the end when the calculation is complete. This should be helpful to db-sync. cc @erikd


ouroboros-consensus-shelley-test/src/Test/ThreadNet/TxGen/Shelley.hs
```diff
@@ -107,7 +108,9 @@ genTx _cfg slotNo TickedShelleyLedgerState { tickedShelleyLedgerState } genEnv =
     ledgerEnv :: SL.LedgerEnv (MockShelley h)
     ledgerEnv = SL.LedgerEnv {
         ledgerSlotNo   = slotNo
-      , ledgerIx       = 0 -- TODO Ix
+      -- The use of 'mkTxIxPartial' is acceptible in test code,
+      -- especially below where we know that 0 is in the domain.
+      , ledgerIx       = SL.mkTxIxPartial 0
```
lehins, line 113: This way there is no need for a comment ;) 
```suggestion
      , ledgerIx       = minBound
```

JaredCorduan: much better, thank you @lehins

nfrisby (approved):
Very tiny diff. Approved! Thanks.

JaredCorduan:
bors merge

nc6:
bors retry

nc6:
bors retry

JaredCorduan:
It looks like the ledger state serialization may have changed, I will investigate.

JaredCorduan:
I have figured out where the change was in the ledger state serialization. We had to change the type of the answer computed by the reward pulser in order to produce the new incremental reward events.  See https://github.com/input-output-hk/cardano-ledger/pull/2615/files#diff-8c3fd71a89804a4cb9120c509e7561def5b5cd65c4b2af88b41b8a7f43e4069eR86-R87

Along the way, I spotted two mistakes that I would like to correct before we merge this PR. Both have to do with a bogus list length in the cbor encoding. The mistake made it past our round trip tests since they were self consistent, albeit wrong. see [this](https://github.com/input-output-hk/cardano-ledger/blob/aebd64e015ec0825776c256faed9d8632712beb0/eras/shelley/impl/src/Cardano/Ledger/Shelley/LedgerState.hs#L393) and [this](https://github.com/input-output-hk/cardano-ledger/blob/aebd64e015ec0825776c256faed9d8632712beb0/eras/shelley/impl/src/Cardano/Ledger/Shelley/PoolRank.hs#L252).

JaredCorduan:
It would be great if https://github.com/input-output-hk/cardano-ledger/pull/2651 could also make its way into this update as well.

nc6:
bors merge

JaredCorduan:
thank you for pushing this along @nc6 !

PR #3603, opened 2022-02-08
IOSimPOR - race exploration with partial order reduction

coot:
Subsumes #3463.

This branch backports `IOSim` fixes to `IOSimPOR`, introduces various minor
modifications most of which were discussed in #3463.  Each patch can be
reviewed seprately.

- Add IOSimPOR, an IO simulator that can explore race conditions
- Adapt one of the governor tests to use IOSimPOR
- See no evil, hear no evil, speak no evil
- io-sim-por: share a thunk in prop_shrinkCarefully
- io-sim-por: introduced InternalTypes module
- io-sim-por: fixed a typo
- io-sim-por: refactor happensBeforeStep
- io-sim-por: stylistic changes
- io-sim-por: removed an if clause
- io-sim-por: refactored advanceControl
- io-sim-por: renamed lubVClock as leastUpperBoundVClock
- io-sim-por: renamed ThreadId constructors
- io-sim-por: introduced insertThread
- ouroboros-network: nightly cabal flag
- io-sim-por: added haddocks & notes
- io-sim-por: shuffle module structure
- ouroboros-network: added a todo
- io-sim-por: backport CancelTimeout handling
- io-sim: backport masking state fix
- io-sim-por: backport logging of MaskingState

### DIFF
--- a/io-classes/io-classes.cabal
+++ b/io-classes/io-classes.cabal
@@ -43,6 +43,7 @@ library
                        Control.Monad.Class.MonadThrow
                        Control.Monad.Class.MonadTime
                        Control.Monad.Class.MonadTimer
+                       Control.Monad.Class.MonadTest
   default-language:    Haskell2010
   other-extensions:    CPP
                        TypeFamilies
--- a/io-classes/src/Control/Monad/Class/MonadTest.hs
+++ b/io-classes/src/Control/Monad/Class/MonadTest.hs
@@ -1,0 +1,13 @@
+module Control.Monad.Class.MonadTest ( MonadTest (..) ) where
+
+import           Control.Monad.Reader
+
+class Monad m => MonadTest m where
+  exploreRaces :: m ()
+  exploreRaces = return ()
+
+instance MonadTest IO
+
+instance MonadTest m => MonadTest (ReaderT e m) where
+  exploreRaces = lift exploreRaces
+
--- a/io-sim/how-to-use-IOSimPOR.md
+++ b/io-sim/how-to-use-IOSimPOR.md
@@ -51,22 +51,28 @@ The IOSimPOR API
 
 The common way to use IOSim is to call
 
-   runSimTrace :: forall a. (forall s. IOSim s a) -> Trace a
+```hs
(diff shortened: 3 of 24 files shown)


rjmh (commented):
There's a reference to a "test thread" in a comment on line 1361... maybe change it to "non-racy thread"?

rjmh (commented):
This haddock comment:
    ControlAwait [ScheduleMod]
                     -- ^ if the current control is 'ControlAwait', it will
                     -- make the thread (found in the first `ScheduleMod`) to
                     -- sleep until its 'threadStepId' matches.
isn't accurate. The effect of ControlAwait is to allow normal scheduling to proceed, until the thread found in the first ScheduleMod reaches the given threadStepId. At that point the thread is put to sleep, until after the steps in the ScheduleMod have been followed.

rjmh (commented):
I don't think this can be right. With the change, CancelTimeout can wake up other threads, but the effect recorded here only includes the write effect to the TVars, not the wakeup effect. I would expect this to cause replay to get out of sync in some cases, because if the first step of a woken thread races with other actions in the same step of the thread that called cancelTimeout, then race reversal might try to reverse those steps. That may fail, because the second thread won't be running before the cancelTimer has taken place. This would show up as a test failure caused by a failed assertion in reschedule, when we try to schedule a thread that is not in the runqueue.

I think the correct change is just to add wakeupEffects to the step effect, for the threads that are woken.

rjmh:
I've been through all the changes now, and I don't see any other problems. The only really important thing was to track thread wake-up effects in the new version of CancelTimeout.

rjmh:
...and I've now realized the suggestion I made for fixing the new CancelTimeout is not correct, or at least, not sufficient.

When threads are woken up by cancelling a timeout, then their thread vector clocks must be advanced by lubbing with the vector clock of the thread that cancelled the timer, because the timer cancellation "happens before" the next step in each of those threads. It's not enough just to add a wakeup effect, unfortunately, we have to explicitly update the vector clocks too.

coot:
bors merge

PR #3604, opened 2022-02-08
Docs: Snapshots and LedgerDBs structure

Jasagredo:
- Document what snapshots have inside and what is the decision process for creating snapshots one way or the other. 
- Document how the LedgerDBs look like inside, following the already existing comment for the legacy DB. 
- Rename some sections in `InMemory` as they were clearly outdated and mixed up.


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -116,16 +116,7 @@ import           Ouroboros.Consensus.Util.Versioned
...
-                               (WithOrigin SlotNo)
-                               (Checkpoint (l ValuesMK))
-                               (Checkpoint (l ValuesMK))
-
-type ModernLedgerDB l = DbChangelog l
-
--- | Internal state of the ledger DB
---
--- The ledger DB looks like
+-- | The ledger DB looks like
```
dnadales, line 119: "The legacy ledger DB looks" like?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
 --
 -- > L3 |> []
+type LegacyLedgerDB l = AnchoredSeq
+                               (WithOrigin SlotNo)
+                               (Checkpoint (l ValuesMK))
+                               (Checkpoint (l ValuesMK))
+
+-- | Type alias for the new-style database
+--
+-- This type is internally a DbChangelog which as an example, with @k = 3@ could look like:
```
dnadales, line 165:
```suggestion
-- This type is a 'DbChangelog'. We illustrate its contents below, where @k = 3@:
```

dnadales, line 165: I wonder if this comment belongs in the definition of `DbChangelog`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
+--
+-- > stateAnchor   | diskAnchor | states                     | tableDiffs
+-- > ----------------------------------------------------------------------------
+-- >      0        |      0     | [ L0]                     | [ ]
```
dnadales, line 169: Spacing: we should be consistent: either we use a space at the beginning and at the end or no space at all.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
+-- >      0        |      0     | [ L0, L1]                 | [ D1]
+-- >      0        |      0     | [ L0, L1, L2]             | [ D1, D2]
+-- >      0        |      0     | [ L0, L1, L2, L3]         | [ D1, D2, D3]
+-- >      1        |      0     | [ L0, L1, L2, L3, L4]     | [ D1, D2, D3, D4]
+-- >      2        |      0     | [ L0, L1, L2, L3, L4, L5] | [ D1, D2, D3, D4, D5]    (*)
+-- >      2        |      2     | [ L2, L3, L4, L5]         | [ D3, D4, D5]   -- flush (**)
+-- >      3        |      2     | [ L2, L3, L4, L5, L6]     | [ D3, D4, D5, D6]
+--
+-- The disk anchor moves when we flush data to the disk, and the state anchor
```
dnadales, line 178:
```suggestion
-- The disk anchor moves when we flush data to disk, and the state anchor
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
+-- points always to the state that represents the tip of the logical immutable
+-- database. Notice that @seqNo (last states) - stateAnchor@ is usually @k@
+-- (unless rollbacks or corruption take place). We cannot roll back more than
```
dnadales, line 181:
```suggestion
-- except when rollbacks or data corruption take place. We cannot roll back more than
```

and shouldn't we mention that it'll be `<k` if we replay from a snapshot, or if we're simply starting from Genesis?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
+-- @k@ blocks, this means that under a rollback of @k@ blocks at (*), the
```
dnadales, line 182:
```suggestion
-- @k@ blocks. This means that after a rollback of @k@ blocks at (*), the
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
+--
+-- >      2        |      0     | [ L0, L1, L2]             | [ D1, D2]
+--
+-- And a rollback of @k@ blocks at (**) will look something like this:
+--
+-- >      2        |      0     | [ L2]                     | [ ]
+--
+-- Notice how the states list always contains the in-memory state of the anchor,
+-- but the table differences might not contain the differences for that anchor
+-- if they have not been flushed to the backend.
```
dnadales, line 193: If they have or have not? I thought that the differences won't be there if we flush them to disk.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
+-- database will look something like this:
+--
+-- >      2        |      0     | [ L0, L1, L2]             | [ D1, D2]
+--
+-- And a rollback of @k@ blocks at (**) will look something like this:
+--
+-- >      2        |      0     | [ L2]                     | [ ]
+--
+-- Notice how the states list always contains the in-memory state of the anchor,
+-- but the table differences might not contain the differences for that anchor
```
dnadales, line 192: We might need to define or exemplify what is the difference for the anchor. I guess if the anchor is `Li` the corresponding difference is `Di`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
+-- if they have not been flushed to the backend.
+--
+-- The new-style database has to be coupled with the @BackingStore@ which
+-- provides the pointers to the on-disk data.
+type ModernLedgerDB l = DbChangelog l
```
dnadales, line 197: Why not simply `LedgerDB`?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -164,9 +155,53 @@ type ModernLedgerDB l = DbChangelog l
...
+--
+-- And a rollback of @k@ blocks at (**) will look something like this:
+--
+-- >      2        |      0     | [ L2]                     | [ ]
+--
+-- Notice how the states list always contains the in-memory state of the anchor,
+-- but the table differences might not contain the differences for that anchor
+-- if they have not been flushed to the backend.
+--
+-- The new-style database has to be coupled with the @BackingStore@ which
```
dnadales, line 195: I wonder if this comment is relevant here: there's no reference to the backing store, or to the pointers to the on-disk data.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -262,19 +282,39 @@ class Monad m => ResolvesBlocks m blk | m -> blk where
 instance Monad m => ResolvesBlocks (ReaderT (ResolveBlock m blk) m) blk where
   resolveBlock r = ReaderT $ \f -> f r
 
+-- Quite a specific instance so we can satisfy the fundep
```
dnadales, line 285: Should we clarify which fundep we're referring to?

dnadales, line 285:
```suggestion
-- Specific instance so we can satisfy the functional dependency
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -789,8 +904,8 @@ data TraceReplayEvent blk =
     -- | There was a LedgerDB snapshot on disk corresponding to the given tip.
     -- We're replaying more recent blocks against it.
   | ReplayFromSnapshot
-        DiskSnapshot
-        (RealPoint blk)
+        DiskSnapshot -- ^ The snapshot we are replaying from
```
dnadales, line 907: Alignment?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
 {-# LANGUAGE TupleSections              #-}
 {-# LANGUAGE TypeApplications           #-}
 
+{-|
+  = Snapshot managing and contents
+
+  Snapshotting a ledger state means saving a copy of the in-memory part of the
+  ledger state serialized as a file in the disk, as well as flushing differences
```
dnadales, line 24:
```suggestion
  ledger state serialized as a file on disk, as well as flushing differences
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+  between the last snapshotted ledger state and the one that we are snapshotting
+  now onto the BackingStore. While the coupled implementation is in place, the
```
dnadales, line 26:
```suggestion
  now. While the coupled implementation is in place, the
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+        >     │   ├── tables
+        >     │   └── state
+        >     ├── <slotNumber2>
+        >     │   ├── tables
+        >     │   └── state
+        >     └── <slotNumber3>
+        >         ├── tables
+        >         └── state
+
+  2. Depending on the snapshots found, there are two paths:
```
dnadales, line 51:
```suggestion
  2. Depending on the snapshots found, there are two possibilities:
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+
+       - If there is no snapshot to load, create a new @BackingStore@ with the
```
dnadales, line 53: If we use single quotes I think the haddock will reference to the `BackingStore` haddocks.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+              contains a full LedgerState snapshot in the same style as before
+              the UTxO HD implementation (i.e. in-memory part of the LedgerState
+              (equal among implementations) + UTxO map).
+
+            - Otherwise, it contains a serialization of the in-memory part of the
+              LedgerState, without a UTxO map.
+
+        In both cases, the snapshot will be deserialized as @ExtLedgerState blk
+        EmptyMK@. In case we found an snapshot, we will overwrite the
+        @BackingStore@ tables with the @tables@ file from said snapshot.
```
dnadales, line 69: Maybe we should mention briefly the abstraction that the backing store represents, and what is its relation w.r.t. the tables it contains. Also shouldn't we describe in the beginning how we initialize this backing store we're overwriting in this step? Is it an empty one? Or one that is in whatever state it was left when the node was shut down? 

Another question is: if we're overwriting the tables of the backing store, would it make sense to always start with an initialized backing store? (eg an empty one)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+  3. Depending on whether the node is requested to run both databases and whether
+     or not the snapshot has a UTxO inside, the node will:
+
+       - __Run both__ + snapshot has UTxO inside: Extract the tables from the
+         deserialized snapshot to create a @ExtLedgerState blk ValuesMK@ which
+         can be 1. used in the legacy database and 2. after stripping out the
+         @ValuesMK@, used in the modern database.
+
+       - __Run both__ + snapshot doesn't have UTxO inside: an error will be
+         thrown as we cannot/don't think we should reconstruct an in-memory UTxO
```
dnadales, line 80:
```suggestion
         thrown as we cannot reconstruct an in-memory UTxO
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+         directly from the @BackingStore@.
+
+       - Otherwise: Strip out the values inside the @ExtLedgerState blk EmptyMK@
+         to make sure that we don't have an in-memory UTxO map and use it as the
+         anchor for the modern database.
```
dnadales, line 85: How can we distinguish the case in which we have an empty UTxO map just because there are no transactions (admittedly a pathological case) and the case in which the map is empty because we didn't store the legacy ledger state?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+  == Taking snapshots at startup
+
+  Due to the nature of the modern database having to carry around all the
+  differences between the last snapshotted state and the current tip, there is a
+  need to flush when replaying the chain as otherwise, for example on a replay
+  from genesis to the tip, we would carry millions of differences in memory.
+
+  Because of this, when we are replaying blocks we will flush regularly. As the
+  last snapshot that was taken lives in a @\<slotNumber\>/tables@ file, there is
+  no risk of destroying it by flushing. Only when we finish replaying blocks and
```
dnadales, line 129: Sorry, I don't get this: how can we destroy a snapshot by flushing?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+  start the background threads (and specifically the @copyAndSnapshotRunner@),
+  we will take a snapshot of the current immutable database anchor as described
+  above.
```
dnadales, line 132: Is it possible that in the future we take snapshots while replaying?

dnadales: Actually, I wonder why not implement this behaviour now :thinking:

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,121 @@
...
+
+       - If the legacy database is enabled, the state to serialize is the anchor
+         of the legacy database (which should be the same as the one from the
+         modern database) with the tables stowed.
+
+       - Otherwise, the state to serialize is the modern database anchor which
+         requires no further modifications.
+
+  3. A directory is created named after the slot number of the ledger state
+     being snapshotted, and the serialization from the previous step is written
```
dnadales, line 112: Isn't `3` part of `2` above?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: DbChangelog to me is just a data structure. It is this module the one that defines how it works and is used.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: I was consistent with what was already there, spaces on the left, no spaces on the right

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: :+1: to both

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: If they *have not*, well spotted

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: Added

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: Because `LedgerDB` is already taken by the `LedgerDB` datatype. Or you mean something different?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: I think it is relevant. The backing store is used in this module, and although not mentioned here, it has to be known and understood by the reader in order to use this module. And I don't think we should explain here what it is, that should live within the BackingStore definition I think

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: :man_shrugging: I just moved this instance from some lines below.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
Jasagredo: This one
> Or one that is in whatever state it was left when the node was shut down?
Not sure if we should mention that here though. I can see benefits on doing it and also on not doing it but instead adding that to the BackingStore definition

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
Jasagredo: The UTxO map would be the one from the legacy ledger and the only possible case in which it would be empty is if every account had a balance of `0` which hopefully does never happen.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
Jasagredo: Well, this precisely says that we cannot. I added a small comment to clarify, but it means that "each snapshot is in a separate directory so when we store a snapshot we won't overwrite other snapshots files"

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
Jasagredo: Could be, don't see why not

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
Jasagredo: Not to me. 2 specifies what is going to be serialized, and 3 says where those things will be placed in the file system.

dnadales (approved):

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
nfrisby: I intended to rename `LedgerDB` shortly after we merge. Every other `DB` in our code is a mutable thing, so I don't want the name of an immutable data structure to end with `DB`. It causes plenty of confusion. Even more so since "ledger dee bee" is a reasonable pronunciation of both `LedgerDB` and or `LgrDB` .

(But I don't want to do that refactoring on our feature branch. Too conflicty.)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,122 @@
 {-# LANGUAGE TupleSections              #-}
 {-# LANGUAGE TypeApplications           #-}
 
+{-|
+  = Snapshot managing and contents
+
+  Snapshotting a ledger state means saving a copy of the in-memory part of the
+  ledger state serialized as a file on disk, as well as flushing differences
+  between the last snapshotted ledger state and the one that we are snapshotting
```
nfrisby, line 25: The snapshot also contains a copy (aka "restore point" aka `tables` below) of the matching on-disk state. Just flushing is not enough, we need to a make a copy that won't be altered by subsequent rights.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,122 @@
...
+       - If there is a snapshot found, then deserialize the @state@ file, which
+         contains either:
+
+            - If the node was running __with the legacy database enabled__, it
+              contains a full LedgerState snapshot in the same style as before
+              the UTxO HD implementation (i.e. in-memory part of the LedgerState
+              (equal among implementations) + UTxO map).
+
+            - Otherwise, it contains a serialization of the in-memory part of the
+              LedgerState, without a UTxO map.
```
nfrisby, line 65: `without a UTxO map` -> `with an empty UTxO map` (at least until Ledger changes their types)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,122 @@
...
+
+        In both cases, the snapshot will be deserialized as @ExtLedgerState blk
+        EmptyMK@. In case we found an snapshot, we will overwrite the
```
nfrisby, line 68: `we will overwrite`: eg using `rm -f tables; cp <snapshot>/tables -r .` is one option that ought to work. But I think instead the backing store implementation should be the one to create copies and load in/reset to/restore from a copy. EG some backing store (eg RocksDB, if I recall correctly) might be able to only store a "restore point" identifier in the snapshot's `tables` file and the restore operation would just be a more efficient reset to the identified former state.

"we will overwrite" might be simple enough wording, but only if it isn't taken literally.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,122 @@
...
+
+  3. Depending on whether the node is requested to run both databases and whether
+     or not the snapshot has a UTxO inside, the node will:
+
+       - __Run both__ + snapshot has UTxO inside: Extract the tables from the
+         deserialized snapshot to create a @ExtLedgerState blk ValuesMK@ which
+         can be 1. used in the legacy database and 2. after stripping out the
+         @ValuesMK@, used in the modern database.
+
+       - __Run both__ + snapshot doesn't have UTxO inside: an error will be
```
nfrisby, line 80: `doesn't have UTxO inside` -> `has an empty UTxO map`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,122 @@
...
+  dictates to do so. Whenever the chain grows past @k@ blocks, said runner will
+  copy the blocks which are more than @k@ blocks from the tip (i.e. the ones
+  that must be considered immutable) to the immutable database and then:
+
+  1. Flush differences in the modern DB up to the immutable db tip.
+
+  2. If dictated by the disk policy, serialize an @ExtLedgerState blk EmptyMK@ containing:
+
+       - If the legacy database is enabled, the state to serialize is the anchor
+         of the legacy database (which should be the same as the one from the
```
nfrisby, line 106: `which should be the same` -> `which, because of the just-prior flush, should be the same`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -17,6 +17,122 @@
...
+     A directory is created named after the slot number of the ledger state
+     being snapshotted, and the serialization from above is written into the
+     @\<slotNumber\>/state@ file and the @BackingStore@ tables are copied into
+     the @\<slotNumber\>/tables@ file.
+
+  3. There is a maximum number of snapshots that should exist in the disk at any
+     time, dictated by the @DiskPolicy@, so if needed, we will trim out old
+     snapshots.
+
+  == Taking snapshots at startup
```
nfrisby, line 121: `at startup` -> `during startup`

nfrisby: After reading it, I see that this section explains that we only take one snapshot, when we're done replaying.

Maybe the section title should be "Flush during startup and snapshot at the end of startup".

Want to mention that we _could_ take snapshots during replay, but haven't yet decided on that?

Jasagredo:
Github didn't allow me to answer to your comments, @nfrisby , but I added all of them :+1:

Jasagredo:
After approval by @dnadales and @nfrisby , I merge this into the `inmem` branch

PR #3607, opened 2022-02-09
DRAFT Dougwilson/cad 3875 integrate lmdb into consensus

duog:
Work in progress.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
@@ -215,6 +215,13 @@ instance (SingleEraBlock x, TableStuff (LedgerState x)) => TableStuff (LedgerSta
   mapLedgerTables  f = coerce $ mapLedgerTables  @(LedgerState x) f
   zipLedgerTables  f = coerce $ zipLedgerTables  @(LedgerState x) f
   foldLedgerTables f = coerce $ foldLedgerTables @(LedgerState x) f
+  traverseLedgerTables f (LedgerTablesOne tbls)
+    = LedgerTablesOne <$> traverseLedgerTables f tbls
+  zip2ALedgerTables f
+    (LedgerTablesOne tbls1)
+    (LedgerTablesOne tbls2)
+    = LedgerTablesOne <$> zip2ALedgerTables f tbls1 tbls2
+  namesLedgerTables = coerce $ namesLedgerTables @(LedgerState x)
 
 deriving instance Eq (LedgerTables (LedgerState x) mk) => Eq (LedgerTables (LedgerState (HardForkBlock '[x])) mk)
 
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
@@ -139,7 +139,9 @@ instance ( CanHardFork xs
          , NoThunks          (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
          , FromCBOR (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
          , ToCBOR   (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
-         , StowableLedgerTables (LedgerState (HardForkBlock xs))
+         , PreApplyBlock (LedgerState (HardForkBlock xs)) (HardForkBlock xs)
+         , PreLedgerSupportsMempool (HardForkBlock xs)
+         , TickedTableStuff (LedgerState (HardForkBlock xs))
          )
   => IsLedger (LedgerState (HardForkBlock xs)) where
   type LedgerErr (LedgerState (HardForkBlock xs)) = HardForkLedgerError  xs
@@ -270,14 +272,6 @@ instance
         versionZeroProductFromCBOR "LedgerTablesOne" 1
       $ LedgerTablesOne <$> fromCBOR
 
-instance
-     StowableLedgerTables (LedgerState x)
-  => StowableLedgerTables (LedgerState (HardForkBlock '[x])) where
-  stowLedgerTables   st =
-      st `withOneState` stowLedgerTables (projectOneState st)
-  unstowLedgerTables st =
-      st `withOneState` unstowLedgerTables (projectOneState st)
-
 {-------------------------------------------------------------------------------
   ApplyBlock
 -------------------------------------------------------------------------------}
@@ -287,7 +281,8 @@ instance ( CanHardFork xs
          , NoThunks (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
          , FromCBOR (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
          , ToCBOR   (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
-         , StowableLedgerTables (LedgerState (HardForkBlock xs))
+         , TickedTableStuff (LedgerState (HardForkBlock xs))
+         , PreLedgerSupportsMempool (HardForkBlock xs)
          , PreApplyBlock (LedgerState (HardForkBlock xs)) (HardForkBlock xs)
          )
       => ApplyBlock (LedgerState (HardForkBlock xs)) (HardForkBlock xs) where
@@ -369,8 +364,8 @@ instance ( CanHardFork xs
          , NoThunks (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
          , FromCBOR (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
          , ToCBOR   (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
-         , StowableLedgerTables (LedgerState (HardForkBlock xs))
          , PreApplyBlock (LedgerState (HardForkBlock xs)) (HardForkBlock xs)
+         , PreLedgerSupportsMempool (HardForkBlock xs)
          ) => UpdateLedger (HardForkBlock xs)
 
 {-------------------------------------------------------------------------------
@@ -447,7 +442,6 @@ type CanHardFork' xs =
   , NoThunks (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
   , FromCBOR (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
   , ToCBOR   (LedgerTables (LedgerState (HardForkBlock xs)) ValuesMK)
-  , StowableLedgerTables (LedgerState (HardForkBlock xs))
   , PreApplyBlock (LedgerState (HardForkBlock xs)) (HardForkBlock xs)
   , PreLedgerSupportsMempool (HardForkBlock xs)
   )
@@ -514,6 +508,34 @@ instance CanHardFork' xs => LedgerSupportsProtocol (HardForkBlock xs) where
         where
           cfg' :: LedgerConfig blk
           cfg' = completeLedgerConfig' ei cfg
+
+instance
+     CanHardFork' xs
+  => StowableLedgerTables (LedgerState (HardForkBlock xs)) where
+  stowLedgerTables st =
+      withLedgerTables
+        (HardForkLedgerState $ HardForkState tele')
+        emptyLedgerTables
+    where
+      HardForkLedgerState (HardForkState tele) = st
+
+      tele' =
+        hcmap
+          proxySingle
+          (\current -> current{currentState = Flip $ stowLedgerTables $ unFlip $ currentState current})
+          tele
+  unstowLedgerTables st =
+      withLedgerTables
+        (HardForkLedgerState $ HardForkState tele')
+        emptyLedgerTables
+    where
+      HardForkLedgerState (HardForkState tele) = st
+
+      tele' =
+        hcmap
+          proxySingle
+          (\current -> current{currentState = Flip $ unstowLedgerTables $ unFlip $ currentState current})
+          tele
 
 {-------------------------------------------------------------------------------
   Annotated forecasts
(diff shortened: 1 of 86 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Ledger.hs
```diff
@@ -215,6 +215,13 @@ instance (SingleEraBlock x, TableStuff (LedgerState x)) => TableStuff (LedgerSta
   mapLedgerTables  f = coerce $ mapLedgerTables  @(LedgerState x) f
   zipLedgerTables  f = coerce $ zipLedgerTables  @(LedgerState x) f
   foldLedgerTables f = coerce $ foldLedgerTables @(LedgerState x) f
+  traverseLedgerTables f (LedgerTablesOne tbls)
+    = LedgerTablesOne <$> traverseLedgerTables f tbls
+  zip2ALedgerTables f
```
nfrisby, line 220: Minor: I would expect `zipLedgerTablesA` based on the existing patterns in `base` eg

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -288,6 +292,10 @@ applyChainTick ::
   -> Ticked1 l mk
 applyChainTick = lrResult ..: applyChainTickLedgerResult
 
+
+class (Ord k, Serialise k, Serialise v) => LedgerConstraint k v
```
nfrisby, line 296: Before we merge, we should avoid using `Serialise` for this.

My suggestion is a newtype (like my `LmdbBox` type from the db-analyser patch branch you saw) that's wrapped and unwrapped immediately before passing/after receiving data from `lmdb-simple`. That newtype would have a `Serialise` instance (since `lmdb-simple` demands it) that directly defers to our preferred class (which can start as `ToCBOR` and `FromCBOR`, but maybe should ultimately be our `EncodeDisk` class? I'm unsure).

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -498,6 +539,8 @@ data ApplyMapKind :: MapKind -> Type -> Type -> Type where
   ApplyTrackingMK :: !(UtxoValues  k v) -> !(UtxoDiff k v) -> ApplyMapKind TrackingMK   k v
   ApplyValuesMK   :: !(UtxoValues  k v)                    -> ApplyMapKind ValuesMK     k v
   ApplyRewoundMK  :: !(RewoundKeys k v)                    -> ApplyMapKind RewoundMK    k v
+  ApplyLMDBMK     :: !String -> !(LMDB.Database k v)       -> ApplyMapKind LMDBMK       k v
```
nfrisby, line 542: Hmm. We probably can't serialize this, so the closed instance for `ApplyMapKind` will probably need to be divided into several open instances now. 😬

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Dual.hs
```diff
@@ -427,6 +428,21 @@ instance Bridge m a => TableStuff (LedgerState (DualBlock m a)) where
...
+
+  zip2ALedgerTables
+    f
+    (DualBlockLedgerTables mainL auxL)
+    (DualBlockLedgerTables mainR auxR) =
+      liftA2 DualBlockLedgerTables
+        (zip2ALedgerTables f mainL mainR)
+        (zip2ALedgerTables f auxL  auxR)
+
+  namesLedgerTables = error "DualBlock:namesLedgerTable:not sure how to combine these"
```
nfrisby, line 444: I think:

```
zipLedgerTables (\(ApplyNameMK l) (ApplyNameMK r) -> ApplyNameMK $ "Dual-" <> l <> "-" <> r) (namesLedgerTables @m) (namesLedgerTables @a)
```

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -489,6 +528,8 @@ data MapKind = DiffMK
              | SeqDiffMK
              | TrackingMK
              | ValuesMK
+             | LMDBMK
+             | NamesMK
```
nfrisby, line 532: Minor: I suggest `NameMK` instead of the plural. Thought process: it's one name per table.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Extended.hs
```diff
@@ -189,9 +189,13 @@ instance (LedgerSupportsProtocol blk, TableStuff (LedgerState blk)) => TableStuf
       ExtLedgerState (lstate `withLedgerTables` tables) hstate
 
   pureLedgerTables f = coerce $ pureLedgerTables @(LedgerState blk) f
-  mapLedgerTables  f = coerce $ mapLedgerTables  @(LedgerState blk) f
   zipLedgerTables  f = coerce $ zipLedgerTables  @(LedgerState blk) f
   foldLedgerTables f = coerce $ foldLedgerTables @(LedgerState blk) f
+  traverseLedgerTables f (ExtLedgerStateTables tbls) =
```
nfrisby, line 194: You could also `coerce` these two, I think.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/FS/API.hs
```diff
@@ -147,6 +147,9 @@ data HasFS m h = HasFS {
 
     -- | Useful for better error reporting
   , mkFsErrorPath            :: FsPath -> FsErrorPath
+
+    -- | TODO
+  , unsafeToFilePath         :: FsPath -> m FilePath
```
nfrisby, line 152: Yeah, I haven't thought through this wrinkle yet :(

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,372 @@
...
+    Just v -> Map.insert k v m
+
+writeLMDBTable :: (Serialise k, Serialise v)
+  => LMDB.Database k v
+  -> HD.UtxoDiff k v
+  -> LMDB.Transaction LMDB.ReadWrite ()
+writeLMDBTable db (HD.UtxoDiff m) = void $ Map.traverseWithKey go m where
+  go k (HD.UtxoEntryDiff v reason) = LMDB.put db k new_val where
+    new_val
+      | reason `elem` [HD.UedsDel, HD.UedsInsAndDel]
```
nfrisby, line 113: `UedsInsAndDel` can be a no-op; no `put` call necessary.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,372 @@
...
+  FS.doesFileExist fs path >>= \b -> when b $ Exn.throw $ DbErrStr $ "guardDbDir:must be a directory:" <> show path
+  FS.doesDirectoryExist fs path >>= \case
+    True | GDDMustNotExist <- gdd ->Exn.throw $ DbErrStr $ "guardDbDir: Must not exist: " <> show path
+    _ -> pure ()
+  FS.createDirectoryIfMissing fs True path
+  FS.unsafeToFilePath fs path
+
+defaultLMDBLimits :: LMDB.Limits
+defaultLMDBLimits = LMDB.defaultLimits
+  { LMDB.mapSize = 6_000_000_000
```
nfrisby, line 187: We'll need to figure out how to maintain this `mapSize` 😬 Maybe the user should set it in a config file?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,372 @@
...
+
+  pure HD.BackingStore{..}
+
+
+-- | ValueHandles hold an Async which is holding a transaction open.
+-- We maintain a map of them in the Db, (keyed on their Seq No) and
+-- they have a reference counting TVar.
+-- This means that multiple BackingStoreValueHandles will reference the
+-- same ValueHandle and thus we keep only one transaction per seq no alive.
+data ValueHandle = ValueHandle
```
nfrisby, line 282: Is there a limit on how many read txs can be open? I don't expect there will be many value handles open at a time. All the ones I've used so far should be ephemeral, except for the `LocalStateQuery` mini protocol server, but there should only ever be one of those, and it's a trusted client, so we can expect they won't hold the vh open too long.

So maybe the sharing and refcounting is unnecessary complexity?

(I can imagine other reasons it's necessary; just hoping for simpler 😄.)

nfrisby: If it's just because `lmdb-simple` doesn't export its internals, then maintaining a simple fork that just exposes its internal modules instead of jumping through this kind of `Async` hoop might be worth a while soon enough.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -801,3 +813,7 @@ data TraceReplayEvent blk =
         (ReplayStart blk) -- ^ the block at which this replay started
         (ReplayGoal blk)  -- ^ the block at the tip of the ImmutableDB
   deriving (Generic, Eq, Show)
+
+data BackingStoreSelector m where
+  LMDBBackingStore :: !FsPath -> BackingStoreSelector IO
```
nfrisby, line 818: Ah ha, I was worrying about how to expose `IO` for LMDB C calls. I hadn't though of using the selector to expose `IO`. Nice and tidy! 👏

nfrisby (commented):
Looking good! Thanks very much for sharing the Draft.

I few minor things, and a prompted for a couple possible simplifications.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
duog: Are you sure it needs to be serialised?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
duog: Yes, the whole limits needs to be configurable. I had thought this could perhaps come in from a config file through the BackendSlector

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Dual.hs
duog: Unfortunately I don't think this quite works. the two namesLedgerTables are different data types, they may well have different numbers and types of tables.

duog: I'm going with `namesLedgerTables = DualBlockLedgerTables namesLedgerTables namesLedgerTables`  for now

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
nfrisby: Right now, the in-memory back-end serialises a `ValuesMK`. But we could probably arrange it so that we don't propagate a `ToCBOR (ApplyMapKind ...)` constraint of any kind.

In which case, yes, no instances open-or-closed needed for serializing `ApplyMapKind`

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Dual.hs
nfrisby: > Unfortunately I don't think this quite works. the two namesLedgerTables are different data types, they may well have different numbers and types of tables.

Hmm, that's true in general, yeah. However, I think for our intended uses case (there's a [simulation](https://en.wikipedia.org/wiki/Simulation_(computer_science)) between the two blocks), it's fair -- especially for now -- to assume that they have the same tables.

That said, I think what you went with already matches the existing `LedgerTables (LedgerState (DualBlock m a)) mk` instance perfectly.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
nfrisby: For the UTxO map, if there is an existing value on disk, then no valid diff can involving inserting at that key. So those three steps as listed cannot arise from a valid chain suffix (which these must be, since they're being placed in the immutable ChainDB).

Does that logic seem sound to you?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
nfrisby: Yeah, `MonadIO` is fine instead of `IO`. The other args make sense too.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
nfrisby: (I'm leaving this Conversation open as a TODO.)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
nfrisby: (I'm leaving this Conversation open as a TODO.)

dnadales:
> Work in progress.

Would it make sense to simply mark the PR as draft in github?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -489,6 +528,8 @@ data MapKind = DiffMK
              | SeqDiffMK
              | TrackingMK
              | ValuesMK
+             | LMDBMK
```
dnadales, line 531: Is this going to receive a less specific name or be replaced by something else?

Jasagredo:
I think we can close this one as it was superseeded by #3710 
@dnadales @jorisdral

PR #3624, opened 2022-02-16
Fix: `test-storage:LedgerDB` test suite

Jasagredo:
Adds the `dbBackingStore` field to the database used in the tests plus integrates it in the operations.

### DIFF
--- a/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
+++ b/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
@@ -282,9 +282,9 @@ fromInitLog  InitFromGenesis          = MockFromGenesis
 fromInitLog (InitFromSnapshot ss tip) = MockFromSnapshot ss tip
 fromInitLog (InitFailure ss err log') =
     case err of
-      InitFailureRead _err -> MockReadFailure     ss     (fromInitLog log')
---      InitFailureTooRecent tip -> MockTooRecent       ss tip (fromInitLog log')   -- TODO Javier's PR removed InitFailureTooRecent
-      InitFailureGenesis   -> MockGenesisSnapshot ss     (fromInitLog log')
+      InitFailureRead _err     -> MockReadFailure     ss     (fromInitLog log')
+      InitFailureTooRecent tip -> MockTooRecent       ss tip (fromInitLog log')
+      InitFailureGenesis       -> MockGenesisSnapshot ss     (fromInitLog log')
 
 mockInitLog :: Mock -> MockInitLog MockSnap
 mockInitLog Mock{..} = go (Map.toDescList mockSnaps)
@@ -478,13 +478,17 @@ data StandaloneDB m = DB {
 
       -- | LedgerDB config
     , dbLedgerDbCfg :: LedgerDbCfg (ExtLedgerState TestBlock)
+
+    , dbBackingStore :: StrictTVar m (LedgerBackingStore m (ExtLedgerState TestBlock))
     }
 
 initStandaloneDB :: forall m. IOLike m => DbEnv m -> m (StandaloneDB m)
 initStandaloneDB dbEnv@DbEnv{..} = do
     dbBlocks <- uncheckedNewTVarM Map.empty
     dbState  <- uncheckedNewTVarM (initChain, initDB)
-
+    dbBackingStore <- uncheckedNewTVarM =<< newBackingStore
+                                              (error "New backing store doesn't use HasFS for now")
+                                              (ExtLedgerStateTables NoTestLedgerTables)
     let dbResolve :: ResolveBlock m TestBlock
         dbResolve r = atomically $ getBlock r <$> readTVar dbBlocks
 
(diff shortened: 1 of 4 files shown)


ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -478,13 +478,18 @@ data StandaloneDB m = DB {
...
+    , dbBackingStore :: StrictTVar m (LedgerBackingStore m (ExtLedgerState TestBlock))
     }
 
 initStandaloneDB :: forall m. IOLike m => DbEnv m -> m (StandaloneDB m)
 initStandaloneDB dbEnv@DbEnv{..} = do
     dbBlocks <- uncheckedNewTVarM Map.empty
     dbState  <- uncheckedNewTVarM (initChain, initDB)
-
+    dbBackingStore <- uncheckedNewTVarM =<< newBackingStore
+                                              (error "New backing store doesn't use HasFS for now")
```
dnadales, line 491: We don't use it because it is an in-memory backing store?

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -608,16 +613,26 @@ runDB standalone@DB{..} cmd =
                 (map ApplyVal bs)
                 db
     go hasFS Snap = do
-        (_, db) <- atomically $ readTVar dbState
+        -- TODO: Maybe @Flush@ should be its own command but that would require
```
dnadales, line 616: Can we discuss this comment?

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -608,16 +613,26 @@ runDB standalone@DB{..} cmd =
...
+        -- the implementation to change so that 'ledgerDbOldest' doesn't require
+        -- a 'flush' to happen just before it is called.
+        (toFlush, bs, db') <- atomically $ do
```
dnadales, line 619: Maybe we should comment on why do we need to obtain the `toFlush`, backing store, and db state in the same atomic block.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -681,8 +698,10 @@ runDB standalone@DB{..} cmd =
                           return $ MaybeErr (Right ())
 
     truncateSnapshot :: SomeHasFS m -> DiskSnapshot -> m ()
-    truncateSnapshot (SomeHasFS hasFS@HasFS{..}) ss =
-        withFile hasFS (snapshotToDirPath ss) (AppendMode AllowExisting) $ \h ->
+    truncateSnapshot (SomeHasFS hasFS@HasFS{..}) ss = do
+        withFile hasFS (snapshotToStatePath ss) (AppendMode AllowExisting) $ \h ->
```
dnadales, line 702: Why do we have the nested `withFile` now?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -906,10 +906,10 @@ extendDbChangelog dblog newState =
     ext (ApplySeqDiffMK sq) (ApplyDiffMK diff) =
       ApplySeqDiffMK $ extendSeqUtxoDiff sq slot diff
 
-pruneDbChangelog ::
+pruneVolatileDbChangelog ::
```
dnadales, line 909: The name might be a bit confusing. Do we mean here the volatile DB changelog or the volatile part of the DB changelog?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -685,6 +685,9 @@ ledgerDbAnchor =
 -- | Information about the state of the most recently flushed ledger
 --
 -- This is what will be serialized when snapshotting.
+--
+-- PRECONDITION: if you are running the legacy ledger, then you must flush
+-- before calling this function
```
dnadales, line 690: Maybe we should explain why do we require this precondition.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
nfrisby: > the volatile DB changelog or the volatile part of the DB changelog

I _think_ I would interpret both of those as the same thing. What distinction did you have in mind?

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
nfrisby: They probably don't need to be, since this test is single-threaded. But conceptually the read-write should be atomic.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
nfrisby: It's not nested; it's two different files. (The new one is the file that the in-memory backing store writes when we take a snapshot.)

dnadales: Oh I see, sorry I misread this then.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
dnadales: Well, the changelog also contains an immutable part, doesn't it? Everything that comes before the state anchor.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
Jasagredo: Maybe abstract this so that it just receives a TVar and then we factor it out here and in LgrDB?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
Jasagredo: Add this (or some rewording of it) to the haddock above

dnadales (approved):

duog:
I'm seeing a test failure
```
          PostconditionFailed "AnnotateC \"real response didn't match model response\"" <omitted> /= Ok
          Use --quickcheck-replay=990702 to reproduce.
          Use -p '/ChainDB q-s-m.sequential/' to rerun this test only.
```

Jasagredo:
> I'm seeing a test failure on this branch:
> 
> ```
>           PostconditionFailed "AnnotateC \"real response didn't match model response\"" <omitted> /= Ok
>           Use --quickcheck-replay=990702 to reproduce.
>           Use -p '/ChainDB q-s-m.sequential/' to rerun this test only.
> ```
> 
> [test-output.txt](https://github.com/input-output-hk/ouroboros-network/files/8088040/test-output.txt)

@duog Indeed, this PR only fixes the `LedgerDB` part of the test-suite. The ChainDB part still requires some work

Jasagredo:
Merging after rebasing

PR #3631, opened 2022-02-23
Open the MapKind data type

nfrisby:
Fixes CAD-4008.

Draft PR, as of yet. This commit is the result of changing `Basics.hs` so that the `MapKind` is open but also so that most of the exported symbols can be used as the have been. As a result, `Basics.hs` is the only file with several changes. A follow-up commit could start doing broad but mechanical renamings/fixups throughout the rest of the codebase now.

There were a few non-mechanical changes:

  * Had to stop using `SingI` directly in most cases. It was replaced with the new (and quite similar) `IsApplyMapKind`.
  
  * Some quantified constraints no longer worked as-is. I think it's possible to enrich them -- something like quantified constraints whose context also involves quantified constraints -- but I didn't want to push the indirection that far (QCs already have terrible error messages as it is :frown). So instead, I simply monomorphized those constraints and listed out the ones that the `ouroboros-consensus` library code requires. I think I have a working set only because `cabal build all` succeeds.


nfrisby:
This commit can be reviewed. As the current PR description says, I intended to add some more that make more changes throughout, but they'll be mechanical.

This also might be a good time to introduce a standalone `PairMK` type, `ConstMK` type, etc, so that we can have a minimal set of fully expressive (ie binary that can combine to achieve n-ary) combinators in `TableStuff`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -310,7 +328,7 @@ class ( ShowLedgerState (LedgerTables l)
   -- the 'SingI' constraint. Unfortunately, that is not always the case. The
```
duog, line 328:
```suggestion
  -- the 'IsApplyMapKind' constraint. Unfortunately, that is not always the case. The
```

duog: Also I'm not convinced by this comment. `l mk` fails to determine `mk` only when `LedgerTables l` is trivial, in which case `projectLedgerTables` and `withLedgerTables` can be trivially implemented.

duog (commented):
This looks good, although I don't understand why `IsApplyMapKind` (or the singleton stuff it replaces) is necessary on `projectLedgerTables` and `withLedgerTables`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
nfrisby: I _think_ it boils down to the following, though I haven't dotted my is and crossed my ts.

Byron has no ledger tables. Thus a value of type `LedgerTables (LedgerState ByronBlock) mk` gives you know what to determine `mk`.

Cardano always has ledger tables (ie the Shelley UTxO map), even before the hard fork transition from Byron to Shelley.

Our current design involves compositional definitions for `HardForkBlock` which delegate to code on the composed eras (such as `ByronBlock`). Converting a Byron `LedgerTables` to a Cardano `LedgerTables` (eg after applying a block during the Byron era) involves creating an "empty" value of the Shelley UTxO map for whichever `mk` is relevant.

I don't think I've understood the detailed dynamics past that. I would be delighted to remove these constraints.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
```diff
@@ -88,11 +88,16 @@ newtype instance LedgerState (HardForkBlock xs) mk = HardForkLedgerState {
...
+deriving newtype instance CanHardFork xs => NoThunks (LedgerState (HardForkBlock xs) EmptyMK)
 
-instance (SingI mk, CanHardFork xs) => Show (LedgerState (HardForkBlock xs) mk) where
-  showsPrec p = showParen (p >= 11) . showsLedgerState sing
+deriving stock   instance CanHardFork xs => Eq       (LedgerState (HardForkBlock xs) ValuesMK)
+deriving newtype instance CanHardFork xs => NoThunks (LedgerState (HardForkBlock xs) ValuesMK)
+
+deriving newtype instance CanHardFork xs => NoThunks (LedgerState (HardForkBlock xs) SeqDiffMK)
+
+instance (SingI mk, CanHardFork xs) => Show (LedgerState (HardForkBlock xs) (ApplyMapKind' mk)) where
```
dnadales, line 99: Why didn't we need to replace `SingI` by `IsApplyMapKind` here? :thinking:

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -397,37 +418,39 @@ zipOverLedgerTables f l tables2 =
...
 -- involving double-ticked types.
 class TableStuff l => TickedTableStuff (l :: LedgerStateKind) where
-  -- | NOTE: The 'SingI' constraint is here for the same reason its on
-  -- 'projectLedgerTables'
-  projectLedgerTablesTicked :: SingI mk => Ticked1 l mk  -> LedgerTables l mk
-  withLedgerTablesTicked    ::             Ticked1 l any -> LedgerTables l mk -> Ticked1 l mk
+  -- | NOTE: The 'IsApplyMapKind mk2' constraint is here for the same reason
+  -- it's on 'projectLedgerTables'
+  projectLedgerTablesTicked :: IsApplyMapKind mk => Ticked1 l mk  -> LedgerTables l mk
+  -- | NOTE: The 'IsApplyMapKind mk2' constraint is here for the same reason
```
dnadales, line 424: Dangling comment?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -484,30 +507,42 @@ forgetTickedLedgerStateTracking = mapOverLedgerTablesTicked valuesTrackingMK
...
+              | EmptyMK'
+              | KeysMK'
+              | QueryMK'
+              | RewoundMK'
+              | SeqDiffMK'
+              | TrackingMK'
+              | ValuesMK'
+
+type DiffMK     = ApplyMapKind' DiffMK'
+type EmptyMK    = ApplyMapKind' EmptyMK'
```
dnadales, line 523: I'm a bit worried about having the ticked and unticked versions of `ApplyMapKind` and the constructors of `MapKind'`. For instance `EmptyMK` actually represents an application of the map kind to a promoted constructor that is called`EmptyMK'`.  But I don't have an alternative so I'm just complaining about the Government without going to vote :)

dnadales: Maybe: 

```
type AppEmptyMK    = ApplyMapKind' EmptyMK
```

:ballot_box:

dnadales: Or we might just want to remove the extra level of indirection.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -537,16 +572,16 @@ mapValuesAppliedMK f = \case
   ApplyQueryAllMK         -> ApplyQueryAllMK
   ApplyQuerySomeMK ks     -> ApplyQuerySomeMK (castUtxoKeys ks)
 
-diffKeysMK :: ApplyMapKind DiffMK k v -> ApplyMapKind KeysMK k v
+diffKeysMK :: DiffMK k v -> KeysMK k v
```
dnadales, line 575: Here we use the `DiffMK` type alias, whereas above we use `ApplyMapKind' KeysMK'` for instance. Would it make sense to use either one or the other?

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Basics.hs
nfrisby: I could have, but then it would have required `UndecidableInstances` which is one of the few LANGUAGE pragmas I'm hesitant to introduce without good cause. So what I'm done here is _almost_ equivalent but doesn't frighten GHC.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -209,8 +222,13 @@ type family LedgerCfg (l :: LedgerStateKind) :: Type
 
 class ( -- Requirements on the ledger state itself
         ShowLedgerState                     l
-      , forall mk. Eq                      (l mk)
-      , forall mk. Typeable mk => NoThunks (l mk)
+--      , forall mk. IsApplyMapKind mk                => Eq       (l mk)
```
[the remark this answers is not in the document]
nfrisby: No ultimately, no. Should be deleted before this PR moves beyond Draft status.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
nfrisby: No, but the `mk2` is a typo. Should be `mk`.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -484,30 +507,42 @@ forgetTickedLedgerStateTracking = mapOverLedgerTablesTicked valuesTrackingMK
...
+data MapKind' = DiffMK'
+              | EmptyMK'
+              | KeysMK'
+              | QueryMK'
+              | RewoundMK'
+              | SeqDiffMK'
+              | TrackingMK'
+              | ValuesMK'
+
+type DiffMK     = ApplyMapKind' DiffMK'
```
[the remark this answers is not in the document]
nfrisby: Richard Eisenberg has stated in a few places that he thinks the tick prefix was a mistake. I therefore avoid it (agreeing with him, and I think it's principally just immodular noise).

So I introduced the ..... hmm, actually I didn't. Well then, I intend to introduce `-Wno-unticked-promoted-constructors`, perhaps even in the `.cabal` file.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
nfrisby: I accidentally removed most `ApplyMapKind` within this module. Before this PR is merged, subsequent commits that only rename things or inline type synonyms etc (entirely mechanical) should address these things. We ultimately want to remove `ApplyMapKind`, leaving `DiffMK` etc as the only "public" interface.

nfrisby:
I want to emphasize that this PR should gain more commits before it is merged. And those commits should be rote, where the only decision being made is the final syntax of the `Basics.hs` exports and then the rest is just mechanically updating all the other modules accordingly.

One syntax-related commits is whether we want to split `ApplyMapKind'` into separate data types definitions; that's an option. For now, to minimize this first commit, I've left it intact.

nfrisby:
I rebased onto the tip of the target branch (whoopsies, hadn't realized I hadn't already) -- the conflicts were just in the import list, easy to resolve

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
duog: I see now. Permit me to restate to check my understanding:
 the `TableStuff (CardanoBlock c)` instance in `ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs` can't be defined without `polyEmptyLedgerTables`, which certainly requires `IsApplyMapKind mk` for access to `emptyAppliedMK`. 
 
 In my branch I have added an `ApplyMapKind` constructor `ApplyMkName :: String -> ApplyMapKind NameMK k v`. With an open map kind, instead we would define the ver useful `newtype ConstMk a k v = ConsMk a`. 
Clearly We can never `polyEmptyLedgerTables @ (ConstMap a)` for arbitrary a.

nfrisby:
We're going to merge this into the integration branch without the follow-up commits, so that we cal then merge PR #3630. Then we can do the follow-up commits in a follow-up PR.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
nfrisby: Yes, perhaps we could replace the `TableStuff CardanoBlock` instance's need for `IsAppliedMapKind` with `HasEmptyMK`. I'd have to spend more time looking into it to be sure.

nfrisby: Feel free to try adding such a commit to this branch, if that's reasonable priority for you.

duog: I gave it a little try yesterday. The singletons stuff is used by serilalisation/deserialisation and I didn't see an easy way to escape that.

nfrisby:
I merged and opened https://input-output.atlassian.net/browse/CAD-4020 for the follow-ups.

PR #3637, opened 2022-02-25
Add: PULL_REQUEST_TEMPLATE for GitHub

Jasagredo:
New features and breaking changes are sometimes not easily visible for outer people that might update/integrate a dependency on our repository.

This is a first idea for a pull request template that might help ease these situations. In particular Consensus plans on maintaining its `interface-CHANGELOG.md` file and this is a good way to make sure it is updated if needed.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,17 @@
+# Checklist
+
+- Branch
+    - [ ] Commit sequence broadly makes sense
+    - [ ] Commits have useful messages
+    - [ ] New tests are added if needed and existing tests are updated
+    - [ ] If this branch changes Consensus and has any consequences for downstream repositories or end users, said changes must be documented in [`interface-CHANGELOG.md`](../ouroboros-consensus/docs/interface-CHANGELOG.md)
+    - [ ] If this branch changes Network and has any consequences for downstream repositories or end users, said changes must be documented in [`interface-CHANGELOG.md`](../docs/interface-CHANGELOG.md)
+    - [ ] If serialization changes, user-facing consequences (e.g. replay from genesis) are confirmed to be intentional.
+- Pull Request
+    - [ ] Self-reviewed the diff
+    - [ ] Useful pull request description at least containing the following information:
+      - What does this PR change?
+      - Why these changes were needed?
+      - How does this affect downstream repositories and/or end-users?
+      - Which ticket does this PR close (if any)? If it does, is it [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)?
+    - [ ] Reviewer requested


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+# Checklist
+
+- Branch
+    - [ ] Commit sequence broadly makes sense
+    - [ ] Commits have useful messages
+    - [ ] At every intermediate commit, the libraries must compile. It is fine if tests don't compile/pass at these commits
```
coot, line 6: We haven't been following that.  Instead, we often refactored one library and in the following commit updated all the dependencies. This makes it much nicer for a review, though slightly less convenient when bisecting trying to find which change introduced a bug.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+    - [ ] New tests are added if needed and existing tests are updated
+    - [ ] If this branch changes Consensus and has any consequences for downstream repositories or end users, said changes must be documented in [`interface-CHANGELOG.md`](../ouroboros-consensus/docs/interface-CHANGELOG.md)
```
coot, line 8: Could you also drop a similar line about network related changes, they should all be recorded in `../docs/interface-CHANGELOG.md`.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+- Pull Request
+    - [ ] Self-reviewed the diff
+    - [ ] Useful pull request description at least containing the following information:
+      - What does this PR change?
+      - Why these changes were needed?
+      - How does this affect downstream repositories and/or end-users?
+      - Which ticket does this PR close (if any)?
```
coot, line 15:
```suggestion
      - [ ] Which ticket does this PR close (if any)? If it does is it [linked](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).
```

.github/PULL_REQUEST_TEMPLATE.md
Jasagredo: Removed :+1:

.github/PULL_REQUEST_TEMPLATE.md
Jasagredo: Added :+1:

.github/PULL_REQUEST_TEMPLATE.md
Jasagredo: Added :+1:

coot (approved):
LGTM

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+# Checklist
+
+- Branch
+    - [ ] Commit sequence broadly makes sense
+    - [ ] Commits have useful messages
+    - [ ] New tests are added if needed and existing tests are updated
+    - [ ] If this branch changes Consensus and has any consequences for downstream repositories or end users, said changes must be documented in [`interface-CHANGELOG.md`](../ouroboros-consensus/docs/interface-CHANGELOG.md)
+    - [ ] If this branch changes Network and has any consequences for downstream repositories or end users, said changes must be documented in [`interface-CHANGELOG.md`](../docs/interface-CHANGELOG.md)    
+- Pull Request
```
nfrisby, line 9: I think "Self-reviewed the diff" and "Reviewer requested" seem a bit excessive, but that's not to say they won't be helpful occasionally.

nfrisby (approved):
Thanks @Jasagredo looks good to me 👍

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+# Checklist
+
+- Branch
```
nfrisby, line 3: Ah, one thing to add: if serialization changes (eg there are diffs in the golden test files), we need to confirm any user-facing consequences (eg replay from genesis block) are intentional

Jasagredo: Added

.github/PULL_REQUEST_TEMPLATE.md
Jasagredo: Yeah, maybe, but it is not that bad to check two boxes and might be helpful sometimes

PR #3638, opened 2022-02-25
Don't overwrite localrootpeers lookup results

karknu:
Instead of updating the rootPeersGroupsVar every thread monitoring a
domain had their own cached copy of rootPeersGroups which it applied
changes to. This meant that only one thread's results where used,
causing the other localrootpeers to go away. However the missing
localrootpeers would still be known peers, which meant that they could
still be randomly promoted.


coot (approved):
LGTM

Issue #3524, opened 2021-12-03
Domain Names for LocalRoot peers


karknu:
Domain Names for LocalRoot peers are unreliable.

Workaround: Use a single flat list of IP addresses for now.

karknu:
Fixed in p2p-master-1.33.0 . jbgi is unable to reproduce on IOHK's testnet p2p nodes.

PR #3641, opened 2022-02-28
Don't overwrite localrootpeers lookup results

karknu:
Instead of updating the rootPeersGroupsVar every thread monitoring a
domain had their own cached copy of rootPeersGroups which it applied
changes to. This meant that only one thread's results where used,
causing the other localrootpeers to go away. However the missing
localrootpeers would still be known peers, which meant that they could
still be randomly promoted.


karknu:
Fixes  https://github.com/input-output-hk/ouroboros-network/issues/3524

bolt12 (approved):
LGTM

coot (approved):

PR #3643, opened 2022-02-28
RootPeersDNS: garbage collect DNS results && test single source of truth

### LINKED ISSUE #3650: DNS entries are not garbage collected
The local root peers provider never expire old entries from the dns map,
[ref](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs?plain=1#L235).

Note that internally we need to distinguish static addresses and addresses obtained from dns lookups.

bolt12:
This PR extends the existing test suite for RootPeersDNS by making the MockRoots more robust, and adding a test to check if localRootPeersProvider has a single source of truth. It also fixes #3650.

This was motivated by a bug found by @karknu and fixed in #3641 .

### DIFF
--- a/ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
+++ b/ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
@@ -25,9 +25,11 @@ import           Data.Foldable (foldl', toList)
 import           Data.Functor (void)
 import           Data.IP (fromHostAddress, toIPv4w, toSockAddr)
 import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.List.Trace as Trace
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import           Data.Maybe (catMaybes)
+import           Data.Proxy (Proxy(..))
 import           Data.Sequence (Seq)
 import qualified Data.Sequence as Seq
 import           Data.Set (Set)
@@ -42,7 +44,7 @@ import           Control.Exception (throw)
 import           Control.Monad.Class.MonadAsync
 import qualified Control.Monad.Class.MonadSTM as LazySTM
 import           Control.Monad.Class.MonadSTM.Strict (MonadSTM, newTVarIO,
-                     readTVar)
+                     readTVar, traceTVarIO, MonadTraceSTM, TraceValue(..))
 import           Control.Monad.Class.MonadThrow
 import           Control.Monad.Class.MonadTime (Time (..))
 import           Control.Monad.Class.MonadTimer
@@ -71,6 +73,8 @@ tests =
                       prop_local_resolvesDomainsCorrectly
        , testProperty "updates domains correctly"
                       prop_local_updatesDomainsCorrectly
+       , testProperty "localRootPeersProvider has single source of truth"
+                      prop_local_singleSourceOfTruth
        ]
     , testGroup "publicRootPeersProvider"
        [ testProperty "resolves domains correctly"
@@ -295,12 +299,17 @@ mockDNSActions dnsMapScript dnsTimeoutScript dnsLookupDelayScript =
        Just (Left e)  -> return ([e], [])
        Just (Right a) -> return ([], a)
 
+-- TODO: Change this once upgraded to most recent GHC version
+newtype Solo a = Solo { unSolo :: a }
+  deriving (Show, Eq)
+
 -- | 'localRootPeersProvider' running with a given MockRoots env
 --
 mockLocalRootPeersProvider :: forall m.
-                              ( MonadAsync m
-                              , MonadDelay m
-                              , MonadTimer m
+                              ( MonadAsync    m
+                              , MonadDelay    m
+                              , MonadTimer    m
+                              , MonadTraceSTM m
                               , Eq (Async m Void)
                               )
                            => Tracer m (TraceLocalRootPeers SockAddr Failure)
@@ -315,6 +324,9 @@ mockLocalRootPeersProvider tracer (MockRoots localRootPeers dnsMapScript _ _)
       dnsLookupDelayScriptVar <- initScript' dnsLookupDelayScript
       localRootPeersVar <- newTVarIO localRootPeers
       resultVar <- newTVarIO mempty
+      _ <- traceTVarIO proxy
+                       resultVar
+                       (\_ a -> pure $ TraceDynamic (Solo a))
 
       void $ MonadTimer.timeout 3600 $
         localRootPeersProvider tracer
@@ -325,6 +337,9 @@ mockLocalRootPeersProvider tracer (MockRoots localRootPeers dnsMapScript _ _)
                                                dnsLookupDelayScriptVar)
                                (readTVar localRootPeersVar)
                                resultVar
+  where
+    proxy :: Proxy m
+    proxy = Proxy
 
 -- | 'publicRootPeersProvider' running with a given MockRoots env.
 --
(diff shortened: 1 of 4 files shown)


ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -290,12 +291,17 @@ mockDNSActions dnsMap dnsTimeoutScript dnsLookupDelayScript =
        Just (Left e)  -> return ([e], [])
        Just (Right a) -> return ([], a)
 
+-- TODO: Change this once upgraded to most recent base version 4.16
+data Solo a = Solo a
+  deriving (Show, Eq)
```
coot, line 296: Why not import it?

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -676,7 +688,7 @@ prop_local_singleSourceOfTruth mockRoots
       $ foldl'
           (\(r, prev) t ->
             case t of
-              TraceLocalRootGroups curr ->
+              Solo curr ->
```
coot, line 691: This should be `Solo (TraceLocalRootGroups curr)`, isn't it?

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
coot: [Solo](https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Tuple.html#t:Solo)

bolt12: Only available in base 4.16, would have to upgrade bounds I think, or it is due to how nix is configured.

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: No, I am tracking it like `Solo curr`.

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
coot: Not with base; it's version is bound to [ghc version](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history), we'll need to wait for `9.2`.

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -288,12 +291,17 @@ mockDNSActions dnsMap dnsTimeoutScript dnsLookupDelayScript =
        Just (Left e)  -> return ([e], [])
        Just (Right a) -> return ([], a)
 
+-- TODO: Change this once upgraded to most recent base version 4.16
+data Solo a = Solo a
```
coot, line 295:
```suggestion
newtype Solo a = Solo a
```

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -307,6 +315,9 @@ mockLocalRootPeersProvider tracer (MockRoots localRootPeers dnsMap _ _)
       dnsLookupDelayScriptVar <- initScript' dnsLookupDelayScript
       localRootPeersVar <- newTVarIO localRootPeers
       resultVar <- newTVarIO mempty
+      _ <- traceTVarIO proxy
+                       (toLazyTVar resultVar)
+                       (\_ a -> pure $ Solo a)
```
coot, line 320: :+1:

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
coot: Yes, it makes more sense to do so.

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -642,6 +656,52 @@ prop_local_updatesDomainsCorrectly mockRoots@(MockRoots lrp _ _ _)
...
+           $ mockLocalRootPeersProvider tracerTraceLocalRoots
+                                        mockRoots
+                                        dnsTimeoutScript
+                                        dnsLookupDelayScript
+
+     in fst
+      $ foldl'
+          (\(r, prev) t ->
+            case t of
+              Solo curr ->
```
coot, line 690:
```suggestion
          (\(r, prev) (Solo curr) ->
```

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -642,6 +656,52 @@ prop_local_updatesDomainsCorrectly mockRoots@(MockRoots lrp _ _ _)
...
+--
+prop_local_singleSourceOfTruth :: MockRoots
+                               -> Script DNSTimeout
+                               -> Script DNSLookupDelay
+                               -> Property
+prop_local_singleSourceOfTruth mockRoots
+                               dnsTimeoutScript
+                               dnsLookupDelayScript =
+    let tr = selectTraceEventsDynamic
+              @() @(Solo (Seq (Int, Map SockAddr PeerAdvertise)))
```
coot, line 679: btw, one can use `@_` to ignore type parameters.

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
coot: I would also rename `prev` to `prevSet`, or even better rename `prev` to `localRootsSet` and `currentSet` to `localRootsSet'`.

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -642,6 +656,52 @@ prop_local_updatesDomainsCorrectly mockRoots@(MockRoots lrp _ _ _)
...
+     in fst
+      $ foldl'
+          (\(r, prev) t ->
+            case t of
+              Solo curr ->
+                let currSet = Set.fromList
+                            . concatMap (Map.keys . snd)
+                            . toList
+                            $ curr
+                    isMonotonic = prev `Set.isSubsetOf` currSet
```
coot, line 695: This is not the right property, it only works because of #3650.

bolt12: Right, but if you are saying that we can not go for a monotonicity check because older entries might be removed then it probably is impossible to distinguish a single source of truth as well, no?

How can I just by observing the contents of the TVar tell if its a SST ?

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
coot: ^ ping

bolt12: done!

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: done!

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: done!

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: Oh right! :D

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: hmmm it sounds tricky, but doable! I'll think about this!

bolt12: I have been thinking and experimenting about this and the best way to do this is once we have garbage collection we update this test to accommodate it. We should have a trace for when and which DNS entries were gc'd and then we would update our expected `localRootPeerSet` to be monotonic modulo garbage collection.

coot (approved):
LGTM

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -459,30 +484,36 @@ prop_local_preservesIPs mockRoots@(MockRoots localRoots _ _ _)
...
+          localGroupEventsAddresses (_, s) =
+              Set.fromList
+            $ concatMap (Map.keys . snd)
+            $ toList s
+
+          localRootAddressesSet = localRootAddresses localRoots
+          localGroupEventsAddressesSet = localGroupEventsAddresses x
+       in counterexample (show localRootAddressesSet ++ " is not subset of "
+                         ++ show localGroupEventsAddressesSet)
+        $ localRootAddressesSet `Set.isSubsetOf` localGroupEventsAddressesSet
```
coot, line 515: :+1:

bolt12:
bors merge

coot:
bors cancel

bolt12:
bors merge

PR #3645, opened 2022-02-28
Fixes so prototype cardano-node can sync Byron from mainnet using in-mem backing store

nfrisby:
Fixes CAD-3904.

These commits include small patches that ultimately enabled a `cardano-node` prototype (branch `nfrisby/CAD-3904-utxohd-inmem-only` in that repo) to sync Byron from `mainnet`. (It failed on the first non-empty Shelley block; CAD-4018 tracks that.)

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
@@ -734,6 +734,13 @@ instance
       decodeArityAndTag len tag = do
         len' <- CBOR.decodeListLen
         tag' <- CBOR.decodeWord8
+        -- @len@ here ought to match the @length xs@ in @encodeArityAndTag@ in
+        -- the corresponding 'ToCBOR' instance, so we need to add one in order
+        -- to match the length recorded in the CBOR stream (ie @len'@)
+        --
+        -- This use of 'when' corresponds to the use of 'decodeListLenOf'
+        -- throughout most of our CBOR decoders: it catches encoder/decoder
+        -- mismatches.
         when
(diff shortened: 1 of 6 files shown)


nfrisby:
Everything builds on this commit, but, as expected, some tests still fail; these changes were only driven by getting the `cardano-node` branch to successfully sync (only Byron, as it turns out) via `./cardano-node.exe run --topology ./configuration/cardano/mainnet-topology.json --database-path ./state --port 3001 --config ./configuration/cardano/mainnet-config.yaml  --socket-path \\.\pipe\cardano-node`

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -654,8 +655,8 @@ instance
         len' <- CBOR.decodeListLen
         tag' <- CBOR.decodeWord8
         when
-          (len /= 1 + len' || tag /= tag')
-          (fail $ "decode @ApplyMapKind " <> show (smk, len', tag'))
+          (1 + len /= len' || tag /= tag')
```
dnadales, line 658: Could we comment on why we check this assertion here?

Jasagredo (approved):
Nick and I went through this on a call. Nick will add a comment on the CBOR check and then merge

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
nfrisby: Done!

nfrisby:
bors merge

nfrisby:
bors cancel

PR #3648, opened 2022-03-02
Extract tcp_info for mux tcp bearer

karknu:
Extract and trace TCP rtt, rtt var and congestion window for tcp
sockets.
This is a costly debug trace that only works on linux and requires one
to compile with the flag "tracetcpinfo".

### DIFF
--- a/network-mux/src/Network/Mux/TCPInfo.hsc
+++ b/network-mux/src/Network/Mux/TCPInfo.hsc
@@ -1,0 +1,122 @@
+-- This is a non-portable linux only interface.
+
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE GADTs               #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE MonadComprehensions #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE PatternSynonyms     #-}
+{-# LANGUAGE Rank2Types          #-}
+{-# LANGUAGE RecordWildCards     #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving  #-}
+{-# LANGUAGE TypeApplications    #-}
+
+#include "HsNet.h"
+
+module Network.Mux.TCPInfo
+  ( StructTCPInfo (..)
+  , SocketOption(TCPInfoSocketOption)
+  ) where
+
+import           Foreign.C
+import           Foreign.Storable (Storable (..))
+import           Network.Socket(SocketOption(..))
+
+
+pattern TCPInfoSocketOption :: SocketOption
+pattern TCPInfoSocketOption = SockOpt 6 11
+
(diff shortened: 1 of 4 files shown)


network-mux/src/Network/Mux/TCPInfo.hsc
```diff
@@ -0,0 +1,136 @@
+{-# LANGUAGE CPP                 #-}
+
+#ifdef linux_HOST_OS
```
coot, line 3: Instead of a top level #ifdef we can conditionally include this module in `other-modules`.

karknu: That file defines a `StructTCPInfo` regardless of OS which is used when defining the `MuxTraceTCPInfo`.

karknu: I've changed to a conditional inclusion of TCPInfo in the cabal file and updated the definition of MuxTraceTCPInfo

network-mux/src/Network/Mux/Trace.hs
```diff
@@ -184,4 +194,17 @@ instance Show MuxTrace where
     show (MuxTraceStartedOnDemand mid dir) = printf "Started on demand (%s) in %s" (show mid) (show dir)
     show (MuxTraceTerminating mid dir) = printf "Terminating (%s) in %s" (show mid) (show dir)
     show MuxTraceShutdown = "Mux shutdown"
+#ifdef linux_HOST_OS
+    show (MuxTraceTCPInfo StructTCPInfo { tcpi_snd_mss, tcpi_rcv_mss, tcpi_lost, tcpi_retrans
+                                  , tcpi_rtt, tcpi_rttvar, tcpi_snd_cwnd } len)
```
coot, line 199:
```suggestion
    show (MuxTraceTCPInfo StructTCPInfo
            { tcpi_snd_mss, tcpi_rcv_mss, tcpi_lost, tcpi_retrans
            , tcpi_rtt, tcpi_rttvar, tcpi_snd_cwnd }
            len)
```

network-mux/src/Network/Mux/Trace.hs
```diff
@@ -147,6 +152,11 @@ data MuxTrace =
     | MuxTraceStartedOnDemand MiniProtocolNum MiniProtocolDir
     | MuxTraceTerminating MiniProtocolNum MiniProtocolDir
     | MuxTraceShutdown
+#ifdef linux_HOST_OS
+    | MuxTraceTCPInfo StructTCPInfo Word16
+#else
+    | MuxTraceTCPInfo Word16
+#endif
```
coot, line 159: I haven't though about this, but with your original solution one was able to unify this constructor to
```
    | MuxTraceTCPInfo StructTCPInfo Word16
```

You can leave it as is, and I'll fix it next week (by providing two modules `Network.Mux.TCPInfo.Linux` and a public `Network.Mux.TCPInfo`).

coot (approved):
LGTM

network-mux/src/Network/Mux/Trace.hs
karknu: Sounds good.

PR #3652, opened 2022-03-03
Docs: Consensus onboarding docs update

Jasagredo:
Add clarification for multi-user nix installations.
Also update the Consensus team members.

### DIFF
--- a/ouroboros-consensus/docs/Onboarding.md
+++ b/ouroboros-consensus/docs/Onboarding.md
@@ -35,15 +35,18 @@ nix-env (Nix) 2.3.10
 > you will have to "compile the whole world" (including few GHC versions) before
 > you will be able to build, test and run our project.
 > If configured correctly, your fresh build (that means build from completely new
-> machine) should take ~15 minutes
+> machine) should take ~15 minutes, otherwise, it can take several hours.
 
-Open your `~/.config/nix/nix.conf` and add/modify following entries
+If you are using a single-user installation of Nix, open your
+`~/.config/nix/nix.conf` and add/modify following entries
 
 ```
 substituters = https://cache.nixos.org/ https://hydra.iohk.io/
 trusted-substituters = https://cache.nixos.org/ https://hydra.iohk.io/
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/docs/Onboarding.md
```diff
@@ -37,14 +37,17 @@ nix-env (Nix) 2.3.10
...
+If you are using a single-user installation of Nix, open your
+`~/.config/nix/nix.conf` and add/modify following entries
 
 ```
 substituters = https://cache.nixos.org/ https://hydra.iohk.io/
 trusted-substituters = https://cache.nixos.org/ https://hydra.iohk.io/
 trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
 ```
 
+If instead you are using a multi-user installation, add/modify said entries in `/etc/nix/nix.conf`.
```
dnadales, line 49: Would it make sense to refer to the cache misses that one might run into if this is not correctly set?

dnadales (approved):

ouroboros-consensus/docs/Onboarding.md
Jasagredo: I think so. Will add it later today :+1:

amesgen (approved):
As said, I think we could remove `trusted-substitutors` here, but it shouldn't do any harm in any case.

```
       • trusted-substituters

         A  list  of  URLs of substituters, separated by whitespace. These are
         not used by default, but can be enabled by users of the Nix daemon by
         specifying  --option  substituters urls on the command line. Unprivi‐
         leged users are only allowed to pass a subset of the URLs  listed  in
         substituters and trusted-substituters.
```

ouroboros-consensus/docs/Onboarding.md
Jasagredo: Done :ok_hand:

PR #3654, opened 2022-03-03
io-classes: added strict versions of traceTVar & traceTMVar

coot:


bolt12 (approved):
LGTM, but it seems CI is failing

PR #3661, opened 2022-03-10
Update downstream dependencies.

nc6:
This bumps:
- base
- ledger
- plutus

In addition, the index-state and pinned nix dependencies are synced with
those in base and ledger (and node).

A spurious upper bound on the `semialign` dependecy is dropped, since
this prevented building with aeson > 2.0.0. which is needed for other
components.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,4 +1,4 @@
-index-state: 2022-01-17T00:00:00Z
+index-state: 2022-02-18T00:00:00Z
 
 packages: ./typed-protocols
           ./typed-protocols-cborg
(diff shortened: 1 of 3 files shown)


dnadales (approved):

cabal.project
```diff
@@ -160,17 +160,11 @@ allow-newer:
   monoidal-containers:aeson,
```
amesgen, line 160: I think this could also be removed, see https://hackage.haskell.org/package/monoidal-containers-0.6.2.0/dependencies

amesgen (approved):

coot:
Github CI is failing:
```
 Resolving dependencies...
cabal.exe: Could not resolve dependencies:
[__0] next goal: cardano-crypto-class (user goal)
[__0] rejecting: cardano-crypto-class-2.0.0 (conflict: pkg-config package
libsecp256k1-any, not found in the pkg-config database)
[__0] fail (backjumping, conflict set: cardano-crypto-class)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: cardano-crypto-class
```

PR #3662, opened 2022-03-10
Removed ST effects from IOSimPOR

coot:
ST effects are thread local, they never race with other steps.  This patch
reverts a previous fix commited in PR #3647.

Thanks for `IOSimPOR` we also found out that the transition:
`UnnegotiatedSt Inbound → TerminatingSt` should be valid, similarly to one for
outbound connections (see PR #3652, commit fbcb6aada).


rjmh:
This looks good to me.

bolt12 (approved):

coot:
bors merge

PR #3670, opened 2022-03-17
CAD-4018 UTxO-HD Fix shelley bug

Jasagredo:
Ticking a ledger state now goes from `l EmptyMK` to `l ValuesMK`. Those differences are prepended to the ones produces by applying the block and therefore recorded in the `DbChangelog` as they should.

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
@@ -81,6 +81,7 @@ import           Ouroboros.Consensus.HardFork.History (Bound (boundSlot),
                      addSlots)
 import           Ouroboros.Consensus.HardFork.Simple
 import           Ouroboros.Consensus.Ledger.Abstract
+import qualified Ouroboros.Consensus.Storage.LedgerDB.HD as HD
 import           Ouroboros.Consensus.TypeFamilyWrappers
 import           Ouroboros.Consensus.Util (eitherToMaybe)
 import           Ouroboros.Consensus.Util.RedundantConstraints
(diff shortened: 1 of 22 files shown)


ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -705,31 +706,32 @@ translateLedgerStateByronToShelleyWrapper :: forall c.
...
+                , shelleyLedgerState =
+                      SL.translateToShelleyLedgerState
+                        (shelleyLedgerGenesis cfgShelley)
+                        epochNo
+                        (byronLedgerState ledgerByron)
+                , shelleyLedgerTransition =
+                      ShelleyTransitionInfo{shelleyAfterVoting = 0}
+                , shelleyLedgerTables = translateNoTables (projectLedgerTables ledgerByron)
+                }
+          , translateLedgerTablesWith = translateNoTables
```
Jasagredo, line 727: Maybe inline `polyEmptyLedgerTables`?

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -909,8 +939,22 @@ translateLedgerStateMaryToAlonzoWrapper ::
...
+              . unFlip
+              . unComp
+              . SL.translateEra' (getAlonzoTranslationContext cfgAlonzo)
+              . Comp
+              . Flip
+        , translateLedgerTablesWith =
+            \ShelleyLedgerTables { shelleyUTxOTable = ApplyValuesMK (HD.UtxoValues vals) } ->
+             ShelleyLedgerTables { shelleyUTxOTable = ApplyValuesMK
+                                                    . HD.UtxoValues
+                                                    . fmap Alonzo.translateTxOut
```
Jasagredo, line 954: Perhaps `translateTxOut` should be performed through `SL.translateEra' ()` as it seems an instance could be defined for that one on the ledger side. This would continue with the same pattern as eras above

dnadales (approved):
Reviewed during a call.

PR #3676, opened 2022-03-30
IOSim's MVar

coot:


coot:
`MVar`s were added in input-output-hk/io-sim#10.

PR #3680, opened 2022-03-30
Performance improvements of IOSim

### LINKED ISSUE #3673: Performance review of IOSim

coot:
Resolves #3673.

### DIFF
--- a/io-sim/src/Control/Monad/IOSim/Internal.hs
+++ b/io-sim/src/Control/Monad/IOSim/Internal.hs
@@ -207,7 +207,7 @@ schedule !thread@Thread{
 
       ForkFrame -> do
         -- this thread is done
-        trace <- deschedule Terminated thread simstate
+        !trace <- deschedule Terminated thread simstate
         return $ SimTrace time tid tlbl EventThreadFinished
                $ SimTrace time tid tlbl (EventDeschedule Terminated)
                $ trace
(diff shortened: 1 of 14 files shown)


io-sim/src/Control/Monad/IOSim/Internal.hs
```diff
@@ -198,7 +198,8 @@ schedule !thread@Thread{
   assert (invariant (Just thread) simstate) $
   case action of
 
-    Return x -> case ctl of
+    Return x -> {-# SCC "schedule.Return" #-}
```
bolt12, line 201: Is this necessary to stay? (The same for all changes in this commit 1d5d3763e94b5a16b9e2177b5d40351d5157079b)

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
```diff
@@ -1057,7 +1057,9 @@ execAtomically time tid tlbl nextVid0 action0 k0 =
        -> ST s (SimTrace c)
     go ctl !read !written writtenSeq createdSeq !nextVid action = assert localInvariant $
                                                        case action of
-      ReturnStm x -> case ctl of
+      ReturnStm x ->
+        {-# SCC "execAtomically.go.ReturnStm" #-}
```
bolt12, line 1061: Is this necessary to stay? (The same for the rest of this commit 22a4ed1d56c926bf89038f12a699e213f33823a8)

bolt12 (approved):
LGTM! Just some small things

io-sim/src/Control/Monad/IOSim/Internal.hs
coot: It doesn't cost us to keep it, and might prove to be useful in the future.

io-sim/src/Control/Monad/IOSimPOR/Internal.hs
coot: As above.

coot:
bors merge

PR #3681, opened 2022-03-31
Fix the ChainDB test suite

dnadales:


ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -398,6 +398,33 @@ run env@ChainDBEnv { varDB, .. } cmd =
...
+-- no longer the case since the ledger DB can be altered as the result of taking
+-- snapshots or opening the ledger DB (for instance when we process the
+-- 'WipeVolatileDB' command). Taking snapshots or opening the ledger DB cause
+-- the ledger DB to be flushed, which modifies its sequence of volatile and
+-- immutable states.
+--
+-- The model does not have information about when the flushes occur and it
+-- cannot infer that information in a reliable way since this depends on the low
+-- level details of operations such as opening the ledger DB. Therefore, we
+-- assume that the 'GetLedgerDB' command should return a flushed ledger DB, and
```
nfrisby, line 414: My only feedback is that this wording (partly my fault for naming the upstream functions) may lead someone to think that a "flushed ledger DB" implies that the backing store was updated. In this case, it's not: "flushed ledger DB" means the ledger db that _would_ result from actually flushing its changes.

Naming is hard. Dunno if there's an easy fix. Not 100% crucial.

nfrisby (approved):
Approved. That comment is _*excellent*_! Thank you!

Jasagredo (approved):

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
dnadales: That's a good point. I can add a TODO in the feature branch to rename `ledgerDbFlush` so that we avoid this confusion. I have to think about what a suitable name would be.

PR #3683, opened 2022-04-01
Updated cardano-base & avoid libsecp256k1 on Windows

### LINKED ISSUE #3678: Install libsecp256k1 on Windows in GH actions.
CI is failing due to missing `libsecp256k1` dependency introduced in #3661.

coot:
Closes #3678, `libsecp256k1` is not available on Windows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -130,12 +130,27 @@ jobs:
       run: |
         cat ./cabal.project.local.ci >> ./cabal.project.local
 
+    - name: Configure (Linux)
+      if: matrix.os == "ubuntu-20.04"
         cat >> cabal.project.local <<EOF
         package cardano-crypto-praos
           flags: -external-libsodium-vrf
         EOF
 
         cat ./cabal.project.local
+
+    - name: Configure (Windows)
+      if: matrix.os == 'windows-latest'
+      run: |
+        echo "FIXME: can't install from sources, so instead libsecp256k1 support is disabled"
+        cat >> cabal.project.local <<EOF
+        package cardano-crypto-praos
+          flags: -external-libsodium-vrf
+        package cardano-crypto-class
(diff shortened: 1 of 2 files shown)


amesgen:
FYI: In https://github.com/input-output-hk/haskell.nix/pull/1424, @hamishmack added the possibility to download a Windows binary of libsecp256k1; this might be an alternative to this PR.

hamishmack:
I wonder if these need to append to PKC_CONFIG_PATH somehow?

https://github.com/input-output-hk/ouroboros-network/blob/de22b3505ab84a98701f45411fb4fa21db0e2aec/.github/workflows/build.yml#L73

https://github.com/input-output-hk/ouroboros-network/blob/de22b3505ab84a98701f45411fb4fa21db0e2aec/.github/workflows/build.yml#L123

hamishmack:
Can you `echo "$PKG_CONFIG_PATH"` in the failing step to see what value it has?

amesgen:
> Can you `echo "$PKG_CONFIG_PATH"` in the failing step to see what value it has?

See https://github.com/input-output-hk/ouroboros-network/runs/5816059242?check_suite_focus=true

hamishmack:
I wonder if it should be: 

```
export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:$(readlink -f secp256k1/lib/pkgconfig)"
```

To keep libsodium in the path.  But if that is the case why was it secp256k1 that it could not find?

amesgen:
Just pushed 4e887f581f7dcc457985f5303d9080a57a5275bf to append to PKG_CONFIG_PATH, but I agree that one would expect that libsodium and not libsecp256k1 could not be found before.

Also, I opened #3687 which I guess is a better place to discuss this.

coot:
Closing this one in favour of #3687.

PR #3688, opened 2022-04-04
Block diffusion pipelining

nfrisby:
This PR adds block diffusion pipelining, primarily by finally introducing the long-existing concept of _the tentative chain_ into the code.

Main changes:

- add the tentative chain header to the ChainDB state.

- have the `Follower`s for the Node-To-Node ChainSync server follow the tentative chain instead of only the selected chain, thus sending the tentative header before the underlying block has been validated

- this now means some honest nodes will send us an invalid block, when the block is tentative, so we adjust ChainSync and BlockFetch clients to allow that in limited number of scenarios necessary for the common pipelining scenarios

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
@@ -56,6 +56,7 @@ module Ouroboros.Consensus.Storage.ChainDB.Impl.Types (
   , TraceInitChainSelEvent (..)
   , TraceIteratorEvent (..)
   , TraceOpenEvent (..)
+  , TracePipeliningEvent (..)
   , TraceValidationEvent (..)
   ) where
 
@@ -644,6 +645,11 @@ data TraceAddBlockEvent blk =
     -- This is done for all blocks from the future each time a new block is
     -- added.
   | ChainSelectionForFutureBlock (RealPoint blk)
+
+    -- | The tentative header (in the context of diffusion pipelining) has been
+    -- updated.
+  | PipeliningEvent (TracePipeliningEvent blk)
+
   deriving (Generic)
 
 deriving instance
@@ -695,6 +701,18 @@ deriving instance
   ( Show (Header           blk)
   , LedgerSupportsProtocol blk
   ) => Show (TraceValidationEvent blk)
+
+data TracePipeliningEvent blk =
+    -- | A new tentative header got set.
+    SetTentativeHeader (Header blk)
+    -- | The body of tentative block turned out to be invalid.
+  | TrapTentativeHeader (Header blk)
+    -- | We selected a new (better) chain, which cleared the previous tentative
+    -- header.
+  | OutdatedTentativeHeader (Header blk)
+
+deriving stock instance Eq   (Header blk) => Eq   (TracePipeliningEvent blk)
+deriving stock instance Show (Header blk) => Show (TracePipeliningEvent blk)
 
 data TraceInitChainSelEvent blk =
     StartedInitChainSelection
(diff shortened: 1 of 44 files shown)


amesgen:
> Esgen thinks the CI failure on [4cc618f](https://github.com/input-output-hk/ouroboros-network/commit/4cc618f426b80b9a351edd6522c7551d88240407) is spurious. He's going to write an explanation here soon.

See #3689 (which reproduces the issue on current master).

amesgen (approved):
~~The GH Windows failure is unrelated (see #3678),~~ (EDIT: now fixed) so LGTM from my side 🎉

nfrisby:
Status update: the first benchmarking run showed that we're not actually pipelining. We believe we've fixed it and are now awaiting a second benchmarking run. We'll be pushing those commits here soon.

nfrisby:
An (optional) idea: split 4fa8cda198a1c875fb6 into a commit that just undoes/reverts changes and a second commit that adds the new behavior.

nfrisby:
Minor: 4fa8cda198a1c875fb6 has some extra layer of `do` in the `clearTentativeHeader` definition. I'm guessing the tracer commit puts that to use, but it looks a little odd isolated in the previous commit

amesgen:
> An (optional) idea: split [4fa8cda](https://github.com/input-output-hk/ouroboros-network/commit/4fa8cda198a1c875fb6884e429192703004f6a91) into a commit that just undoes/reverts changes and a second commit that adds the new behavior.

I can do that if it is helpful for reviewing :+1:. In the end, we want to squash [4fa8cda](https://github.com/input-output-hk/ouroboros-network/pull/3688/commits/4fa8cda198a1c875fb6884e429192703004f6a91) into [aa43e5d](https://github.com/input-output-hk/ouroboros-network/pull/3688/commits/aa43e5dc21fafd0464206a7423a65b0ccd0b6f78), so it won't matter there.

> Minor: [4fa8cda](https://github.com/input-output-hk/ouroboros-network/commit/4fa8cda198a1c875fb6884e429192703004f6a91) has some extra layer of `do` in the `clearTentativeHeader` definition. I'm guessing the tracer commit puts that to use, but it looks a little odd isolated in the previous commit

Ah yes, I swapped the two commits (such that the squashing becomes trivial), will change that.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -195,6 +199,13 @@ data ChainDbEnv m blk = CDB
     --
     -- Note that the \"immutable\" block will /never/ be /more/ than @k@
     -- blocks back, as opposed to the anchor point of 'cdbChain'.
+  , cdbTentativeState  :: !(StrictTVar m (TentativeState blk))
+  , cdbTentativeHeader :: !(StrictTVar m (Maybe (Header blk)))
```
nfrisby, line 203: @amesgen We should use a `MaybeHeader` type here that is strict.

Also, the Haddock is stale: we do read from it as of your promptness fix

nfrisby (commented):
I rebased onto `origin/master` and did one last pass. I found a tvar with a `Maybe` and a stale comment.

@amesgen would you fix that (preferably squash the fix) and then merge?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
nfrisby: Esgen noted that I neglected to pull before rebasing. That explains the stale comment. But the `TVar Maybe` remains.

I have re-rebased (it was entirely automatic this time).

nfrisby:
bors merge

nfrisby:
Thank @amesgen for preparing this one with me! The benchmarking results show improvement, so we're good to merge. We still have some room for improvement in the benchmarks and new tests, but this PR is a solid first step.

PR #3692, opened 2022-04-04
Fix has fs testsuite

dnadales:
The problem was that we were seeking a file that was removed via `removeDirectoryRecursively`. The mock does not allow us to remove files that are opened, so I changed `removeDirectoryRecursively` to make sure we don't delete opened files.

### DIFF
--- a/ouroboros-consensus-test/src/Test/Util/FS/Sim/FsTree.hs
+++ b/ouroboros-consensus-test/src/Test/Util/FS/Sim/FsTree.hs
@@ -30,6 +30,8 @@ module Test.Util.FS.Sim.FsTree (
   , removeFile
   , renameFile
   , replace
+    -- * Path-listing
+  , find
     -- * Pretty-printing
   , pretty
   ) where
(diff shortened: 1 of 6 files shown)


ouroboros-consensus-test/src/Test/Util/FS/Sim/FsTree.hs
```diff
@@ -281,6 +299,65 @@ renameFile fpOld fpNew tree = do
...
+        findFsTree :: FsTree a -> [[Text]]
+        findFsTree (File   _      ) = []
+        findFsTree (Folder folder') = findTree folder'
+
+findExample :: [FsPath]
+findExample =
+      either (error . show) id
+    $ find (fsPathFromList ["var"]) (example :: FsTree ())
+
+examplePathsNotFound :: [FsPath]
```
dnadales, line 335: These would probably have to go to tests.

ouroboros-consensus-test/src/Test/Util/FS/Sim/FsTree.hs
```diff
@@ -69,7 +76,18 @@ example =
               ])
           ])
       , ("var", Folder $ M.fromList [
-            ("log",  Folder mempty)
+            ("log",  Folder $ M.fromList [
```
dnadales, line 79: This should go to the `FsTree` unit tests (if we have those).

ouroboros-consensus-test/src/Test/Util/FS/Sim/MockFS.hs
```diff
@@ -754,29 +754,32 @@ doesFileExist fp = readMockFS $ \fs ->
...
+               , fsErrorNo     = Nothing
+               , fsErrorStack  = prettyCallStack
+               , fsLimitation  = True
+               }
+  --      _ | fp `S.member` openFilePaths fs
+        _ | reachablePaths `S.intersection` openFilePaths fs /= mempty
+          -> throwError FsError {
+                 fsErrorType   = FsIllegalOperation
+               , fsErrorPath   = fsToFsErrorPathUnmounted fp
+               , fsErrorString = "cannot remove an open file"
```
dnadales, line 775: Would it make sense to add the opened files to the error? As a string?

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/FsTree.hs
```diff
@@ -0,0 +1,128 @@
...
+      , testCase "boom" $ ["boom"] `shouldReportMissingFileIn` example
+      ]
+    ]
+
+checkResultsOfFind :: [Text] -> FsTree () -> [FsPath] -> Assertion
+checkResultsOfFind fp fs expectedResult = do
+    (expectedResult \\ filePathsFound)
+      `shouldBeEmptyOtherwise` "Not all expected paths were found"
+    (filePathsFound \\ expectedResult)
+      `shouldBeEmptyOtherwise` "Find returned unexpeced paths"
```
dnadales, line 39:
```suggestion
      `shouldBeEmptyOtherwise` "Find returned unexpected paths"
```

Jasagredo (approved):

ouroboros-consensus-test/src/Test/Util/FS/Sim/MockFS.hs
```diff
@@ -754,29 +754,36 @@ doesFileExist fp = readMockFS $ \fs ->
...
-      _ -> do
-        files' <- checkFsTree $ FS.removeDirRecursive fp (mockFiles fs)
-        return ((), fs { mockFiles = files' })
+removeDirectoryRecursive fp = do
+    modifyMockFS $ \fs -> do
+      reachablePaths <- fmap S.fromList $ checkFsTree $ FS.find fp (mockFiles fs)
+      let openReachablePaths = reachablePaths `S.intersection` openFilePaths fs
+      case fsPathToList fp of
+        []
+          -> throwError FsError {
```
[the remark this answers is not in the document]
dnadales: As I hypothesized, the precondition filters now the case in which we delete an open handle recursively:

```haskell

precondition m@Model{..} (At cmd) =
            QSM.forall (handles cmd) (`QSM.member` RE.keys knownHandles)
    QSM.:&& QSM.Boolean (Mock.numOpenHandles mockFS < maxNumOpenHandles)
    QSM.:&& QSM.Not (knownLimitation m (At cmd))
    
knownLimitation model cmd =
    case getResp resp of
      Left FsError{..} -> QSM.Boolean fsLimitation

removeDirectoryRecursive fp =
    modifyMockFS $ \fs -> case fsPathToList fp of
      []
...
      _ | fp `S.member` openFilePaths fs
        -> throwError FsError {
               fsErrorType   = FsIllegalOperation
             , fsErrorPath   = fsToFsErrorPathUnmounted fp
             , fsErrorString = "cannot remove an open file"
             , fsErrorNo     = Nothing
             , fsErrorStack  = prettyCallStack
             , fsLimitation  = True
             }    
``` 
Note that we set `fsLimitation` to True when throwing the error (this was already the case before this PR and this is consistent with the limitation that we can't remove a file if there's an open handle associated with it.

PR #3701, opened 2022-04-06
Keep and forward keys when applying a block

Jasagredo:
Keys were being dropped and therefore applying a block would fail. This
keeps them in `UnforwardedReadSets`.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
@@ -30,6 +30,7 @@ module Ouroboros.Consensus.Storage.LedgerDB.HD (
     -- * Combinators
   , RewoundKeys (..)
   , forwardValues
+  , forwardValuesAndKeys
   , mapRewoundKeys
   , restrictValues
   , rewindKeys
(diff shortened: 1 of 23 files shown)


Jasagredo:
Note that the diff looks longer because I had to define `zipLedgerTables2` on every `TableStuff` instance.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -339,6 +340,30 @@ forwardValues (UtxoValues values) (UtxoDiff diffs) =
       UedsIns       -> error "impossible! duplicate insert of key"
       UedsInsAndDel -> error "impossible! duplicate insert of key"
 
+-- | Transport a set of values by applying a valid difference as well as
+-- creating the values for keys that must be created in said differences.
+--
+-- Note that this fails via 'error' if the diff is invalid, eg it deletes a key
+-- that is not present in the argument or inserts a key that is already in the
+-- argument.
+forwardValuesAndKeys :: forall k v. (Ord k, HasCallStack) => UtxoValues k v -> UtxoKeys k v -> UtxoDiff k v -> UtxoValues k v
```
dnadales, line 349: Would it make sense to make a note of the coupling that exists between `RewoundKeySets`, `UnforwardedReadSets`, and this function?

dnadales (approved):

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
Jasagredo: I think it should be enough with what I wrote now

PR #3702, opened 2022-04-08
CAD-4155 UTxO-HD: Ticking and apply now produce diffs

Jasagredo:
In the light of the Shelley to Allegra transition requiring to delete
values, it was decided that both operations on the ledger would return
differences, and we would inject the AVVMs in the shelley -> allegra
transition.


dnadales (approved):
Reviewed on a call with Javier.

PR #3705, opened 2022-04-12
UTxO-HD CAD-4157 introduce dead AVVMs

Jasagredo:
# Current state

As for now translating a LedgerState only accepts an `l Empty` state, we are injecting with `stowLedgerTables` the `deadCardanoAVVM`s harcoded into the source before calling the translation function, so that they are destroyed and added to the reserves.

After the translation, we contruct a `DiffMK` out of thin air from the `deadCardanoAVVM`s harcoded into the source.

```diff
          translateLedgerStateWith = \_epochNo ->
-                noNewTickingValues
+                flip withLedgerTables (ShelleyLedgerTables
+                                      $ ApplyDiffMK
+                                       $ HD.UtxoDiff
+                                       $ Map.map (`HD.UtxoEntryDiff` HD.UedsDel)
+                                       $ Map.map (unTxOutWrapper . SL.translateEra' () . TxOutWrapper)
+                                       $ SL.unUTxO deadCardanoAVVM
+                                      )
              . unFlip
              . unComp
              . SL.translateEra' ()
              . Comp
              . Flip
+              . stowLedgerTables
+              . flip withLedgerTables (ShelleyLedgerTables $ ApplyValuesMK $ HD.UtxoValues $ SL.unUTxO deadCardanoAVVM)
```

# Intermediate state

After removing the `DeadCardanoAVVMs.hs` module, the ledger must keep track of those AVVMs starting from the Byron->Shelley transition, where they were already made unspendable, all the way up to the Shelley->Allegra transition, where they were destroyed (and moved into the reserves).

For that, Ledger will have to store said values in a field somewhere inside the `NewEpochState`, and it should on its own, feed those values into the Shelley->Allegra transition. From the Consensus side, we should get a function that provides us with those values from inside the `NewEpochState`, to be used somehow along this example:

```diff
 translateLedgerStateShelleyToAllegraWrapper =
     ignoringBoth $
       TranslateLedgerState {
-          translateLedgerStateWith = \_epochNo ->
+          translateLedgerStateWith = \_epochNo ls ->
                 flip withLedgerTables (ShelleyLedgerTables
                                        $ ApplyDiffMK
                                        $ HD.UtxoDiff
                                        $ Map.map (`HD.UtxoEntryDiff` HD.UedsDel)
-                                       $ Map.map (unTxOutWrapper . SL.translateEra' () . TxOutWrapper)
-                                       $ SL.unUTxO deadCardanoAVVM
+                                       $ SL.getRedeemersToDelete $ shelleyLedgerState ls
                                       )
               . unFlip
               . unComp
               . SL.translateEra' ()
               . Comp
               . Flip
-              . stowLedgerTables
-              . flip withLedgerTables (ShelleyLedgerTables $ ApplyValuesMK $ HD.UtxoValues $ SL.unUTxO deadCardanoAVVM)
+              $ ls
         , translateLedgerTablesWith =
```

The `stowLedgerTables` at the bottom would be no longer needed as the Ledger would already inject the needed values on its own, and the new function `getRedeemersToDelete` would return a `SplitMap.SplitMap (TxIn (Crypto AllegraEra)) (Core.TxOut AllegraEra)` which can be directly turned into a `DiffMK`.

# Desired final state

The ledger functions for apply block, ticking and translating an era should have the following shape:

```
... :: ... -> LedgerState l ValuesMK -> LedgerState l DiffMK
```

So in particular, the `SL.translateEra` instance 
```
instance ... => SL.TranslateEra era (Flip LedgerState (ApplyMapKind' mk) :.: ShelleyBlock) where
```
should already return a `DiffMK` containing the deletions of the AVVM addresses. The resulting snippet would look something like this:

```diff
 translateLedgerStateShelleyToAllegraWrapper =
     ignoringBoth $
       TranslateLedgerState {
           translateLedgerStateWith = \_epochNo ->
-                flip withLedgerTables (ShelleyLedgerTables
-                                       $ ApplyDiffMK
-                                       $ HD.UtxoDiff
-                                       $ Map.map (`HD.UtxoEntryDiff` HD.UedsDel)
-                                       $ Map.map (unTxOutWrapper . SL.translateEra' () . TxOutWrapper)
-                                       $ SL.unUTxO deadCardanoAVVM
-                                      )
               . unFlip
               . unComp
               . SL.translateEra' ()
               . Comp
               . Flip
-              . stowLedgerTables
-              . flip withLedgerTables (ShelleyLedgerTables $ ApplyValuesMK $ HD.UtxoValues $ SL.unUTxO deadCardanoAVVM)
         , translateLedgerTablesWith =
```


dnadales (approved):
Reviewed during a call with Javier.

PR #3708, opened 2022-04-13
Fix performance regression

dnadales:


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -667,9 +664,9 @@ ledgerDbCurrentValues =
...
 ledgerDbCurrent :: GetTip (l EmptyMK) => LedgerDB l -> l EmptyMK
 ledgerDbCurrent =
-    either unCheckpoint unCheckpoint
+    either unDbChangelogState unDbChangelogState
   . AS.head
-  . dbChangelogVolatileCheckpoints
+  . changelogVolatileStates
   . ledgerDbChangelog
 
 -- | Information about the state of the ledger at the anchor
```
Jasagredo, line 672: I cannot comment on un-edited lines, but I would also change `ledgerDbAnchor` to use `unDbChangelogState` and avoid `dbChangelogVolatileCheckpoints`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -101,6 +101,7 @@ import           Ouroboros.Network.AnchoredSeq (Anchorable (..),
 import qualified Ouroboros.Network.AnchoredSeq as AS
 
 import           Control.Exception
+import           Data.Coerce (coerce)
 import           Ouroboros.Consensus.Block
```
Jasagredo, line 105:
```suggestion
import           Control.Exception
import           Data.Coerce (coerce)

import           Ouroboros.Consensus.Block
```

Perhaps keep the separation. See point 15 in https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/StyleGuide.md#formatting

Jasagredo (commented):
I would remove also the need in `ledgerDbBimap` for the conversion. I'm pretty sure we can just wipe away `dbChangelogVolatileCheckpoints` and just use the anchored sequence of DbChangelogStates

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -746,7 +743,7 @@ ledgerDbMaxRollback :: (GetTip (l ValuesMK), GetTip (l EmptyMK)) => LedgerDB l -
 ledgerDbMaxRollback LedgerDB{..} =
   let
     old = fromIntegral . AS.length <$> ledgerDbCheckpoints
-    new = fromIntegral $ AS.length $ dbChangelogVolatileCheckpoints ledgerDbChangelog
+    new = fromIntegral $ AS.length $ changelogVolatileStates ledgerDbChangelog
```
dnadales, line 746: Discussed with Nick: revert to `dbChangelogVolatileCheckpoints` so that we don't have to reach for the internals.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -667,9 +664,9 @@ ledgerDbCurrentValues =
 -- | The ledger state at the tip of the chain
 ledgerDbCurrent :: GetTip (l EmptyMK) => LedgerDB l -> l EmptyMK
 ledgerDbCurrent =
-    either unCheckpoint unCheckpoint
+    either unDbChangelogState unDbChangelogState
```
dnadales, line 667: Discussed with Nick: revert to dbChangelogVolatileCheckpoints so that we don't have to reach for the internals.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
dnadales: Nick pointed out that it's better to use `dbChangelogVolatileCheckpoints` so that we use LedgerDB API functions without reaching for the internals.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/InMemory.hs
```diff
@@ -598,11 +599,7 @@ dbChangelogVolatileCheckpoints ::
        (WithOrigin SlotNo)
        (Checkpoint (l EmptyMK))
        (Checkpoint (l EmptyMK))
-dbChangelogVolatileCheckpoints =
-    AS.bimapPreservingMeasure
-      (\(DbChangelogState l) -> Checkpoint l)
-      (\(DbChangelogState l) -> Checkpoint l)
-  . changelogVolatileStates
+dbChangelogVolatileCheckpoints = coerce changelogVolatileStates
```
dnadales, line 602: Discussed with Nick: inline and remove `dbChangelogVolatileCheckpoints`.

Jasagredo (approved):
Reviewed on a call

PR #3710, opened 2022-04-13
CAD-4180 Rebase LMDB branch

Jasagredo:


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+{-------------------------------------------------------------------------------
+ Db settings
+-------------------------------------------------------------------------------}
+
+readDbSettings ::
+     LMDB.Database () DbState
+  -> LMDB.Transaction mode DbState
+readDbSettings db = readDbSettingsMaybeNull db >>= \case
+  Just x -> pure x
+  -- FIXME: if this (and other errors) are thrown, the async thread will die but
```
Jasagredo, line 269: This is an important FIXME, probably worth fixing before merging. The 'happy' path however works fine.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+
+guardDbDir :: MonadIO m => GuardDbDir -> FS.SomeHasFS m -> FS.FsPath -> m FilePath
+guardDbDir gdd (FS.SomeHasFS fs) path = do
+  FS.doesFileExist fs path >>= \b -> when b $ Exn.throw $ DbErrStr $ "guardDbDir:must be a directory:" <> show path
+  FS.doesDirectoryExist fs path >>= \case
+    True | GDDMustNotExist <- gdd ->
+           -- TODO Should throw a DbErrDbExists exception
+           -- Callers should delete the directory if they want to restore a snapshot
+           FS.removeDirectoryRecursive fs path
+    False | GDDMustExist <- gdd -> Exn.throw $ DbErrDbDoesntExist path
```
Jasagredo, line 303: This `throw`s exceptions, but probably the function should instead be `-> ExceptT DBErr m FilePath` to be able to act on the errors instead of needing to be in IO to catch them.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+    Just _ -> Exn.throw $ DbErrStr "initFromVals: db already had state"
+
+initFromLMDBs :: MonadIO m
+  => Trace.Tracer IO TraceDb
+  -> FS.SomeHasFS m
+  -> FS.FsPath
+  -> FS.FsPath
+  -> m ()
+initFromLMDBs tracer shfs from0 to0 = do
+  liftIO $ Trace.traceWith tracer $ TDBInitialisingFromLMDB from0
```
Jasagredo, line 337: I _think_ that we should be able to provide a tracer that was transformed into `m` with `natTracer`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+  liftIO $ Exn.bracket
+    (LMDB.openEnvironment from defaultLMDBLimits) -- TODO assess limits, in particular this could fail if the db is too big
+    LMDB.closeEnvironment $ \e -> lmdbCopy tracer e to
+
+
+lmdbCopy :: MonadIO m
+  => Trace.Tracer m TraceDb
+  -> LMDB.Environment LMDB.ReadWrite
+  -> FilePath
+  -> m ()
```
Jasagredo, line 349: Again I think there is a mixture of `m` and `IO` here. Above we use `liftIO` on this, so `m ~ IO`. Is it used somewhere else that requires `m /~ IO`?

Jasagredo, line 349: Also I'm unsure if when we use this, we are actually storing the new environment somewhere or is it being lost? I imagined `lmdbCopy` should return `m (LMDB.Environment LMDB.ReadWrite)` so that we keep track of the new copied DB, and in case we are not interested, we close that environment afterwards. Or maybe not?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+      Trace.traceWith tracer TVHSubmissionStarted
+      ret_tmvar <- IOLike.newEmptyTMVarIO
+      IOLike.atomically . IOLike.writeTBQueue q . Just $ SomeDbSubmission t ret_tmvar
+      r <- IOLike.atomically $ IOLike.takeTMVar ret_tmvar
+      Trace.traceWith tracer TVHSubmissionEnded
+      pure r
+
+    vhClose = do
+      Trace.traceWith tracer TVHClosing
+      IOLike.cancel the_async
```
Jasagredo, line 509: The async thread is being cancelled instead of closed (by sending a Nothing). Why is this the case?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+    vhSubmit t = do
+      Trace.traceWith tracer TVHSubmissionStarted
+      ret_tmvar <- IOLike.newEmptyTMVarIO
+      IOLike.atomically . IOLike.writeTBQueue q . Just $ SomeDbSubmission t ret_tmvar
+      r <- IOLike.atomically $ IOLike.takeTMVar ret_tmvar
+      Trace.traceWith tracer TVHSubmissionEnded
+      pure r
+
+    vhClose = do
+      Trace.traceWith tracer TVHClosing
```
Jasagredo, line 508: I have a feeling that we are doing a bit of an overkill with so much tracing in this module.

Jasagredo (commented):
I think the rest looks fine, modulo the questions we already were discussing
- Reason for using TBQueues
- Time spent in `atomically` and `async`
- Time spent in IO

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+  , LMDBValueHandle
+  , DbErr
+  , LMDBInit(..)
+  , LMDBLimits
+  , TraceDb(..)
+  , defaultLMDBLimits -- TODO this is just for convenience, should remove
+  ) where
+
+import qualified Codec.Serialise as S (Serialise(..))
+import qualified Control.Exception as Exn
```
dnadales, line 27: We should be consistent with other imports of `Control.Exception`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+  | TTOWrite
+  deriving (Show)
+
+{-------------------------------------------------------------------------------
+ Errors
+-------------------------------------------------------------------------------}
+
+data DbErr = DbErrStr !String
+  | DbErrNoSettings
+  | DbErrNonMonotonicSeq !(WithOrigin SlotNo) !(WithOrigin SlotNo)
```
dnadales, line 107: We should probably explain what this error means.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+
+instance Exn.Exception DbErr
+
+{-------------------------------------------------------------------------------
+  Database definition
+-------------------------------------------------------------------------------}
+
+data Db l = Db
+  { dbEnv           :: !(LMDB.Environment LMDB.ReadWrite) -- ^ The LMDB environment is a pointer to the directory that contains the DB.
+  , dbSettings      :: !(LMDB.Database () DbState) -- ^ A database with only one key and one value, for the current sequence number on the DB.
```
dnadales, line 125: We should explain what the current sequence number is.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+
+lmdbReadTable :: LedgerConstraint k v
+  => LMDB.Database k v
+  -> HD.UtxoKeys k v
+  -> LMDB.Transaction mode (HD.UtxoValues k v)
+lmdbReadTable db0 (HD.UtxoKeys keys) =
+    HD.UtxoValues <$> foldlM' go Map.empty (Set.toList keys)
+  where
+    db = coerceDatabase db0
+    go m k = LMDB.get db (LmdbBox k) <&> \case
```
dnadales, line 176: If the following makes sense we should probably add a TODO and JIRA ticket: I wonder if we should be issuing a single C call to get a bunch of keys in one go.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+      Just (LmdbBox v) -> Map.insert k v m
+
+lmdbWriteTable :: LedgerConstraint k v
+  => LMDB.Database k v
+  -> HD.UtxoDiff k v
+  -> LMDB.Transaction LMDB.ReadWrite ()
+lmdbWriteTable db0 (HD.UtxoDiff m) =
+    void $ Map.traverseWithKey go m
+  where
+    db = coerceDatabase db0
```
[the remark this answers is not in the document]
dnadales: Good point. We should vanish unsafe coercions if possible.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+    go k (HD.UtxoEntryDiff v reason) =
+      LMDB.put db (LmdbBox k) (if reason `elem` [HD.UedsDel, HD.UedsInsAndDel] then Nothing else Just (LmdbBox v))
```
dnadales, line 189: We should match exactly on `UedsIns` as well as in the other cases. Otherwise we might be inserting under the wrong circumstances if we add an additional constructor to `HD.UtxoEntryDiffState`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+
+lmdbInitRangeReadTable :: LedgerConstraint k v
+  => Int
```
dnadales, line 192: We need to explain what this int represents.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+  -> LedgerTables l ValuesMK
+  -> Db l
+  -> IO ()
+initFromVals dbsSeq vals Db{..} = LMDB.readWriteTransaction dbEnv $
+  withDbSettingsRWMaybeNull dbSettings $ \case
+    Nothing -> zipLedgerTablesA initLMDBTable dbBackingTables vals $> ((), DbState{dbsSeq})
+    Just _ -> Exn.throw $ DbErrStr "initFromVals: db already had state"
+
+initFromLMDBs :: MonadIO m
+  => Trace.Tracer IO TraceDb
```
dnadales, line 331: We need to comment these params.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
dnadales: Might be. Returning the environment will require using the resource registry right? Since in this case we won't be able to use `bracket`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+    LMDB.closeEnvironment $ \e -> lmdbCopy tracer e to
+
+
+lmdbCopy :: MonadIO m
+  => Trace.Tracer m TraceDb
+  -> LMDB.Environment LMDB.ReadWrite
+  -> FilePath
+  -> m ()
+lmdbCopy tracer dbEnv to = do
+  -- TODO This copying is gross. Tests depend on it, but I wish I was only responsible for copying the database here
```
dnadales, line 351: If only the tests depend on it, can we move this to the tests? :thinking: I wonder how much of this module we will have to expose to achieve this.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+-- TODO 50% of total time is spent somewhere else. Where?
+
+-- | Initialise a backing store
+newLMDBBackingStore :: forall m l. (TableStuff l, MonadIO m, IOLike m)
+  => Trace.Tracer m TraceDb
+  -> LMDBLimits
+  -> FS.SomeHasFS m
+  -> LMDBInit l
+  -> m (LMDBBackingStore l m)
+newLMDBBackingStore tracer limits sfs init_db = do
```
dnadales, line 375: We need to consider documenting some parameters. In particular `sfs` could have a more descriptive name.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,545 @@
...
+  -- Since the tracer we were passed runs in m, we use this async to match the
+  -- Tracer IO we want to use with the Tracer m we have
+  -- TODO exception safety, we should trace_async 'onException'
+  --
+
+  -- TODO why are we using TBQueues?? is it really needed?
+
+  -- TODO This tracing logic also could use a review (perhaps use `natTracer` to transform the tracer?)
+  trace_q <- liftIO $ IOLike.newTBQueueIO 1
+  trace_async <- IOLike.async . void . runMaybeT . forever $ do
```
dnadales, line 390: In addition to the comment above I wonder why do we need to trace asynchronously.

dnadales: After reviewing the uses of `dbTracer` in this module I think we should be  able to get rid of this additional complexity.

dnadales (commented):
- We need to add LMDB installation instructions for cabal users.
- It seems we're testing OnDisk with the in-memory and LMDB backend :+1: 
- We need to format the code properly (eg avoid snake case)
- Maybe the backing store needs to add a tracer.

ouroboros-consensus-test/src/Test/Tasty/Traceable.hs
```diff
@@ -0,0 +1,36 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+-- |
+{-# OPTIONS_GHC -Wno-missing-export-lists #-}
+
+module Test.Tasty.Traceable where
```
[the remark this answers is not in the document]
dnadales: We should also use explicit exports.

ouroboros-consensus-test/src/Test/Tasty/Traceable.hs
```diff
@@ -0,0 +1,36 @@
...
+import Test.Tasty.QuickCheck (QC(QC))
+
+newtype ShowTrace = ShowTrace Bool
+  deriving stock (Eq, Show)
+
+instance Tasty.IsOption ShowTrace where
+  defaultValue = coerce False
+  parseValue = coerce safeReadBool
+  optionName = coerce "trace"
+  optionHelp = coerce "Use a Control.Tracer.stdoutTracer where possible"
```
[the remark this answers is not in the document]
dnadales:
```suggestion
  optionHelp = coerce "This option allows the user to specify if the property should show a trace. It is to the property to choose how to act on this flag."
```

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -88,14 +91,19 @@ import           Test.Util.TestBlock hiding (TestBlock, TestBlockCodecConfig,
...
 
+import qualified Test.Tasty.Traceable as TTT
 {-------------------------------------------------------------------------------
   Top-level tests
 -------------------------------------------------------------------------------}
 
 tests :: TestTree
-tests = testGroup "OnDisk" [
-      testProperty "LedgerSimple" prop_sequential
+tests = testGroup "OnDisk"
```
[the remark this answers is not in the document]
dnadales: Hmmm on a second thought we don't want to share the same LMDB database across each QC test run.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -814,16 +828,21 @@ data StandaloneDB m = DB {
     , dbLedgerDbCfg :: LedgerDbCfg (ExtLedgerState TestBlock)
 
     , dbBackingStore :: StrictTVar m (LedgerBackingStore m (ExtLedgerState TestBlock))
+
+      -- | needed for restore TODO does this really belong here?
```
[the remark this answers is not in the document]
dnadales: Upon a brief inspection of the code it seems this field is not out of place here :heavy_check_mark:

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -1313,23 +1339,57 @@ sm secParam db = StateMachine {
...
+        HD.bsClose bs
+        dbCleanup $ dbEnv db
     }
 
-prop_sequential :: SecurityParam -> QC.Property
-prop_sequential secParam = QC.withMaxSuccess 100000 $
-    forAllCommands (sm secParam dbUnused) Nothing $ \cmds ->
-      QC.monadicIO (propCmds secParam cmds)
-
+prop_sequential :: (SecurityParam -> IO (DbEnv IO)) -> SecurityParam -> QC.Property
```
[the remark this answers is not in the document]
jorisdral: I believe we use `prop_sequential` to test for arbitrary values of `SecurityParam` at lines 101-107, so we can't omit it here. We also can't remove it from `DbEnv IO` because `initStandaloneDB` depends on it. We could test for arbitrary values of `DbEnv IO` that contain arbitrary values of `SecurityParam`, but I don't think we can come up with a sensible `Arbitrary` instance for it. Maybe we can still make it a bit nicer with minor refactoring?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/FS/API.hs
```diff
@@ -147,6 +147,9 @@ data HasFS m h = HasFS {
 
     -- | Useful for better error reporting
   , mkFsErrorPath            :: FsPath -> FsErrorPath
+
+    -- | TODO
+  , unsafeToFilePath         :: !(FsPath -> m FilePath)
```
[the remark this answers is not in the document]
dnadales: We need at least comment on why do we need this.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -0,0 +1,499 @@
...
+  -> LMDB.Environment LMDB.ReadOnly
+  -> IOLike.TVar m (Map Int (ValueHandle m))
+  -> m (ValueHandle m)
+mkValueHandle dbTracer0 dbEnv dbOpenHandles = do
+  let LMDB.Env env = dbEnv
+  readOnlyTx <- liftIO $ LMDB.mdb_txn_begin env Nothing (LMDB.isReadOnlyEnvironment dbEnv)
+
+  (r, traces) <- IOLike.atomically $ IOLike.stateTVar dbOpenHandles $ \x0 ->
+    let
+      vh_id = maybe 0 ((+1) . fst) $ Map.lookupMax x0
```
jorisdral, line 445: Matter of taste: Extract this conversion to value handle ID to a top-level function?

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -975,17 +995,23 @@ runDB standalone@DB{..} cmd =
...
     go hasFS Restore = do
+        LedgerBackingStore old_db <- atomically . readTVar $ dbBackingStore
+        HD.bsClose old_db
         (initLog, db, _replayed, backingStore) <-
           initLedgerDB
             nullTracer
-            nullTracer
+            (Trace.Tracer $ \case
+               LMDBEvent e -> Trace.runTracer (dbTracer dbEnv) e
+               _ -> pure ()) -- TODO gross
```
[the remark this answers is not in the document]
jorisdral: We have (at least) two options: trace `LMDB.TraceDb` events only, or `TraceEvent blk` as well. Seeing as this implementation seems to be intended to only trace `LMDB.TraceDb`, I'd propose to do only a minor refactoring to use an abastraction like `Trace.traceMaybe` or `Trace.squelchUnless`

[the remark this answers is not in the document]
jorisdral: I see now that my last comment was a bit naive: `traceMaybe` is not sufficient because you still have to embed `Tracer m TraceDb` in `Tracer m TraceEvent`. I did find that the [`traceMaybe` function from the original `contra-variant` package](https://github.com/avieth/contra-tracer/blob/c8fc0aec6a57e3011aab907bdf48bbdb9a3224b2/src/Control/Tracer.hs#L217) would do exactly this embedding for us. Would it be worth it to upstream this function to the IOHK-fork of `contra-variant`?

jorisdral (commented):
I've added some more remarks after having worked on CAD-4303 (minor LMDB refactorings)

ouroboros-consensus-test/src/Test/Tasty/Traceable.hs
```diff
@@ -0,0 +1,36 @@
...
+import Data.Tagged
+import Data.Coerce
+import qualified Test.QuickCheck as QC
+import Test.Tasty
+import Test.Tasty.QuickCheck (QC(QC))
+
+newtype ShowTrace = ShowTrace Bool
+  deriving stock (Eq, Show)
+
+instance Tasty.IsOption ShowTrace where
```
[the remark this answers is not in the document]
jorisdral: I've added a Todo that mentions the possibility of improving this option such that it will also allow showing traces only when the test fails.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -121,3 +121,29 @@ executable db-analyser
                        -threaded
                        -rtsopts
                        "-with-rtsopts=-T -I0 -N2 -A16m"
+
+executable ledger-db-backends-checker
```
dnadales, line 125: Did we document this in the readme or somewhere?

ouroboros-consensus-test/ouroboros-consensus-test.cabal
```diff
@@ -279,6 +283,7 @@ test-suite test-storage
                        -Wmissing-export-lists
                        -Wno-unticked-promoted-constructors
                        -fno-ignore-asserts
+                       -threaded
```
dnadales, line 286: Do you know what did we add this?

dnadales (approved):

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
jorisdral: Not AFAIK, we should add it to `ouroboros-consensus-cardano/README.md`

ouroboros-consensus-test/ouroboros-consensus-test.cabal
jorisdral: The LMDB bindings won't work if we don't set the `-threaded` option

```bash
$ cabal run ouroboros-consensus-test:test-storage -- -p "LMDB" 

TDBOpening
ouroboros-storage
  Storage
    LedgerDB
      HD
        LMDB
          simple insert and read: FAIL
            Exception: user error (RTS doesn't support multiple OS threads (use ghc -threaded when linking))
            Use -p '/LMDB/&&/simple insert and read/' to rerun this test only.
      OnDisk
        LedgerSimple-LMDB:        FAIL
          *** Failed! Exception: 'user error (RTS doesn't support multiple OS threads (use ghc -threaded when linking))' (after 1 test and 1 shrink):
          SecurityParam 0
          Commands { unCommands = [] }
          Use --quickcheck-replay=160168 to reproduce.
          Use -p '/LMDB/&&/LedgerSimple-LMDB/' to rerun this test only.

2 out of 2 tests failed (0.00s)
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/Common.hs
```diff
@@ -7,6 +7,7 @@
 {-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE NamedFieldPuns             #-}
 {-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE DataKinds                  #-}
```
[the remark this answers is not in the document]
jorisdral: Not needed

jorisdral:
Closed because of #3786

PR #3711, opened 2022-04-13
CAD-4179 Refine Shelley->Allegra to use the Ledger AVVMs

Jasagredo:
Implement the Shelley -> Allegra transition using the function exposed by the ledger instead of the hard-coded version.


ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -848,43 +848,54 @@ translateLedgerStateShelleyToAllegraWrapper ::
...
+              -- provides them to us through 'shelleyToAllegraAVVMsToDelete'.
               --
-              -- Once the ledger keeps track of these values, this operation
-              -- will no longer be needed.
-              . stowLedgerTables
-              . flip withLedgerTables (ShelleyLedgerTables $ ApplyValuesMK $ HD.UtxoValues $ SL.unUTxO deadCardanoAVVM)
+              -- In the long run, the ledger will already use ledger states
+              -- parametrized by the map kind and therefore will already provide
+              -- the differences in this translation.
+              let avvms           = SL.unUTxO $ fromSMaybe
```
nfrisby, line 861: FYI, one of my requests on Clarke's PR was to use a type family which would avoid the need for this `error` case here.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -848,43 +848,54 @@ translateLedgerStateShelleyToAllegraWrapper ::
...
+                              . ApplyValuesMK
+                              . HD.UtxoValues
+                              $ avvms
+
+                  resultingState = unFlip . unComp
+                                 . SL.translateEra' ()
+                                 . Comp   . Flip
+                                 $ stowedState
+
+              in resultingState `withLedgerTables` avvmsAsDeletions
```
nfrisby, line 897: There may be a simple answer to this that I could find with some care, but I'm just going to ask instead: Should this instead be combining `avvmsAsDeletions` with whatever was in `projectLedgerTables resultingState`?

nfrisby (commented):
I did a first pass. I expect upstream changes soon, so I'll hold off on Approving/Requesting Changes etc.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
Jasagredo: Technically yes, although for now it is only UTxO and we know the `projectLedgerTables resultingState` will be empty. I kinda noted that in the comment above, specifying that this is a special case.

dnadales (approved):
Reviewed during a call.

PR #3712, opened 2022-04-13
Fix Haddocks CI by installing secp256k1

amesgen:
Started failing since #3661

### DIFF
--- a/.github/workflows/github-page.yml
+++ b/.github/workflows/github-page.yml
@@ -29,39 +29,10 @@ jobs:
 
     - name: Select build directory
       run: |
-        if [ "$OS" == Windows_NT ]; then
-          CABAL_BUILDDIR="D:\\a\\_temp\\dist"
-        else
-          CABAL_BUILDDIR="dist-newstyle"
-        fi
+        CABAL_BUILDDIR="dist-newstyle"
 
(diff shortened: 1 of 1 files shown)


Jasagredo:
I re-added the `nix` label after seeing that it is described to contain everything related to "ci, and also nix"

Jasagredo (approved):
This looks correct to me as it seems to work on the `build` workflows.

Perhaps we could also use this opportunity to remove the steps related to windows from this workflow as the matrix.os only specifies linux?

coot (approved):

amesgen:
> Perhaps we could also use this opportunity to remove the steps related to windows from this workflow as the matrix.os only specifies linux?

Sounds good, will do, thanks :+1:

.github/workflows/github-page.yml
```diff
@@ -82,6 +57,22 @@ jobs:
         sudo apt-get -y remove --purge software-properties-common
         sudo apt-get -y autoremove
 
+    - name: Install secp256k1
+      if: matrix.os == 'ubuntu-latest'
```
amesgen, line 61: Just noted that this was `ubuntu-20.04` before, which is incorrect in this PR. Now fixed.

PR #3717, opened 2022-04-20
CAD-4098 add Flush cmd to OnDisk tests

Jasagredo:


ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -944,29 +953,29 @@ runDB standalone@DB{..} cmd =
...
           bs <- readTVar dbBackingStore
           let (toFlush, db') = ledgerDbFlush DbChangelogFlushAllImmutable db
           modifyTVar dbState (\(rs, _) -> (rs, db'))
-          pure (toFlush, bs, db')
+          pure (toFlush, bs)
         flush bs toFlush
-
+        pure Flushed 
+    go hasFS Snap = do
+        _ <- go hasFS Flush
```
Jasagredo, line 967:
```
-- TODO: Maybe @Flush@ should be its own command but that would require
-- the implementation to change so that 'ledgerDbOldest' doesn't require
-- a 'flush' to happen just before it is called.
```

Add some rewording of this comment back

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/OnDisk.hs
```diff
@@ -1189,7 +1198,8 @@ shrinker _ (At cmd) =
     case cmd of
       Current      -> []
       Push _b      -> []
-      Snap         -> []
+      Snap         -> [At Flush]
```
dnadales, line 1201: Hmm, I'm not sure about this. Specially now that we removed the flush operation before taking a snapshot.

PR #3669, opened 2022-03-17
Reduce Nix closure size

amesgen:
Reduce the closure size of `release.nix` (useful for remote building) by filtering the source (such that e.g. `dist-newstyle` and `.git` are not considered to be part of the derivation).

Before:
```
 $ nix path-info --derivation -Sh $(nix-instantiate release.nix)
/nix/store/833krq3p3jr7s7mbi3lwhb5791xv5d54-github-required.drv                      	 913.3M
/nix/store/940dy107863a26g6rs1h7ab2n8k9zn7d-ouroboros-network-0.1.0.0-tests-win64.drv	  21.4M
```
After:
```
/nix/store/4xchhdk69d1wprccm4m5f22dnjskkmih-github-required.drv                      	  54.7M
/nix/store/940dy107863a26g6rs1h7ab2n8k9zn7d-ouroboros-network-0.1.0.0-tests-win64.drv	  21.4M
```

PR #3720, opened 2022-04-21
Fix and unify stylish-haskell checks 

amesgen:
Previously, the nix stylish-haskell checks where only checking that the files *can* be formatted (no errors due to e.g. CPP), but not that they actually are formatted, as `stylish-haskell -i` will succeed with exit code 0 even if it some of its input files were not already formatted. This PR fixes this by using the `exit_code: error_on_format` of stylish-haskell.

In addition:
  - We reuse the scripts in `scripts/ci` such that people can easily run stylish-haskell in the same way as it is checked in CI.
  - We use `cleanGit` to only copy the git-tracked files to the nix store. Previously, the entire local development directory would be copied (easily >1GB). This subsumes #3669.
  - `checks.styles.check-stylish-network` is removed and part of `nix-build -A checks.styles.check-stylish`.

To test this:
  1. Introduce some change which makes the file unformatted.
  2. Check that `nix-build -A checks.styles.check-stylish` fails.


amesgen:
There seem to be some additional files which have to be excluded due to CPP, will fix.

amesgen:
Hmm, the Hydra checkout of the repo does not include the `.git` dir, so `git diff` does not work. Looking for a workaround...

amesgen:
Now using the `exit_code: error_on_format` option of stylish-haskell.

amesgen:
This is now ready for review, sorry for not opening this as a Draft PR first. The Hydra failure for test-storage is not related (known to be flaky).

coot (approved):
LGTM

amesgen:
Windows CI seems to be failing due to an OOM error. I am merging it manually, as we will have bors run when merging #3719 either way.

PR #3722, opened 2022-04-26
Adding diff to checks

craigem:
I envisage that this PR will supersede [#3719](https://github.com/input-output-hk/ouroboros-network/pull/3719) and provide the following: 

- [X] Diff output for changes

Not set to **Draft** as it needs to run in CI.

Diff output has been tested:

* **check-stylish:** a test [formatting change](https://hydra.iohk.io/build/14042003/nixlog/1) I made
* **check-stylish-network:** real [formatting issues](https://hydra.iohk.io/build/14041511/nixlog/1) it finds

It also lays the basic framework by which CI scripts can be unified into one source, rather than two.


coot (approved):
LGTM

nix/check-stylish-network.nix
```diff
@@ -15,5 +15,15 @@ runCommand "check-stylish-network" {
   fd -p network-mux -e hs -E Setup.hs -E network-mux/src/Network/Mux/Bearer/Pipe.hs -E network-mux/src/Network/Mux/Channel.hs -X stylish-haskell -c .stylish-haskell-network.yaml -i
   fd -p ouroboros-network -e hs -E Setup.hs -X stylish-haskell -c .stylish-haskell-network.yaml -i
```
coot, line 16: In #3719 I noticed that the above does matches all the files if the parent directory is also called `ouroboros-network`.  It should be done as in this [patch](https://github.com/input-output-hk/ouroboros-network/pull/3719/commits/3167ff4bb12113f7aa63f52edb85bd7ae7c20cc9).  I think you can just cherry-pick it.

coot (changes requested):

nix/check-stylish-network.nix
craigem: Cherries picked! :smiley:

coot (approved):
LGTM

PR #3727, opened 2022-04-29
Public Root Peers target Diffusion tests

bolt12:


ouroboros-network/test/Test/Ouroboros/Network/PeerSelection.hs
```diff
@@ -752,6 +757,169 @@ check_governor_connstatus _ trace0 =
...
+                   <*> requestOpportunities
+                   <*> requestOpportunitiesIgnoredTooLong)
+
+-- | A variant of 'prop_governor_target_established_below' but for the target
+-- that any public root peers should become established.
+--
+-- We do not need separate above and below variants of this property since it
+-- is not possible to exceed the target.
+--
+prop_governor_target_established_public :: GovernorMockEnvironment -> Property
```
coot, line 828: Hmm, what if we adopt the following naming convention:
`prop_diffusion_target_established_below` where we use `prop_diffusion` instead of `prop_governor` with the same suffix as the original test.

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -382,6 +386,186 @@ prop_diffusion_dns_can_recover defaultBearerInfo diffScript =
...
+            lastTime = fst
+                     . last
+                     $ evsList
+         in classifySimulatedTime lastTime
+          $ classifyNumberOfEvents (length evsList)
+          $ verify_target_established_public ev
+        )
+      <$> events
+  where
+    verify_target_established_public :: Events DiffusionTestTrace -> Property
```
coot, line 440: Could this be exported by the `Test.Ouroboros.Network.Governor` module.  The same for the other such functions.

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection.hs
bolt12: I am not sure why because these are not run in Diffusion, so it makes sense to be `prop_governor`

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
bolt12: Not quite because although the code looks pretty much the same we are using different ways of obtaining the trace!

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection.hs
```diff
@@ -752,6 +757,169 @@ check_governor_connstatus _ trace0 =
 -- Progress properties
 --
 
+-- | Check that the governor can hit (but not overshoot) its target for the
+-- number of root peers. This has to be bounded by what is possible: we cannot
+-- always find enough peers, and when we can, some of them fail.
```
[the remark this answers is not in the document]
bolt12: Done!

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection.hs
coot: you're right, ok.

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
coot: I'd extract the core function which does the validation, and export it.  This will allows us to use it in the two different contexts.  But this can be done in a separate PR if you prefer.

coot (approved):
LGTM

bolt12:
bors merge

PR #3734, opened 2022-05-05
ChainSync server: prevent duplicate `RollForward` messages

amesgen:
Closes [CAD-4228](https://input-output.atlassian.net/browse/CAD-4228).

#### Problem description

Consider the following very common scenario prior to this PR:

 1. Assume that the node is idle, having selected a chain with tip A

 2. It receives a new block B that extends A

 3. We set the tentative header to B

 3. This wakes up the ChainSync server thread (in an `STM` transaction
    in `instructionHelper`)

     - The roll state is updated from `RollForwardFrom A` to `RollForwardFrom B`.

     - The chain update `AddBlock B` is emitted.

 4. After successful validation, the node selects B which involves a
    call to `fhSwitchFork` with intersection point A.

     - Because the intersection point is now behind the roll state
       point, the roll state is set to `RollBackTo A`.

     - Hence, the follower emits/the ChainSync server sends a
       `RollBack A`, immediately followed by another `AddBlock B`.

These extra messages are redundant, and will be no longer sent with
this PR.

#### Diagnosis

The arguments to `fhSwitchFork` are unaware of the tentative header,
so for a *tentative* follower it looks like a rollback is necessary.

Before pipelining, the call to `fhSwitchFork` was a no-op when rolling
forward by a block because the passed intersection point did not
precede (in fact, was equal to) the roll state point.

#### Possible solutions

(note that these are not mutually exclusive)

 1. Make the call to `fhSwitchFork` with arguments aware of the
    tentative header.

 2. Don't call `fhSwitchFork` when selecting our tentative header.

 3. Don't call `fhSwitchFork` when extending our chain.

We went with the third, as it stresses that `fhSwitchFork` is only
necessary when actually switching to a fork, and it is most
straightforward to implement.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
@@ -376,6 +376,11 @@ olderThanK hdr isEBB immBlockNo
   where
     bNo = blockNo hdr
 
+-- | When we switch to a new selected chain, we are either extending the current
+-- chain by adding blocks on top or we are switching to a fork.
+data ChainSwitchType = AddingBlocks | SwitchingToAFork
+  deriving (Show, Eq)
+
 -- | Return the new tip.
 chainSelectionForFutureBlocks
   :: ( IOLike m
@@ -601,7 +606,7 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
                 switchTo
                   validatedChainDiff
                   (varTentativeHeader chainSelEnv)
-                  AddedToCurrentChain
+                  AddingBlocks
       where
         chainSelEnv = mkChainSelEnv curChainAndLedger
         curChain    = VF.validatedFragment curChainAndLedger
@@ -662,7 +667,7 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
                 switchTo
                   validatedChainDiff
                   (varTentativeHeader chainSelEnv)
-                  SwitchedToAFork
+                  SwitchingToAFork
       where
         chainSelEnv = mkChainSelEnv curChainAndLedger
         curChain    = VF.validatedFragment curChainAndLedger
@@ -714,16 +719,9 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
          -- ^ Chain and ledger to switch to
       -> StrictTVar m (StrictMaybe (Header blk))
          -- ^ Tentative header
-      -> (    [LedgerEvent blk]
-           -> NewTipInfo blk
-           -> AnchoredFragment (Header blk)
-           -> AnchoredFragment (Header blk)
-           -> TraceAddBlockEvent blk
-         )
-         -- ^ Given the 'NewTipInfo', the previous chain, and the new chain,
-         -- return the event to trace when we switched to the new chain.
+      -> ChainSwitchType
       -> m (Point blk)
-    switchTo vChainDiff varTentativeHeader mkTraceEvent = do
+    switchTo vChainDiff varTentativeHeader chainSwitchType = do
         (curChain, newChain, events, prevTentativeHeader) <- atomically $ do
           curChain  <- readTVar         cdbChain -- Not Query.getCurrentChain!
           curLedger <- LgrDB.getCurrent cdbLgrDB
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -745,17 +742,21 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
...
 
-              -- Update the followers
-              --
-              -- 'Follower.switchFork' needs to know the intersection point
-              -- (@ipoint@) between the old and the current chain.
-              let ipoint = castPoint $ Diff.getAnchorPoint chainDiff
-              followerHandles <- Map.elems <$> readTVar cdbFollowers
-              forM_ followerHandles $ \followerHandle ->
-                fhSwitchFork followerHandle ipoint newChain
+              when (chainSwitchType == SwitchingToAFork) $ do
```
nfrisby, line 745: I would rather have a `case` with exhaustive explicit alternatives (mostly trivial), so that any changes to the data type will necessitate intentional (probably trivial) changes here.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -714,16 +717,10 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
...
-      -> (    [LedgerEvent blk]
-           -> NewTipInfo blk
-           -> AnchoredFragment (Header blk)
-           -> AnchoredFragment (Header blk)
-           -> TraceAddBlockEvent blk
-         )
-         -- ^ Given the 'NewTipInfo', the previous chain, and the new chain,
-         -- return the event to trace when we switched to the new chain.
+      -> ChainSwitchType
+         -- ^ Whether we are extending the current chain or switching to a fork.
```
nfrisby, line 721: I think this comment would be better suited as the Haddock on `ChainSwitchType`.

nfrisby (commented):
I was hoping to review all your neglected PRs tonight, but I've run out of steam: I don't currently have the focus to properly confirm that removing these `fhSwitchFork` calls is the right way forward. I'll try to get it tomorrow.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -745,17 +743,23 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
...
-              -- Update the followers
-              --
-              -- 'Follower.switchFork' needs to know the intersection point
-              -- (@ipoint@) between the old and the current chain.
-              let ipoint = castPoint $ Diff.getAnchorPoint chainDiff
-              followerHandles <- Map.elems <$> readTVar cdbFollowers
-              forM_ followerHandles $ \followerHandle ->
-                fhSwitchFork followerHandle ipoint newChain
+              case chainSwitchType of
+                AddingBlock      -> pure ()
```
nfrisby, line 747: Please add a comment here about why it's safe/important to skip `fhSwitchFork` here.

And update the PR description.

(As we just agreed on the call; this is merely reifying that.)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -376,6 +376,11 @@ olderThanK hdr isEBB immBlockNo
   where
     bNo = blockNo hdr
 
+-- | When we switch to a new selected chain, we are either extending the current
```
nfrisby, line 379: We might be adding more than one block to the chain.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -745,17 +743,28 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish = do
...
-              -- Update the followers
-              --
-              -- 'Follower.switchFork' needs to know the intersection point
-              -- (@ipoint@) between the old and the current chain.
-              let ipoint = castPoint $ Diff.getAnchorPoint chainDiff
-              followerHandles <- Map.elems <$> readTVar cdbFollowers
-              forM_ followerHandles $ \followerHandle ->
-                fhSwitchFork followerHandle ipoint newChain
+              case chainSwitchType of
+                -- When adding a block, the intersection point of the old and
```
nfrisby, line 747: We might be adding more than one block. The logic is correct still, but the wording shouldn't say "adding a block"

nfrisby (approved):
Just minor things! Thanks much. Approving (and pinging you on Slack to clarify).

Where you wrote "Previously" in the PR description, I think you mean "before pipelining" not just "before this PR". Please replace that word.

amesgen:
bors merge

amesgen:
bors retry

PR #3737, opened 2022-05-09
Optimise caching

newhoggy:


.github/workflows/build.yml
```diff
@@ -142,18 +142,28 @@ jobs:
...
     - name: Update Hackage index
       run: cabal update
 
     - name: Cabal Configure
       run: cabal --builddir="$CABAL_BUILDDIR" configure --enable-tests
 
+    - name: Record dependencies
+      run: |
+        cabal build all --dry-run
+        cat ${{ env.PLAN_JSON }} | jq -r '."install-plan"[].id' | sort | uniq > dependencies.txt
```
amesgen, line 154: Where does `PLAN_JSON` come from?

amesgen (approved):
The CI run succeeded and the cache got updated (and the cache name now depends on the set of dependencies, so when they change, we will store a new cache entry, in contrast to the current situation), so it seems like this should work.
Thanks, LGTM!

.github/workflows/build.yml
newhoggy: Good spot.  I've hard coded it to `dist-newstyle/cache/plan.json` now.

amesgen:
@newhoggy What are we waiting for to merge this PR? Bors timeouts are again plaguing us in other PRs.

angerman:
LGTM. I'm still suspicious of the GH Actions. But for a first approximation of correctness this should be fine. My core concern is that this is not identical to how we build our artifacts we distribute.

amesgen:
Yes, I consider this PR to be a intermittent alleviation of our CI woes, and I am looking forward to running the testing artifacts built by Hydra :+1:

PR #3741, opened 2022-05-10
Provide correct bound for the VRF check.

nc6:
Since in Praos the leader value doens't derive directly from the VRF
evaluation but from a hash, the bounds are different. This change (in
conjunction with a related change in the ledger) means that one has to
provide the bounds explicitly when comparing the leader value.


ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/VRF.hs
```diff
@@ -100,9 +103,11 @@ vrfLeaderValue ::
   Crypto c =>
   proxy c ->
   CertifiedVRF (VRF c) InputVRF ->
-  Natural
-vrfLeaderValue p =
-  bytesToNatural . hashToBytes . hashVRF p SVRFLeader
+  BoundedNatural
+vrfLeaderValue p cvrf =
+  assertBoundedNatural
+    (8 * fromIntegral (sizeHash (Proxy @(HASH c))))
```
kderme, line 109: Shouldn't be 2 ^ this?

nc6: Yes, yes it should!

dcoutts (approved):
LGTM.

Including the referenced ledger commit to adjust `checkLeaderValue` to use `BoundedNatural`.

nc6:
bors merge

nc6:
bors merge

nc6:
bors retry

nc6:
bors retry

nc6:
bors retry

PR #3744, opened 2022-05-16
Remove submission queue

Jasagredo:


Jasagredo:
I'm don't particularly enjoy the exception mechanism used in the LMDB module. We just rely on throwing exceptions into IO with `Exception.throw` but then we should wrap every call to LMDB code in a `catch` or sth like that.

cc @jorisdral @dnadales

dnadales (approved):
:muscle:

PR #3754, opened 2022-05-18
Bugfix in HFC: do not consider the last known era to be eternal

nfrisby:
This supersedes PR #3750. And it unblocks the Vasil HF.

It's in Draft only because I've not yet concluded what we need to do about the node-to-client version, since I'm changing the serialization scheme used in the answer to the `GetInterpreter` query. And because I haven't written a good enough explanation in this PR description -- I'm opening it now, so that it's available to downstream folk.


ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/State/Infra.hs
```diff
@@ -170,32 +170,41 @@ reconstructSummary (History.Shape shape) transition (HardForkState st) =
...
+                    , eraEnd    = applySafeZone
+                                    nextParams
+                                    nextStart
+                                    (boundSlot nextStart)
+                    }
+              ExactlyNil               ->
+                -- HOWEVER, this code doesn't know what that era is! This can
+                -- arise when a node has not updated its code despite an
+                -- imminent hard fork.
+                NonEmptyOne EraSummary {
```
nc6, line 203: This seems an interesting case - how do we know about the hard fork before the node is updated? We could know about the imminent PV bump, but a PV bump does not imply a hard fork. We only relate a PV bump to a hard fork through declarations in code.

nc6 (commented):
Looks reasonable. I've left one comment about a case which seems to me to be impossible.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/State/Infra.hs
```diff
@@ -170,32 +170,41 @@ reconstructSummary (History.Shape shape) transition (HardForkState st) =
     go :: Exactly xs' EraParams
        -> Telescope (K Past) (Current f) xs'
        -> NonEmpty xs' EraSummary
+    go ExactlyNil              t                   = case t of {}
```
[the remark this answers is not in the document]
nfrisby: Yeah, I was confused at one point until I noticed it was located below the interesting cases, so I just pulled it to the top.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/State/Infra.hs
nfrisby: Ah, OK. I've looked more carefully now at how `TransitionKnown` _actually_ arises.

First step backwards from my focus on`reconstructSummary` ("the HFC part"):

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/State.hs#L112-L145

Second step ("the `ShelleyBlock` part"): 

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/ShelleyHFC.hs#L175-L187

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/ShelleyHFC.hs#L207-L217

Third step ("the `protocolInfoCardano` part"):

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L940-L961

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L632-L638

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L659-L665

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L683-L689

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L707-L713

https://github.com/input-output-hk/ouroboros-network/blob/dfa2697c2bcdefdc0a97db7f988bcf90aaff47b5/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs#L536-L555

Fourth step ("the `cardano-node` part"):

https://github.com/input-output-hk/cardano-node/blob/0b32f893a170866deca50dc9ef303c0a11ba37f0/cardano-node/src/Cardano/Node/Protocol/Cardano.hs#L217-L284

I think what you're observing is that:

* We wouldn't reach the `TransitionKnown & ExactlyNil` case of `reconstructSummary.go` because code with the last known era `X` would not even have a `ProtocolTransitionParamsShelleyBased` argument for the `protocolInfoCardano` -- unless something had gone catastrophically off the rails in our PR review process.

I would agree with that, with a couple caveats:

* I agree because of the `ShelleyBlock` instance and the pattern in `protocolInfoCardano` -- but that's "by convention".

* Moreover, `reconstructSummary` is more abstract than `ShelleyBlock`.

I think, at the very least, I should write a comment in this case of `reconstructSummary.go` explaining (as in this comment) why we think it's impossible for `CardanoBlock`.

nfrisby: I've force pushed a commit that adds a corresponding paragraph to the comment.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Summary.hs
```diff
@@ -525,50 +516,27 @@ instance SListI xs => Serialise (Summary xs) where
...
   -- - If @|xs| > |ys|@, by definition @n < |xs|@, and hence we will not modify
   --   the era summary: this is what we wanted.
   --
-  -- - If @|xs| == |ys|@, then at most @n == |xs|@, in which case we might
-  --   "modify" the final era to be unbounded. But in this case, the server will
-  --   consider that era to be final as well, and so it will _already_ be
-  --   unbounded: effectively this means that this means we will leave the
-  --   summary unchanged.
+  -- - If @|xs| == |ys|@, then at most @n == |xs|@.
   decode = do
```
nfrisby, line 532: I think it'd be OK to push this change without introducing a version gate and explicit backwards-compatibility logic.

* This is only used by node-to-client messaging: nodes do not send `Summary`s to other nodes.
* Specifically, nodes serialize `Summary` and clients deserialize it (via the LocalStateQuery mini protocol).
* This PR only changes the behavior of `decode`.
* We're changing it by removing a post-processing step.

The consequences of a "mismatched" node and client would be:

* if an old node is communicating with an new client, then this change is inconsequential: the old node is already sending a `Summary` that ends with an eternal Alonzo and the new client would deserialize it exactly as such (which is also how the old client would deserialize it).

* if a new node was communicating with an old client, the old client would deserialize the exact same `Summary` that it would deserialize if it were communicating with an old node. The new node would send an `Alonzo` that is not eternal (just because Alonzo is no longer the last known era -- so irrespective of this PR), but the older client would `fixEndBound` so that it was. The old node would have sent an eternal `Alonzo`, which the old client would also redundantly call `fixEndBound` on.

* Note that the previous bullet is true regardless of whether the new node has reached Babbage. If it has, then it sends a longer `Summary`, but the old client would trim off the Babbage part, leaving a non-eternal Alonzo. If it hasn't, then the node would send a `Summary` ending in a non-eternal Alonzo. Either way, the old client deserializes a `Summary` that ends with Alonzo doctored by `fixEndBound` to be eternal.

In summary: only a new client communicating with a the new node would deserialize a non-eternal Alonzo. All other configurations have the client deserializing an eternal Alonzo. So we effectively get backwards-compatibility for free, here (due to the old code's logic).

I haven't yet catalogued the ways in which the known clients (eg Cardano API, wallet, etc) use the `Summary`. I imagine some usages could give results affected by this last era. But what harm could that actually do? EG maybe it would somewhat contribute to the operator being able to _create_ transactions with invalid tx validity intervals. But -- at least before the first release after `mainnet` reaches Babbage -- that's unproblematic because of PR input-output-hk/cardano-ledger#2785: nodes will continue to accept such validity ranges in Alonzo.

nfrisby: Once `mainnet` reaches Babbage, no one can be running an "old node" nor an "old client", because those things wouldn't be able to handle the Babbage txs and blocks that `mainnet` is now working with. And at that point we can considering revisiting input-output-hk/cardano-ledger#2785 to only extend the `EpochInfo` for a finite set of grandfathered-in txs.

So the possibility of this "mismatched" scenario is temporary.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/State/Infra.hs
nc6: That's a very comprehensive explanation! I agree with you that it makes sense to both leave this case in as-is and explain in comments why it's a rare case to see.

nc6 (approved):

dcoutts (approved):
I agree this is the right thing.

There will be an observable change in behaviour for new clients talking to new nodes, where they will no longer be able to translate slot <-> time beyond the usual forecast window. This is of course the change we want, but it's possible that some clients were incorrectly trying to do conversions further into the future than ought to be possible. So we will need to rely on integration testing for those clients against the new node to check that they were not doing this. It ought to be noisy if they were.

nfrisby:
bors merge

nfrisby:
bors retry

nfrisby:
bors cancel

nfrisby:
bors merge

PR #3757, opened 2022-05-19
Extract generation of `ChainUpdate`s from the ChainSync client test

amesgen:
Closes [CAD-4313](https://input-output.atlassian.net/browse/CAD-4313).

This will be used by the upcoming consensus-level BlockFetch client tests.

### DIFF
--- a/ouroboros-consensus-test/src/Test/Util/ChainUpdates.hs
+++ b/ouroboros-consensus-test/src/Test/Util/ChainUpdates.hs
@@ -1,0 +1,188 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts   #-}
+{-# LANGUAGE LambdaCase         #-}
+
+-- | Generate sequences of updates to model an evolving chain
+module Test.Util.ChainUpdates (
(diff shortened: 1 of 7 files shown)


ouroboros-consensus-test/src/Test/Util/ChainUpdates.hs
```diff
@@ -0,0 +1,184 @@
...
+    tot = sum (map fst xs0)
+
+    pick n ((k,x):xs)
+      | n <= k    = x
+      | otherwise = pick (n-k) xs
+    pick _ _  = error "pick used with empty list"
+
+-- | Test that applying the generated updates gives us the same chain as
+-- '_currentChain'.
+prop_genChainUpdates :: SecurityParam -> Int -> Property
```
amesgen, line 180: It seemed like the easiest option to put this here (as this uses `ChainUpdateState` and related functions, which are not exported); but we could also create a `Test.Util.ChainUpdates.Internal` module as an alternative.

nfrisby (approved):
We reviewed on a call. Approving. Thanks!

amesgen:
bors merge

PR #3761, opened 2022-05-20
Fix no self in root peers in Diffusion tests

bolt12:


bolt12:
This bug is due to a node trying to connect to itself in our simulation environment. In a real scenario this would be seen as a simultaneous open, but our simulated socket sees `ip:port` and `ip:port'` as being different machines and so both `connect` and `accept` call succeed in the same machine.

coot (approved):

bolt12:
bors retry

PR #3765, opened 2022-05-20
CAD-4398 Add LMDB to github actions CI

Jasagredo:
Together with #3733, these should bring CI back to life for the UTxO-HD LMDB branch


Jasagredo:
Still fails to find `lmdb` library

angerman:
The behaviour here stumps be enough that I've resorted to try this in a VM now.

PR #3766, opened 2022-05-20
Add `COMPLETE` pragma for `FallingEdge` pattern synonym

amesgen:


nfrisby (approved):
> If a function definition matches on all the constructors specified in the pragma then the compiler will produce no warning.

So the fact that `FallingEdge` fixes `a` to `()` is sufficient for the logic here. 👍

amesgen:
bors merge

amesgen:
bors retry

amesgen:
bors retry

PR #3767, opened 2022-05-21
CAD-4307 LMDB exception handling

jorisdral:
This PR comes up with a strategy for exception handling in the `LedgerDB.HD.LMDB` module.

#### Error propagation

The `LMDB` module uses `Control.Exception.throw` to throw exceptions, which will be propagated up to the nearest handler or kill the thread. Using `ExceptT` instead would force us to act on exceptions, so it might be beneficial to let (parts of) the `LMDB` module run in `ExceptT DbErr`. This depends in part on the following:

There is no meaningful way to recover from most exceptions from within the `LMDB` module. Most should be fatal for the node as well: If we can't find a UTxO on-disk while we are trying to read it, then the node probably has no meaningful way to recover from this exception. However, when we try to replay a snapshot and encounter a "fatal" exception, we could instead discard the current snapshot and try a different one. The question is whether it would be better to let the node die anyway, since the snapshot could still be there on the next node startup.

- [x] Decide which errors should be "fatal".
- [x] Depending on the previous, decide whether it makes sense to have parts run in  `ExceptT DbErr`. 

#### Other

- [x] Decide whether we should store the database settings in the database itself.
- [x] Decide on a strategy for `LMDBLimits`.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
@@ -49,7 +49,7 @@ import qualified Ouroboros.Consensus.Storage.FS.API.Types as FS
 import qualified Ouroboros.Consensus.Storage.LedgerDB.HD as HD
 import qualified Ouroboros.Consensus.Storage.LedgerDB.HD.BackingStore as HD
 import           Ouroboros.Consensus.Util (foldlM')
-import           Ouroboros.Consensus.Util.IOLike (IOLike, bracket, onException)
+import           Ouroboros.Consensus.Util.IOLike (IOLike, MonadCatch(..), MonadThrow(..), bracket, onException)
 
 import qualified Database.LMDB.Raw as LMDB
 import qualified Database.LMDB.Simple as LMDB
@@ -87,6 +87,7 @@ data TraceDb
   | TDBValueHandle Int TraceValueHandle
   | TDBTableOp     TraceTableOp
   | TDBInitialisingFromLMDB !FS.FsPath
+  | TDBInitialisingFromLMDBDone !FS.FsPath
   deriving (Show, Eq)
 
 data TraceTableOp = TTO
@@ -151,8 +152,6 @@ defaultLMDBLimits = LMDB.defaultLimits
   }
 
 -- TODO a version field.
--- TODO(jdral): I've made a distinction between DbSettings and DbState because
--- /settings/ should probably not be updated on every flush.
 data DbSettings = DbSettings
 
 newtype DbState = DbState {
@@ -392,18 +391,45 @@ withDbStateRWMaybeNull db f  =
 
 data GuardDbDir  = GDDMustExist | GDDMustNotExist
 
-guardDbDir :: MonadIO m => GuardDbDir -> FS.SomeHasFS m -> FS.FsPath -> m FilePath
+-- | Guard for the existence/non-existence of a database directory,
+-- and create it if missing.
+guardDbDir ::
+     (MonadIO m, IOLike m)
+  => GuardDbDir
+  -> FS.SomeHasFS m
+  -> FS.FsPath
+  -> m FilePath
 guardDbDir gdd (FS.SomeHasFS fs) path = do
-  FS.doesFileExist fs path >>= \b -> when b $ Exn.throw $ DbErrStr $ "guardDbDir:must be a directory:" <> show path
-  FS.doesDirectoryExist fs path >>= \case
-    True | GDDMustNotExist <- gdd ->
-           -- TODO Should throw a DbErrDbExists exception
-           -- Callers should delete the directory if they want to restore a snapshot
-           FS.removeDirectoryRecursive fs path
-    False | GDDMustExist <- gdd -> Exn.throw $ DbErrDbDoesntExist path
-    _ -> pure ()
+  fileEx <- FS.doesFileExist fs path
+  when fileEx $
+    Exn.throw $ DbErrStr $ "guardDbDir:must be a directory: " <> show path
+  dirEx <- FS.doesDirectoryExist fs path
+  case dirEx of
+    True  | GDDMustNotExist <- gdd -> throwIO $ DbErrDbExists path
+    False | GDDMustExist    <- gdd -> throwIO $ DbErrDbDoesntExist path
+    _                              -> pure ()
   FS.createDirectoryIfMissing fs True path
   FS.unsafeToFilePath fs path
+
+-- | Same as @`guardDbDir`@, but retries the guard if we can make meaningful
+-- changes to the filesystem before we perform the retry.
+--
+-- Note: We only retry if a database directory exists while it shoudn't: In
+-- this case, we remove the directory recursively before retrying the guard.
+guardDbDirWithRetry ::
+     (MonadIO m, IOLike m)
+  => GuardDbDir
+  -> FS.SomeHasFS m
+  -> FS.FsPath
+  -> m FilePath
+guardDbDirWithRetry gdd shfs@(FS.SomeHasFS fs) path =
+    handle retryHandler (guardDbDir gdd shfs path)
+  where
+    retryHandler e = case (gdd, e) of
+      (GDDMustNotExist, DbErrDbExists _path) -> do
+        FS.removeDirectoryRecursive fs path
+        guardDbDir GDDMustNotExist shfs path
+      _ -> throwIO e
 
 {-------------------------------------------------------------------------------
  Initialize an LMDB
@@ -428,20 +454,27 @@ initFromVals dbsSeq vals Db{..} = liftIO $ LMDB.readWriteTransaction dbEnv $
                 $> ((), DbState{dbsSeq})
     Just _ -> Exn.throw $ DbErrStr "initFromVals: db already had state"
 
-initFromLMDBs :: (MonadIO m, IOLike m)
+initFromLMDBs ::
+     (MonadIO m, IOLike m)
   => Trace.Tracer m TraceDb
+  -> LMDBLimits
   -> FS.SomeHasFS m
   -> FS.FsPath
   -> FS.FsPath
   -> m ()
-initFromLMDBs tracer shfs from0 to0 = do
-  Trace.traceWith tracer $ TDBInitialisingFromLMDB from0
-  from <- guardDbDir GDDMustExist shfs from0
-  to <- guardDbDir GDDMustNotExist shfs to0
-  bracket
-    (liftIO $ LMDB.openEnvironment from defaultLMDBLimits) -- TODO assess limits, in particular this could fail if the db is too big
-    (liftIO . LMDB.closeEnvironment)
-    (flip (lmdbCopy tracer) to)
+initFromLMDBs tracer limits shfs from0 to0 = do
+    Trace.traceWith tracer $ TDBInitialisingFromLMDB from0
+    from <- guardDbDir GDDMustExist shfs from0
+    to <- guardDbDirWithRetry GDDMustNotExist shfs to0
+    bracket
+      -- TODO assess limits, in particular this could fail if the db is too big
+      -- TODO(jdral): Should we read from some configuration/settings file what
+      -- the limits of the "from" database are? Do we want to support
+      -- interoperablity between databases with different limits?
+      (liftIO $ LMDB.openEnvironment from limits)
+      (liftIO . LMDB.closeEnvironment)
+      (flip (lmdbCopy tracer) to)
+    Trace.traceWith tracer $ TDBInitialisingFromLMDBDone to0
 
 lmdbCopy :: MonadIO m
   => Trace.Tracer m TraceDb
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -132,8 +145,11 @@ defaultLMDBLimits = LMDB.defaultLimits
   , LMDB.maxReaders = 16
   }
 
-newtype DbState  = DbState
-  { dbsSeq :: WithOrigin SlotNo  -- TODO a version field
+-- TODO a version field.
+data DbSettings = DbSettings
```
jorisdral, line 149: The settings are currently stored in the database itself, but this may not be sensible: We can only access them *after* we have opened both the `LMDB` environment and corresponding `LMDB` settings database, while these settings could be required *before* opening the environment.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -306,24 +348,31 @@ initFromVals ::
...
-    (flip (lmdbCopy tracer) to)
+initFromLMDBs tracer limits shfs from0 to0 = do
+    Trace.traceWith tracer $ TDBInitialisingFromLMDB from0
+    from <- guardDbDir GDDMustExist shfs from0
+    to <- guardDbDirWithRetry GDDMustNotExist shfs to0
+    bracket
+      -- TODO assess limits, in particular this could fail if the db is too big
+      -- TODO(jdral): Should we read from some configuration/settings file what
+      -- the limits of the "from" database are? Do we want to support
+      -- interoperablity between databases with different limits?
```
jorisdral, line 371: This requires a strategy.

dnadales:
Some observations on the PR's comment:

> There is no meaningful way to recover from most exceptions from within the LMDB module. Most should be fatal for the node as well: If we can't find a UTxO on-disk while we are trying to read it, then the node probably has no meaningful way to recover from this exception.

I agree with this.


> However, when we try to replay a snapshot and encounter a "fatal" exception, we could instead discard the current snapshot and try a different one.

I think that if a snapshot causes a fatal error in the LMDB module then there must be something wrong in the database itself, so we should not proceed anyway. 

> The question is whether it would be better to let the node die anyway, since the snapshot could still be there on the next node startup.

I'd be inclined to let the node die, for the reason I described above.

  
> Other
>  Decide whether we should store the database settings in the database itself.
>  Decide on a strategy for LMDBLimits.

Are these tasks related to this PR? They seem out of scope

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -81,6 +81,7 @@ data TraceDb
   | TDBValueHandle Int TraceValueHandle
   | TDBTableOp     TraceTableOp
   | TDBInitialisingFromLMDB !FS.FsPath
+  | TDBInitialisingFromLMDBDone !FS.FsPath
```
dnadales, line 84: We don't use `Done` in the events above, but the past tense of the verb instead. Maybe we want to do the same here, ie `TDBInitialisedFromLMDB`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -96,8 +97,13 @@ data TDBTableOp
  Errors
 -------------------------------------------------------------------------------}
 
-data DbErr = DbErrStr !String
+data DbErr =
```
dnadales, line 100: Should we add comments for the new error constructors?

jorisdral:
@dnadales They are somewhat related in the sense that we may want to throw exceptions when settings are missing/incompatible with an existing database/problematic for some other reason, or if the limits are somehow incompatible with an existing database. I figured it might be good to come up with a strategy here such that we can also come up with sensible exceptions. I'd be happy to move them out of the scope of this PR though.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -96,8 +97,13 @@ data TDBTableOp
...
+    DbErrStr !String
   | DbErrNoSettings
+  | DbErrNoDbState
+  -- | The sequence number of a @`Db`@ should be monotonically increasing
+  -- across calls to @`bsWrite`@, since we use @`bsWrite`@ to flush
+  -- /immutable/ changes.
```
dnadales, line 106: Here we might want to explain how are the monotonically increasing property and the flushing related. I guess this has to do with the fact that we cannot flush "older" sequence numbers. Maybe we could add a small clarifying sentence.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
dnadales: Good point. Shouldn't the settings be passed in a configuration file? 

Also do these settings refer exclusively to the LMDB settings? If so we might want to rename this data type.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -252,42 +268,68 @@ initRangeReadLMDBTable count (LMDBMK _        db)                      = ApplyVa
 rangeReadLMDBTable     count (LMDBMK _        db) (ApplyKeysMK   keys) = ApplyValuesMK    <$> lmdbRangeReadTable count db keys
 
 {-------------------------------------------------------------------------------
- Db settings
+ Db state
```
dnadales, line 271: To discuss over a call, why did we change `settings` to `state`?

dnadales:
> @dnadales They are somewhat related in the sense that we may want to throw exceptions when settings are missing/incompatible with an existing database/problematic for some other reason, or if the limits are somehow incompatible with an existing database. I figured it might be good to come up with a strategy here such that we can also come up with sensible exceptions. I'd be happy to move them out of the scope of this PR though.

Makes sense. Thanks.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -252,42 +268,68 @@ initRangeReadLMDBTable count (LMDBMK _        db)                      = ApplyVa
...
-  FS.doesDirectoryExist fs path >>= \case
-    True | GDDMustNotExist <- gdd ->
-           -- TODO Should throw a DbErrDbExists exception
-           -- Callers should delete the directory if they want to restore a snapshot
-           FS.removeDirectoryRecursive fs path
-    False | GDDMustExist <- gdd -> Exn.throw $ DbErrDbDoesntExist path
-    _ -> pure ()
+  fileEx <- FS.doesFileExist fs path
+  when fileEx $
+    Exn.throw $ DbErrStr $ "guardDbDir:must be a directory: " <> show path
```
dnadales, line 305: I'm a bit confused when we use `throwIO` and `Exn.throw`. Shouldn't we use the former?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -252,42 +268,68 @@ initRangeReadLMDBTable count (LMDBMK _        db)                      = ApplyVa
...
+    True  | GDDMustNotExist <- gdd -> throwIO $ DbErrDbExists path
+    False | GDDMustExist    <- gdd -> throwIO $ DbErrDbDoesntExist path
+    _                              -> pure ()
   FS.createDirectoryIfMissing fs True path
   FS.unsafeToFilePath fs path
 
+-- | Same as @`guardDbDir`@, but retries the guard if we can make meaningful
+-- changes to the filesystem before we perform the retry.
+--
+-- Note: We only retry if a database directory exists while it shoudn't: In
```
dnadales, line 317: I wonder if this is what we want. Don't we risk deleting a database just because, for instance, we passed the wrong flags?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
dnadales: On a second pass I think this makes sense, but I'd still like to confirm I understood it correctly: we're basically doing this because `settings` was a misnomer.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -365,22 +414,23 @@ newLMDBBackingStore dbTracer limits sfs init_db = do
...
+  dbFilePath <- guardDbDirWithRetry gdd sfs path
 
   -- copy from another lmdb path if appropriate
   copy_db_action
 
   -- open this database
   dbEnv <- liftIO $ LMDB.openEnvironment dbFilePath limits
 
-  -- the LMDB.Database that holds the DbState (i.e. sequence number)
+  dbSettings <- liftIO $ LMDB.readWriteTransaction dbEnv $ LMDB.getDatabase (Just "_dbsettings")
```
dnadales, line 430: Do we do anything with this? If not, how will we use this?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
```diff
@@ -306,24 +348,31 @@ initFromVals ::
...
   -> FS.SomeHasFS m
   -> FS.FsPath
   -> FS.FsPath
   -> m ()
-initFromLMDBs tracer shfs from0 to0 = do
-  Trace.traceWith tracer $ TDBInitialisingFromLMDB from0
-  from <- guardDbDir GDDMustExist shfs from0
-  to <- guardDbDir GDDMustNotExist shfs to0
-  bracket
-    (liftIO $ LMDB.openEnvironment from defaultLMDBLimits) -- TODO assess limits, in particular this could fail if the db is too big
```
dnadales, line 324: So if I understand this correctly, before this PR we were using two limits: the one passed to `newLMDBBackingStore`, and `defaultLMDBLimits`. Now we're using the same limits, which are the ones passed to the `newLMDBBackingStore` function, right?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
dnadales: Don't we assume the limits to be given at this point? And isn't this a problem to be solved at the node?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
dnadales: Like I mentioned in a previous comment, I wonder if we could assume these settings are passed to `newLMDBBackingStore`.

dnadales (commented):
Thanks for this PR Joris :pray: 

I left some comments that we could discuss over a call together with @Jasagredo

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
jorisdral: You're right, I'm not being consistent with this yet. [Control-Exception:throwIO](https://hackage.haskell.org/package/base-4.14.1.0/docs/Control-Exception.html#v:throwIO) does recommend using `throwIO` instead of `throw` when in `IO`, so I'll do a pass over the code to make the usage of `throwIO` and `throw` consistent

jorisdral (commented):
Replies to review comments

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
jorisdral: We decided to keep the settings around, such that we can use it in the future.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
jorisdral: We assume that the limits are given to us.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/LMDB.hs
jorisdral: We decided to favour using `throwIO` (or `liftIO . throwIO`)

jorisdral:
> This PR comes up with a strategy for exception handling in the `LedgerDB.HD.LMDB` module.
> #### Error propagation
> ...

The strategy here is that exceptions should be fatal to the node. We put in effort to make the exception messages descriptive, and to have a unified strategy of raising exceptions.

> #### Other
> ...

We leave the settings in the database, but we do not use them as of yet. In the future, we might use the settings to keep track of a versioning scheme in case we have to make changes that are incompatible with existing LMDB databases.

With regards to the `LMDBLimits`, the strategy is that we assume that the limits are always given to us by the node.

dnadales (approved):

Jasagredo (approved):

PR #3770, opened 2022-05-25
Provide an 'EpochInfo' that can fail to ledger.

nc6:
The ledger now supports an `EpochInfo (Either Text)`, and uses the
`Left` constructor to indicate a horizon error. Unfortunately, while the
epoch info being provided had the correct _type_, it was implemented in
a degenerate way - the `Left` constructor would never be used and
instead a pure error was thrown on failures.

This PR instead provides a correct `EpochInfo` to the ledger, where the
failure is reflected in the `Left` constructor.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs
@@ -50,11 +50,13 @@ import qualified Codec.CBOR.Decoding as CBOR
 import           Codec.CBOR.Encoding (Encoding)
 import qualified Codec.CBOR.Encoding as CBOR
 import           Codec.Serialise (decode, encode)
+import           Control.Arrow (left)
 import qualified Control.Exception as Exception
 import           Control.Monad.Except
 import           Data.Coerce (coerce)
 import           Data.Functor ((<&>))
 import           Data.Functor.Identity
+import qualified Data.Text as Text
 import           Data.Word
 import           GHC.Generics (Generic)
 import           GHC.Records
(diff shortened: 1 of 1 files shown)


dcoutts (approved):
LGTM.

To be more comprehensive I think we should audit the other uses of `toPureEpochInfo`, and indeed perhaps rename it to make it clear it throws exceptions. We should file a TODO issue for that, it doesn't need to be in this PR of course.

nc6:
I did go through the other uses of `toPureEpochInfo` and think they are all legit. We otherwise use it only for epoch/slot computations for the current slot.

nc6:
bors merge

nfrisby (approved):
Hmm. I wonder if `SL.mkShelleyGlobals` used to take a pure `EpochInfo` and was changed to take an `Either` one, and the integration PR for that changed missed this because `toPureEpochInfo` is polymorphic.

So, we could try changing `toPureEpochInfo` to return `Identity` instead of `Applicative f => f` to try to prevent similar from happening again.

nfrisby:
> I did go through the other uses of `toPureEpochInfo` and think they are all legit. We otherwise use it only for epoch/slot computations for the current slot.

I see a few remaining uses of `toPureEpochInfo`.

* `tickChainDepState @Praos` passes it to `isNewEpoch` to check whether we're ticking past an epoch boundary. It converts the slot we're ticking to an epoch (which could fail) and compares epochs -- I think it should instead convert the end of the current epoch to a slot (which can't fail) and then compare slots

* `checkIsLeader @TPraos` uses it to compute the `firstSlot` (of the epoch) of the given `slot` when calling `SL.lookupInOverlaySchedule firstSlot ...`. As in the previous case, that translation can also fail, but I don't see an alternative that can't fail.

* `tickChainDepState @TPraos` uses it as does `tickChainDepState @Praos` above

So I think we should fixup the `tickChainDepState` calls to use unfailing translations instead of failing translations.

nc6:
bors retry

PR #3775, opened 2022-05-27
Fix NodeToNodeVersion for Babbage and P2P.

coot:
- Use NodeToNodeV_9 for babbage and NodeToNodeV_10 for p2p
- In non P2P mode use InitiatorOnlyDiffusionMode in all handshake negotiations

### DIFF
--- a/ouroboros-network-testing/src/Ouroboros/Network/Testing/Utils.hs
+++ b/ouroboros-network-testing/src/Ouroboros/Network/Testing/Utils.hs
@@ -5,7 +5,34 @@
 {-# LANGUAGE LambdaCase                 #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 
-module Ouroboros.Network.Testing.Utils where
+module Ouroboros.Network.Testing.Utils
+  ( -- * Arbitrary Delays
+    Delay (..)
+  , genDelayWithPrecision
+  , SmallDelay (..)
+    -- * QuickCheck Utils
+  , arbitrarySubset
+  , shrinkVector
+  , prop_shrink_valid
+  , prop_shrink_nonequal
+  -- * Tracing Utils
+  , WithName (..)
+  , WithTime (..)
+  , tracerWithName
+  , tracerWithTime
+  , tracerWithTimeName
+  , swapTimeWithName
+  , swapNameWithTime
+  , splitWithNameTrace
+  -- * Tracers
+  , debugTracer
+  , sayTracer
+  -- * Tasty Utils
+  , nightlyTest
+  , ignoreTest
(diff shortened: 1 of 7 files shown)


amesgen (commented):
FYI: We added `isPipeliningEnabled` here:

https://github.com/input-output-hk/ouroboros-network/blob/3c043a3acfdab0405198b1d3380610f0a55eb437/ouroboros-network/src/Ouroboros/Network/NodeToNode/Version.hs#L132-L135

The check for whether P2P is enabled could also be moved there, to centralize the feature checks by version.

coot:
Thanks @amesgen, that's a good idea.

coot:
Actually, I think that both `nodeDataFlow` and `localDataFlow` should belong to `Diffusion.P2P`. They are not exported, and there's no reason to make them available for downstream dependencies.

dcoutts (approved):
Sounds plausible to me.

amesgen:
> Actually, I think that both `nodeDataFlow` and `localDataFlow` should belong to `Diffusion.P2P`. They are not exported, and there's no reason to make them available for downstream dependencies.

I wasn't necessarily suggesting to move these functions, but to add sth like `isP2PEnabled` next to `isPipeliningEnabled` and use that instead of `nv >=`-like checks.

coot:
Yes, but we only do that in `nodeDataFlow`; putting it in `Ouroboros.Network.NodeToNode` would suggest that this is required by downstream dependencies or by other parts of the `ouroboros-network`.

coot:
There was a CI [failure](https://hydra.iohk.io/build/15138118/nixlog/18), but it's fixed by PR #3772 (which fixes test itself, not the code).

coot:
bors merge

dnadales (approved):

ouroboros-network-testing/src/Ouroboros/Network/Testing/Utils.hs
```diff
@@ -5,7 +5,34 @@
...
+  , tracerWithTimeName
+  , swapTimeWithName
+  , swapNameWithTime
+  , splitWithNameTrace
+  -- * Tracers
+  , debugTracer
+  , sayTracer
+  -- * Tasty Utils
+  , nightlyTest
+  , ignoreTest
```
coot, line 32: The export list was added, to actually export `ignoreTest`.

coot:
bors merge

coot:
bors merge

kevinhammond (commented):
, (NodeToNodeV_10, CardanoNodeToNodeVersion6)

Would be worth commenting to explain the two identical eras.

Otherwise looks good

PR #3781, opened 2022-05-31
New function getOpCertCounters

newhoggy:
to get Operational Certificate Counters over any protocol

### DIFF
--- a/ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Common.hs
+++ b/ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Common.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE DerivingStrategies         #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeFamilies               #-}
 
 -- | Various things common to iterations of the Praos protocol.
 module Ouroboros.Consensus.Protocol.Praos.Common (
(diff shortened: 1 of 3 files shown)


nfrisby (approved):
Approved. Thanks for being proactive!

This does indeed seem like a continuation of the concern initially raised in Issue input-output-hk/cardano-node#3864, so this is an appropriate place to bundle this addition. We can continue to do so until we have a comprehensive list of the ways the node was relying on `TPraos` and then we'll be able to determine whether this interface is indeed the appropriate way to abstract over `TPraos` vs `Praos` vs etc. (See the discussion at input-output-hk/cardano-node#3864 for more.)

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/TPraos.hs
```diff
@@ -531,6 +533,18 @@ instance SL.PraosCrypto c => PraosProtocolSupportsNode (TPraos c) where
         , ticknStatePrevHashNonce
         } = csTickn
 
+  getOpCertCounters _prx cdst = opcertCounters
+    where
+      TPraosState { tpraosStateChainDepState } = cdst
+      SL.ChainDepState {
+        SL.csProtocol
+        } = tpraosStateChainDepState
+      SL.PrtclState
```
nfrisby, line 542: I'd generally prefer record syntax for this pattern, but the field we want has a unique type, so no real risk here.

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -108,4 +111,8 @@ data PraosNonces = PraosNonces {
 -- adding to it here under time pressure. See
 -- <https://github.com/input-output-hk/cardano-node/issues/3864>
 class ConsensusProtocol p => PraosProtocolSupportsNode p where
+  type PraosProtocolSupportsNodeCrypto p
```
nfrisby, line 114: I had hopes that some `p -> c` mapping already existed, but it does not. So I think it's reasonable to include this here, and I also agree it is wise to name is so specifically to this `PraosProtocolSupportsNode` use case, which we haven't given much forethought yet. 👍

PR #3782, opened 2022-05-31
cardano: bump latestReleasedNodeVersion to include Babbage without having to enable testing features

nfrisby:
While reviewing Marcin's latest PR #3775, we noticed the latest released version was never updated to include Babbage.

This PR updates it, so that Babbage is on by default. The test flag will then only have the effect of enabling P2P.


coot (approved):
LGTM

dcoutts (approved):

nfrisby:
It failed due to a cached `ouroboros-network` test failure on Hydra. I told Hydra to restart that failure (after notifying #network about it on the Slack).

nfrisby:
bors retry

PR #3788, opened 2022-06-03
Make ouroboros-consensus-protocol-test a standalone package inside of a sub-library.

KtorZ:
Sub-libraries are not supported by stack (see https://github.com/commercialhaskell/stack/issues/5318) and provide here very little benefits other than being able to nest files under one directory. Plus, this is inconsistent with how current test libraries (e.g. `ouroboros-consensus-shelley-test`) are handled.

Feel free to reject this PR and close it saying "we don't want to support stack". I am simply opening to raise awareness about this problem and because I had to fork the repository to do this anyway.


hasufell:
@KtorZ there's a (drunk) PR attempt to fix this in stack, have you tried it?

https://github.com/commercialhaskell/stack/pull/5659

KtorZ:
@hasufell yup, saw that! Yet "drunk attempt" kind of drove me away and I feared that it may be more effort to get this working that actually forking the repository and removing the sub-libraries. I was happy to see that Hai did submit a PR and I guess he hit the same wall as I did with Plutus :shrug: ... I did also fork Plutus for the same reasons but removing the sub-libraries was a bit too much effort there. 

Perhaps it's better to pick up Hai's PR and get it implemented in stack after all.

dnadales:
Thank you @KtorZ :raised_hands: The change seems sensible. We'll look into the hydra error and review this.

amesgen (changes requested):
Thanks, seems nice to make life easier for Stack users :+1:

I think that adding ouroboros-consensus-protocol-test to cabal.project will fix the Hydra error, see https://github.com/input-output-hk/ouroboros-network/commit/2ced2ba41e9ab0e2d69e5f2370fb65a463b32397 (just cherry-pick/amend).

amesgen:

PR #3789, opened 2022-06-03
Hydra: don't run style checks on mingwW64

amesgen:
Previously, the `styles.check-nixfmt` check also ran when cross-compiling to mingwW64 (which is of course unnecessary). This failed, as some dependencies of nixfmt can't be cross-compiled, but the job was not marked as required, so it didn't prevent `hydra:required` from going green.

This check was never supposed to run on mingwW64, but it was not properly excluded due to what seems to be a copy-paste error of `stylePaths` in `noCrossBuild`. This PR fixes this by excluding `checks.styles.*` from being cross-compiled.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -92,10 +92,6 @@ let
       ([ "exes" ] ++ path)
       ([ "haskellPackages" (head path) "components" "exes" ] ++ (tail path))
     ];
-    stylePaths = path: [
-      ([ "checks" "styles" ] ++ path)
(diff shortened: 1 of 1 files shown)


Jasagredo:
I think this can be merged right away. It won't do any harm (`nixfmt` is still checked in Linux at least).

Jasagredo (approved):

amesgen:
bors merge

PR #3790, opened 2022-06-03
Hydra: disable creation of golden files if missing

Jasagredo:


amesgen (approved):
Great catch! I think this is the best way to fix this ATM, LGTM!

Ideally though, this should be set as the default in `Test.Util.Serialisation.Golden` using [`localOption`](https://hackage.haskell.org/package/tasty-1.4.2.3/docs/Test-Tasty.html#v:localOption) as it is evidently very easy to forget to adapt this Nix code. Sadly, `tasty-golden` does not expose [`NoCreateFile`](https://github.com/UnkindPartition/tasty-golden/blob/3282bce3dc422ba304b9a6c949ecae7f26cb9e95/Test/Tasty/Golden/Internal.hs#L44), but I will comment on https://github.com/UnkindPartition/tasty-golden/issues/50 to see whether people agree to expose it.

Jasagredo:
https://hydra.iohk.io/build/15377668/nixlog/1 Indeed this fails as expected :warning:

PR #3791, opened 2022-06-03
CAD-4378 Investigate LMDB Max_Map_Size

jorisdral:
### Background on `mapsize`

View the LMDB API documentation for information about the `mapsize` setting for LMDB: [`mdb_env_set_mapsize`](http://www.lmdb.tech/doc/group__mdb.html#gaa2506ec8dab3d969b0e609cd82e619e5).
Or, view the LMDB source code: [`mdb_env_set_mapsize`](https://git.openldap.org/openldap/openldap/-/blob/mdb.master/libraries/liblmdb/mdb.c#L4562).

### Summary of PR

At the moment, we have not determined an appropriate value for the `mapsize` of the LMDB backing store. We could set this to a very large value, such that we do not have to update the `mapsize` dynamically when the database exceeds it. We have to ensure, however, that this does not negatively impact performance, memory usage and storage usage.

This PR includes a small executable that opens an `LMDB` database with an input `mapsize` and performs a given sequence of write/read commands. For different combinations of `mapsize` settings and sequences of commands, we: i) profile performance, ii) profile memory usage, iii) check the size of the on-disk database. We compare the results of these analyses to get an understanding of the `mapsize` setting's impact on the three metrics we listed before.

**The tentative conclusion of the results is that the `mapsize` does not influence any of the 3 metrics (significantly).**

We propose to keep the small executable around in `ouroboros-consensus/docs/utxo-hd/mapsize-profiling`.

### Acceptance checklist

- [x] Come up with a recommendation for the `mapsize` that should be used on the mainnet.

### DIFF
--- a/ouroboros-consensus/docs/utxo-hd/mapsize-profiling/cabal.project
+++ b/ouroboros-consensus/docs/utxo-hd/mapsize-profiling/cabal.project
@@ -1,0 +1,17 @@
+index-state: 2022-02-18T00:00:00Z
+
+profiling: True
+
+packages: .
+
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/haskell-lmdb
+  tag: d9ba8e382161aec999de12ae46e98c0eb83096b9
+  --sha256: 173dbqpn9czyjib6lnqa8f7ai66n3r4k44h5hvyl321ljychn5fm
+
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/lmdb-simple
+  tag: 5b2c622c1cf43deca081139b2d1d1eb9fc991064
+  --sha256: 0n96rzj4901y0q2l1fb6ffjp1kvgqchliig9id6jaq05xn4v386s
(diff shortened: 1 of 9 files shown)


coot (commented):
@jorisdral this seems to be a large PR.  Could you split it in a smaller logical patches. I only care that each patch should compile the component that it modifies (e.g. `ouroboros-network:lib:ouroboros-network`, and then update `ouroboros-network:test` in a following patch if that would make it too large).

Splitting a large patch into smaller one usually takes some time and effort (sometimes is non-trivial), but it pays back in quicker and more thorough reviews.

This [PR](https://github.com/input-output-hk/ouroboros-network/pull/3680/commits) might be a good example.

jorisdral:
> @jorisdral this seems to be a large PR. Could you split it in a smaller logical patches. I only care that each patch should compile the component that it modifies (e.g. `ouroboros-network:lib:ouroboros-network`, and then update `ouroboros-network:test` in a following patch if that would make it too large).
> 
> Splitting a large patch into smaller one usually takes some time and effort (sometimes is non-trivial), but it pays back in quicker and more thorough reviews.
> 
> This [PR](https://github.com/input-output-hk/ouroboros-network/pull/3680/commits) might be a good example.

Hi @coot, thanks for the feedback! I'm closing the PR, since it has been superseded/has become outdated. Concerning the large diff: I think that a recent rebase of `feature/utxo-hd` on `master` is the culprit. Possibly, this has added reviewers other than @dnadales and @Jasagredo to the PR as well, since I have experienced that in the past when rebasing the target branch of a PR.

I will clean up properly next time!

PR #3795, opened 2022-06-07
CAD-3915 config flags in node

Jasagredo:
- `DISABLE_DUAL_LEDGER` removed.
- `RunAlsoLegacy` defaulted to `RunOnlyNew` always.
- New `BackingStoreSelectorFlag` datatype to be able to parse them without the `m` monad.


Jasagredo:
> LGTM - - Though I wonder if the `BackingStoreSelectorFlag` should live in the node instead, since it is not being used by consensus

I moved it to `cardano-node`, Thanks!

PR #3798, opened 2022-06-08
Improvements to db-analyser

nfrisby:
This PR improves `db-analyser` (a tool used by devs) in the following ways:

- The `--checkThunks` analysis no longer trivial fails
- The `cardano` command now expects a single config argument: `--config`. The parsed fields are a subset of the schema that `exe:cardano-node` expects from its `config.json` file. This replaces the handful of config arguments the `db-analyser cardano` command used to require.
- The `cardano` command also now respects the `Test*HardForkAtEpoch` fields in the identified `--config` file; this is new functionality and necessary for running `db-analyser` on the chains resulting Benchmarking Team's workflows that "skip" over Byron etc.
- Added a new analysis `--repro-mempool-and-forge`, which maintains a mempool from the transactions in each block of the chain we're analyzing. It then makes the same two calls that we have found dominate the "forge time" metric in Benchmarking runs and reports their duration according to `getMonotonicTime`. This was developed for Jira ticket CAD-4214, but Ledger simultaneously fixed the high forge times we have seen. We're merging it now in case it's useful later.


nfrisby:
The commits are well-organized and have good commit messages; please see those to guide your review (eg why I removed `Alonzo.hs`). Thanks!

amesgen (approved):
Reviewed on a call, looks great, LGTM!

Issue #3159, opened 2021-05-17
Enable block production dynamically


coot:
For redunancy reasons SPOs are running backup block production nodes.
Currently, with a firewall rules, they can prevent relays to connect to the
backup node, and thus prevent the duplicate blocks to be diffused.   This will
no longer work with p2p nodes, as relays will be able to reuse inbound
connection from the block producer.  For this reason we would like a block
producing node to be able to start without being able to produce blocks and
start when it receives a SIGHUP signal.

This feature should be implemented against `p2p-master` branches (both in
`ouroboros-network` and `cardano-node`).

nfrisby:
This sounds reasonable. Though the choice of `SIGHUP` seems a bit odd to me; maybe a `USR*` instead? (Disclaimer: I'm unaware of any conventions here, just "hangup" sounds wrong here.)

Also, is sending it a signal _that_ much more convenient than restarting the backup non-BP node with a slightly altered config file/argument that enables BP? IE is it sufficiently more convenient to justify the additional complexity in the implementation? Signals are usually a pain. (Slippery slope though: will we also want to be able to similarly stop the block producer, eg once the primary one is back online?)

AndrewWestberg:
Signals are very simple to send. SIGHUP is used in many other processes to mean "reload your config". For example, prometheus does this.

```
kill -SIGHUP <pid>
```

karknu:
As Andrew mentions it is common for unix daemon's to support reloading their config files upon receiving `SIGHUP`. For p2p we've added support for reloading the topology file. What we're after here is a reloading a config file, not simply switching between block production enabled/disabled upon receiving the signal.

nfrisby:
Thanks @AndrewWestberg @karknu for elaborating.

> What we're after here is a reloading a config file, not simply switching between block production enabled/disabled upon receiving the signal.

@karknu @coot Would you update this Issue title and description according to this clarification ^^^? That's quite a bit more general than the current wording, if I understand correctly. Thanks.

-----

[...]

PR #3800, opened 2022-06-08
Control block forging through NodeKernel

coot:
Address #3159 on the `ouroboros-consensus` side.

- Fixed a typo
- Added setBlockForging to NodeKernel
- Updated ouroboros-consensus-test
- Updated ouroboros-consensus-mock
- Updated ouroboros-consensus-mock-test
- Updated ouroboros-consensus-byron
- Updated ouroboros-consensus-byron-test
- Updated ouroboros-consensus-shelley
- Updated ouroboros-consensus-shelley-test
- Updated ouroboros-consensus-cardano library
- Updated ouroboros-consensus-cardano:db-analyzer
- Updated ouroboros-consensus-cardano-test

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus-byron-test/test/Test/ThreadNet/Byron.hs
+++ b/ouroboros-consensus-byron-test/test/Test/ThreadNet/Byron.hs
@@ -22,7 +22,7 @@ import qualified Data.ByteString as BS
 import           Data.Coerce (coerce)
 import           Data.Functor ((<&>))
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (listToMaybe, mapMaybe)
+import           Data.Maybe (mapMaybe)
 import qualified Data.Set as Set
 import           Data.Word (Word64)
 
@@ -1290,27 +1290,27 @@ mkRekeyUpd
   => Genesis.Config
   -> Genesis.GeneratedSecrets
   -> CoreNodeId
-  -> ProtocolInfo m ByronBlock
+  -> ProtocolInfo ByronBlock
+  -> m [BlockForging m ByronBlock]
   -> EpochNo
   -> Crypto.SignKeyDSIGN Crypto.ByronDSIGN
   -> m (Maybe (TestNodeInitialization m ByronBlock))
-mkRekeyUpd genesisConfig genesisSecrets cid pInfo eno newSK =
-    pInfoBlockForging pInfo <&> \blockForging ->
-      case listToMaybe blockForging of
-        Nothing -> Nothing
-        Just _  ->
-          let genSK = genesisSecretFor genesisConfig genesisSecrets cid
-              creds' = updSignKey genSK bcfg cid (coerce eno) newSK
-              blockForging' =
-                byronBlockForging
-                  (TxLimits.mkOverrides TxLimits.noOverridesMeasure)
-                  creds'
-              pInfo' = pInfo { pInfoBlockForging = return [blockForging'] }
+mkRekeyUpd genesisConfig genesisSecrets cid pInfo blockForging eno newSK = do
+  blockForging <&> \case
+    []    -> Nothing
+    (_:_) ->
+      let genSK = genesisSecretFor genesisConfig genesisSecrets cid
+          creds' = updSignKey genSK bcfg cid (coerce eno) newSK
+          blockForging' =
+            byronBlockForging
+              (TxLimits.mkOverrides TxLimits.noOverridesMeasure)
+              creds'
 
-          in Just TestNodeInitialization
-            { tniCrucialTxs = [dlgTx (blcDlgCert creds')]
-            , tniProtocolInfo = pInfo'
-            }
+      in Just TestNodeInitialization
+        { tniCrucialTxs = [dlgTx (blcDlgCert creds')]
+        , tniProtocolInfo = pInfo
+        , tniBlockForging = pure [blockForging']
+        }
   where
     bcfg = configBlock (pInfoConfig pInfo)
 
--- a/ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
+++ b/ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
@@ -22,6 +22,7 @@ import           Test.Tasty.QuickCheck
 
 import           Cardano.Slotting.Slot (EpochSize (..), SlotNo (..))
 
+import           Ouroboros.Consensus.Block.Forging (BlockForging)
 import           Ouroboros.Consensus.BlockchainTime
 import           Ouroboros.Consensus.Config.SecurityParam
 import           Ouroboros.Consensus.Ledger.SupportsMempool (extractTxs)
@@ -473,7 +474,8 @@ mkProtocolCardanoAndHardForkTxs
     protocolParamsByronShelley =
     TestNodeInitialization
       { tniCrucialTxs   = crucialTxs
-      , tniProtocolInfo = pInfo
+      , tniProtocolInfo = protocolInfo
+      , tniBlockForging = blockForging
       }
   where
     crucialTxs :: [GenTx (CardanoBlock c)]
@@ -491,8 +493,24 @@ mkProtocolCardanoAndHardForkTxs
               generatedSecretsByron
               propPV
 
-    pInfo :: ProtocolInfo m (CardanoBlock c)
-    pInfo = protocolInfoCardano
+    protocolInfo :: ProtocolInfo (CardanoBlock c)
+    blockForging :: m [BlockForging m (CardanoBlock c)]
+    (protocolInfo, blockForging) = protocolInfoCardano
+              paramsByron
+              paramsShelleyBased
+              paramsShelley
+              paramsAllegra
+              paramsMary
+              paramsAlonzo
+              paramsBabbage
+              transitionShelley
+              transitionAllegra
+              transitionMary
+              transitionAlonzo
+              transitionBabbage
+
+    paramsByron :: ProtocolParamsByron
+    paramsByron =
         ProtocolParamsByron {
             byronGenesis                = genesisByron
             -- Trivialize the PBFT signature window so that the forks induced by
@@ -503,47 +521,79 @@ mkProtocolCardanoAndHardForkTxs
           , byronLeaderCredentials      = Just leaderCredentialsByron
           , byronMaxTxCapacityOverrides = TxLimits.mkOverrides TxLimits.noOverridesMeasure
           }
+
+    paramsShelleyBased :: ProtocolParamsShelleyBased (ShelleyEra c)
+    paramsShelleyBased =
         ProtocolParamsShelleyBased {
             shelleyBasedGenesis           = genesisShelley
           , shelleyBasedInitialNonce      = initialNonce
           , shelleyBasedLeaderCredentials = [leaderCredentialsShelley]
           }
+
+    paramsShelley :: ProtocolParamsShelley c
+    paramsShelley =
         ProtocolParamsShelley {
             shelleyProtVer                = SL.ProtVer shelleyMajorVersion 0
           , shelleyMaxTxCapacityOverrides = TxLimits.mkOverrides TxLimits.noOverridesMeasure
           }
+
+    paramsAllegra :: ProtocolParamsAllegra c
+    paramsAllegra =
         ProtocolParamsAllegra {
             allegraProtVer                = SL.ProtVer allegraMajorVersion 0
           , allegraMaxTxCapacityOverrides = TxLimits.mkOverrides TxLimits.noOverridesMeasure
           }
+
+    paramsMary :: ProtocolParamsMary c
+    paramsMary =
         ProtocolParamsMary {
             maryProtVer                   = SL.ProtVer maryMajorVersion    0
           , maryMaxTxCapacityOverrides    = TxLimits.mkOverrides TxLimits.noOverridesMeasure
           }
+
+    paramsAlonzo :: ProtocolParamsAlonzo c
+    paramsAlonzo =
         ProtocolParamsAlonzo {
             alonzoProtVer                 = SL.ProtVer alonzoMajorVersion  0
           , alonzoMaxTxCapacityOverrides  = TxLimits.mkOverrides TxLimits.noOverridesMeasure
           }
+
+    paramsBabbage :: ProtocolParamsBabbage c
+    paramsBabbage =
         ProtocolParamsBabbage {
             babbageProtVer                = SL.ProtVer babbageMajorVersion  0
           , babbageMaxTxCapacityOverrides  = TxLimits.mkOverrides TxLimits.noOverridesMeasure
           }
-        protocolParamsByronShelley
+
+    transitionShelley :: ProtocolTransitionParamsShelleyBased (ShelleyEra c)
+    transitionShelley = protocolParamsByronShelley
+
+    transitionAllegra :: ProtocolTransitionParamsShelleyBased (AllegraEra c)
+    transitionAllegra =
         ProtocolTransitionParamsShelleyBased {
             transitionTranslationContext = ()
           , transitionTrigger            =
               TriggerHardForkAtVersion allegraMajorVersion
           }
+
+    transitionMary :: ProtocolTransitionParamsShelleyBased (MaryEra c)
+    transitionMary =
         ProtocolTransitionParamsShelleyBased {
             transitionTranslationContext = ()
           , transitionTrigger            =
               TriggerHardForkAtVersion maryMajorVersion
           }
+
+    transitionAlonzo :: ProtocolTransitionParamsShelleyBased (AlonzoEra c)
+    transitionAlonzo =
         ProtocolTransitionParamsShelleyBased {
             transitionTranslationContext = Alonzo.degenerateAlonzoGenesis
           , transitionTrigger            =
               TriggerHardForkAtVersion alonzoMajorVersion
           }
+
+    transitionBabbage :: ProtocolTransitionParamsShelleyBased (BabbageEra c)
+    transitionBabbage =
         ProtocolTransitionParamsShelleyBased {
             transitionTranslationContext = Alonzo.degenerateAlonzoGenesis
           , transitionTrigger            =
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
@@ -514,7 +514,9 @@ protocolInfoCardano ::
   -> ProtocolTransitionParamsShelleyBased (MaryEra c)
   -> ProtocolTransitionParamsShelleyBased (AlonzoEra c)
   -> ProtocolTransitionParamsShelleyBased (BabbageEra c)
-  -> ProtocolInfo m (CardanoBlock c)
+  -> ( ProtocolInfo      (CardanoBlock c)
+     , m [BlockForging m (CardanoBlock c)]
+     )
 protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
                         byronGenesis                = genesisByron
                       , byronLeaderCredentials      = mCredsByron
@@ -569,12 +571,13 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
   , length credssShelleyBased > 1
   = error "Multiple Shelley-based credentials not allowed for mainnet"
   | otherwise
-  = assertWithMsg (validateGenesis genesisShelley) $
-    ProtocolInfo {
+  = assertWithMsg (validateGenesis genesisShelley)
+    ( ProtocolInfo {
         pInfoConfig       = cfg
       , pInfoInitLedger   = initExtLedgerStateCardano
-      , pInfoBlockForging = blockForging
       }
+    , blockForging
+    )
   where
     -- The major protocol version of the last era is the maximum major protocol
     -- version we support.
@@ -605,7 +608,7 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
           , topLevelConfigBlock    = blockConfigByron
           }
       , pInfoInitLedger = initExtLedgerStateByron
-      } = protocolInfoByron @m protocolParamsByron
+      } = protocolInfoByron protocolParamsByron
 
     partialConsensusConfigByron :: PartialConsensusConfig (BlockProtocol ByronBlock)
     partialConsensusConfigByron = consensusConfigByron
--- a/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
+++ b/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
@@ -35,7 +35,7 @@ import           Data.Functor.Identity (Identity)
 import           Data.List (sortOn)
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (fromMaybe)
+import           Data.Maybe (fromJust, fromMaybe)
 import           Data.Ord (Down (..))
 import           Data.Proxy
 import           Data.Sequence.Strict (StrictSeq)
@@ -379,7 +379,8 @@ run env@ChainDBEnv { varDB, .. } cmd =
     advanceAndAdd :: ChainDBState m blk -> SlotNo -> blk -> m (Point blk)
     advanceAndAdd ChainDBState { chainDB } newCurSlot blk = do
       atomically $ modifyTVar varCurSlot (max newCurSlot)
-      addBlock chainDB InvalidBlockPunishment.noPunishment blk
+      -- `blockProcessed` always returns 'Just'
+      fromJust <$> addBlock chainDB InvalidBlockPunishment.noPunishment blk
 
     wipeVolatileDB :: ChainDBState m blk -> m (Point blk)
     wipeVolatileDB st = do
--- a/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
+++ b/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
@@ -35,7 +35,7 @@ import           Data.Functor.Identity (Identity)
 import           Data.List (sortOn)
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
-import           Data.Maybe (fromJust, fromMaybe)
+import           Data.Maybe (fromMaybe)
 import           Data.Ord (Down (..))
 import           Data.Proxy
 import           Data.Sequence.Strict (StrictSeq)
@@ -380,7 +380,10 @@ run env@ChainDBEnv { varDB, .. } cmd =
     advanceAndAdd ChainDBState { chainDB } newCurSlot blk = do
       atomically $ modifyTVar varCurSlot (max newCurSlot)
       -- `blockProcessed` always returns 'Just'
-      fromJust <$> addBlock chainDB InvalidBlockPunishment.noPunishment blk
+      res <- addBlock chainDB InvalidBlockPunishment.noPunishment blk
+      return $ case res of
+        Nothing -> error "advanceAndAdd: block not added"
+        Just pt -> pt
 
     wipeVolatileDB :: ChainDBState m blk -> m (Point blk)
     wipeVolatileDB st = do
--- a/ouroboros-consensus/docs/interface-CHANGELOG.md
+++ b/ouroboros-consensus/docs/interface-CHANGELOG.md
@@ -56,6 +56,14 @@ may appear out of chronological order.
 The internals of each entry are organized similar to
 https://keepachangelog.com/en/1.1.0/, adapted to our plan explained above.
 
+## Circa 2022-11-07
+
+## Added
+
+- Added `setBlockForging` to `NodeKernel` which must be used to set / control
+  block forging of the consensus layer.
+- We removed the `pInfoBlockForging` record field from the `ProtocolInfo` type.
+- `ProtocolInfo` no longer needs the monad parameter, so we've removed that too.
 ## Circa 2022-09-30
 
 ### Fixed
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Embed/Binary.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Embed/Binary.hs
@@ -32,19 +32,23 @@ protocolInfoBinary ::
      forall m blk1 blk2.
      (CanHardFork '[blk1, blk2], Monad m)
      -- First era
-  => ProtocolInfo m blk1
+  => ProtocolInfo blk1
+  -> m [BlockForging m blk1]
   -> History.EraParams
   -> (ConsensusConfig (BlockProtocol blk1) -> PartialConsensusConfig (BlockProtocol blk1))
   -> (LedgerConfig blk1 -> PartialLedgerConfig blk1)
      -- Second era
-  -> ProtocolInfo m blk2
+  -> ProtocolInfo blk2
+  -> m [BlockForging m blk2]
   -> History.EraParams
   -> (ConsensusConfig (BlockProtocol blk2) -> PartialConsensusConfig (BlockProtocol blk2))
   -> (LedgerConfig blk2 -> PartialLedgerConfig blk2)
-  -> ProtocolInfo m (HardForkBlock '[blk1, blk2])
-protocolInfoBinary protocolInfo1 eraParams1 toPartialConsensusConfig1 toPartialLedgerConfig1
-                   protocolInfo2 eraParams2 toPartialConsensusConfig2 toPartialLedgerConfig2 =
-    ProtocolInfo {
+  -> ( ProtocolInfo (HardForkBlock '[blk1, blk2])
+     , m [BlockForging m (HardForkBlock '[blk1, blk2])]
+     )
+protocolInfoBinary protocolInfo1 blockForging1 eraParams1 toPartialConsensusConfig1 toPartialLedgerConfig1
+                   protocolInfo2 blockForging2 eraParams2 toPartialConsensusConfig2 toPartialLedgerConfig2 =
+    ( ProtocolInfo {
         pInfoConfig = TopLevelConfig {
             topLevelConfigProtocol = HardForkConsensusConfig {
                 hardForkConsensusConfigK      = k
@@ -86,9 +90,9 @@ protocolInfoBinary protocolInfo1 eraParams1 toPartialConsensusConfig1 toPartialL
                   WrapChainDepState $
                     headerStateChainDep initHeaderState1
           }
-      , pInfoBlockForging =
-          alignWith alignBlockForging <$> blockForging1 <*> blockForging2
       }
+    , alignWith alignBlockForging <$> blockForging1 <*> blockForging2
+    )
   where
     ProtocolInfo {
         pInfoConfig = TopLevelConfig {
@@ -102,7 +106,6 @@ protocolInfoBinary protocolInfo1 eraParams1 toPartialConsensusConfig1 toPartialL
             ledgerState = initLedgerState1
           , headerState = initHeaderState1
           }
-      , pInfoBlockForging = blockForging1
       } = protocolInfo1
 
     ProtocolInfo {
@@ -113,7 +116,6 @@ protocolInfoBinary protocolInfo1 eraParams1 toPartialConsensusConfig1 toPartialL
           , topLevelConfigCodec    = codecConfig2
           , topLevelConfigStorage  = storageConfig2
           }
-      , pInfoBlockForging = blockForging2
       } = protocolInfo2
 
     k1, k2, k :: SecurityParam
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Node/ProtocolInfo.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Node/ProtocolInfo.hs
@@ -35,10 +35,9 @@ enumCoreNodes (NumCoreNodes numNodes) =
 -------------------------------------------------------------------------------}
 
 -- | Data required to run the specified protocol.
-data ProtocolInfo m b = ProtocolInfo {
-        pInfoConfig       :: TopLevelConfig b
-      , pInfoInitLedger   :: ExtLedgerState b -- ^ At genesis
-      , pInfoBlockForging :: m [BlockForging m b]
+data ProtocolInfo b = ProtocolInfo {
+        pInfoConfig     :: TopLevelConfig b
+      , pInfoInitLedger :: ExtLedgerState b -- ^ At genesis
       }
 
 -- | Data required by clients of a node running the specified protocol.
--- a/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns          #-}
 {-# LANGUAGE ConstraintKinds       #-}
 {-# LANGUAGE DataKinds             #-}
 {-# LANGUAGE DerivingVia           #-}
@@ -30,15 +31,19 @@ import           Control.DeepSeq (force)
 import           Control.Monad
 import           Control.Monad.Except
 import           Data.Bifunctor (second)
+import           Data.Foldable (traverse_)
 import           Data.Hashable (Hashable)
 import           Data.List.NonEmpty (NonEmpty)
 import           Data.Map.Strict (Map)
 import           Data.Maybe (isJust, mapMaybe)
 import           Data.Proxy
 import qualified Data.Text as Text
+import           Data.Void
 import           System.Random (StdGen)
 
 import           Control.Tracer
+
+import qualified Control.Concurrent.Class.MonadSTM as LazySTM
 
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment,
                      AnchoredSeq (..))
@@ -107,6 +112,9 @@ data NodeKernel m remotePeer localPeer blk = NodeKernel {
 
       -- | The node's tracers
     , getTracers             :: Tracers m remotePeer localPeer blk
+
+      -- | Set block forging
+    , setBlockForging        :: [BlockForging m blk] -> m ()
     }
 
 -- | Arguments required when initializing a node
@@ -118,7 +126,6 @@ data NodeKernelArgs m remotePeer localPeer blk = NodeKernelArgs {
     , chainDB                 :: ChainDB m blk
     , initChainDB             :: StorageConfig blk -> InitChainDB m blk -> m ()
     , blockFetchSize          :: Header blk -> SizeInBytes
-    , blockForging            :: [BlockForging m blk]
     , mempoolCapacityOverride :: MempoolCapacityBytesOverride
     , miniProtocolParameters  :: MiniProtocolParameters
     , blockFetchConfiguration :: BlockFetchConfiguration
@@ -136,15 +143,19 @@ initNodeKernel
     => NodeKernelArgs m remotePeer localPeer blk
     -> m (NodeKernel m remotePeer localPeer blk)
 initNodeKernel args@NodeKernelArgs { registry, cfg, tracers
-                                   , blockForging, chainDB, initChainDB
+                                   , chainDB, initChainDB
                                    , blockFetchConfiguration
                                    } = do
 
+    -- using a lazy 'TVar', 'BlockForging' does not have a 'NoThunks' instance.
+    blockForgingVar :: LazySTM.TMVar m [BlockForging m blk] <- LazySTM.newTMVarIO []
     initChainDB (configStorage cfg) (InitChainDB.fromFull chainDB)
 
     st <- initInternalState args
 
-    mapM_ (forkBlockForging st) blockForging
+    void $ forkLinkedThread registry "NodeKernel.blockForging" $
+                            blockForgingController st (LazySTM.takeTMVar blockForgingVar)
+
 
     let IS { blockFetchInterface, fetchClientRegistry, varCandidates,
              mempool } = st
@@ -167,7 +178,20 @@ initNodeKernel args@NodeKernelArgs { registry, cfg, tracers
       , getFetchMode           = readFetchMode blockFetchInterface
       , getNodeCandidates      = varCandidates
       , getTracers             = tracers
+      , setBlockForging        = \a -> atomically . LazySTM.putTMVar blockForgingVar $! a
       }
+  where
+    blockForgingController :: InternalState m remotePeer localPeer blk
+                           -> STM m [BlockForging m blk]
+                           -> m Void
+    blockForgingController st getBlockForging = go []
+      where
+        go :: [Thread m Void] -> m Void
+        go !forgingThreads = do
+          blockForging <- atomically getBlockForging
+          traverse_ cancelThread forgingThreads
+          blockForging' <- traverse (forkBlockForging st) blockForging
+          go blockForging'
 
 {-------------------------------------------------------------------------------
   Internal node components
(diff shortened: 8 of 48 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Embed/Binary.hs
```diff
@@ -124,6 +123,17 @@ protocolInfoBinary protocolInfo1 eraParams1 toPartialConsensusConfig1 toPartialL
     shape :: History.Shape '[blk1, blk2]
     shape = History.Shape $ exactlyTwo eraParams1 eraParams2
 
+
+blockForgingBinary :: forall blk1 blk2 m.
```
nfrisby, line 127: Minor: style is to indent the whole type, as in `protocolInfoBinary` above.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Embed/Binary.hs
```diff
@@ -124,6 +123,17 @@ protocolInfoBinary protocolInfo1 eraParams1 toPartialConsensusConfig1 toPartialL
...
+                      ( Monad m
+                      , CanHardFork '[blk1, blk2]
+                      )
+                   => m [BlockForging m blk1]
+                   -> m [BlockForging m blk2]
+                   -> m [BlockForging m (HardForkBlock '[blk1, blk2])]
+blockForgingBinary blockForging1 blockForging2 =
+          alignWith alignBlockForging <$> blockForging1 <*> blockForging2
```
nfrisby, line 135: Minor: style is just a 4 space indent

ouroboros-consensus/src/Ouroboros/Consensus/Node/ProtocolInfo.hs
```diff
@@ -35,10 +35,9 @@ enumCoreNodes (NumCoreNodes numNodes) =
 -------------------------------------------------------------------------------}
 
 -- | Data required to run the specified protocol.
-data ProtocolInfo m b = ProtocolInfo {
+data ProtocolInfo b = ProtocolInfo {
```
nfrisby, line 38: This type is a prominent part of our interface, and its used in much downstream code/repos.

1) Removing a now-dead parameter is a nice simplification, but it'll be onerous to integrate downstream. Perhaps we should open that PR (once this PR gets far enough through the process).

2) This PR should explain the change in [`interfaceCHAN](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-consensus/docs/interface-CHANGELOG.md)

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -175,7 +186,20 @@ initNodeKernel args@NodeKernelArgs { registry, cfg, tracers
...
       , getTracers             = tracers
+      , setBlockForging        = \a -> atomically . LazySTM.writeTVar blockForgingVar $! a
       }
+  where
+    blockForgingController :: InternalState m remotePeer localPeer blk
+                           -> STM m [BlockForging m blk]
+                           -> m Void
+    blockForgingController st getBlockForging = go []
+      where
+        go :: [Thread m Void] -> m Void
```
nfrisby, line 197: As written, this is a busy loop, no?

I think this may be a good use case for https://github.com/input-output-hk/ouroboros-network/blob/327a90918caabaa3d5fadb54de2701ca4fc7a83f/ouroboros-consensus/src/Ouroboros/Consensus/Util/STM.hs#L158

Or perhaps this should be a `TMVar`? And thus (at most) one cycling of the threads per `setBlockForging`

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -175,7 +186,20 @@ initNodeKernel args@NodeKernelArgs { registry, cfg, tracers
...
+    blockForgingController :: InternalState m remotePeer localPeer blk
+                           -> STM m [BlockForging m blk]
+                           -> m Void
+    blockForgingController st getBlockForging = go []
+      where
+        go :: [Thread m Void] -> m Void 
+        go !forgingThreads = do
+          blockForging <- atomically getBlockForging
+          traverse_ cancelThread forgingThreads
+          blockForging' <- traverse (forkBlockForging st) blockForging
```
nfrisby, line 201: I think it's good that we wait until all threads are dead before we spawn any new threads. 👍

ouroboros-consensus/docs/interface-CHANGELOG.md
```diff
@@ -56,6 +56,13 @@ may appear out of chronological order.
 The internals of each entry are organized similar to
 https://keepachangelog.com/en/1.1.0/, adapted to our plan explained above.
 
+## Circa 2022-06-08
+
+## Added
+
+- Added `setBlockForgnig` to `NodeKernel` which must be used to set / control
```
nfrisby, line 63: typo: `setBlockForgnig` -> `setBlockForging`

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
coot: Fixed.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Embed/Binary.hs
coot: Fixed

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Embed/Binary.hs
coot: I think you use the same style as we do, 6 when there is a where clause indented with 4 spaces.

ouroboros-consensus/src/Ouroboros/Consensus/Node/ProtocolInfo.hs
coot: That's my plan, still we need to use the new interface in `cardano-node` to reload block forging.

ouroboros-consensus/docs/interface-CHANGELOG.md
coot: Fixed.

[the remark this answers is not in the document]
coot: Fixed.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
coot: I haven't used `Watcher` because it seemed to me to be designed for a more complex usecase.  We don't need a finger print, and we just need to block on the `TMVar` until it changes.  However, in the future, if the `TMVar` holds configuration from which `BlockForging` record is constructed, then a `Watcher` will actually make sense.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Embed/Binary.hs
nfrisby: We use 4 above and below `where`. Thanks!

image

ouroboros-consensus/src/Ouroboros/Consensus/Node/ProtocolInfo.hs
nfrisby: Yeah. The only alternative I've imagined (and which I'm not excited about) would be to leave the `m` on `ProtocolInfo`, even unused. Just to avoid having to update the type signatures.

Especially if we're the ones making the downstream PRs, then the confusion/etc around that is probably not worth it. But if others are coping with the change, then it might save them some work. Just pointing out the option.

ouroboros-consensus/docs/interface-CHANGELOG.md
```diff
@@ -56,6 +56,15 @@ may appear out of chronological order.
 The internals of each entry are organized similar to
 https://keepachangelog.com/en/1.1.0/, adapted to our plan explained above.
 
+## Circa 2022-06-08
+
+## Added
+
+- Added `setBlockForging` to `NodeKernel` which must be used to set / control
+  block forging of the consensus layer.
+- `ProtocolInfo` does not have the `pInfoBlockForging` record field, as a result
```
nfrisby, line 65: I'd replace

> `ProtocolInfo` does not have the `pInfoBlockForging` record field

with

> We removed the `pInfoBlockForging` record field from the `ProtocolInfo` type

and I'd replace 

> it is no longer polymorphic in the underlying monad.

with

> `ProtocolInfo` no longer needs the monad parameter, so we've removed that too.

ouroboros-consensus/src/Ouroboros/Consensus/Node/ProtocolInfo.hs
coot: That's an easy change, in `cardano-node` there were just [one place](https://github.com/input-output-hk/cardano-node/pull/4024/commits/829fbc6761e31985e1c54df2a78af53a311b5322) to update, it's probably not that difficult in `cardano-wallet` and `cardano-db-sync`.

A draft pr in the node: https://github.com/input-output-hk/cardano-node/pull/4024

ouroboros-consensus/docs/interface-CHANGELOG.md
coot: agreed.

ouroboros-consensus/src/Ouroboros/Consensus/Node/ProtocolInfo.hs
nfrisby: I had overlooked the fact that this whole workstream required a Node PR. And as a bonus: the necessary changes in Node were indeed simple, for this signature change.

I hope it's that simple in other downstream code!

ouroboros-consensus-byron-test/test/Test/ThreadNet/Byron.hs
```diff
@@ -1290,26 +1289,26 @@ mkRekeyUpd
...
+  -> m [BlockForging m ByronBlock]
   -> EpochNo
   -> Crypto.SignKeyDSIGN Crypto.ByronDSIGN
   -> m (Maybe (TestNodeInitialization m ByronBlock))
-mkRekeyUpd genesisConfig genesisSecrets cid pInfo eno newSK =
-    pInfoBlockForging pInfo <&> \blockForging ->
-      case listToMaybe blockForging of
-        Nothing -> Nothing
+mkRekeyUpd genesisConfig genesisSecrets cid pInfo blockForging eno newSK = do
+  listToMaybe <$> blockForging >>= \case
```
nfrisby, line 1298: Minor: you could leave it as `<&>`, I think.

ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
```diff
@@ -488,8 +490,39 @@ mkProtocolCardanoAndHardForkTxs
               generatedSecretsByron
               propPV
 
-    pInfo :: ProtocolInfo m (CardanoBlock c)
+    pInfo :: ProtocolInfo (CardanoBlock c)
     pInfo = protocolInfoCardano
+              paramsByron
```
nfrisby, line 495: Now that I've seen these parameter duplication happen in so many places, I'm wondering if introducing a

```
foo :: ... -> (ProtocolInfo blk, m [BlockForging m blk]
```

scheme and defining/using that throughout would be a worthwhile investment to cut down on repetition at the various use sites.

The `protocolInfo*` functions used to serve this "all of the config" purpose. But this PR splits that into a slimmer `protocolInfo*` and the new `blockForging*` alongside it. So the proposal here is to re-introduce a `protocolInfoAndBlockForging*` so that we again have an "all of the config" combinator, which ought to save some noise at the numerous sites where we do in fact need both.

nfrisby: This pattern would also reduce the duplication I'm seeing in the new definitions of `protoclInfo*` and `blockForging*` functions: you're having to project out the same config parameters, etc, in each definition. Instead, you could essentially rename the existing `protocolInfo*` to `protocolInfoAndBlockForging*`, and munge its return type and body top-level expression from `ProtocolInfo{x,y,z} to `(y, ProtocolInfo{x,z})` and then define the new `protoclInfo*` and `blockForging*` as easy projections of `protocolInfoAndBlockForging*`.

nfrisby: Another possible benefit: would the various calls to `plainTestNodeInitialization` be easier to update (just the renaming `protocolInfo*` to `protocolInfoAndBlockForging*`) if we changed its single argument type to match the return type of `protocolInfoAndBlockForging*`?

nfrisby (dismissed):
Thank you so much for all the heavy-lifting you've done here in the Consensus code! Your diff is really clean, and the effort that took is much appreciated 🙏.

I am Requesting Changes only because of the amount of copy-paste code I'm seeing; lots of "parameter projection" code was duplicated between the new `protocolInfo*` and `blockForging*` functions, which have been split apart from the old `protocolInfo*` functions (whose single definition was able to share the parameter projections about the two halves that are separated in the new code). See the bigger Conversation about it below.

ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
coot: That would've been simpler, but I thought that the next step should be to pass the configuration which allows to construct `BlockForging` not `BlockForging` itself - `BlockForging`.  Isn't `BlockForging` in `cardano-node` an abstraction leak?

coot: When we pass configuration then it's possible to block the listener until it actually receives something different than the current configuration (although that's right now is not that easy, because many types are missing `Eq` instances).

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
coot: How this will work if the forging thread is creating a block, would that block be created or would it be discarded?

coot:
I am getting some test failures:

`ouroboros-consensus-byron-test`
--------------------------------

```
byron
  Byron
    simple convergence: FAIL (295.37s)
      *** Failed! Falsified (after 4 tests):
      TestSetup {setupEBBs = NoEBBs, setupK = SecurityParam 1, setupTestConfig = TestConfig {initSeed = Seed 4918591009071197805, nodeTopology = NodeTopology (fromList [(CoreNodeId 0,fromList []),(CoreNodeId 1,fromList [CoreNodeId 0]),(CoreNodeId 2,fromList [CoreNodeId 0,CoreNodeId 1]),(CoreNodeId 3,fromList [CoreNodeId 1])]), numCoreNodes = NumCoreNodes 4, numSlots = NumSlots 67}, setupNodeJoinPlan = NodeJoinPlan (fromList [(CoreNodeId 0,SlotNo 1),(CoreNodeId 1,SlotNo 1),(CoreNodeId 2,SlotNo 1),(CoreNodeId 3,SlotNo 35)]), setupNodeRestarts = NodeRestarts (fromList [(SlotNo 21,fromList [(CoreNodeId 1,NodeRestart)]),(SlotNo 25,fromList [(CoreNodeId 1,NodeRestart)]),(SlotNo 28,fromList [(CoreNodeId 0,NodeRestart)]),(SlotNo 29,fromList [(CoreNodeId 1,NodeRestart)]),(SlotNo 33,fromList [(CoreNodeId 1,NodeRestart)]),(SlotNo 34,fromList [(CoreNodeId 2,NodeRestart)]),(SlotNo 35,fromList [(CoreNodeId 3,NodeRekey)]),(SlotNo 37,fromList [(CoreNodeId 1,NodeRestart)]),(SlotNo 44,fromList [(CoreNodeId 0,NodeRestart)]),(SlotNo 45,fromList [(CoreNodeId 1,NodeRestart)]),(SlotNo 55,fromList [(CoreNodeId 3,NodeRestart)]),(SlotNo 56,fromList [(CoreNodeId 0,NodeRestart)]),(SlotNo 62,fromList [(CoreNodeId 2,NodeRestart)]),(SlotNo 63,fromList [(Cor

... 
     
      consensus expected: True
      maxForkLength: 0
      There were unexpected CannotForges: fromList [(SlotNo 55,[PBftCannotForgeInvalidDelegation (KeyHash {unKeyHash = 911965b94fe206522fe8fb1683abf0bd39d9f092e9c9553ddbe35ae0})]),(SlotNo 59,[PBftCannotForgeInvalidDelegation (KeyHash {unKeyHash = 911965b94fe206522fe8fb1683abf0bd39d9f092e9c9553ddbe35ae0})]),(SlotNo 63,[PBftCannotForgeInvalidDelegation (KeyHash {unKeyHash = 911965b94fe206522fe8fb1683abf0bd39d9f092e9c9553ddbe35ae0}),PBftCannotForgeInvalidDelegation (KeyHash {unKeyHash = 911965b94fe206522fe8fb1683abf0bd39d9f092e9c9553ddbe35ae0})])]
      Use --quickcheck-replay=254433 to reproduce.
```
It's interesting that the in the slot `55` the node was not able to produce a block, and that's exactly the sl

The full [ouroboros-consensus-byron-test log](https://github.com/input-output-hk/ouroboros-network/files/8900064/byron.log).

`ouroboros-consensus-cardano-test`
----------------------------------

And also another one in `ouroboros-consensus-cardano-test`:
```
          Exception thrown while showing test case:
            Assertion failed
              assert, called at src/Cardano/Crypto/KES/Mock.hs:98:9 in cardano-crypto-class-2.0.0-49b4213a40d8c0265da39810113c14e195218a76babd0ce512365b752abc1e6e:Cardano.Crypto.
KES.Mock
              signKES, called at src/Cardano/Crypto/KES/Class.hs:355:40 in cardano-crypto-class-2.0.0-49b4213a40d8c0265da39810113c14e195218a76babd0ce512365b752abc1e6e:Cardano.Cry
pto.KES.Class
          
          Use --quickcheck-replay=900463 to reproduce.
          Use -p '/SerialiseDisk.roundtrip Header/' to rerun this test only.
```

The last one is because [this assertion](https://github.com/input-output-hk/cardano-base/blob/master/cardano-crypto-class/src/Cardano/Crypto/KES/Mock.hs#L98=) failure.

The full [ouroboros-consensus-cardano-test log](https://github.com/input-output-hk/ouroboros-network/files/8900071/cardano.log).

ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
nfrisby: Even if the pass in values affecting the `BlockForging` that change instead of passing in the `BlockForging` itself, we'll still need to have a function somewhere that constructs the `BlockForging` value, right? So I think the de-duplication opportunity would still apply there, wouldn't it?

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -873,14 +928,18 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
...
-          let maxKESEvo :: Word64
+          let praosMaxKESEvo = SL.sgMaxKESEvolutions genesisShelley
+
+              maxKESEvo :: Word64
               maxKESEvo = assert (tpraosMaxKESEvo == praosMaxKESEvo) praosMaxKESEvo
 
               startPeriod :: Absolute.KESPeriod
               startPeriod = Absolute.ocertKESPeriod $ praosCanBeLeaderOpCert canBeLeader
 
           HotKey.mkHotKey @m @c initSignKey startPeriod maxKESEvo
```
nfrisby, line 939: I'm guessing this is the link to the KES-related assertion failure. `mkHotKey` includes a method that calls `signedKES`.

https://github.com/input-output-hk/ouroboros-network/blob/a7f023edadbccf37d980e061c29fb49079bf439e/ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Ledger/HotKey.hs#L190

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -459,10 +483,9 @@ forkBlockForging
...
     -> BlockForging m blk
-    -> m ()
+    -> m (Thread m Void)
 forkBlockForging IS{..} blockForging =
-      void
-    $ forkLinkedWatcher registry threadLabel
+      forkLinkedWatcher registry threadLabel
     $ knownSlotWatcher btime
     $ withEarlyExit_ . go
   where
```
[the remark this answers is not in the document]
nfrisby: I need to re-read our docs regarding KES/hot keys.

ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
coot: Fixed: now the functions return a tuple `(ProtocolInfo blk, m [BlockForging m blk])`; the slightly annoying thing is that chaining a function `f` which produces such a tuple with `fst` leaves `GHC` puzzled how to instantiate `m`.

ouroboros-consensus-byron-test/test/Test/ThreadNet/Byron.hs
coot: Fixed.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
coot: so we need something like:

```haskell
 bracket (atomically $ ChainDB.addBlockAsync ...)
         (\_ -> return ())
         (\result -> do curTip <- atomically $ ChainDB.blockProcessed result
                        ...
                        lift $ removeTxs ...
                        
         )
```
For this to be safe in presence of async exceptions the finaliser must be either a non blocking call, or we need to use `uninterruptibleMask`.  I suspect `ChainDB.blockProcessed` is a blocking call, `removeTxs` if it has access to some `TMVar` could also be blocking.

`addBlockAsync` is quite likely non-blocking (is it?), `bracket` will mask exceptions when it is executed so an async exception will not hit when it's running.

nfrisby:
Ah ha! Hydra built green! 🙌

nfrisby:
I'm mentally gassed at this point in my day. I'll do a last pass tomorrow during/after our call.

ouroboros-consensus-cardano-test/test/Test/ThreadNet/Cardano.hs
nfrisby: >chaining a function `f` which produces such a tuple with `fst` leaves GHC puzzled how to instantiate `m`

Ah, yes, that is a bother

ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
```diff
@@ -243,9 +243,9 @@ mkCardanoProtocolInfo ::
   -> SL.AlonzoGenesis
   -> Nonce
   -> NP STA (CardanoShelleyEras StandardCrypto)
-  -> ProtocolInfo IO (CardanoBlock StandardCrypto)
+  -> ProtocolInfo (CardanoBlock StandardCrypto)
 mkCardanoProtocolInfo genesisByron signatureThreshold genesisShelley genesisAlonzo initialNonce hardForkTriggers =
-    protocolInfoCardano
+    fst $ protocolInfoCardano @_ @IO
```
nfrisby, line 248: Here's an example of where you have to specify `m` for GHC.

This is only necessary because in this case the arguments to `protocolInfoCardano` have no credentials and hence do not themselves fix `m`, right? In general, if you pass in credentials, then GHC will know the `m`.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: This is the last remaining issue -- otherwise the diff looks good to me.

coot: This is fixed, although not with a `bracket` but with `uninterruptibleMask_`.

coot: See the [commit](https://github.com/input-output-hk/ouroboros-network/pull/3800/commits/7356c6431c79fc1fa4c15273b9761e31086e2339).

ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
coot: Yes, that's right.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -642,40 +642,48 @@ forkBlockForging IS{..} blockForging =
...
-        -- e.g., @DualBlock@.
-        trace $ TraceAdoptedBlock currentSlot newBlock txs
+        -- 'addBlockAsync' is a non-blocking action, so `mask_` would suffice,
+        -- but the finalizer is a blocking operation, hence we need to use
+        -- 'uninterruptibleMask_' to make sure that async exceptions do not
+        -- interrupt it.
+        uninterruptibleMask_ $ do
+          result <- lift $ ChainDB.addBlockAsync chainDB noPunish newBlock
+          -- Block until we have processed the block
+          curTip <- lift $ atomically $ ChainDB.blockProcessed result
```
nfrisby, line 655: I think this block is assuming that `blockProcessed` is guaranteed to become unblocked.

But I'm not sure that's true. In particular, I doubt the corresponding ChainDB code is blocking async exceptions. So, suppose:

- We call `addBlockAsync`.
- The ChainDB shuts down (eg the users requests shutdown, eg a fatal error, etc) before unblocking `blockProcessed`.

That would leave this thread deadlocked, wouldn't it?

... maybe that's OK? This thread doesn't own any resources (eg that would need some logic to close gracefully). So this thread would only die when the main GHC thread ends (ie when the OS reaps the process).

I _think_ the ChainDB only has a single lifetime (ie it only shuts down if the node is shutting down), so `blockProcessed` will only deadlock if the node is shutting down---ie there's no risk we'll accumulate a bunch of zombie threads.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: That logic sounds good to me, with one potential caveat, which I spelled out in this other Thread: https://github.com/input-output-hk/ouroboros-network/pull/3800#discussion_r997642861 But, as I suggested there, I think it's OK as-is. Feels too subtle though :(

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: However, opting into this stuck thread behavior seems like asking for trouble in the future.

The alternative is to not mask async exns here. It seems acceptable to allow for the thread to die between adding the block and being notified that it was processed. The continuation only matters for i) completeness of log message and ii) removing the offending transactions from the mempool in the unlikely case that a bug has caused the node to forge a block that it then identifies as invalid. Unless the operator is rotating their block credentials with high frequency, those same txs would probably be included in the next block, and so would get caught by the continuation then.

So, the options and their downsides are:

- include the mask: we are opting-in to some possible unusual stuck-thread behavior when the ChainDB shutsdown while/immediately after the node forges a block

- exclude the mask: when the user signals a credentials change immediately after the node forges an invalid block (which would require additional bug somewhere in the forge logic), the node would miss two consecutive forging opportunities instead of just the first.

- ??? is there a Pareto-dominant option I didn't consider here?

cc @coot

nfrisby: Ah, nevermind. https://hackage.haskell.org/package/base-4.17.0.0/docs/Control-Exception.html#t:BlockedIndefinitelyOnSTM If we have the mask (option 1, as on your current diff), then I think the RTS will raise `BlockedIndefinitelyOnSTM` (*synch*ronously, _I think_?) in this situation if the ChainDB shutsdown while the thread is waiting on `blockProcessed`. It involves more moving parts than I'd like, but there all being used how they're intended.

coot: > synchronously, I think?

Does it matter how the exception is raised? I'd expect that in this case such case RTS sets the thread to throw an exception, and does not wait for it to be raised (like the `throwTo` does).

Anyway relaying on `BlockedIndefinitelyOnSTM` makes me shiver :wink:. 

But all this discussion points to a problem with the current approach.  The thread which actually adds block should inform the other end of the queue that it died, e.g. `blockProcessed` ought to return `Maybe (Point blk)`, if `Nothing` is returned the block producing thread would know that the thread that was adding a block died and it could terminate as well, without being blocked.

nfrisby: > Does it matter how the exception is raised? I'd expect that in this case such case RTS sets the thread to throw an exception, and does not wait for it to be raised (like the `throwTo` does).

Well, if the spec is that it would be raised _synchronously_ (eg it's considered as if the STM transaction itself decided to raise it), then I'd be confident that the masking state wouldn't the thread from receiving it while stuck in the STM transaction. That's all I was thinking.

-----

> But all this discussion points to a problem with the current approach.

Hmm, yes, I like that solution (Pareto-dominant! 👏).

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -425,28 +425,31 @@ forkBlockForging IS{..} blockForging =
...
-                --
-                -- Remove all the transactions in that block, otherwise we'll
-                -- run the risk of forging the same invalid block again. This
-                -- means that we'll throw away some good transactions in the
-                -- process.
-                lift $ removeTxs mempool (map (txId . txForgetValidated) txs)
-            exitEarly
+          mbCurTip <- lift $ atomically $ ChainDB.blockProcessed result
+          case mbCurTip of
+            Nothing -> trace (TraceDidntAdoptBlock currentSlot newBlock)
```
coot, line 430: @nfrisby should we have a separate trace, so it can be distinguished from the other places which can emit it.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -535,7 +535,11 @@ addBlockRunner
...
-      blockRealPoint $ blockToAdd blkToAdd
-    addBlockSync cdb blkToAdd
+    -- if the `addBlockSync` does not complete because it was killed by an async
+    -- exception (or it errored), notify the blocked thread)
+    bracketOnError (getBlockToAdd cdbBlocksToAdd)
+                   (\blkToAdd -> atomically $ tryPutTMVar (varBlockProcessed blkToAdd) Nothing)
+                   (\blkToAdd -> do
+                     trace $ PoppedBlockFromQueue $ FallingEdgeWith $
+                             blockRealPoint $ blockToAdd blkToAdd
+                     addBlockSync cdb blkToAdd)
```
coot, line 545: `bracketOnError` masks exceptions when `getBlockToAdd` blocks, this is ok (since `getBlockToAdd` blocks and thus is interruptible); once `getBlockToAdd` returns if there's an async exception that kills the thread it will guarantee that the `varBlockProcessed` `TMVar` is filled with `Nothing` (unless it was already filled by `addBlockSync`).

nfrisby: Hmm. Do we also need the shutdown logic of the `addBlockRunner` thread to flush the `cdbBlocksToAdd` queue, sending `Nothing` to every request it hadn't started processing before it begin terminating?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
```diff
@@ -437,7 +439,7 @@ addBlockWaitWrittenToDisk chainDB punish blk = do
 
 -- | Add a block synchronously: wait until the block has been processed (see
 -- 'blockProcessed'). The new tip of the ChainDB is returned.
-addBlock :: IOLike m => ChainDB m blk -> InvalidBlockPunishment m -> blk -> m (Point blk)
+addBlock :: IOLike m => ChainDB m blk -> InvalidBlockPunishment m -> blk -> m (Maybe (Point blk))
```
nfrisby, line 442: 1) `grep` hasn't yet found a use of this function that is obviously part of the implementation. I'm suspecting it's only for tests.

2) Therefore, I'm less worried about muddying its type with this partiality that should only happen when the whole node is shutting down. However, if we ever do want to remove the `Maybe` you're adding here, since its semantics is "return once the block is processed", I think it'd be fine to raise an exception instead of ever returning `Nothing`.

nfrisby: The definition `addBlock_ x y z = void $ addBlock x y z` below forces this question a little more:

- Should `addBlock_` return quietly when `addBlock` returned `Nothing`?

- Or should it instead raise that hypothetical exception I had proposed in the previous message?

(I'm also not definitely seeing `addBlock_` used in the actual implementation.)

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: Yeah, good idea. I could definitely envision that a more specific tracer event being in a future log could help someone confidently explain some otherwise confusing behavior.

nfrisby (changes requested):
Thanks! Nice and clean commits. I'm Requesting Changes mainly because of the question about flushing the blocks-to-add queue during shutdown.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
nfrisby: Great. Yes, looks good. I'm Resolving this Thread, since I just started some new Threads for the new code.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
coot: :100:

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
nfrisby: > do we ever restart the ChainDB? The queue is created when we start the ChainDB, if it's has the same life time as ChainDB then it doesn't matter, new ChainDB will anyway use new queue. But I suspect we never re-open ChainDB, is this right?

I'm pretty sure we never restart the `ChainDB`. (Maybe in tests?)

- Suppose the queue has two elements, `[A, B]`. (Recall that the block forging thread and the BlockFetch clients add to this queue concurrently.)
- Suppose `addBlockRunner` pops `B`, but the system goes down before it's done.
- If `A` was added by the block forging thread, then the code currently in the diff would never send `Nothing` to `A`'s promise; and so the block forging thread would be a zombie.

coot: Right, we indeed need to notify all the threads which posted a message in the queue.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
coot: `addBlock_` is used by [`initNodeKernel`](https://input-output-hk.github.io/ouroboros-network/ouroboros-consensus/Ouroboros-Consensus-NodeKernel.html#v:initNodeKernel) via [`fromFull` ](https://input-output-hk.github.io/ouroboros-network/ouroboros-consensus/Ouroboros-Consensus-Storage-ChainDB-Init.html#v:fromFull).  So these function should not be partial.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -379,7 +379,8 @@ run env@ChainDBEnv { varDB, .. } cmd =
     advanceAndAdd :: ChainDBState m blk -> SlotNo -> blk -> m (Point blk)
     advanceAndAdd ChainDBState { chainDB } newCurSlot blk = do
       atomically $ modifyTVar varCurSlot (max newCurSlot)
-      addBlock chainDB InvalidBlockPunishment.noPunishment blk
+      -- `blockProcessed` always returns 'Just'
+      fromJust <$> addBlock chainDB InvalidBlockPunishment.noPunishment blk
```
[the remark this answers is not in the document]
coot: Closing this as the discussion continues in the comment near `addBlock`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
coot: Fixed.

nfrisby:
> But I suspect we never re-open ChainDB, is this right?

That is right. Some tests do it, but not the implementation.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
nfrisby: This message is just exhaustively checking for uses.

`ouroboros-consensus*` defines four `addBlock` functions:

- `addBlock`          380 `ouroboros-network/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/Model.hs` signature
- `addBlock`          440 `ouroboros-network/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs` signature
- `addBlock`           24 `ouroboros-network/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Init.hs` selector
- `addBlock`           63 `ouroboros-network/ouroboros-consensus/src/Ouroboros/Consensus/Storage/VolatileDB/Impl/FileInfo.hs` signature

The second one is our topic here and neighbors the one and only `addBlock_` wrapper.

The same name is also defined in some `ouroboros-network*` packages.

`git grep -w -e addBlock origin/master -- 'ouroboros*hs'` reveals these uses:

- `API.addBlock_` used in `API.fromChain`, which seems to be dead code (maybe downstream repo?)
- `API.addBlock_` used in `Init.fromFull`, see below.
- `API.addBlock_` used in `Test.Ouroboros.Storage.ChainDB.FollowerPromptness`
- `API.addBlock` used in `Test.Ouroboros.Storage.ChainDB.StateMachine`
- `API.addBlock` used in `Test.ThreadNet.Network`
- `API.addBlock` used to define `API.toChain`, which is only used in `Test.ThreadNet.Network`.

- `FileInfo.addBlock` used in `Ouroboros.Consensus.Storage.VolatileDB.Impl`

- `Init.addBlock` used in `nodeInitChainDB @ByronBlock`
- `Init.addBlock` used in `nodeInitChainDB @DualByronBlock`
- `Init.addBlock` composed in `nodeInitChainDB @HardForkBlock`

- `Model.addBlock` also used in `Test.Ouroboros.Storage.ChainDB.StateMachine`

- `Ouroboros.Network.MockChain.Chain.addBlock` used in `Test.Util.ChainUpdates`

- `Test.Ouroboros.Storage.ChainDB.Model{,.Test}` defines and uses its own `addBlock`

(There's a second `fromChain`, which is declared in `Ouroboros/Consensus/HeaderStateHistory.hs`, but it's unrelated and only used in `Test.Consensus.MiniProtocol.ChainSync.Client`.)

That confirms `Init.fromFull` as the only interesting one, as you had noted.

It's used to initialize the node in `NodeKernel.hs`; in this call:

https://github.com/input-output-hk/ouroboros-network/blob/72863b0fc78abdc2b8e29f0dda96c06da3dd11d0/ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs#L138-L143

`git grep` indicates that `initChainDB` is only defined as `nodeInitChainDB`. So the use-def chain looks like this: `initChainDB`, `nodeInitChainDB`, `Init.addBlock` (for Byron types), `Init.fromFull`, `API.addBlock_`, `API.addBlock`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -492,6 +494,14 @@ addBlockToAdd tracer (BlocksToAdd queue) punish blk = do
...
 getBlockToAdd (BlocksToAdd queue) = atomically $ readTBQueue queue
 
+-- | Flush the 'BlocksToAdd' queue and notify the waiting threads.
+--
+closeBlocksToAdd :: IOLike m => BlocksToAdd m blk -> STM m ()
+closeBlocksToAdd (BlocksToAdd queue) = do
+  as <- flushTBQueue queue
+  traverse_ (\a -> tryPutTMVar (varBlockProcessed a) Nothing) as
+
+
```
nfrisby, line 504: Consensus style: please remove the extra newline here.

ouroboros-consensus/src/Ouroboros/Consensus/NodeKernel.hs
```diff
@@ -392,40 +415,51 @@ forkBlockForging IS{..} blockForging =
...
+        -- but the finalizer is a blocking operation, hence we need to use
+        -- 'uninterruptibleMask_' to make sure that async exceptions do not
+        -- interrupt it.
+        uninterruptibleMask_ $ do
+          result <- lift $ ChainDB.addBlockAsync chainDB noPunish newBlock
+          -- Block until we have processed the block
+          mbCurTip <- lift $ atomically $ ChainDB.blockProcessed result
+          case mbCurTip of
+            Nothing -> trace (TraceAdoptionThreadDied currentSlot newBlock)
+                    >> exitEarly
```
nfrisby, line 431: Consensus style: please use `do` instead of `>>`.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Background.hs
```diff
@@ -535,7 +535,13 @@ addBlockRunner
 addBlockRunner cdb@CDB{..} = forever $ do
     let trace = traceWith cdbTracer . TraceAddBlockEvent
     trace $ PoppedBlockFromQueue RisingEdge
-    blkToAdd <- getBlockToAdd cdbBlocksToAdd
-    trace $ PoppedBlockFromQueue $ FallingEdgeWith $
-      blockRealPoint $ blockToAdd blkToAdd
-    addBlockSync cdb blkToAdd
+    -- if the `addBlockSync` does not complete because it was killed by an async
+    -- exception (or it errored), notify the blocked thread
```
nfrisby, line 539: Please update the comment to say `notify all the threads that are waiting for an element in the queue`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
nfrisby: Looking more closely at `nodeInitChainDB @ByronBlock` --- which is ultimately the only interesting transitive use of `API.addBlock` (see previous message) --- we see it's merely adding the slot 0 Epoch Boundary Block.

https://github.com/input-output-hk/ouroboros-network/blob/72863b0fc78abdc2b8e29f0dda96c06da3dd11d0/ouroboros-consensus-byron/src/Ouroboros/Consensus/Byron/Node.hs#L273-L282

There's no way to recover, if that fails.

---

So: the only real use of `addBlock` in the system has no expectation of failure and no useful way to recover if it did fail.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
```diff
@@ -492,6 +494,14 @@ addBlockToAdd tracer (BlocksToAdd queue) punish blk = do
 getBlockToAdd :: IOLike m => BlocksToAdd m blk -> m (BlockToAdd m blk)
 getBlockToAdd (BlocksToAdd queue) = atomically $ readTBQueue queue
 
+-- | Flush the 'BlocksToAdd' queue and notify the waiting threads.
+--
+closeBlocksToAdd :: IOLike m => BlocksToAdd m blk -> STM m ()
```
nfrisby, line 499: Bah; this doesn't seem enough. We have a race now, don't we?

- Some threads are adding tasks to the queue; these are BlockFetch clients and the `NodeKernel.hs` forge.
- One thread is popping one task from the queue at a time

If the popping thread is killed mid-pop, then it notifies the unlucky owner of the task that got interrupted. And it also flushes the queue, similarly notifying all other task owners. And then the popping thread will terminate, since `bracket*` re-raises the exception.

But there's no guarantee the popping thread will be the last to terminate. So even when it's gone, other threads may be adding to the queue.

The first option that comes to mind:

- ~Complicate the queue by adding state indicating whether it's "open" or "closed" and have the `addBlockRunner` close the queue when it flushes it.~ Hmm... the `ChainDB` already has an open/closed state; maybe the `addBlockRunner` dying is either reason enough to fully close the ChainDB or can only actually happen when the ChainDB is closed/closing or something like that?

- ~But now `addBlockAsync` will _also_ be partial, since you can't add to a closed queue.~ Now `addBlockAsync :: ... -> m (AddBlockPromise m blk)` would create a degenerate "promise" (immediately filled with `False` and `Nothing`) when asked to add a block to a closed queue.

nfrisby (changes requested):
Some style and a new problematic observation regarding concurrency :(

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
nfrisby: Because there is a failure case in the actual implementation, my preference is to keep the `Maybe` layer you've added here (but as a custom `Maybe`-isomorph with clearly named constructors), and to propagate it all the way to the problematic call-site.

EG we'd need to update at least `Init.addBlock` to return `Maybe ()` instead of `()`. Perhaps also doing that for `nodeInitChainDB`.

But I'm not certain about this. I think I tend to reach for explicit error types more than most. So maybe a `UnrecoverableFailureDuringNodeInit` would be a fine exception, and it could be wired into the definition of `Init.addBlock :: ... -> m ()`?

What do you think, @coot

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
bolt12: I didn't really grasp the problem here. Isn't the point of STM to be atomic, I don't think we'll get a Async exception mid-pop, but what can happen is the thread getting interrupted while blocked on waiting for popping. And if that's the case the cleanup handle won't even run.

bolt12 (commented):
@nfrisby I am taking over this PR and I had some comments, could you spare some time to reply so I feel confident in making the changes needed?

Notice that I also need to rebase (eek!) this very old PR, but I figure addressing the issues first will be better

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs
coot: @nfrisby is right, we need to make sure no new block can be accepted by the db after `closeBlocksToAdd` is called by `addBlockRunner`. @bolt12 it's not about the thread itself, it's about all the other concurrent writes to the `ChainDB` that are done by all block-fetch clients.

@nfrisby The exception will propagate and eventually `ChainDB.closeDB` will be called.  It seems to me that we don't have access to the `TVar` which holds `ChainDbState` in the context of `addBlockRunner`, so keeping the state of the queue might be an easier option to implement (as in the suggestion you stroked).

PR #3844, opened 2022-06-28
Inbound governor idle timeout

### LINKED ISSUE #3843: Change local_PROTOCOL_IDLE_TIMEOUT
The [`local_PROTOCOL_IDLE_TIMEOUT`](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs?plain=1#L258-L259=) should not be set for `node-to-client` protocol.

coot:
For local connections the inbound governor should not force the idleness
timeout.  Some clients relay letting the connection idle after it was
negotiated and before any mini-protocol is started.

We keep the connection manager idle timeout, this is only used for
outbound connections, and thus not related to `node-to-client` protocol.

Fixes #3843


coot:
bors merge

PR #3850, opened 2022-06-29
Refactor FirstToFinish and LastToFinish instances

coot:


monoidal-synchronisation/src/Data/Monoid/Synchronisation.hs
```diff
@@ -48,12 +56,23 @@ newtype FirstToFinish m a = FirstToFinish { runFirstToFinish :: m a }
...
                    , Traversable
                    )
-  deriving Semigroup     via (Alt m a)
-  deriving Monoid        via (Alt m a)
   deriving Foldable      via (Alt m)
   deriving Contravariant via (Alt m)
 
 
+instance Semigroup (FirstToFinish IO a) where
+    FirstToFinish a <> FirstToFinish b = FirstToFinish $ either id id <$> race a b
```
coot, line 64: We can use similar approach to `LastToFinish`.

network-mux/src/Network/Mux.hs
```diff
@@ -203,6 +204,8 @@ runMux :: forall m mode.
           , MonadTime  m
           , MonadTimer m
           , MonadMask m
+          , forall a stm. stm ~ STM m => Semigroup (FirstToFinish stm a)
```
coot, line 207: Use a type alias for this constraint if possible.

bolt12 (commented):
What's the difference between `LastToFinish` and `LastToFinishM`?

coot:
The difference is explained [here](https://ouroboros-network.cardano.intersectmbo.org/monoidal-synchronisation/Data-Monoid-Synchronisation.html#t:LastToFinish).

PR #3855, opened 2022-07-01
Network StyleGuide

coot:
# Description

The networking style guide based on `ouroboros-consensus/docs/StyleGuide.md`.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
    - [] Useful pull request description at least containing the following information:
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/docs/StyleGuide.md
+++ b/docs/StyleGuide.md
@@ -115,13 +115,31 @@ the rules below, it is good practice to update the code's style to match them.
    c. We indent `data` definitions with multiple constructors as follows:
 
       ```haskell
-      data Foo =
-          Bar Int Int
+      data Foo
+        = Bar Int Int
         | Baz
             Int
             Int
             (Maybe Bool)
             [Foo]
+      ```
+
+      or
+
+      ```haskell
+      data Foo
+        = -- | Bar
+          --
+          Bar Int Int
+
+          -- | Baz
+          --
+        | Baz
+            Int
+            Int
+            (Maybe Bool)
+            [Foo]
+
       ```
 
       Note the argument of `Baz` being indented by two spaces.
@@ -616,8 +634,8 @@ the rules below, it is good practice to update the code's style to match them.
       :: Word -- ^ Max size
       -> ..
 
-    data Foo =
-        -- | Foo
+    data Foo
+      = -- | Foo
         --
         -- ..
         Foo
--- a/docs/StyleGuide.md
+++ b/docs/StyleGuide.md
@@ -209,7 +209,7 @@ the rules below, it is good practice to update the code's style to match them.
         A x -> A_body
         B y ->
           let bl = bl_body y
-          in  B_body
+          in B_body
       ```
 
       Note that we align `B_body` with `bl` in the `let` block. At the moment we
--- a/docs/StyleGuide.md
+++ b/docs/StyleGuide.md
@@ -1,0 +1,1142 @@
+# Network Style Guide
+
+This style guide is based on [consensus][consensus-style-guide].  Our guiding
+principles and enforcement through PR reviews are the same, so please read the
+relevant sections of that document, we just repeat here the most important
+topics.  However our formatting rules are slightly different.
+
+## Guiding principles
+
+We value the following principles in the networking team:
+
+* __Optimise for clarity__
+* __Consistency__: inconsistent style, especially within a single module,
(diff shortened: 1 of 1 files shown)


docs/StyleGuide.md
```diff
@@ -0,0 +1,1142 @@
...
+          }
+        deriving stock    (Show, Eq, Generic)
+        deriving anyclass (NoThunks, NFData)
+
+      newtype Foo = Foo {
+            unFoo :: Int
+          }
+        deriving stock   (Show)
+        deriving newtype (Eq)
+        deriving NoThunks via InspectHeapNamed "Foo" Foo
```
bolt12, line 82: No particular ordering on these right?

docs/StyleGuide.md
```diff
@@ -0,0 +1,1142 @@
...
+
+      ```haskell
+      data Foo =
+          Bar Int Int
+        | Baz
+            Int
+            Int
+            (Maybe Bool)
+            [Foo]
+      ```
```
bolt12, line 125: I would prefer

```haskell
data Foo = Bar Int Int
         | Baz Int
               Int
               (Maybe Bool)
               [Foo]
```

If space allows, I just find it more readable with the bar aligned with the equals sign and the first argument in the same line as the constructor but not always that is possible due to space constraints

docs/StyleGuide.md
```diff
@@ -0,0 +1,1142 @@
...
+    ```haskell
+    -- | Short title
+    --
+    -- Longer description .. 'Foo' .. "Data.String" .. @a@ .. /not/ ..
+    -- __never__ .. called \"foo bars\" .. alternative style " foo bars "
+    -- .. @'Foo' a@
+    --
+    -- > foo bar baz
+    --
+    -- .. ends here.
```
bolt12, line 600: Don't we usually have an empty comment line between the haddock/description and the function type signature like this:

```haskell
-- | Description ....
-- 
function :: A -> B
function = ...
```

docs/StyleGuide.md
```diff
@@ -0,0 +1,1142 @@
...
+
+    ```haskell
+    data API m = API {
+          foo_ :: HasCallStack => Maybe a -> m a
+        }
+    foo :: HasCallStack => API m -> Maybe a -> m a
+    foo = foo_
+    ```
+
+    Without the extra wrapper `foo`, the call stack would only start at `_foo`,
```
bolt12, line 1047: `foo_` and not `_foo`

bolt12 (approved):
Looks good! I just have some small preference differences but they are that that strong of a preference to me! 

Regarding records, I prefer hanging ',' style visually, but can understand the reasoning for the other ones 😄

docs/StyleGuide.md
coot: good question.  The order should be dictated by the type classes, not the `deriving` clause syntax, hence leaving it unspecified makes more sense to me.  What do you think?

bolt12: I agree!

docs/StyleGuide.md
coot: I think this one with haddocks:

```haskell
data Foo
    = -- | Bar
      --
      Bar Int Int
      
    | -- | Baz
      --
      Baz Int
          Int
          (Maybe Bool)
          [Foo]
```

will be more consistent than one where `=` on the `data Foo =` line.

bolt12: Yes I agree!

docs/StyleGuide.md
```diff
@@ -0,0 +1,1142 @@
...
+      ```haskell
+      case x of y
+        A x -> A_body
+        B y ->
+          let bl = bl_body y
+          in  B_body
+      ```
+
+      Note that we align `B_body` with `bl` in the `let` block. At the moment we
+      are not being very consistent with this.
```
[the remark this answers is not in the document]
bolt12: Okay!

docs/StyleGuide.md
coot: Resolved.

docs/StyleGuide.md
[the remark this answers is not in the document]
coot: btw, the space after `in` is ok, we're just not saying it's mandatory.

coot:
bors merge

PR #3882, opened 2022-07-11
CAD-4199 Anti-diff prototype for `SeqUtxoDiff`.

### LINKED ISSUE #3946: Anti-diff prototype
# Problematic performance for sequences of diffs

We replayed the mainnet chain from Genesis and reached Alonzo. During our performance tests we found out that 20% of the time is spent in the flushing operation. This can be brought down to 14% by flushing less frequently, but after a certain threshold decreasing the flushing frequency no longer decreases the time spent in this operation. This threshold was found to be flushing around 1K blocks.

The vast majority of the cost belongs to `Map.unionWith` which is used to implement the `<>` method for the `Semigroup` instance for `SeqUtxoDiff`, which in turn is called by the split operation of the fingertree data structure (which forces the computation of the intermediate measures of the fingertree via the `Semigroup` `SudMeasure` instance) which in turn is needed to flush.

Therefore we have to try to design an alternative approach to maintaining the sequence of differences (which is the data structure used when flushing to disk).

Maybe such an alternative is not feasible, or has other tradeoffs that are less acceptable than the slowdown that the current flush operation has. In this case we might need to optimize the existing implementation, and in the worst-case scenario we might need to accept the performance loss.

## Goals
- [x] (1) Elaborate a design and prototype to determine if this approach is feasible.
- [x] (2) If the approach is feasible, compare the performance of the prototype's "new" approach to the performance of the "legacy" implementation.

# (1) Prototyping the anti-diff approach

One approach to solving the problem described above is to keep track of a total/top-level cumulative diff, instead of caching cumulative diffs for sub-sequences of the diff sequence. If we can represent diffs as "negative" diffs, we could update this top-level cumulative diff by adding/subtracting diffs, instead of reconstructing the cumulative diff from precomposed subs-sequences of diffs. This should lead to a constant number of sum operations, instead of a logarithmic number of them.

In this approach:
* anti/inverse diffs are applied to either the left or right side of the total cumulative diff to allow for fast splits when rolling back or flushing to disk.
* new differences that arise from block applications are applied to the right side.

## Example

The example below illustrates the anti-diff approach. We consider a number of consensus-relevant operations on diff sequences and how these affect the top-level cumulative diff. These operations are: Push, flush and rollback. 

```
Individual diffs:  [1, 2, 3, 4]
Cached total diff: 1+2+3+4
```

### Adding/pushing diffs: add diffs to the right end of the diff sequence

```
[1, 2, 3, 4] -> [1, 2, 3, 4, 5]
1+2+3+4      -> 1+2+3+4+5
```

### Flushing diffs: split the diff sequence/take diffs on the left end of the diff sequence.

We have (at least) two options:
* We create fresh diffs on the left side, and we sum its inverse on the left of the right side.
```
[1, 2, 3, 4, 5] -> ([1], [2, 3, 4, 5])
1+2+3+4+5       -> (1,   (-1) + (1+2+3+4+5))
```
```
[1, 2, 3, 4, 5] -> ([1, 2], [3, 4, 5])
1+2+3+4+5       -> (1+2,    (-(1+2)) + (1+2+3+4+5))
```

* We sum the inverse on the (right of) the left side, and we sum the inverse on the (left of) the right side
```
[1, 2, 3, 4, 5] -> ([1],                     [2, 3, 4, 5])
1+2+3+4+5       -> ((1+2+3+4+5) - (2+3+4+5), (-1) + (1+2+3+4+5))
```
```
[1, 2, 3, 4, 5] -> ([1, 2],                  [3, 4, 5])
1+2+3+4+5       -> ((1+2+3+4+5) - (3+4+5), (-(1+2)) + (1+2+3+4+5))
```

We should use the first option: we expect the left side of the split to be the smaller than the right side. That is, the remaining diff sequence should be much larger than the part that we flush.

### Rolling back diffs: split the diff sequence/take diffs on the right end of the sequence

We have (at least) two options:
* We create a fresh top-level measure after rolling back.
```
[1, 2, 3, 4, 5] -> [1, 2, 3] -> [1, 2, 3]
1+2+3+4+5       -> {}        -> 1+2+3
```
* We sum the inverse on (the right of) the total cumulative diff. 
```
[1, 2, 3, 4, 5] -> [1, 2, 3]
1+2+3+4+5       -> (1+2+3+4+5) + (-(4+5))
```

We should use the second option: we expect the right side of the split to be smaller than the left side. That is, the remaining diff sequence should be much larger than the part that we roll back.

# (2) Comparing performance

Finally, we need a benchmark setup where we can sample the performance of the current "legacy" implementation, and the "new" anti-diff approach. 

The benchmark should measure the time of applying a sequence of the following operations to a diff sequence:
* Push (insert on the right).
* Flush (split the sequence at a given point/take on the right side).
* Rollback (split the sequence at a given point/take on the left side).
* Forward (apply the differences in the sequence up to the tip).

The benchmarks should present at least the results for this configuration:
* An equal number of forward and pushes.
* 1 flush per every 10 pushes.
* 1 rollback every 100 pushes

jorisdral:
NOTE: This PR description is under continuous construction.

# Description

This PR resolves #3946. In summary, this PR:
* (1) Implements a prototype for an alternative way of keeping track of sequences of (UTxO) diffs, which uses abstract algebra (i.e., `Group` properties) as a means for optimisation.
* (2) Implements a benchmark suite that measures and compares the performance of the "legacy" and "new" implementations.

## (1) Prototyping the anti-diff approach

See commit history. 

## (2) Comparing performance

Tentative benchmarking results warrant cautious positivity: The anti-diff approach shows a ~0.42x improvement over the legacy approach. Some variation in the improvement factor across runs of the benchmark suite could be explained by the fact that there is some randomness in the construction of the benchmarking scenario. Nonetheless, I have observed that the improvement factor is often in the range `[0.40, 0.50]`.

Additional experimentation should show whether these results hold up when we change the configuration of the benchmarks. Examples of configuration options are the length of the command sequence, how many key-value pairs are deleted/inserted on a push, the probability distribution for rollback sizes, the security parameter `k`, and more. See `GenConfig` in the benchmarking code for more information about configuration options. 

```bash
$ cabal run ouroboros-consensus-test:bench-storage
Up to date
All
  Ouroboros
    Consensus
      Storage
        LedgerDB
          HD
            DiffSeq/Diff operations
              Comparative performance analysis
                AntiDiff
                  Benchmark interpret:                                       OK (1.12s)
                    34.2 ms ± 1.9 ms, 0.43x
                  Sanity check:                                              OK (0.03s)
                    +++ OK, passed 1 test.
                LegacyDiff
                  Benchmark interpret:                                       OK (1.23s)
                    79.0 ms ± 2.0 ms
                  Sanity check:                                              OK (0.07s)
                    +++ OK, passed 1 test.
                interpret AntiDiff == interpret LegacyDiff:                  OK (0.11s)
                  +++ OK, passed 1 test.
              deterministic: interpret AntiDiff == interpret LegacyDiff:     OK (20.28s)
                +++ OK, passed 100 tests.
              non-deterministic: interpret AntiDiff == interpret LegacyDiff: OK (18.87s)
                +++ OK, passed 100 tests.

All 7 tests passed (41.92s)
```
EDIT (24 Aug 2022): Since recent changes have been made, the improvement factor is now in the range `[0.60, 0.70]`.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested


ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+      if n == 0 then
+        0
+      else
+        fromIntegral k / fromIntegral n
+    nr :: Float
+    nr = fromIntegral (nrCommands conf)
+
+  fgen (conf {nrCommands = round $ p * nr})
+
+type CmdGen v a = StateT (Model Int v) Gen a
```
jorisdral, line 382: Consider using a newtype.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+
+-- | Generate a list of commands in non-deterministic order.
+genCmds :: (Eq v, Arbitrary v) => GenConfig -> CmdGen v [Cmd 'New Int v]
+genCmds conf = mapM (const $ genCmd conf) [1 .. nrCommands conf]
+
+-- | Generate a list of commands in deterministic order.
+genCmds' :: forall v. (Eq v, Arbitrary v) => GenConfig -> CmdGen v [Cmd 'New Int v]
```
jorisdral, line 389: This function should be documented better -- it is not clear how the deterministic order is achieved.

Furthermore, "deterministic order" might be confusing: In this case, the order of commands is deterministic, but the inputs generated for each command are random.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+      nrg <- gets nrGenerated
+      let nrc = nrCommands conf
+
+      if nrg < nrc then
+        gen
+      else
+        pure []
+
+    singleton :: a -> [a]
+    singleton x = [x]
```
jorisdral, line 432: A newer version of `base` should have a `singleton` function. Add a FIXME to remove this function when we depend on this newer version of `base`.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+    genN n gen = concat <$> replicateM n gen
+
+genCmd :: (Eq v, Arbitrary v) => GenConfig -> CmdGen v (Cmd 'New Int v)
+genCmd conf = do
+    frequency' [
+        (100, genPush conf)
+      , (10, genFlush conf)
+      , (1, genRollback conf)
+      , (100, genForward conf)
+      ]
```
jorisdral, line 444: Document why we use these frequencies

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+
+{-------------------------------------------------------------------------------
+  Command generator: Individual command generation
+-------------------------------------------------------------------------------}
+
+-- | Generate a @'Push'@ command.
+--
+-- > data Cmd (i :: Imp) k v =
+-- >    Push (DiffSeq i k v) (SlotNo i) (Diff i k v)
```
jorisdral, line 453:
```suggestion
-- >    Push (SlotNo i) (Diff i k v)
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+-- >    -- ...
+--
+-- Steps to generate a @'Push'@ command:
+-- * Forward backing values.
+-- * "Simulate a transaction" by generating a diff that:
+--   1. Deletes one or more values that exist in the forwarded backing values.
+--   2. Inserts one or more values with globally unique keys.
+-- * Return the previous diff sequence, a strictly increasing slot number, and
+--   the new diff.
```
jorisdral, line 462:
```suggestion
-- * Return a strictly increasing slot number, and the new diff.
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+  vs <- gets backingValues
+  kc <- gets keyCounter
+
+  let
+    _vs'@(TT.TableValues m) = TT.forwardValues vs (DS.cumulativeDiff ds)
+
+  taken :: [(Int, v)] <-
+    take nrToDelete <$> lift (shuffle (Map.toList m))
+  made :: [(Int, v)]  <-
+    lift $ mapM genKeyValue [kc .. kc + nrToInsert - 1]
```
jorisdral, line 486: Extract into separate, documented functions

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+
+  let
+    d = TT.TableDiff . MapDiff.Diff $
+         Map.fromList [(k, MapDiff.singletonDelete v) | (k,v) <- taken]
+      <> Map.fromList [(k, MapDiff.singletonInsert v) | (k,v) <- made]
```
jorisdral, line 491: Extract into separate, documented function

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+    , nrGenerated = nrGenerated st + 1
+    })
+
+  pure $ Push (fromIntegral t) d
+
+-- | Generate a @'Flush'@ command.
+--
+-- >  data Cmd (i :: Imp) k v =
+-- >    -- ...
+-- >    | Flush Int (DiffSeq i k v)
```
jorisdral, line 506:
```suggestion
-- >    | Flush Int
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+-- >    -- ...
+--
+-- Steps to generate a @'Flush'@ command:
+-- * Compute how many diffs @n@ in the diff sequence have a slot number that
+--   exceeds the security parameter @k@.
+-- * Return the current diff sequence and @n@.
```
jorisdral, line 512:
```suggestion
-- * Return @n@.
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+    , nrGenerated = nrGenerated st + 1
+    })
+
+  pure $ Flush n
+
+-- | Generate a @'Rollback'@ command.
+--
+-- >  data Cmd (i :: Imp) k v =
+-- >    -- ...
+-- >    | Rollback Int (DiffSeq i k v)
```
jorisdral, line 537:
```suggestion
-- >    | Rollback Int
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+-- >    -- ...
+--
+-- Steps to generate a @'Rollback'@ command:
+-- * Pick how much to rollback, i.e. an integer @n@.
+--   * With high probability, we perform small rollbacks of around @0.05 * k@.
+--   * With low probability, we perform large rollbacks of around @0.95 * k@.
+-- * Return the current diff sequence and @n@.
```
jorisdral, line 544:
```suggestion
-- * Return @n@.
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+--
+-- >  data Cmd (i :: Imp) k v =
+-- >    -- ...
+-- >    | Rollback Int (DiffSeq i k v)
+-- >    -- ...
+--
+-- Steps to generate a @'Rollback'@ command:
+-- * Pick how much to rollback, i.e. an integer @n@.
+--   * With high probability, we perform small rollbacks of around @0.05 * k@.
+--   * With low probability, we perform large rollbacks of around @0.95 * k@.
```
jorisdral, line 543:
```suggestion
```
These numbers are configurable, so shouldn't be part of the comment.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+
+    pure $ Rollback n
+  where
+    ofK :: Float -> Int
+    ofK p =
+      let
+        x :: Float
+        x = p * fromIntegral k
+      in
+        round x
```
jorisdral, line 576: Document

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+
+-- | Generate a @'Forward'@ command.
+--
+-- >  data Cmd (i :: Imp) k v =
+-- >    -- ...
+-- >    | Forward (Values i k v) (Keys i k v) (DiffSeq i k v)
```
jorisdral, line 582:
```suggestion
-- >    | Forward (Values i k v) (Keys i k v)
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+-- | Sanity checks for interpretation as fold.
+sanityCheck ::
+     ( Eq k, Eq v, Show k, Show v
+     , Isomorphism (DiffSeq i k v) (DS.DiffSeq 'TS.UTxO k v)
+     )
+  => Model k v
+  -> [Cmd i k v]
+  -> DiffSeq i k v
+  -> Property
+sanityCheck model _cmds result = diffs model === to result
```
jorisdral, line 631: One sanity check that could/should be included is one that checks whether the model and interpretation match for every step of the interpretation function

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+-- >  data Cmd (i :: Imp) k v =
+-- >    -- ...
+-- >    | Forward (Values i k v) (Keys i k v) (DiffSeq i k v)
+--
+-- Steps to generate a @'Forward'@ command:
+-- * Determine which keys to forward.
+-- * Retrieve (a subset of) the backing values from the model, i.e.,
+--   /read/ the keys (obtained in previous step) from the backing values.
+-- * Retrieve the current diff sequence from the model.
+-- * Return the previous three combined.
```
jorisdral, line 589:
```suggestion
-- * Return the previous two combined.
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+-- BOOKKEEPING: None, since forwarding doesn't alter any model state.
+--
+-- Note: The keys to forward should be a superset of the keys of the backing
+-- values that we retrieve from the model.
```
jorisdral, line 593: Why? Because the set of keys is used to "read" backing values.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+    ForwardConfig{nrToForward} = forwardConfig conf
+
+  ds  <- gets diffs
+  bvs <- gets backingValues
+
+  let
+    d              = DS.cumulativeDiff ds
+    TT.TableKeys s = TT.diffKeys d
+
+  ksList <- take nrToForward <$> lift (shuffle (toList s))
```
jorisdral, line 606: Extract into a separate, documented function

ouroboros-consensus-test/bench-storage/Main.hs
```diff
@@ -0,0 +1,19 @@
+module Main (main) where
+
+--import           Criterion.Main
```
jorisdral, line 3:
```suggestion
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,631 @@
...
+{-# LANGUAGE MultiParamTypeClasses   #-}
+{-# LANGUAGE NamedFieldPuns          #-}
+{-# LANGUAGE ScopedTypeVariables     #-}
+{-# LANGUAGE StandaloneDeriving      #-}
+{-# LANGUAGE TupleSections           #-}
+{-# LANGUAGE TypeApplications        #-}
+{-# LANGUAGE TypeFamilyDependencies  #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
+
+{-# LANGUAGE UndecidableInstances    #-}
```
jorisdral, line 19: Make into one list.

ouroboros-consensus-test/ouroboros-consensus-test.cabal
```diff
@@ -317,3 +325,39 @@ test-suite test-infra
...
+benchmark bench-storage
+  type:                exitcode-stdio-1.0
+  hs-source-dirs:      bench-storage
+  main-is:             Main.hs
+  other-modules:
+                       Bench.Ouroboros.Consensus.Storage.LedgerDB.HD
+  build-depends:       base
+                     , bifunctors
+                     , containers
+                     , criterion
```
jorisdral, line 338:
```suggestion
```

ouroboros-consensus-test/src/Test/Util/Orphans/DiffSeq/Arbitrary.hs
```diff
@@ -0,0 +1,44 @@
...
+  Diffs
+------------------------------------------------------------------------------}
+
+deriving newtype instance (Ord k, Arbitrary k, Arbitrary v)
+                       => Arbitrary (MapDiff.Diff k v)
+deriving newtype instance (Arbitrary v) => Arbitrary (MapDiff.DiffHistory v)
+instance (Arbitrary v) => Arbitrary (MapDiff.DiffEntry v) where
+  arbitrary = do
+    constr <- elements [MapDiff.Insert, MapDiff.Delete]
+    constr <$> arbitrary
```
jorisdral, line 27: This is not the most general arbitrary instance. Instead, create a `newtype` wrapper that has this instance's behaviour.

ouroboros-consensus-test/src/Test/Util/Orphans/DiffSeq/Arbitrary.hs
```diff
@@ -0,0 +1,44 @@
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Test.Util.Orphans.DiffSeq.Arbitrary () where
```
jorisdral, line 7: Is it necessary to have this as a separate module, instead of having these instances in `Test.Util.Orphans.Arbitrary`?

ouroboros-consensus-test/src/Test/Util/Orphans/Isomorphism.hs
```diff
@@ -0,0 +1,125 @@
...
+{------------------------------------------------------------------------------
+  @'Isomorphism'@ class
+------------------------------------------------------------------------------}
+
+-- FIXME(jdral): This class is based on the @IsomorphicTo@ class from REF. We
+-- can not use the package at this time because the cabal update index is too
+-- outdated. Once the index has been updated to a more recent one, we should
+-- switch to this package.
+--
+-- REF: https://hackage.haskell.org/package/isomorphism-class-0.1.0.5/docs/IsomorphismClass.html
```
jorisdral, line 35: Should be fixable when `feature/utxo-hd` is rebased on the most recent version of `master`.

ouroboros-consensus-test/src/Test/Util/Orphans/Isomorphism.hs
```diff
@@ -0,0 +1,125 @@
...
+instance Isomorphism (TT.TableDiff ts k v) (HD.UtxoDiff k v) where
+  to (TT.TableDiff (MapDiff.Diff m)) = HD.UtxoDiff (fmap to' m)
+    where
+      to' :: MapDiff.DiffHistory v -> HD.UtxoEntryDiff v
+      to' = \case
+        MapDiff.DiffHistory (_xs Seq.:|> x) ->
+          to'' x
+        _ ->
+          error "A DiffHistory is isomorphic to a UtxoEntryDiff under the \
+                \ assumption that diff histories contain exactly one element."
```
jorisdral, line 84: Explain why we "assume" this in a comment.

ouroboros-consensus-test/src/Test/Util/Orphans/NFData.hs
```diff
@@ -0,0 +1,75 @@
...
+
+
+-- | FIXME: We should verify that this instance is sufficient. Unfortunately,
+-- we can not automaticaly derive the @'NFData'@ instance, since the
+-- 'fingertree' package does not export all necessary constructors to
+-- automatically derive the NFData instance for the underlying
+-- @'FingerTree'@.
+instance (NFData a, NFData v, FT.Measured v a)
+      => NFData (FT.StrictFingerTree v a) where
+  rnf ft = rnf (FT.measure ft) `seq` rnf (toList ft)
```
jorisdral, line 35: This FIXME should be fixed before merging this PR.

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -252,6 +255,8 @@ library
 
                        -- Strict wrapper around SOP
                        Data.SOP.Strict
+                       Data.FingerTree.TopMeasured.Strict
+                       Data.Map.Diff.Strict
```
jorisdral, line 259: Where should these modules live? They are not `consensus`-specific.

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -235,6 +235,9 @@ library
                        Ouroboros.Consensus.Storage.LedgerDB.DiskPolicy
                        Ouroboros.Consensus.Storage.LedgerDB.HD
                        Ouroboros.Consensus.Storage.LedgerDB.HD.BackingStore
+                       Ouroboros.Consensus.Storage.LedgerDB.HD.DiffSeq
```
jorisdral, line 238: Can we make this module `consensus`-agnostic, or is this unnecessary abstraction? Specifically, should we make a general-purpose diff sequence datatype?

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -312,6 +317,7 @@ library
                      , directory         >=1.3   && <1.4
                      , filelock
                      , filepath          >=1.4   && <1.5
+                     , groups
```
jorisdral, line 320: Should we pin this dependency?

ouroboros-consensus/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,164 @@
...
+import           Data.Group
+import           GHC.Generics (Generic)
+
+import           NoThunks.Class (NoThunks (..), noThunksInValues)
+
+import           Data.FingerTree.Strict (Measured)
+import qualified Data.FingerTree.Strict as FT
+
+-- TODO(jdral): Should we force strictness anywhere? Or more generally, where
+-- should we guide evaluation?
```
jorisdral, line 37: I think this TODO is fixed, unless someone has other insights

ouroboros-consensus/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,164 @@
...
+
+{-------------------------------------------------------------------------------
+  Measuring
+-------------------------------------------------------------------------------}
+
+-- | Internal measures are used by the internal @'StrictFingerTree'@.
+type InternalMeasured v a = Measured v a
+
+measureInternal :: Measured v a => a -> v
+measureInternal = FT.measure
```
jorisdral, line 74: Do we really need to make this distinction? I'd say, let's remove this.

ouroboros-consensus/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,164 @@
...
+
+-- | All values of type @Alt@ are internal-measured.
```
jorisdral, line 76: The `Alt` datatype is now called `StrictFingerTree`. We should ensure that all references to `Alt` are removed in comments (and other code)

ouroboros-consensus/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,164 @@
...
+-- > @instance Measured T  a where -- ...@
+-- > @instance Measured T' a where -- ...@
+-- Furthermore, we want the top-level measure to be a @'Group'@ instead of a
+-- @'Monoid'@.
+class Group v => TopMeasured v a | a -> v where
+  measureTop :: a -> v
+
+-- | All values of type @Alt@ are top-measured.
+instance TopMeasured vt a => TopMeasured vt (StrictFingerTree vt vi a) where
+  measureTop = tm
```
jorisdral, line 94: We can probably auto-derive this.

ouroboros-consensus/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,164 @@
...
+
+-- | Wrapper for when we need both @TopMeasured@ and @InternalMeasured@
+type SuperMeasured vt vi a = (TopMeasured vt a, InternalMeasured vi a)
```
jorisdral, line 97: Reconsider this name

ouroboros-consensus/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,164 @@
...
+
+{-------------------------------------------------------------------------------
+  API
```
jorisdral, line 100: This is a minimal API based on what `consensus` requires. Should we extend this API or is it out of the scope of this PR? I can imagine that we would want to extend this API only when we decide whether to integrate the new anti-diff approach in the `consensus` code. This integration would also be a separate issue/PR.

ouroboros-consensus/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,164 @@
...
+  -> (v01 -> v02)
+  -> StrictFingerTree v01 v1 a1
+  -> StrictFingerTree v02 v2 a2
+fmap'' f g (StrictFingerTree v0 sft) = StrictFingerTree v0' sft'
+  where
+    sft' = FT.fmap' f sft
+    v0' = g v0
+
+
+-- TODO: Choice of left- or right-Delta should depend on sizes.
```
jorisdral, line 148: It is worth looking into this, and seeing what effect it has on the performance

ouroboros-consensus/src/Data/Map/Diff/Strict.hs
```diff
@@ -0,0 +1,137 @@
...
+  deriving stock (Generic, Show, Eq)
+  deriving anyclass (NoThunks)
+
+-- | A history of changes to a value in a key-value store.
+--
+-- A history has an implicit sense of ordering according to time: from left to
+-- right. This means that the left-most element in the history is the
+-- /earliest/ change, while the right-most element in the history is the
+-- /latest/ change.
+newtype DiffHistory v = DiffHistory (Seq (DiffEntry v))
```
jorisdral, line 44: Reconsider this name

ouroboros-consensus/src/Data/Map/Diff/Strict.hs
```diff
@@ -0,0 +1,137 @@
...
+singletonInsert = singleton . Insert
+
+singletonDelete :: v -> DiffHistory v
+singletonDelete = singleton . Delete
+
+-- | A change to a value in a key-value store.
+--
+-- Note: updates are equivalent to inserts, since we consider them to
+-- overwrite previous values.
+data DiffEntry v = Insert !v | Delete !v
```
jorisdral, line 61: Reconsider this name

ouroboros-consensus/src/Data/Map/Diff/Strict.hs
```diff
@@ -0,0 +1,137 @@
...
+instance (Ord k, Eq v) => Semigroup (Diff k v) where
+  Diff m1 <> Diff m2 = Diff $
+    Merge.merge
+      Merge.preserveMissing
+      Merge.preserveMissing
+      (Merge.zipWithMaybeMatched
+        (\_k h1 h2 -> isNonEmptyHistory $ h1 <> h2)
+      )
+      m1
+      m2
```
jorisdral, line 78: Document why this is how we merge diffs. For example, why do we use `isNonEmptyHistory`?

ouroboros-consensus/src/Data/Map/Diff/Strict.hs
```diff
@@ -0,0 +1,137 @@
...
+
+instance (Ord k, Eq v) => Monoid (Diff k v) where
+  mempty = Diff mempty
+
+instance (Ord k, Eq v) => Group (Diff k v) where
+  invert (Diff m) = Diff $ fmap invert m
```
jorisdral, line 84: We can probably auto-derive these.

ouroboros-consensus/src/Data/Map/Diff/Strict.hs
```diff
@@ -0,0 +1,137 @@
...
+  Class instances for @'DiffHistory'@
+------------------------------------------------------------------------------}
+
+instance Eq v => Semigroup (DiffHistory v) where
+  DiffHistory s1 <> DiffHistory s2 = DiffHistory $ s1 `mappend'` s2
+    where
+      mappend' Empty Empty                   = Empty
+      -- At the ``touching'' ends of the sequences, take off diff entries that
+      -- are each other's inverse until we find two non-inverse entries. In this
+      -- case, we can not continue so we return the concatenated remainders.
```
jorisdral, line 96: Give examples

ouroboros-consensus/src/Data/Map/Diff/Strict.hs
```diff
@@ -0,0 +1,137 @@
...
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GADTs                      #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE KindSignatures             #-}
+{-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+
+module Data.Map.Diff.Strict (
```
jorisdral, line 13: A module level haddock comment explaining the approach taken here (supplemented with examples) would probably be useful to readers

ouroboros-consensus/src/Data/Map/Diff/Strict.hs
```diff
@@ -0,0 +1,137 @@
...
+      -- At the ``touching'' ends of the sequences, take off diff entries that
+      -- are each other's inverse until we find two non-inverse entries. In this
+      -- case, we can not continue so we return the concatenated remainders.
+      mappend' xs0@(xs :|> x) ys0@(y :<| ys)
+        | areInverses x y                    = mappend' xs ys
+        | otherwise                          = xs0 Seq.>< ys0
+      mappend' xs ys                         = xs Seq.>< ys
+
+instance Eq v => Monoid (DiffHistory v) where
+  mempty = DiffHistory $ mempty
```
jorisdral, line 103: We can probably autoderive this

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/ToStore.hs
```diff
@@ -0,0 +1,3 @@
+module Ouroboros.Consensus.Storage.LedgerDB.HD.ToStore (ToStoreKind (UTxO)) where
```
jorisdral, line 1: Seems unnecessary to have a separate module just for this datatype, but it prevents cyclic dependencies

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/TableTypes.hs
```diff
@@ -0,0 +1,129 @@
...
+
+{-------------------------------------------------------------------------------
+  Constructing table contents
+-------------------------------------------------------------------------------}
+
+valuesFromList :: Ord k => [(k, v)] -> TableValues ts k v
+valuesFromList = TableValues . Map.fromList
+
+{-------------------------------------------------------------------------------
+  Interactions between keys, values and diffs
```
jorisdral, line 69: This does not really correctly summarise the functions in this section

jorisdral: Furthermore, these functions need to be documented.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,289 @@
...
+import           Data.Group
+import           Data.Monoid (Sum (..))
+import           Data.Semigroup (Max (..))
+import           GHC.Generics (Generic)
+import           NoThunks.Class (NoThunks)
+
+import           Data.FingerTree.TopMeasured.Strict hiding (split)
+import qualified Data.FingerTree.TopMeasured.Strict as TMFT (split)
+
+-- import           Cardano.Binary (FromCBOR (..), ToCBOR (..))
```
jorisdral, line 57:
```suggestion
```

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,289 @@
...
+
+import           Ouroboros.Consensus.Storage.LedgerDB.HD.TableTypes
+import           Ouroboros.Consensus.Storage.LedgerDB.HD.ToStore (ToStoreKind)
+
+
+{-------------------------------------------------------------------------------
+  Sequences of diffs
+-------------------------------------------------------------------------------}
+
+newtype DiffSeq (ts :: ToStoreKind) k v =
```
jorisdral, line 68: Again, can we make this more general and not `consensus`-specific?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,289 @@
...
+-- | Short-hand for @'InternalMeasured'@.
+type IM (ts :: ToStoreKind) k v =
+  InternalMeasured (InternalMeasure ts  k v) (Element ts k v)
+
+-- | Short-hand for @'SuperMeasured'@.
+type SM (ts :: ToStoreKind) k v =
+  (TM ts k v, IM ts k v)
+
+{-------------------------------------------------------------------------------
+  API: type classes
```
jorisdral, line 173: These type classes seem nonsensical because they only have one instance (for the `DiffSeq` datatype above), but they could be an ingredient for generalising the `DiffSeq` datatype

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,289 @@
...
+instance HasLength ts k v where
+  getLength (TopMeasure s _) = s
+  getInternalMeasureLength (InternalMeasure s _) = s
+
+instance HasSlot ts k v where
+  getElementSlot (Element sl _) = sl
+  getInternalMeasureSlot (InternalMeasure _ sl) = sl
+
+{-------------------------------------------------------------------------------
+  API: derived functions
```
jorisdral, line 203: These functions should be documented

jorisdral (commented):
Self-review

ouroboros-consensus/src/Data/FingerTree/TopMeasured/Strict.hs
jorisdral: Elaborated on this in the comment, so it is left as future work.

ouroboros-consensus/src/Data/Map/Diff/Strict.hs
jorisdral: Elaborated on the use of `isNonEmptyHistory` in a comment.

anti-diff/src/AntiDiff/Util/Auto.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE DerivingStrategies         #-}
```
dnadales, line 1: Do we use this?

anti-diff/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,194 @@
...
+split ::
+     ( SuperMeasured vt vi a
+     , Group vt
+     )
+  => (vi -> Bool)
+  -> StrictFingerTree vt vi a
+  -> ( StrictFingerTree vt vi a
+     , StrictFingerTree vt vi a
+     )
+split p = split' p f
```
dnadales, line 131: Is this the only place where we use the `invert` operation?

dnadales (commented):
I didn't have time to complete the review. Here are some top-level considerations:

- I think we should split this large PR and use stacked PR's. A sequence of PR's that comes to mind is:
    - create a benchmarking setup
    - write generators for the commands
    - benchmark the cumulative difference (aka legacy) implementation
    - write the anti-diff prototype 
    - hook the prototype up to the benchmarks 
 - An action point for me would be to understand why do we still need to use fingertrees, I would have expected sequences would have sufficed if we have inverse diff operations.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,723 @@
...
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE TupleSections              #-}
+{-# LANGUAGE TypeApplications           #-}
+{-# LANGUAGE TypeFamilyDependencies     #-}
+{-# LANGUAGE UndecidableInstances       #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+{- | Comparative benchmarks for legacy/anti-diff approaches to diff sequences.
```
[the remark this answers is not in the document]
jorisdral: In a way, the anti-diff propotype can also be viewed as being cumulative, so I'm not sure about the name -- I do agree that "legacy" is probably not descriptive enough. Another description that comes to mind is "caching", but again the two approaches perform caching, just using a different method. So, maybe we should somehow it by its caching method? Something like "sub-cached"?

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,723 @@
...
+    Push     (SlotNo i)      (Diff i k v)
+  | Flush    Int
+  | Rollback Int
+  | Forward  (Values i k v)  (Keys i k v)
+  deriving stock Generic
+
+-- TODO: Add trace to custom instance that shows when forced, don't keep in
+-- benchmark code.
+-- TODO: Look at output with -ddump-deriv.
+deriving anyclass instance ( NFData (DiffSeq i k v)
```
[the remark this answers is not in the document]
jorisdral: Commands are part of the input to the benchmarks, and if we do not evaluate them before running the benchmark the benchmark measurements would possibly be contaminated with time spent evaluating parts of the inputs

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,723 @@
...
+  => [Cmd i k v]
+  -> DiffSeq i k v
+interpret = foldl go mempty
+  where
+    go :: DiffSeq i k v -> Cmd i k v -> DiffSeq i k v
+    go ds = \case
+      Push sl d     -> push ds sl d
+      -- We @'deepseq'@ the left part of the flush result, because that is the
+      -- part that would be written to disk, for which it must be fully
+      -- evaluated.
```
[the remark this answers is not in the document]
jorisdral: Time spent evaluating `l` is implementation-dependent, and since we flush we know `l` will be fully evaluated before writing `l` to disk. However, @Jasagredo pointed out that we should actually not be evaluating `l`, but rather the cumulative diff of `l`.

jorisdral:
> I think we should split this large PR and use stacked PR's. A sequence of PR's that comes to mind is:
>   * create a benchmarking setup
>   * write generators for the commands
>   * benchmark the cumulative difference (aka legacy) implementation
>   * write the anti-diff prototype
>   * hook the prototype up to the benchmarks
> ...

As discussed with @Jasagredo, I will use [jdral/3946-anti-diff-prototype-main](https://github.com/input-output-hk/ouroboros-network/tree/jdral/3946-anti-diff-prototype-main) as a target for the stacked PRs.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,723 @@
...
+      -- can forget about it.
+      Rollback n    -> let (l, _r) = rollback n ds
+                       in  l
+      -- FIXME(jdral): Should we @'seq'@ or @'deepseq'@ the forwarding result?
+      -- Or, is there something else we should do with the result of @vs'@?
+      Forward vs ks -> let vs' = forward vs ks ds
+                       in  vs' `seq` ds
+
+-- | Common interface to diff sequence implementations.
+class Comparable (i :: Imp) where
```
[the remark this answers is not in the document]
jorisdral: I've rewritten the type class to this:

```haskell
type DiffSeqKind = Type -> Type -> Type

-- | Common interface to diff sequence implementations.
class IsDiffSeq (ds :: DiffSeqKind) k v where
  type Diff ds k v   = r | r -> ds k v
  type Values ds k v = r | r -> ds k v
  type Keys ds k v   = r | r -> ds k v
  type SlotNo ds     = r | r -> ds
```

Hopefully, this is more descriptive of what the type class represents

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,723 @@
...
+
+{- | Comparative benchmarks for legacy/anti-diff approaches to diff sequences.
+
+  The benchmarks measure the time needed to evaluate a sequence of operations
+  on diff sequences (and auxiliary datatypes). A sequence of operations should
+  mimick how diff sequences are used in Consensus.
+
+  We run these benchmarks for two approaches to diff sequences: the legacy and
+  anti-diff approaches. We compare performance across these two approaches.
+  We suspect that the anti-diff approach should be more performant.
```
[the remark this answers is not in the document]
jorisdral: I have removed the "suspect" part. I can also imagine that we would not want to keep the old implementation around if we decide to replace it with the anti-diff implementation

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,723 @@
...
+
+  The operations can be of four types:
+  * Push (insert on the right)
+  * Flush (split the sequence at a given point/take on the right side)
+  * Rollback (split the sequence at a given point/take on the left side)
+  * Forward (apply the cumulative diff up to the tip)
```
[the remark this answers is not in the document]
jorisdral: I'm not exactly sure what you mean, what would that operation look like for a diff sequence?

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,723 @@
...
+
+  Notes:
```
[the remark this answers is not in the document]
jorisdral: Done!

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -1,8 +1,10 @@
-{-# LANGUAGE DeriveAnyClass        #-}
-{-# LANGUAGE DeriveGeneric         #-}
-{-# LANGUAGE LambdaCase            #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE DeriveAnyClass             #-}
```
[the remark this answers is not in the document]
jorisdral: Adding additional language pragma's like `GeneralisedNewtypeDeriving` makes `stylish-haskell` reformat all of the language pragmas

anti-diff/src/Data/Map/Diff/Strict/Internal.hs
```diff
@@ -0,0 +1,352 @@
...
+forwardValuesAndKeys ::
+     (Ord k, Eq v, HasCallStack)
+  => Values k v
+  -> Keys k v
+  -> Diff k v
+  -> Values k v
+forwardValuesAndKeys v@(Values values) (Keys keys) (Diff diffs) =
+  forwardValues
+    v
+    (Diff $ diffs `Map.restrictKeys` (Map.keysSet values `Set.union` keys))
```
jorisdral, line 283: Should these return a `Maybe`, instead of throwing errors?

anti-diff/src/AntiDiff/Util/Auto.hs
jorisdral: I'm not sure if you refer to the module or language extension -- If you refer to the module, then yes it is being used in tests for (semi-)group(-oid) laws

anti-diff/src/AntiDiff/Util/X.hs
```diff
@@ -0,0 +1,12 @@
+-- | Utilities for very randomly generating very small integers.
+module AntiDiff.Util.X (X (..)) where
+
+import           Test.QuickCheck
+
+
+
+newtype X = X Int
```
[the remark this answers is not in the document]
jorisdral: I will remove this module, since we can use `Small`

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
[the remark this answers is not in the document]
jorisdral: I think these changes were left over from previous version of the prototype, but we do not need these changes

anti-diff/src/Data/FingerTree/TopMeasured/Strict.hs
```diff
@@ -0,0 +1,194 @@
+{-# LANGUAGE BangPatterns           #-}
+{-# LANGUAGE ConstraintKinds        #-}
+{-# LANGUAGE DeriveGeneric          #-}
+{-# LANGUAGE DerivingStrategies     #-}
+{-# LANGUAGE FlexibleContexts       #-}
+{-# LANGUAGE FlexibleInstances      #-}
+{-# LANGUAGE FunctionalDependencies #-}
+
+-- Note: Parts of the documentation are based on/are directly copied from
```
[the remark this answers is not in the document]
jorisdral: This may tie into the action point you mentioned: we define a thin wrapper around strict fingertrees

anti-diff/src/Data/Map/Diff/Strict/Internal.hs
```diff
@@ -0,0 +1,353 @@
...
+import           NoThunks.Class (NoThunks)
+
+{------------------------------------------------------------------------------
+  General-purposes diffs for key-value stores
+------------------------------------------------------------------------------}
+
+-- | A diff for key-value stores.
+--
+-- INVARIANT: If a key @k@ is present in the @Map@, then the @DiffHistory@ is
+-- non-empty. This prevents the @Map@ from getting bloated with empty diff
```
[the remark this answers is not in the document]
jorisdral: I've rephrased this

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,723 @@
...
+  -> Benchmark
+benchComparative name boundsMay setup =
+  bgroup name [
+    env setup $ \ ~(newCmds, legacyCmds, model) ->
+      bgroup "Comparative performance analysis" [
+        bgroup "AntiDiff" [
+          maybe bcompare (uncurry bcompareWithin) boundsMay target $
+            bench "Benchmark interpret" $
+                  nf interpret newCmds
+          , testProperty "Sanity check" $
```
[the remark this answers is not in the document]
jorisdral: It is not passed to the sanity check part, but the indentation is off

anti-diff/src/Data/Map/Diff/Strict/Internal.hs
```diff
@@ -0,0 +1,353 @@
...
+-- such, it is the using code's responsibility to ensure that the inputs given
+-- to the sum lead to sensible results.
+--
+-- Note: We do not cancel out consecutive elements in @h1@ and @h2@
+-- individually. It is only at the border between @h1@ and @h2@ that we cancel
+-- out elements (see the third example given above).
+instance Eq v => Semigroup (DiffHistory v) where
+  DiffHistory s1 <> DiffHistory s2 = DiffHistory $ s1 `mappend'` s2
+    where
+      -- At the ``touching'' ends of the sequences, take off diff entries that
```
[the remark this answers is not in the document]
jorisdral: It is an optimisation because the aim was that combining inserts/deletes with anti-inserts/anti-deletes should make the diff history smaller, like it's subtracting elements. Just from the top of my head, it is also essential to its correctness because it must satisfy the `Group` laws: `invert x <> x == mempty` and `x <> invert x == mempty`

ouroboros-consensus-test/ouroboros-consensus-test.cabal
```diff
@@ -317,3 +327,41 @@ test-suite test-infra
...
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+                       -Wpartial-fields
+                       -Widentities
+                       -Wredundant-constraints
+                       -Wmissing-export-lists
+                       -Wno-unticked-promoted-constructors
+                       -fno-ignore-asserts
+
+--mixins:              tasty-bench (Test.Tasty.Bench as Criterion, Test.Tasty.Bench as Criterion.Main, Test.Tasty.Bench as Gauge, Test.Tasty.Bench as Gauge.Main)
```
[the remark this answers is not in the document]
jorisdral: I have removed this

jorisdral:
Closed in favour of #3997.

PR #3891, opened 2022-07-12
Fix Praos obsolete node check

JaredCorduan:
# Description

The obsolete node check in the `Praos` ledger protocol is suppose to compare the current major protocol version (from the protocol parameters in the ledger state) with the global "max major protocol version" value. Currently, however, it is checking that the global max major value is not bigger than the protocol version listed in the block header (which is not supposed to have any semantic meaning).

This PR fixes the check to perform the same logic done by the `TPraos` ledger protocol, ie the logic in the ledger function `chainChecks`. To do this, `Views.LedgerView` now contains the current protocol parameters.

 The consequence of the incorrect obsolete node check is that nodes will not properly halt after a hard fork when they do not have the required software for the hard fork.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested


ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Views.hs
```diff
@@ -43,6 +44,8 @@ data LedgerView crypto = LedgerView
     -- | Maximum header size
     lvMaxHeaderSize :: !Natural,
     -- | Maximum block body size
-    lvMaxBodySize   :: !Natural
+    lvMaxBodySize   :: !Natural,
+    -- | Current protocol version
+    lvProtocolVersion   :: !ProtVer
```
nfrisby, line 49: Style: please align the `::`s

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
```diff
@@ -73,8 +73,8 @@ instance PraosCrypto c => ProtocolHeaderSupportsEnvelope (Praos c) where
         BlockSizeTooLarge (bhviewBSize bhv) maxBodySize
     where
       pp = praosParams cfg
-      (MaxMajorProtVer m) = praosMaxMajorPV pp
-      (ProtVer maxpv _) = hbProtVer $ headerBody hdr
+      (MaxMajorProtVer maxpv) = praosMaxMajorPV pp
```
nfrisby, line 76: Hmm. I understand that this check was missing and needs to be added, but it's not obvious to me that this is the right place to do it.

This method and class are about check the header. But the `m` and `maxpv` being checked in this new logic no don't depend on the header at all 🤔 

EG see the comments at the definition of `ProtocolHeaderSupportsEnvelope` and `envelopeChecks`

https://github.com/input-output-hk/ouroboros-network/blob/0a3a61ae80244dde943843fd39ef17cd85260980/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/Abstract.hs#L100-L126

nfrisby (commented):
Thanks, Jared! Nice PR.

I did make one significant comment.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
JaredCorduan: I agree @nfrisby, this isn't a block header concern and is a bit weird to live here. The main reason to put it here is for consistency with `TPraos`.

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Views.hs
JaredCorduan: I ran `./scripts/ci/check-stylish.sh`

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
nfrisby: I see. So `TPraos` is implemented this way:

https://github.com/input-output-hk/ouroboros-network/blob/021c35a09861ca91c59db63d1d68a887cdbeb9ad/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/TPraos.hs#L49-L55

where `SL.chainChecks` is:

https://github.com/input-output-hk/cardano-ledger/blob/c1897b5176ac8a452dc173997bb23f981f85b197/eras/shelley/impl/src/Cardano/Ledger/Chain.hs#L68-L83

```haskell
chainChecks ::
  MonadError ChainPredicateFailure m =>
  Natural ->
  ChainChecksPParams ->
  BHeaderView crypto ->
  m ()
chainChecks maxpv ccd bhv = do
  unless (m <= maxpv) $ throwError (ObsoleteNodeCHAIN m maxpv)
  ...
  where
    (ProtVer m _) = ccProtocolVersion ccd
```

And yeah, that's also an odd "header" check.

I opened a Tech Debt issue to revisit this https://input-output.atlassian.net/browse/CAD-4742

nfrisby (approved):
Looks good to me, thanks Jared 👍

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
JaredCorduan: yep, exactly. my first reaction was "why, oh why, isn't Praos using `chainChecks`"? the answer to that, I found by looking around, seemed to be that the ledger's "ledger view" didn't make as much sense inside consensus (but would have saved us from this bug).

I'm glad you made a ticket, I like the idea of changing both TPraos and Praos to have this check elsewhere (but it will take some thought, and updating the specs).

JaredCorduan:
bors merge

amesgen:
bors retry

amesgen:
There is a rework of the GH Actions Windows CI coming up, which will hopefully resolve this :pray:

amesgen (commented):
This should be good to go now, the necessary CI improvements have been merged.

nfrisby:
bors retry

PR #3903, opened 2022-07-15
Fix GetUTxOBy(Address|TxIn)

Jasagredo:
# Description

There were three crucial problems fixed in this PR:
1. GetUtxoByAddress combined the UTxO inside the retrieved ledger states and not the tables. An `stowLedgerTables` and some magic was required.
2. GetUtxoByTxIn suffered from a very similar problem
3. `sgInitialFunds` were being lost because we updated the UTxO inside the genesis ledger state but then never combined that with the values that already existed outside the ledger state in the tables.

# Checklist

- Branch
    - [X] New tests are added if needed and existing tests are updated
    - [X] If this branch changes Consensus and has any consequences for downstream repositories or end users, said changes must be documented in [`interface-CHANGELOG.md`](../ouroboros-consensus/docs/interface-CHANGELOG.md)
    - [X] If this branch changes Network and has any consequences for downstream repositories or end users, said changes must be documented in [`interface-CHANGELOG.md`](../docs/interface-CHANGELOG.md)
    - [X] If serialization changes, user-facing consequences (e.g. replay from genesis) are confirmed to be intentional.
- Pull Request
    - [X] Self-reviewed the diff
    - [X] Useful pull request description at least containing the following information:
      - What does this PR change?
    - [X] Reviewer requested

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
@@ -62,7 +62,8 @@ import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Config
 import qualified Ouroboros.Consensus.HardFork.History as History
 import           Ouroboros.Consensus.HeaderValidation
-import           Ouroboros.Consensus.Ledger.Basics (ValuesMK)
+import           Ouroboros.Consensus.Ledger.Basics (StowableLedgerTables (..),
+                     ValuesMK, forgetLedgerTables)
 import           Ouroboros.Consensus.Ledger.Extended
 import qualified Ouroboros.Consensus.Mempool.TxLimits as TxLimits
 import           Ouroboros.Consensus.Node.NetworkProtocolVersion
@@ -816,9 +817,20 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
 
         register ::
              (EraCrypto era ~ c, ShelleyBasedEra era)
-          => Flip LedgerState mk (ShelleyBlock proto era)
-          -> Flip LedgerState mk (ShelleyBlock proto era)
-        register (Flip st) = Flip st {
+          => Flip LedgerState ValuesMK (ShelleyBlock proto era)
+          -> Flip LedgerState ValuesMK (ShelleyBlock proto era)
+        -- Due to UTxO-HD there is a subtlety here. The functions that register
+        -- the initial funds work on the UTxO inside the LedgerState instead of
+        -- on the tables. This implies that we have to first stow the tables,
+        -- modify the values and then unstow them. However because
+        -- 'unstowLedgerTables' expects an 'EmptyMK' state, we have to
+        -- 'forgetLedgerTables' just to make the types match.
+        --
+        -- It doesn't seem wise to modify the type of 'unstowLedgerTables' to
+        -- accept any 'mk' as now we get the guarantee that whenever we are
+        -- unstowing, we are doing it on an EmptyMK and only in this case (which
+        -- must happen only on tests) we do this trickery.
+        register (Flip st) = Flip $ unstowLedgerTables $ forgetLedgerTables $ st {
               Shelley.shelleyLedgerState =
                 -- We must first register the initial funds, because the stake
                 -- information depends on it.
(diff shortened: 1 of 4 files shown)


ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -816,17 +816,19 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
 
         register ::
              (EraCrypto era ~ c, ShelleyBasedEra era)
-          => Flip LedgerState mk (ShelleyBlock proto era)
-          -> Flip LedgerState mk (ShelleyBlock proto era)
-        register (Flip st) = Flip st {
+          => Flip LedgerState ValuesMK (ShelleyBlock proto era)
+          -> Flip LedgerState ValuesMK (ShelleyBlock proto era)
+        register (Flip st) = Flip $ unstowLedgerTables $ forgetLedgerTables $ st {
```
dnadales, line 821: I think what's happening here is not trivial, so I wonder if we should comment why do we do the `unstow;stow` sequence.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -239,10 +239,8 @@ instance ShelleyCompatible proto era => QueryLedger (ShelleyBlock proto era) whe
           getProposedPPUpdates st
         GetStakeDistribution ->
           SL.poolsByTotalStakeFraction globals st
-        GetUTxOByAddress addrs ->
-          SL.getFilteredUTxO st addrs
-        GetUTxOWhole -> -- FIXME:  if we never match this case then we should probaly clarify this to avoid confusions when reading this part of the code.
-          SL.getUTxO st
+        GetUTxOByAddress{} -> error "Unreachable"
```
dnadales, line 242: This can be reached if we define some instances. Maybe we should explain why we shouldn't reach this case.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
```diff
@@ -285,6 +285,8 @@ class QuerySat (mk :: MapKind) (fp :: FootprintL)
 
 instance QuerySat mk SmallL
 
+instance QuerySat EmptyMK LargeL
```
dnadales, line 288: I'm not sure about this. How can empty tables satisfy large query?

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
Jasagredo: Added a comment

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
Jasagredo: Rephrased the `error` message

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Query.hs
Jasagredo: Added a comment explaining it

dnadales (approved):

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -239,10 +239,10 @@ instance ShelleyCompatible proto era => QueryLedger (ShelleyBlock proto era) whe
           getProposedPPUpdates st
         GetStakeDistribution ->
           SL.poolsByTotalStakeFraction globals st
-        GetUTxOByAddress addrs ->
-          SL.getFilteredUTxO st addrs
-        GetUTxOWhole -> -- FIXME:  if we never match this case then we should probaly clarify this to avoid confusions when reading this part of the code.
-          SL.getUTxO st
+        GetUTxOByAddress{} ->
```
Jimbo4350, line 242: Does this impact the cli?

Jasagredo: The CLI will be impacted because of other things (now this query is quite expensive) but we are still working on this. This error call just indicates that this path should be unreachable, as a GetUTxOByAddress query should be handled by the `handleWholeQuery` handler instead.

PR #3918, opened 2022-07-22
Throw error when ledger db is incompatible

Jasagredo:
# Description

Previously, this would fail with an difficult to decipher error. Now it is much more understandable

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/BackingStore.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/BackingStore.hs
@@ -162,6 +162,16 @@ data TVarBackingStoreDeserialiseExn =
   deriving anyclass (Exn.Exception)
   deriving stock    (Show)
 
+newtype StoreDirIsIncompatible = StoreDirIsIncompatible FilePath
+  deriving anyclass (Exn.Exception)
+
+instance Show StoreDirIsIncompatible where
+  show (StoreDirIsIncompatible p) =
+       "In-Memory database not found in the database directory: "
+    <> show p
+    <> ".\nPre-UTxO-HD and LMDB implementations are incompatible with the In-Memory \
+       \ implementation. Please delete your ledger database directory."
+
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/BackingStore.hs
```diff
@@ -182,6 +182,14 @@ newTVarBackingStore lookup_ rangeRead_ forwardValues_ enc dec initialization = d
...
       (slot, values) <- case initialization of
         Left (FS.SomeHasFS fs, BackingStorePath path) -> do
+          dir <- FS.unsafeToFilePath fs path
+          tvarFileExists <- FS.doesFileExist fs (extendPath path)
+          unless
+            tvarFileExists
+            (error $ "In-Memory database not found in the database directory: "
+                   <> show dir
+                   <> ".\nPre-UTxO-HD and LMDB implementations are incompatible with the In-Memory \
+                      \ implementation. Please delete your ledger database directory." )
```
jorisdral, line 192: Possibly use `Exn.throw` with a custom datatype?

jorisdral (approved):
LGTM, but I'm wondering whether we should have a backingstore-independent approach that checks if a snapshot is of the desired backingstore-type before trying to initialise a store. In the current approach, each backing store implementation would have the responsibility of doing the check

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/BackingStore.hs
Jasagredo: Sounds reasonable

Jasagredo: Done!

Jasagredo:
> backingstore-independent approach that checks if a snapshot is of the desired backingstore-type before trying to initialise a store

Probably this would be the right way to go, indeed.

PR #3932, opened 2022-08-01
New query GetPoolDistr

newhoggy:
# Description

This introduce a new query GetPoolDistr which extracts the stake snapshot. This query is required to improve the CPU and memory efficiency of the query `leadership-schedule` command in the CLI.

# Checklist

- Branch
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

## `cardano-api` changes

There will be a new `GetPoolDistr` which uses a lot less CPU and memory:

```
GetPoolDistr
    :: SL.KeyHash 'SL.StakePool (EraCrypto era)
    -> BlockQuery (ShelleyBlock proto era)
                  (SL.PoolDistr (EraCrypto era))
```

This query takes a set of pool ids so it is possible to query multiple pools at once.

## `cardano-cli` changes

A existing command `query leadership-schedule` will be be functionally unchanged, but use much less CPU and memory.

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
@@ -210,6 +210,11 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
     -> BlockQuery (ShelleyBlock proto era)
                   (StakeSnapshots (EraCrypto era))
 
+  GetPoolDistr
+    :: Maybe (Set (SL.KeyHash 'SL.StakePool (EraCrypto era)))
+    -> BlockQuery (ShelleyBlock proto era)
+                  (SL.PoolDistr (EraCrypto era))
+
   -- WARNING: please add new queries to the end of the list and stick to this
   -- order in all other pattern matches on queries. This helps in particular
   -- with the en/decoders, as we want the CBOR tags to be ordered.
@@ -330,6 +335,12 @@ instance (ShelleyCompatible proto era, ProtoCrypto proto ~ crypto) => QueryLedge
                 , ssSetTotal       = getAllStake _pstakeSet
                 , ssGoTotal        = getAllStake _pstakeGo
                 }
+        GetPoolDistr mPoolIds ->
+          let poolDistr = SL.calculatePoolDistr . SL._pstakeSet . SL.esSnapshots $ getEpochState st in
+          case mPoolIds of
+            Just poolIds -> SL.PoolDistr $ Map.restrictKeys (SL.unPoolDistr poolDistr) poolIds
+            Nothing -> poolDistr
+
     where
       lcfg    = configLedger $ getExtLedgerCfg cfg
       globals = shelleyLedgerGlobals lcfg
@@ -451,6 +462,13 @@ instance SameDepIndex (BlockQuery (ShelleyBlock proto era)) where
     = Nothing
   sameDepIndex (GetStakeSnapshots _) _
     = Nothing
+  sameDepIndex (GetPoolDistr poolids) (GetPoolDistr poolids')
+    | poolids == poolids'
+    = Just Refl
+    | otherwise
+    = Nothing
+  sameDepIndex (GetPoolDistr _) _
+    = Nothing
 
 deriving instance Eq   (BlockQuery (ShelleyBlock proto era) result)
 deriving instance Show (BlockQuery (ShelleyBlock proto era) result)
@@ -478,6 +496,7 @@ instance ShelleyCompatible proto era => ShowQuery (BlockQuery (ShelleyBlock prot
       GetRewardInfoPools                         -> show
       GetPoolState {}                            -> show
       GetStakeSnapshots {}                       -> show
+      GetPoolDistr {}                            -> show
 
 -- | Is the given query supported by the given 'ShelleyNodeToClientVersion'?
 querySupportedVersion :: BlockQuery (ShelleyBlock proto era) result -> ShelleyNodeToClientVersion -> Bool
@@ -503,6 +522,7 @@ querySupportedVersion = \case
     GetRewardInfoPools                         -> (>= v5)
     GetPoolState {}                            -> (>= v6)
     GetStakeSnapshots {}                       -> (>= v6)
+    GetPoolDistr {}                            -> (>= v6)
     -- WARNING: when adding a new query, a new @ShelleyNodeToClientVersionX@
     -- must be added. See #2830 for a template on how to do this.
   where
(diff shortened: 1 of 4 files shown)


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -287,6 +297,8 @@ instance ShelleyCompatible proto era => QueryLedger (ShelleyBlock proto era) whe
       hst = headerState ext
       st  = shelleyLedgerState lst
 
+
```
nfrisby, line 300: minor: extra newlines

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -193,6 +193,11 @@ data instance BlockQuery (ShelleyBlock proto era) :: Type -> Type where
     -> BlockQuery (ShelleyBlock proto era)
                   (SL.PState (EraCrypto era))
 
+  GetPoolDistr
+    :: Maybe (Set (SL.KeyHash 'SL.StakePool (EraCrypto era)))
+    -> BlockQuery (ShelleyBlock proto era)
+                  (SL.PoolDistr (EraCrypto era))
```
nfrisby, line 199: @JaredCorduan @lehins I think I asked a similar question about `PState` on a recent&similar PR, but this time it's about `PoolDistr`:

Do we want to expose `PoolDistr` in the query API? That's promising some stability in it's interface.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -273,6 +278,11 @@ instance ShelleyCompatible proto era => QueryLedger (ShelleyBlock proto era) whe
                 , SL._retiring = Map.restrictKeys (SL._retiring dpsPState) poolIds
                 }
             Nothing -> dpsPState
+        GetPoolDistr mPoolIds ->
+          let poolDistr = SL.calculatePoolDistr . SL._pstakeSet . SL.esSnapshots $ getEpochState st in
```
nfrisby, line 282: Is this `calculatePoolDistr` perhaps too expensive?

nfrisby (changes requested):
I'm Requesting Changes, mostly to ensure that we have a good answer to whether `calculatePoolDistr` is prohibitively expensive.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
JaredCorduan: I think it is okay, but I do not know for sure.

Our definition of "big" (ie the data structures that we know will be finding a home one day on disk) is "at least the size of the number of registered stake credentials". The `PoolDistr` is a consolidation of the per-stake credential stake distribution, which is itself not "big", but is derived from a "big" structure. I am currently expecting that we will keep the pool distribution in memory, and that this query will not be disrupted. But it is intimately related to a "big" data structure that we do know will be disrupted.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
JaredCorduan: It might be, but I don't know that we have any clear requirements on what is too expensive. Note that `SL._pstakeSet` is one of the data structure that we anticipate being moved to disk.

nfrisby: This is for `query leadership-schedule`, so I think we have a good amount of headroom -- I wouldn't expect SPOs etc to be spamming this command (though it's easy enough to do that accidentally within a script).

newhoggy: Is there a way to make this cheaper if we're only interested in a subset of pool ids?

JaredCorduan: great question! The following might be faster:

Make a new version of `calculatePoolDistr` with signature
```
calculatePoolDistr' :: SnapShot crypto -> Set (KeyHash 'Staking crypto) -> PoolDistr crypto
```
which is identical to `calculatePoolDistr` except that after looking up the credential in the delegation map (see [here](https://github.com/input-output-hk/cardano-ledger/blob/dd4b6e31a05e3c02c6f00f0d53cf20e3b8b74468/eras/shelley/impl/src/Cardano/Ledger/Shelley/Rules/NewEpoch.hs#L214)), it also looks up the pool ID in the set of pool IDs provided by the user (and discards the result if not found).

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
```diff
@@ -273,6 +279,9 @@ instance ShelleyCompatible proto era => QueryLedger (ShelleyBlock proto era) whe
                 , SL._retiring = Map.restrictKeys (SL._retiring dpsPState) poolIds
                 }
             Nothing -> dpsPState
+        GetPoolDistr mPoolIds ->
+          let stakeSet = SL._pstakeSet . SL.esSnapshots $ getEpochState st in
+          SL.calculatePoolDistr' (maybe (const True) (flip Set.member) mPoolIds) stakeSet
```
newhoggy, line 284: Using the more efficient `calculatePoolDistr'` function.

newhoggy:
I reverted back to use `calculatePoolDistr` because some refactoring in `cardono-ledger` blocks merging the PR otherwise.  I would like to merge this version and in a future PR switch to use `calculatePoolDistr'`

ouroboros-network/src/Ouroboros/Network/NodeToClient/Version.hs
```diff
@@ -35,7 +35,7 @@ data NodeToClientVersion
     | NodeToClientV_13
     -- ^ enabled @CardanoNodeToClientVersion9@, i.e., Babbage
     | NodeToClientV_14
-    -- ^ added @GetPoolState, @GetSnapshots
+    -- ^ added @GetPoolDistr, @GetPoolState, @GetSnapshots
```
coot, line 38: :+1: `NodeToClientV_14` hasn't been yet released. `node-1.35.x` comes with `NodeToClientV_13` as the latest version.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Query.hs
kevinhammond: I'd look at use cases before optimising too much. I think the two likely cases are:

- all pools
- a specific pool (and then repeat for multi-pool operators)

The command is likely to be used occasionally rather than frequently, I expect

nfrisby (approved):
Marcin's commented that the NTC version was correct. So that's good.

The question of `calculatePoolDistr`'s performance remains unresolved, but I don't think it's a blocker for this PR. We'll need to assess it, though. @newhoggy can you pursue that with some measurements etc after merging this (or before, if that's convenient)?

newhoggy:
Yep.  I will evaluate after merging.

There is a [follow up PR](https://github.com/input-output-hk/ouroboros-network/pull/3949) that improves the performance, but that's blocked on a refactoring PR.

newhoggy:
I didn't notice much difference in CPU/memory usage of the new version of `cardano-node`.  I think it's because `query leadershipship-schedule` was already very expensive to start with.

PR #3951, opened 2022-08-12
Performant in-memory implementation

Jasagredo:
# Description

The InMemory implementation was based on using a `TVar` whose contents were then serialized from time to time. This implied keeping track of differences a la UTxO-HD. In fact, this is not necessary and the `InMemory` implementation can run without tables at all, in the same style as before UTxO-HD. This PR implements changes for achieving this.


ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -348,17 +351,24 @@ initLedgerDB replayTracer
              bss = do
     listSnapshots hasFS >>= tryNewestFirst id
   where
+
+    f :: m (ExtLedgerState blk wt ValuesMK)
+    f = case findOutWT (Proxy @wt) of
```
Jasagredo, line 356: Probably worth renaming this

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/LgrDB.hs
```diff
@@ -182,7 +188,6 @@ defaultArgs lgrHasFS diskPolicy bss = LgrDbArgs {
     , lgrGenesis        = NoDefault
     , lgrHasFS
     , lgrTopLevelConfig = NoDefault
-    , lgrTraceLedger    = nullTracer
```
Jasagredo, line 185: This was set to the null tracer and never ever customized.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/ChainSel.hs
```diff
@@ -782,7 +787,6 @@ chainSelectionForBlock cdb@CDB{..} blockCache hdr punish =
           mkTraceEvent events (mkNewTipInfo newLedger) curChain newChain
         whenJust (strictMaybeToMaybe prevTentativeHeader) $ traceWith $
           PipeliningEvent . OutdatedTentativeHeader >$< addBlockTracer
-        traceWith cdbTraceLedger newLedger
```
Jasagredo, line 785: No point in tracing always with the null tracer (plus tracing a whole LedgerDB is an overkill :thinking: )

PR #3955, opened 2022-08-16
Batch transactions in the mempool

Jasagredo:
# Description

_description of the pull request, if it fixes a particular issue it should link
the PR to a particular issue, see
[ref](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword=)_

# Checklist

- Branch
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested


Jasagredo:
Under a *VERY* specific situation we encounter the errors which we are looking for:
- Node adds tx `8922e396` which produces TxOuts `8922e396 (TxIx 0-10)`
- Node adds txs `18b1134e...d07cc4f2bf` which consume TxOuts `8922 (0-9)`
- Node adopts a block which contains tx `8922`
- Node resyncs the ledger state, revalidating `18b1134e...d07cc4f2bf`
- Node tries to add `a003b45` which consumes TxOut `8922 (10)`
- `getStatePair` reports the following:
```
The transactions to add are these: 
- `a003b45`: `[8922e3 (TxIx 10)]`

The transactions to keep are these: 

The keys needed by all of them are these:
- `[8922e3 (TxIx 10)]`

Current values and diffs in mempool:
- `CardanoLedgerTables [] []`
```
So:
- `18b1134e...d07cc4f2bf` which previously were valid, are now invalid
- Therefore TxOuts `8922 (0-9)` must be gone when resyncing?

The code for resyncing basically calls `getStatePair` and then runs the transactions. How could this be called and still get a wrong version of the changelog?

`implSyncWithLedger` is invoked by `forkSyncStateOnTipPointChange` when the result from this pseudo-call changes:
```
ledgerTipPoint . ledgerState <$> fmap LedgerDB.ledgerDbCurrent . LgrDB.getCurrent cdbLgrDB $ chainDB
```
And then `getStatePair` through `ChainDB.getLedgerStateForTxs` and `Query.getLedgerStateForKeys` does the following:
```
ldb0 <- atomically $ LgrDB.getCurrent cdbLgrDB
...
(a, ks) <- m (LedgerDB.ledgerDbCurrent ldb0)
```

So here comes the feared question. Might the dbchangelog be outdated when pushing blocks?

Investigating...

Jasagredo:
Tracing the `dbChangelog` that is set when doing chain selection shows that indeed the `dbChangelog` is correct (or at least it is not empty as I was expecting). A more important detail: adding tracers to the chain selection code (which slow down the code), results *in no tx rejection anymore*. It seems there is a race on the LedgerDB?

Jasagredo:
So after some time of investigation I can conclude the following in this regard:

The logic underneath seems correct. Every transaction that is reported as failing, is indeed already included in a block by the other node.

So it seems to be just a matter of performance and not of bugs. Need to re-evaluate the priorities after this revelation.

Issue #3689, opened 2022-04-04
ChainDB q-s-m failure with `GetIsValid`


amesgen:
### To reproduce

Run 
```
cabal run test-storage -- -p 'ChainDB q-s-m' --quickcheck-replay=455411
```
on fa10cb4eef1e7d3e095cec3c2bb1210774b7e5fa (master at time of writing). This yields
```
(PredicateC (Resp {getResp = Right (IsValid (IsValidResult {real = True, isValid = Just True}))}

:/=          Resp {getResp = Right (IsValid (IsValidResult {real = False, isValid = Nothing}))}))
```

This is exactly what #3651 ought to have fixed.

(edit NSF: just an update: it also fails on today's `master` tip, 03f935ec6, if you revert PR 3743, which disables this aspect of the test, before running with that seed)

### Cursory glance

The QSM commands look like this:
```
 - Add valid   block Genesis >: A with hash 2173152789646624309
 - Add valid   block A       >: B with hash 9122512639099877219
 - Add valid   block B       >: C with hash -7915818155603348863
 - Add invalid block C       >: D with hash -1393368485677491682
 - Add valid   block A       >: E with hash -2651776708921877814
 - Add invalid block E       >: F with hash 128720366063791916
 - Check if E is valid
    - ChainSel says: yes, valid
    - Model says:    unknown

Chain tree:
  Genesis >: A >: B >: C  :> D*
                  ∧    ∧
               >: E >: F*
* invalid
```

My interpretation: The model is wrong here, we have to validate `E` as the chain diff with rollback 2 and fragment `E >: F` is an improvement over our currently selected chain `A >: B >: C` as `C < F`. So even though we end up not selecting it as `F` is invalid, ChainSel still learns that `E` is valid.

We already have some asymmetry when comparing these validity results:

https://github.com/input-output-hk/ouroboros-network/blob/fa10cb4eef1e7d3e095cec3c2bb1210774b7e5fa/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs#L501-L504

Concretely, it is for the "other way around": when ChainSel does not know whether a block is valid, we intentionally don't check whether the model agrees.

### Full log

<details>
<summary>Click to expand</summary>

```
      ChainDB q-s-m
        sequential:
Model {
  dbModel = Model {
    volatileDbBlocks = Map.fromList
      [],
    immutableDbChain = Genesis,
    cps = ChainProducerState {
[...]

PR #3989, opened 2022-09-05
Add function `getAskBlockValidity` to ChainDB API to replace `getIsValid` and `getIsInvalidBlock`.

### LINKED ISSUE #3974: Enrich getIsInvalidBlock to subsume getIsValid
If `geIsInvalidBlock` could distinguish between "I don't know" and "definitely valid" (both of which are currently `Nothing`), then `getIsValid` (which seems to only be used in tests) would be redundant.

bartfrenk:
# Description

Extend ChainDB API with a function to replace both `getIsInvalidBlock` and `getIsValid`.

Closes #3974.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
@@ -9,7 +9,8 @@
 -- | Queries
 module Ouroboros.Consensus.Storage.ChainDB.Impl.Query (
     -- * Queries
-    getBlockComponent
+    getAskBlockValidity
+  , getBlockComponent
   , getCurrentChain
   , getIsFetched
   , getIsInvalidBlock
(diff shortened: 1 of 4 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/Impl/Query.hs
```diff
@@ -150,9 +152,26 @@ getIsInvalidBlock ::
...
+getAskBlockValidity CDB{..} = do
+  prevApplied <- LgrDB.getPrevApplied cdbLgrDB
+  invalid     <- readTVar cdbInvalid
+  let fn = \pt@(RealPoint _ hash) ->
+        case Map.lookup hash $ forgetFingerprint invalid of
+          Just invalidBlockInfo -> Just $ Invalid (invalidBlockReason invalidBlockInfo)
+          Nothing -> if Set.member pt prevApplied
+            then Just Valid
+            else Nothing
+  pure $ fmap (const fn) invalid
```
Jasagredo, line 172: So this `fmap :: (a -> b) -> WithFingerprint a -> WithFingerprint b` is the one thing that makes the fingerprint change when the `cdbInvalid` value changes.

Do we really wait on this? Or is the `StrictTVar m (WithFingerprint (InvalidBlocks blk))` even redundant and it could just be `StrictTVar m (InvalidBlocks blk)`?

Jasagredo (commented):
I think in general this is indeed the proposed way in #3689.1. Now all usages of the other functions should be replaced.

After this, there is still the question on whether we should still allow for discrepancies between the model and the system, but I guess that can go in another PR.

nfrisby:
In today's Planning call, we decided for Bart to try an alternative route: leaving the functions as-is and improving the documentation to better motivate that their shape (Bart had some doubts/concerns about this PR that could motivate leaving the code as-is).

PR #3992, opened 2022-09-07
Comment on why we should keep getIsInvalidBlock separate

### LINKED ISSUE #3974: Enrich getIsInvalidBlock to subsume getIsValid

bartfrenk:
Add comment why we shouldn't `getIsValid` into `getIsInvalidBlock` in the ChainDB API.

Closes #3974.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
@@ -311,15 +311,28 @@ data ChainDB m blk = ChainDB {
       -- Blocks unknown to the ChainDB will result in 'False'.
       --
       -- If the hash corresponds to a block that is known to be invalid, but
-      -- is now older than @k@, this function may return 'False'.
+      -- is now older than 'k', this function may return 'False'.
       --
-      -- Whenever a new invalid block is added, the @Fingerprint@ will be
+      -- Whenever a new invalid block is added, the 'Fingerprint' will be
       -- changed. This is useful when \"watching\" this function in a
       -- transaction.
       --
       -- Note that when invalid blocks are garbage collected and thus no
       -- longer detected by this function, the 'Fingerprint' doesn't have to
       -- change, since the function will not detect new invalid blocks.
+      --
+      -- It might seem natural to have this function also return whether the
+      -- ChainDB knows that a block is valid, thereby subsuming the 'getIsValid'
+      -- function and simplifying the API. However, this adds the overhead of
+      -- checking whether the block is valid for blocks that are not known to be
+      -- invalid that does not give useful information to current clients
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Util/STM.hs
```diff
@@ -38,7 +38,7 @@ import           Ouroboros.Consensus.Util.ResourceRegistry
   Misc
 -------------------------------------------------------------------------------}
 
--- | Wait until the TVar changed
+-- | Wait until a function of the TVar changed
```
Jasagredo, line 41: leftover? as it retries, it will wait until the TVars read by `getA` change, so I think it was correct previously

Jasagredo (approved):

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
```diff
@@ -306,6 +306,18 @@ data ChainDB m blk = ChainDB {
         -> BlockComponent blk b
         -> m (Follower m blk b)
 
+      -- It might seem natural to have this function also return whether the
```
nfrisby, line 309: Were you intending for this comment to be part of the Haddock or no? Right now it's just a plain comment, not Haddock.

Whether or not it's Haddock, I'd encourage using single quotes instead of the at-sign. You also used the Markdown backtick, and I think we'd prefer single quotes for that one too.

nfrisby: Would you also mention that, at least currently, only testing requires the more information predicate?

nfrisby (changes requested):
See my two requests below.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/ChainDB/API.hs
```diff
@@ -311,15 +311,28 @@ data ChainDB m blk = ChainDB {
       -- Blocks unknown to the ChainDB will result in 'False'.
       --
       -- If the hash corresponds to a block that is known to be invalid, but
-      -- is now older than @k@, this function may return 'False'.
+      -- is now older than 'k', this function may return 'False'.
```
nfrisby, line 314: I would leave this as `@k@`. `@` is for typewriter font (a la "`tt`"). `'` quotes are for top-level identifiers we expect to be in-scope. That's not the case here.

nfrisby (approved):
Thanks! Approved.

PR #3997, opened 2022-09-08
Anti-diff prototype for `SeqUtxoDiff` (broken up in parts).

### LINKED ISSUE #3946: Anti-diff prototype

jorisdral:
# Description

Resolves #3946. 

This PR breaks up #3882 into several commits that should be reviewed independently. The parts are based on the list described [in this PR comment](https://github.com/input-output-hk/ouroboros-network/pull/3882#issuecomment-1238102283):

## Part 1: Scaffolding for an anti-diff benchmarking setup. 

This commit defines scaffolding for a benchmarking setup for the anti-diff prototype, which should allow for running comparative benchmarks. 

In particular, this commit:
* Defines a top-level module comment describing the approach taken in the benchmark setup.
* Defines commands, which are used as inputs to the benchmarking setup.
* Defines a common interface to diff sequence implementations for the benchmarks to use, which allows straightforward plugging in of different implementations.
* Implements a "mocked", simplified version of a diff sequence.
* Sets up top-level definitions for generators that are used to generate inputs to the benchmarks.

## Part 2: Implement generators for benchmark inputs. 

This commits implements generators for commands, which are used as inputs to the benchmarking setup. 

In particular, this commit:
* Defines types of configuration parameters for the generators.
* Implements a default set of configuration parameters.
* Defines generators for each specific command type.
* Defines the generator that simulates the warmup phase.
* Defines a top-level generator that can be run in two modes: in `BenchMode`, commands are generated in a deterministic order, whereas in `TestMode`, commands are generated in a non-deterministic order. Furthermore, the generator will replace some of the configuration parameters by QuickCheck sized ones when running in `TestMode`.

## Part 3: Benchmark sub-cached (aka: "legacy") implementation

This commits hooks up the sub-cached (aka: "legacy") implementation to the benchmarking setup. 

In particular, this commit:
* Slightly alters the benchmarking setup to allow for comparative benchmarks, which compares the performance of the mock implementation to the sub-cached implementation.
* Implements sanity checks.
* Implements property tests.
* Defines an `Isomorphism` type class, which we use to define a mapping between sub-cached and mock diff sequences. This mapping is used to convert benchmark inputs that contain mock-types (i.e., types that are specific to mock diff sequences) to sub-cached-types (i.e., types that are specific to sub-cached diff sequences).

## Part 4: Implement semigroupoid package. 

This commit implements the semigroupoid package, which defines type classes for groupoid-like structures (i.e., semigroup, monoid, group, semigroupoid, groupoid). In addition, the package implements utilities for testing laws of groupoid-like structures (associativity, identity, etc.). We need these two things for Part 5. 

In particular, this commit:
* Defines a type class interface for semigroupoids and groupoids.
* Defines a type for automatically deriving groupoid-like type class instances from group-like class instances: semigroups are semigroupoids, and groups are groupoids.
* Defines tests for type class laws for the given type classes: semigroup, monoid, group, semigroupoid and groupoid.
* Tests these laws for a simple type.

##  Part 5: Implement anti-diffs and strict fingertrees with root measures.

This commit implements anti-diffs, and a variant of strict fingertrees that can store a "root measure": a measure stored only at the top level (or root) of the strict fingertree.

In particular, this commit:
* Defines strict fingertrees with root measures.
* Defines ``relatively performant'' split functions on strict fingertrees.
* Defines diff datatype for key-value stores, which can store a sequence of changes to specific key-value pairs.
* Implements functions for forwarding of key-value stores through diffs.
* Test groupoid-like structure laws for both datatypes.

## Part 6: Implement consensus-specific diff sequence based on anti-diffs and strict fingertrees with root measures.

This commit uses the datatypes from Part 5 to create a consensus-specific diff sequence.

In particular, this commit:
* Implement a consensus-specific diff sequence: a strict fingertree of key-value store diffs with a root measure that is the combined, total diff.
* Test groupoid-like structure laws for the consensus-specific diff sequence.

## Part 7: Hook up consensus-specific diff sequence to benchmarks.

We hook up consensus-specific diff sequences that are based on anti-diffs to the benchmark setup to obtain results about comparative performance. The results are listed in [this update post](https://github.com/input-output-hk/cardano-updates/blob/consensus/update-2022-08-31/posts/2022-08-31-consensus.markdown).

In particular, this commit:
* Removes mocked diff sequences entirely.
* Replace mocked diff sequences by anti-diff diff sequences in the benchmarking setup.
* Replace mocked diff sequences by anti-diff diff sequences in Isomorphism instances.
* Add more concrete configurations to run comparative benchmarks with.
* Slightly enhance sanity checks for benchmarks.

## Part 8: General cleanup.

This is a proposed commit that performs general cleanup, keeping only the changes that we want on the feature branch.

## Notes in response to PR comments

* The *sub-cached* implementation is now called the *intermediate-sum* implementation. See the module description of the benchmark module for a more elaborate description of the implementation and the naming choice.

## Epilogue: next steps

Possible follow-up PRs:
* Integration of new diff sequence implementation in consensus.
* Removal of "legacy" diff sequence implementation.
* What to do with the comparative benchmarks.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested


ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+
+  === Stateful generators
+
+  The stateful generators simulate how consensus manipulates diff sequences. The
+  top-level stateful generator runs in three parts:
+  * Generate an initial model.
+  * Run the warmup phase using the initial model.
+  * Generate a sequence of commands using the warmed up model.
+
+  The warmup phase ensures that we get past the point where the simulation is
```
dnadales, line 78: Why isn't the warmed-up diff seq part of the initial model?

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+  than @k@: before that, flushes will not have any effect. To be precise, the
+  warmup phase corresponds to running the simulation/generator that generates
+  Push commands for exactly @k@ steps and throwing away the commands we generate
+  in the process.
+
+  We use the commands that we generate after the warmup phase as inputs to the
+  benchmarks, as well as the state of the diff sequence right after the warmup
+  phase. Convince yourself that interpretation of these commands only make sense
+  with respect to the "warmed up" diff sequence.
+-}
```
dnadales, line 89: Great description :heart:

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+  The warmup phase ensures that we get past the point where the simulation is
+  still "warming up", i.e., when the length of the diff sequence is still less
+  than @k@: before that, flushes will not have any effect. To be precise, the
+  warmup phase corresponds to running the simulation/generator that generates
+  Push commands for exactly @k@ steps and throwing away the commands we generate
+  in the process.
+
+  We use the commands that we generate after the warmup phase as inputs to the
+  benchmarks, as well as the state of the diff sequence right after the warmup
+  phase. Convince yourself that interpretation of these commands only make sense
```
dnadales, line 87: Was the last sentence intended to be a `TODO`?

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+  with respect to the "warmed up" diff sequence.
+-}
+module Bench.Ouroboros.Consensus.Storage.LedgerDB.HD (benchmarks) where
+
+import           Control.DeepSeq
```
dnadales, line 92: Would it make sense to make these imports explicit?

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+
+import           QuickCheck.GenT
+import           Test.QuickCheck hiding (Result, choose, frequency, getSize,
+                     resize, shuffle, sized, variant)
+import           Test.Tasty.Bench
+import           Test.Tasty.QuickCheck hiding (choose, frequency, resize, sized,
+                     variant)
+
+import           Test.Util.MockDiffSeq
+
```
dnadales, line 116: Extra spaces?

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+    nrCommands        :: Int
+    -- | Security parameter @k@
+  , securityParameter :: Int
+  }
+
+{-------------------------------------------------------------------------------
+  Specific configurations
+-------------------------------------------------------------------------------}
+
+gcDefault :: GenConfig
```
dnadales, line 422: Do we want to align these fields?

ouroboros-consensus-test/ouroboros-consensus-test.cabal
```diff
@@ -317,3 +319,38 @@ test-suite test-infra
...
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+                       -Wpartial-fields
+                       -Widentities
+                       -Wredundant-constraints
+                       -Wmissing-export-lists
+                       -Wno-unticked-promoted-constructors
+                       -fno-ignore-asserts
+                       -rtsopts
+                       "-with-rtsopts=-A32m"
```
dnadales, line 360: Why did we set the allocation area to this? My guess: to prevent GC from interfering with the benchmarks. If so, did we confirm that this setting works? Maybe we should add a comment on this.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+-- | Simulate the warmup phase where there have not yet been at least @k@
+-- pushes.
+warmup :: CmdGen k v ()
+warmup = error "warmup: not implemented"
+
+{-------------------------------------------------------------------------------
+  Sanity checks
+-------------------------------------------------------------------------------}
+
+-- | Sanity check benchmark inputs.
```
dnadales, line 899: We need to comment on what sanity-checking the benchmarking environment means.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+-------------------------------------------------------------------------------}
+
+-- | Model for keeping track of state during generation of the command sequence.
+data Model k v = Model {
+    -- | The current diff sequence.
+    diffs         :: MockDiffSeq k v
+    -- | The tip of the diff sequence.
+  , tip           :: Int
+    -- | Values that have been flushed to a backing store.
+  , backingValues :: MockValues k v
```
dnadales: Would `flushedValues` describe this variable better?

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+{-------------------------------------------------------------------------------
+  Commands
+-------------------------------------------------------------------------------}
+
+type Block (ds :: DiffSeqKind) k v = (SlotNo ds, Diff ds k v)
+
+data Cmd (ds :: DiffSeqKind) k v =
+    Push     (SlotNo ds)     (Diff ds k v)
+  | Flush    Int
+  | Rollback Int             [Block ds k v]
```
dnadales, line 213: I wonder why we need the list of blocks to roll forward here.

ouroboros-consensus-test/src/Test/Util/MockDiffSeq.hs
```diff
@@ -0,0 +1,196 @@
...
+  }
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass NFData
+
+newtype MockLength = MockLength { unMockLength :: Int }
+  deriving stock (Show, Eq, Ord, Generic)
+  deriving newtype Num
+  deriving anyclass NFData
+
+newtype MockDiff k v = MockDiff (Map k (Seq (MockDiffEntry v)))
```
dnadales: Why do we need a `Seq` here? Isn't it sufficient to keep track of the last insertion or deletion of a key?

ouroboros-consensus-test/src/Test/Util/MockDiffSeq.hs
```diff
@@ -0,0 +1,196 @@
...
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE RankNTypes                 #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE StandaloneKindSignatures   #-}
+{-# LANGUAGE TupleSections              #-}
+
+module Test.Util.MockDiffSeq (
```
dnadales: Maybe we should explain what is the role of the mock diff in the benchmarks.

dnadales: My guess: it serves as the naïve implementation of diff sequences. But if I'm right, why do we need fingertrees at all :thinking:

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+  caches when we perform a split or append. In Consensus (post UTxO-HD), we
+  often split and append sequences of diffs: we split when we flush diffs or
+  roll back blocks, and we append when we push blocks or roll forward blocks.
+  The new /anti-diff/ implementation should reduce the overhead of this caching
+  since it only caches the total diff, and not the cumulative diffs of subtrees.
+  This total diff can be updated when appending by /summing/ diffs, whereas the
+  total diff can be updated when splitting by /subtracting/ diffs. As such, the
+  anti-diff implementation uses /Group/ properties of its definition of diffs.
+  The sub-cached implementation uses /Monoid/ properties of its definition of
+  diffs: it only uses summing to reconstruct a cumulative diff from other diffs.
```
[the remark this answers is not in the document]
dnadales: In other words, stating that the anti-diff operation is the inverse operation of the Group.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+
+  === What we measure
+
+  Diff sequences are manipulated by client code in Consensus. To mimick how
```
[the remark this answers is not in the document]
dnadales:
```haskell
-- We want to benchmark only the operations on sequence of differences. However,
-- these operations need to be consistent with the diff sequence to which they
-- are applied, for instance, we should only delete values that were previously
-- inserted in the diff-seq or we should only insert keys once. In other words,
-- there is an interdependency between the generated operations. For this
-- reason, commands are generated in a stateful manner, but we do not bechmkark
-- the generation process. The function that we actually measure the performance
-- ...
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
dnadales: Maybe we could simply remove this.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
dnadales: It seems we do, but maybe we could rename `Block` to something else that does  or remove the type synnonim.

```haskell
    | Switch Int (Pushes ds k v)
    -- ^ @Switch k ps@ switche to a different diff sequence by rolling back @k@ elements, 
    -- and then pushing the differences specified by `Pushes`.
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+  , securityParameter = 2160
+  }
+
+{-------------------------------------------------------------------------------
+  Command generator: Model
+-------------------------------------------------------------------------------}
+
+-- | Model for keeping track of state during generation of the command sequence.
+data Model k v = Model {
+    -- | The current diff sequence.
```
dnadales, line 494: We should double-check the use of this type of comment in the style guide.

ouroboros-consensus-test/src/Test/Util/MockDiffSeq.hs
dnadales: This is removed in latter commits.

ouroboros-consensus-test/src/Test/Util/MockDiffSeq.hs
dnadales: Not needed here, but it might be needed in the final prototype.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+      -- We do not @'deepseq'@ the right part of the rollback result, because we
+      -- are throwing it away in the case of a rollback.
+      Rollback n bs -> let (l, _r) = rollback n ds
+                       in  pushMany l bs
+      -- We @'deepseq'@ the result of forwarding values through a total diff,
+      -- since forwarding can be implementation-specific in (at least) two ways:
+      -- 1. How the total diff is computed.
+      -- 2. How the change to a specific key-value pair is computed.
+      Forward vs ks -> let vs' = forward vs ks ds
+                       in  vs' `deepseq` ds
```
dnadales, line 285: Maybe we should comment on the fact that in the actual system, the result of the forward operation will be used (for instance in block application), and we force its evaluation here to account for the costs of the forward operation.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -379,29 +450,331 @@ data CmdGenResult k v = CmdGenResult {
...
--- | Generate a list of commands.
-genCmds :: CmdGen k v [Cmd MockDiffSeq k v]
-genCmds = error "genCmds: not implemented"
+  where
+    gen = case gmode of
+      BenchMode -> genCmdsDet
+      TestMode  -> genCmdsNonDet
+
+    -- Replace configuration parameters by QuickCheck sized ones if the
+    -- generatore mode is set to @Test@.
```
dnadales, line 603: `generator`

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -379,29 +450,331 @@ data CmdGenResult k v = CmdGenResult {
...
+  nrg <- gets nrGenerated
+
+  if nrg >= nrc then
+    pure []
+  else do
+    dirsMay <- reader directives
+
+    case dirsMay of
+      Nothing       -> error "No directives given."
+      Just []       -> error "Directives should be infinitish length"
```
dnadales, line 636: I don't get why this is an error case.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,40 @@
...
+  , testGroupWithProxy (Proxy @(InternalMeasure X X)) [
+        testSemigroupLaws
+      , testMonoidLaws
+      ]
+  , testGroupWithProxy (Proxy @(Auto (InternalMeasure X X))) [
+        testSemigroupoidLaws
+      ]
+  ]
+
+type X = Small Int
```
dnadales, line 40: What if we omit the type synonim? Maybe it's better to remove this level of indirection. Besides `X` is not very descriptive :)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,281 @@
...
+  , tmDiff   :: !(Diff k v)
+  }
+  deriving stock (Generic, Show, Eq, Functor)
+  deriving anyclass (NoThunks)
+
+data InternalMeasure k v = InternalMeasure {
+    -- | Cumulative length
+    imLength :: {-# UNPACK #-} !Length
+    -- | Right-most slot number
+  , imSlotNo ::                !(Maybe SlotNo)
```
dnadales, line 91: Maybe we should explain what `Nothing` represents.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,281 @@
...
+
+data Element k v = Element {
+    elSlotNo :: {-# UNPACK #-} !SlotNo
+  , elDiff   ::                !(Diff k v)
+  }
+  deriving stock (Generic, Show, Eq, Functor)
+  deriving anyclass (NoThunks)
+
+
+-- | Length as a @'Sum'@.
```
dnadales, line 104: Would it make sense to explain what this length refers to? Eg

```haskell
Length of the sequence of differences
```

I don't get the "as a Sum" part. Does it mean that its Group instance is defined through the Group instance for `Sum`?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,281 @@
...
+import           Data.Bifunctor (Bifunctor (bimap))
+import           Data.Group
+import           Data.Monoid (Sum (..))
+import           Data.Semigroup (Max (..))
+import           GHC.Generics (Generic)
+import           NoThunks.Class (NoThunks)
+
+import           Data.FingerTree.RootMeasured.Strict hiding (split, splitl,
+                     splitlr, splitr)
+import qualified Data.FingerTree.RootMeasured.Strict as RMFT (splitlr)
```
dnadales, line 55: Do we need to explicitly import `splitlr` if we're using a qualified import?

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
```diff
@@ -0,0 +1,250 @@
...
+split p f (SFT vt sft) = (SFT vtLeft left, SFT vtRight right)
+  where
+    (left, right)     = FT.split p sft
+    (vtLeft, vtRight) = f vt (foldMap measureRoot left, foldMap measureRoot right)
+
+-- | Data type representing either /left/ or /right/.
+data LR = L | R
+  deriving (Show, Eq)
+
+-- | Like @'split'@, but we compute to-measures for /free/ trough subtraction of
```
dnadales, line 152: I don't understand the "for free" part. What does it mean?

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
```diff
@@ -0,0 +1,250 @@
...
+  -> ( StrictFingerTree vt vi a
+     , StrictFingerTree vt vi a
+     )
+split p f (SFT vt sft) = (SFT vtLeft left, SFT vtRight right)
+  where
+    (left, right)     = FT.split p sft
+    (vtLeft, vtRight) = f vt (foldMap measureRoot left, foldMap measureRoot right)
+
+-- | Data type representing either /left/ or /right/.
+data LR = L | R
```
dnadales, line 149: Do we need this and `splitlr`? Why not using `splitl` or `splitr` directly?

dnadales (commented):
We still need to:

- Update the CONTRIBUTING doc to describe these benchmarks.
- Write down the complexity of Data.FingerTree.RootMeasured.Strict.split

Other than that, I think we can merge this.

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,980 @@
...
+  than @k@: before that, flushes will not have any effect. To be precise, the
+  warmup phase corresponds to running the simulation/generator that generates
+  Push commands for exactly @k@ steps and throwing away the commands we generate
+  in the process.
+
+  We use the commands that we generate after the warmup phase as inputs to the
+  benchmarks, as well as the state of the diff sequence right after the warmup
+  phase.
+-}
+module Bench.Ouroboros.Consensus.Storage.LedgerDB.HD (benchmarks) where
```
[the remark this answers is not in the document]
jorisdral: Changed the module name to
```haskell
module Bench.Ouroboros.Consensus.Storage.LedgerDB.HD.DiffSeq
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -0,0 +1,412 @@
...
+
+-- | The kind of a diff sequence type that is parameterised by the type of keys
+-- and values respectively.
+--
+-- Note: Diffs in this context are diffs for key-value stores (a.k.a. @'Map'@),
+-- not diffs for arbitrary datatypes.
+type DiffSeqKind = Type -> Type -> Type
+
+-- | Common interface to diff sequence implementations.
+class IsDiffSeq (ds :: DiffSeqKind) k v where
```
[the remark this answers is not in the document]
jorisdral: If I understand correctly, the deps are "necessary" because `ds` is annotated with a kind. I do prefer the current defintion of the type class over one of these alternatives:
```haskell
class IsDiffSeq (ds :: Type) k v where
  type Diff ds k v = ...
  ...
  
class IsDiffSeq (ds :: Type) where
  type Diff ds k v = ...
  ...
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -291,8 +294,49 @@ data GenConfig = GenConfig {
...
+  , pushConfig        :: PushConfig
+  , forwardConfig     :: ForwardConfig
+  , rollbackConfig    :: RollbackConfig
+    -- | A series of directives that determines which commands to generate.
+  , directives        :: Maybe [Directive]
   }
 
+-- | Configuration parameters for @'Push'@ command generation
+data PushConfig = PushConfig {
+    -- | How many key-value pairs to delete
```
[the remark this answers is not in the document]
jorisdral: Yes, but only if `nrDeletes <= nrInitialValues` and `nrDeletes <= nrInserts`, or we could end up in a situation where we are trying to delete more keys than there are currently in `flushedValues` (in which case we delete every key from `flushedValues`).

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
jorisdral: Fixed

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
jorisdral: I hope this comment clarifies why:
```haskell
    -- | An infinite list of directives that determines which commands to
    -- generate.
    --
    -- If we run out of directives without generating @nrCommands@ commands, we
    -- would not know how to proceed generating more commands. Therefore, since
    -- @nrCommands@ can be an arbitrary number, we would want the number of
    -- directives to be /at least/ `nrCommands`.  To ensure that we never run
    -- out of directives, the safest bet is to make the list of directives
    -- infinite, for example by defining it to be an infinitely repeating
    -- series.
    --
    -- @Nothing@ signals that we do not want to use any directives. @Just []@
    -- should result in an error, for the reason we mentioned in the paragraph
    -- above.
  , directives        :: Maybe [Directive]
```

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
```diff
@@ -291,8 +294,49 @@ data GenConfig = GenConfig {
     nrCommands        :: Int
     -- | Security parameter @k@
   , securityParameter :: Int
+    -- | The number of initial backing values
+  , nrInitialValues   :: Int
+  , pushConfig        :: PushConfig
+  , forwardConfig     :: ForwardConfig
+  , rollbackConfig    :: RollbackConfig
+    -- | A series of directives that determines which commands to generate.
+  , directives        :: Maybe [Directive]
```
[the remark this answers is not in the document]
jorisdral: [See this comment](https://github.com/input-output-hk/ouroboros-network/pull/3997/commits/86179817e449cd94add4981a53b30c70e3a34291#r973115944)

ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD.hs
jorisdral: I also update the error description to be:
```haskell
      Just []       -> error "Unexpectedly ran out of directives."
```

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
jorisdral: When applying the `split` function, you have to pass it a function to compute the root measures. However, when we know that the root measures form a `Group`, then you get this function "for free", or in other words, you don't have to think about how to compute the root measures. I will try to clarify this in the documentation

jorisdral: I've slightly altered the `haddock` comment on the `split` function

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
jorisdral: It slightly decrease code duplication in downstream modules liked the `HD.DiffSeq` module in `ouroboros-consensus`. I will look into whether `splitlr` is absolutely necessary, or if there is a better alternative.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: It's not necessary, but all other definitions are imported unqualified and the qualified import is really only necessary for `splitlr` because this module also defines a `splitlr` function

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: -- | Right-most slot number
    --
    -- Empty diff sequences have no right-most slot number, so in that case
    -- @imSlotNo == Nothing@.
  , imSlotNo ::                !(Maybe SlotNo)

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: It means that `Length` is, with respect to group properties, equivalent to a `Sum`. This does not have to be mentioned in this `haddock` comment however, so I have removed it.

```haskell
-- | Length of a sequence of differences.
newtype Length = Length {unLength :: Int}
```

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
```diff
@@ -0,0 +1,251 @@
...
+-- be distinct from the internal measure. For example, we can not create both of
+-- these instances for distinct types @T@ and @T'@:
+--
+-- > @instance Measured T  a where -- ...@
+--
+-- > @instance Measured T' a where -- ...@
+--
+-- Furthermore, we want the root measure to be a @'Group'@ instead of a
+-- @'Monoid'@.
+class Monoid v => RootMeasured v a | a -> v where
```
Jasagredo, line 88: Isn't the comment saying that you would expect requiring `Group v => RootMeasured v a` rather than `Monoid v`?

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
```diff
@@ -0,0 +1,251 @@
...
+import qualified Data.FingerTree.Strict as FT
+
+{-------------------------------------------------------------------------------
+  Strict finger trees with root measures
+-------------------------------------------------------------------------------}
+
+-- | A @StrictFingerTree@ with elements of type @a@, an internal measure of type
+-- @vi@, and a root measure of type @vt@.
+data StrictFingerTree vt vi a = SFT {
+    tm       :: vt
```
Jasagredo, line 49: This is not `RootMeasure` right?
```suggestion
    rm       :: vt
```

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
```diff
@@ -0,0 +1,251 @@
...
+
+{-------------------------------------------------------------------------------
+  Construction
+-------------------------------------------------------------------------------}
+
+infixl 5 |>
+
+-- | /O(1)/. Add an element to the right end of a sequence.
+--
+-- Mnemonic: a triangle with the single element at the pointy end.
```
Jasagredo, line 106: I think these comments are taken directly from `Data.Sequence`. Are they still true? i.e. is this still in `O(1)`?

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
```diff
@@ -0,0 +1,251 @@
...
+(|>) ::
+     SuperMeasured vt vi a
+  => StrictFingerTree vt vi a
+  -> a
+  -> StrictFingerTree vt vi a
+SFT vt sft |> (!a) = SFT (vt <> measureRoot a) (sft FT.|> a)
+
+-- | /O(n)/. Create a sequence from a finite list of elements. The opposite
```
Jasagredo, line 114: Same as previous comment

anti-diff/anti-diff.cabal
```diff
@@ -0,0 +1,109 @@
...
+category:            Network
+build-type:          Simple
+cabal-version:       >=1.10
+
+library
+  hs-source-dirs:      src
+
+  exposed-modules:
+                       Data.FingerTree.RootMeasured.Strict
+                       Data.Map.Diff.Strict
```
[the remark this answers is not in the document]
jorisdral: Done

ouroboros-consensus-test/src/Test/Util/Orphans/Isomorphism.hs
```diff
@@ -0,0 +1,122 @@
...
+
+from :: Isomorphism b a => a -> b
+from = to
+
+isomorphismLaw ::
+     forall a b. (Isomorphism a b, Show a, Eq a)
+  => Proxy b
+  -> a
+  -> Property
+isomorphismLaw _ x = to @b @a (to @a @b x) === x
```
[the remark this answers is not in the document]
jorisdral: It is the type class law that instances of `Isomorphism` should abide (based on [the IsomorphicTo class](https://hackage.haskell.org/package/isomorphism-class-0.1/docs/IsomorphismClass.html#t:IsomorphicTo))

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,286 @@
...
+{-# LANGUAGE DeriveFunctor              #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingVia                #-}
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE TypeFamilies               #-}
+
```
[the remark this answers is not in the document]
jorisdral: Done

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,286 @@
...
+  DiffSeq
+    (StrictFingerTree
+      (RootMeasure k v)
+      (InternalMeasure k v)
+      (Element k v)
+    )
+  deriving stock (Generic, Show, Eq)
+  deriving newtype (Semigroup, Monoid)
+  deriving anyclass (NoThunks)
+
```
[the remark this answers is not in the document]
jorisdral: Done

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,286 @@
...
+
+-- The @'SlotNo'@ is not included in the root measure, since it is
+-- not a @'Group'@ instance.
+data RootMeasure k v = RootMeasure {
+    -- | Cumulative length
+    tmLength :: {-# UNPACK #-} !Length
```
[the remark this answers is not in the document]
jorisdral: Done

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,286 @@
...
+  deriving stock (Generic, Show, Eq, Functor)
+  deriving anyclass (NoThunks)
+
+data Element k v = Element {
+    elSlotNo :: {-# UNPACK #-} !SlotNo
+  , elDiff   ::                !(Diff k v)
+  }
+  deriving stock (Generic, Show, Eq, Functor)
+  deriving anyclass (NoThunks)
+
```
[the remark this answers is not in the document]
jorisdral: Done

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,286 @@
...
+
+-- | Length of a sequence of differences.
+newtype Length = Length {unLength :: Int}
```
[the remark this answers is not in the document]
jorisdral: Done

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,286 @@
...
+  deriving stock (Generic, Show, Eq, Ord)
+  deriving newtype (Num)
+  deriving anyclass (NoThunks)
+  deriving Semigroup via Sum Int
+  deriving Monoid via Sum Int
+  deriving Group via Sum Int
+
+-- | Right-most slot number.
+newtype SlotNo = SlotNo {unSlotNo :: Slot.SlotNo}
```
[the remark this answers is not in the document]
jorisdral: Done

semigroupoid/format-stylish.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+export LC_ALL=C.UTF-8
+# TODO the export of the <= operator TxLimits crashes stylish-haskell
```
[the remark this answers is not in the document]
jorisdral: Done

semigroupoid/src/Data/Semigroupoid/Auto.hs
```diff
@@ -0,0 +1,34 @@
...
+{-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE UndecidableInstances       #-}
+
+-- | Automatically derive type class instances for groupoid-like structures
+-- from type class instances for group-like structures.
+--
+-- This is a consequence of groupoid-like structures being generalised versions
+-- of group-like structures:
+-- * A semigroup is a semigroupoid.
+-- * A group is a groupoid.
```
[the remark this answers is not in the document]
jorisdral: Done

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,286 @@
...
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE TypeFamilies               #-}
+
+{-# LANGUAGE UndecidableInstances       #-}
+
+module Ouroboros.Consensus.Storage.LedgerDB.HD.DiffSeq (
```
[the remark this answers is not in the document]
jorisdral: I will add some more documentation to discuss the approach

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,286 @@
...
+{-------------------------------------------------------------------------------
+  Short-hands for type-class constraints
+-------------------------------------------------------------------------------}
+
+-- | Short-hand for @'SuperMeasured'@.
+type SM k v = SuperMeasured (RootMeasure k v) (InternalMeasure k v) (Element k v)
+
+{-------------------------------------------------------------------------------
+  API: derived functions
+-------------------------------------------------------------------------------}
```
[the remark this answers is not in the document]
jorisdral: I will add some more documentation about this as well

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -11,8 +11,65 @@
...
+
+   === Note on the use of @l@ and @r@ infixes in function names
+
+   Functions like @'splitl'@, @'splitlAt'@ and @'splitrAtFromEnd'@ use infix @l@
+   and @r@ characters to indicate that its time complexity is dependent on
+   either the length of the left part or right part of the split respectively.
+   This means that if we expect the left part of a split to be small relative to
+   the length of the input, then we should use an @l@-variant. If the reverse is
+   true, then we should use an @r@-variant.
+-}
```
jorisdral, line 72: @dnadales @Jasagredo what do you think of this module description?

Added this documentation in response to these two comments:
* https://github.com/input-output-hk/ouroboros-network/pull/3997#discussion_r973984244
* https://github.com/input-output-hk/ouroboros-network/pull/3997#discussion_r973983405

anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
```diff
@@ -105,147 +107,172 @@ infixl 5 |>
...
-     , StrictFingerTree vt vi a
+  -> StrictFingerTree vr vi a
+  -> ( StrictFingerTree vr vi a
+     , StrictFingerTree vr vi a
      )
-splitr p = split p f
-  where
-    f vt (_vtLeft, vtRight) = (vt <> invert vtRight, vtRight)
-
+splitr p = split p SplitRight
```
jorisdral, line 240: @Jasagredo @dnadales what do you think of this re-organisation of the split functions? The documentation should now also comment on the time complexity of these functions.

These changes were made in reponse to:
* https://github.com/input-output-hk/ouroboros-network/pull/3997#discussion_r972911366
* https://github.com/input-output-hk/ouroboros-network/pull/3997#discussion_r972909487

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
dnadales: Some comments:

- Do we need to export `SplitCustom`? Actually... do we need it at all? 
- I'm still struggling to understand why we need `SplitLeft` and `SplitRight`. What if we did not have these?
- Complexity analysis:
    - we need to explain what the variables are, eg `n`, `i`, etc.
    - does the `log` factor comes from `FT.split`? If so we might want to clarify this.

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,354 @@
...
+{-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE TypeFamilies               #-}
+
+{- | Sequences of diffs for ledger tables.
+
+   These diff sequences are an instantiation of a strict finger tree with root
+   measures. The tree/sequence itself contains diffs and slot information, while
```
dnadales, line 17: I don't know if we should refer to `slots` in this level of abstraction. What about something like "sequence numbers"?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,354 @@
...
+   split @ys@ and @zs respectively.
+
+   > splitAt 2 xs = (ys, zs) = ([1..2], [3..n])
+
+   How should we compute we the root measure @rmys@ of @ys@? Since @ys@ is
+   small, we can just compute @rmys = sum [1..2]@. How should we compute the
+   root measure @rmzs@ of @zs@? We should not compute @rmzs = sum [3..n]@ in this
+   case, since @n@ is large. Instead, we compute @rmzs = rmxs - rmys@, which
+   evaluates to its result in time that is linear in the length of @ys@, in this
+   case @O(2)@.
```
dnadales, line 48: `O(1)` actually, since the constant is irrelevant, isn't it?

dnadales (approved):

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,354 @@
...
+   operation to facilitate computing updated root measures.
+
+   === Note on the use of @l@ and @r@ infixes in function names
+
+   Functions like @'splitl'@, @'splitlAt'@ and @'splitrAtFromEnd'@ use infix @l@
+   and @r@ characters to indicate that its time complexity is dependent on
+   either the length of the left part or right part of the split respectively.
+   This means that if we expect the left part of a split to be small relative to
+   the length of the input, then we should use an @l@-variant. If the reverse is
+   true, then we should use an @r@-variant.
```
dnadales, line 71: Are there parts of the code that consider the length of the sequence to use either an `l` or an `r` variant?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -0,0 +1,354 @@
...
+   use the total sum of diffs nearly as often as we split or extend the diff
+   sequence, this proved to be too costly. The single-instance root measure
+   reduces the overhead of this "caching" of intermediate sums of diffs by only
+   using a single total sum of diffs, though augmented with an @'invert'@
+   operation to facilitate computing updated root measures.
+
+   === Note on the use of @l@ and @r@ infixes in function names
+
+   Functions like @'splitl'@, @'splitlAt'@ and @'splitrAtFromEnd'@ use infix @l@
+   and @r@ characters to indicate that its time complexity is dependent on
```
dnadales, line 67:
```suggestion
   and @r@ characters to indicate that its time complexity is dominated by
```

dnadales: Or "determined".

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: * Probably not, but I figured that it would not hurt, what do you think?
* If we use `splitl` and `splitr` directly, then we're writing the same logic twice in downstream modules. Example:
  ```haskell
  splitlAt ::
       SM k v
    => Int
    -> DiffSeq k v
    -> (DiffSeq k v, DiffSeq k v)
  splitlAt =  splitl ((Length n<) . imLength)
  
  splitrAt ::
       SM k v
    => Int
    -> DiffSeq k v
    -> (DiffSeq k v, DiffSeq k v)
  splitrAt = splitr  ((Length n<) . imLength)
  ```
  VS.
  ```haskell
  splitAt ::
       SM k v
    => SRM k v
    -> Int
    -> DiffSeq k v
    -> (DiffSeq k v, DiffSeq k v)
  splitAt f n = split f ((Length n<) . imLength)
  
  splitlAt ::
       SM k v
    => Int
    -> DiffSeq k v
    -> (DiffSeq k v, DiffSeq k v)
  splitlAt = splitAt SplitLeft
  
  splitrAt ::
       SM k v
    => Int
    -> DiffSeq k v
    -> (DiffSeq k v, DiffSeq k v)
  splitrAt = splitAt SplitRight
  ```
  In this case I think it's nice that we can write a general `splitAt` function that abstracts over `SplitRootMeasure`s, and   instantiate the function with specific `SplitRootMeasure`s
* I'll add the notes on complexity analysis to the section header

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: Not currently, but it would not surprise if that at some point we would move to using some heuristic to make the decision, instead of hard-coding which variant we use

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: The specific datatype does contain slot information, so I'm not sure if we can replace it -- Or do you mean to say that the diff sequence should abstract over different kinds of "sequence numbers", like slots?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: Done

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: Replaced by "determined"

PR #4009, opened 2022-09-16
ghc-9.2.4 compatibility

angerman:
This merges #4003 and #4001, and tries to look at the codebase with 9.2.4

I'm not sure the FlexibleContexts fix is proper.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Testing/ConcreteBlock.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Testing/ConcreteBlock.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE UndecidableInstances       #-}
+{-# LANGUAGE FlexibleContexts           #-}
 
 -- This module is for examples and tests (not the library) so orphans are ok
 {-# OPTIONS_GHC -Wno-orphans #-}
@@ -121,7 +122,7 @@ deriving instance Hashable BlockNo
 --
 -- This requires @UndecidableInstances@ because @Hashable (HeaderHash b)@
 -- is no smaller than @Hashable (ChainHash b)@.
-instance Hashable (HeaderHash b) => Hashable (ChainHash b)
+instance (StandardHash b, Hashable (HeaderHash b)) => Hashable (ChainHash b)
  -- use generic instance
 
 -- | The hash of all the information in a 'BlockHeader'.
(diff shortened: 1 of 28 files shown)


angerman:
For reference, this is the error 9.2.4 provides:
```
	src/Ouroboros/Network/Testing/ConcreteBlock.hs:124:10: error:
	    • Non type-variable argument
	        in the constraint: Hashable (HeaderHash b)
	      (Use FlexibleContexts to permit this)
	    • In the instance declaration for ‘Hashable (ChainHash b)’
	    |
	124 | instance Hashable (HeaderHash b) => Hashable (ChainHash b)
	    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
following the `FlexibleContexts` suggestion, GHC then suggests:
```
src/Ouroboros/Network/Testing/ConcreteBlock.hs:125:10: error:
    • Could not deduce (StandardHash b)
        arising from the superclasses of an instance declaration
      from the context: Hashable (HeaderHash b)
        bound by the instance declaration
        at src/Ouroboros/Network/Testing/ConcreteBlock.hs:125:10-58
      Possible fix:
        add (StandardHash b) to the context of the instance declaration
    • In the instance declaration for ‘Hashable (ChainHash b)’
    |
125 | instance Hashable (HeaderHash b) => Hashable (ChainHash b)
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
which leaves us with this fix, but that feels rather fishy.

ouroboros-network/src/Ouroboros/Network/Testing/ConcreteBlock.hs
```diff
@@ -4,6 +4,7 @@
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE UndecidableInstances       #-}
+{-# LANGUAGE FlexibleContexts           #-}
```
amesgen, line 7: This is due to a bug fix in 9.2:

> Previously, `-XUndecidableInstances` accidentally implied `-XFlexibleContexts`. This is now fixed, but it means that some programs will newly require `-XFlexibleContexts`.

https://downloads.haskell.org/~ghc/9.2.4/docs/html/users_guide/9.2.1-notes.html

angerman: Thanks for pointing this out!

PR #4014, opened 2022-09-19
Draft ledger ops benchmark in db-analyser

nfrisby:
This PR is draft because

- it's experimental
- it targets an old commit; I using all upstream commits that the `master` tip of `cardano-node` refers to

### DIFF
--- a/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
+++ b/ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
@@ -82,9 +82,12 @@ executable db-analyser
                      , cardano-crypto-class
                      , cardano-crypto-wrapper
                      , cardano-ledger-alonzo
+                     , cardano-ledger-babbage
                      , cardano-ledger-byron
                      , cardano-ledger-core
                      , cardano-ledger-shelley
+                     , cardano-ledger-shelley-ma
+                     , cardano-slotting
                      , cborg
                      , containers
                      , contra-tracer
--- a/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
+++ b/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
@@ -450,7 +450,7 @@ traceLedgerProcessing
   for each block
 -------------------------------------------------------------------------------}
 
-data St blk = St !IOLike.Time !GC.RTSStats !(ExtLedgerState blk)
+data St blk = St {- !IOLike.Time -} !GC.RTSStats !(ExtLedgerState blk)
 
 infixl :&
 pattern (:&) :: a -> b -> (a, b)
@@ -468,16 +468,18 @@ benchmarkLedgerOps
     IO.hSetBuffering IO.stdout (IO.BlockBuffering $ Just $ 4*1024*1024)   -- useful?
     putStrLn $ unwords $ reverse $ "...era-specific stats" : theHeadings
 
-    prevt5 <- IOLike.getMonotonicTime
     rtsStats <- GC.getRTSStats
-    let st0 = St prevt5 rtsStats initLedger
+    let st0 = St rtsStats initLedger
 
     void $ processAll db registry GetBlock initLedger limit st0 process
   where
     ccfg = topLevelConfigProtocol cfg
     lcfg = topLevelConfigLedger   cfg
 
-    showTimeDiff t' t = show (fromEnum (t' `IOLike.diffTime` t) `div` 1000000)
+    _showTimeDiff t' t = show $ fromEnum (t' `IOLike.diffTime` t) `div` 1000000
+
+    -- nanoseconds from RTS
+    showNsDiff t' t = show $ (t' - t) `div` 1000
 
     (theHeadings, theCells) =
         let infixl `o`
@@ -485,14 +487,15 @@ benchmarkLedgerOps
               ( heading : headings
               , \(ts, t') ->
                   let (cells, t) = acc ts
-                  in (showTimeDiff t' t : cells, t')
+--                  in (showTimeDiff t' t : cells, t')
+                  in (showNsDiff t' t : cells, t')
               )
-            acc0 (                                 rp :& slotGap :& whole :&    majGcCount :&    gc_us, t0) =
-                ( reverse [condense (realPointSlot rp),  slotGap,   whole, show majGcCount, show gc_us]
-                , t0
+            acc0 (                                 rp :& slotGap :& both :& mut :& gc :& majGcCount, t') =
+                ( reverse [condense (realPointSlot rp),  slotGap,   both,   mut,   gc,   majGcCount]
+                , t'
                 )
         in
-        (reverse [                              "slot", "slotGap", "whole",    "majGcCount",    "gc"], acc0)
+        (reverse [                              "slot", "slotGap", "both", "mut", "gc", "majGcCount"], acc0)
           `o` "forecast"
           `o` "headerTick"
           `o` "headerApply"
@@ -505,11 +508,11 @@ benchmarkLedgerOps
         Slotting.At (SlotNo j) -> i - j
 
     process :: St blk -> blk -> IO (St blk)
-    process (St prevt5 prevRtsStats extLdgrSt) blk = do
+    process (St prevRtsStats extLdgrSt) blk = do
         let slot = blockSlot      blk
             rp   = blockRealPoint blk
 
-        t0 <- IOLike.getMonotonicTime
+        t0 <- GC.mutator_elapsed_ns <$> GC.getRTSStats
 
         -- forecast the LedgerView
         let forecaster = ledgerViewForecastAt lcfg (ledgerState extLdgrSt)
@@ -517,40 +520,47 @@ benchmarkLedgerOps
             Left err -> fail $ "benchmark doesn't support headers beyond the forecast limit: " <> show rp <> " " <> show err
             Right !x -> pure x
 
-        t1 <- IOLike.getMonotonicTime
+        t1 <- GC.mutator_elapsed_ns <$> GC.getRTSStats
 
         -- tick the HeaderState
         let !tickedHeaderState = tickHeaderState ccfg tickedLedgerView slot (headerState extLdgrSt)
 
-        t2 <- IOLike.getMonotonicTime
+        t2 <- GC.mutator_elapsed_ns <$> GC.getRTSStats
 
         -- apply the header
         !headerState' <- case runExcept $ validateHeader cfg tickedLedgerView (getHeader blk) tickedHeaderState of
             Left err -> fail $ "benchmark doesn't support invalid headers: " <> show rp <> " " <> show err
             Right x -> pure x
 
-        t3 <- IOLike.getMonotonicTime
+        t3 <- GC.mutator_elapsed_ns <$> GC.getRTSStats
 
         -- tick the ledger state
         let !tickedLdgrSt = applyChainTick lcfg slot (ledgerState extLdgrSt)
 
-        t4 <- IOLike.getMonotonicTime
+        t4 <- GC.mutator_elapsed_ns <$> GC.getRTSStats
 
         -- apply the block
         !ldgrSt' <- case runExcept (lrResult <$> applyBlockLedgerResult lcfg blk tickedLdgrSt) of
             Left err -> fail $ "benchmark doesn't support invalid blocks: " <> show rp <> " " <> show err
             Right x -> pure x
 
-        t5 <- IOLike.getMonotonicTime
-
-        rtsStats <- GC.getRTSStats
+        rtsStats5 <- GC.getRTSStats
+        let t5 = GC.mutator_elapsed_ns rtsStats5
 
         putStrLn
           $ unwords
-          $ reverse (fst $ theCells $ rp :& (slot `slotCount` getTipSlot extLdgrSt) :& showTimeDiff t5 prevt5 :& ((GC.gc_elapsed_ns rtsStats - GC.gc_elapsed_ns prevRtsStats) `div` 1000) :& (GC.major_gcs rtsStats - GC.major_gcs prevRtsStats) :& t0 :& t1 :& t2 :& t3 :& t4 :& t5)
+          $ reverse (fst $ theCells $
+                           rp
+                        :& (slot `slotCount` getTipSlot extLdgrSt)
+                        :& showNsDiff (GC.elapsed_ns rtsStats5) (GC.elapsed_ns prevRtsStats)
+                        :& showNsDiff t5 (GC.mutator_elapsed_ns prevRtsStats)
+                        :& showNsDiff (GC.gc_elapsed_ns rtsStats5) (GC.gc_elapsed_ns prevRtsStats)
+                        :& show (GC.major_gcs rtsStats5 - GC.major_gcs prevRtsStats)
+                        :& t0 :& t1 :& t2 :& t3 :& t4 :& t5
+                    )
             ++ HasAnalysis.blockStats blk
 
-        pure $ St t5 rtsStats $ ExtLedgerState ldgrSt' headerState'
+        pure $ St rtsStats5 $ ExtLedgerState ldgrSt' headerState'
 
 {-------------------------------------------------------------------------------
   Analysis: reforge the blocks, via the mempool
--- a/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
+++ b/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
@@ -3,10 +3,12 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE PatternSynonyms     #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections       #-}
 {-# LANGUAGE TypeApplications    #-}
+
 module Analysis (
     AnalysisEnv (..)
   , AnalysisName (..)
@@ -21,16 +23,22 @@ import           Data.List (intercalate)
 import qualified Data.Map.Strict as Map
 import           Data.Word (Word16, Word64)
 import qualified Debug.Trace as Debug
+import qualified GHC.Stats as GC
 import           NoThunks.Class (noThunks)
+import qualified System.IO as IO
 
+import qualified Cardano.Slotting.Slot as Slotting
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Config
+import           Ouroboros.Consensus.Forecast (forecastFor)
 import           Ouroboros.Consensus.HeaderValidation (HasAnnTip (..),
-                     HeaderState (..), annTipPoint)
+                     HeaderState (..), annTipPoint, tickHeaderState,
+                     validateHeader)
 import           Ouroboros.Consensus.Ledger.Abstract (LedgerCfg, LedgerConfig,
-                     applyChainTick, tickThenApplyLedgerResult, tickThenReapply)
+                     applyBlockLedgerResult, applyChainTick,
+                     tickThenApplyLedgerResult, tickThenReapply)
 import           Ouroboros.Consensus.Ledger.Basics (LedgerResult (..),
-                     LedgerState)
+                     LedgerState, getTipSlot)
 import           Ouroboros.Consensus.Ledger.Extended
 import           Ouroboros.Consensus.Ledger.SupportsMempool
                      (LedgerSupportsMempool)
@@ -43,6 +51,7 @@ import qualified Ouroboros.Consensus.Mempool.TxSeq as MP
 import           Ouroboros.Consensus.Storage.Common (BlockComponent (..),
                      StreamFrom (..))
 import           Ouroboros.Consensus.Storage.FS.API (SomeHasFS (..))
+import           Ouroboros.Consensus.Util.Condense (condense)
 import qualified Ouroboros.Consensus.Util.IOLike as IOLike
 import           Ouroboros.Consensus.Util.ResourceRegistry
 
@@ -75,6 +84,7 @@ data AnalysisName =
   | CountBlocks
   | CheckNoThunksEvery Word64
   | TraceLedgerProcessing
+  | BenchmarkLedgerOps
   | ReproMempoolAndForge Int
   deriving Show
 
@@ -103,6 +113,7 @@ runAnalysis analysisName env@(AnalysisEnv { tracer }) = do
     go CountBlocks                 = countBlocks env
     go (CheckNoThunksEvery nBks)   = checkNoThunksEvery nBks env
     go TraceLedgerProcessing       = traceLedgerProcessing env
+    go BenchmarkLedgerOps          = benchmarkLedgerOps env
     go (ReproMempoolAndForge nBks) = reproMempoolForge nBks env
 
 type Analysis blk = AnalysisEnv IO blk -> IO ()
@@ -433,6 +444,113 @@ traceLedgerProcessing
               HasAnalysis.WithLedgerState blk (ledgerState oldLedger) (ledgerState newLedger))
       mapM_ Debug.traceMarkerIO traces
       return $ newLedger
+
+{-------------------------------------------------------------------------------
+  Analysis: maintain a ledger state and time the five major ledger calculations
+  for each block
+-------------------------------------------------------------------------------}
+
+data St blk = St !IOLike.Time !GC.RTSStats !(ExtLedgerState blk)
+
+infixl :&
+pattern (:&) :: a -> b -> (a, b)
+pattern x :& y = (x, y)
+{-# COMPLETE (:&) #-}
+
+benchmarkLedgerOps ::
+  forall blk.
+     ( HasAnalysis blk
+     , LedgerSupportsProtocol blk
+     )
+  => Analysis blk
+benchmarkLedgerOps
+  (AnalysisEnv {db, registry, initLedger, cfg, limit}) = do
+    IO.hSetBuffering IO.stdout (IO.BlockBuffering $ Just $ 4*1024*1024)   -- useful?
+    putStrLn $ unwords $ reverse $ "...era-specific stats" : theHeadings
+
+    prevt5 <- IOLike.getMonotonicTime
+    rtsStats <- GC.getRTSStats
+    let st0 = St prevt5 rtsStats initLedger
+
+    void $ processAll db registry GetBlock initLedger limit st0 process
+  where
+    ccfg = topLevelConfigProtocol cfg
+    lcfg = topLevelConfigLedger   cfg
+
+    showTimeDiff t' t = show (fromEnum (t' `IOLike.diffTime` t) `div` 1000000)
+
+    (theHeadings, theCells) =
+        let infixl `o`
+            (headings, acc) `o` heading =
+              ( heading : headings
+              , \(ts, t') ->
+                  let (cells, t) = acc ts
+                  in (showTimeDiff t' t : cells, t')
+              )
+            acc0 (                                 rp :& slotGap :& whole :&    majGcCount :&    gc_us, t0) =
+                ( reverse [condense (realPointSlot rp),  slotGap,   whole, show majGcCount, show gc_us]
+                , t0
+                )
+        in
+        (reverse [                              "slot", "slotGap", "whole",    "majGcCount",    "gc"], acc0)
+          `o` "forecast"
+          `o` "headerTick"
+          `o` "headerApply"
+          `o` "blockTick"
+          `o` "blockApply"
+
+    slotCount :: SlotNo -> WithOrigin SlotNo -> String
+    slotCount (SlotNo i) = show . \case
+        Slotting.Origin        -> i
+        Slotting.At (SlotNo j) -> i - j
+
+    process :: St blk -> blk -> IO (St blk)
+    process (St prevt5 prevRtsStats extLdgrSt) blk = do
+        let slot = blockSlot      blk
+            rp   = blockRealPoint blk
+
+        t0 <- IOLike.getMonotonicTime
+
+        -- forecast the LedgerView
+        let forecaster = ledgerViewForecastAt lcfg (ledgerState extLdgrSt)
+        !tickedLedgerView <- case runExcept $ forecastFor forecaster slot of
+            Left err -> fail $ "benchmark doesn't support headers beyond the forecast limit: " <> show rp <> " " <> show err
+            Right !x -> pure x
+
+        t1 <- IOLike.getMonotonicTime
+
+        -- tick the HeaderState
+        let !tickedHeaderState = tickHeaderState ccfg tickedLedgerView slot (headerState extLdgrSt)
+
+        t2 <- IOLike.getMonotonicTime
+
+        -- apply the header
+        !headerState' <- case runExcept $ validateHeader cfg tickedLedgerView (getHeader blk) tickedHeaderState of
+            Left err -> fail $ "benchmark doesn't support invalid headers: " <> show rp <> " " <> show err
+            Right x -> pure x
+
+        t3 <- IOLike.getMonotonicTime
+
+        -- tick the ledger state
+        let !tickedLdgrSt = applyChainTick lcfg slot (ledgerState extLdgrSt)
+
+        t4 <- IOLike.getMonotonicTime
+
+        -- apply the block
+        !ldgrSt' <- case runExcept (lrResult <$> applyBlockLedgerResult lcfg blk tickedLdgrSt) of
+            Left err -> fail $ "benchmark doesn't support invalid blocks: " <> show rp <> " " <> show err
+            Right x -> pure x
+
+        t5 <- IOLike.getMonotonicTime
+
+        rtsStats <- GC.getRTSStats
+
+        putStrLn
+          $ unwords
+          $ reverse (fst $ theCells $ rp :& (slot `slotCount` getTipSlot extLdgrSt) :& showTimeDiff t5 prevt5 :& ((GC.gc_elapsed_ns rtsStats - GC.gc_elapsed_ns prevRtsStats) `div` 1000) :& (GC.major_gcs rtsStats - GC.major_gcs prevRtsStats) :& t0 :& t1 :& t2 :& t3 :& t4 :& t5)
+            ++ HasAnalysis.blockStats blk
+
+        pure $ St t5 rtsStats $ ExtLedgerState ldgrSt' headerState'
 
 {-------------------------------------------------------------------------------
   Analysis: reforge the blocks, via the mempool
--- a/ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
+++ b/ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
@@ -27,6 +27,7 @@ import           Cardano.Crypto (RequiresNetworkMagic (..))
 import qualified Cardano.Crypto as Crypto
 import qualified Cardano.Crypto.Hash.Class as CryptoClass
 import qualified Cardano.Ledger.Alonzo.Genesis as SL (AlonzoGenesis)
+import qualified Cardano.Ledger.Alonzo.Tx (ValidatedTx(..))
 import           Cardano.Ledger.Crypto
 import qualified Cardano.Ledger.Era as Core
 import           Control.Monad (when)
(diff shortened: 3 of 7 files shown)


nfrisby:
I invoke the tool like this:

```
$(cabal list-bin exe:db-analyser) --db state-node-mainnet/db-mainnet/ --only-immutable-db cardano --config mainnet-config.json --analyse-from 42942407 --benchmark-ledger-ops 2>/dev/null 1>db-analyser.log
```

and then feed it to this Gnuplot script

```
set terminal png size 1600,800
set border back
set output "bench-ledger-ops.png"

set xlabel "slot"
set ylabel "dur (μs)"

set multiplot layout 2,1

plot 'db-analyser.log' using 1:3 with impulses title "both",\
     'db-analyser.log' using 1:4 title "mut",\
     'db-analyser.log' using 1:5 title "gc"

plot 'db-analyser.log' using 1:4 with impulses title "mut",\
     'db-analyser.log' using 1:7 title "forcst",\
     'db-analyser.log' using 1:8 title "hdrTik",\
     'db-analyser.log' using 1:9 title "hdrApp",\
     'db-analyser.log' using 1:10 title "blkTik",\
     'db-analyser.log' using 1:11 title "blkApp" 

unset multiplot
```

to yield the following plots

bench-ledger-ops

The x-axis is the slot number of the block, the y-axis is duration in *microseconds*. Legend:

- `both` means `elapsed_ns` from https://hackage.haskell.org/package/base-4.17.0.0/docs/GHC-Stats.html
- `mut` means `mutator_elapsed_ns`
- `gc` means `gc_elapsed_ns`
- the second plot series corresponds to the steps in code being instrumented:

https://github.com/input-output-hk/ouroboros-network/blob/3961b4d6bce2b90a95df22fd0226e2873638a758/ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs#L515-L548

nfrisby:
bench-ledger-ops

nfrisby:
Here's the entire history. The only thing you might add to this plot is the Byron epoch delimiters---we chose not too because they'd be very dense.

bench-ledger-ops

nfrisby:
And the same plot, for just the most youngest100k blocks.
bench-ledger-ops-most-recent-100k-blocks

nfrisby:
And finally for just the youngest 500 blocks
bench-ledger-ops-most-recent-500-blocks

nfrisby:
Here's my latest gnuplot script, which I invoke like this `nix-shell -p gnuplot --run "gnuplot -e \"benchmarkLedgerOpsLogFilePath='db-analyser7.log'\" gp.scr"`

```
set terminal pngcairo size 1600,800
set border back
set output "bench-ledger-ops.png"

set xlabel "slot"
set ylabel "dur (μs)"

set multiplot layout 2,1

plot benchmarkLedgerOpsLogFilePath using 1:3 with impulses title "both",\
     benchmarkLedgerOpsLogFilePath using 1:4 title "mut",\
     benchmarkLedgerOpsLogFilePath using 1:5 title "gc"

# draw a vertical line at each every Shelley boundary
do for [sl in "4492800 4924800 5356800 5788800 6220800 6652800 7084800 7516800 7948800 8380800 8812800 9244800 9676800 10108800 10540800 10972800 11404800 11836800 12268800 12700800 13132800 13564800 13996800 14428800 14860800 15292800 15724800 16156800 16588800 17020800 17452800 17884800 18316800 18748800 19180800 19612800 20044800 20476800 20908800 21340800 21772800 22204800 22636800 23068800 23500800 23932800 24364800 24796800 25228800 25660800 26092800 26524800 26956800 27388800 27820800 28252800 28684800 29116800 29548800 29980800 30412800 30844800 31276800 31708800 32140800 32572800 33004800 33436800 33868800 34300800 34732800 35164800 35596800 36028800 36460800 36892800 37324800 37756800 38188800 38620800 39052800 39484800 39916800 40348800 40780800 41212800 41644800 42076800 42508800 42940800 43372800 43804800 44236800 44668800 45100800 45532800 45964800 46396800 46828800 47260800 47692800 48124800 48556800 48988800 49420800 49852800 50284800 50716800 51148800 51580800 52012800 52444800 52876800 53308800 53740800 54172800 54604800 55036800 55468800 55900800 56332800 56764800 57196800 57628800 58060800 58492800 58924800 59356800 59788800 60220800 60652800 61084800 61516800 61948800 62380800 62812800 63244800 63676800 64108800 64540800 64972800 65404800 65836800 66268800 66700800 67132800 67564800 67996800 68428800 68860800 69292800 69724800 70156800 70588800 71020800 71452800 71884800 72316800 72748800 73180800 73612800 74044800 74476800 74908800 75340800 75772800 76204800 76636800 77068800 77500800 77932800 78364800 78796800 79228800 79660800 80092800 80524800 80956800 81388800 81820800 82252800 82684800 83116800 83548800 83980800 84412800 84844800 85276800 85708800 86140800 86572800 87004800 87436800"] {
  set arrow from sl, screen 0.5 to sl, screen 0.6 nohead ls 28
  set arrow from sl, screen 0.0 to sl, screen 0.1 nohead ls 28
}

# draw a heavier line for the epoch transitions that are also HFC era transitions
do for [sl in "4492800 16588800 23068800 39916800"] {
  set arrow from sl, screen 0.5 to sl, screen 0.575 nohead ls 28 linewidth 5
  set arrow from sl, screen 0.0 to sl, screen 0.075 nohead ls 28 linewidth 5
}

plot benchmarkLedgerOpsLogFilePath using 1:4 with impulses title "mut",\
     benchmarkLedgerOpsLogFilePath using 1:7 title "forcst",\
     benchmarkLedgerOpsLogFilePath using 1:8 title "hdrTik",\
     benchmarkLedgerOpsLogFilePath using 1:9 title "hdrApp",\
     benchmarkLedgerOpsLogFilePath using 1:10 title "blkTik",\
     benchmarkLedgerOpsLogFilePath using 1:11 title "blkApp" 

unset multiplot

# the first shelley epoch was 208, and it began in slot 4492800
# there are 10k/f = 2160 * 10 * 20 = 432000 slots in each Shelley epoch

# these are the first epoches of each ledger era
# (see https://github.com/input-output-hk/cardano-ledger/wiki/First-Block-of-Each-Era)
# Byron		0
# Shelley	208
# Allegra 	236
# Mary	 	251
# Alonzo 	290
# Alonzo' 	298

# these are the first slots of each Shelley epoch
# 208 4492800
# 209 4924800
# 210 5356800
# 211 5788800
# 212 6220800
# 213 6652800
# 214 7084800
# 215 7516800
# 216 7948800
# 217 8380800
# 218 8812800
# 219 9244800
# 220 9676800
# 221 10108800
# 222 10540800
# 223 10972800
# 224 11404800
# 225 11836800
# 226 12268800
# 227 12700800
# 228 13132800
# 229 13564800
# 230 13996800
# 231 14428800
# 232 14860800
# 233 15292800
# 234 15724800
# 235 16156800
# 236 16588800
# 237 17020800
# 238 17452800
# 239 17884800
# 240 18316800
# 241 18748800
# 242 19180800
# 243 19612800
# 244 20044800
# 245 20476800
# 246 20908800
# 247 21340800
# 248 21772800
# 249 22204800
# 250 22636800
# 251 23068800
# 252 23500800
# 253 23932800
# 254 24364800
# 255 24796800
# 256 25228800
# 257 25660800
# 258 26092800
# 259 26524800
# 260 26956800
# 261 27388800
# 262 27820800
# 263 28252800
# 264 28684800
# 265 29116800
# 266 29548800
# 267 29980800
# 268 30412800
# 269 30844800
# 270 31276800
# 271 31708800
# 272 32140800
# 273 32572800
# 274 33004800
# 275 33436800
# 276 33868800
# 277 34300800
# 278 34732800
# 279 35164800
# 280 35596800
# 281 36028800
# 282 36460800
# 283 36892800
# 284 37324800
# 285 37756800
# 286 38188800
# 287 38620800
# 288 39052800
# 289 39484800
# 290 39916800
# 291 40348800
# 292 40780800
# 293 41212800
# 294 41644800
# 295 42076800
# 296 42508800
# 297 42940800
# 298 43372800
# 299 43804800
# 300 44236800
# 301 44668800
# 302 45100800
# 303 45532800
# 304 45964800
# 305 46396800
# 306 46828800
# 307 47260800
# 308 47692800
# 309 48124800
# 310 48556800
# 311 48988800
# 312 49420800
# 313 49852800
# 314 50284800
# 315 50716800
# 316 51148800
# 317 51580800
# 318 52012800
# 319 52444800
# 320 52876800
# 321 53308800
# 322 53740800
# 323 54172800
# 324 54604800
# 325 55036800
# 326 55468800
# 327 55900800
# 328 56332800
# 329 56764800
# 330 57196800
# 331 57628800
# 332 58060800
# 333 58492800
# 334 58924800
# 335 59356800
# 336 59788800
# 337 60220800
# 338 60652800
# 339 61084800
# 340 61516800
# 341 61948800
# 342 62380800
# 343 62812800
# 344 63244800
# 345 63676800
# 346 64108800
# 347 64540800
# 348 64972800
# 349 65404800
# 350 65836800
# 351 66268800
# 352 66700800
# 353 67132800
# 354 67564800
# 355 67996800
# 356 68428800
# 357 68860800
# 358 69292800
# 359 69724800
# 360 70156800
# 361 70588800
# 362 71020800
# 363 71452800
# 364 71884800
# 365 72316800
# 366 72748800
# 367 73180800
# 368 73612800
# 369 74044800
# 370 74476800
# 371 74908800
# 372 75340800
# 373 75772800
# 374 76204800
# 375 76636800
# 376 77068800
# 377 77500800
# 378 77932800
# 379 78364800
# 380 78796800
# 381 79228800
# 382 79660800
# 383 80092800
# 384 80524800
# 385 80956800
# 386 81388800
# 387 81820800
# 388 82252800
# 389 82684800
# 390 83116800
# 391 83548800
# 392 83980800
# 393 84412800
# 394 84844800
# 395 85276800
# 396 85708800
# 397 86140800
# 398 86572800
# 399 87004800
# 400 87436800
```

dnadales:
Some observations from these results:

- We cannot conclude anything about ledger vs consensus overhead from these benchmarks. We are not measuring the consensus overhead. I think measuring this requires more careful analysis. Maybe we should create a new issue.
- We seem to have a strong indication that HFC does not introduce significant overhead, at least for the operations benchmarked here. Why? Because we do not see a discrete increase in the total time these operations take as we cross the epoch boundaries.

dnadales:
Possible ways to drive this PR home:

- [ ] Incorporate the Gnuplot scripts into the repository. This requires that we adjust the plot boundaries based on the range of the input data.
- [ ] Improve structure (if applicable).
- [ ] Add more comments.
- [ ] Use more descriptive variable names.
- [ ] Document this feature (eg what the benchmark does, how to run it).
- [ ] Decide where to merge this PR: we should avoid developing on a release branch, however merging this to master will probably render `db-analyser` unusable since we won't have ledger or consensus version compatible with the Chain DB that was created using a previous version. So we should re-consider moving this tool somewhere else (eg `cardano-node` or to a separate repository).

Example of a place were we could add more comments:

```haskell
        -- why do we return this:
        pure $ St rtsStats5 $ ExtLedgerState ldgrSt' headerState'
```

We might consider adding some ASCII version of these drawings to the documentation 

image
image

nfrisby:
FWIW, my initial reaction was the same as Arnaud's:

> None beside sticking to gnuplot :slightly_smiling_face: Yes the doc is not straightforward but no graphing library in Haskell I am aware of is on par anyway. I would be interested if there were an alternative in Haskell though :slightly_smiling_face:
Alternatively, python as some great plotting libraries

I've also had pleasant experiences with some JS libraries (plop your data in a JSON array, copy it in to a HTML file, and boom you have interactive plots).

I would hesitate to commit to any plotting library in Haskell -- I expect it all to inevitably break 😢.

But that's easy enough: just also commit the gnuplot script as a back-up?

Are you liking the vega interface?

dnadales:
@nfrisby one of the next steps I had in mind was to output each line as a JSON representation of a `SlotDataPoint`, eg:

```haskell
data SlotDataPoint =
    SlotDataPoint
      { slot      :: Int
      , totalTime :: Int
      , mut       :: Int
      , gc        :: Int
       -- ... 
      }
```

and this is the format that the new plotting and analyzing tool will ingest. In this way, we make sure that if we change the output data in the `db-analyser` we will detect this in the tools that ingest this data. We can convert this data back to a CSV, but as a separate step (for instance when converting the file to something `gnuplot` can understand).

nfrisby:
> Option 0: remove the vega-lite implementation and generate a gnu-plot script (where I would basically accommodate the epoch and era boundaries in the plot to cover only the data being plotted).

Why do you have to do that? Can you plot them all (like my gnuplot script does) and the set the x-axis bounds explicitly to automatically crop out the "extra" epoch/era boundaries?

> Option 1: keep the vega-lite implementation, and also generate the gnu-plot script.

Why _generate_ the gnuplot script? Can't a static script suffice? Or maybe one that takes `gnuplot` parameters (you can set variable names on the command-line invocation)?

-----

I don't have a preference.

Tangent: The JSON format is a lot of per-line overhead, so that'll slow things down for longer runs and swell file sizes. Neither of which really matters, but I personally prefer the X-separated values format for tabular data.

dnadales:
> > Option 0: remove the vega-lite implementation and generate a gnu-plot script (where I would basically accommodate the epoch and era boundaries in the plot to cover only the data being plotted).
> 
> Why do you have to do that? Can you plot them all (like my gnuplot script does) and the set the x-axis bounds explicitly to automatically crop out the "extra" epoch/era boundaries?
> 
Won't you end up with vertical lines outside the x-axis as we can see in some of the plots above? Is there a way to prevent this in gnuplot? 

> I don't have a preference.
> 
In that case I'm going to keep the vega-lite backend just because it gives us the possibility of zooming in and the plots show a tooltip with the data values when the mouse pointer hovers over the data point.

> Tangent: The JSON format is a lot of per-line overhead, so that'll slow things down for longer runs and swell file sizes. Neither of which really matters, but I personally prefer the X-separated values format for tabular data.

Good point. This is the problem I want to solve: I'd like to couple the plotting program with the benchmarking program in such a way that if the latter outputs the data in a different format (eg swapping the order of the columns), we can detect this discrepancy at compile time. I guess I could define a [`Read`](https://www.stackage.org/haddock/lts-19.29/base-4.15.1.0/Prelude.html#t:Read) instance for `SlotDataPoint`, and in the benchmarking program collect the benchmarking data for a point as a `SlotDataPoint` value, and output it as a single line (as it is now).

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +449,160 @@ traceLedgerProcessing
...
+            Slotting.Origin        -> i
+            Slotting.At (SlotNo j) -> i - j
+
+        IO.hPutStrLn outFileHandle (   DP.showData slotDataPoint separator
+                                    ++ intercalate separator (HasAnalysis.blockStats blk)
+                                   )
+
+        pure $ BenchmarkLedgerOpsState currentRtsStats $ ExtLedgerState ldgrSt' headerSt'
+      where
+        forecastTheLedgerView slot rp = do
```
dnadales, line 544: Add local bindings.

Jasagredo:
Related PR that removes the forecasting step in the forging logic #4099

dnadales:
I ditched the `vega-lite` integration, maybe we're never gonna need it after all, and if so I can bring it back

dnadales:
So I think to merge this PR we need:

- [ ] An approval from @nfrisby 
- [ ] Decide where to merge it. Just rebase on top of `master`?

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +448,187 @@ traceLedgerProcessing
...
+        let slot = blockSlot      blk
+        (!tkLdgrView, tForecast) <- forecast            slot prevLedgerState
+        (!tkHdrSt,    tHdrTick)  <- tickTheHeaderState  slot prevLedgerState tkLdgrView
+        (!hdrSt',     tHdrApp)   <- applyTheHeader                           tkLdgrView tkHdrSt
+        (!tkLdgrSt,   tBlkTick)  <- tickTheLedgerState  slot prevLedgerState
+        (!ldgrSt',    tBlkApp)   <- applyTheBlock                                       tkLdgrSt
+
+        currentRtsStats <- GC.getRTSStats
+        let
+          slotDataPoint = DP.mkSlotDataPoint
```
nfrisby, line 523: I think the risk of passing arguments here to `mkSlotDataPoint` in the wrong order drastically out-weighs the risk of forgetting to use `showData`. So I suggest eliminating `mkSlotDataPoint` and using record expression syntax here.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +448,187 @@ traceLedgerProcessing
...
+benchmarkLedgerOps ::
+  forall blk.
+     ( HasAnalysis blk
+     , LedgerSupportsProtocol blk
+     )
+  => Analysis blk
+benchmarkLedgerOps AnalysisEnv {db, registry, initLedger, cfg, limit} =
+    -- TODO: at the moment we do not offer a configuration option for changing
+    -- the location of the output file.
+    IO.withFile benchmarkLedgerOpsOutputPath IO.WriteMode $ \outFileHandle -> do
```
nfrisby, line 490: Why not just `stdout`? I think that's much simpler and easier to anticipate.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +448,187 @@ traceLedgerProcessing
...
+
+      let separator = "\t"
```
nfrisby, line 492: Just move this to the `where` clause? That way you don't need to pass it to `process`.

nfrisby:
I'm anticipating that rebasing will let us drop the `Use a relative path for era config files` commit, since I'm pretty sure Karg already did that on `master`.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +448,187 @@ traceLedgerProcessing
...
+  - Block forging.
+  - Block validation.
+
+-------------------------------------------------------------------------------}
+
+data BenchmarkLedgerOpsState blk = BenchmarkLedgerOpsState {
+       -- | RTS stats that correspond to either the point in time in which the
+       -- benchmarking process started, or the end of the previous block
+       -- application.
+       prevRtsStats    :: !GC.RTSStats
```
nfrisby, line 473: I think we could drop this state data type if we call `GC.getRTSStats` six times, once as the very first and once as the very last thing (other than subtractions and rendering) in each application of `process`. And that would avoid measuring the costs of everything (eg `processAll` etc) other than the 5 steps done within `process`.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +448,187 @@ traceLedgerProcessing
...
+        applyTheHeader ::
+             Ticked (LedgerView (BlockProtocol blk))
+          -> Ticked (HeaderState blk)
+          -> IO (HeaderState blk, GC.RtsTime)
+        applyTheHeader tickedLedgerView tickedHeaderState = time chrono $ do
+          case runExcept $ validateHeader cfg tickedLedgerView (getHeader blk) tickedHeaderState of
+            Left err -> fail $ "benchmark doesn't support invalid headers: " <> show rp <> " " <> show err
+            Right x -> pure x
+
+        tickTheLedgerState ::
```
nfrisby, line 580: I personally find the local bindings a lot more noisy and distracting than just inlining them into the above.

I feel that most strongly about the two `tick*` bindings: they'd be very legible one-liners if inlined.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +448,187 @@ traceLedgerProcessing
...
+
+        applyTheBlock ::
+             Ticked (LedgerState blk)
+          -> IO (LedgerState blk, GC.RtsTime)
+        applyTheBlock tickedLedgerSt = time chrono $ do
+          case runExcept (lrResult <$> applyBlockLedgerResult lcfg blk tickedLedgerSt) of
+            Left err -> fail $ "benchmark doesn't support invalid blocks: " <> show rp <> " " <> show err
+            Right x -> pure x
+
+newtype Chronometer = Chronometer { mutatorElapsedTimeNS :: IORef GC.RtsTime }
```
nfrisby, line 595: You could use a `StateT IO` stack if you don't want to go this "objecty" route.

ouroboros-consensus-cardano/tools/db-analyser/HasAnalysis.hs
```diff
@@ -34,6 +37,8 @@ class (HasAnnTip blk, GetPrevHash blk) => HasAnalysis blk where
   -- | Emit trace markers at points in processing.
   emitTraces     :: WithLedgerState blk -> [String]
 
+  blockStats     :: blk -> [String]
```
nfrisby, line 40: Perhaps a comment explaining that this was introduced for the sake of the `BenchmarkLedgerOps` analysis pass.

nfrisby (commented):
I did a review pass on the diff.

nfrisby:
> Won't you end up with vertical lines outside the x-axis as we can see in some of the plots above? Is there a way to prevent this in gnuplot?

The gp script does. But i) I don't think it's a problem worth fixing and ii) if we did the math to plot the marks as a normal `plot` instead of as "`arrow`"s, then it would be cropped away by explicitly setting the x-axis. But that math (or maybe a condition in the loop that skips before xlo and after xhi) is just the kind of complexity that felt excessive 🤷

nfrisby:
> This is the problem I want to solve: I'd like to couple the plotting program with the benchmarking program in such a way that if the latter outputs the data in a different format (eg swapping the order of the columns), we can detect this discrepancy at compile time

Yeah, I also feel that. Since this is a diagnostic, I think at run-time instead of at compile-time would be fine too (eg CI should fail even everything built).

.... aha! We should be using `column()` in the gnuplot scripts instead of the column numbers. I'm surprised I didn't know about this already (maybe I had forgotten it)? https://stackoverflow.com/a/70279186

ouroboros-consensus-cardano/tools/db-analyser/scripts/plot-ledger-ops-cost.gp
```diff
@@ -0,0 +1,263 @@
...
+set terminal pngcairo size 1600,800
+set border back
+set output out_file
+
+set xlabel "slot"
+set ylabel "dur (μs)"
+
+set multiplot layout 2,1
+
+plot bench_data using 1:3 with impulses title "both",\
```
nfrisby, line 33: We should use `column()` here https://stackoverflow.com/a/70279186

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
dnadales: Because then one has to fiddle with re-directions and has interference from other processes that print to stdout, like `cabal`. I find it more ergonomic this way, provided we document correctly where the CSV file is stored. What do you think?

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
dnadales: I introduced the local bindings so that it becomes clear how the data flows between the different steps. If I inline those then we will lose this. I think this gives the reader an idea of the main steps involved, and she can go to the local definitions for more details:

```haskell
        (!tkLdgrView, tForecast) <- forecast            slot prevLedgerState
        (!tkHdrSt,    tHdrTick)  <- tickTheHeaderState  slot prevLedgerState tkLdgrView
        (!hdrSt',     tHdrApp)   <- applyTheHeader                           tkLdgrView tkHdrSt
        (!tkLdgrSt,   tBlkTick)  <- tickTheLedgerState  slot prevLedgerState
        (!ldgrSt',    tBlkApp)   <- applyTheBlock                                       tkLdgrSt
```

dnadales: In particular, after doing this refactoring I realized that for ticking the ledger state and applying the block we didn't need the other steps.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
dnadales: Good point. I don't have any preference, but I wonder if we need this IORef at all, isn't the cost of the extra `getRTSStats` negligible when compared with the other ledger ops? I'd love to get rid of the `Chronometer` :stopwatch:

ouroboros-consensus-cardano/tools/db-analyser/HasAnalysis.hs
dnadales: Absolutely! I overlooked this :/ Thanks.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
dnadales: Would this be overkill? 

```haskell
slotDataPoint = DP.mkSlotDataPoint
                            (realPointSlot rp) -- We're ok here, there's only one `Slot` type
                            (SlotGap   $ slot `slotCount` getTipSlot prevLedgerState)
                            (TotalTime $ GC.elapsed_ns         currentRtsStats `nsDiff` GC.elapsed_ns prevRtsStats)

newtype SlotGap = ...
```

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
nfrisby: I would guess it's relatively very low-cost. Moreover, "one extra" doesn't matter: we'll be calculating the difference between them, so the last one's cost is skipped (in the measurements, but not in the analysis' actual run-time).

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
nfrisby: Hmm. Maybe include a mandatory `--outfile` argument and give `--outfile -` the common `stdout` semantics?

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
nfrisby: I think so, yes.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
dnadales: Done.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
dnadales: Yeah, sounds sensisble. I removed it.

dnadales:
@nfrisby and I agree it's best to merge this into `master`, so I'll do that once I get approval from @nfrisby

ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
```diff
@@ -122,12 +128,19 @@ instance HasProtocolInfo (CardanoBlock StandardCrypto) where
     CardanoConfig {..} <- either (error . show) return =<<
       Aeson.eitherDecodeFileStrict' configFile
 
+    -- The paths to the eras configuration files (eg @byronGenesisPath@) are
```
nfrisby, line 131: I think this change is already merged into `master`.

See my top-level Review comment for more info.

nfrisby (commented):
Nothing major. Looks good!

I realized something, though: this PR is not aimed at the `master` branch. So, we should:

- change the PR to target `master` (.... I'm not even sure that's possible, but I'll try)

- rebase the commits onto `master` (which will be a bit of a bother since `db-analyser` was relocated to `ouroboros-consensus-cardano-tools`).

- also, the PR description is stale.

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
```diff
@@ -78,32 +78,39 @@ executable db-analyser
...
-                     , cardano-crypto-class
-                     , cardano-crypto-wrapper
-                     , cardano-ledger-alonzo
-                     , cardano-ledger-byron
-                     , cardano-ledger-core
-                     , cardano-ledger-shelley
                      , cborg
                      , containers
                      , contra-tracer
+                     , filepath          >=1.4   && <1.5
```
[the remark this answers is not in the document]
dnadales: Makes sense. 

This is out of scope, but do you know what do we want to do about the version bounds above the line on which you commented? 

```cabal
    base              >=4.9   && <4.15
                     , bytestring        >=0.10  && <0.11
                     , cborg             >=0.2.2 && <0.3
                     , containers        >=0.5   && <0.7
                     , mtl               >=2.2   && <2.3
                     , nothunks
                     , these             >=1.1   && <1.2
```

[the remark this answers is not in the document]
dnadales: Hmmm, look at this https://github.com/input-output-hk/ouroboros-network/blob/d7c63c3f545d5fcf72fc2c6390e3652f9fb95dfa/ouroboros-consensus/ouroboros-consensus.cabal#L295

I think we need to address this issue separately, namely, decide where to put the version bounds. Maybe as a part of https://github.com/input-output-hk/ouroboros-network/issues/4185

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +447,159 @@ traceLedgerProcessing
...
+  We focus on these 5 operations because they are involved in:
+
+  - Chain syncing.
+  - Block forging.
+  - Block validation.
+
+-------------------------------------------------------------------------------}
+
+newtype BenchmarkLedgerOpsState blk = BenchmarkLedgerOpsState {
+       -- | Intermediate ledger state to which the next block will be applied.
```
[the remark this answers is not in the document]
dnadales: Nothing :) That's a vestige from a previous version in which the state was richer IIRC. I'll remove it.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +447,159 @@ traceLedgerProcessing
...
+       prevLedgerState :: ExtLedgerState blk
+     }
+
+benchmarkLedgerOpsOutputPath :: String
```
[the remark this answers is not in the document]
dnadales: Yes, I overlooked this. Thanks!

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +447,159 @@ traceLedgerProcessing
...
+            , gc          = GC.gc_elapsed_ns      currentRtsStats `nsDiff` GC.gc_elapsed_ns prevRtsStats
+            , majGcCount  = GC.major_gcs          currentRtsStats    -     GC.major_gcs prevRtsStats
+            , forecast    = tForecast `div` 1000
+            , headerTick  = tHdrTick  `div` 1000
+            , headerApply = tHdrApp   `div` 1000
+            , blockTick   = tBlkTick  `div` 1000
+            , blockApply  = tBlkApp   `div` 1000
+            }
+          -- Compute the difference between two 'RtsTime's and convert it to microseconds.
+          nsDiff t t' = (t - t') `div` 1000
```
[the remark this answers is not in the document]
dnadales: > I know I chose it, but that was young foolish me from a couple months ago

:laughing: Actually, I should have looked at this more critically.  I updated this to use `-` and `div` directly.

ouroboros-consensus-cardano/tools/db-analyser/Block/Cardano.hs
dnadales: Sure. Let's see which conflicts I get when I rebase on top of master :)

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +447,159 @@ traceLedgerProcessing
...
+
+    process ::
+         IO.Handle
+      -> BenchmarkLedgerOpsState blk
+      -> blk
+      -> IO (BenchmarkLedgerOpsState blk)
+    process outFileHandle BenchmarkLedgerOpsState {prevLedgerState} blk = do
+        prevRtsStats <- GC.getRTSStats
+        let slot = blockSlot      blk
+        (!tkLdgrView, tForecast) <- forecast            slot prevLedgerState
```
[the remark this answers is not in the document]
dnadales: > Also, we specifically do not want these ! bang patterns to be necessary here: in order for the time wrapper to measure the actual calculation, the strictness patterns must have been inside the argument passed to time 

:100: 

I need your help understanding what is the right thing to do here. I see two options, which might be mutually exclusive: 

0. Add `!` inside the arguments to time, eg

```haskell
applyTheHeader tickedLedgerView tickedHeaderState = time $ do
          case ... of -- NO STRICNESS PATTERNS HERE
            Left err -> fail... -- NO STRICNESS PATTERNS HERE
            Right x -> pure $! x  -- STRICTNESS PATTERN HERE!!!
```

1. Add the strictness pattern to `time` directly:

```haskell
time act = do
    tPrev <- GC.mutator_elapsed_ns <$> GC.getRTSStats
    !r <- act
    tNow <- GC.mutator_elapsed_ns <$> GC.getRTSStats
    pure (r, tNow - tPrev)
```

If Option 1 makes Option 0 redundant (which is my current understanding), I'd be inclined towards using the strictness annotation in function `time`. In this way, we do not run the risk of forgetting a bang pattern inside the arguments passed to `time`.

What do you think?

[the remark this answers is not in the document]
dnadales: My proposal: https://github.com/input-output-hk/ouroboros-network/pull/4014/commits/2a3ab055b6d8206e8f860c046db23d5ef521aea4

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +447,159 @@ traceLedgerProcessing
...
+        (!tkHdrSt,    tHdrTick)  <- tickTheHeaderState  slot prevLedgerState tkLdgrView
+        (!hdrSt',     tHdrApp)   <- applyTheHeader                           tkLdgrView tkHdrSt
+        (!tkLdgrSt,   tBlkTick)  <- tickTheLedgerState  slot prevLedgerState
+        (!ldgrSt',    tBlkApp)   <- applyTheBlock                                       tkLdgrSt
+
+        currentRtsStats <- GC.getRTSStats
+        let
+          slotDataPoint =
```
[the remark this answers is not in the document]
dnadales: What about https://github.com/input-output-hk/ouroboros-network/pull/4014/commits/1bbaa890385b24be06a52a9787b607bb6e84653d

ouroboros-consensus-cardano/tools/db-analyser/Analysis/BenchmarkLedgerOps/SlotDataPoint.hs
```diff
@@ -0,0 +1,103 @@
...
+
+-- | Return the headers that correspond to the fields of 'SlotDataPoint'.
+--
+-- The position of each header matches the position in which the corresponding
+-- field value is returned in 'showData'. Eg, if show headers returns:
+--
+-- > "slot slotGap totalTime" ...
+--
+-- then the third value returned by 'showData' will correspond to 'totalTime'.
+showHeaders :: String -> String
```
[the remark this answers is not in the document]
dnadales: This is a great idea :heart: I incorporated this suggestion and replaced the use of `String` with `Text`.

ouroboros-consensus-cardano/tools/db-analyser/Analysis.hs
```diff
@@ -434,6 +447,159 @@ traceLedgerProcessing
...
+benchmarkLedgerOps ::
+  forall blk.
+     ( HasAnalysis blk
+     , LedgerSupportsProtocol blk
+     )
+  => Maybe FilePath -> Analysis blk
+benchmarkLedgerOps mOutfile AnalysisEnv {db, registry, initLedger, cfg, limit} =
+    withFile mOutfile $ \outFileHandle -> do
+
+      let
```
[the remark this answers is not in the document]
dnadales: > Nothing major. Looks good!
> 
Thank you Nick. I incorporated your comments and I'll rebase onto `master` and change the PR target.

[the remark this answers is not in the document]
dnadales: I have no idea what GH does when I click on "quote reply" :roll_eyes:

ouroboros-consensus-cardano/ouroboros-consensus-cardano.cabal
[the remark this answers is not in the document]
nfrisby: As of the rebase, switching this diff to Karg's new `ouroboros-consensus-cardano-tools` package, this concern is resolved _en passant_.

dnadales (approved):

nfrisby (commented):
Damian and had a couple calls on this. I Approve 🙌 (though he'll have to click that, since I opened the PR originally.)

PR #4028, opened 2022-09-21
consensus: make TickHardForkLedgerView fields strict

nfrisby:
While preparing PR #4014, I noticed the `forecast` function for `CardanoBlock` was lazy. This small patch makes it strict.

This is a Draft PR because I'm not totally sure we want it to be strict.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Protocol/LedgerView.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Protocol/LedgerView.hs
@@ -46,8 +46,8 @@ type HardForkLedgerView = HardForkLedgerView_ WrapLedgerView
 -------------------------------------------------------------------------------}
 
 data instance Ticked (HardForkLedgerView_ f xs) = TickedHardForkLedgerView {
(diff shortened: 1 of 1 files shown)


nfrisby:
I think this will only cause minor improvements to the performance. The main benefit is that is much less confusing for profiling, etc.

So: let's merge it and await the normal per-release benchmarking to catch any regression.

If that benchmark does show regressions, then we can suggest removing it with a trivial git revert on a branch and trying that in the next benchmark run.

dnadales:
> I think this will only cause minor improvements to the performance. The main benefit is that is much less confusing for profiling, etc.
> 
> So: let's merge it and await the normal per-release benchmarking to catch any regression.
> 
> If that benchmark does show regressions, then we can suggest removing it with a trivial git revert on a branch and trying that in the next benchmark run.

I'd rather we detect any performance regressions sooner rather than later, but if we can't run a full-blown benchmark we will have to merge. I asked @kosyrevSerge if we could do something to try to detect any performance regressions earlier, but if we can't, then let's go ahead and merge.

dnadales (approved):

nfrisby:
bors retry

Issue #3943, opened 2022-08-08
Pending to be discarded: Implement performant version using WithoutLedgerTables


Jasagredo:
The In-Memory backend is slow and in reality there is not much gain in comparison with the Legacy Ledger for the users that use the In-Memory backend. Therefore it makes more sense to have only two options from now on:

- In-Memory backend: works almost the same way as the Legacy Ledger and all the operations on tables are trivially no-ops.
- LMDB backend: uses all the UTxO-HD machinery.

This implies using a new datatype:

```haskell
data SwitchLedgerTables = WithLedgerTables | WithoutLedgerTables

class TableStuff l wt where 
    ...

instance TableStuff l WithoutLedgerTables where
    data instance LedgerTables ... = NoLedgerTables
    ...
```

The expectation is that this will achieve similar performance as the pre-UTxO-HD node.

Jasagredo:
This needs also the companion PR https://github.com/input-output-hk/cardano-node/pull/4304

Jasagredo:
This is blocked on the benchmarking of the anti-diff implementation

Jasagredo:
Blocked on #4206

Jasagredo:
Blocked on #4040

PR #4036, opened 2022-09-22
Integrate anti-diff prototype in `consensus` packages

### LINKED ISSUE #4010: Integrate anti-diff prototype
#3997 implements and micro-benchmarks the anti-diff prototype as described in #3946: an alternative implementation of sequences of (UTxO) differences, which should be more performant than the "intermediate-sum" implementation (#3997 elaborates on the naming). To get  a sense of the real performance gain, we aim to benchmark either `cardano-node`, or `db-analyser`. Before this can happen, we should integrate the anti-diff prototype in all of `consensus`.

jorisdral:
# Description

Resolves #4010. This is a follow-up PR to #3997.

This PR integrates the anti-diff diff sequences in consensus. In addition to integration, this PR includes two major changes:

* Anti-diff invariants are now enforced through non-empty sequences.
* The `DiffSeq` datatype now checks more slot-related invariants.

The need for these changes only became apparent during integration. The invariants are now more heavily enforced, and should prevent misuse of the datatypes.

My advice is to look at the changes on a per-commit basis, which should make the diffs easier to parse.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested


Jasagredo:
Linked #3943 as WithoutLedgerTables becomes obsolete once this is merged

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -375,14 +376,14 @@ class ( ShowLedgerState (LedgerTables l)
 
   pureLedgerTables ::
        (forall k v.
-            Ord k
+            (Ord k, Eq v)
```
Jasagredo, line 379: I'm unsure why are these Eq constraints propagated if we are not comparing the values in the end?

Jasagredo, line 379: I'm unsure why are these Eq constraints propagated if we are not comparing the values in the end?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
```diff
@@ -744,6 +744,12 @@ reapplyTrackingTicked after before =
     zipOverLedgerTablesTicked rawReapplyTracking after
   $ projectLedgerTables before
 
+rawTranslateDiff ::
+     (v -> v')
+  -> DiffMK k v
+  -> DiffMK k v'
+rawTranslateDiff f (ApplyDiffMK d) = ApplyDiffMK $ fmap f d
```
Jasagredo, line 751: This name confuses me as we talked about `Translate` but this is not related to the `Translate` operation is it?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -653,33 +653,35 @@ mkDiskLedgerView (LedgerBackingStoreValueHandle seqNo vh, ldb, close) =
...
+      getSum $ DS.unsafeFoldMapDiffEntry (Sum . oneIfDel) d
       where
-        oneIfDel (HD.UtxoEntryDiff _v diffstate) = case diffstate of
-          HD.UedsDel       -> 1
-          HD.UedsIns       -> 0
-          HD.UedsInsAndDel -> 0
+        oneIfDel x = case x of
+          DS.Delete _           -> 1
+          DS.Insert _           -> 0
+          DS.UnsafeAntiDelete _ -> error "Found UnsafeAntiDelete"
```
Jasagredo, line 682: Something a bit more explanatory?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -359,6 +360,8 @@ translateShelleyTables ::
      ( SL.TranslateEra     era TxOutWrapper
      , SL.TranslationError era TxOutWrapper ~ Void
      , EraCrypto (SL.PreviousEra era) ~ EraCrypto era
+     , Eq (Cardano.Ledger.Core.TxOut (SL.PreviousEra era))
```
Jasagredo, line 363: Same question. Why do we need these Eq constraints?

Jasagredo (commented):
I am in general concerned that we are loosing the previous implementation. We might as well just delete the `HD` module right? 

Would it be possible to unify the interface and define a `CPP` flag that switches implementations? Is it even worth? Ping @dnadales

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
Jasagredo: I then saw that this is used when defining `translateLedgerTablesWith` in era transitions `NP`s

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -687,7 +687,7 @@ instance CardanoHardForkConstraints c => LedgerTablesCanHardFork (CardanoEras c)
       byron = InjectLedgerTables $ \NoByronLedgerTables -> polyEmptyLedgerTables
 
       shelley ::
-           forall era proto. SL.Crypto era ~ c
+           forall era proto. (SL.Crypto era ~ c, Eq (Core.TxOut era))
```
dnadales, line 690: Why do we need this constraint now? We should check with the ledger that this is not too strong.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
dnadales: Can't we define a `Functor (DiffMK k)` instance?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -142,18 +154,23 @@ data RootMeasure k v = RootMeasure {
 
 data InternalMeasure k v = InternalMeasure {
     -- | Cumulative length
-    imLength :: {-# UNPACK #-} !Length
-    -- | Right-most slot number
+    imLength  :: {-# UNPACK #-} !Length
+    -- | Left-most slot number (or lower bound)
```
dnadales, line 158:
```suggestion
    -- | Leftmost slot number (or lower bound)
```

Do we need the hyphen?

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
```diff
@@ -142,18 +154,23 @@ data RootMeasure k v = RootMeasure {
...
     --
     -- Empty diff sequences have no right-most slot number, so in that case
     -- @imSlotNo == Nothing@.
-  , imSlotNo ::                !(Maybe SlotNo)
+  , imSlotNoL ::                !(Maybe SlotNoLB)
+    -- | Right-most slot number (or upper bound)
```
dnadales, line 163:
```suggestion
    -- | Rightmost slot number (or upper bound)
```

Jasagredo:
The groupoid laws seem to fail on Hydra? https://hydra.iohk.io/build/19760053/nixlog/8

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
jorisdral: Good catch, I will remove them again.

This might be a bit of tangent, but I was wondering if it is necessary to have this Rank-2 type here, and why we could not have `k` and `v` mentioned in the header of `TableStuff`. It's somewhat cumbersome that we have to add new constraints on `k` and `v` in `pureLedgerTables` every time we want to apply `pureLedgerTables` to some new function that places these additional constraints on `k` and `v`. Would that be an alternative, or would other parts of the code break?

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
jorisdral: > Can't we define a `Functor (DiffMK k)` instance?

Will do

> This name confuses me ...

I added this function with the aim of having self-documenting code: I wanted to make explicit that mapping `(v -> v')` over a diff is an era translation. However, if it's confusing I'll remove it of course.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
jorisdral: I think it can be removed because we don't perform the anti-diff sanity checks anymore

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
jorisdral: I am unsure, but I will remove them

jorisdral:
> The groupoid laws seem to fail on Hydra? https://hydra.iohk.io/build/19760053/nixlog/8

You're right, and it should fail as far as I can tell with the current `Arbitrary` instances. I'll have to look into whether this is a bug or a feature

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
jorisdral: > Good catch, I will remove them again.

I was wrong. The constraints are there because there are other anti-diff functions that require the equality constraint, such as `diff` and the `Semigroup`, `Monoid` and `Group` instance functions

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
jorisdral: > I think these can be removed as well

Again, I stand corrected. See https://github.com/input-output-hk/ouroboros-network/pull/4036#discussion_r991936963

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
jorisdral: Again, I stand corrected. See https://github.com/input-output-hk/ouroboros-network/pull/4036#discussion_r991936963

jorisdral:
> I am in general concerned that we are loosing the previous implementation. We might as well just delete the `HD` module right?
> 
> Would it be possible to unify the interface and define a `CPP` flag that switches implementations? Is it even worth? Ping @dnadales

If you consider it a risk to remove the previous implementation, we could keep it around for a while on the feature branch at least

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
jorisdral: > Why do we need this constraint now? We should check with the ledger that this is not too strong.

We need this constraint now because of `mapValuesAppliedMK`, which needs `Eq` specifically in the case that it is mapping over a `SeqDiffMK`

JaredCorduan: @jorisdral I do not think that having the constraint will cause us any problems. Ledger has the same constraint in several places, such as the `EraTxOut` class.

dnadales:
> > I am in general concerned that we are loosing the previous implementation. We might as well just delete the `HD` module right?
> > Would it be possible to unify the interface and define a `CPP` flag that switches implementations? Is it even worth? Ping @dnadales
> 
> If you consider it a risk to remove the previous implementation, we could keep it around for a while on the feature branch at least

I'd say we should keep HD around definitely during the prototyping phase. @jorisdral can you remind me if anti-diffs are guaranteed to exist for any choice of `k` `v`? If so I do not see a reason why we should keep the old implementation around afterwards.

ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
dnadales: I wonder if we can remove some of the `Eq` constraints. For instance, does the patch below make sense? (`ouroboros-consensus` compiles with it, but I haven't tried the rest of the code, however, the question is rather about the purpose of the `Eq` constraint).

```diff
diff --git a/anti-diff/src/Data/Map/Diff/Strict.hs b/anti-diff/src/Data/Map/Diff/Strict.hs
index 0ea9cf480..e63cf065b 100644
--- a/anti-diff/src/Data/Map/Diff/Strict.hs
+++ b/anti-diff/src/Data/Map/Diff/Strict.hs
@@ -88,7 +88,7 @@ import           NoThunks.Class (NoThunks (..), noThunksInValues)
 -- @'DiffHistory'@ is non-empty. This prevents the @'Map'@ from getting bloated with
 -- empty diff histories.
 newtype Diff k v = Diff (Map k (NEDiffHistory v))
-  deriving stock (Generic, Show, Eq, Functor)
+  deriving stock (Generic, Show, Functor)
   deriving anyclass (NoThunks)
 
 -- | A history of changes to a value in a key-value store.
@@ -101,20 +101,20 @@ newtype UnsafeDiffHistory t v = UnsafeDiffHistory (t (DiffEntry v))
 
 deriving stock instance Show v => Show (UnsafeDiffHistory Seq v)
 deriving stock instance Show v => Show (UnsafeDiffHistory NESeq v)
-deriving stock instance Eq v => Eq (UnsafeDiffHistory Seq v)
-deriving stock instance Eq v => Eq (UnsafeDiffHistory NESeq v)
+-- deriving stock instance Eq v => Eq (UnsafeDiffHistory Seq v)
+-- deriving stock instance Eq v => Eq (UnsafeDiffHistory NESeq v)
 
 {-# COMPLETE DiffHistory #-}
 
 newtype DiffHistory v = MkDiffHistory (UnsafeDiffHistory Seq v)
-  deriving stock (Generic, Show, Eq, Functor)
+  deriving stock (Generic, Show, Functor)
   deriving newtype (NoThunks, Foldable)
 
 {-# COMPLETE NEDiffHistory #-}
 
 -- | A non-empty diff history.
 newtype NEDiffHistory v = MkNEDiffHistory (UnsafeDiffHistory NESeq v)
-  deriving stock (Generic, Show, Eq, Functor)
+  deriving stock (Generic, Show, Functor)
   deriving newtype (NoThunks, Foldable)
 
 pattern DiffHistory :: Seq (DiffEntry v) -> DiffHistory v
@@ -142,7 +142,7 @@ data DiffEntry v =
     | Delete !v
     | UnsafeAntiInsert !v
     | UnsafeAntiDelete !v
-  deriving stock (Generic, Show, Eq, Functor, Foldable)
+  deriving stock (Generic, Show, Functor, Foldable)
   deriving anyclass (NoThunks)
 
 {------------------------------------------------------------------------------
@@ -290,8 +290,17 @@ invertDiffEntry = \case
 --
 -- For simplicity, we simply compare the inverse of the first argument to the
 -- second argument. That is, inversion should be invertible.
+--
+-- TODO: we only use this in teststing code it seems. I wonder if we can make
+-- this def local.
 areInverses :: Eq v => DiffEntry v -> DiffEntry v -> Bool
-areInverses e1 e2 = invertDiffEntry e1 == e2
+areInverses (Insert x) (UnsafeAntiInsert y) = x == y
+  -- Delete x           -> UnsafeAntiDelete x
+  -- UnsafeAntiInsert x -> Insert x
+  -- UnsafeAntiDelete x -> Delete x
+
+
+-- e1 e2 = invertDiffEntry e1 == e2
 
 {------------------------------------------------------------------------------
   Values and keys
@@ -399,7 +408,7 @@ data ApplyDiffError k v =
   | BadDelMatchingKey k v v (NEDiffHistory v)
   | BadDelInsMatchingKey k v v v (NEDiffHistory v)
   | InsDelMatchingKey k v (NEDiffHistory v)
-  deriving (Show, Eq)
+  deriving (Show)
 
 -- | Applies a diff to values, performs sanity checks.
 --
diff --git a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
index a635f81f3..ec7e4296c 100644
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Ledger/Basics.hs
@@ -821,16 +821,16 @@ mapValuesAppliedMK f = \case
   ApplyQueryAllMK      -> ApplyQueryAllMK
   ApplyQuerySomeMK vs  -> ApplyQuerySomeMK (fmap f vs)
 
-instance (Ord k, Eq v) => Eq (ApplyMapKind' mk k v) where
-  ApplyEmptyMK          == _                     = True
-  ApplyKeysMK   l       == ApplyKeysMK   r       = l == r
-  ApplyValuesMK l       == ApplyValuesMK r       = l == r
-  ApplyTrackingMK l1 l2 == ApplyTrackingMK r1 r2 = l1 == r1 && l2 == r2
-  ApplyDiffMK l         == ApplyDiffMK r         = l == r
-  ApplySeqDiffMK l      == ApplySeqDiffMK r      = l == r
-  ApplyQueryAllMK       == ApplyQueryAllMK       = True
-  ApplyQuerySomeMK l    == ApplyQuerySomeMK r    = l == r
-  _                     == _                     = False
+-- instance (Ord k, Eq v) => Eq (ApplyMapKind' mk k v) where
+--   ApplyEmptyMK          == _                     = True
+--   ApplyKeysMK   l       == ApplyKeysMK   r       = l == r
+--   ApplyValuesMK l       == ApplyValuesMK r       = l == r
+--   ApplyTrackingMK l1 l2 == ApplyTrackingMK r1 r2 = l1 == r1 && l2 == r2
+--   ApplyDiffMK l         == ApplyDiffMK r         = l == r
+--   ApplySeqDiffMK l      == ApplySeqDiffMK r      = l == r
+--   ApplyQueryAllMK       == ApplyQueryAllMK       = True
+--   ApplyQuerySomeMK l    == ApplyQuerySomeMK r    = l == r
+--   _                     == _                     = False
 
 instance (Ord k, NoThunks k, NoThunks v) => NoThunks (ApplyMapKind' mk k v) where
   wNoThunks ctxt   = NoThunks.allNoThunks . \case
diff --git a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
index 82e502a85..559f88865 100644
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
@@ -138,7 +138,7 @@ newtype DiffSeq k v =
       (InternalMeasure k v)
       (Element k v)
     )
-  deriving stock (Generic, Show, Eq)
+  deriving stock (Generic, Show)
   deriving anyclass (NoThunks)
 
 -- The @'SlotNo'@ is not included in the root measure, since it is
@@ -149,7 +149,7 @@ data RootMeasure k v = RootMeasure {
     -- | Cumulative diff
   , rmDiff   :: !(Diff k v)
   }
-  deriving stock (Generic, Show, Eq, Functor)
+  deriving stock (Generic, Show, Functor)
   deriving anyclass (NoThunks)
 
 data InternalMeasure k v = InternalMeasure {
@@ -173,7 +173,7 @@ data Element k v = Element {
     elSlotNo :: {-# UNPACK #-} !Slot.SlotNo
   , elDiff   ::                !(Diff k v)
   }
-  deriving stock (Generic, Show, Eq, Functor)
+  deriving stock (Generic, Show, Functor)
   deriving anyclass (NoThunks)
 
 -- | Length of a sequence of differences.

```

dnadales: The test suite will fail to compile with the patch above, but maybe the question is why we need to test the root measure 

```text
test-storage/Test/Ouroboros/Storage/LedgerDB/HD/DiffSeq.hs:23:9: error:
    • No instance for (Eq (RootMeasure Key Val))
        arising from a use of ‘testSemigroupLaws’
    • In the expression: testSemigroupLaws
      In the second argument of ‘testGroupWithProxy’, namely
        ‘[testSemigroupLaws, testMonoidLaws, testGroupLaws]’
      In the expression:
        testGroupWithProxy
          (Proxy @(RootMeasure Key Val))
          [testSemigroupLaws, testMonoidLaws, testGroupLaws]
   |
23 |         testSemigroupLaws
   |         ^^^^^^^^^^^^^^^^^
```

dnadales: The patch I proposed will also break this:

```text

src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs:269:10: error:
    • Could not deduce (Eq
                          (ApplyMapKind'
                             'Ouroboros.Consensus.Ledger.Basics.EmptyMK'
                             (SL.TxIn (Core.Crypto era))
                             (Core.TxOut era)))
        arising from the superclasses of an instance declaration
      from the context: ShelleyBasedEra era
        bound by the instance declaration
        at src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs:269:10-81
    • In the instance declaration for
        ‘TableStuff (LedgerState (ShelleyBlock proto era))’
    |
269 | instance ShelleyBasedEra era => TableStuff (LedgerState (ShelleyBlock proto era)) where
```

dnadales: What I don't get is why the `Eq` constraints that I removed are not needed in `ouroboros-consensus`, where the core logic of the LedgerDB is implemented, but they are needed by the clients.

dnadales: I hope you guys don't mind that I mark this as unresolved for now.

dnadales: Sorry for the extra noise. One important aspect that I overlooked is that at the moment you need the equality constraints to be able to invert values, eg:

```haskell
areInverses :: Eq v => DiffEntry v -> DiffEntry v -> Bool
areInverses (Insert x) (UnsafeAntiInsert y) = x == y
```

So this means that we can only use anti-diffs with values that have an `Eq` instance, right? In that case that is an argument for keeping the old implementation around. 

Coming back to the original question of this thread, I wonder if it'd be possible to re-engineer this in such a way that:

- define table stuff independently of the `Eq` constraint.
- if client code wants to use the anti-diff implementation they will need to provide these `Eq` instances.

jorisdral:
> > > I am in general concerned that we are loosing the previous implementation. We might as well just delete the `HD` module right?
> > > Would it be possible to unify the interface and define a `CPP` flag that switches implementations? Is it even worth? Ping @dnadales
> > 
> > 
> > If you consider it a risk to remove the previous implementation, we could keep it around for a while on the feature branch at least
> 
> I'd say we should keep HD around definitely during the prototyping phase. @jorisdral can you remind me if anti-diffs are guaranteed to exist for any choice of `k` `v`? If so I do not see a reason why we should keep the old implementation around afterwards.

Anti-diffs are guaranteed to exist for any choice of `k` `v` (modulo some constraints like `Ord`)

PR #4039, opened 2022-09-22
db-analyser: use reapply not apply for --store-ledger

nfrisby:
This should have always been `reapply` instead of `apply`.


dnadales (approved):
Indeed!

PR #4073, opened 2022-10-13
Test that UTxO tables are updated correctly on era transitions

### LINKED ISSUE #4043: Cardano eras transitions tests
- [x] Write tests for all the Cardano transitions. 
    - [x] Byron to Shelley
    - [x] Shelley to Allegra
    - [x] Allegra to Mary
    - [x] Mary to Alonzo 
    - [x] Alonzo to Babbage
- [x] Add to the `AddingAnEra.md` document instructions to add a new era translation test when adding a new era.

If we can generate arbitrary ledger states of each Cardano era we can write property tests. Otherwise, we'll have to resort to unit tests.

# Byron to Shelley 
Ensure that *all the UTxO* is transferred from in-memory Byron state (which has no tables) to the ledger tables of the Shelley state. The result must consist only of insertions (which are present in the resulting diffs).

# Shelley to Allegra
Ensure that the *avvm* addresses appear as deletions in the resulting diffs.

# All remaining transitions
Ensure that the resulting diff is empty, which means no modification to the ledger tables will be made from one era to the next.

bartfrenk:
# Description

Test that UTxO tables are computed correctly on era transitions, for all era transitions.

Closes #4043.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
+++ b/ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
@@ -1,14 +1,15 @@
-{-# LANGUAGE DataKinds          #-}
-{-# LANGUAGE FlexibleInstances  #-}
-{-# LANGUAGE GADTs              #-}
-{-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TypeApplications   #-}
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs             #-}
+{-# LANGUAGE NamedFieldPuns    #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Test.Consensus.Byron.Generators (
     RegularBlock (..)
   , epochSlots
+  , genByronLedgerConfig
+  , genByronLedgerState
   , k
   , protocolMagicId
   ) where
@@ -19,7 +20,9 @@ import qualified Data.Map.Strict as Map
 
 import           Cardano.Binary (fromCBOR, toCBOR)
 import           Cardano.Chain.Block (ABlockOrBoundary (..),
-                     ABlockOrBoundaryHdr (..))
+                     ABlockOrBoundaryHdr (..), ChainValidationState (..),
+                     cvsPreviousHash)
+
 import qualified Cardano.Chain.Block as CC.Block
 import qualified Cardano.Chain.Byron.API as API
 import           Cardano.Chain.Common (KeyHash)
@@ -27,6 +30,7 @@ import qualified Cardano.Chain.Delegation as CC.Del
 import qualified Cardano.Chain.Delegation.Validation.Activation as CC.Act
 import qualified Cardano.Chain.Delegation.Validation.Interface as CC.DI
 import qualified Cardano.Chain.Delegation.Validation.Scheduling as CC.Sched
+import           Cardano.Chain.Genesis as Byron
 import qualified Cardano.Chain.Genesis as CC.Genesis
 import           Cardano.Chain.Slotting (EpochNumber, EpochSlots (..),
                      SlotNumber)
@@ -36,6 +40,7 @@ import qualified Cardano.Chain.Update.Validation.Interface as CC.UPI
 import qualified Cardano.Chain.Update.Validation.Registration as CC.Reg
 import           Cardano.Crypto (ProtocolMagicId (..))
 import           Cardano.Crypto.Hashing (Hash)
+import           Cardano.Slotting.Slot (WithOrigin (..))
 
 import           Ouroboros.Consensus.Block
 import           Ouroboros.Consensus.Config.SecurityParam
@@ -55,6 +60,7 @@ import           Test.QuickCheck.Hedgehog (hedgehog)
 import qualified Test.Cardano.Chain.Block.Gen as CC
 import qualified Test.Cardano.Chain.Common.Gen as CC
 import qualified Test.Cardano.Chain.Delegation.Gen as CC
+import qualified Test.Cardano.Chain.Genesis.Gen as CC
 import qualified Test.Cardano.Chain.MempoolPayload.Gen as CC
 import qualified Test.Cardano.Chain.Slotting.Gen as CC
 import qualified Test.Cardano.Chain.UTxO.Gen as CC
@@ -271,6 +277,28 @@ instance Arbitrary ByronTransition where
 
 instance Arbitrary (LedgerState ByronBlock EmptyMK) where
   arbitrary = ByronLedgerState <$> arbitrary <*> arbitrary <*> arbitrary
+
+-- | Generator for a Byron ledger state in which the tip of the ledger given by
+-- `byronLedgerTipBlockNo` matches that derived from the `ChainValidationState` in
+-- `byronLedgerState`.
+genByronLedgerState :: Gen (LedgerState ByronBlock EmptyMK)
+genByronLedgerState = do
+  chainValidationState <- arbitrary
+  ledgerTransition <- arbitrary
+  ledgerTipBlockNo <- genLedgerTipBlockNo chainValidationState
+  pure $ ByronLedgerState {
+      byronLedgerTipBlockNo = ledgerTipBlockNo
+    , byronLedgerState = chainValidationState
+    , byronLedgerTransition = ledgerTransition
+    }
+  where
+    genLedgerTipBlockNo ChainValidationState { cvsPreviousHash } =
+      case cvsPreviousHash of
+        Left _  -> pure Origin
+        Right _ -> At <$> arbitrary
+
+genByronLedgerConfig :: Gen Byron.Config
+genByronLedgerConfig = hedgehog $ CC.genConfig protocolMagicId
 
 instance Arbitrary (TipInfoIsEBB ByronBlock) where
   arbitrary = TipInfoIsEBB <$> arbitrary <*> elements [IsEBB, IsNotEBB]
--- a/ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
+++ b/ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
@@ -30,7 +30,7 @@ import qualified Cardano.Chain.Delegation as CC.Del
 import qualified Cardano.Chain.Delegation.Validation.Activation as CC.Act
 import qualified Cardano.Chain.Delegation.Validation.Interface as CC.DI
 import qualified Cardano.Chain.Delegation.Validation.Scheduling as CC.Sched
-import           Cardano.Chain.Genesis as Byron
+import qualified Cardano.Chain.Genesis as Byron
 import qualified Cardano.Chain.Genesis as CC.Genesis
 import           Cardano.Chain.Slotting (EpochNumber, EpochSlots (..),
                      SlotNumber)
@@ -288,7 +288,7 @@ genByronLedgerState = do
   ledgerTipBlockNo <- genLedgerTipBlockNo chainValidationState
   pure $ ByronLedgerState {
       byronLedgerTipBlockNo = ledgerTipBlockNo
-    , byronLedgerState = chainValidationState
+    , byronLedgerState      = chainValidationState
     , byronLedgerTransition = ledgerTransition
     }
   where
--- a/ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
+++ b/ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
@@ -279,8 +279,8 @@ instance Arbitrary (LedgerState ByronBlock EmptyMK) where
   arbitrary = ByronLedgerState <$> arbitrary <*> arbitrary <*> arbitrary
 
 -- | Generator for a Byron ledger state in which the tip of the ledger given by
--- `byronLedgerTipBlockNo` matches that derived from the `ChainValidationState` in
--- `byronLedgerState`.
+-- `byronLedgerTipBlockNo` is consistent with the chain validation state, i.e., if there is no
+-- previous block, the ledger tip wil be `Origin`.
 genByronLedgerState :: Gen (LedgerState ByronBlock EmptyMK)
 genByronLedgerState = do
   chainValidationState <- arbitrary
--- a/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
+++ b/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
@@ -26,6 +26,7 @@ import           Cardano.Slotting.Slot (EpochNo (..))
 
 import qualified Cardano.Chain.Block as Byron
 import qualified Cardano.Chain.UTxO as Byron
+
 import           Cardano.Ledger.Alonzo ()
 import           Cardano.Ledger.Alonzo.Genesis (AlonzoGenesis (..))
 import           Cardano.Ledger.Alonzo.Scripts (CostModels (..), ExUnits (..),
@@ -39,8 +40,9 @@ import           Cardano.Ledger.Shelley.API
                      ShelleyGenesisStaking (..), TxIn (..),
                      translateCompactTxOutByronToShelley,
                      translateTxIdByronToShelley)
+import           Cardano.Ledger.Shelley.LedgerState hiding (LedgerState)
 import           Cardano.Ledger.Shelley.PParams (emptyPParams)
-import           Cardano.Ledger.Shelley.UTxO
+import           Cardano.Ledger.Shelley.UTxO (UTxO (..))
 
 import           Ouroboros.Consensus.BlockchainTime.WallClock.Types
                      (slotLengthFromSec)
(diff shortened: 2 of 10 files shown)


ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
```diff
@@ -272,6 +278,28 @@ instance Arbitrary ByronTransition where
 instance Arbitrary (LedgerState ByronBlock EmptyMK) where
   arbitrary = ByronLedgerState <$> arbitrary <*> arbitrary <*> arbitrary
 
+-- | Generator for a Byron ledger state in which the tip of the ledger given by
+-- `byronLedgerTipBlockNo` matches that derived from the `ChainValidationState` in
```
dnadales, line 282: We should explain what *matches* means here. Maybe something like: 

> block number of the ledger tip is consistent with the chain validation state, this is, if there is no previous block, then the ledger tip will be origin

ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
```diff
@@ -272,6 +278,28 @@ instance Arbitrary ByronTransition where
...
+-- `byronLedgerState`.
+genByronLedgerState :: Gen (LedgerState ByronBlock EmptyMK)
```
dnadales, line 284: While it makes sense that the tip block number matches the chain state, I wonder if we should generate ledger states in which this is not the case. Don't we want to test our system under these circumstances?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+  ) where
+
+import qualified Data.ListMap as ListMap
+import           Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+import           Test.QuickCheck
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+
+import qualified Cardano.Chain.Block as Byron
```
dnadales, line 18: Just to check: does this group of imports correspond to `cardano-ledger`?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+import           Test.Consensus.Cardano.MockCrypto (MockCryptoCompatByron)
+import           Test.Consensus.Shelley.Generators ()
+import           Test.Consensus.Shelley.MockCrypto
+import           Test.Util.Time (dawnOfTime)
+
+import qualified Cardano.Ledger.Block as Core
+import qualified Cardano.Ledger.Core as Core
+import qualified Cardano.Ledger.Era as Core
+
+type Crypto = MockCryptoCompatByron
```
dnadales, line 69: I wonder if we should remove this level of indirection.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+type Era    = ShelleyEra Crypto
+type Proto  = TPraos Crypto
+
+data TestSetup = TestSetup {
```
dnadales, line 73: Are we going to have one test setup per-each era transition?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+    TestSetup <$> genByronLedgerConfig
+              <*> pure fixedShelleyLedgerConfig
+              <*> genByronLedgerState
+
+tests :: TestTree
+tests = testGroup "ByronToShelley" [
+    testProperty "test" prop_byronUtxosMatchThoseInShelleyTable
+  ]
+
+prop_byronUtxosMatchThoseInShelleyTable :: TestSetup -> Property
```
dnadales, line 95: I wonder if we should comment the background of this test: why do we need it?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+prop_byronUtxosMatchThoseInShelleyTable TestSetup {..} =
```
dnadales, line 96: Should we avoid wildcards, so that it is clearer where each identified is declared?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+
+fixedShelleyLedgerConfig :: Core.TranslationContext era ~ ()
+                         => ShelleyLedgerConfig era
+fixedShelleyLedgerConfig = mkShelleyLedgerConfig
+      shelleyGenesis
+      ()
+      (fixedEpochInfo (sgEpochLength shelleyGenesis) tpraosSlotLength)
+      (MaxMajorProtVer 1000)
+
+shelleyGenesis :: ShelleyGenesis era
```
dnadales, line 206: If we use a fixed Shelley genesis we should justify why. I think in this case the justification might be that we are interested in the initial UTxO set, which is not affected by the genesis config. But we need to confirm this.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+  -> LedgerState (ShelleyBlock Proto (ShelleyEra Crypto)) EmptyMK
+  -> LedgerState (ShelleyBlock Proto (AllegraEra Crypto)) DiffMK
+translateLedgerStateShelleyToAllegra shelleyConfig allegraConfig epochNo shelleyLedgerState =
+  let translateLedgerState = shelleyToAllegraTranslateLedgerState shelleyConfig allegraConfig
+  in translateLedgerStateWith translateLedgerState epochNo shelleyLedgerState
+
+genShelleyLedgerState :: CanMock proto era => Gen (LedgerState (ShelleyBlock proto era) EmptyMK)
+genShelleyLedgerState = arbitrary
+
+run :: IO ()
```
dnadales, line 169: Is this temporary?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+
+tests :: TestTree
+tests = testGroup "ByronToShelley" [
+    testProperty "test" prop_byronUtxosMatchThoseInShelleyTable
+  ]
+
+prop_byronUtxosMatchThoseInShelleyTable :: TestSetup -> Property
+prop_byronUtxosMatchThoseInShelleyTable TestSetup {..} =
+  let
+    epochNo = EpochNo 10
```
dnadales, line 98: What is the effect of the epoch number in this property, and/or in the ledger state translation?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+    gen :: Gen (LedgerState (ShelleyBlock Proto (AllegraEra Crypto)) DiffMK)
+    gen = do
+      let shelleyC = fixedShelleyLedgerConfig
+      let allegraC = fixedShelleyLedgerConfig
+      state <- genShelleyLedgerState
+      pure $ translateLedgerStateShelleyToAllegra shelleyC allegraC epochNo state
+    epochNo :: EpochNo
+    epochNo = EpochNo 10
+
+translateToShelleyLedgerState
```
dnadales, line 181: What about?

```suggestion
translateByronToShelleyLedgerState
```

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,226 @@
...
+  :: LedgerConfig ByronBlock
+  -> LedgerConfig (ShelleyBlock Proto Era)
+  -> EpochNo
+  -> LedgerState ByronBlock EmptyMK
+  -> LedgerState (ShelleyBlock Proto Era) DiffMK
+translateToShelleyLedgerState byronConfig shelleyConfig epochNo byronLedgerState =
```
dnadales, line 187: Would it make sense to make these definitions local to their use site?

dnadales (commented):
Looks good so far.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
bartfrenk: Yes

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
bartfrenk: I think it is unnecessary indirection, but it reduces noise in type signatures. I am impartial to this one.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
bartfrenk: I have checked the imports, although from the style guide is not completely clear to me how to order imports (I'd prefer it to be simpler, but that is another discussion).

ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
bartfrenk: The translation then fails; my assumption that "invalid" ledger states are only due to programming errors, in which case I don't think we should test the behaviour, but that assumption might be wrong.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
dnadales: I'd prefer skipping the extra level of indirection, but that's subjective. @Jasagredo preferences?

ouroboros-consensus-byron-test/src/Test/Consensus/Byron/Generators.hs
```diff
@@ -272,6 +278,28 @@ instance Arbitrary ByronTransition where
...
+genByronLedgerState = do
+  chainValidationState <- arbitrary
+  ledgerTransition <- arbitrary
+  ledgerTipBlockNo <- genLedgerTipBlockNo chainValidationState
+  pure $ ByronLedgerState {
+      byronLedgerTipBlockNo = ledgerTipBlockNo
```
dnadales, line 290: I think we try to align this in other parts of the code base.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
dnadales: So do they correspond to `cardano-ledger`? :smile: To put it differently, why do we have this group of imports. My assumption was that you grouped the import in this way because they all belong to the same package. I'm not objecting to this grouping ;)

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,320 @@
...
+import           Ouroboros.Consensus.Byron.Ledger (ByronBlock, byronLedgerState)
+
+import           Ouroboros.Consensus.Shelley.Eras
+import           Ouroboros.Consensus.Shelley.HFEras ()
+import           Ouroboros.Consensus.Shelley.Ledger (ShelleyBlock,
+                     ShelleyLedgerConfig, mkShelleyLedgerConfig,
+                     shelleyLedgerState, shelleyLedgerTables, shelleyUTxOTable)
+import           Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
+
+-- TODO: Not sure where to put these
```
dnadales, line 74: I think they're ok here. @Jasagredo ?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,320 @@
...
+
+-- Definitions to make the signatures a bit less unwieldy
+type Crypto = MockCryptoCompatByron
+type Proto  = TPraos Crypto
+
+tests :: TestTree
+tests = testGroup "UpdateTablesOnEraTransition" $ toTestTrees predicates tls
+  where
+    tls = translateLedgerState hardForkEraTranslation
+    predicates :: InPairs TranslationPredicate (CardanoEras Crypto)
```
dnadales, line 93: Why not define a `[TestTree]` directly? Is `toTestTrees` necessary?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,320 @@
...
+  , tsDestLedgerConfig :: LedgerConfig dest
+  , tsSrcLedgerState   :: LedgerState src EmptyMK
+  , tsEpochNo          :: EpochNo
+}
+
+deriving instance ( Show (LedgerConfig src)
+                  , Show (LedgerConfig dest)
+                  , Show (LedgerState src EmptyMK)) => Show (TestSetup src dest)
+
+-- todo: Useful to merge some of these instances?
```
dnadales, line 232: I don't see how we could merge the Byron to Shelley instances. I'm not sure about the others. Do you know what prevents us from defining a:

```haskell 
instance Arbitrary (TestSetup (ShelleyBlock p era0) (ShelleyBlock p era1))
```

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,320 @@
...
+-- transitions does not depend on the configurations of any of the ledgers involved..
+-- TODO: Check this
+fixedShelleyLedgerConfig :: Core.TranslationContext era -> ShelleyLedgerConfig era
+fixedShelleyLedgerConfig translationContext = mkShelleyLedgerConfig
+      shelleyGenesis
+      translationContext
+      (fixedEpochInfo (sgEpochLength shelleyGenesis) (slotLengthFromSec 2))
+      (MaxMajorProtVer 1000)
+  where
+    shelleyGenesis = ShelleyGenesis {
```
dnadales, line 286: We should align these.

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
```diff
@@ -346,6 +350,10 @@ allDisjoint = go Set.empty
     go _   []       = True
     go acc (xs:xss) = Set.disjoint acc xs && go (Set.union acc xs) xss
 
+-- | Map over keys and values
```
dnadales, line 353: Why is this called a `dimap` and not a `bimap`? My guess: `dimap` does not satisfy the bifunctor laws (I haven't checked).

dnadales (approved):

dnadales:
I forgot to add these remaining tasks:

- [ ] Add to the AddingAnEra.md document instructions to add a new era translation test when adding a new era.
- [ ] Use quickcheck's [`cover`](https://www.stackage.org/haddock/lts-19.29/QuickCheck-2.14.2/Test-QuickCheck.html#v:cover) to check that we're testing important cases.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
```diff
@@ -0,0 +1,320 @@
...
+}
+
+toTestTrees
+  :: (AllPairs TestSetup Arbitrary xs, AllPairs TestSetup Show xs)
+  => InPairs TranslationPredicate xs
+  -> InPairs (RequiringBoth WrapLedgerConfig TranslateLedgerState) xs
+  -> [TestTree]
+toTestTrees predicates translations = mkTestTrees $ zipInPairs predicates translations
+  where
+    fn
```
Jasagredo, line 177: I would avoid naming this `fn` because we are heavily into `generics-sop` in these hard fork tests and there is actually a function named like that in SOP: https://hackage.haskell.org/package/generics-sop-0.5.1.2/docs/Generics-SOP.html#v:fn

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
bartfrenk: I agree with both, but I also tend to read files as @Jasagredo. So just to close the discussion I am keeping it as it is.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
bartfrenk: I have no preference. The prefixes in this case make it clear.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
bartfrenk: Well, I couldn't do it.

ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
bartfrenk: It is a profunctor I think, in the sense that it contravariant in the first argument, covariant in the latter. The profunctor map in https://hackage.haskell.org/package/profunctors-5.6.2/docs/Data-Profunctor.html is called `dimap`.

ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/Util/InPairs.hs
```diff
@@ -120,3 +123,15 @@ requiringBoth = flip go
     go :: InPairs (RequiringBoth h f) xs -> NP h xs -> InPairs f xs
     go PNil         _              = PNil
     go (PCons f fs) (x :* y :* zs) = PCons (provideBoth f x y) (go fs (y :* zs))
+
+{-------------------------------------------------------------------------------
+  Constraints
+-------------------------------------------------------------------------------}
+
+type family AllPairs
```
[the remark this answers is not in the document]
bartfrenk: This is adapted from Maguire's Thinking with types. I can kind-of see how to do it with classes, but I like the terseness of this definition, and I think, but do not know, the class-one will be more convoluted.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Translation.hs
bartfrenk: To define the properties I need to have access to the appropriate `TranslateEra x y` values. I chose the direction of zipping the predicates with what is loosely, `InPairs TranslateEra (CardanoEras c)`.

dnadales:
We should also test with empty UTxO, shouldn't we? 

```text
      +++ OK, passed 100 tests (100% UTxO set is not empty).
```

It seems we aren't at the moment.

PR #4087, opened 2022-10-19
Run GitHub Actions on pull requests

coot:


.github/workflows/build.yml
```diff
@@ -1,6 +1,6 @@
 name: Haskell CI
 
-on: [push]
+on: [push, pull_request]
```
amesgen, line 3: Right now, this means that in PRs created from branches in this repository, we run the GH Actions checks twice (as in this PR). One solution would be to restrict `push` to certain branches ([docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-to-specific-branches-occurs)).

bolt12 (approved):

.github/workflows/build.yml
coot: Indeed that's annoying (that's why we just use `push`). Restricting `push` would also be a bit annoying.  I often create PRs early, and push to the branch, which triggers the CI.  I wonder if this would work:
```yaml
on:
  pull_request:
    types: [opened, reopened, synchronize]
```
But `synchronize` probably doesn't do the same thing as `push`.

coot: `cardano-ledger` is using this trick to avoid this:
```
on:
  push:
    branches: [ '*' ]
  pull_request:
    branches: [ "master", "release/*" ]
```
I think it works for PRs that are opened directly from `master` branches of forked repos, but what if one opens a PR from another branch?

coot: What about this: https://github.com/input-output-hk/ouroboros-network/pull/4094
I need to check if it will work for a PR from a fork (especially the branch restriction rules).

coot:
bors merge

.github/workflows/build.yml
```diff
@@ -1,6 +1,10 @@
 name: Haskell CI
 
-on: [push]
+on:
+  push:
+    branches: ['*']
```
amesgen, line 5: Ah, I think the issue with the Hydra timeout is that this trigger does not fire on commits on `bors/staging`, as the `*` glob does not match the `/` (see [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)), so it probably has to be included in addition.

amesgen (changes requested):

.github/workflows/build.yml
coot: Oh, so it means if one pushes a branch `coot/feature-x` it will not trigger as well.

amesgen: Yes, that is my understanding, one would have to use branch names without `/` when one wants to run CI without creating a PR.

coot: I checked on another repo that at least when when one pushes to a branch with open PR, then just `pull_request` is enough to trigger CI.  I previously thought that this is only done when `push` is set as well.

I think it's ok to be explicit in `push` and allow all pull request, e.g.

```yaml
on:
  push:
    branches:
      - 'bors/*'
  pull_request:
```

It seems the ledger team is not really using `bors` for merging, that's why it wasn't an issue for them.

coot: Now should not trigger the same workflow even for PRs from forks.

amesgen (approved):

coot:
bors merge

PR #4098, opened 2022-10-24
Clarify current VRF tie breaking implementation

amesgen:
# Description

Update/correct potentially misleading names of VRF-related functions (in particular `csvLeaderVRF`), and provide context for why a particular VRF value is used for tie breaking.   

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Common.hs
+++ b/ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Common.hs
@@ -49,13 +49,14 @@ newtype MaxMajorProtVer = MaxMajorProtVer
 -- 1. By chain length, with longer chains always preferred.
 -- 2. If the tip of each chain was issued by the same agent, then we prefer
 --    the chain whose tip has the highest ocert issue number.
--- 3. By the leader value of the chain tip, with lower values preferred.
+-- 3. By a VRF value from the chain tip, with lower values preferred. See
+--    @pTieBreakVRFValue@ for which one is used.
 data PraosChainSelectView c = PraosChainSelectView
   { csvChainLength :: BlockNo,
     csvSlotNo      :: SlotNo,
     csvIssuer      :: SL.VKey 'SL.BlockIssuer c,
     csvIssueNo     :: Word64,
-    csvLeaderVRF   :: VRF.OutputVRF (VRF c)
+    csvTieBreakVRF :: VRF.OutputVRF (VRF c)
   }
   deriving (Show, Eq, Generic, NoThunks)
 
@@ -64,7 +65,7 @@ instance Crypto c => Ord (PraosChainSelectView c) where
     mconcat
       [ compare `on` csvChainLength,
         whenSame csvIssuer (compare `on` csvIssueNo),
-        compare `on` Down . csvLeaderVRF
+        compare `on` Down . csvTieBreakVRF
       ]
     where
       -- When the @a@s are equal, use the given comparison function,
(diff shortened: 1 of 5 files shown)


Jasagredo (approved):
Looks reasonable to me

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Common.hs
```diff
@@ -49,13 +49,14 @@ newtype MaxMajorProtVer = MaxMajorProtVer
...
--- 3. By the leader value of the chain tip, with lower values preferred.
+-- 3. By a VRF value from the chain tip, with lower values preferred. See
+--    @pTieBreakVRFValue@ for which one is used.
 data PraosChainSelectView c = PraosChainSelectView
   { csvChainLength :: BlockNo,
     csvSlotNo      :: SlotNo,
     csvIssuer      :: SL.VKey 'SL.BlockIssuer c,
     csvIssueNo     :: Word64,
-    csvLeaderVRF   :: VRF.OutputVRF (VRF c)
+    csvTieBreakVRF :: VRF.OutputVRF (VRF c)
```
dnadales, line 59: Why do we think this is a better name than the previous one? Aren't in any case `csvChainLength` and `csvSlotNo` tie breakers? What about `csvChainTipVRF`?

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Protocol/Praos.hs
```diff
@@ -138,7 +138,13 @@ instance PraosCrypto c => ProtocolHeaderSupportsProtocol (Praos c) where
       }
   pHeaderIssuer = hbVk . headerBody
   pHeaderIssueNo = SL.ocertN . hbOCert . headerBody
-  pHeaderVRFValue = certifiedOutput . hbVrfRes . headerBody
+
+  -- This is the "unified" VRF value, prior to range extension which yields e.g.
+  -- the leader VRF value used for slot election.
+  --
+  -- In the future, we might want to use a dedicated range-extended VRF value
```
dnadales, line 145: In the future or in `TPraos`, or "in future protocols"?

dnadales (approved):
I left one small question. Thanks for adding those comments to `pTieBreakVRFValue`.

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Common.hs
amesgen: In `TPraos`, `csvLeaderVRF` is the correct name, as we actually use the leader VRF value for tie breaking. But in `Praos` (which uses the same `SelectView`, i.e. this one), we no longer use the leader VRF, so this name is incorrect in that situation.

Hence, I chose a name indicating the purpose of the VRF in the context of chain selection, as it is the final tie breaker, compared to e.g. `csvChainLength` which causes the ties in the first place. (Side remark: `csvSlotNo` is not used for anything, we could remove it.)

But we don't have to stress its role as a tie breaker in the field name, e.g. `csvChainTipVRF` also sounds reasonable, as it also (intentionally!) does not mention which exact VRF from the header is used.

dnadales: > Hence, I chose a name indicating the purpose of the VRF in the context of chain selection, as it is the final tie breaker,

I see, that makes sense. I'm ok with that name. Thanks!

amesgen:
bors merge

nfrisby:
I'm not 100% sure, but I suspect the rename of `csvTieBreakVRF` might require a (small) integration in eg `cardano-node` and so should be mentioned in the `interface-CHANGELOG.md`.

Did you check for that sort of thing already, @amesgen ?

amesgen:
Grepping for `csvLeaderVRF` does not reveal anything in node (and also not in a few other IOG repos I have checked out). From my intuition, a `SelectView` should be a purely Consensus-internal thing, so I would actually be surprised/consider it tech debt if it would show up in consumers of the Cardano protocol. Where you thinking about a specific way it might be used in e.g. the node?

PR #4111, opened 2022-10-26
make maxMajorProtVer easier to correctly maintain

nfrisby:
See the message of the single commit in this PR.

### DIFF
--- a/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
+++ b/ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
@@ -579,7 +579,22 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
     -- The major protocol version of the last era is the maximum major protocol
     -- version we support.
     maxMajorProtVer :: MaxMajorProtVer
-    maxMajorProtVer = MaxMajorProtVer (pvMajor protVerBabbage)
+    maxMajorProtVer =
+          MaxMajorProtVer
+        $ pvMajor
+        $ nonEmptyLast
+        $ exactlyWeakenNonEmpty
+        $ protVers
+      where
+        protVers :: Exactly (CardanoShelleyEras StandardCrypto) ProtVer
+        protVers = Exactly $
+          -- ensure that these have the same order as 'CardanoShelleyEras'!
+          K protVerShelley :*
+          K protVerAllegra :*
+          K protVerMary :*
+          K protVerAlonzo :*
+          K protVerBabbage :*
+          Nil
 
     -- Byron
 


ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -579,7 +579,18 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
     -- The major protocol version of the last era is the maximum major protocol
     -- version we support.
     maxMajorProtVer :: MaxMajorProtVer
-    maxMajorProtVer = MaxMajorProtVer (pvMajor protVerBabbage)
+    maxMajorProtVer =
+        nonEmptyLast (exactlyWeakenNonEmpty maxMajorProtVers)
+      where
+        protVers :: Exactly (CardanoShelleyEras StandardCrypto) MaxMajorProtVer
+        protVers = Exactly $
+          -- ensure that these have the same order as 'CardanoShelleyEras'!
```
amesgen, line 587: As discussed, a more invasive approach that does not have this (very minor) potential for future incorrect updates would be to modify `ProtocolParams*` to be type-family based and take an `NP` of those as an argument to `protocolInfoCardano`, which the last protocol version can then be derived from.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
```diff
@@ -579,7 +579,18 @@ protocolInfoCardano protocolParamsByron@ProtocolParamsByron {
...
+          protVerShelley :*
+          protVerAllegra :*
+          protVerMary :*
+          protVerAlonzo :*
+          protVerBabbage :*
+          Nil
```
amesgen, line 593: This fails to compile; this should work:
```suggestion
          MaxMajorProtVer
        . pvMajor
        . nonEmptyLast
        . exactlyWeakenNonEmpty
        $ protVers
      where
        protVers :: Exactly (CardanoShelleyEras StandardCrypto) ProtVer
        protVers = Exactly $
          -- ensure that these have the same order as 'CardanoShelleyEras'!
             K protVerShelley
          :* K protVerAllegra
          :* K protVerMary
          :* K protVerAlonzo
          :* K protVerBabbage
          :* Nil
```

amesgen (changes requested):
Nice, this will make it very unlikely that we forget to update this in the future!

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/Node.hs
nfrisby: I very much agree. Let's merge this PR now, and open a TechDebt for the follow-up (as it involves at least some `cardano-node` changes etc)

amesgen (approved):

PR #4143, opened 2022-11-10
Fix `Arbitrary` instance for keys and values in `DiffSeq` benchmarks

jorisdral:
# Description

This is a fix related to benchmarks that were introduced in https://github.com/input-output-hk/ouroboros-network/pull/3997.

In the `DiffSeq` benchmarks, we set out to generate keys and values as uniformly random 32-byte and 64-byte `ShortByteStrings` respectively, which model 32-byte and 64-byte cryptographic hashes, by generating and packing 32 and 64 `Word8`s respectively:

```haskell
newtype Key = Key ShortByteString
newtype Value = Value ShortByteString

instance Arbitrary Key where
  arbitrary = Key . B.pack <$> replicateM 32 arbitrary

instance Arbitrary Value where
  arbitrary = Value . B.pack <$> replicateM 64 arbitrary
```

However, in Quickcheck (2.14.2), the `Arbitrary` instance for `Word8` is:

```haskell
instance Arbitrary Word8 where
  arbitrary = arbitrarySizedBoundedIntegral
  shrink    = shrinkIntegral
```

where `arbitrarySizedBoundedIntegral` is:

```haskell
-- | Generates an integral number from a bounded domain. The number is
-- chosen from the entire range of the type, but small numbers are
-- generated more often than big numbers. Inspired by demands from
-- Phil Wadler.
arbitrarySizedBoundedIntegral :: (Bounded a, Integral a) => Gen a
```

This means that we are currently not picking keys and values from a uniform distribution. Instead, we should use:

```haskell
-- | Generates an integral number. The number is chosen uniformly from
-- the entire range of the type. You may want to use
-- 'arbitrarySizedBoundedIntegral' instead.
arbitraryBoundedIntegral :: (Bounded a, Integral a) => Gen a
```

So, now we have:

```haskell
instance Arbitrary Key where
  arbitrary = Key . B.pack <$> replicateM 32 arbitraryBoundedIntegral

instance Arbitrary Value where
  arbitrary = Value . B.pack <$> replicateM 64 arbitraryBoundedIntegral
```

## Changes in benchmark results

Performance gains decrease with respect to previous results as listed in https://github.com/input-output-hk/cardano-updates/blob/consensus/update-2022-08-31/posts/2022-08-31-consensus.markdown. Still, speedups are in the range of `[3.33, 4.75]` across the different configurations.

|Configuration|Implementation|Mean CPU time|2*Stdev (CPU time)|
|-|-|-|-|
|Default|AntiDiff|2.974 s (0.26x)|273 ms|
|Default|LegacyDiff|11.416 s|11 ms|
|No switches|AntiDiff|2.553 s (0.24x)|156 ms|
|No switches|LegacyDiff|10.838 s|65 ms|
|Only small switches|AntiDiff|2.442 s (0.21x)|14 ms|
|Only small switches|LegacyDiff|11.760 s|331 ms|
|No switches, larger flushes (every 100 pushes)|AntiDiff|2.261 s (0.30x)|45 ms|
|No switches, larger flushes (every 100 pushes)|LegacyDiff|7.531 s|612 ms|

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
+++ b/ouroboros-consensus-test/bench-storage/Bench/Ouroboros/Consensus/Storage/LedgerDB/HD/DiffSeq.hs
@@ -117,8 +117,9 @@ import           QuickCheck.GenT (Arbitrary (arbitrary), Gen, MonadGen (..),
 import           Test.Tasty.Bench (Benchmark, bcompare, bcompareWithin, bench,
                      bgroup, env, nf)
 import           Test.Tasty.QuickCheck (Positive (getPositive), Property,
-                     Small (getSmall), conjoin, counterexample, forAll,
-                     generate, once, testProperty, (.&&.), (===))
+                     Small (getSmall), arbitraryBoundedIntegral, conjoin,
+                     counterexample, forAll, generate, once, testProperty,
+                     (.&&.), (===))
 
 import           Ouroboros.Consensus.Block (SlotNo (..))
 import qualified Ouroboros.Consensus.Storage.LedgerDB.HD as HD
@@ -553,12 +554,12 @@ newtype Value = Value ShortByteString
 
 -- | UTxO keys are expected to be 32 byte cryptographic hashes.
 instance Arbitrary Key where
-  arbitrary = Key . B.pack <$> replicateM 32 arbitrary
+  arbitrary = Key . B.pack <$> replicateM 32 arbitraryBoundedIntegral
 
 -- | UTxO entries, i.e., key-value pairs, are expected to be around 100-128 bytes
 -- in size. (@32 + 64 ~= 100@)
 instance Arbitrary Value where
-  arbitrary = Value . B.pack <$> replicateM 64 arbitrary
+  arbitrary = Value . B.pack <$> replicateM 64 arbitraryBoundedIntegral
 
 {-------------------------------------------------------------------------------
   Command generator: Monad transformer stack


Jasagredo (approved):
Nice catch

dnadales (commented):
I think the comment you wrote in this PR is excellent. We should include part of the rational for using `arbitraryBoundedIntegral` here, otherwise, this knowledge might get lost (even though it might be in the git history).

jorisdral:
> I think the comment you wrote in this PR is excellent. We should include part of the rational for using arbitraryBoundedIntegral here, otherwise, this knowledge might get lost (even though it might be in the git history).

Good point. The rationale is that we try to simulate a consensus workload as best as possible. In this case, we are simulating operations on a sequence of UTxO (i.e., key-value pairs) differences. UTxO keys are 32-byte cryptographic hashes, where each hash is equally likely to occur, and so if we generate random UTxO-like keys, we should ensure that we pick the keys uniformly random from the whole range of possible 32-byte keys. Previously, we were using `arbitrarySizedBoundedIntegral` to generate the keys:  though this picks from the full range of 32-byte keys, it does not pick from a uniform distribution because smaller keys are more likely to be picked. `arbitraryBoundedIntegral` does pick from a uniform distribution, so that is the change we made.

dnadales:
> > I think the comment you wrote in this PR is excellent. We should include part of the rational for using arbitraryBoundedIntegral here, otherwise, this knowledge might get lost (even though it might be in the git history).
> 
> Good point. The rationale is that we try to simulate a consensus workload as best as possible. In this case, we are simulating operations on a sequence of UTxO (i.e., key-value pairs) differences. UTxO keys are 32-byte cryptographic hashes, where each hash is equally likely to occur, and so if we generate random UTxO-like keys, we should ensure that we pick the keys uniformly random from the whole range of possible 32-byte keys. Previously, we were using `arbitrarySizedBoundedIntegral` to generate the keys: though this picks from the full range of 32-byte keys, it does not pick from a uniform distribution because smaller keys are more likely to be picked. `arbitraryBoundedIntegral` does pick from a uniform distribution, so that is the change we made.

I was a bit vague about what I meant by "here" :grimacing: I think we should have put this as a comment in the code. Do you think that makes sense @jorisdral ? If so, could you submit a tiny PR?

PR #4144, opened 2022-11-14
Add support for CHaP and ghc-9.2

erikd:
Builds with both ghc-8.10 and ghc-9.2.

Currently uses a `source-package-repository` reference to a `cardano-ledger` PR that has not yet been merged.


angerman:
The windows failures might partially be due to https://github.com/msys2/MINGW-packages/issues/10837#issuecomment-1047105402; more investigation needed.

ouroboros-network-framework/src/Ouroboros/Network/ErrorPolicy.hs
```diff
@@ -248,9 +249,14 @@ completeApplicationTx ErrorPolicies {epConErrorPolicies} (ConnectionError t addr
               carState =
                 case ps of
                   PeerStates peerStates -> PeerStates $ Map.update fn addr peerStates
-                  ThrowException{}      -> ps,
-              carThreads = Set.empty,
-              carTrace   = Just $
+#if __GLASGOW_HASKELL__ < 900 
+                  -- GHC 9 is certain this pattern is
+                  -- not used. GHC 8 apparently can't
+                  -- agree. m(
```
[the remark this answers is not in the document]
angerman: you don't like my ascii facepalm?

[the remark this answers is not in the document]
coot: oh I see ok, I didn't know that's how facepalm looks like in ascii.

ouroboros-network/src/Ouroboros/Network/Protocol/LocalTxMonitor/Type.hs
```diff
@@ -220,7 +220,9 @@ instance Protocol (LocalTxMonitor txid tx slot) where
   data NobodyHasAgency st where
     TokDone  :: NobodyHasAgency StDone
 
-  exclusionLemma_ClientAndServerHaveAgency TokIdle tok = case tok of {}
+  exclusionLemma_ClientAndServerHaveAgency TokIdle tok     = case tok of {}
+
+  exclusionLemma_ClientAndServerHaveAgency TokAcquired tok = case tok of {}
 
   exclusionLemma_NobodyAndClientHaveAgency TokDone tok = case tok of {}
```
coot, line 227: let's keep the code aligned
```suggestion
  exclusionLemma_ClientAndServerHaveAgency TokIdle     tok = case tok of {}

  exclusionLemma_ClientAndServerHaveAgency TokAcquired tok = case tok of {}

  exclusionLemma_NobodyAndClientHaveAgency TokDone     tok = case tok of {}
```

cabal.project
```diff
@@ -81,52 +81,55 @@ constraints:
...
+  , cardano-ledger-shelley-test >= 0.1.1.1
+  , cardano-data >= 0.1.1.1
+  , cardano-ledger-core >= 0.1.1.1
+  , cardano-ledger-pretty >= 0.1.1.1
+  , cardano-protocol-tpraos >= 0.1.1.1
+  , non-integral >= 0.1.1.1
+  , set-algebra >= 0.1.1.1
+  , small-steps >= 0.1.1.1
+  , small-steps-test >= 0.1.1.1
+  , vector-map >= 0.1.1.1
```
coot, line 132: @dnadales you should distribute these constraints in `ouroboros-consensus` packges.

@angerman I'd prefer `^>=` so we don't get CI build failures as soon as a new version of `cardano-ledger` or `plutus` is published to `CHaP.

coot (changes requested):

angerman:
bors merge

coot (approved):
LGTM

angerman:
bors merge

amesgen:
@angerman Is it intentional that we don't test with GHC 9.2 on Hydra/is there are a follow-up PR already planned? We don't run the Consensus tests on GH Actions, so this means that GHC 9.2 is untested on CI for Consensus.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/TxGen/Shelley.hs
```diff
@@ -62,8 +61,6 @@ instance HashAlgorithm h => TxGen (ShelleyBlock (TPraos (MockCrypto h)) (MockShe
       --
       -- When fixed, remove the line below to re-enable the transaction
       -- generator.
-      | otherwise               = pure []
-
```
amesgen, line 66: This is a logic change, isn't it?

amesgen (changes requested):
Usually, PRs that touch both Network and Consensus should be reviewed by team members from both Consensus and Network (see also the codeowners) -- maybe we should make that more clear.

amesgen:
bors cancel

ouroboros-consensus-shelley-test/src/Test/ThreadNet/TxGen/Shelley.hs
erikd: Furthermore, if the code above the deleted code  did not cover **all** cases, there would have been a compiler warning and there wasn't.

I also find it useful to convert `-Wall` warnings to errors using `-Werror` . I trust GHC far more than I trust myself.

amesgen: The problem is not that not all cases all covered, but rather that there is a logic change 

The code in question looked like this:

```haskell
  testGenTxs _coreNodeId _numCoreNodes curSlotNo cfg extra lst
      | stgeStartAt > curSlotNo = pure []

      -- TODO Temporarily disable the transaction generator until we fix the
      -- failing assertion in TxSubmission.Inbound, see #2680.
      --
      -- When fixed, remove the line below to re-enable the transaction
      -- generator.
      | otherwise               = pure []

      | otherwise               = do
      n <- choose (0, 20)
      go [] n $ applyChainTick lcfg curSlotNo lst
```

The first `| otherwise` clause was intentionally added in order to work around an existing bug (#2680). Now the pattern match redundancy checker got smarter in GHC 9.2 compared to 8.10, and detects that the second `| otherwise` clause is unreachable (which is intentional here!).

To fix this, you can use a more "complicated" expression that evaluates to `True` instead of `otherwise`, e.g. `() == ()` (or even the new [`considerAccessible`](https://hackage.haskell.org/package/base-4.17.0.0/docs/GHC-Exts.html#v:considerAccessible) that has exactly this purpose).

Does that clear it up?

---

> I also find it useful to convert `-Wall` warnings to errors using `-Werror`.

Agreed, we do that on Hydra, but this PR does not add 9.2 to Hydra CI (see https://github.com/input-output-hk/ouroboros-network/pull/4144#issuecomment-1322006034).

erikd: The current way the last case is disabled is confusing to people not a part of this project in that it current generates a compiler warning that someone not familiar with the code thinks they should remove.

angerman:
@amesgen who can review this? It's overwhelming mechanical changes to add 9.2 compatibility, and it's blocking downstream work on node.

ouroboros-consensus-shelley-test/src/Test/ThreadNet/TxGen/Shelley.hs
erikd: I have come up with an alternative that both maintains the logic and satisfies GHC.

amesgen: Thanks, using an `if True` is another way to "outsmart" even the improved GHC 9.2 pattern match redundancy checker :smile:

amesgen:
> who can review this? It's overwhelming mechanical changes to add 9.2 compatibility, and it's blocking downstream work on node.

I think the only thing missing now is GHC 9.2 on Nix/Hydra -- see https://github.com/input-output-hk/ouroboros-network/pull/4144#issuecomment-1322006034, and this is also the only place where we have `-Werror` enabled for the Consensus packages: While testing everything locally, I found a few warnings in the Consensus code (also pushed a commit fixing those, feel free to squash).

ouroboros-consensus-cardano-tools/app/DBAnalyser/Parsers.hs
```diff
@@ -1,8 +1,12 @@
 {-# LANGUAGE ApplicativeDo #-}
+{-# LANGUAGE CPP           #-}
 
 module DBAnalyser.Parsers (parseCmdLine) where
 
+#if __GLASGOW_HASKELL__ < 900
```
nfrisby, line 6: Could we avoid CPP here if we used a qualified import?

A very rushed experiment suggests that making this a qualified import and similarly qualifying the `asum` occurrences in this module (I assume there aren't many) would avoid the warning.

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos.hs
```diff
@@ -362,13 +362,13 @@ deriving instance PraosCrypto c => NoThunks (PraosValidationErr c)
...
 
 instance PraosCrypto c => ConsensusProtocol (Praos c) where
-  type ChainDepState _ = PraosState c
-  type IsLeader _ = PraosIsLeader c
-  type CanBeLeader _ = PraosCanBeLeader c
-  type SelectView _ = PraosChainSelectView c
-  type LedgerView _ = Views.LedgerView c
-  type ValidationErr _ = PraosValidationErr c
-  type ValidateView _ = PraosValidateView c
+  type ChainDepState (Praos c) = PraosState c
```
nfrisby, line 365: Happy to see this, thanks.

ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/ShelleyHFC.hs
```diff
@@ -317,6 +317,7 @@ translateLedgerViewAcrossShelley =
 
 instance ( ShelleyBasedEra era
          , ShelleyBasedEra (SL.PreviousEra era)
+         , SL.Era (SL.PreviousEra era)
```
nfrisby, line 320: I don't immediately recognize why this is necessary. (note to self)

ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
```diff
@@ -7,6 +7,7 @@
 {-# LANGUAGE LambdaCase           #-}
 {-# LANGUAGE NamedFieldPuns       #-}
 {-# LANGUAGE PatternSynonyms      #-}
+{-# LANGUAGE PolyKinds            #-}
```
nfrisby, line 10: I don't immediately recognize why this is necessary. (note to self)

ouroboros-consensus-test/src/Test/Util/TestBlock.hs
```diff
@@ -500,17 +500,17 @@ instance PayloadSemantics ptype => UpdateLedger (TestBlockWith ptype)
 instance InspectLedger (TestBlockWith ptype) where
   -- Defaults are fine
 
-instance (Typeable ptype, PayloadSemantics ptype) => HasAnnTip (TestBlockWith ptype) where
```
nfrisby, line 503: I don't immediately recognize why this is necessary. (note to self)

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -657,6 +660,22 @@ iters = bifoldMap (:[]) (const [])
 flrs :: Bitraversable t => t it flr -> [flr]
 flrs = bifoldMap (const []) (:[])
 
+{-------------------------------------------------------------------------------
```
nfrisby, line 663: Is GHC 9.2 more sensitive about how the TH splices delineate the module's declarations now?

ouroboros-consensus/src/Ouroboros/Consensus/HeaderStateHistory.hs
```diff
@@ -107,7 +107,7 @@ cast (HeaderStateHistory history) =
 -- rewinding that header state again by @j@ where @i + j > k@ is not possible
 -- and will yield 'Nothing'.
 rewind ::
-     forall blk. (BlockSupportsProtocol blk, HasAnnTip blk)
```
nfrisby, line 110: Was there a bug in GHC <9.2 that caused it to be missing unnecessary constraints or something?

nfrisby (approved):
I Approve the Consensus changes.

There are several bits I don't already understand the reason for (see my comments), but they're all minor, so not blocking this PR. Thanks!

nfrisby:
Just making sure GitHub notifies people about my edit just now to https://github.com/input-output-hk/ouroboros-network/pull/4144#pullrequestreview-1188175154 regarding Esgen's reviews.

ouroboros-consensus/src/Ouroboros/Consensus/HeaderStateHistory.hs
erikd: Yes, in earlier GHCs that was incorrectly considered an necessary constraint. It never was necessary and with 9.2 it is flagged as redundant.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
erikd: Yes.

erikd:
> Usually, PRs that touch both Network and Consensus should be reviewed by team members from both Consensus and Network (see also the codeowners) -- maybe we should make that more clear.

That should be resolved in the `CODEOWNERS` file. Please update it, or provide a diff on what it should be.

angerman:
> > who can review this? It's overwhelming mechanical changes to add 9.2 compatibility, and it's blocking downstream work on node.
> 
> 
> 
> I think the only thing missing now is GHC 9.2 on Nix/Hydra -- see https://github.com/input-output-hk/ouroboros-network/pull/4144#issuecomment-1322006034, and this is also the only place where we have `-Werror` enabled for the Consensus packages: While testing everything locally, I found a few warnings in the Consensus code (also pushed a commit fixing those, feel free to squash).

As we are going to lose hydra in a week anyway, I don't think we should spend much time on hydra right now. There is a different Cicero PR I believe.

angerman:
@amesgen, @nfrisby, @coot could we run those tests on GHA if we had our own runners?

amesgen (approved):
Thanks, LGTM!
  
> As we are going to lose hydra in a week anyway, I don't think we should spend much time on hydra right now. There is a different Cicero PR I believe.

Ok, I ran all tests locally using GHC 9.2, so I think it is fine to postpone it and include it in the mentioned #4108 :+1:

amesgen:
> @amesgen, @nfrisby, @coot could we run those tests on GHA if we had our own runners?

If they were sufficiently performant, I think that would remove the reason why we currently don't do that. Note however that the Network tests *are* run on GHA even today, in contrast to the Consensus tests.

amesgen:
bors retry

PR #4146, opened 2022-11-14
Let ledger determine whether Shelley forecasting succeeds

amesgen:
# Description

Remove a pure exception in Shelley forecasting by letting the ledger determine whether a forecasting operation succeeds instead of reimplementing the stability window check manually.

Historically, it seems this was done this way as `at`, the slot we are forecasting from, was not simply the tip slot of the given ledger state as it is now, but rather an explicit argument:

https://github.com/input-output-hk/ouroboros-network/blob/0934a3cb1e24ecbfcbab4e40522d99ffcd60feaf/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/Ledger.hs#L322-L326

TODO Potential implications to (double)check:

 - How fast does the Ledger short-circuit on failure?
 - Do we need the new `assert`?
 - Should we also assert that `for < maxFor` on success?
 - Right now, there is no obvious function `FutureLedgerViewError -> OutsideForecastRange`, espc. as `FutureLedgerViewError` is backed by a (not statically guaranteed to be empty!) list. How desirable is it to change this?

# Checklist

- Branch
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/SupportsProtocol.hs
+++ b/ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/SupportsProtocol.hs
@@ -22,6 +22,7 @@ module Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol () where
 
 import qualified Cardano.Ledger.Shelley.API as SL
 import qualified Cardano.Protocol.TPraos.API as SL
+import           Control.Exception (assert)
 import           Control.Monad.Except (MonadError (throwError))
 import           Data.Coerce (coerce)
 import           GHC.Records (getField)
@@ -59,10 +60,12 @@ instance
     if
         | NotOrigin for == at ->
           return $ TPraos.TickedPraosLedgerView $ SL.currentLedgerView shelleyLedgerState
-        | for < maxFor ->
-          return $ futureLedgerView for
+        | Right forecastedLedgerView <-
+            SL.futureLedgerView globals shelleyLedgerState for ->
+          return $ TPraos.TickedPraosLedgerView forecastedLedgerView
         | otherwise ->
           throwError $
+            assert (for >= maxFor)
             OutsideForecastRange
               { outsideForecastAt = at,
                 outsideForecastMaxFor = maxFor,
@@ -73,13 +76,6 @@ instance
       globals = shelleyLedgerGlobals cfg
       swindow = SL.stabilityWindow globals
       at = ledgerTipSlot ledgerState
-
-      futureLedgerView :: SlotNo -> Ticked (SL.LedgerView (EraCrypto era))
-      futureLedgerView =
-        either
-          (\e -> error ("futureLedgerView failed: " <> show e))
-          TPraos.TickedPraosLedgerView
-          . SL.futureLedgerView globals shelleyLedgerState
 
       -- Exclusive upper bound
       maxFor :: SlotNo


ouroboros-consensus-shelley/src/Ouroboros/Consensus/Shelley/Ledger/SupportsProtocol.hs
```diff
@@ -22,6 +22,7 @@ module Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol () where
```
nfrisby, line 22: > * How fast does the Ledger short-circuit on failure?
> * Do we need the new assert?
> * Should we also assert that for `< maxFor` on success?
> * Right now, there is no obvious function `FutureLedgerViewError -> OutsideForecastRange`, espc. as `FutureLedgerViewError` is backed by a (not statically guaranteed to be empty!) list. How desirable is it to change this?

* I think it would be pretty fast; using the `EpochInfo` to ask for the epoch of the slot is (currently) one of the first things the `TICKF` rule does. Hmm.

* I like the new assert.

* The `error` you removed essentially is an _assert that `, maxFor` on success_. Hence, I don't think we should be removing that `either (error ...)` call.

-----

This gets into tricky territory, in my opinion. Ideally, this "`OutsideForecastRange` iff slot is `> maxFor`" is something we'd like to prove. Unfortunately, even testing it confidently is difficult, because of the cardinality of the involved data types (esp in the rich `Cardano` type).

So, if the `futureLedgerView` somehow violates the "`OutsideForecastRange` iff `< maxFor`" spec, what do we want to ensure? The current code ensures the ledger won't accidentally forecast _too far_. On the other hand, if the ledger fails to forecast _far enough_, we raise a run-time error. The pure type of `ledgerViewForecast` currently doesn't allow for anything else, really---should we enrich it to that we could eg log a loud warning instead of crashing? Or is this the kind of bug that does justify crashing this high availability system?

amesgen: We discussed this in a call yesterday -- one crucial observation (also confirmed by actually running the code) is that the ledger will never return a `Left` from `SL.futureLedgerView` when the input is well-formed (which it always is from the side of consensus). Rather, it will throw a exception in pure code when it e.g. can't fetch information of the `for` slot from the `EpochInfo`.

Hence, the patch as it is written in this PR does not work, and will lead to pure `PastHorizon` exceptions being thrown when forecasting for slots beyond the stability window. @nfrisby is currently investigating adapting the ledger in such a way that it properly respects the "monadicity" of the `EpochInfo`, such that the approach of this PR (not having to reimplement the stability window check in Consensus) would actually work.

amesgen: FTR: This is futher explained in #4174. I will now repurpose this PR to add a comment pointing to that section.

amesgen: Hmm, that is literally a one line change, seems better to directly do that in #4174.

amesgen:
See https://github.com/input-output-hk/ouroboros-network/pull/4146#discussion_r1026172255

PR #4155, opened 2022-11-17
Refactor network & consensus packages

coot:
A refactorisation of `ouroboros-network` and `ouroboros-consensus`.  It includes the following packages:

* `ouroboros-network-api` - no other ouroboros depenencies
* `ouroboros-network-framework`
* `ouroboros-network-mock` - mock chain
* `ouroboros-network-protocols` - implementation of all (except Handshake) mini-protocols, also includes an `ouroboros-network-protocols:testlib` visible testing library.
* `ouroboros-network` - peer selection & diffusion
* `ouroboros-consensus-diffusion` - the top level package which was created out of `ouroboros-consensus`

pkgs


abailly-iohk:
The idea is to have a separate package that ties the know between network and consensus? Seems like a good idea to me.

abailly-iohk:
I am not too sure what's behind each package, it would help to group them as `Network` and `Consensus` I guess. 
My intuition is that the `Consensus` component and the `Network` component should depend on some `Interface` package that contains only the "interfaces" between the 2 components, and the "wiring" would be done in another component, possibly in the test packages or the node. 
But perhaps this is a different line of thought than yours @coot ?

coot:
Future tasks:

* the network team should move the simulated snocket into its own package, this might remove the `ouroboros-network-framework -> ouroboros-network-testing` dependency
* the consensus team might want to try remove the `ouroboros-consensus -> ouroboros-network-mock` dependency

Then the production code will be isolated from testing.

coot:
> But perhaps this is a different line of thought than yours @coot?

The API used by `ouroboros-consensus` is isolated in `ouroboros-network-api` package.  `ouroboros-consensus` depends on:
* `ouroboros-network-api`
* `ouroboros-network-protocols`

This is a bit masked by the fact that `ouroboros-network-protocols` also depends on `ouroboros-network-framework`.  I should look if it's easy to solve.

coot:
> I am not too sure what's behind each package, it would help to group them as Network and Consensus I guess.

They are grouped using naming convetion: all `network` packages match: `ouroboros-network*`
All `cosnensus` packages match `ouroboros-cosnensus*`.  Plus the top level `ouroboros-node`.  There are also generic packages like:
* `network-mux`
* `monoidal-synchronisation`
* `ntp-client`
* `cardano-client` - top level integration package for `cardano-db-sync` and maybe `cardano-wallet`

The first two will at some point be put in seprate git repos and published on `Hackage`.

coot:
`ouroboros-network-api` pulls `TraceLabelPeer` type from `network-mux` (no reason to solve this dependency).

coot:
Another remark:
`ouroboros-consensus-test` depends on `ouroboros-network-framework` because it pulls the [`Channel`](https://input-output-hk.github.io/ouroboros-network/ouroboros-network-framework/Ouroboros-Network-Channel.html#t:Channel) abstraction.

coot:
Requires #4159

ouroboros-node/ouroboros-node.cabal
```diff
@@ -0,0 +1,96 @@
+cabal-version:       3.0
+name:                ouroboros-node
```
nfrisby, line 2: Would you update the PR description to say `ouroboros-node` instead of `ouroboros-consensus-node`?

ouroboros-network-api/ouroboros-network-api.cabal
```diff
@@ -0,0 +1,83 @@
...
+                       -Wno-unticked-promoted-constructors
+                       -Wcompat
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+                       -Wpartial-fields
+                       -Widentities
+                       -Wredundant-constraints
+  if flag(asserts)
+    ghc-options:       -fno-ignore-asserts
+  build-depends:       text              >=1.2 && <1.3,
```
nfrisby, line 82: Why two different `build-depends` stanzas?

ouroboros-network-api/src/Ouroboros/Network/NodeToNode/Version.hs
```diff
@@ -145,8 +144,11 @@ nodeToNodeCodecCBORTerm _version
 
 data ConnectionMode = UnidirectionalMode | DuplexMode
 
+
 -- | Check whether a version enabling diffusion pipelining has been
 -- negotiated.
+--
+-- TODO: this ought to be defined in `ouroboros-consensus` or `ouroboros-consensus-node`
```
nfrisby, line 151: This comment says `ouroboros-consensus-node`, but the packaged is named `ouroboros-node`

ouroboros-network-api/src/Ouroboros/Network/SizeInBytes.hs
```diff
@@ -0,0 +1,5 @@
+module Ouroboros.Network.SizeInBytes (SizeInBytes) where
```
nfrisby, line 1: Seems like we could upstream this type definition to `cardano-base`, e?

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/BlockFetch/ClientInterface.hs
```diff
@@ -23,7 +23,10 @@ import           GHC.Stack (HasCallStack)
 import           Ouroboros.Network.AnchoredFragment (AnchoredFragment)
 import qualified Ouroboros.Network.AnchoredFragment as AF
 import           Ouroboros.Network.Block (MaxSlotNo)
-import           Ouroboros.Network.BlockFetch
+import           Ouroboros.Network.BlockFetch.ConsensusInterface
```
nfrisby, line 26: Glad to see this sort of refined import statement!

cardano-client/cardano-client.cabal
```diff
@@ -22,9 +22,10 @@ library
                        bytestring        >=0.10 && <0.11,
                        containers,
                        io-classes,
-                       ouroboros-consensus,
+                       ouroboros-network-api,
                        ouroboros-network,
                        ouroboros-network-framework,
+                       ouroboros-node,
```
nfrisby, line 28: Is `ouroboros-node` perhaps too general of a name? My line of thought:

> there are a _lot_ of things that the `cardano-node` package handles for Cardano that this `ouroboros-node` does _not_ handle for a general node based in Ouroboros

nix/ouroboros-network.nix
```diff
@@ -36,9 +36,26 @@ let
     modules = [
 
       {
-        # Compile all local packages with -Werror:
+        # Compile all local packages with -Werror and only
```
nfrisby, line 39: I suspect the rest of the Consensus team would also like to enable `--Wunused-packages` everywhere. Perhaps we can simplify this Nix by just setting the flag for all packages instead of just `-network` packages?

amesgen: I think we should set these flags in the respective `.cabal` files. Only `-Werror` is added via the Nix infra for CI as it is very inconvenient to have it enabled when developing locally.

ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -331,9 +319,9 @@ library
 
                      , io-classes       ^>=0.3
                      , typed-protocols
-                     , network-mux
-                     , ouroboros-network-framework ^>=0.2
-                     , ouroboros-network ^>=0.2
```
nfrisby, line 336: Let's remove the upper bounds in this `.cabal` file --- none of our others have them; I think they're appropriate in the `cabal.project` file now, right?

cc @dnadales @amesgen

nfrisby (commented):
I'm Approving the Consensus changes 👍 

I didn't click Approve, because someone else needs to do that for the Network changes.

ouroboros-consensus/ouroboros-consensus.cabal
nfrisby: We discussed this --- the plan is the _opposite_. The bounds are only temporarily in `cabal.project`.

My desiderata:

- each upstream package's bound is stated in _one place_ (this makes bumps trivial to do, and clear in diffs)
- there is some logic behind where that is

We had two ideas for this:

- choose one `.cabal` file as the "owner" of each upstream dep (eg `ouroboros-consensus-shelley.cabal` would own the Shelley-related ledger packages)

- reorganize all Consensus packages into a single `.cabal` file and use the multi-library support with reusable stanzas that include the bounded `build-depend`s

We'll address this shortly in a follow-up PR --- it's related to but beyond the scope of this one.

nfrisby (changes requested):
Ah, actually, the introduction of `ouroboros-node` requires an entry in `INTERFACE-changelog.md`.

ouroboros-network-api/ouroboros-network-api.cabal
coot: copy pasting bug :). fixed.

ouroboros-network-api/src/Ouroboros/Network/SizeInBytes.hs
coot: Yes, it makes sense.  Should it be a newtype? Then we could control which instances it has.

nix/ouroboros-network.nix
coot: I moved the `-Wunused-packages` to `*.cabal` files and removed this `nix` change.

cardano-client/cardano-client.cabal
coot: I renamed it to `ouroboros-consensus-diffusion`.

ouroboros-node/ouroboros-node.cabal
coot: There's no description, but also there's no `ouroboros-node` anywhere.

bolt12 (approved):
Looks good to me! I would just merge `ouroboros-network-protocols` & `ouroboros-network-protocols-test` into one package. I'd prefer if the tests for something would be in the same package as the functions/datatypes that are tested 😄

coot:
@bolt12 I merged the two packages.

ouroboros-node/ouroboros-node.cabal
nfrisby: By "PR description" I meant the thing at the top of this PR page. It looks like this to me:

image

ouroboros-network-api/src/Ouroboros/Network/SizeInBytes.hs
nfrisby: I don't have a strong preference, but I would lean `newtype` myself.

nfrisby (changes requested):
Thanks for addressing my comments.

There's one left: the typo in PR description (I commented there to try to clarify). I care because these end up in the merge commit messages.

Edit: this one https://github.com/input-output-hk/ouroboros-network/pull/4155#discussion_r1030790605

ouroboros-consensus/ouroboros-consensus.cabal
amesgen: See #4185

ouroboros-node/ouroboros-node.cabal
coot: The full graph is >10MB and is refused by github.

coot:
I fixed `ghc-9.2` support.  The `-Wunused-packages` was improved in `9.2` (as compared to `8.10`).

coot:
I also deleted the `script/test.sh`.

ouroboros-network-api/src/Ouroboros/Network/SizeInBytes.hs
nfrisby: I'm going to Approve without this being resolved.

nfrisby (approved):
Approved 👏

ouroboros-network-api/src/Ouroboros/Network/SizeInBytes.hs
coot: This deserves a separate PR, it will be every so slightly easier when rebasing other PRs over these changes.

coot:
bors merge

PR #4166, opened 2022-11-19
GHA: fix caching and others

angerman:
This tries to simplify the GHA further.
- fix the Caching logic, by using output instead of a mixture of output and env.
- investigate if we can use ghcup for consistency on linux _and_ windows, instead of haskell actions + ghcup on windows.


newhoggy (approved):

angerman:
bors merge

PR #4174, opened 2022-11-22
docs/HardWonWisdom.md: on how Ledger uses EpochInfo

nfrisby:
Starting a new document to capture small bites of knowledge as we accumulate them.

Step one is recording them at all; this file is supposed to make that easy.
Step two would be to occasionally extract cohesive insights from this bag, creating new documents or refining old ones accordingly.

### DIFF
--- a/ouroboros-consensus/docs/HardWonWisdom.md
+++ b/ouroboros-consensus/docs/HardWonWisdom.md
@@ -1,0 +1,101 @@
+This document contains a loosely organized list of small summaries of realizations we had while working on the code base.
+
+We do eventually want to integrate these lessons learned into more coherent and targetted documents.
+But step one one is to record them at all; this file is supposed to make that easy.
+Step two will be to occasionally extract cohesive insights from this bag, creating new documents or refining old ones accordingly.
+
+## Why doesn't Ledger code ever return `PastHorizonException`?
+
+One of the `HardForkBlock` combinator's major responsibilities is providing an `EpochInfo` to the ledger code.
+This `EpochInfo` uses the `Either` monad to return `Right` only when the query can be answered with certainty.
+For more information on when that is, see the Consensus Report and the recordings of Edsko's presentations at the Weekly IOG Seminar.
+
+However, most of the the ledger code that interacts with the given `EpochInfo` assumes it cannot fail by using `epochInfoPure`.
+
+```haskell
+data Globals = Globals { epochInfo :: !(EpochInfo (Either Text)), ... }  
+
+epochInfoPure :: Globals -> EpochInfo Identity
+epochInfoPure = hoistEpochInfo (either (throw . EpochErr) pure) . epochInfo
+```
+
+Thus, it is the responsibility of the calling code (eg the Consensus code) to check that the `HardForkBlock`-derived `EpochInfo` will not fail when its invoking the ledger rules.
+One example we've been looking at recently is the invocation of the `TICKF` rule, which is the `ledgerViewForecastAt` definition below.
+
+```haskell
+data Forecast a = Forecast {
+      forecastAt  :: WithOrigin SlotNo
+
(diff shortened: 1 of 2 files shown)


dnadales (approved):
Good idea, however, as you stated in the PR desc we should not forget to curate the relevant bit and pieces and place them in more appropriate places if there are any (eg code comments or the report).

ouroboros-consensus/docs/HardWonWisdom.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Why doesn't Ledger code ever return `PastHorizonException`?
+
+One of the `HardForkBlock` combinator's major responsibilities is providing an `EpochInfo` to the ledger code.
+This `EpochInfo` uses the `Either` monad to return `Right` only when the query can be answered with certainty.
+For more information on when that is, see the Consensus Report and the recordings of Edsko's presentations at the Weekly IOG Seminar.
+
+However, most of the the ledger code that interacts with the given `EpochInfo` assumes it cannot fail by using `epochInfoPure`.
+
+```
```
amesgen, line 11:
```suggestion
```haskell
```

ouroboros-consensus/docs/HardWonWisdom.md
```diff
@@ -0,0 +1,97 @@
...
+data Globals = Globals { epochInfo :: !(EpochInfo (Either Text)), ... }  
+
+epochInfoPure :: Globals -> EpochInfo Identity
+epochInfoPure = hoistEpochInfo (either (throw . EpochErr) pure) . epochInfo
+```
+
+Thus, it is the responsibility of the calling code (eg the Consensus code) to check that the `HardForkBlock`-derived `EpochInfo` will not fail when its invoking the ledger rules.
+One example we've been looking at recently is the invocation of the `TICKF` rule, which is the `ledgerViewForecastAt` definition below.
+
+```
```
amesgen, line 21:
```suggestion
```haskell
```

ouroboros-consensus/docs/HardWonWisdom.md
```diff
@@ -0,0 +1,97 @@
...
+It is not obvious how to do that when `EpochInfo` returns `Left`.
+While it may be possible to work something out, such as perhaps the `TICKF` rule emits a new `PredicateFailure` and simply leaves the `NewEpochState` unchanged, no idea seems natural to us.
+They all involve creating an incorrect `LedgerView` and then only throwing it away at the last moment.
+And notably, the rest of the rule could emit additional errors, all of which would be presumably spurious given that we already know the very first check (ie querying the `EpochInfo`) failed.
+
+So instead, for now at least, the Consensus code must do its own checking before invoking Ledger code that relies on the `EpochInfo` never failing.
+
+- By design, any invocation of Ledger code that causes the `EpochInfo` to return `Left` would be a _bug_, with no obvious way to recover.
+  Thus this isn't a particularly burdensome requirement; not much extra code (such as `for < maxFor` above).
+  And also pure exceptions are reasonable choice.
```
amesgen, line 82: Maybe call them "imprecise exceptions" as in the [original paper](https://www.microsoft.com/en-us/research/wp-content/uploads/1999/05/except.pdf)? Apparaently, they are sometimes also called "impure exceptions":

> A small point on terminology: in the safe-exceptions framework, imprecise exceptions are called “impure exceptions”. I am not fond of this terminology: as I’ve argued at the beginning of this post, imprecise exceptions are no more impure than infinite loops. They are just more useful.

https://www.tweag.io/blog/2020-04-16-exceptions-in-haskell/

amesgen (approved):
Thanks, very good exposition :+1:

I closed #4146 and pushed a tiny commit referencing this new section at the relevant code snippet, as also remarked by @dnadales.

amesgen:
bors retry

PR #4195, opened 2022-11-28
ouroboros-network-api: make SizeInBytes a newtype

coot:


ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Node/Serialisation.hs
```diff
@@ -75,7 +75,7 @@ instance Serialise ext => SerialiseNodeToNodeConstraints (MockBlock ext) where
       7 {- CBOR-in-CBOR -} + 1 {- encodeListLen 2 -} + hdrSize + bodySize
     where
       hdrSize  = fromIntegral (Lazy.length (serialise hdr))
-      bodySize = simpleBodySize (simpleHeaderStd hdr)
+      bodySize = fromIntegral (simpleBodySize (simpleHeaderStd hdr))
```
dnadales, line 78: Why do we need `fromIntegral` here?

ouroboros-network-api/src/Ouroboros/Network/SizeInBytes.hs
```diff
@@ -1,5 +1,15 @@
...
 import           Data.Word (Word32)
 
-type SizeInBytes = Word32
+import           NoThunks.Class (NoThunks (..))
+
+newtype SizeInBytes = SizeInBytes { getSizeInBytes :: Word32 }
+  deriving (Show, Eq, Ord)
+  deriving Enum     via Word32
+  deriving Num      via Word32
+  deriving Real     via Word32
```
dnadales, line 13: Do we need this instance? It seems strange that `SizeInBytes` is an instance of `Real`.

dnadales (commented):
Thanks! Only have one question:

What prevents us from addressing these TODO's is this PR?

ouroboros-network-api/src/Ouroboros/Network/SizeInBytes.hs
coot: `Real` just gives access to `toRational :: SizeInBytes -> Rational` which is useful when one needs to do some division.

ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Node/Serialisation.hs
coot: `estimateBlockSize` supposed to create `SizeInBytes` but `simpleBodySize` gives `Word32`.

coot:
> What prevents us from addressing these TODO's is this PR?

Because that's consensus internals, but feel free to fix it :smile:.

coot:
The TODO's introduced in this PR are recorded in #4216.

PR #4207, opened 2022-12-01
Add ReleaseProcess.md

### LINKED ISSUE #4204: Decide on a versioning scheme for the Consensus packages
The goal of this ticket is to decide on how to version packages in the release branches and on master, and how strict the inter-package bounds should be.

Related to #4185.

---

On Tuesday, we discussed this and came to the following preliminary conclusions:

 - Right now, we plan to version the consensus packages in two groups: one version for the "generic" consensus packages (first group above) and Cardano-specific ones (snd group)
Creating a new release for CHaP works by creating a commit in a release branch that bumps the relevant version: entries in the affected .cabal files which is then referenced in CHaP. Note that these commits are an exception to the usual rule that all commits should be backported/cherry-picked from master.
 - We proposed multiple ways how to manage the versions on the master branch:
    - Permanently unchanged bogus version, i.e. 0 or some high number.
    - The major version of the most recent active release branch.
    - GHC-like versioning scheme, i.e. master has odd major version numbers, while release branches have even ones, and the most active release branch has exactly one less than master . Example: 0.5 on master, 0.4.* on release branch, and 0.2.* on an older release branch.

nfrisby:
Closes #4204

I wrote the `ReleaseProcessDecision.md` to capture and edit the notes from the various couple sync calls where we discussed how to version our packages.

Those discussions expanded a bit into how to maintain/prepare releases. That results in the `ReleaseProcess.md` file, which documents our process---it's not just a record of/catalyst for discussion like `ReleaseProcessDecision.md` was.

### DIFF
--- a/ouroboros-consensus/docs/ReleaseProcess.md
+++ b/ouroboros-consensus/docs/ReleaseProcess.md
@@ -121,7 +121,7 @@ This scheme allows for multiple commits to declare their package has the same ve
 This means the mapping from commit hash <-> released version number is unfortunately not one-to-one.
 There is obviously some possibility for confusion there.
 However, we think the probability is low: if users only retrieve our code from the package repositories that we insert our release into, then the mapping will be one-to-one (as enforced by those repositories).
-Despite it being relatively easy to preclude this risk (add an extra `.0` dimension to the just-released versions immediately after announcing them, remove it when preparing the next release, but immediately add it back, etc -- thus between-release commits are always versions MAJOR.MINOR.PATCH.0), we're choosing not to.
+Despite it being relatively easy to preclude this risk (add an extra .0 dimension to the just-released versions immediately after announcing them, remove it when preparing the next release, but immediately add it back, etc -- thus between-release commits are always versions MAJOR.MINOR.PATCH.0), we're choosing not to.
 The mitigation is quite simple, but that extra version dimension is unfamiliar to the broader community and so is likely to cause confusion of its own.
 We are, though, ready to try this mitigation if the lack of one-to-one mapping does end up causing enough confusion.
 
@@ -155,34 +155,34 @@ Consider just one package, FOO.
 Suppose we have released the following versions: 1.0.0 and 2.0.0.
 The commit history might look like the following linear chain, with the tags and branch pointers annotated.
 
-`
+```
 D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0, branch: release-FOO-2.0.x, branch: main)
 C - replace that confusing feature with much nicer one
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
 A - ...
-`
+```
 
 If we subsequently merge a PR that fixes a minor bug present in 2.0.0, it'd then look like this.
 
-`
+```
 E - fix that off-by-one error (branch: main)
 D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0, branch: release-FOO-2.0.x)
 C - replace that confusing feature with much nicer one
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
 A - ...
-`
+```
 
-We could have also advanced `release-FOO-2.0.x`, but the relevent RULES above do not _require_ doing so until we release version 2.0.1.
+We could have also advanced release-FOO-2.0.x, but the relevent RULES above do not _require_ doing so until we release version 2.0.1.
 Once we do that, we'd have the following.
 
-`
+```
 F - the release of FOO-2.0.1 (tag: release-FOO-2.0.1, branch: release-FOO-2.0.x, branch: main)
 E - fix that off-by-one error
 D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0)
 C - replace that confusing feature with much nicer one
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
 A - ...
-`
+```
 
 We must advance release-FOO-2.0.x to the F commit, because we released F as a version 2.0.X and so the RULE above requires that the F commit is included in the first-parent history of release-FOO-2.0.x.
 
@@ -191,39 +191,39 @@ We might want to fix it because our users aren't yet ready to integrate the new 
 Thus, we'd merge a bugfix PR directly into the release-1.0.x branch.
 Note that we should merge bugfixes into main branch and then backport them onto a release branch when we can, but in this example the buggy feature no longer exists on the main branch.
 
-`
+```
 G - fix confusion in the old logic (branch: release-FOO-1.0.x)
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
 A - ...
-`
+```
 
 Once that patch passes validation, we could then release 1.0.1.
 
-`
+```
 H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1, branch: release-FOO-1.0.x)
 G - fix confusion in the old logic
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
 A - ...
-`
+```
 
 Now suppose we could have easily backported E onto 1.0.x as well.
 If we're making another release of the 1.0 version, our users would likely appreciate us including as many bugfixes as we can.
 
-`
+```
 I - cherry-pick of E (branch: release-FOO-1.0.x)
 H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1)
 G - fix confusion in the old logic
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
 A - ...
-`
+```
 
 And shortly the following.
 
-`
+```
 J - the release of FOO-1.0.2 (tag: release-FOO-1.0.2, branch: release-FOO-1.0.x)
 I - cherry-pick of E
 H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1)
 G - fix confusion in the old logic
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
 A - ...
-`
+```
--- a/ouroboros-consensus/docs/ReleaseProcess.md
+++ b/ouroboros-consensus/docs/ReleaseProcess.md
@@ -22,7 +22,7 @@ If you search for "RULE:" in this document, you'll find the key statements.
 ## Rules for Branches
 
 The Consensus team will sometimes be maintaining multiple MAJOR.MINOR versions of the same package simultaneously.
-EG We might release FOO-2.4.1 on Monday, release FOO-2.3.7 on Wednesday, and release FOO-2.4.2 on Friday, etc.
+EG If we were currently maintaining a 2.3 version and a 2.4 version simultaneously, then we might release FOO-2.4.1 on Monday, release FOO-2.3.7 on Wednesday, and release FOO-2.4.2 on Friday, etc.
 
 - Whenever we're maintaining only the one greatest-ever MAJOR.MINOR version of a package, then all work on that package will happen on the main branch.
 - Whenever we're also maintaining some lesser MAJOR.MINOR versions of a package, then some work for that package will also be happening on the respective _release branches_.
@@ -32,6 +32,7 @@ A _release branch_ is a branch dedicated to the maintenance of some older releas
 RULE: If we have released a package FOO-X.Y.Z, then the release-FOO-X.Y.x branch MUST exist and its first-parent history MUST include the commit released as FOO-X.Y.Z.
 For example, when releasing version 2.3.0 of a package named FOO, we must create the branch named release-FOO-2.3.x.
 And when we later release 2.3.1 of FOO, then we'd need to advance the release-FOO-2.3.x branch to point at the commit being released as FOO-2.3.1.
+See the "Release Branch Example" below for a more thorough example.
 
 There are two kinds of work on release branches.
 Most of the time, that work is immitating some work that was done on the main branch.
@@ -45,13 +46,13 @@ EG Consider the following possible timeline.
 - We release FOO-2.3.7 on Wednesday.
 
 *Remark*.
-Not every first-parent commit in the history of the release branch is a release commit.
+Not every first-parent commit in the history of the release branch was announced as a release of a package.
 We will be merging multiple PRs into the release branch in order to prepare the next release from it, so some commits will just be the intermediates between two releases.
 
 *Remark*.
 The release branch for the greatest MAJOR.MINOR version of a package is somewhat degenerate.
 It's either equal to or a prefix of the main branch.
-The only time it MUST be updated to the tip of the main branch is when we cut a release of the package from the main branch, to satisfy the rule above about all X.Y.Z release commits being on the release-X.Y.x branch.
+The only time it MUST be updated to the tip of the main branch is when we cut a release of the package from the main branch, to satisfy the rule above about all commits that were released as some version X.Y.Z being on the release-X.Y.x branch.
 
 ## Rules for PRs
 
@@ -106,6 +107,7 @@ We prepare that release as follows.
 - RULE: We tag that resulting merge commit as release-PKG-A.B.C; one tag per package PKG that is being released (ie FOO and/or BAR).
 - RULE: If C is 0, then we also create the release branch release-PKG-A.B.x from this new commit.
     - For example, when releasing version 2.3.0 of a package named FOO, we'd create the branch named release-FOO-2.3.x.
+      See the "Release Branch Example" below.
 - RULE: Finally, we announce this commit hash as the new release of these packages.
   EG We insert these package's new versions into Hackage, [CHaP](https://github.com/input-output-hk/cardano-haskell-packages), etc.
 
@@ -142,3 +144,86 @@ The only refinement to the wording in the above sections is that we update the `
 This means some packages will occasionally get vacuous versions bumps.
 That's obnoxious to downstream users and would be easy to interpret as evidence the Consensus team has made a mistake in during that release.
 However, we either mitigate this by choosing our bundle partitioning to make it unlikely (put coupled packages into the same bundle) or else accept the oddity of the vacuous version bumps as an acceptable cost for the bundling's other advantages.
+
+## Some Concrete Examples
+
+Some of the rules above are much easier to internalize via concrete examples.
+
+### Release Branch Example
+
+Consider just one package, FOO.
+Suppose we have released the following versions: 1.0.0 and 2.0.0.
+The commit history might look like the following linear chain, with the tags and branch pointers annotated.
+
+`
+D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0, branch: release-FOO-2.0.x, branch: main)
+C - replace that confusing feature with much nicer one
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
+A - ...
+`
+
+If we subsequently merge a PR that fixes a minor bug present in 2.0.0, it'd then look like this.
+
+`
+E - fix that off-by-one error (branch: main)
+D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0, branch: release-FOO-2.0.x)
+C - replace that confusing feature with much nicer one
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
+A - ...
+`
+
+We could have also advanced `release-FOO-2.0.x`, but the relevent RULES above do not _require_ doing so until we release version 2.0.1.
+Once we do that, we'd have the following.
+
+`
+F - the release of FOO-2.0.1 (tag: release-FOO-2.0.1, branch: release-FOO-2.0.x, branch: main)
+E - fix that off-by-one error
+D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0)
+C - replace that confusing feature with much nicer one
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
+A - ...
+`
+
+We must advance release-FOO-2.0.x to the F commit, because we released F as a version 2.0.X and so the RULE above requires that the F commit is included in the first-parent history of release-FOO-2.0.x.
+
+Suppose we then find a bad bug in feature that commit C had replace.
+We might want to fix it because our users aren't yet ready to integrate the new feature from 2.0 version into their code.
+Thus, we'd merge a bugfix PR directly into the release-1.0.x branch.
+Note that we should merge bugfixes into main branch and then backport them onto a release branch when we can, but in this example the buggy feature no longer exists on the main branch.
+
+`
+G - fix confusion in the old logic (branch: release-FOO-1.0.x)
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
+A - ...
+`
+
+Once that patch passes validation, we could then release 1.0.1.
+
+`
+H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1, branch: release-FOO-1.0.x)
+G - fix confusion in the old logic
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
+A - ...
+`
+
+Now suppose we could have easily backported E onto 1.0.x as well.
+If we're making another release of the 1.0 version, our users would likely appreciate us including as many bugfixes as we can.
+
+`
+I - cherry-pick of E (branch: release-FOO-1.0.x)
+H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1)
+G - fix confusion in the old logic
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
+A - ...
+`
+
+And shortly the following.
+
+`
+J - the release of FOO-1.0.2 (tag: release-FOO-1.0.2, branch: release-FOO-1.0.x)
+I - cherry-pick of E
+H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1)
+G - fix confusion in the old logic
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
+A - ...
+`
--- a/ouroboros-consensus/docs/ReleaseProcess.md
+++ b/ouroboros-consensus/docs/ReleaseProcess.md
@@ -29,9 +29,9 @@ EG If we were currently maintaining a 2.3 version and a 2.4 version simultaneous
 
 A _release branch_ is a branch dedicated to the maintenance of some older release of a package.
 
-RULE: If we have released a package FOO-X.Y.Z, then the release-FOO-X.Y.x branch MUST exist and its first-parent history MUST include the commit released as FOO-X.Y.Z.
-For example, when releasing version 2.3.0 of a package named FOO, we must create the branch named release-FOO-2.3.x.
-And when we later release 2.3.1 of FOO, then we'd need to advance the release-FOO-2.3.x branch to point at the commit being released as FOO-2.3.1.
+RULE: If we have released a package FOO-X.Y.Z, then the release-FOO-X.x.x branch MUST exist and its first-parent history MUST include the commit released as FOO-X.Y.Z.
+For example, when releasing version 2.0.0 of a package named FOO, we must create the branch named release-FOO-2.x.x.
+And when we later release 2.3.1 of FOO (and we've never released a 2.Y.Z with Y>3), then we'd need to advance the release-FOO-2.x.x branch to point at the commit being released as FOO-2.3.1.
 See the "Release Branch Example" below for a more thorough example.
 
 There are two kinds of work on release branches.
@@ -52,7 +52,7 @@ We will be merging multiple PRs into the release branch in order to prepare the 
 *Remark*.
 The release branch for the greatest MAJOR.MINOR version of a package is somewhat degenerate.
 It's either equal to or a prefix of the main branch.
-The only time it MUST be updated to the tip of the main branch is when we cut a release of the package from the main branch, to satisfy the rule above about all commits that were released as some version X.Y.Z being on the release-X.Y.x branch.
+The only time it MUST be updated to the tip of the main branch is when we cut a release of the package from the main branch, to satisfy the rule above about all commits that were released as some version X.Y.Z being on the release-X.x.x branch.
 
 ## Rules for PRs
 
@@ -105,8 +105,8 @@ We prepare that release as follows.
     - RULE: It updates the versions of the packages being released as described above.
     - RULE: It flushes the pending changelog entries of each package being released into the `CHANGELOG.md` file for that package.
 - RULE: We tag that resulting merge commit as release-PKG-A.B.C; one tag per package PKG that is being released (ie FOO and/or BAR).
-- RULE: If C is 0, then we also create the release branch release-PKG-A.B.x from this new commit.
-    - For example, when releasing version 2.3.0 of a package named FOO, we'd create the branch named release-FOO-2.3.x.
+- RULE: If C is 0, then we also create the release branch release-PKG-A.x.x from this new commit.
+    - For example, when releasing version 2.0.0 of a package named FOO, we'd create the branch named release-FOO-2.x.x.
       See the "Release Branch Example" below.
 - RULE: Finally, we announce this commit hash as the new release of these packages.
   EG We insert these package's new versions into Hackage, [CHaP](https://github.com/input-output-hk/cardano-haskell-packages), etc.
@@ -156,9 +156,9 @@ Suppose we have released the following versions: 1.0.0 and 2.0.0.
 The commit history might look like the following linear chain, with the tags and branch pointers annotated.
 
 ```
-D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0, branch: release-FOO-2.0.x, branch: main)
+D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0, branch: release-FOO-2.x.x, branch: main)
 C - replace that confusing feature with much nicer one
-B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.x.x)
 A - ...
 ```
 
@@ -166,33 +166,33 @@ If we subsequently merge a PR that fixes a minor bug present in 2.0.0, it'd then
 
 ```
 E - fix that off-by-one error (branch: main)
-D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0, branch: release-FOO-2.0.x)
+D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0, branch: release-FOO-2.x.x)
 C - replace that confusing feature with much nicer one
-B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.x.x)
 A - ...
 ```
 
-We could have also advanced the release-FOO-2.0.x branch, but the relevant RULES above do not _require_ doing so until we release version 2.0.1.
+We could have also advanced the release-FOO-2.x.x branch, but the relevant RULES above do not _require_ doing so until we release version 2.0.1.
 Once we do that, we'd have the following.
 
 ```
-F - the release of FOO-2.0.1 (tag: release-FOO-2.0.1, branch: release-FOO-2.0.x, branch: main)
+F - the release of FOO-2.0.1 (tag: release-FOO-2.0.1, branch: release-FOO-2.x.x, branch: main)
 E - fix that off-by-one error
 D - the release of FOO-2.0.0 (tag: release-FOO-2.0.0)
 C - replace that confusing feature with much nicer one
-B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
+B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.x.x)
 A - ...
 ```
 
-We must advance release-FOO-2.0.x to the F commit, because we released F as a version 2.0.X and so the RULE above requires that the F commit is included in the first-parent history of release-FOO-2.0.x.
+We must advance release-FOO-2.x.x to the F commit, because we released F as a version 2.0.X and so the RULE above requires that the F commit is included in the first-parent history of release-FOO-2.x.x.
 
 Suppose we then find a bad bug in feature that commit C had replaced.
 We might want to fix it because our users aren't yet ready to integrate the new feature from 2.0 version into their code.
-Thus, we'd merge a bugfix PR directly into the release-1.0.x branch.
+Thus, we'd merge a bugfix PR directly into the release-1.x.x branch.
 Note that we should merge bugfixes into the main branch and then backport them onto a release branch when that is possible, but in this case the buggy feature no longer exists on the main branch.
 
 ```
-G - fix confusion in the old logic (branch: release-FOO-1.0.x)
+G - fix confusion in the old logic (branch: release-FOO-1.x.x)
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
 A - ...
 ```
@@ -200,17 +200,17 @@ A - ...
 Once that patch passes validation, we could then release 1.0.1.
 
 ```
-H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1, branch: release-FOO-1.0.x)
+H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1, branch: release-FOO-1.x.x)
 G - fix confusion in the old logic
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
 A - ...
 ```
 
-Now suppose we belatedly realized we can easily backport E onto 1.0.x as well.
+Now suppose we belatedly realized we can easily backport E onto 1.x.x as well.
 If we're making another release of the 1.0 version, our users would likely appreciate us including as many bugfixes as we can.
 
 ```
-I - cherry-pick of E (branch: release-FOO-1.0.x)
+I - cherry-pick of E (branch: release-FOO-1.x.x)
 H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1)
 G - fix confusion in the old logic
 B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
@@ -220,7 +220,7 @@ A - ...
 And shortly the following.
 
 ```
-J - the release of FOO-1.0.2 (tag: release-FOO-1.0.2, branch: release-FOO-1.0.x)
+J - the release of FOO-1.0.2 (tag: release-FOO-1.0.2, branch: release-FOO-1.x.x)
 I - cherry-pick of E
 H - the release of FOO-1.0.1 (tag: release-FOO-1.0.1)
 G - fix confusion in the old logic
--- a/ouroboros-consensus/docs/ReleaseProcess.md
+++ b/ouroboros-consensus/docs/ReleaseProcess.md
@@ -172,7 +172,7 @@ B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0, branch: release-FOO-1.0.x)
 A - ...
 ```
 
-We could have also advanced release-FOO-2.0.x, but the relevent RULES above do not _require_ doing so until we release version 2.0.1.
+We could have also advanced the release-FOO-2.0.x branch, but the relevant RULES above do not _require_ doing so until we release version 2.0.1.
 Once we do that, we'd have the following.
 
 ```
@@ -186,10 +186,10 @@ A - ...
 
 We must advance release-FOO-2.0.x to the F commit, because we released F as a version 2.0.X and so the RULE above requires that the F commit is included in the first-parent history of release-FOO-2.0.x.
 
-Suppose we then find a bad bug in feature that commit C had replace.
+Suppose we then find a bad bug in feature that commit C had replaced.
 We might want to fix it because our users aren't yet ready to integrate the new feature from 2.0 version into their code.
 Thus, we'd merge a bugfix PR directly into the release-1.0.x branch.
-Note that we should merge bugfixes into main branch and then backport them onto a release branch when we can, but in this example the buggy feature no longer exists on the main branch.
+Note that we should merge bugfixes into the main branch and then backport them onto a release branch when that is possible, but in this case the buggy feature no longer exists on the main branch.
 
 ```
 G - fix confusion in the old logic (branch: release-FOO-1.0.x)
@@ -206,7 +206,7 @@ B - the release of FOO-1.0.0 (tag: release-FOO-1.0.0)
 A - ...
 ```
 
-Now suppose we could have easily backported E onto 1.0.x as well.
+Now suppose we belatedly realized we can easily backport E onto 1.0.x as well.
 If we're making another release of the 1.0 version, our users would likely appreciate us including as many bugfixes as we can.
 
 ```
--- a/ouroboros-consensus/docs/ReleaseProcess.md
+++ b/ouroboros-consensus/docs/ReleaseProcess.md
@@ -30,6 +30,8 @@ EG We might release FOO-2.4.1 on Monday, release FOO-2.3.7 on Wednesday, and rel
 A _release branch_ is a branch dedicated to the maintenance of some older release of a package.
 
 RULE: If we have released a package FOO-X.Y.Z, then the release-FOO-X.Y.x branch MUST exist and its first-parent history MUST include the commit released as FOO-X.Y.Z.
+For example, when releasing version 2.3.0 of a package named FOO, we must create the branch named release-FOO-2.3.x.
+And when we later release 2.3.1 of FOO, then we'd need to advance the release-FOO-2.3.x branch to point at the commit being released as FOO-2.3.1.
 
 There are two kinds of work on release branches.
 Most of the time, that work is immitating some work that was done on the main branch.
@@ -103,12 +105,14 @@ We prepare that release as follows.
     - RULE: It flushes the pending changelog entries of each package being released into the `CHANGELOG.md` file for that package.
 - RULE: We tag that resulting merge commit as release-PKG-A.B.C; one tag per package PKG that is being released (ie FOO and/or BAR).
 - RULE: If C is 0, then we also create the release branch release-PKG-A.B.x from this new commit.
+    - For example, when releasing version 2.3.0 of a package named FOO, we'd create the branch named release-FOO-2.3.x.
 - RULE: Finally, we announce this commit hash as the new release of these packages.
   EG We insert these package's new versions into Hackage, [CHaP](https://github.com/input-output-hk/cardano-haskell-packages), etc.
 
 *Remark*.
-To explicitly clarify: after a release of a package, there will be zero pending changelog entries for that package; they were destroyed when incorprated into the `CHANGELOG.md` file making the release.
-But there may be pending changelog entries of other packages, those that we're included in this release.
+To explicitly clarify: after a release of a package, there will be zero pending changelog entries for that package.
+When making the release, those entries were first incorporated into the `CHANGELOG.md` file and then the individual files containing those pending entires (see `scriv`'s mechanism) were removed (a la `git rm`).
+But there may be pending changelog entries of other packages, those that were not included in this release.
 
 *Remark*.
 This scheme allows for multiple commits to declare their package has the same version number as some released version, even if those commits have made alterations to the package.
--- a/ouroboros-consensus/docs/ReleaseProcess.md
+++ b/ouroboros-consensus/docs/ReleaseProcess.md
@@ -1,0 +1,140 @@
+# How Consensus Makes Releases
+
+This document explains how the Consensus team uses Pull Requests (PRs), version numbers, and branches to prepare releases of our packages.
+
+Let us assume our repository contains two packages FOO and BAR.
+Two is enough to explain the complexities.
+The resulting rules can be generalized to any number of packages, as explained at the end of the document.
+
+*Remark*.
+If you disagree with the decisions in this document, please consider reviewing our brainstorming notes and discussions, available at [this document's PR](https://github.com/input-output-hk/ouroboros-network/pull/4207).
+Even if you choose not to do so, please share your concern with us.
+
+## Notation
+
+In this document, we will assume each version number is a MAJOR.MINOR.PATCH triple.
+This is easy to generalize to other conventional notations, such as [the PVP](https://pvp.haskell.org/)'s MAJOR.MAJOR.MINOR.PATCH quadruples.
+
+We refer throughout this document to _the main branch_ (aka `master`, aka `main`).
+
+If you search for "RULE:" in this document, you'll find the key statements.
+
+## Rules for Branches
+
+The Consensus team will sometimes be maintaining multiple MAJOR.MINOR versions of the same package simultaneously.
+EG We might release FOO-2.4.1 on Monday, release FOO-2.3.7 on Wednesday, and release FOO-2.4.2 on Friday, etc.
+
+- Whenever we're maintaining only the one greatest-ever MAJOR.MINOR version of a package, then all work on that package will happen on the main branch.
+- Whenever we're also maintaining some lesser MAJOR.MINOR versions of a package, then some work for that package will also be happening on the respective _release branches_.
+
+A _release branch_ is a branch dedicated to the maintenance of some older release of a package.
+
+RULE: If we have released a package FOO-X.Y.Z, then the release-FOO-X.Y.x branch MUST exist and its first-parent history MUST include the commit released as FOO-X.Y.Z.
+
+There are two kinds of work on release branches.
+Most of the time, that work is immitating some work that was done on the main branch.
+EG we fixed a bug on the main branch and we're backporting that fix onto a release branch (hopefully it's a trivial cherry-pick).
+Rarely, though, there will be fresh work done on a release branch.
+EG Consider the following possible timeline.
+
+- We release FOO-2.4.1 on Monday, and FOO-2.4.1 completely reworked some functionality.
+- We then realize on Tuesday that there was a severe bug in the old functionality.
+- We fix the bug by merging a PR that targets the FOO-2.3.x release branch (since the bug no longer exists on the main branch!)
+- We release FOO-2.3.7 on Wednesday.
+
+*Remark*.
+Not every first-parent commit in the history of the release branch is a release commit.
+We will be merging multiple PRs into the release branch in order to prepare the next release from it, so some commits will just be the intermediates between two releases.
+
+*Remark*.
+The release branch for the greatest MAJOR.MINOR version of a package is somewhat degenerate.
+It's either equal to or a prefix of the main branch.
+The only time it MUST be updated to the tip of the main branch is when we cut a release of the package from the main branch, to satisfy the rule above about all X.Y.Z release commits being on the release-X.Y.x branch.
+
+## Rules for PRs
+
+We classify each PR as either a _fresh_ PR or as a _backport_ PR.
+
+- A fresh PR is doing work that has never been done before (new feature, new bugfix, etc).
+  Except in the rare circumstances discussed in the previous section, every fresh PR will target the main branch.
+- The primary objective of a backport PR is merely to immitate some fresh PR.
+
+The rules are then as follows.
+
+- RULE: A fresh PR MUST target the main branch, if possible.
+- RULE: A PR MUST NOT be merged into a branch unless it should be included in the next release from that branch.
+- We maintain a changelog for each package (one for FOO and one for BAR).
+- RULE: A PR MUST add a pending changelog entry for each package that it alters.
+- RULE: Each pending changelog entry MUST at the very least classify the alteration to that package as a _patch_ level change (eg a bugfix), a _minor_ change (eg a non-breaking interface change), or a _major_ change (eg a breaking interface change).
+
+*Remark*.
+Notice that we do not allow merging a PR that should not be included the subsequent release.
+This means some PR may be "done" but still should not be merged.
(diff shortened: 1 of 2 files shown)


ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,34 @@
+# Choosing a Versioning Schemes
+
+This document records our discussions around choosing how to version our packages.
+
+## Desiderata
+
+- *Simplicity and Familiarity*. We'd like the versioning scheme to be simple to explain, and ideally already well-established.
+
+- *Ease of execution*. We'd like the process of cutting a release to merely involve following a very simple checklist. We'd like it to require as few inputs, discussions, decisions, etc.
```
nfrisby, line 9: The main thing we've discussed that is missing from the document is deciding that some packages will advance in lock-step. We had considered this partitioning:

```
ouroboros-consensus.cabal
ouroboros-consensus-diffusion.cabal
ouroboros-consensus-test.cabal
ouroboros-consensus-mock.cabal
ouroboros-consensus-mock-test.cabal
ouroboros-consensus-protocol.cabal

ouroboros-consensus-byron.cabal
ouroboros-consensus-byron-test.cabal
ouroboros-consensus-byronspec.cabal
ouroboros-consensus-cardano.cabal
ouroboros-consensus-cardano-test.cabal
ouroboros-consensus-cardano-tools.cabal
ouroboros-consensus-shelley.cabal
ouroboros-consensus-shelley-test.cabal```

File: ouroboros-consensus/docs/VersioningSchemeDecision.md [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+
+Minor versions will be odd for packages on master and even for releases (like the GHC Team's scheme).
+To cut a release, create branch release/XXX pointing at the desired master commit and then add a commit to release/XXX that bumps all versions to the next even number, and also immediately create a subsequent master commit that advances all the appropriate versions to the next odd number.
+
+## Proposal NonZero
+
+Master always has degenerate versions on it: everything is version 0.
+To cut a release, create branch release/XXX pointing at the desired master commit and then add a commit to release/XXX that advances all appropriate versions COMPARED TO their value in the previous release branch.
+
+## Proposal Dimension
```
nfrisby, line 31: I currently favor `Proposal Dimension`, FWIW.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,34 @@
+# Choosing a Versioning Schemes
+
+This document records our discussions around choosing how to version our packages.
```
nfrisby, line 3: Once we've made a decision inside Consensus, we can invite Marcin et al to ask whether they agree/can refine?

ouroboros-consensus/docs/VersioningSchemeDecision.md
yogeshsajanikar: I favor this one too! It also allows adding patch on release (hotfix).

abailly-iohk (commented):
Why create release branches? Why not simply _tag_ the release from master? Do you foresee a situation where you would need to fix specific versions and not others? Consensus is a library which currently is only used in a single place (cardano-node).
For the quickcheck-dynamic library, we've setup a simple scheme where versions are just tags on master and pushed to hackage from CI: https://github.com/input-output-hk/quickcheck-dynamic/blob/main/.github/workflows/release.yaml 
Perhaps a simple scheme like that could work for consensus?
Also, are you using SemVer or PVP?

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,41 @@
...
+
+- *Distinguished development versions*. We'd like to distinguish between the two possible semantics of version number.
+
+    - The version of a released thing identifies some immutable thing, which is always somehow more refined than any release that has a lesser version number.
+
+    - A development version refers to some _mutable_ thing that is improving during the time between two releases, eg the version on master. IE there will usually be multiple different commits that all have the same development version number.
+
+## Proposal Simplest
+
+To cut a release, merely create branch release/XXX pointing at the desired commit on master and also immediately create a subsequent master commit that advances all the appropriate versions.
```
amesgen, line 19: > create branch release/XXX

Just to be sure: In all proposals listed here, cutting a release involves "create branch release/XXX", i.e. you would create a new branch for every version -- is this intentional?

E.g. right now, we just have *one* release branch ([`release/cardano-node-1.35.x`](https://github.com/input-output-hk/ouroboros-network/tree/release/cardano-node-1.35.x)), which was forked off of `master`, and only contains backported commits and version bumps, which are referenced in CHaP.

ouroboros-consensus/docs/VersioningSchemeDecision.md
amesgen: > It also allows adding patch on release (hotfix).

I think whether we are able to create patch releases is mostly orthogonal to the different proposals here -- i.e. Proposal Dimension and Proposal NonZero only differ in what version is written down in the `.cabal` files on `master`, but the versions in CHaP would/could be exactly the same.

ouroboros-consensus/docs/VersioningSchemeDecision.md
lehins: I'd suggest using tags, instead of branches for releases.

coot: I am also in favour of `tags`, but branches might be needed at times so I wouldn't rule them out.  This is because the networking team might push changes to master which we might not want to release in a hot fix release, but rather a major release; or the other way around: like currently with `1.35.5` where I needed to push >100 commits to the release branch to avoid releasing consensus changes.  *That's actually another argument to split network & consensus into two repos (I am starting to convince myself that's actually a good thing to do).*

If we had two repositories, we would integrate network changes to master, but then to try it out, we would likely cherry pick changes on the most recent `ouroboros-consensus` released with `cardano-node`.  Since now only `ouroboros-consensus-diffusion` depends on `ouroboros-network`, merging our changes to `master` might be enough to avoid using not yet released `ouroboros-consensus` with our new stuff. If we add a new version of `NodeToNodeVersion` (or `NodeToClientVersion`) then we will also need a newer version of `ouroboros-consensus-cardano` for which we might need to use a release branch.

lehins: > This is because the networking team might push changes to master which we might not want to release in a hot fix release, but rather a major release; or the other way around: 

In such situation you can always create a branch from a tag, work on it and once such fix has finalized and release was made it should be turned into another tag. Branches are intended for mutable things, while tags for immutable, such as releases.

coot: Do you want to release all the consensus packages at the same time, or only release packages which have some changes? Probably the former, but it's better to be explicit about it.

nfrisby: Sorry, yes, I wrote too loosely. I should have consistently said:

> To cut a release of a branch new minor version, create a branch release/foo.xx and ...

I think that's our plan (and we'll tag increments within that branch with concrete verssions `release/foo.1.37.3`).

lehins:
I always assumed versioning scheme should be decided at the CHaP's level. This way all packages in IO could conform to it and there would be no surprises or inconsistencies. I was a strong proponent of PVP for Haskell packages, as the rest of the Haskell community relies on it.

Some things discussed in this document go against common versioning schemes like PVP and SemVer. For example if you release all packages in the repo at once and they all have the same version, then you are back to the same scheme as it was prior to CHaPs using one version for the whole repo (previously it was git sha), except with a more convoluted mechanism.

ouroboros-consensus/docs/VersioningSchemeDecision.md
lehins: :slightly_smiling_face: perhaps `release/foo-A.B.3`

ouroboros-consensus/docs/VersioningSchemeDecision.md
coot: I also like this one the most, it's quite clean.  It has the nice side effect that all your packages will have the same `major` version if I understand correctly.  This makes it quite nice for downstream dependencies.  Further more if you will have some reasonable constraints between all your packages, then declaration of which version of `consensus` is used might be as simple as constraining only one of `consensus` packages in `cardano-node` repo.

nfrisby:
Both https://github.com/input-output-hk/ouroboros-network/pull/4207#pullrequestreview-1201498810 and https://github.com/input-output-hk/ouroboros-network/pull/4207#issuecomment-1334403218 seem out of scope to me.

That's probably my fault, for using shorthand (approximately _synecdoche_) in the write-up.

To clarify, this PR is focused on question: for each individual package, how should the version listed in the `.cabal` file on the `master` branch relate to the version in the `.cabal` file in the latest release?

In particular, I think this question is almost entirely independent of branches versus tags, SemVer versus PVP, etc.

Specific conflicts:

- I think SemVer requires `x.y.z`, and Proposal Dimensions would only have `x.y` on the `master` branch, but every _release_ would be `x.y.z`. So that seems compatible to me.

- SemVer and PVP, as I recall, at least _primarily_ focus on  when to increment which dimension of the version. However, I don't think it would be a earth-shattering violation to increase the version by more than one (cf Proposal Parity), or to increase it more than strictly necessary (cf the comment about keeping subsets of our packages in lock-step).

Thanks for the feedback! I am happy to have it, just not necessarily on _this_ PR :D (again: my fault).

ch1bo (commented):
Not sure if any of your proposals already covers this approach: Whenever a change is done on `master` (or a feature branch before), consider how it impacts the version number, i.e. bump the corresponding digit for breaking changes or patches when you do break an interface or just patch something. The `master` version then indicates the version number which would be true if you just release today (e.g. by tagging a commit). We use that also in [hydra](https://github.com/input-output-hk/hydra/blob/master/CONTRIBUTING.md#versioning--changelog)

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,54 @@
...
+This document records our discussions around choosing how to version our packages.
+
+In its current state, the document is primarily focused on one question: for each individual package, how should the version listed in the .cabal file on the `master` branch relate to the version in the .cabal file in the latest release?
+In particular, I think this question is almost entirely independent of usings `git` branches versus `git` tags, SemVer versus PVP, etc.
+This document should eventually specify that level of detail as well.
+
+## Desiderata
+
+- *Conventional Version Number*.
+  We'd like our release versions to have the standard shape of `A.B.C`, where in particular increments of `C` principally represent bugfixes.
```
michaelpj, line 12: I think this does bear on PVP, since PVP has two major versions so `C` would be a minor version rather than a patch version.

michaelpj (commented):
I think you're missing an option, let's call it "Proposal Retrograde": when you want to make a release, check all the changes that have been made since the last release to decide whether a minor/major version bump is needed, perform it, and then create the release branch.

That and "Simplest" are the two I've seen most often, basically: either bump your versions just _after_ releasing or just _before_ releasing. The advantage of doing it just after is that it's easy to remember, and your master version is always on a newer version; the advantage of doing it just before is that it's easier to compute the version bump that is _required_, if you're doing major/minor/patch versions specially (hard to do in advance!).

Parity also has the forward-looking problem: GHC essentially commits that the next release off master will _always_ be a second-major version, and all minor/patch releases happen off a release branch. That means that they can correctly guess the "next" version ahead of time. That may or may not be true for you.

michaelpj:
> Whenever a change is done on master (or a feature branch before), consider how it impacts the version number, i.e. bump the corresponding digit for breaking changes or patches when you do break an interface or just patch something.

The downside of this is that if you do *two* changes that require a minor bump, then you might accidentally bump the minor version twice. You could fix that up before you do a release, but then I'm not sure it's better than calculating the version bump just before you release.

ouroboros-consensus/docs/VersioningSchemeDecision.md
dnadales: I'm also ok with this proposal.

dnadales: However, I like this option better:

>  Whenever a change is done on master (or a feature branch before), consider how it impacts the version number, i.e. bump the corresponding digit for breaking changes or patches when you do break an interface or just patch something. The master version then indicates the version number which would be true if you just release today (e.g. by tagging a commit).

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: For the record, I think 2 dimensions versus 3 dimensions is a superficial difference. The differences discussed at https://pvp.haskell.org/faq/#how-does-the-pvp-relate-to-semantic-versioning-semver all seem secondary to me.

nfrisby: I've spelled addressed this somewhat-directly in the Desiderata section now.

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: I cleaned it up to just say "announce a commit" instead of talking about branches, tags, etc. Thanks!

nfrisby:
I've added Proposal124. It's certainly the most involved, but as far as I can tell, it does actually addresses all of the discussions on this PR that were about the focus question so far:

- Sebastian's https://github.com/input-output-hk/ouroboros-network/pull/4207#pullrequestreview-1202111566 and https://github.com/input-output-hk/ouroboros-network/pull/4207#issuecomment-1335051619 and https://github.com/input-output-hk/ouroboros-network/pull/4207#discussion_r1038060644
- Michael's https://github.com/input-output-hk/ouroboros-network/pull/4207#pullrequestreview-1202281210

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,271 @@
...
+To cut a release from a commit COMMIT on `main` that declares version `A`, announce a new non-`main` commit that extends COMMIT merely to declare version `A.0.0`.
+Also immediately update the `main` version to `A.0.0.2718`.
+
+To cut a release from a commit COMMIT on `main` that declares version `A.B`, announce a new non-`main` commit that extends COMMIT merely to declare version `A.B.0`.
+Also immediately update the `main` version to `A.B.0.2718`.
+
+To cut a release from a commit COMMIT on `main` that declares version `A.B.C.2718`, announce a new non-`main` commit that extends COMMIT merely to declare version `A.B.(C+1)`.
+Also immediately update the `main` version to `A.B.(C+1).2718`.
+
+The above is summarized in the following diagram, which relies on this legend depicting the semantics of the one kind of node (state) and the semantics of the two kinds of edges (transitions).
```
nfrisby, line 169: I'm quite proud of this Mermaid diagram. I think it does a really good job of explaining the proposal. The only two improvements I can imagine:

- better node positioning (I don't know how with Mermaid)
- better classifications of the "merge a PR" transitions; "bugfix", "backwards-compat" and "backwards-incompat" are familiar and suggestive but perhaps overly-precise?

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+Each `B`-level PR must update the `main` version from `A.B.C` to `A.(B+1).0` unless `A.B.0` is already greater than the previous release.
+
+To cut a release from a commit COMMIT on `main` that declares `A.B.C`, announce COMMIT.
+Also immediately update the `main` version to `A.B.(C+1)`.
+
+Cons:
+
+- `main` versions and release versions are not distinguished.
+
+- Some `main` commits will declare version `A.B.(C+1)` even if that version is never officially released or is created only later by backporting _different_ (patch) commits to a previous release branch.
```
michaelpj, line 68: oh, that's quite nasty, I hadn't thought of that!

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+- It incurs spurious increments of `A.B` when the only differences between two releases were patch PRs.
+
+## Proposal Dimension124
+
+Proposal Dimension above has the downside that it incurs spurious increments of `A.B` when the only differences between two releases were patch PRs.
+The following enrichment adds the minimal amount of additional complexity to avoid that without losing any invariants.
+
+Each `main` version is `A`, `A.B` where `B>0`, or `A.B.C.2718`; it is never three-dimensional.
+Each release version is the usual `A.B.C`.
+`2718` is just a recognizable magic number; it's so large it's unlikely to come up in actual versioning and it's the first digits of _e_, which is related to _growth_, which is what the `main` branch is for.
```
michaelpj, line 142: Why not just use 0?

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+
+This document records our discussions around choosing how to version our packages.
+
+In its current state, the document is primarily focused on one question: for each individual package, how should the version listed in the `.cabal` file on the main integration branch relate to the version in the `.cabal` file in the latest release?
+In particular, I think this question is almost entirely independent of usings `git` branches versus `git` tags for releases, SemVer versus PVP, etc.
+This document should eventually specify that level of detail as well.
+
+Given its current focus, without loss of generality, suppose the repository contains only one package while considering the proposals below.
+
+## Desiderata
```
zliu41, line 11: Should scalability be one of the desiderata? Does a certain proposal scale to 20 commits per day, and multiple releases per week?

If a proposal requires each PR to set the `main` version based on either the current `main` version of HEAD, or the latest release version, it probably doesn't.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+
+```
+Prelude Data.Version> makeVersion [1,2] < makeVersion [1,2,0]
+True
+```
+
+Each `main` version has only two dimensions: `A.B`.
+Each release version has at least three `A.B.C`, where `C` can be `0`.
+
+PRs do not alter the `main` version.
```
zliu41, line 116: Thank god

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+
+To cut a release from a commit COMMIT on `main` that declares version `A.B`, announce a new non-`main` commit that extends COMMIT merely to declare version `A.B.0`.
+Also immediately update the `main` version to `next(A.B)`.
```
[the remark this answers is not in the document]
zliu41: This reads like you are releasing version `A.B.0`, _then_ updating the `main` version to `next(A.B)`. If so, how do you decide whether to increment `A` or `B`? There would be no change yet since the release.

I think you probably mean deciding it _before_ the release, and either release `A+1.B.0` or `A.B+1.0`. But then I don't know what's the purpose of the commit that merely declares version `A.B.0`.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+
+- It's a very mechanical state machine, easy to execute and also easy to immediately recognize which state it's in.
+
+Cons:
+
+- The multi-sorted state transition system prevents any explanation from being comparatively small.
+
+- This scheme is certainly not already well-established!
+
+- It incurs spurious increments of `A.B` when the only differences between two releases were patch PRs.
```
zliu41, line 133: I don't understand why this has to be the case. Can you not, in this proposal, treat `C` the same as `A` and `B`, and replace `A.B` with `A.B.C` across the board?

zliu41 (commented):
It seems you are proposing to always release from `main` and the version number always increases, i.e., you'd never release 3.6.0, and then later release 3.5.2 by patching 3.5.1. That is totally fine especially for libraries, but it's worth explicitly calling out. Given this, some occurrences of `next(A.B)` can be changed to `next(A.B.C)`, since there's no need to differentiate major/minor releases vs. patch releases.

dnadales:
> Why create release branches? Why not simply _tag_ the release from master? Do you foresee a situation where you would need to fix specific versions and not others? 

Maybe because of what Marcin describes [here](https://github.com/input-output-hk/ouroboros-network/pull/4207#discussion_r1037518173)?

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+This is similar to the GHC Team's scheme.
+
+PRs do not alter the `main` version.
+
+To cut a release from a commit COMMIT on `main` that declares version `A.B` (where `B` is necessarily odd), announce a new non-`main` commit that extends COMMIT merely to declare version `X.Y = next(A.B)` (where `Y` is necessarily even).
+Also immediately update the `main` version to `X.(Y+1)`.
+
+Cons:
+
+- Commits on `main` that only include patches since the previous release would still spuriously include the `Y+1` increment in the `B` dimension.
```
dnadales, line 83: I wonder in which cases this will be an actual problem. Also, if we bumped `Y` on `main`, making the current version `X.(Y+1).Z`, I would expect that commits that include only patches would bump `Z` (which I understand it is not what is proposed here).

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Proposal NonZero
+
+Each `main` version is degenerate; eg it is always version `0`.
+Each release version is the usual `A.B.C`.
+
+PRs do not alter the `main` version.
```
dnadales, line 90: I wonder why we insist that PR's don't alter the `main` version. Doesn't this complicate the decision process of which version components to bump when making a release?

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+
+Pros:
+
+- It enforces all desired invariants.
+
+- It's a very mechanical state machine, easy to execute and also easy to immediately recognize which state it's in.
+
+Cons:
+
+- The multi-sorted state transition system prevents any explanation from being comparatively small.
```
dnadales, line 129: Could you expand this explanation? I don't get this.

ouroboros-consensus/docs/VersioningSchemeDecision.md
dnadales, line 142: DAMMERUNG SCP, right?

dnadales: Oh, nevermind, I just read your explanation later on :)

ouroboros-consensus/docs/VersioningSchemeDecision.md
dnadales: The diagram definitely helped me to understand the approach being proposed.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+## Proposal Dimension124
+
+Proposal Dimension above has the downside that it incurs spurious increments of `A.B` when the only differences between two releases were patch PRs.
+The following enrichment adds the minimal amount of additional complexity to avoid that without losing any invariants.
+
+Each `main` version is `A`, `A.B` where `B>0`, or `A.B.C.2718`; it is never three-dimensional.
+Each release version is the usual `A.B.C`.
+`2718` is just a recognizable magic number; it's so large it's unlikely to come up in actual versioning and it's the first digits of _e_, which is related to _growth_, which is what the `main` branch is for.
+
+PRs alter the `main` version as indicated in the following diagram.
```
dnadales, line 144: I can't reconcile the code block below with the diagram. I don't think I get the following diagram.

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: Yeah, this has lead to the same questions from many people.

My thought process: people will see `A.B.C.0` in a version and assume that that component can vary, since it looks so familiar. EG They'd expect `A.B.C.1`, `A.B.C.2`, etc.

I figured with `.2718` it would at least cause people to pause and ask "Wait, what? That's odd. What am I missing". (And it has, even already lol.)

nfrisby:
Thanks for the reviews! I haven't yet read the most recent ones, but I did discuss with some Consensus members this morning. That lead to some suggested improvements 🙌 

So just now I pushed proposals Redimensional and Redimensional124. The first includes a fully standalone motivation --- so you don't have to review the whole document in order to judge these new two.

cc @amesgen @dnadales @yogeshsajanikar @michaelpj @zliu41

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,397 @@
...
+One could also dismiss this concern about tooling, though, because only devs (here or downstream) should ever be building dev versions, and they should only do so explicitly (eg building from source by building a git worktree or by listing a specific git commit as the dependency).
+As long as no one installs a dev version package into a repository, there's no risk of a tool that picks amongst versions from some repository even considering a dev version.
+Dev versions are never _released_, and so should never be in any public package repository such as Hackage or [CHaP](https://github.com/input-output-hk/cardano-haskell-packages).
+The most likely way we can imagine a dev version might end up in a repository would be if a developer/user who doesn't know any better builds from source and installs a dev version into their machine's local package database.
+That single risk seems manageable.
+
+This suggests the following proposal.
+- Each release version is `A.B.C`.
+- Each dev version is `A.B.C.D`.
+- Any PR that alters a package in a way that should affect the next release version should update its version as follows.
```
nfrisby, line 352: TODO. This line still incurs the concern that @michaelpj and @zliu41 raised [here](https://github.com/input-output-hk/ouroboros-network/pull/4207/#discussion_r1040345279) and https://github.com/input-output-hk/ouroboros-network/pull/4207/#discussion_r1039426571.

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: That's a great point, yes!

I just added commit 217f273b2c9d06c5c4dc59690ad3b5d9c07d952f, which favors that desideratum, I think.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,286 @@
...
+
+    - You are of course free to decide any PR should increment `A` or `B` even if the proposed rules do not require it.
+      Ultimately, the only risk of such spurious increases is confusing/inconveniencing downstream users.
+
+    - Note, in the context of [the PVP](https://pvp.haskell.org/) eg, our `A` dimension here would itself denote a pair.
+
+- *Simplicity and Familiarity*.
+  We'd like the versioning scheme to be simple to explain and ideally already well-established.
+
+- *Ease of Execution*.
```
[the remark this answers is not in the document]
nfrisby: That's a great point, yes!

I just added commit 217f273b2c9d06c5c4dc59690ad3b5d9c07d952f, which favors that desideratum, I think.

ouroboros-consensus/docs/VersioningSchemeDecision.md
michaelpj: If they assume it can vary, and it won't, I think nothing goes wrong, and they can just continue on. Whereas if `.2718` makes them go "wait what?" then you're going to get _questions_.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,422 @@
...
+- Each release version is `A.B.C`.
+- Each dev version is `A.B.C.D`.
+- Any PR that alters a package in a way that should affect the next release version should update its version as follows.
+    - If the version was `A.B.C` before the PR, bump it to `A.B.C.0`.
+    - If the version was `A.B.C.D` before the PR, bump it to `A.B.C.(D+1)`.
+        - This rule isn't fundamental, but it ensures that every PR touches the version, which in turn ensures `git` will raise merge conflicts for backports/rebases/cherry-picking/etc.
+          Those conflicts force the developer to make adjustments to versions.
+          Those conflicts are necessary and also trivial to resolve but would be easy to overlook without this rule leveraging `git` to remind us.
+          Additionally, this means that no two code-changing commits in the subset of a single branch will declare the same version---or set of versions if the repo has multiple packages---which seems intuitive.
+        - Similar conflicts will also require updating open version-altering PRs after each version-altering PR is merged.
```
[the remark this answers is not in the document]
michaelpj: Requiring FF merges is bad enough IMO, having it so you can't even update your PR in one step but you _also_ have to fix a pointless version number conflict sounds _awful_.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,422 @@
...
+      `(A+1).0.0` if there were breaking changes, else `A.(B+1).0` if there were some new features, else `A.B.(C+1)`.
+      (Note that `.D` is present, so at least one version-influencing change was made.
+      If the version in the commit was still `A.B.C`, why are you announcing a new release?)
+    - Announce that commit as a release of the packages that have changed since the previous release.
+      (In a monorepo, there's just the one.
+      But in a polyrepo, there may be packages that haven't changed since the previous release.)
+
+## Proposal EasierRedimensional
+
+There is a variant of the above that means only release PRs alter the version, which is more scalable.
```
michaelpj, line 374: I would say this this is instead a variant of RisingEdge that avoids one of the cons (that main and release versions look the same). It's RisingEdge, except that just after you do the release you bump to a "dev version" by adding zeros until you get to the fourth zero.

michaelpj: RisingEdge continues to be my preference for simplicity, this seems like an acceptably small tweak.

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: I think because:

- It's confusing; people tend to expect versions to evolve in a minimal way.
- Spurious increases in bounds cause unnecessary busy work for downstream .cabal files that use (relatively) strong upper bounds on our libraries.

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: "Compared to most proposals, this proposal has more moving parts and so it difficult to describe briefly, which rigorous jargon, diagrams etc."

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: Just to clarify:

> except that just after you do the release you bump to a "dev version" by adding zeros until you get to the fourth zero.

Every release version has three components (or 4, for PVP). So you only ever add one `.0`.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,465 @@
...
+    - Merge a fresh PR that doesn't alter the package but does bump dev version to the next release version, as follows.
+        - Bump `A` to `A.0.0`.
+        - Bump `A.B` to `A.B.0`.
+        - Bump `A.B.C.0` to `A.B.(C+1)`.
+    - Announce that commit as a release of the packages that have changed since the previous release.
+      (In a monorepo, there's just the one.
+      But in a polyrepo, there may be packages that haven't changed since the previous release.)
+    - Announce that commit.
+
+## Proposal ScrivvyRedimensional
```
nfrisby, line 424: Thanks to @michaelpj for pointing me at and @zliu41 for writing https://github.com/input-output-hk/plutus/blob/master/doc/read-the-docs-site/RELEASE.adoc

Reading through it and the scripts inspired this proposal, which I think avoids the spurious git conflicts of Proposal Redimensional.

cc @dnadales @amesgen

ouroboros-consensus/docs/VersioningSchemeDecision.md
michaelpj: Ah right. It would also work if you didn't do that, and you could release e.g. just `A.B`. Then I guess you would say that a "dev version" is one with a trailing 0. But your way is also fine.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,465 @@
...
+
+Maintain a `wip-status` directory for each package.
+
+When merging a PR, add one of these empty files as appropriate to each package the PR alters.
```
michaelpj, line 428: Tweak: why not just suggest that people add a string to the end of the changelog entry indicating what kind of bump it is?

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,465 @@
...
+
+- Bump each package version according to the contents of its `wip-status` (leave it unchaged if the directory is empty).
+- Remove all files from the three `wip-status/{major,minor,patch}` directories.
+
+_Remark_.
+If you maintain a separate changelog per package, then `scriv`'s [Categories](https://github.com/nedbat/scriv/blob/cd79a2a618eb752075bbf45e93ad6b7576a2924a/docs/concepts.rst#categories) can be leveraged for this, instead of these empty files.
+
+Pros:
+
+- A PR's contribution to the subsequent release's version bump is judged when reviewing the PR instead of being assessed (many days) later.
```
michaelpj, line 449: I do like this, might steal it for `plutus`.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,465 @@
...
+- Cut a new release as follows.
+    - Merge a fresh PR that doesn't alter the package but does bump dev version `A.B.C.D` to the next release version, depending on the contents of the relevant PRs since the previous release:
+      `(A+1).0.0` if there were breaking changes, else `A.(B+1).0` if there were some new features, else `A.B.(C+1)`.
+      (Note that `.D` is present, so at least one version-influencing change was made.
+      If the version in the commit was still `A.B.C`, why are you announcing a new release?)
+    - Announce that commit as a release of the packages that have changed since the previous release.
+      (In a monorepo, there's just the one.
+      But in a polyrepo, there may be packages that haven't changed since the previous release.)
+
+## Proposal EasierRedimensional
```
michaelpj, line 372: I continue to think this is a variant of RisingEdge but whatever

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,465 @@
...
+    - Note, in the context of [the PVP](https://pvp.haskell.org/) eg, our `A` dimension here would itself denote a pair.
+
+- *Simplicity and Familiarity*.
+  We'd like the versioning scheme to be simple to explain and ideally already well-established.
+
+- *Ease of Execution*.
+  We'd like the process of cutting a release to merely involve following a very simple checklist.
+  We'd like it to require as few inputs, discussions, decisions, etc as possible.
+
+- *Distinguished Development Versions*.
```
michaelpj, line 31: I had a somewhat evil idea. 

```
flag allow-dev-version
   description: Allow the use of development versions of this package.
   manual: True
   default: false
```

Then when you are on a development version you have:
```
if !flag(allow-dev-version)
  buildable: False
```
... and not when you do a release. You can set it locally in your `cabal.project` so you can always use the versions, but downstream won't be able to use them without explicitly setting `allow-dev-version`.

So the release process is something like:
- Make a commit C1 which bumps the version numbers 
- Make a commit C2 which removes all the `allow-dev-version` guards
- Tag that commit as the release
- Revert C2

michaelpj: (I also want to register that I'm less convinced than you that this is a big problem)

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,465 @@
...
+When merging a PR, add one of these empty files as appropriate to each package the PR alters.
+
+- `<package>/wip-status/major/<scriv-id>`
+- `<package>/wip-status/minor/<scriv-id>`
+- `<package>/wip-status/patch/<scriv-id>`
+
+(`<scriv-id>` is the timestamp, commit handle, and branch name, [like `scriv`](https://github.com/nedbat/scriv/blob/cd79a2a618eb752075bbf45e93ad6b7576a2924a/docs/index.rst#getting-started).)
+
+In particular, if the PR should add `wip-status/minor/<this-PR-id>`, then it should do so even if `wip-status/major` is non-empty.
+That would ensure that backporting this PR onto a branch with an empty `major` directory would still appropriately alter the `wip-status` directory.
```
dnadales, line 437: I don't get why this is important.

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: I think this sentence below is relevant:

> If you maintain a separate changelog per package, ...

I guess if your changelog entries were formulaic enough that you could "automatically" determine the way in which the PR affects the version of each package the PR affects---that'd be sufficient too.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,577 @@
...
+This short document therefore assembles what prior art we have found into such a resource.
+
+Executive summary: the challenge of maintaining versions is qualitatively equivalent to the challenge of maintaining a changelog.
+
+### Design Space Overview
+
+It's a surprisingly rich design space, but the following concepts delineate its perimeter and the main trends within it.
+
+A principal dichotomy among possible version declaration maintenance schemes has become apparent during our research.
+- Definition _RisingEdge_. The version declaration is altered by (some of the) development PRs; it's already correct when it comes time to announce a release.
```
michaelpj, line 480: These definitions are reversed compared to the previous draft! Which perhaps suggests they're not that obvious as names :)

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,577 @@
...
+- Add a commit that sets those versions accordingly.
+
+If you're envisioning a single bumplog file, then you might be wondering why this scheme wouldn't also incur spurious merge conflicts among PRs, etc.
+Fortunately, this is already a solved problem for changelogs; see tools like https://scriv.readthedocs.io/en/latest et al.
+
+_Remark_.
+We chose 90% and 95% based on the use of a degenerate changelog.
+Mature/stable projects likely already have a non-degenerate changelog that would make the pre-release assessment required by FallingEdge very simple, if not totally trivial.
+Thus, such projects' teams might assign slighty better and/or worse scores to EasyPR and/or
+EasyRelease to this scheme, depending on how they choose to resolve the tension between the bumplog and the structure/thoroughness of their actual changelog.
```
michaelpj, line 544: Perhaps you might consider it a point in favour of having the changelog and bumplog merged that it nudges people to make changelog entries! It's easy to think "oh, this doesn't need a changelog entry, it's so minor", but if you have to edit it _anyway_ to say "patch bump"... you might as well write a sentence about what changed.

michaelpj (approved):
I like the combination of "set version just before release" and "use the changelog to keep track of what kind of version bump you will need" :+1:

ouroboros-consensus/docs/VersioningSchemeDecision.md
nfrisby: Yes, I realized that same thing last night!

And agree that it means they're not good names.

zliu41 (commented):
FallingEdge with changelog/bumplog sounds good.

Can we enforce editing changelog/bumplog? E.g., add a CI check that fails if the PR does not produce a new changelog/bumplog entry, unless the commit message contains `DISABLE_CHANGELOG_CHECK`, or something like that. Because if changelog/bumplog is used to determine the next version, then its accuracy is important.

ouroboros-consensus/docs/VersioningSchemeDecision.md
Jasagredo: I would not be opposed to this. I think it is a nice use of the `buildable` flag and can be made consistent in the whole project. Configure it once, be safe forever :smile:

Jasagredo:
> FallingEdge with changelog/bumplog sounds good.

I adhere to @zliu41's comment. The CI check is maybe too much, adding a checkbox on the PR template maybe is enough?

Jasagredo (commented):
>  FallingEdge with changelog/bumplog sounds good.

I adhere to @zliu41's comment. The CI check is maybe too much, adding a checkbox on the PR template maybe is enough?

ouroboros-consensus/docs/VersioningSchemeDecision.md
coot: Observation: so a new minor release `A.B.(C+1)` (with `B` even) will only be possible from a release branch hooked on `A.B.0`.  This is compatible with the branching model we use right now.

ouroboros-consensus/docs/VersioningSchemeDecision.md
```diff
@@ -0,0 +1,577 @@
...
+Even if we assume an ideal commit messages and/or PR descriptions, there will likely be a burdensome number of them.
+The slightly more realistic assumptions of commit messages and/or PR descriptions that do have ideal content but don't have perfectly uniform structure means the assessor would still have to thoroughly read each, judging what the content requires of the version bump under consideration.
+
+However, there is one standard log that can often make this assessment much easier: the changelog.
+It's not automatic, but it is often structured around "additions/changes/etc", which are exactly what determines the next version under a policy like SemVer, for example.
+
+Thus our final scheme to consider is a variant of FallingEdge that also requires/assumes the maintenance of a _bumplog_, which is merely a degenerate changelog that contains nothing more than what's necessary to determine the next version number for each changed packaged (ie one classification of `patch`, `minor`, and `major` for each merged PR).
+
```
coot, line 524: Is `FallingEdge` like `FalingEdgePatch` but PR author doesn't bump the major version (thus regular PRs never bump versions at all)?

ouroboros-consensus/docs/VersioningSchemeDecision.md
coot: I've been thinking about exactly the same, e.g. the next unpublished section of changelog could be split into two lists of major and minor changes.

nfrisby:
I've marked this PR ready for review. The main content is the new `ReleaseProcess.md` file; the other one is just for historical reference if we end up later questioning the content of `ReleaseProcess.md`.

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+
+The Consensus team will sometimes be maintaining multiple MAJOR.MINOR versions of the same package simultaneously.
+EG We might release FOO-2.4.1 on Monday, release FOO-2.3.7 on Wednesday, and release FOO-2.4.2 on Friday, etc.
+
+- Whenever we're maintaining only the one greatest-ever MAJOR.MINOR version of a package, then all work on that package will happen on the main branch.
+- Whenever we're also maintaining some lesser MAJOR.MINOR versions of a package, then some work for that package will also be happening on the respective _release branches_.
+
+A _release branch_ is a branch dedicated to the maintenance of some older release of a package.
+
+RULE: If we have released a package FOO-X.Y.Z, then the release-FOO-X.Y.x branch MUST exist and its first-parent history MUST include the commit released as FOO-X.Y.Z.
```
jorisdral, line 32: It was not entirely clear to me from reading up to this point why the branch is named release-FOO-X.Y.**x**, though it is explained later. May be good to mention here what **x** represents (I think just a literal `x` character?) or to refer to the later section where it is explained

jorisdral (approved):

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+This is easy to generalize to other conventional notations, such as [the PVP](https://pvp.haskell.org/)'s MAJOR.MAJOR.MINOR.PATCH quadruples.
+
+We refer throughout this document to _the main branch_ (aka `master`, aka `main`).
+
+If you search for "RULE:" in this document, you'll find the key statements.
+
+## Rules for Branches
+
+The Consensus team will sometimes be maintaining multiple MAJOR.MINOR versions of the same package simultaneously.
+EG We might release FOO-2.4.1 on Monday, release FOO-2.3.7 on Wednesday, and release FOO-2.4.2 on Friday, etc.
```
dnadales, line 25: I don't get how the example relates to releasing versions simultaneously, and why we choose that ordering. In particular, the fact that we released `FOO-2.4.1` indicates that we were maintaining a previous version `< 2.4.1`?

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+
+- Whenever we're maintaining only the one greatest-ever MAJOR.MINOR version of a package, then all work on that package will happen on the main branch.
```
dnadales, line 27: Something like this? 

```text
            FOO-2.3.6 --- FOO-2.3.7 <- release branch 
             /                                            
-------FOO-2.3.? ------  .... ------ FOO-2.4.1  
                                        ^
                                        |
                                   main branch
```

Assuming `FOO-2.4.1` is the greatest-ever version, if we have to maintain this version, then we develop on `main`. If we have to maintain `2.3.7`, then we work on the release branch.

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+- We release FOO-2.3.7 on Wednesday.
+
+*Remark*.
+Not every first-parent commit in the history of the release branch is a release commit.
+We will be merging multiple PRs into the release branch in order to prepare the next release from it, so some commits will just be the intermediates between two releases.
+
+*Remark*.
+The release branch for the greatest MAJOR.MINOR version of a package is somewhat degenerate.
+It's either equal to or a prefix of the main branch.
+The only time it MUST be updated to the tip of the main branch is when we cut a release of the package from the main branch, to satisfy the rule above about all X.Y.Z release commits being on the release-X.Y.x branch.
```
dnadales, line 52: It'd be nice to have an example of this.

dnadales: Similar to the example above:

```text
            FOO-2.3.6 --- FOO-2.3.7 <- release branch 
             /                                            
-------FOO-2.3.? ------  .... ------ FOO-2.4.1 --- (what happens with the commits here?)  
                                        ^
                                        |
                                   main branch = release branch? 
```                                   

do we need to create a branch when we release `2.4.1`? If so, what happens to the commits on `main`?

ouroboros-consensus/docs/ReleaseProcess.md
dnadales, line 32: >  MUST include the commit released as FOO-X.Y.Z

What is "the commit released as FOO-X.Y.Z"?

dnadales: Do you mean the release commit?

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+EG Consider the following possible timeline.
+
+- We release FOO-2.4.1 on Monday, and FOO-2.4.1 completely reworked some functionality.
+- We then realize on Tuesday that there was a severe bug in the old functionality.
+- We fix the bug by merging a PR that targets the FOO-2.3.x release branch (since the bug no longer exists on the main branch!)
+- We release FOO-2.3.7 on Wednesday.
+
+*Remark*.
+Not every first-parent commit in the history of the release branch is a release commit.
+We will be merging multiple PRs into the release branch in order to prepare the next release from it, so some commits will just be the intermediates between two releases.
```
dnadales, line 47: > We will be merging multiple PRs into the release branch in order to prepare the next release from it

by means of cherry-picking commits from `main`?

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+
+- RULE: A fresh PR MUST target the main branch, if possible.
+- RULE: A PR MUST NOT be merged into a branch unless it should be included in the next release from that branch.
+- We maintain a changelog for each package (one for FOO and one for BAR).
+- RULE: A PR MUST add a pending changelog entry for each package that it alters.
+- RULE: Each pending changelog entry MUST at the very least classify the alteration to that package as a _patch_ level change (eg a bugfix), a _minor_ change (eg a non-breaking interface change), or a _major_ change (eg a breaking interface change).
+
+*Remark*.
+Notice that we do not allow merging a PR that should not be included the subsequent release.
+This means some PR may be "done" but still should not be merged.
```
dnadales, line 72: So according to this rule, if we are not sure we will be releasing a new feature, eg UTxO-HD, the feature has to live in a branch, right?

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+    - (Note that there must be some alterations to each package we are including in the release, since otherwise we wouldn't be including it in the release---but mind the _bundles_ mentioned below.)
+    - Let X.Y.Z be the version of the package currently declared on this branch.
+    - RULE: If any alteration was major-level, then we bump to (X+1).0.0.
+    - RULE: Else if any alteration was minor-level, then we bump to X.(Y+1).0.
+    - RULE: Otherwise all alterations were patch-level, so we bump to X.Y.(Z+1).
+- RULE: We merge one final PR, which MUST do exactly the following and nothing more.
+    - RULE: It updates the versions of the packages being released as described above.
+    - RULE: It flushes the pending changelog entries of each package being released into the `CHANGELOG.md` file for that package.
+- RULE: We tag that resulting merge commit as release-PKG-A.B.C; one tag per package PKG that is being released (ie FOO and/or BAR).
+- RULE: If C is 0, then we also create the release branch release-PKG-A.B.x from this new commit.
```
dnadales, line 105: What is branch `release-PKG-A.B.x` for? And does `x` represent a particular value, or it's `x` verbatim?

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+- RULE: Finally, we announce this commit hash as the new release of these packages.
+  EG We insert these package's new versions into Hackage, [CHaP](https://github.com/input-output-hk/cardano-haskell-packages), etc.
+
+*Remark*.
+To explicitly clarify: after a release of a package, there will be zero pending changelog entries for that package; they were destroyed when incorprated into the `CHANGELOG.md` file making the release.
+But there may be pending changelog entries of other packages, those that we're included in this release.
```
dnadales, line 111: Those that were not included?

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,140 @@
...
+- RULE: We merge one final PR, which MUST do exactly the following and nothing more.
+    - RULE: It updates the versions of the packages being released as described above.
+    - RULE: It flushes the pending changelog entries of each package being released into the `CHANGELOG.md` file for that package.
+- RULE: We tag that resulting merge commit as release-PKG-A.B.C; one tag per package PKG that is being released (ie FOO and/or BAR).
+- RULE: If C is 0, then we also create the release branch release-PKG-A.B.x from this new commit.
+- RULE: Finally, we announce this commit hash as the new release of these packages.
+  EG We insert these package's new versions into Hackage, [CHaP](https://github.com/input-output-hk/cardano-haskell-packages), etc.
+
+*Remark*.
+To explicitly clarify: after a release of a package, there will be zero pending changelog entries for that package; they were destroyed when incorprated into the `CHANGELOG.md` file making the release.
```
dnadales, line 110:
```suggestion
To explicitly clarify: after a release of a package, there will be zero pending changelog entries for that package; they were merged into the `CHANGELOG.md` file when making the release.
```

dnadales (commented):
Thank you Nick! I left some comments and questions.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: Yes, yeesh. A bad typo. Thanks.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: I expanded this sentence to be much more explicit.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: I've added a sub-bullet here with a concrete example. Does that clarify?

@jorisdral commented similarly about `.x`. This `.x` suffix is used in other IOG repos, eg:

image

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: Yes. Please confirm: that's what we decided on in our call, right?

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: The plan is for the commits to be cherry-picked, yeah. See _fresh_ versus _backport_ in the `Notation` section above.

> There are two kinds of work on release branches.
> Most of the time, that work is immitating some work that was done on the main branch.
EG we fixed a bug on the main branch and we're backporting that fix onto a release branch (hopefully it's a trivial cherry-pick).
> Rarely, though, there will be fresh work done on a release branch.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: Yes. I think "the commit released as FOO-X.Y.Z" is easer to read than "the FOO-X.Y.Z release commit".

Do you disagree?

nfrisby: Thanks; yeah, you and Damian both pointed that out. I'll add some concrete examples making that clear.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: I expanded that last sentence to the following. Does that clarify?

> EG If we were currently maintaining a 2.3 version and a 2.4 version simultaneously, then we might release FOO-2.4.1 on Monday, release FOO-2.3.7 on Wednesday, and release FOO-2.4.2 on Friday, etc.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: > If we have to maintain 2.3.7, then we work on the release branch.

If necessary, yes (see my reply to your other comment below). But hopefully that "work on the release branch" will just be cherry-picking some commits from the main branch.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: For now, for consistency, I've removed all occurrences of the noun phrase "release commit" (there were two).

We can reverse that if I'm outvoted :P

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: Hmm. If, _unlike UTxO HD_, we do have an easy way to "include the feature but definitely hidden behind a flag", then we could go ahead and merge that. But we'd only merge it if we are _intending_ for it to be in the next release: the "experimental flag" will be part of the next release. (That's the point of the flag.)

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: I pushed up a longer example. Does it suffice for this?

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: I pushed up a longer example. Does it suffice for this?

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: I pushed up a longer example. Does it suffice for this?

ouroboros-consensus/docs/ReleaseProcess.md
dnadales: >> What is branch release-PKG-A.B.x for?
> I've added a sub-bullet here with a concrete example. Does that clarify?

Is this the clarification? 

> EG We insert these package's new versions into Hackage

So the answer, in the example, would be: we use this branch so that CHaP can pick the new version up?

ouroboros-consensus/docs/ReleaseProcess.md
dnadales: > Yes. Please confirm: that's what we decided on in our call, right?

I think so, but I'll ask the rest.

ouroboros-consensus/docs/ReleaseProcess.md
dnadales: Alright, clear. Thanks!

ouroboros-consensus/docs/ReleaseProcess.md
dnadales: > I expanded that last sentence to the following. Does that clarify? 

It does! Thanks!

ouroboros-consensus/docs/ReleaseProcess.md
dnadales: I guess you refer to [this](https://github.com/input-output-hk/ouroboros-network/pull/4207/commits/02482cf7fbabf1b6e4441dd6d730f36c121bb45e#diff-a3b11ca853f39e9e181a2ad2868b36c39f946100f19907a46e41607456b48cd9R148) right? That example is really helpful, I just want to confirm my assumption: at the end of the example we'll end up with this commit structure, right?

```
F   J
|   |
E   I 
|   | 
D   H  
|   | 
C   G
| /
B
|
A
```

dnadales (approved):

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,229 @@
...
+  Except in the rare circumstances discussed in the previous section, every fresh PR will target the main branch.
+- The primary objective of a backport PR is merely to immitate some fresh PR.
+
+The rules are then as follows.
+
+- RULE: A fresh PR MUST target the main branch, if possible.
+- RULE: A PR MUST NOT be merged into a branch unless it should be included in the next release from that branch.
+- We maintain a changelog for each package (one for FOO and one for BAR).
+- RULE: A PR MUST add a pending changelog entry for each package that it alters.
+- RULE: Each pending changelog entry MUST at the very least classify the alteration to that package as a _patch_ level change (eg a bugfix), a _minor_ change (eg a non-breaking interface change), or a _major_ change (eg a breaking interface change).
```
amesgen, line 71: Do we want to add a RULE that merge bot such as [Mergify](https://docs.mergify.com/actions/backport/#backport-action) must be used if possible (i.e. no conflicts)?

ouroboros-consensus/docs/ReleaseProcess.md
```diff
@@ -0,0 +1,229 @@
...
+
+The Consensus team will sometimes be maintaining multiple MAJOR.MINOR versions of the same package simultaneously.
+EG If we were currently maintaining a 2.3 version and a 2.4 version simultaneously, then we might release FOO-2.4.1 on Monday, release FOO-2.3.7 on Wednesday, and release FOO-2.4.2 on Friday, etc.
+
+- Whenever we're maintaining only the one greatest-ever MAJOR.MINOR version of a package, then all work on that package will happen on the main branch.
+- Whenever we're also maintaining some lesser MAJOR.MINOR versions of a package, then some work for that package will also be happening on the respective _release branches_.
+
+A _release branch_ is a branch dedicated to the maintenance of some older release of a package.
+
+RULE: If we have released a package FOO-X.Y.Z, then the release-FOO-X.Y.x branch MUST exist and its first-parent history MUST include the commit released as FOO-X.Y.Z.
```
amesgen, line 32: I am not opposed, but what would be a use case for backporting some change to a *proper* subset of minor versions of a given major release? IIUC, our users should always be able to upgrade from one minor versions to a later one under the same major version. If this extra flexibility is not required, we could just have one branch per major version.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: The clarification was about only about the `.x` question:

> For example, when releasing version 2.3.0 of a package named FOO, we'd create the branch named release-FOO-2.3.x.

The answer to "What is branch release-PKG-A.B.x for?" is explained earlier in the document ("## Rules for Branches") and it's demonstrated in action in the example at the bottom ("### Release Branch Example")

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: Hmm. I hadn't thought that through---I merely mimicked what I've seen in the `cardano-node` and `cardano-ledger` repositories.

Please confirm I'm understanding your point correctly. It should be as simple for the user to upgrade from A.B.X to A.(B+1).Y as it would be for them to upgrade to A.B.(C+1). And therefore we shouldn't bother maintain release branches for each MAJOR.MINOR pair, only one per MAJOR. Is that right?

In the abstract, that does seem right to me, because of the backwards-compatibility semantics of minor version bumps. But I wonder if there's some surprise lurking there that I'm failing to anticipate. ... maybe the downstream user has other constraints (eg regulatory/large company bullshit SoPs/etc) preventing them from taking the minor bump upgrade?

We can cross that bridge when/if we come to it, though. I'll adjust the text now.

amesgen: > Please confirm I'm understanding your point correctly. It should be as simple for the user to upgrade from A.B.X to A.(B+1).Y as it would be for them to upgrade to A.B.(X+1). And therefore we shouldn't bother maintaining release branches for each MAJOR.MINOR pair, only one per MAJOR. Is that right?

Exactly, thanks for spelling that out explicitly.

ouroboros-consensus/docs/ReleaseProcess.md
nfrisby: This document is only about which branches&versions should exist and how they should relate. How you actually merge the PRs seems beyond the scope of that. What do you think?

ouroboros-consensus/docs/ReleaseProcess.md
dnadales: Yes.

ouroboros-consensus/docs/ReleaseProcess.md
dnadales: I see. Thanks.

nfrisby:
bors merge

PR #4213, opened 2022-12-05
Make the `forkTxProducer` in ThreadNet tests get a `BackingStoreValueHandle` on each slot

Jasagredo:
# Description

Retrieve a BackingStoreValueHandle on each act of the watcher in `forkTxProducer` in the ThreadNet tests

### DIFF
--- a/ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
+++ b/ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
@@ -658,13 +658,13 @@ runThreadNetwork systemTime ThreadNetworkArgs
                    -> OracularClock m
                    -> TopLevelConfig blk
                    -> Seed
-                   -> DiskLedgerView m (ExtLedgerState blk)
+                   -> m (DiskLedgerView m (ExtLedgerState blk))
                       -- ^ How to get the current ledger state
                    -> Mempool m blk
                    -> m ()
-    forkTxProducer coreNodeId registry clock cfg nodeSeed dlv mempool =
+    forkTxProducer coreNodeId registry clock cfg nodeSeed mdlv mempool =
         void $ OracularClock.forkEachSlot registry clock "txProducer" $ \curSlotNo -> do
-          let (DiskLedgerView emptySt _ doRangeQuery _) = dlv
+          DiskLedgerView emptySt _ doRangeQuery _ <- mdlv
           fullLedgerSt <- fmap ledgerState $ do
                 -- FIXME: we know that the range query implemetation will add at
                 -- most 1 to the number of requested keys, hence the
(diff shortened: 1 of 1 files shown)


ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
```diff
@@ -1044,29 +1044,10 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
-             )
-        extractBsValueHandle = \case
-            StaticLeft bsValueHandle -> bsValueHandle
-      seBsValueHandle <-
-          fmap extractBsValueHandle
-        $ ChainDB.getLedgerBackingStoreValueHandle chainDB registry (StaticLeft ())
+      let getValueHandle = do
+            bsvh <- ChainDB.getLedgerBackingStoreValueHandle chainDB registry (StaticLeft ())
+            pure $ mkDiskLedgerView $ case bsvh of
+              StaticLeft v -> v
```
jorisdral, line 1050: What do you think of adding these functions to `Ouroboros.Consensus.Util`:

```haskell
staticFromLeft :: StaticEither 'False l r -> l
staticFromLeft = \case (StaticLeft x) -> x

staticFromRight :: StaticEither 'True l r -> r
staticFromRight = \case (StaticRight x) -> x
```

jorisdral (approved):
Could you elaborate on why this change is necessary? (not that I'm doubting it, but I don't know the motivation)

ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
```diff
@@ -658,13 +658,13 @@ runThreadNetwork systemTime ThreadNetworkArgs
...
                    -> Seed
-                   -> DiskLedgerView m (ExtLedgerState blk)
+                   -> m (DiskLedgerView m (ExtLedgerState blk))
                       -- ^ How to get the current ledger state
                    -> Mempool m blk
                    -> m ()
     forkTxProducer coreNodeId registry clock cfg nodeSeed dlv mempool =
         void $ OracularClock.forkEachSlot registry clock "txProducer" $ \curSlotNo -> do
-          let (DiskLedgerView emptySt _ doRangeQuery _) = dlv
+          DiskLedgerView emptySt _ doRangeQuery _ <- dlv
```
dnadales, line 667: Minor: would it make sense to replace `dlv` with `mdlv`?

ouroboros-consensus-test/src/Test/ThreadNet/Network.hs
dnadales: I like the idea.

dnadales (approved):
Thanks!

PR #4239, opened 2022-12-21
Add component diagram to the Consensus' Haddock

### LINKED ISSUE #4198: Redraw the diagram from network design doc linking to code and put in the top-level haddock
The [network-design document](https://hydra.iohk.io/job/Cardano/ouroboros-network/native.network-docs.x86_64-linux/latest/download/1) contains a diagram that shows an overview of the main consensus components. This provides a great overview of the consensus layer that should be made available in the [top-level haddock documentation](https://input-output-hk.github.io/ouroboros-network/).

The goal of this issue is to update ` scripts/prolog` to incorporate a (possibly improved or more detailed) version of such diagram. The components in the diagram should link to the haddock documentation of the corresponding modules.

abailly-iohk:
# Description

This small PR adds a C4-style Component diagram to generated documentation, replicating some figure from the Consensus report with  hyperlinks to relevant Haddock's page. The goal is to have more of those kind of diagrams to better communicate the architecture of the consensus.

fix #4198

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/scripts/c4-component.puml
+++ b/scripts/c4-component.puml
@@ -1,0 +1,42 @@
+@startuml Consensus
+!include https://raw.githubusercontent.com/plantuml-stdlib/C4-PlantUML/master/C4_Component.puml
+HIDE_STEREOTYPE()
+
+ComponentDb(chainDB, "Chain DB", "FS", "Stores chain of blocks", $link="https://input-output-hk.github.io/ouroboros-network/ouroboros-consensus/Ouroboros-Consensus-Storage-ChainDB.html")
+
+
+Component(mempool, "Mempool", "", "Holds new transactions past the current tip of the chain", $link="https://input-output-hk.github.io/ouroboros-network/ouroboros-consensus/Ouroboros-Consensus-Mempool.html")
+Component(syncer, "Mempool Syncer", "", "Synchronizes the mempool with latest state of the chain", $link="https://input-output-hk.github.io/ouroboros-network/ouroboros-consensus/Ouroboros-Consensus-Mempool-Impl.html")
+Component(block_creation, "Block Creator", "", "Creates new block out of txs in the mempool", $link = "https://input-output-hk.github.io/ouroboros-network/ouroboros-consensus/Ouroboros-Consensus-Block-Forging.html")
+Component(candidates, "Chain Candidates", "", "Maintains candidate chains from peers")
+Component(block_download, "Block Download", "", "Select which blocks to downloads")
+
+Component(tx_sub_in, "Tx Submission Client", "", "Receives new transactions from connected peers", $link="https://input-output-hk.github.io/ouroboros-network/ouroboros-network-protocols/Ouroboros-Network-Protocol-TxSubmission2-Client.html")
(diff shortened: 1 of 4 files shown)


scripts/c4-component.puml
```diff
@@ -0,0 +1,42 @@
+@startuml Consensus
```
dnadales, line 1: The graph looks amazing. Can we add instructions somewhere that describe the tool we use for producing this diagram, and how to generate it? Also, we should add a checklist item to the PR template to make sure this diagram stays current.

abailly-iohk: Good point! How about adding comments to the `.puml` file?

dnadales: Sounds good!

dnadales (approved):

Jasagredo:
I think this is a super useful diagram, however I see some improvements here and there. This is a rather unsorted collection so I will list them here.

## Direction of relations/arrows

I find the directions of the arrows a bit confusing. Let me explain myself:

- For example on the one hand: `TxSubmission Client` adds transactions to the `Mempool`. This one is OK.
- on the other hand, `Block creator` reads transactions from the `Mempool`, but the arrow says that `Mempool` reads txs to/from/for the `Block creator`. I think this arrow is inverted. Either I would invert the arrow or I would label it `serves txs` instead of `reads txs`.
- Similarly between ChainDB and Mempool Syncer.
- Similarly between ChainSync Client and ChainDB
- Between ChainDB and ChainSyncServer I would say `serves headers` (note that now it says "blocks"), and then the component for chain sync server already specifies that it propagates them to clients.
- Similarly with ChainDB -> BlockFetchServer

## Other notes
- I think `Block Download` and `Block Fetch Client` are the same thing.
- I think it would be useful to replace TxSubmissionClient with `TxSubmission NTN client` which receives transactions from peers (nodes) and `TxSubmission NTC server` which receives transactions from local clients (wallets)
- I would try to place ChainSyncServer and ChainSyncClient one next to the other
- It might be useful to use colors? Like ChainSync in one color, BlockFetch in another.
- Another component might be Chain Selection which would read blocks from the chain DB and update its state based on the protocol. Technically `BlockFetch` doesn't extend the chain, it pushes new blocks to the VolatileDB and then Chain selection runs to maybe adopt the new chain which might be an extension.
- Might be an overkill but maybe we can depict also the distinction between VolatileDB, ImmutableDB and LedgerDB? In particular
    - ChainSync provides things from both the Immutable and VolatileDB.
    - BlockFetch adds blocks to the volatileDB
    - Chain Selection takes blocks from the volatileDB and updates the LedgerDB when choosing a candidate.
    - Mempool reads the tip of the LedgerDB and runs transactions on top of it.
    - Block creator adds a block to the volatileDB the same way BlockFetch does.

Those are the notes I can think of atm. Let me know what you think.

abailly-iohk:
I will upgrade the PR following your and @dnadales 's suggestions

Jasagredo (commented):
I should have actually used the Review commenting instead of adding a comment alone. If and when you need a re-review please request it again.

dnadales (approved):
Looks great. Thanks!

PR #4249, opened 2023-01-02
Happy new year :)

coot:

PR #4261, opened 2023-01-10
Add cabal check CI

newhoggy:
# Description

If cabal files are malformed, it can cause downstream projects to not be able to use this project.  Adding the CI check will ensure the cabal files are never malformed so downstream projects can update their dependencies freely.

* Fix malformed cabal files
* Bring cabal files up to standard (syntactically) for publishing
* Add cabal file check so that new and existing cabal files do not regress below standard

# Checklist

- Branch
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested


.github/workflows/build.yml
```diff
@@ -143,13 +88,16 @@ jobs:
     - name: Record dependencies
       id: record-deps
       run: |
+        # The tests call out to msys2 commands. We generally do not want to mix toolchains, so
+        # we are very deliberate about only adding msys64 to the path where absolutely necessary.
+        ${{ (runner.os == 'Windows' && '$env:PATH=("C:\msys64\mingw64\bin;{0}" -f $env:PATH)') || '' }}
```
coot, line 93: We add this to almost every section, is it necessary?  Can we alternatively have commands that do not depend on this, then modify `PATH` and then commands which require it?

.github/workflows/check.yml
```diff
@@ -0,0 +1,155 @@
...
+    - name: Cabal check
+      run: |
+        for x in $(find . -name '*.cabal' | cut -c 3-); do
+          (
+            d=$(dirname $x)
+            echo "== $d =="
+            cd $d
+            cabal check
+          )
+        done
```
amesgen, line 155: AFAIU, `cabal check` does not compile anything, so I think most of this setup in this CI workflow is not necessary.

Also, can you move this to e.g. `./scripts/ci/check-cabal.sh` such that it is easy to run locally and does not get out of sync?

amesgen (changes requested):
Thanks a lot of cleaning up the `.cabal` files!

All other formatting/linting checks reside in Nix (see e.g. [`./nix/check-stylish.nix`](https://github.com/input-output-hk/ouroboros-network/blob/db633ec71eff9b2b7797ad0cce0c130771b8cf0c/scripts/ci/check-stylish.sh)), so what about also putting this one there?

.github/workflows/build.yml
newhoggy: Fixed!

.github/workflows/check.yml
newhoggy: Done!

cardano-ping/CHANGELOG.md
```diff
@@ -1,5 +0,0 @@
-# Revision history for cardano-ping
-
-## 0.1.0.0 -- 2022-12-14
-
-* This code was originally from the cardano-ping executable component of the `network-mux` package.
```
coot, line 5: Instead of removing it we should add `extra-source-files: CHANGELOG.md`.  The same in all other removed `CHANGELOG.md` files.

cardano-client/cardano-client.cabal
```diff
@@ -1,24 +1,23 @@
...
+category:               Network
+build-type:             Simple
 
 library
   hs-source-dirs:      src
   exposed-modules:     Cardano.Client.Subscription
   default-language:    Haskell2010
-  build-depends:       base,
+  build-depends:       base >= 4.14 && < 4.17,
                        bytestring        >=0.10 && <0.12,
```
coot, line 21:
```suggestion
  build-depends:       base              >= 4.14 && < 4.17,
                       bytestring        >=0.10 && <0.12,
```

scripts/check-cabal.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
amesgen, line 1: Note that the workflow currently tries to execute `scripts/check-cabal-files.sh` instead of this.

Also, the other scripts are in the `ci` subdirectory of `./scripts` instead of directly in it.

coot (changes requested):

.github/workflows/check-cabal-files.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    env:
+      # current ref from: 27.02.2022
+      SECP256K1_REF: ac83be33d0956faf6b7f61a60ab524ef7d6a473a
+
+      SECP_CACHE_VERSION: "2022-12-30"
+
+      PKG_CONFIG_PATH: /usr/lib/pkgconfig:/usr/local/lib/pkgconfig
+
+      LD_LIBRARY_PATH: /usr/local/lib
```
amesgen, line 18: Given that the new workflow file is now so small (e.g. this `env:` section can even be still removed I think), what about adding this as a job/step in the existing build workflow? This way, we will have less duplication of e.g. GHC/Cabal versions.

Alternatively, we could just run it via Nix, do you have thoughts on that (see also the [review message](https://github.com/input-output-hk/ouroboros-network/pull/4261#pullrequestreview-1241825272))?

cardano-ping/CHANGELOG.md
newhoggy: I thought to add, but it seems a lot of these files are never updated.

cardano-client/cardano-client.cabal
newhoggy: Fixed!

scripts/check-cabal.sh
newhoggy: Fixed!

.github/workflows/check-cabal-files.yml
newhoggy: It might be advantageous to be able to control the `cabal` version.  For example `cabal check` behaves differently between `cabal-3.6.2.0` and `cabal-3.8.1.0` despite `cabal` being able to compile the project in either case.

newhoggy: The `env` is removed now.

cardano-ping/CHANGELOG.md
newhoggy: I'm going to go for `CHANGELOG.md` since that's consistent with `interface-CHANGELOG.md`

coot: I want to maintain them, I already renamed them in #4249.

.github/workflows/check-cabal-files.yml
```diff
@@ -0,0 +1,20 @@
...
+    - name: Install Haskell
+      uses: input-output-hk/setup-haskell@v1
+      id: setup-haskell
+      with:
+        cabal-version: "3.8.1.0"
+
+    - uses: actions/checkout@v3
+
+    - name: Cabal check
+      run: ./scripts/check-cabal-files.sh
```
amesgen, line 20: Can you move this to `build.yml` as a separate job? Advantages:

 - The `on:` triggers are reused (e.g. they are incorrect here, as they are not triggered on PRs).
 - We don't have to update the branch protection rules (see the "Required" tags) as well as `bors.toml` (once we start using it again).

scripts/check-cabal-files.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
amesgen, line 1: Quoting from https://github.com/input-output-hk/ouroboros-network/pull/4261#discussion_r1065554986:

> Also, the other scripts are in the `ci` subdirectory of `./scripts` instead of directly in it.

Can you move it there, too?

amesgen (changes requested):

.github/workflows/check-cabal-files.yml
newhoggy: Moved!

scripts/check-cabal-files.sh
newhoggy: Moved!

bolt12 (approved):
LGTM thank you! :D

ouroboros-consensus/docs/report/unsoundswitch/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Revision history for unsoundswitch
```
amesgen, line 1: unsoundswitch is a purely internal thing; in particular, it is not published to CHaP, so I don't think it needs a changelog.

amesgen (changes requested):
Can you revert adding all of the `CHANGELOG.md` files for the Consensus packages? We do not yet have finally decided on how to track changes across the different packages, so we would like to move this to a dedicated PR, such that this PR is not blocked. Thanks!

(Hopefully last time requesting changes :smile:)

bolt12:
How's this PR coming a long? It'd be nice to be able to compile cardano-node against most recent ouroboros-network master branch :)

amesgen (approved):
Eventually, the CI changes should also be applied to the Haddock CI (see also #4299), but that doesn't have to happen in this PR.

ouroboros-consensus-shelley/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Revision history for ouroboros-consensus-shelley
+
+## 0.1.0.0 -- 2020-03-24
+
+* Initial release
```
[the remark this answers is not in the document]
newhoggy: Fixed!

PR #4263, opened 2023-01-10
`StaticEither` accessors

jorisdral:
# Description

The `StaticEither` datatype encodes its "left-ness" or "right-ness" in a type-level `Bool`.

```haskell
data StaticEither :: Bool -> Type -> Type -> Type where
  StaticLeft  :: l -> StaticEither False l r
  StaticRight :: r -> StaticEither True  l r
```

As a consequence, when the type-level `Bool` becomes concrete (i.e., `'True` or `'False'`), we can only match on one of the constructors.

```haskell
case sl of StaticLeft x -> x
let StaticLeft x = sl
case sr of StaticRight x -> x
let StaticRight x = sr 
```

We add two helper functions, that are analogous to `fromLeft` and `fromRight`, but without the default value that `fromLeft` and `fromRight` require to be total functions.

```haskell
fromStaticLeft :: StaticEither 'False l r -> l
fromStaticLeft = \case (StaticLeft x) -> x

fromStaticRight :: StaticEither 'True l r -> r
fromStaticRight = \case (StaticRight x) -> x
```

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
@@ -453,7 +453,7 @@ instance Bifunctor (StaticEither b) where
   second f (StaticRight r) = StaticRight (f r)
 
 fromStaticLeft :: StaticEither 'False l r -> l
-fromStaticLeft = \case (StaticLeft x) -> x
+fromStaticLeft (StaticLeft x) = x
 
 fromStaticRight :: StaticEither 'True l r -> r
-fromStaticRight = \case (StaticRight x) -> x
+fromStaticRight (StaticRight x) = x
--- a/ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
@@ -18,7 +18,6 @@ module Ouroboros.Consensus.Util (
   , Some (..)
   , SomePair (..)
   , SomeSecond (..)
-  , StaticEither (..)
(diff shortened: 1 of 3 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Util.hs
```diff
@@ -444,3 +451,9 @@ instance Bifunctor (StaticEither b) where
 
   second _ (StaticLeft l)  = StaticLeft l
   second f (StaticRight r) = StaticRight (f r)
+
+fromStaticLeft :: StaticEither 'False l r -> l
+fromStaticLeft = \case (StaticLeft x) -> x
+
+fromStaticRight :: StaticEither 'True l r -> r
+fromStaticRight = \case (StaticRight x) -> x
```
amesgen, line 459: Is there a particular reason why this written using `LambdaCase`, and not e.g. plainly like this:
```suggestion
fromStaticLeft :: StaticEither 'False l r -> l
fromStaticLeft (StaticLeft x) = x

fromStaticRight :: StaticEither 'True l r -> r
fromStaticRight (StaticRight x) = x
```

jorisdral: I copied what I had suggested a while ago, but I don't remember why I chose to use `LambdaCase`. Remove the `LambdaCase`!

Jasagredo (approved):
Looks good

PR #4269, opened 2023-01-11
Split function for anti-diff finger tree that is based on heuristics

### LINKED ISSUE #4268: Improved splits for anti-diff finger trees
Anti-diff finger trees are distinct from regular finger trees in the sense that we also a store a measure in the root of the tree, which is distinct from the measures that we store in the nodes of the tree. This "root measure" should be a `Group`, such that if we split/uncons/unsnoc anti-diff finger trees, we can construct the new root measure using subtraction. For splits in particular, performance depends on which part of the split we use to subtract.

```haskell
-- | /O(log(min(i,n-i))) + O(i)/. Specialisation of @'split'@ that is fast if
-- @i@ is small.
splitl ::
     SuperMeasured vr vi a
  => (vi -> Bool)
  -> StrictFingerTree vr vi a
  -> ( StrictFingerTree vr vi a
     , StrictFingerTree vr vi a
     )

-- | /O(log(min(i,n-i))) + O(n-i)/. Specialisation of @'split'@ that is fast if
-- if @i@ is large.
splitr ::
     SuperMeasured vr vi a
  => (vi -> Bool)
  -> StrictFingerTree vr vi a
  -> ( StrictFingerTree vr vi a
     , StrictFingerTree vr vi a
     )
```

The downside to these two functions is that performance depends on educated guesses/estimations of which part of the split is going to be smallest. In case we do not know the shape of the result, we would prefer to compute a heuristic that determines which part of the split we use to subtract.

jorisdral:
# Description

Closes #4268.

We add a `splitSized` function that computes a split and the corresponding split root measures using a heuristic, which the `Sized` class provides. One example of a sensible `Sized` instance would be one that computes the length of a finger tree. We can view the `splitSized` function as an equivalent function that delegates to `splitl` and `splitr` based on the heuristic.

We also perform comparative benchmarks, where we compare the performance of `splitl`, `splitr` and `splitSized` for three scenarios: 
1. `splitl'`'s worst-case running time: a split at the end of the finger tree.
2. `splitr's`'s worst-case running time: a split at the start of the finger tree.
3. A split the middle of the finger tree.

In scenario 1, `splitl` performs much better than `splitr`, which was expected, and `splitc` performs only slightly worse than `splitl`. In scenario 2, `splitl` performs much worse than `splitr`, which is expected, and `splitSized` performs only slightly worse than `splitr`. In Scenario 3, we see nearly identical performance for all three split functions.

# Checklist

- Branch
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
+++ b/anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE BangPatterns           #-}
 {-# LANGUAGE ConstraintKinds        #-}
 {-# LANGUAGE DeriveGeneric          #-}
-{-# LANGUAGE DerivingStrategies     #-}
 {-# LANGUAGE FlexibleContexts       #-}
 {-# LANGUAGE FlexibleInstances      #-}
 {-# LANGUAGE FunctionalDependencies #-}
@@ -19,8 +18,10 @@ module Data.FingerTree.RootMeasured.Strict (
   , fromList
   , (|>)
     -- * Splitting
+  , Sized (..)
   , SplitRootMeasure (..)
   , split
+  , splitSized
   , splitl
   , splitr
     -- * Maps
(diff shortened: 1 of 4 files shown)


anti-diff/src/Data/FingerTree/RootMeasured/Strict.hs
```diff
@@ -176,39 +169,58 @@ newtype SplitRootMeasure vr vi a = SplitRootMeasure {
...
+-- delegates to @'splitl'@ or @'splitr'@ based on whether @i@ or @n-i@ are
+-- smallest respectively.
+splitSized ::
+     (SuperMeasured vr vi a, Sized vi)
+  => (vi -> Bool)
+  -> StrictFingerTree vr vi a
+  -> ( StrictFingerTree vr vi a
+     , StrictFingerTree vr vi a
+     )
+splitSized p = split p $ SplitRootMeasure $  \vr (l, r) ->
```
Jasagredo, line 214:
```suggestion
splitSized p = split p $ SplitRootMeasure $ \vr (l, r) ->
```

Jasagredo (approved):
Looks good to me. Thanks!

PR #4274, opened 2023-01-11
PBT for laws on TableStuff and StowableLedgerTables

### LINKED ISSUE #4270: Contracts on the instances of `StowableLedgerTables`, at least that `unstow . stow= id`, `stow . unstow` will overwrite the ledger tables

Jasagredo:
# Description

Implements PBT for laws on `TableStuff` and `StowableLedgerTables` for the Cardano blocks.

Closes #4270 and #4271

### DIFF
--- a/ouroboros-consensus-byron-test/test/Test/Consensus/Byron/LedgerTables.hs
+++ b/ouroboros-consensus-byron-test/test/Test/Consensus/Byron/LedgerTables.hs
@@ -1,0 +1,17 @@
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Consensus.Byron.LedgerTables (tests) where
+
+import           Ouroboros.Consensus.Byron.Ledger
+
+import           Test.Consensus.Byron.Generators ()
+import           Test.LedgerTables
+
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+
+tests :: TestTree
+tests = testGroup "LedgerTables"
+  [ testProperty "Stowable laws" (prop_stowable_laws @ByronBlock)
+  , testProperty "TableStuff laws" (prop_tablestuff_laws @ByronBlock)
+  ]
--- a/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
+++ b/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
@@ -1,0 +1,90 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE TypeApplications  #-}
+{-# LANGUAGE TypeFamilies      #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Test.Consensus.Cardano.LedgerTables (tests) where
+
+import           Data.Map.Diff.Strict (Values (..))
+import qualified Data.Map.Strict as Map
+import           Data.SOP.Strict
+
+import           Ouroboros.Consensus.Cardano.Block
+import           Ouroboros.Consensus.Cardano.CanHardFork
+import           Ouroboros.Consensus.HardFork.Combinator
+import           Ouroboros.Consensus.HardFork.Combinator.State.Types
+import           Ouroboros.Consensus.HardFork.Combinator.Util.Functors
+import           Ouroboros.Consensus.HardFork.Combinator.Util.Telescope
+import           Ouroboros.Consensus.Ledger.Tables
+import           Ouroboros.Consensus.Protocol.Praos.Translate ()
+import           Ouroboros.Consensus.Shelley.HFEras ()
+import           Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
+import           Ouroboros.Consensus.Shelley.ShelleyHFC (ShelleyTxOut (..))
+import           Ouroboros.Consensus.Util.SOP
+
+import qualified Cardano.Ledger.Shelley.API as SL
+
+import           Test.Cardano.Ledger.Alonzo.Serialisation.Generators ()
+import           Test.Cardano.Ledger.Babbage.Serialisation.Generators ()
+
+import           Test.LedgerTables
+
+import           Test.Consensus.Byron.Generators ()
+import           Test.Consensus.Cardano.MockCrypto (MockCryptoCompatByron)
+import           Test.Consensus.Shelley.Generators ()
+
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+
+type Crypto = MockCryptoCompatByron
+
+tests :: TestTree
+tests = testGroup "LedgerTables"
+  [ testGroup "Cardano"
+    [ testProperty "Stowable laws" (prop_stowable_laws @(CardanoBlock Crypto))
+    , testProperty "TableStuff laws"
+        (\(InSameEra ls tbs) -> prop_tablestuff_laws @(CardanoBlock Crypto) ls tbs)
+    ]
+  ]
+
+instance Arbitrary (LedgerTables (LedgerState (CardanoBlock Crypto)) ValuesMK) where
+  arbitrary = projectLedgerTables . unstowLedgerTables <$> arbitrary
+
+instance Show (LedgerTables (LedgerState (CardanoBlock Crypto)) EmptyMK) where
+  show x = showsLedgerState x ""
+
+instance Show (LedgerTables (LedgerState (CardanoBlock Crypto)) ValuesMK) where
+  show x = showsLedgerState x ""
+
+-- | We need to generate Ledger states and tables in the same era or otherwise
+-- it wouldn't make sense to try to attach them to a hard fork ledger state,
+-- this data definition does just that.
+data InSameEra =
+  InSameEra
+    !(LedgerState (CardanoBlock Crypto) EmptyMK)
+    !(LedgerTables (LedgerState (CardanoBlock Crypto)) ValuesMK)
+  deriving Show
+
+instance Arbitrary InSameEra where
+  arbitrary = do
+    ls <- arbitrary
+    tbs <- arbitrary `suchThat` (sameEra (getHardForkState $ hardForkLedgerStatePerEra ls) . cardanoUTxOTable)
+    pure $ InSameEra ls tbs
+
+-- | This function asserts that:
+--
+-- - Producing a table to be attached to a Byron HardFork State requires the
+--   table to be empty as that is the only kind of table such a state can have.
+--
+-- - Producing a table to be attached to a Shelley HardFork State requires all
+-- - the entries in the map to be on the same era as the state. We do so by
+-- - checking that all of them and the state produce the same index through
+-- - @'nsToIndex'@.
+sameEra ::
+     Telescope (K Past) (Current (Flip LedgerState EmptyMK)) (CardanoEras Crypto)
+  -> ApplyMapKind ValuesMK (SL.TxIn Crypto) (ShelleyTxOut (ShelleyBasedEras Crypto))
+  -> Bool
+sameEra (TZ{})      (ApplyValuesMK (Values m))      = Map.null m
+sameEra (TS _ tele) (ApplyValuesMK (Values values)) =
+    all (nsToIndex (tip tele) ==) $ map (nsToIndex . unShelleyTxOut . snd) (Map.toList values)
--- a/ouroboros-consensus-shelley-test/test/Test/Consensus/Shelley/LedgerTables.hs
+++ b/ouroboros-consensus-shelley-test/test/Test/Consensus/Shelley/LedgerTables.hs
@@ -1,0 +1,73 @@
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE TypeFamilies         #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Test.Consensus.Shelley.LedgerTables (tests) where
+
+import           Ouroboros.Consensus.Ledger.Tables
+import           Ouroboros.Consensus.Shelley.Eras
+import           Ouroboros.Consensus.Shelley.Ledger
+
+import           Cardano.Crypto.Hash (ShortHash)
+
+import           Ouroboros.Consensus.Protocol.Praos (Praos)
+import           Ouroboros.Consensus.Protocol.Praos.Translate ()
+import           Ouroboros.Consensus.Protocol.TPraos (TPraos)
+import           Ouroboros.Consensus.Shelley.HFEras ()
+import           Ouroboros.Consensus.Shelley.Ledger.SupportsProtocol ()
+
+import           Test.Cardano.Ledger.Alonzo.Serialisation.Generators ()
+import           Test.Cardano.Ledger.Babbage.Serialisation.Generators ()
+
+import           Test.LedgerTables
+
+import           Test.Consensus.Shelley.Generators ()
+import           Test.Consensus.Shelley.MockCrypto (CanMock, MockCrypto)
+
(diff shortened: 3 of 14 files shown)


ouroboros-consensus-test/src/Test/LedgerTables.hs
```diff
@@ -0,0 +1,51 @@
...
+     ( TableStuff (LedgerState blk)
+     , Eq (LedgerTables (LedgerState blk) EmptyMK)
+     , Show (LedgerTables (LedgerState blk) EmptyMK)
+     , Show (LedgerTables (LedgerState blk) ValuesMK)
+     )
+  => LedgerState blk EmptyMK
+  -> LedgerTables (LedgerState blk) ValuesMK
+  -> Property
+prop_tablestuff_laws = \ls tbs ->
+  compareTables (ls `withLedgerTables` (projectLedgerTables ls)) ls .&&.
```
jorisdral, line 50: Why not compare the ledger states instead of the ledger tables? This would give a stronger property

ouroboros-consensus-test/src/Test/LedgerTables.hs
```diff
@@ -0,0 +1,51 @@
...
+     , Show (LedgerTables (LedgerState blk) ValuesMK)
+     )
+  => LedgerState blk EmptyMK
+  -> Property
+prop_stowable_laws = \ls ->
+  compareTables (stowLedgerTables (unstowLedgerTables ls)) ls .&&.
+  let ls' = unstowLedgerTables ls
+  in compareTables (unstowLedgerTables (stowLedgerTables ls')) ls'
+
+prop_tablestuff_laws ::
```
jorisdral, line 40: Maybe add a comment to these properties that describes the laws, since it is a little hard to read from the code itself what the laws are that are being tested. Something like these crude examples from the issue title: `with . project = id` and `unstow . stow= id`

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
```diff
@@ -0,0 +1,79 @@
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE TypeFamilies         #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
```
jorisdral, line 4:
```suggestion
{-# OPTIONS_GHC -Wno-orphans #-}

```

ouroboros-consensus-shelley-test/test/Test/Consensus/Shelley/LedgerTables.hs
```diff
@@ -0,0 +1,74 @@
...
+    ]
+  ]
+
+
+instance
+  ( CanMock proto era
+  , Arbitrary (LedgerState (ShelleyBlock proto era) EmptyMK)
+  , StowableLedgerTables (LedgerState (ShelleyBlock proto era))
+  )
+  => Arbitrary (LedgerTables (LedgerState (ShelleyBlock proto era)) ValuesMK) where
```
jorisdral, line 65: The style guide says to do this

```suggestion
instance ( CanMock proto era
         , Arbitrary (LedgerState (ShelleyBlock proto era) EmptyMK)
         , StowableLedgerTables (LedgerState (ShelleyBlock proto era))
         ) => Arbitrary (LedgerTables (LedgerState (ShelleyBlock proto era)) ValuesMK) where
```

ouroboros-consensus-shelley-test/test/Test/Consensus/Shelley/LedgerTables.hs
```diff
@@ -0,0 +1,74 @@
...
+  arbitrary = projectLedgerTables . unstowLedgerTables <$> arbitrary
+
+instance ShelleyBasedEra era =>
+         Show (LedgerTables (LedgerState (ShelleyBlock proto era)) EmptyMK) where
```
jorisdral, line 69:
```suggestion
instance ShelleyBasedEra era 
      => Show (LedgerTables (LedgerState (ShelleyBlock proto era)) EmptyMK) where
```

And idem for other instances

ouroboros-consensus-shelley-test/test/Test/Consensus/Shelley/LedgerTables.hs
```diff
@@ -0,0 +1,74 @@
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE TypeFamilies         #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
```
jorisdral, line 6:
```suggestion
{-# OPTIONS_GHC -Wno-orphans #-}

```

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
```diff
@@ -0,0 +1,79 @@
...
+  show x = showsLedgerState x ""
+
+-- | We need to generate Ledger states and tables in the same era or otherwise
+-- it wouldn't make sense to try to attach them to a hard fork ledger state,
+-- this data definition does just that.
+data InSameEra =
+  InSameEra ( LedgerState (CardanoBlock Crypto) EmptyMK
+            , LedgerTables (LedgerState (CardanoBlock Crypto)) ValuesMK
+            )
+  deriving (Show)
```
jorisdral, line 65: Make it a `newtype`, or make it a product type without the tuple

jorisdral, line 65:
```suggestion
data InSameEra =
  InSameEra ( LedgerState (CardanoBlock Crypto) EmptyMK
            , LedgerTables (LedgerState (CardanoBlock Crypto)) ValuesMK
            )
  deriving Show
```

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
```diff
@@ -0,0 +1,79 @@
...
+
+instance Arbitrary InSameEra where
+  arbitrary = do
+    ls <- arbitrary
+    tbs <- arbitrary `suchThat` (sameEra (getHardForkState $ hardForkLedgerStatePerEra ls) . cardanoUTxOTable)
+    pure $ InSameEra (ls, tbs)
+
+sameEra ::
```
jorisdral, line 73: Maybe a small comment that describes "how" the function determines that all are in the same era?

jorisdral (commented):
Nice! Maybe we can make a follow-up issue to make a list of laws beyond the ones that we have now?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
Jasagredo: Done

ouroboros-consensus-test/src/Test/LedgerTables.hs
Jasagredo: Added

ouroboros-consensus-test/src/Test/LedgerTables.hs
Jasagredo: I added a comment explaining why. Basically the ledger state can be arbitrarily complex/big and comparing the whole ledger state takes a non-trivial amount of time more. But it can be done

jorisdral (approved):

ouroboros-consensus-byron-test/test/Test/Consensus/Byron/LedgerTables.hs
```diff
@@ -0,0 +1,17 @@
...
+import           Test.Consensus.Byron.Generators ()
+import           Test.LedgerTables
+
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+
+tests :: TestTree
+tests = testGroup "LedgerTables"
+  [ testProperty "Stowable laws" (prop_stowable_laws @ByronBlock)
+  , testProperty "TableStuff laws" (prop_tablestuff_laws @ByronBlock)
```
dnadales, line 16: Nothing we can do now, but I wonder if we should have renamed table stuff earlier :grimacing:

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
```diff
@@ -0,0 +1,90 @@
...
+tests = testGroup "LedgerTables"
+  [ testGroup "Cardano"
+    [ testProperty "Stowable laws" (prop_stowable_laws @(CardanoBlock Crypto))
+    , testProperty "TableStuff laws"
+        (\(InSameEra ls tbs) -> prop_tablestuff_laws @(CardanoBlock Crypto) ls tbs)
+    ]
+  ]
+
+instance Arbitrary (LedgerTables (LedgerState (CardanoBlock Crypto)) ValuesMK) where
+  arbitrary = projectLedgerTables . unstowLedgerTables <$> arbitrary
```
dnadales, line 52: We're defining the `Arbitrary` instance for `LedgerTables CardanoBlock` in terms of the larger `LedgerState CardanoBlock`. I would expect we have the arbitrary definition here and use it in the `Arbitrary` instance for `LedgerState`.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
```diff
@@ -0,0 +1,90 @@
...
+-- | This function asserts that:
+--
+-- - Producing a table to be attached to a Byron HardFork State requires the
+--   table to be empty as that is the only kind of table such a state can have.
+--
+-- - Producing a table to be attached to a Shelley HardFork State requires all
+-- - the entries in the map to be on the same era as the state. We do so by
+-- - checking that all of them and the state produce the same index through
+-- - @'nsToIndex'@.
+sameEra ::
```
dnadales, line 84: Hmmm, we have a run-time check for this. Can't this function be defined at the type level?

ouroboros-consensus-test/src/Test/LedgerTables.hs
```diff
@@ -0,0 +1,63 @@
...
+
+import           Data.Function (on)
+
+import           Ouroboros.Consensus.Ledger.Basics
+import           Ouroboros.Consensus.Ledger.Tables
+
+import           Test.QuickCheck
+
+-- | We will compare the Ledger Tables of the result because the comparison for
+-- the rest of the LedgerState takes considerably more time to run.
```
dnadales, line 16:
```suggestion
-- | We compare the Ledger Tables of the result because the comparison with
-- the rest of the LedgerState takes considerably more time to run.
```

ouroboros-consensus-test/src/Test/LedgerTables.hs
```diff
@@ -0,0 +1,63 @@
...
+     , TableStuff (LedgerState blk)
+     , Eq (LedgerTables (LedgerState blk) mk)
+     , Show (LedgerTables (LedgerState blk) mk)
+     )
+  => LedgerState blk mk
+  -> LedgerState blk mk
+  -> Property
+compareTables = (===) `on` projectLedgerTables
+
+-- | The StowableLedgerTables instances should follow these two laws:
```
dnadales, line 28: This comment is really helpful :heart:

ouroboros-consensus-test/src/Test/LedgerTables.hs
```diff
@@ -0,0 +1,63 @@
...
+     , StowableLedgerTables (LedgerState blk)
+     , Eq (LedgerTables (LedgerState blk) EmptyMK)
+     , Show (LedgerTables (LedgerState blk) EmptyMK)
+     , Show (LedgerTables (LedgerState blk) ValuesMK)
+     )
+  => LedgerState blk EmptyMK
+  -> Property
+prop_stowable_laws = \ls ->
+  compareTables (stowLedgerTables (unstowLedgerTables ls)) ls .&&.
+  let ls' = unstowLedgerTables ls
```
dnadales, line 44: Why do we need this `unstow`? This differs from the comment. Maybe we should clarify this.

ouroboros-consensus-test/src/Test/LedgerTables.hs
```diff
@@ -0,0 +1,63 @@
...
+     ( TableStuff (LedgerState blk)
+     , StowableLedgerTables (LedgerState blk)
+     , Eq (LedgerTables (LedgerState blk) EmptyMK)
+     , Show (LedgerTables (LedgerState blk) EmptyMK)
+     , Show (LedgerTables (LedgerState blk) ValuesMK)
+     )
+  => LedgerState blk EmptyMK
+  -> Property
+prop_stowable_laws = \ls ->
+  compareTables (stowLedgerTables (unstowLedgerTables ls)) ls .&&.
```
dnadales, line 43: If we'd introduce an operator for `compareTables` maybe this could be rewritten as:

```haskell
stowLedgerTables (unstowLedgerTables ls) ==? ls
```

or alternatively:

```haskell
stowLedgerTables (unstowLedgerTables ls) `compareTables` ls
```

ouroboros-consensus-test/src/Test/LedgerTables.hs
dnadales: But besides the runtime cost, isn't comparing /just the ledger tables/ the right thing to do in this case?

dnadales (approved):
We're missing the definition of `shrink` in several `Arbitrary` instances. Any chance we can define them in this PR?

Other than that it looks great!

ouroboros-consensus-test/src/Test/LedgerTables.hs
jorisdral: >But besides the runtime cost, isn't comparing /just the ledger tables/ the right thing to do in this case?

Since we're testing `withLedgerTables . projectLedgerTabels == id`, I can agree that it is the right thing to do. Still, maybe another law could be added which states that `withLedgerTables` should *only* change the ledger tables and not any other part of the ledger state

ouroboros-consensus-byron-test/test/Test/Consensus/Byron/LedgerTables.hs
jorisdral: `HasTables`/`HasLedgerTables` :upside_down_face:

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
Jasagredo: This is not exactly the way it is. We are defining the `Arbitrary` instance for `LedgerTables CardanoBlock` in terms of the instance `Arbitrary (LedgerState (HardForkBlock xs) mk)` which defer to each generator for each era. As the Cardano tables are defined for the Cardano Block and not compositionally I don't think we can flip the relation here

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/LedgerTables.hs
Jasagredo: I don't think we can. We cannot introspect into which index an NS is at, i.e. for this:
```
Z (I 'x')      :: NS I       '[ Char, Bool ]
S (Z (I True)) :: NS I       '[ Char, Bool ]
```
they both have the same type, so we have to pattern match on the value to find out the index.

We could leverage the `NS` to carry an index but I think that would trash away the NS purpose and the `nsToIndex` is intended to be used precisely for this

ouroboros-consensus-test/src/Test/LedgerTables.hs
Jasagredo: I will keep this as is. We can add the full check later.

PR #4275, opened 2023-01-11
Test.ChainProducerState: add FlexibleContexts

locallycompact:
This fixes an issue on ghc 9.4


coot:
I checked that the package builds with `ghc-9.4` and `ghc-9.6` without this change.

Issue #3596, opened 2022-02-04
Light Peer Sharing


coot:
Make inbound connections available to the outbound governor.

PR #4277, opened 2023-01-12
Implement Light Peer Sharing

bolt12:
This PR introduces Light Peer Sharing on top of the already existing Peer Sharing feature. They're actually somewhat independent, but one should be careful with this feature since it will be prone to Eclipse attacks without the Eclipse Evasion implementation.

The main objective here is to use the information that the Inbound Governor has about incoming connections and feed that information directly to the Peer Selection Governor (aka Outbound Governor). Right now the Outbound Governor only has peers information (Known / Established / Active Peers Set etc..) from local configuration and/or ledger peers. With Light Peer Sharing, when a node connects to us we will make an effort to add it to our Known Peer Set, being yet another source of peers for fulfilling this target and possibly share across the network

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested


bolt12:
Closes #3596

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -441,18 +441,23 @@ peerSelectionGovernor :: (MonadAsync m, MonadLabelledSTM m, MonadMask m,
...
+peerSelectionGovernor tracer debugTracer countersTracer fuzzRng mbLocalIPv4
+                      mbLocalIPv6 stateVar actions policy =
     JobPool.withJobPool $ \jobPool -> do
       localPeers <- map (\(target, _) -> (target, 0))
                 <$> atomically (readLocalRootPeers actions)
       peerSelectionGovernorLoop
         tracer
         debugTracer
         countersTracer
+        mbLocalIPv4
```
njd42, line 459: Question - why are these now necessary? are you implying that the `peerSelectionGovernor` needs to know "its own addresses" to operate?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
```diff
@@ -176,9 +183,14 @@ jobPeerShare PeerSelectionActions{requestPeerShare}
                -- which means other smaller valency nodes won't be shared that
                -- much. So any chance we can have of filtering them from peer
                -- sharing results is welcome.
+               --
+               -- We also filter ourselves from the result set.
                newPeers    = [ p | Right (PeerSharingResult ps) <- totalResults
                                  , p <- ps
-                                 , not (isKnownLedgerPeer p (knownPeers st)) ]
+                                 , not (isKnownLedgerPeer p (knownPeers st))
+                                 , Just p /= mbLocalIPv4
```
njd42, line 191: I'm not certain this will give the behaviour you want under various deployment scenarios. 

It will "work" for a single connected host directly with a public IP address, in other scenarios loopback connections may well be established.

What is the consequence of creating a loopback connection?

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -137,6 +137,8 @@ tests =
   , testProperty "cm & ig timeouts enforced"
                  prop_diffusion_timeouts_enforced
   , testProperty "unit #4177" unit_4177
+  , testProperty "never connect to self"
```
njd42, line 140: What is so about self-connecting? (see comment earlier)

njd42 (commented):
Looks good - just the question, for me, about the consequences of self-connection. Is this something we just want to avoid in tests (as kernel would disallow) or is it something that we need to avoid in operation (given that for NAT and multi-homed hosts it could happen)

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor/ControlChannel.hs
```diff
@@ -53,31 +55,41 @@ instance Show peerAddr
...
+    ControlChannel (NewConnection peerAddr (Handle muxMode peerAddr versionData bytes m a b)) m
+
+-- | Control Channel between Server and Outbound Governor.
+--
+-- Control channel that is meant to share inbound connections with the Peer
+-- Selection Governor. So the consumer is the Governor and Producer is the
+-- Server.
+--
+type GovernorControlChannel peerAddr m =
+    ControlChannel (peerAddr, PeerSharing) m
```
coot, line 74: If we want to reuse this type in the `outbound-governor` we also should rename this module; otherwise it's confusing to use something below `Ouroboros.Network.InboundGovernor` in the `outbound-governor`.

Since both channels are used by the `connection-manager`, maybe we should rename this module to `Ouroboros.Network.ConnectionManager.ControlChannel`.

If we are at this maybe there's actually a better name name `ControlChannel`, it's not much about controlling something but passing information / notifications around.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor/ControlChannel.hs
```diff
@@ -53,31 +55,41 @@ instance Show peerAddr
...
 
 
--- | A Server control channel which instantiates 'handle'.
+-- | A Server control channel which instantiates to 'NewConnection' and 'Handle'.
+--
+-- It allows to pass 'STM' transactions which will resolve to 'NewConnection'.
+-- Server's monitoring thread is the consumer of these messages; there are two
+-- producers: accept loop and connection handler for outbound connections.
 --
 type ServerControlChannel (muxMode :: MuxMode) peerAddr versionData bytes m a b =
```
coot, line 64: What about renaming this type to `InboundGovernorControlChannel` and the other one to `OutboundGovernorChannel`.

ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
```diff
@@ -90,6 +91,16 @@ data ServerArguments (muxMode  :: MuxMode) socket peerAddr versionData versionNu
       serverControlChannel        :: ServerControlChannel muxMode peerAddr versionData
                                                           bytes m a b,
 
+      -- | Governor control channel var is passed as an argument; this allows the Server
+      -- to communicate with the Outbound Governor telling it about new inbound
+      -- connections to add to the Known Peers Set.
+      --
+      governorControlChannel        :: GovernorControlChannel peerAddr m,
```
coot, line 98: I wouldn't do that by the server.  The connection manager is passing outbound duplex connections to the inbound governor, so why not let it pass inbound duplex connections to the outbound governor.  This will be more symmetric.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -176,10 +180,18 @@ inboundGovernor trTracer tracer serverControlChannel inboundIdleTimeout
...
+            (Handle csMux muxBundle _ versionData)) -> do
 
               traceWith tracer (TrNewConnection provenance connId)
 
+              -- Comunicate this new inbound connection to the Outbound Governor
+              when (provenance == Inbound) $
+                atomically $
+                  ControlChannel.writeMessage governorControlChannel
+                                              ( remoteAddress connId
+                                              , getPeerSharing versionData
```
coot, line 192: We should only pass duplex connections to the outbound governor.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -91,6 +94,34 @@ jobs jobPool st =
...
+                      }
+                      st@PeerSelectionState {
+                        knownPeers
+                      } =
+    Guarded Nothing $ do
+      (addr, ps) <- readNewInboundConnection
+      return $ \_ ->
+        let -- If peer happens to already be present in the Known Peer set
+            -- 'insert' is going to do its due diligence before adding.
+            newEntry    = Map.singleton addr (ps, DoAdvertisePeer, IsNotLedgerPeer)
```
coot, line 115: So we now have `PeerSharing` and `AdvertisePeer`. `PeerSharing` means if one is running the peer sharing protocol and `AdvertisePeer` is whether the peer should be shared, isn't it?  Don't we need to advertise `AdvertisePeer` through handshake too?  If a BP connects to a relay and it doesn't wishes to be advertised, how can it be achieved?

coot (changes requested):

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor/ControlChannel.hs
bolt12: That's actually better!

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
bolt12: Docs for `NewConnection` say:

> A request to start mini-protocol bundle, either from the server or from
> connection manager after a duplex connection was negotiated.

Also the comment above says that:

> new connection has been announced by either accept loop or
> by connection manager (in which case the connection is in
> 'DuplexState').

So I guess this is already happening, right?

ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
bolt12: See comment above. I'll update this comment though

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
coot: This is implemented for passing outbound duplex connections to the inbound governor, but not when passing inbound connections to the outbound governor (the light peer sharing side).

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
bolt12: This is more a test for the test suite as our network simulation isn't really capable of dealing with self connects, so this test is to catch such cases and not have them appear in other tests disguised

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
coot: ok, thanks.  Btw, we should document this somewhere visible to SPOs, e.g. in `cardano-node`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -654,6 +656,7 @@ data TracePeerSelection peeraddr =
      | TracePeerShareRequests     Int Int (Set peeraddr) (Set peeraddr)
      | TracePeerShareResults         [(peeraddr, Either SomeException (PeerSharingResult peeraddr))] --TODO: classify failures
      | TracePeerShareResultsFiltered [peeraddr]
+     | TraceNewInboundConnection peeraddr PeerSharing
```
coot, line 659:
```suggestion
     | TraceKnownInboundPeer peeraddr PeerSharing
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -91,6 +94,34 @@ jobs jobPool st =
...
       return (completion st)
 
+-- | Monitor new inbound connections
+--
+newInboundConnections :: forall m peeraddr peerconn.
+                          (MonadSTM m, Ord peeraddr)
+                      => PeerSelectionActions peeraddr peerconn m
+                      -> PeerSelectionState peeraddr peerconn
+                      -> Guarded (STM m) (TimedDecision m peeraddr peerconn)
+newInboundConnections PeerSelectionActions{
```
coot, line 104:
```suggestion
inboundPeers :: forall m peeraddr peerconn.
                (MonadSTM m, Ord peeraddr)
             => PeerSelectionActions peeraddr peerconn m
             -> PeerSelectionState peeraddr peerconn
             -> Guarded (STM m) (TimedDecision m peeraddr peerconn)
inboundPeers PeerSelectionActions {
```

coot (changes requested):
A few minor remarks.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
bolt12: I'll make a note and as soon as I am able to write a PR for cardano node I'll include it!

bolt12: I did this on the Peer Sharing PR that made it to cardano-node. I left it in the commit message

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -135,7 +138,10 @@ data ConnectionManagerArguments handlerTrace socket peerAddr handle handleError
         -- | Prune policy
         --
         cmPrunePolicy         :: PrunePolicy peerAddr (STM m),
-        cmConnectionsLimits   :: AcceptedConnectionsLimit
+        cmConnectionsLimits   :: AcceptedConnectionsLimit,
+
+        -- | How to extract PeerSharing information from versionData
+        cmGetPeerSharing        :: versionData -> PeerSharing
```
coot, line 144:
```suggestion
        cmGetPeerSharing      :: versionData -> PeerSharing
```
Let's align the `::`.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -501,6 +504,27 @@ runResponder mux
...
+    InformationChannel (NewConnectionInfo peerAddr (Handle muxMode peerAddr versionData bytes m a b)) m
+
+-- | Control Channel between Server and Outbound Governor.
+--
+-- Control channel that is meant to share inbound connections with the Peer
+-- Selection Governor. So the consumer is the Governor and Producer is the
+-- Server.
+--
+type OutboundGovernorInfoChannel peerAddr m =
+    InformationChannel (peerAddr, PeerSharing) m
```
coot, line 527: I think a better place for these two type aliases is the module where `InformationChannel` is defined.  It's a bit confusing why `OutboundGovernorInfoChannel` is defined in `Ouroboros.Network.InboundGovernor`.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor/ControlChannel.hs
coot: The `Ouroboros.Network.InboundGovernor.InboundGovernorInfoChannel` should be moved here.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor/ControlChannel.hs
coot: The `Ouroboros.Network.InboundGovernor.OutboundGovernorInfoChannel` should replace this type alias here.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -566,10 +566,11 @@ peerSelectionGovernorLoop tracer
...
       -- All the alternative potentially-blocking decisions.
-         Monitor.connections          actions st
-      <> Monitor.jobs                 jobPool st
-      <> Monitor.targetPeers          actions st
-      <> Monitor.localRoots           actions policy st
+         Monitor.connections           actions st
+      <> Monitor.jobs                  jobPool st
+      <> Monitor.targetPeers           actions st
+      <> Monitor.localRoots            actions policy st
+      <> Monitor.inboundPeers         actions st
```
coot, line 573:
```suggestion
         Monitor.connections          actions st
      <> Monitor.jobs                 jobPool st
      <> Monitor.targetPeers          actions st
      <> Monitor.localRoots           actions policy st
      <> Monitor.inboundPeers         actions st
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -91,6 +94,34 @@ jobs jobPool st =
...
+          -- 'insert' is going to do its due diligence before adding.
+          newEntry    = Map.singleton addr (ps, DoAdvertisePeer, IsNotLedgerPeer)
+          knownPeers' = KnownPeers.insert newEntry knownPeers
+       in Decision {
+            decisionTrace = [TraceKnownInboundConnection addr ps],
+            decisionJobs = [],
+            decisionState =
+              st {
+                knownPeers = knownPeers'
+              }
```
coot, line 123:
```suggestion
            decisionTrace = [TraceKnownInboundConnection addr ps],
            decisionJobs  = [],
            decisionState = st { knownPeers = knownPeers' }
```

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/MockEnvironment.hs
```diff
@@ -57,11 +57,11 @@ import           Ouroboros.Network.PeerSelection.Governor hiding
...
-import           Ouroboros.Network.Testing.Data.Script (PickScript,
-                     ScriptDelay (..), TimedScript, arbitraryPickScript,
-                     initScript', interpretPickScript, playTimedScript,
-                     prop_shrink_Script, singletonScript, stepScript,
-                     stepScriptSTM)
+import           Ouroboros.Network.Testing.Data.Script (NonEmpty (..),
+                     PickScript, Script (..), ScriptDelay (..), TimedScript,
+                     arbitraryPickScript, arbitraryScriptOf, initScript',
+                     interpretPickScript, playTimedScript, prop_shrink_Script,
+                     singletonScript, stepScript)
```
coot, line 64: :thinking: why we need to import `NonEmpty`? In this module we only add `retry` in one of the records.

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -2659,6 +2663,61 @@ prop_diffusion_timeouts_enforced defaultBearerInfo diffScript =
        in getAllProperty
         $ verifyAllTimeouts True transitionSignal
 
+-- | This property checks that a node never connects to itself.
+--
+-- Connecting to itself means connecting to exactly the same address and port
+-- of a node's listening socket. This is something that in the real world
+-- wouldn't happen since the kernel would disallow it.
```
coot, line 2670: I don't think that's true, have you tried?

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/MockEnvironment.hs
bolt12: I don't know, maybe I forgot to clean this import.

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
bolt12: Right, what I meant here is that our net sim would misbehave in the case of self connects, but you fixed it recently, is that right?

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor/ControlChannel.hs
bolt12: Actually this file should be deleted because I refactored it to InformationChannel.hs

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
bolt12: Bump

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -570,6 +570,7 @@ peerSelectionGovernorLoop tracer
       <> Monitor.jobs                 jobPool st
       <> Monitor.targetPeers          actions st
       <> Monitor.localRoots           actions policy st
+      <> Monitor.inboundPeers         actions st
```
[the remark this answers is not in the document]
coot: If an action blocks then we go to the next one, but if we are bombarded with new connections we might fail to progress on reaching the targets.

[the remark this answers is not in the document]
bolt12: We have rate limiting for accepting inbound connections, so I don't think it won't starve the others, but I can do that! 😄

coot (approved):
LGTM

ouroboros-network/test/Test/Ouroboros/Network/Testnet.hs
coot: Yes, it's fixed, it was a bug in `connection-manager` not `sim-net` though.

coot:
Added to merge queue.

PR #4296, opened 2023-01-20
Add the LedgerHD initial sketch

### LINKED ISSUE #4074: Validate prototype by drafting future steps for other tables
Until now we have only had the UTxO table on the UTxO-HD prototype. In the foreseeable future, we will add more tables to the implementation, possibly delegations. 

We should add some other table to the prototype to make sure that we will be able to do so in the future without major modifications to the prototype. The table might be trivial, but it should give us enough confidence that we will be able to support other types of tables that can capture the other large maps that exist in the ledger state.

# Possible approach 

A possible way to approach this is to use SOP in the `LedgerStateKind`:

```haskell
type MapKind = Type -> Type -> Type -> Type
type LedgerStateKind = NP MapKind '[UTxO :: Type, OtherMap] -> Type

data LedgerTables (LedgerState ByronBlock) mk = NoByronLedgerTables

newtype LedgerTables (LedgerState (ShelleyBlock proto era)) mk = ShelleyLedgerTables { 
  npTables :: ZipUncurry mk [ (SL.TxIn (EraCrypto era), Core.TxOut era)    -- k and v for UTxO
                            , (OtherKey era,            OtherValue era) ]  -- k and v for OtherMap
  }
```

In this way, functions that deal with tables would become `NP`s of functions. For example:

```haskell
mapLedgerTables :: NP (mk1 -.2-> mk2) KVs -> LedgerTables l mk1 -> LedgerTables l mk2
``` 

(`-.2->` is meant to be like `-.->` but applies to 2 types, key and value)

Jasagredo:
Closes #4074

### DIFF
--- a/ouroboros-consensus/docs/future-ledger-hd.md
+++ b/ouroboros-consensus/docs/future-ledger-hd.md
@@ -1,0 +1,258 @@
+# Ledger-HD sketch (UTxO-HD v2)
+
+This document describes the result of the discussion between Ledger and
+Consensus teams on 2022-01-17 about the future steps on UTxO-HD which would not
+make sense to be called this way anymore in future versions and therefore we
+propose Ledger-HD as a replacement.
+
+Below, we outline the tables that are expected to be moved to the disk, their
+dynamics, the main computations that will leave the Ledger and some open
+questions.
+
+This is meant to be just a sketch, details are not worked out yet.
+
+## Scope of Ledger-HD
+
+The plan for Ledger-HD is to move the following tables to the disk (we will show
+the "lenses" that reach each data structure from the new epoch state):
+
+- The unified map of `rewards`, `delegations`, `pointers` and `deposits`
+
+```haskell
+( ne :: NewEpochState era )
+  & ( nesEs     :: NewEpochState      era  -> EpochState         era  )
+  & ( esLState  :: EpochState         era  -> LedgerState        era  )
+  & ( lsDPState :: LedgerState        era  -> DPState (EraCrypto era) )
+  & ( dpsDState :: DPState (EraCrypto era) -> DState  (EraCrypto era) )
+  & ( dsUnified :: DState  (EraCrypto era) -> UMap    (EraCrypto era) )
+
+data UMap c = UMap !(Map (Credential 'Staking c) (Trip c)) !(Map Ptr (Credential 'Staking c))
+```
+
+- The current stake distribution per stake credential
+
+```haskell
+( ne :: NewEpochState era )
+  & ( nesEs           :: NewEpochState era -> EpochState                  era  )
+  & ( esLState        :: EpochState    era -> LedgerState                 era  )
+  & ( lsUTxOState     :: LedgerState   era -> UTxOState                   era  )
+  & ( utxosStakeDistr :: UTxOState     era -> IncrementalStake (EraCrypto era) )
+
+data IncrementalStake c = IStake
+  { credMap :: !(Map (Credential 'Staking c) Coin)
+  , ptrMap  :: !(Map Ptr Coin)
+  }
+```
+
+- The stake snapshots
+
+```haskell
+( ne :: NewEpochState era )
+  & ( nesEs       :: NewEpochState era -> EpochState           era  )
+  & ( esSnapshots :: EpochState    era -> SnapShots (EraCrypto era) )
+
+data SnapShots c = SnapShots
+  { ssStakeMark          :: SnapShot c  -- Lazy on purpose
+  , ssStakeMarkPoolDistr :: PoolDistr c -- Lazy on purpose
+  , ssStakeSet           :: !(SnapShot c)
+  , ssStakeGo            :: !(SnapShot c)
+  , ssFee                :: !Coin
+  }
+
+data SnapShot c = SnapShot
+  { ssStake       :: !(Stake c)
+  , ssDelegations :: !(VMap VB VB (Credential 'Staking c) (KeyHash 'StakePool c))
+  , ssPoolParams  :: !(VMap VB VB (KeyHash 'StakePool c) (PoolParams c))
+  }
+
+newtype Stake c = Stake
+  { unStake :: VMap VB VP (Credential 'Staking c) (CompactForm Coin)
+  }
+
+newtype PoolDistr c = PoolDistr
+  { unPoolDistr ::
+      Map (KeyHash 'StakePool c) (IndividualPoolStake c)
+  }
+
+data IndividualPoolStake c = IndividualPoolStake
+  { individualPoolStake    :: !Rational
+  , individualPoolStakeVrf :: !(Hash c (VerKeyVRF c))
+  }
+```
+
+As noted by [@JaredCorduan](https://github.com/JaredCorduan), after CIP-1694 is
+complete, there will probably also be a `ssStakeMarkDRepDistr :: Map (KeyHash
+'DRep c) Coin` and a `ssDRepDelegations :: VMap VB VB (Credential 'Staking c)
+(KeyHash 'DRep c)`.
+
+- The reward update
+
+```haskell
+( ne :: NewEpochState era )
+  & ( nesRu :: NewEpochState era -> StrictMaybe (PulsingRewUpdate (EraCrypto era)) )
+  
+data RewardUpdate c = RewardUpdate
+  { deltaT    :: !DeltaCoin
+  , deltaR    :: !DeltaCoin
+  , rs        :: !(Map (Credential 'Staking c) (Set (Reward c)))
+  , deltaF    :: !DeltaCoin
+  , nonMyopic :: !(NonMyopic c)
+  }
+```
+
+Where the `PulsingRewUpdate` is just a mechanism in order to pulse through the
+stake snapshot and in the end yield a `RewardUpdate`.
+
+## Dynamics of these maps
+
+### Unified map
+
+On each application of a block (i.e. the `BBODY` rule) and each tick (i.e. the
+`TICK` rules) we know what entries of the unified map are needed to execute the
+rule. Therefore it fits the current design of UTxO-HD, and therefore can follow
+the _same_ pattern as we currently have for the UTxO set:
+
+- Before calling the ledger rule we can query for the needed entries
+- We can present the ledger with the values they asked for, but restricted to
+  the data available in the disk + changelog
+- The ledger will either provide diffs for the given values or will return
+  updated values that we can then diff with the provided ones
+- The resulting differences can be included in the current definition of the
+  `DbChangelog`.
+  
+In particular, we know that the sets required for the deltas on the unified map
+are small and therefore fit the overall design.
+
+> **_PLAN:_** Consensus will have a new table on the `LedgerTables` that will
+> represent the unified map. Perhaps even 4 tables or a table of triplets. It
+> will have its own place on the `DbChangelog` too. The flow above describes the
+> general strategy for calling the `BBODY` rule, i.e. `(re)applyBlockOpts` and
+> for calling the `TICK` rule, i.e. `applyTickOpts`.
+
+### Stake distribution
+
+The update to this map is performed by `updateStakeDistribution`. It is known
+before calling the `BBODY` rule which UTxOs are going to be deleted and the rule
+execution logic itself knows the UTxOs that are going to be added. Therefore we
+can follow a logic similar to the above:
+
+- Before calling the ledger rule we can query for the needed entries
+- We can present the ledger with the values they asked for, but restricted to
+  the data available in the disk + changelog
+- The ledger will either provide diffs for the given values or will return
+  updated values that we can then diff with the provided ones
+- The resulting differences can be included in the current definition of the
+  `DbChangelog`.
+
+> **_PLAN:_** Consensus will have a new table on the `LedgerTables` that will
+> represent the incremental stake distribution (maybe two tables, one for creds
+> one for ptrs). It will have its own place on the `DbChangelog` too. The flow
+> above describes the general strategy for calling the `BBODY` rule, i.e.
+> `(re)applyBlockOpts`.
+
+### Stake snapshots
+
+The snapshots are rotated by the `snapTransition` rule (called by `TICK`). This
+is the most complicated of the three because it involves accessing the unified
+map and the `IncrementalStake` in their entirety in order to fold them. We do
+these two steps:
+
+```
+step1 = (dom activeDelegs ◁ credStake) ∪ (dom activeDelegs ◁ ptrStake)
+step2 =  aggregate (dom activeDelegs ◁ rewards) step1
+```
+
+Where
+- `activeDelegs` comes from the delegations in the unified map,
+- `rewards` come from the unified map,
+- `credStake` and `ptrStake` come from the incremental stake
+
+However, there is an important note here. The `ssStakeMark` is only used to be
+put in `ssStakeSet` on the next snapshot rotation, then `ssStakeSet` is only
+used to be put in `ssStakeGo`, and `ssStakeGo` is only used to prepare the
+`PulsingRewUpdate`. This is done in `PulsingReward.startStep` and it also uses
+the `rewards` map of the unified map. Assuming the reward calculation is done
+outside of the ledger rules, the snapshot is not really needed by the ledger. In
+that case we would avoid providing the whole map to the Ledger because the
+calculation would be performed outside of the ledger rules.
+
+The `ssStakeMarkPoolDistr` field is used to be put in `nesPd` on the
+`NewEpochState` (by `NEWEPOCH`) which later will be used to provide ledger views
+and calculate leader schedules. Note this is purely a Protocol concern and thus
+probably a Consensus concern.
+
+However it seems that the `ssPoolParams` are in fact modified by the Ledger.
+This should not be very problematic as we would know in andvance which pools are
+updating their params, and we could replicate the schema above for this map.
+
+## Reward computation and the ADA pots
+
+The way the rewards computation happens now (see `PulsingRewards` and
+`Shelley.Rules.Rupd` and `Shelley.Rules.Tick`) is that on each ticking we pulse
+a chunk of the rewards update so that when we reach the epoch boundary we want
+to have pulsed through the whole `ssStakeGo` snapshot that was used when
+creating the pulser.
+
+In the end, the reward computation has to produce a `RewardUpdate`:
+```haskell
+data RewardUpdate c = RewardUpdate
+  { deltaT :: !DeltaCoin
+  , deltaR :: !DeltaCoin
+  , rs :: !(Map (Credential 'Staking c) (Set (Reward c)))
+  , deltaF :: !DeltaCoin
+  , nonMyopic :: !(NonMyopic c)
+  }
+```
+
+where we will use the `delta*` fields to update the treasury, reserves and fee
+pot on the `NEWEPOCH` rule. We also use the `rs` field when calling
+`updateRewards` which happens also on `NEWEPOCH`.
+
+
+If Consensus can compute the `RewardUpdate` (possibly on a separate thread that
+traverses the map at its own pace?) then we can provide the `RewardUpdate` to
+the `NEWEPOCH` rule so that the pots can be updated.
+
+> **_PLAN:_** Consensus will compute the rewards outside of the Ledger and will
+> provide the parts of the `RewardUpdate` to the ledger. In particular, on the
+> epoch boundary, it will provide `deltaT`, `deltaR`, and `deltaF`. Ledger will
+> not compute the `RewardUpdate` thus the pulser becomes dead code.
+
+> Moreover, the ledger will provide a function `f :: Block -> Set
+> (StakeCredential c)` so that Consensus can supply Ledger with only the part of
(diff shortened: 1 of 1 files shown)


ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+  & ( esLState  :: EpochState         era  -> LedgerState        era  )
+  & ( lsDPState :: LedgerState        era  -> DPState (EraCrypto era) )
+  & ( dpsDState :: DPState (EraCrypto era) -> DState  (EraCrypto era) )
+  & ( dsUnified :: DState  (EraCrypto era) -> UMap    (EraCrypto era) )
+   
+data UMap c = UMap !(Map (Credential 'Staking c) (Trip c)) !(Map Ptr (Credential 'Staking c)) 
+```
+
+> **_NOTE:_** The `deposits` don't seem to be in-place yet? For me it still is a
+> TriMap
```
JaredCorduan, line 32: it is there in the `master` branch now, maybe you have a older version?

Jasagredo: That would explain it, yeah :smile:

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+
+- The stake distribution per stake credential
```
JaredCorduan, line 34:
```suggestion
- The current stake distribution per stake credential
```

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+- The stake snapshots
+
+```haskell
+( ne :: NewEpochState era )
+  & ( nesEs       :: NewEpochState era -> EpochState           era  )
+  & ( esSnapshots :: EpochState    era -> SnapShots (EraCrypto era) )
+  
+data SnapShots c = SnapShots
+  { ssStakeMark          :: SnapShot c -- Lazy on purpose
+  , ssStakeMarkPoolDistr :: PoolDistr c -- Lazy on purpose
```
JaredCorduan, line 58: Note that after CIP-1694 is complete, we will probably also have a `ssStakeMarkDRepDistr` with a similar type, `Map (KeyHash 'DRep c) Coin`.

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+  , ssStakeSet           :: !(SnapShot c)
+  , ssStakeGo            :: !(SnapShot c)
+  , ssFee                :: !Coin
+  }  
+  
+data SnapShot c = SnapShot
+  { ssStake       :: !(Stake c)
+  , ssDelegations :: !(VMap VB VB (Credential 'Staking c) (KeyHash 'StakePool c))
```
JaredCorduan, line 66: Note that after CIP-1694 is complete, we will probably also have a `ssDRepDelegations` with a similar type, `VMap VB VB (Credential 'Staking c) (KeyHash 'DRep c)`.

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+      Map (KeyHash 'StakePool c) (IndividualPoolStake c)
+  }  
+  
+data IndividualPoolStake c = IndividualPoolStake
+  { individualPoolStake    :: !Rational
+  , individualPoolStakeVrf :: !(Hash c (VerKeyVRF c))
+  }  
+```
+
+> **_TODO:_** All of the maps?
```
JaredCorduan, line 85: We probably also want the reward update:

https://github.com/input-output-hk/cardano-ledger/blob/3c4835dc4fe80a19f5714f576c49926d6a58c563/eras/shelley/impl/src/Cardano/Ledger/Shelley/RewardUpdate.hs#L91

the "lens" for it is `NewEpochState -> Maybe RewardUpdate`.

JaredCorduan (commented):
overall, this looks really excellent, thank you @Jasagredo ! I made a few small comments.

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+
+
+If Consensus can compute the `RewardUpdate` (possibly on a separate thread that
+traverses the map at its own pace?) then we can provide the `RewardUpdate` to
+the `NEWEPOCH` rule so that the pots can be updated.
+
+> **_PLAN:_** Consensus will compute the rewards outside of the Ledger and will
+> provide the `RewardUpdate` to the ledger so that it can modify the pots and
+> also produce diffs for the unified map's `Rewards` field. Ledger will not
+> compute the `RewardUpdate` thus the pulser becomes dead code.
```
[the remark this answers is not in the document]
Jasagredo: I think I misworded the sentence. I was meaning:

> Consensus will compute the rewards outside of the Ledger and will
> provide the `RewardUpdate` to the Ledger so that the Ledger can modify the pots and
> also the Ledger can produce diffs for the unified map's `Rewards` field based on the withdrawals happening in this block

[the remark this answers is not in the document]
JaredCorduan: Sorry if I am being overly picky! This doesn't sound right to me:

> [Consensus] will provide the RewardUpdate to the Ledger

and also this:

> the Ledger can produce diffs for the unified map's Rewards field

---

How about something like this:

> **_PLAN:_** Consensus will compute the rewards outside of the Ledger and will
> provide the _parts_ of the `RewardUpdate` to the ledger. In particular, on the epoch
> boundary, it will provide `deltaT`, `deltaR`, and `deltaF`.
> Ledger will not compute the `RewardUpdate` thus the pulser becomes dead code.
>
> Moreover, the ledger will provide a function `f :: Block -> Set (StakeCredential c)`
> so that Consensus can supply Ledger with only the part of the `UMap` (the unified map)
> that it requires. In particular, Consensus will have to apply the
> `rs :: !(Map (Credential 'Staking c) (Set (Reward c)))` field of the reward update
> to the unified map on the epoch boundary, _prior_ to computing this view for the ledger
> (to ensure that reward withdrawals are correct within the given block).

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+- Before calling the ledger rule we can query for the needed entries
+- We can present the ledger with the values they asked for, but restricted to
+  the data available in the disk + changelog
+- The ledger will either provide diffs for the given values or will return
+  updated values that we can then diff with the provided ones
+- The resulting differences can be included in the current definition of the
+  `DbChangelog`.
+  
+> **_PLAN:_** Consensus will have a new table on the `LedgerTables` that will
+> represent the Incremental stake distribution (maybe two tables, one for creds
```
dnadales, line 125:
```suggestion
> represent the incremental stake distribution (maybe two tables, one for creds
```

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+> one for ptrs). It will have its own place on the `DbChangelog` too. The flow
+> above describes the general strategy for calling the `BBODY` rule, i.e.
+> `(re)applyBlockOpts`.
+
+### Stake snapshots
+
+The snapshots are rotated by the `snapTransition` rule. This is the most
```
dnadales, line 132: Maybe it's worth mentioning where this rule is called (I think by the `TICK` rule, right?)

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+- `activeDelegs` comes from the delegations in the unified map,
+- `rewards` come from the unified map,
+- `credStake` and `ptrStake` come from the incremental stake
+
+However, there is an important note here. The `ssStakeMark` is only used to be
+put in `ssStakeSet` on the next snapshot rotation, then `ssStakeSet` is only
+used to be put in `ssStakeGo`, and `ssStakeGo` is only used to prepare the
+`PulsingRewUpdate`. This is done in `PulsingReward.startStep` and it also uses
+the `rewards` map of the unified map. Assuming the reward calculation is done
+outside of the ledger rules, the snapshot is not really needed by the ledger.
```
dnadales, line 152: From the preceding discussion, it is unclear how we avoid querying the entire delegations map? Is the fact that `activeDelegs` is a small set the key here?

dnadales: Or is the fact that we're doing the calculation outside the ledger rules the key piece of information here? If so we should probably underline the fact that Ledger HD will require a Consensus-Ledger bridge, and hint at where we think this bridge ought to live (ie Consensus or Ledger).

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+
+> **_PLAN:_** Consensus will manage the snapshots to produce stake distribution
+> by pool that can be used by Consensus later to resolve queries about the
+> LeaderSchedule. Ledger will not know about the Snapshots. In particular, the
+> UTxO-HD report includes the concept of Snapshots of tables, which would be
+> used to manage and access these snapshots.
+
+## Open questions
+
+- Should Ledger return the diffs? they actually do it, but they then apply the
```
dnadales, line 214:
```suggestion
- Should the Ledger return the diffs? It actually does it, but it then applies the
```

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+  diffs. If we wanted to return then, there are many intermediate layers through
```
dnadales, line 215:
```suggestion
  diffs. If we wanted to return them, there are many intermediate layers through
```

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+
+> **_TODO:_** All of the maps?
+
+## Dynamics of these maps
+
+### Unified map
+
+On each application of a block (i.e. the `BBODY` rule) and each tick (i.e. the
+`TICK` rules) it can be known what entries of this map are needed to carry the
+execution. Therefore it fits the current design of UTxO-HD.
```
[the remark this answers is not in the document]
Jasagredo: Yes for this map, and the "big" ones we were considering will not be given to the ledger as you correctly inferred below

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,219 @@
...
+On each application of a block (i.e. the `BBODY` rule) and each tick (i.e. the
+`TICK` rules) it can be known what entries of this map are needed to carry the
+execution. Therefore it fits the current design of UTxO-HD. 
+
+- Before calling the ledger rule we can query for the needed entries
+- We can present the ledger with the values they asked for, but restricted to
+  the data available in the disk + changelog
+- The ledger will either provide diffs for the given values or will return
+  updated values that we can then diff with the provided ones
+- The resulting differences can be included in the current definition of the
```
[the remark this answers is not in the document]
Jasagredo: Nope, clarified

ouroboros-consensus/docs/future-ledger-hd.md
Jasagredo: It is the latter, I have clarified it a bit at the bottom.

Jasagredo:
Thanks for your comments @JaredCorduan and @dnadales. I think I have addressed all of them. Also have in mind that once this document is good enough to be considered "a rough sketch of a plan" it should be good to go, we shouldn't invest much time on it.

JaredCorduan (approved):
looks great, thank you @Jasagredo !

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,257 @@
...
+      Map (KeyHash 'StakePool c) (IndividualPoolStake c)
+  }
+
+data IndividualPoolStake c = IndividualPoolStake
+  { individualPoolStake    :: !Rational
+  , individualPoolStakeVrf :: !(Hash c (VerKeyVRF c))
+  }
+```
+
+As noted by @jared.corduan, after CIP-1694 is complete, there will probably also
```
dnadales, line 83:
```suggestion
As noted by [@JaredCorduan](https://github.com/JaredCorduan), after CIP-1694 is complete, there will probably also
```

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,258 @@
...
+- The ledger will either provide diffs for the given values or will return
+  updated values that we can then diff with the provided ones
+- The resulting differences can be included in the current definition of the
+  `DbChangelog`.
+
+> **_PLAN:_** Consensus will have a new table on the `LedgerTables` that will
+> represent the incremental stake distribution (maybe two tables, one for creds
+> one for ptrs). It will have its own place on the `DbChangelog` too. The flow
+> above describes the general strategy for calling the `BBODY` rule, i.e.
+> `(re)applyBlockOpts`.
```
dcoutts, line 151: The incrementally maintained stake distribution should be a table where we normally only need to do monoidal updates. This means we turn a read/write workload into a write only workload. This is covered in the original report and is the motivation for the inclusion of monoidal update tables (including in the prototype).

This is important for performance, otherwise we are going to be doing almost as many lookups in the stake distribution table as we do lookups in the utxo table. Lookups are the bottleneck for a write-optimised disk backend.

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,258 @@
...
+> **_PLAN:_** Consensus will manage the snapshots to produce stake distribution
+> by pool that can be used by Consensus later to resolve queries about the
+> LeaderSchedule. Ledger will not know about the Snapshots. In particular, the
+> UTxO-HD report includes the concept of Snapshots of tables, which would be
+> used to manage and access these snapshots.
+
+Note that this implies creating a new package or component at the
+Consensus-Ledger boundary whose owner would probably be the Consensus team as
+its responsibilities would be related with computations required for the
+Consensus protocol (leader checks, and similar).
```
dcoutts, line 248: I'm slightly unclear about what this section is porposing. The stake distribution by pool is not large, so will not be an on-disk table. It will remain as an in-memory structure as it is now, within the ledger state and ledger view. I don't see the need for any consensus/ledger interface changes.

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Open questions
+
+- Should the Ledger return the diffs? it actually internally compute diffs, but
+  they the diffs are applied to the values before returning. If we wanted to
+  return the diffs instead, there are many intermediate layers through which
+  they will have to be floated, but it should be doable.
```
dcoutts, line 255: If this refers to what I think it does, it's covered in the second report and in the prototype. The solution is to paramaterise the ledger rules by the table type, and then to use a "partial tracking map" which fits the map interface and also tracks the diffs. This gives us the diffs we want without any "real" changes in the ledger rules (just abstracting over the map interface for the on-disk tables).

dcoutts (commented):
A few comments below.

One thing that is not clear from this doc, but needs to become clear is what precisely are the new on-disk tables, and how do they correspond to the existing ledger state.

In the style of the prototype that would be done by paramaterising the ledger state types by the on-disk table types and using them in the appropriate places (at the appropriate types). That would make it precise.

ouroboros-consensus/docs/future-ledger-hd.md
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+newtype Stake c = Stake
+  { unStake :: VMap VB VP (Credential 'Staking c) (CompactForm Coin)
+  }
+
+newtype PoolDistr c = PoolDistr
+  { unPoolDistr ::
+      Map (KeyHash 'StakePool c) (IndividualPoolStake c)
+  }
```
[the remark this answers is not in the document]
JaredCorduan: Here's the funny thing, though, @dcoutts : once the large structures are moved to a database, ledger doesn't even need the per-pool stake distribution! it is _only_ used by consensus for the leader check, and it entirely derived from the other large data structures.

ouroboros-consensus/docs/future-ledger-hd.md
JaredCorduan: Are you suggesting that we do something more like double entry booking, where we (monoidally) send updates to two separate things, additions and deletions, and only combine them when needed?

ouroboros-consensus/docs/future-ledger-hd.md
JaredCorduan: I commented on this above.

ouroboros-consensus/docs/future-ledger-hd.md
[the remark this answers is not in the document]
dcoutts: Ok, but that doesn't obviously mean that it should be moved across the ledger/consensus boundary. I'm not saying it shouldn't but we should be clear about the reasons.

The way I would look at this is that the on-disk tables still belong to the ledger. The consensus provides the ledger with the facility to have on-disk tables (just as it provides state persistence, rollbacks etc). So something being in an on-disk table doesn't itself imply it has moved across the consensus/ledger boundary. It's like the consensus layer is the interpreter and the ledger is the logic or program being interpreted. The consensus layer shouldn't need to know about the content or purpose of the tables, it should just manage them as instructed by the ledger.

ouroboros-consensus/docs/future-ledger-hd.md
dcoutts: I'm not sure about two separate things (what two?) but yes I am suggesting a table that can support monoidal updates, as well as insert/delete/lookup.

So for an incrementally updated stake distribution, that maps stake address to lovelace, it would be updated by the ledger working out the delta in stake arising for a stake address from a tx, and that is the table update. So in particular, the lovelace value of a spent input is a negative delta for that input's stake address, while lovelace outputs contribute positively to the balance of their corresponding stake addresses. Note that no lookup is needed: the sum total of the lovelace for a stake address is not needed.

And the practical consequence is the update-only workload is a write-only workload, which is much cheaper than a read/write workload.

ouroboros-consensus/docs/future-ledger-hd.md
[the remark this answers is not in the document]
JaredCorduan: that's totally reasonable, but we are in the ~gray~ grey zone here for the consensus/ledger split. The "ledger protocol" logic/rules were moved to the consensus repo at the vasil hard fork. and all the per-pool stake distribution really is is an aggregation of the table that we know is not up for debate.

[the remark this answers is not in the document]
dcoutts: For this particular case, of the "protocol" rules needing access to the stake distribution, I think that does exactly fall either side of the boundary.

At a high level we would say that the Praos protocol (which is part of consensus) depends on the ledger for access to the stake distribution (or some suitable notion of a stake distribution). So maintaining that stake distribution (updating it as transactions are validated) should be the responsibility of the ledger, and then it needs to provide it to the Praos layer when needed.

So I think in this case, where one would logically draw the boundary at the high level tells us fairly clearly where we should draw the boundary here.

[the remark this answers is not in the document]
JaredCorduan: Let me try this from another angle. The ledger should never be given the entire per-stake-credential stake distribution (henceforth the PSCSD :grin: ), since that violates the rule that the ledger never hold "large" structures in memory. Instead, the ledger should give consensus instructions for what to do.

The per-pool stake distribution (PPSD) is small, and so the ledger is fine to hold it. But how does it the ledger give consensus instructions on how to make it? The instructions are the delegation map, which could be provided incrementally (each delegation is a "delta", as is each pool retirement"). And this is already a part of the plan, we know that the delegation relation will be a table (probably as a part of the "unified map").

So I can see two ways to get the PPSD to the ledger:
1. it maintains it's own incremental computation, using the txins, txouts, delegations, and retirements. and hopes that this always stays in since with the PSCSD maintained by consensus using the tables and the deltas provided by ledger.
2. consensus uses the PSCSD and the delegation relation to maintain another table for the small PPSD.

And just remember, that the PPSD is not used by any code in the ledger repo any more, it is only used for the leader check. It is not even a member of what consensus deems "the ledger state".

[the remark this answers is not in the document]
dcoutts: Yes, the reasonable designs to consider are continuously incrementally maintaining the by-pool distribution or doing a bulk computation over some phase of the epoch (implemented as an incremental traversal of the snapshot of the previous epoch by-stake-cred distribution).

Continuously incrementally maintaining the by-pool distribution sounds preferable, but also sounds tricky in the presence of fractional delegation. But if you think we can do it, then it means it's all in memory which is a lot easier.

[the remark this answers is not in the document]
JaredCorduan: > And just remember, that the PPSD is not used by any code in the ledger repo any more

@dcoutts as soon as I make a bold claim things begin too change. :laughing:  SPOs will get a stake-based vote in the Voltaire phase.

PR #4301, opened 2023-01-23
UTxO-HD: Reenable mempool forge and benchmark ledger ops

Jasagredo:
# Description

Apart from reenabling the analysis, we are also plotting the time it takes to read and flush the values from the BackingStore.

### DIFF
--- a/ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
+++ b/ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
@@ -558,7 +558,7 @@ traceLedgerProcessing
 
 -------------------------------------------------------------------------------}
 benchmarkLedgerOps ::
-  forall blk.
+     forall blk.
      HasAnalysis blk
   => Maybe FilePath -> Analysis blk
 benchmarkLedgerOps mOutfile AnalysisEnv {db, initLedger, cfg, limit, backing, ledgerDbFS} = do
@@ -609,7 +609,7 @@ benchmarkLedgerOps mOutfile AnalysisEnv {db, initLedger, cfg, limit, backing, le
               tNow <- GC.mutator_elapsed_ns <$> GC.getRTSStats
               pure (r, tNow - tPrev)
 
-        let slot = blockSlot      blk
+        let slot = blockSlot blk
         ((ldb', sinceLast'), tFlush) <- time $
           if sinceLast == 100
           then do
@@ -618,21 +618,26 @@ benchmarkLedgerOps mOutfile AnalysisEnv {db, initLedger, cfg, limit, backing, le
             Ouroboros.Consensus.Storage.LedgerDB.OnDisk.flush backingStore toFlush
             pure (toKeep, 0)
           else pure (ldb, sinceLast + 1)
+
         let prevLedgerState = ledgerDbCurrent ldb'
+        let aks             = rewindTableKeySets (ledgerDbChangelog ldb') (getBlockKeySets blk)
+
         -- We do not use strictness annotation on the resulting tuples since
         -- 'time' takes care of forcing the evaluation of its argument's result.
         (tkLdgrView, tForecast) <- time $ forecast            slot prevLedgerState
         (tkHdrSt,    tHdrTick)  <- time $ tickTheHeaderState  slot prevLedgerState tkLdgrView
         (hdrSt',     tHdrApp)   <- time $ applyTheHeader                           tkLdgrView tkHdrSt
-        let ks = getBlockKeySets blk
-        let aks = rewindTableKeySets (ledgerDbChangelog ldb') ks
-        (urs, tReadDb) <- time $ defaultReadKeySets (readKeySets backingStore) $ readDb aks
+        (urs,        tReadDb)   <- time $ defaultReadKeySets (readKeySets backingStore) $ readDb aks
+
         let fwd = case forwardTableKeySets (ledgerDbChangelog ldb') urs of
               Left e -> error $ "Couldn't forward " ++ show e
               Right v -> v
         let l = ledgerState prevLedgerState `withLedgerTables` unExtLedgerStateTables fwd
+
         (tkLdgrSt,   tBlkTick)  <- time $ tickTheLedgerState  slot (forgetLedgerTables l)
+
         let l' = applyLedgerTablesDiffsTicked l tkLdgrSt
+
         (ldgrSt',    tBlkApp)   <- time $ applyTheBlock                                       l'
 
         let ldgrSt'' = prependLedgerTablesDiffsFromTicked tkLdgrSt ldgrSt'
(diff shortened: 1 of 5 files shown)


ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
...
+          -- result.
+          time act = do
+              tPrev <- GC.mutator_elapsed_ns <$> GC.getRTSStats
+              !r <- act
+              tNow <- GC.mutator_elapsed_ns <$> GC.getRTSStats
+              pure (r, tNow - tPrev)
+
+        let slot = blockSlot      blk
+        ((ldb', sinceLast'), tFlush) <- time $
+          if sinceLast == 100
```
jorisdral, line 614: We should record `100` somewhere as a constant in the `OnDisk` module, and let the actual `flush` function and this local definition refer to it. Otherwise the implementation and this analysis are likely to get out of sync if we change the flush frequency

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
...
+          -- action. This function forces the evaluation of its argument's
+          -- result.
+          time act = do
+              tPrev <- GC.mutator_elapsed_ns <$> GC.getRTSStats
+              !r <- act
+              tNow <- GC.mutator_elapsed_ns <$> GC.getRTSStats
+              pure (r, tNow - tPrev)
+
+        let slot = blockSlot      blk
+        ((ldb', sinceLast'), tFlush) <- time $
```
jorisdral, line 613: In the `OnDisk` module, it looks like the flush check is made after applying a block, instead of before it. It's probably equivalent, but the flushing time could be attributed to a slot/block that is off-by-one compared to actual flush slot/block

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
...
+          -- recorded 'elapsedTime' till the end of the execution of the given
+          -- action. This function forces the evaluation of its argument's
+          -- result.
+          time act = do
+              tPrev <- GC.mutator_elapsed_ns <$> GC.getRTSStats
+              !r <- act
+              tNow <- GC.mutator_elapsed_ns <$> GC.getRTSStats
+              pure (r, tNow - tPrev)
+
+        let slot = blockSlot      blk
```
jorisdral, line 612:
```suggestion
        let slot = blockSlot blk
```

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
...
+          else pure (ldb, sinceLast + 1)
+        let prevLedgerState = ledgerDbCurrent ldb'
+        -- We do not use strictness annotation on the resulting tuples since
+        -- 'time' takes care of forcing the evaluation of its argument's result.
+        (tkLdgrView, tForecast) <- time $ forecast            slot prevLedgerState
+        (tkHdrSt,    tHdrTick)  <- time $ tickTheHeaderState  slot prevLedgerState tkLdgrView
+        (hdrSt',     tHdrApp)   <- time $ applyTheHeader                           tkLdgrView tkHdrSt
+        let ks = getBlockKeySets blk
+        let aks = rewindTableKeySets (ledgerDbChangelog ldb') ks
+        (urs, tReadDb) <- time $ defaultReadKeySets (readKeySets backingStore) $ readDb aks
```
jorisdral, line 629: Align with the other `(,) <- time` patterns. Possibly also align the `=` in let-bindings with the `<-` arrows? Might make the code slightly easier to read, though I haven't tried to see what it looks like

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
...
+        let
+          -- Compute how many nanoseconds the mutator used from the last
+          -- recorded 'elapsedTime' till the end of the execution of the given
+          -- action. This function forces the evaluation of its argument's
+          -- result.
+          time act = do
+              tPrev <- GC.mutator_elapsed_ns <$> GC.getRTSStats
+              !r <- act
+              tNow <- GC.mutator_elapsed_ns <$> GC.getRTSStats
+              pure (r, tNow - tPrev)
```
jorisdral, line 610: I can imagine that in this function, lazy evaluation could warp the measured times. Should we verify/check that with the changes for UTxO HD, we are not accidentally inflating/deflating the measures, for example by not evaluating subcomputation of an `act` enough before passing it to the `time` function?

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
 -------------------------------------------------------------------------------}
 benchmarkLedgerOps ::
   forall blk.
```
jorisdral, line 561:
```suggestion
     forall blk.
```

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
...
+      1 -> pure ReproMempoolForgeOneBlk
+      2 -> pure ReproMempoolForgeTwoBlks
+      _ -> fail $ "--repro-mempool-and-forge only supports"
+               <> "1 or 2 blocks at a time, not " <> show numBlks
+
+    initLedger' <- initialiseLedger ledgerDbFS cfg initLedger
+
+    (ldb, bs) <- ledgerDbAndBackingStore backing ledgerDbFS $ case initLedger of
+                                                                Left s -> Left (initLedger', s)
+                                                                Right v -> Right v
```
jorisdral, line 745: The `case` looks like a common pattern, could be abstracted

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
...
+      , ledgerDbFS
+    } = env
+
+    lCfg :: LedgerConfig blk
+    lCfg = configLedger cfg
+
+    elCfg :: LedgerCfg (ExtLedgerState blk)
+    elCfg = ExtLedgerCfg cfg
+
+    timed :: IO a -> IO (a, IOLike.DiffTime)
```
jorisdral, line 794: Similar comment about lazy evaluation possibly muddling measured time

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
```diff
@@ -574,251 +559,327 @@ traceLedgerProcessing
...
+      -> IO (LedgerDB' blk, Word64)
+    process bs howManyBlocks mempool mbBlk blk' (ldb, sinceLast) = do
+      (ldb', sinceLast') <-
+          if sinceLast == 100
+          then do
+            let (toFlush, toKeep) =
+                  ledgerDbFlush DbChangelogFlushAllImmutable ldb
+            Ouroboros.Consensus.Storage.LedgerDB.OnDisk.flush bs toFlush
+            pure (toKeep, 0)
+          else pure (ldb, sinceLast + 1)
```
jorisdral, line 817: Similar comment as before, should this happen before or after applying txs?

jorisdral (changes requested):
Looking good, I've seen some `benchmarkLedgerOps` graphs as well, and the read/flush operations are a good addition!

dnadales (approved):

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
Jasagredo: Referenced in new issue

ouroboros-consensus-cardano-tools/src/Cardano/Tools/DBAnalyser/Analysis.hs
Jasagredo: I think this should be fine taken with some caution. This is just an indication of the overall performance of the system, so I don't think it is that crucial that it mimicks the behavior of the node?

jorisdral (approved):

PR #4313, opened 2023-01-25
WIP: possible improvement in ChainSync client?

nfrisby:
This is a single-commit PR; see its message.

### DIFF
--- a/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
+++ b/ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
@@ -565,7 +565,7 @@ chainSyncClient mkPipelineDecision0 tracer cfg
           -- ('rollBackward'), so we have nothing to do.
           return $ Just kis
 
-        | Just intersection <- AF.intersectionPoint ourFrag' theirFrag ->
+        | Just (ourPre, _ourSuf, _theirPre, theirSuf) <- AF.intersect ourFrag' theirFrag ->
           -- Our current chain changed, but it still intersects with candidate
           -- fragment, so update the 'ourFrag' field and trim to the
           -- candidate fragment to the same anchor point.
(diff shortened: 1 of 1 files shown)


nfrisby:
Hmm. Actually: does "average-case = worst-case" apply here, meaning that we _don't_ want to share the selection's prefix amongst all the ChainSync clients' `theirFrag` prefixes?

ouroboros-consensus/src/Ouroboros/Consensus/MiniProtocol/ChainSync/Client.hs
```diff
@@ -565,7 +565,7 @@ chainSyncClient mkPipelineDecision0 tracer cfg
           -- ('rollBackward'), so we have nothing to do.
           return $ Just kis
 
-        | Just intersection <- AF.intersectionPoint ourFrag' theirFrag ->
+        | Just (ourPre, _ourSuf, _theirPre, theirSuf) <- AF.intersect ourFrag' theirFrag ->
```
amesgen, line 568: Was just confused when looking at the docs; makes no semantic difference of course:
```suggestion
        | Just (ourPre, _theirPre, _ourSuf, theirSuf) <- AF.intersect ourFrag' theirFrag ->
```

nfrisby: Fixed, thanks.

amesgen (commented):
I think this is a nice change, I could even see defining an abstract function like
```haskell
cross ::
     forall blk. HasHeader blk
  => AnchoredFragment blk
  -> AnchoredFragment blk
  -> Maybe (Point blk, AnchoredFragment blk)
cross c1 c2 = do
    (p1, _p2, _s1, s2) <- AF.intersect c1 c2
    let crossed = fromMaybe (error "impossible") $ AF.join p1 s2
    pure (AF.anchorPoint s2, crossed)
```
to make it more clear that we don't rely on any domain-specific properties of the ChainSync fragments here. Of course, we should keep the fairbairn threshold in mind.

W.r.t. performance, this new implementation should have the same worst-case performance; and I don't immediately see how it might have a better average-case performance (`splitAfterPoint` and `join` both perform sharing with previous fragments). In any case, it should be easy to write a small benchmark (and if we use the `cross` function above, it is easy to swap out the implementation if we intentionally want to make the average case slower again).

nfrisby:
> `splitAfterPoint` and `join` both perform sharing with previous fragments

`cross` causes all clients to share the selection's prefix. `join` causes each client to reuse its own copy of that prefix. So `splitAfterPoint` causes the clients to share less, right?

amesgen:
Superseded by https://github.com/input-output-hk/ouroboros-consensus/pull/222

PR #4345, opened 2023-02-07
Fix line split in CI workflow file

Jasagredo:


.github/workflows/build.yml
```diff
@@ -103,7 +103,8 @@ jobs:
 
     - name: Record dependencies
       id: record-deps
-      run: cabal build all --dry-run
+      run: |
+        cabal build all --dry-run
```
amesgen, line 107: Why did this not make CI fail?

https://github.com/input-output-hk/ouroboros-network/actions/runs/4084241519/jobs/7040712757#step:11:13

amesgen (approved):
Thanks, I guess you did note this just as a side effect, or does this actually fix an issue?

Might be a nice addition to https://noyaml.com

Jasagredo:
> Thanks, I guess you did note this just as a side effect, or does this actually fix an issue?
> 
> 
> 
> Might be a nice addition to https://noyaml.com

I actually found it while thinking what could be causing that CI error on Arnaud's branch, and wanting to check the dependencies file if possible 😄

.github/workflows/build.yml
Jasagredo: Tbh I have no idea. Perhaps Cabal doesn't end with an exit code that is recognized as failing in this case?

Jasagredo:
bors retry

PR #4348, opened 2023-02-08
Add concrete steps for cutting a release

Jasagredo:


dnadales (approved):
Thanks! This is a great start! We can improve this later (eg by linking to the CHaP release process). The only question I have is whether we should either link to the new section at the beginning of the doc, or put this section higher up.

PR #4361, opened 2023-02-13
Enforce changelog check in Bors

Jasagredo:
# Description

Enforce the Consensus changelog consistency check and the cabal files check in Bors.


amesgen (approved):
Let's see how Bors likes it :+1:

PR #4364, opened 2023-02-14
Bump haskell.nix

andreabedini:
The bump is to support cabal file revisions on CHaP which are necessary to get our packages to build without all those manual constraints we have in cabal.project.

I am opening a draft PR to test the waters and warm up the chaces ahead of time since it's relatively big change in haskell.nix. I'll make sure everything works and then mark it as ready.


angerman:
bors try

andreabedini:
Subsumed by #4349 4349

PR #4365, opened 2023-02-14
Bump version of monoidal-synchronisation

coot:


monoidal-synchronisation/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Revision history for monoid-synchronisation
```
Jasagredo, line 1: The name is wrong here
```suggestion
# Revision history for monoidal-synchronisation
```

Jasagredo (approved):

coot:
ouch, I forgot to use `bors` :man_facepalming:

coot:
The problem I have with it, is that in some repos we use it in others (`io-sim`, `typed-protocols`, ...) we don't and it got out of my habit.

PR #4375, opened 2023-02-16
contributing guide

coot:
Add important coding remarks section.  It now just includes a single
our `timeout` & friends functions which are using SI units.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -177,6 +177,22 @@ almost all of our code is written in polymorphic way using `io-classes` which
 comes with [`io-sim`]: `io-classes` expose a very similar API that the `base`,
 `async`, `stm` and `time` packages provides.
 
+### Important Coding Remarks
+
+* Please note that across this code base we are using a custom time functions
+  which are measured in seconds not nano or microseconds as similarly named
+  functions in `base` do.  This applies to:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -177,6 +177,22 @@ almost all of our code is written in polymorphic way using `io-classes` which
 comes with [`io-sim`]: `io-classes` expose a very similar API that the `base`,
 `async`, `stm` and `time` packages provides.
 
+### Important Coding Remarks
+
+* Please note that across this code base we are using a custom time functions
+  which are measured in seconds not nano or microseconds as similarly named
+  functions in `base` do.  This applies to:
+  - `timeout`
```
dnadales, line 185: Would it be too lengthy or disruptive to rename these functions to things like `timeoutInSeconds`?  Or have these functions take a type that requires a `Seconds` constructor.

coot: They are imported from a different module which indicates we are using SI units.  Using newtype wrapper would be rather annoying, but would guarantee correctness.  Something like `timeoutSec` or `timeoutSI` could be a good idea, or maybe we should just always import the module qualified and use it as `SI.timeout`.

coot:
bors merge

Issue #4151, opened 2022-11-16
Groom Issue 4150, VRF and KES


nfrisby:
This task is for @nfrisby to prototype the Issue #4150 integration, in order to refine our estimation of how much work that Issue will actually require.

nfrisby:
@yogeshsajanikar Might do this instead of me. So I'll sketch my plan here.

- Determine which branches of the upstream repos (eg `cardano-base`) we should use to gain access to the new VRF and KES types & impls.

- Do the work of simply getting an `ouroboros-network` commit to build with those dependencies (possibly no effort required here 🤞).

There are then two decisions about how to proceed.

- {temporary decision} Should this 4151 work be done on the Conway draft branch? The new crypto will actually be used by Conway. But it might be simpler to do this POC on Babbage on our `master` branch eg. For the purpose of this Issue, it shouldn't matter. And I think the end result will be flexible enough that it'd be easy to change which era uses which crypto post hoc.

- {longterm design decision} How to express multiple crypto types in the sequence of blocks types passed to `HardForkBlock` when defining `CardanoEras`?

Currently:

```
type CardanoBlock c = HardForkBlock (CardanoEras c)

type CardanoEras c =
  '[ ByronBlock
   , ShelleyBlock (TPraos c) (ShelleyEra c)
   , ShelleyBlock (TPraos c) (AllegraEra c)
   , ShelleyBlock (TPraos c) (MaryEra    c)
   , ShelleyBlock (TPraos c) (AlonzoEra  c)
   , ShelleyBlock (Praos  c) (BabbageEra c)
   ]
```

What we ultimately need is one of the following instead:


```
type CardanoErasX c1 c2 =
  '[ ByronBlock
   , ShelleyBlock (TPraos c1) (ShelleyEra c1)
   , ShelleyBlock (TPraos c1) (AllegraEra c1)
   , ShelleyBlock (TPraos c1) (MaryEra    c1)
   , ShelleyBlock (TPraos c1) (AlonzoEra  c1)
   , ShelleyBlock (Praos  c1) (BabbageEra c1)
   , ShelleyBlock (Praos  c2) (ConwayEra  c2)
   ]
```

or

```
type family PrevCrypto (era :: Type -> Type) (cs :: Type) :: c

type CardanoErasY cs =
  '[ ByronBlock
   , ShelleyBlock (TPraos (PrjCrypto ShelleyEra cs)) (ShelleyEra (PrjCrypto ShelleyEra cs))
[...]

Issue #4150, opened 2022-11-16
Integrate new VRF and KES crypto versions into the Conway Era


nfrisby:
The crypto team has (re-)prepared newer versions of the VRF and KES algorithms/formats for use in eras after Babbage (ie starting with Conway after the Chang HF).

This Issue tracks the Consensus team's work to integrate those changes into the Conway branch. I think the main challenge will be choosing how to increase the expressiveness of the `CardanoBlock` type to have varying crypto `c` arguments to the `ShelleyBlocks` within the list of eras argument to `HardForkBlock`. Byron through Babbage will use `c1` and Conway will use `c2`. It should be just a bit more bookkeeping and maybe a few careful placed `~` constraints, but that's all.

For reference, this is a previous, backed out PR that might be relevant, or point to relevant upstream commits: PR #3801

nfrisby:
For reference, this is a previous, backed out PR that might be relevant, or point to relevant upstream commits: PR #3801

nfrisby:
To clarify, the Crypto team has three PI objectives that will involve Consensus:

> - KES secure forgetting
> - VRF batch-compatibility (requires HF)
> - KES signature size reduction (requires HF)

If I understand correctly, this Issue pertains only to the second to; it does not involve _secure forgetting_.

abailly-iohk:
From @nfrisby 
> This Issue tracks the Consensus team's work to integrate those changes into the Conway branch. I think the main challenge will be choosing how to increase the expressiveness of the `CardanoBlock` type to have varying crypto `c` arguments to the `ShelleyBlocks` within the list of eras argument to `HardForkBlock`. Byron through Babbage will use `c1` and Conway will use `c2`. It should be just a bit more bookkeeping and maybe a few careful placed `~` constraints, but that's all.

abailly-iohk:
I tried latest solution proposed by @yogeshsajanikar in #4151 and it does not work nicely. While I was able to modify `O.C.C.Block` easily to introduce the type families providing an additional level of indirection to resolve the actual crypto used, I got bitten by this error in the `O.C.C.CanHardFork` module:

```
src/Ouroboros/Consensus/Cardano/CanHardFork.hs:308:42: error:
    • Illegal type synonym family application ‘StdCrypto
                                                 c’ in instance:
        CanHardFork (CardanoEras c)
    • In the instance declaration for ‘CanHardFork (CardanoEras c)’
    |
[...]

PR #4386, opened 2023-02-22
Introduces property checking the size of Header's VRF Certs for different Eras

abailly-iohk:
# Description

This PR does 2 things:
* An attempt at minimising the "attack surface" of consensus' dependencies on cardano-ledger crypto stuff related to Praos/TPraos, 
* Introduction of Property to check dispatching of crypto depending on era actually produces the right results.

## Hide Ledger's `PraosCrypto`

The idea is to remove direct dependencies from consensus package to cardano-ledger's `TPraos.API` module. There is a duplication of the `PraosCrypto` typeclass between the two components which does seem unnecessary as ultimately, these crypto primitives should end up in the consensus where they are used and relevant. 

I would have loved to be able to remove the ledger's `PraosCrypto` altogether but of course this is not possible as we use quite a few functions from the ledger that depend on it. So the tiny first step consists in hiding the dependency on `Ledger.PraosCrypto` behind the `Ouroboros.Consensus.Praos.Crypto.PraosCrypto` typeclass (reexported by `Praos` and `TPraos` modules). This simplifies some constraints down the road.

The hypothesis this work is based on is that if we manage to locate all dependencies from consensus to ledger for KES and VRF stuff, then introducing a new `Crypto` typeclass and primitives will have a minimal impact over the codebase and could be done independently of the refactoring in the cardano-ledger.

## Add Property to check "dispatching" of VRF per Era

The goal of #4150 is to ensure different eras can have different crypto implementations for things such as VRF proofs generation and verification, and KES signing. While #4151 provided solutions to this problem, this PR tries to make the problem manifest through a property that, given an arbitrary header from different eras, verifies the correct crypto is used. This is done simply by verifying the sizes of the VRF certificates which are different between plain Praos and Batch compatible VRF. 
 
# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/MockCrypto.hs
+++ b/ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/MockCrypto.hs
@@ -10,8 +10,8 @@ import           Cardano.Crypto.KES (MockKES)
 import           Cardano.Crypto.VRF (MockVRF)
 
 import           Cardano.Ledger.Crypto (Crypto (..))
+import qualified Cardano.Protocol.TPraos.API as Ledger
 
-import qualified Ouroboros.Consensus.Protocol.Praos as Praos
 import qualified Ouroboros.Consensus.Protocol.TPraos as TPraos
 
 -- | A replacement for 'Test.Consensus.Shelley.MockCrypto' that is compatible
@@ -41,6 +41,6 @@ instance Crypto MockCryptoCompatByron where
   type KES      MockCryptoCompatByron = MockKES 10
   type VRF      MockCryptoCompatByron = MockVRF
 
-instance TPraos.PraosCrypto MockCryptoCompatByron
+instance Ledger.PraosCrypto MockCryptoCompatByron
 
-instance Praos.PraosCrypto MockCryptoCompatByron
+instance TPraos.PraosCrypto MockCryptoCompatByron
--- a/ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/MockCrypto.hs
+++ b/ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/MockCrypto.hs
@@ -10,7 +10,7 @@ import           Cardano.Crypto.KES (MockKES)
 import           Cardano.Crypto.VRF (MockVRF)
 
 import           Cardano.Ledger.Crypto (Crypto (..))
-import qualified Cardano.Protocol.TPraos.API as Ledger
+import qualified Cardano.Protocol.TPraos.API as Protocol
 
 import qualified Ouroboros.Consensus.Protocol.TPraos as TPraos
 
@@ -41,6 +41,6 @@ instance Crypto MockCryptoCompatByron where
   type KES      MockCryptoCompatByron = MockKES 10
   type VRF      MockCryptoCompatByron = MockVRF
 
-instance Ledger.PraosCrypto MockCryptoCompatByron
+instance Protocol.PraosCrypto MockCryptoCompatByron
 
 instance TPraos.PraosCrypto MockCryptoCompatByron
--- a/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
+++ b/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
@@ -41,16 +41,20 @@ tests =
 
 -- | Check that Babbage and Conway blocks use different VRF crypto.
 --
+-- This test is based on the following steps:
+--
 -- 1. generate (forge?) babbage or conway headers
---   - those should contain different VRF proofs (because they are supposed to use different algorithms)
+--   - those should contain different VRF proofs (because they are supposed to
+--     use different algorithms)
 --   - the VRF proof should be invalid for the other era
 -- 2. call some header validation functions that's calling VRF certificate check
 -- 3. assert that different VRF function is called for each era
 --    - the header should be valid
 --
--- * why not mock everything? because it does not check that we are implementing things correctly for Conway
---   we would like to test the dispatching induced by this type, to make it clear that Conway relies on different
---   crypto primitives
+-- * why not mock everything? because it does not check that we are
+--   implementing things correctly for Conway we would like to test
+--   the dispatchign induced by this type, to make clear Conway relies
+--   on different crypto primitives
 --
 -- What needs to change is this type:
 --
--- a/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
+++ b/ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
@@ -1,15 +1,13 @@
 {-# LANGUAGE FlexibleContexts      #-}
 {-# LANGUAGE FlexibleInstances     #-}
 {-# LANGUAGE GADTs                 #-}
-
+{-# LANGUAGE LambdaCase            #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns        #-}
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE PatternSynonyms       #-}
 {-# LANGUAGE RankNTypes            #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
-
-
 {-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE UndecidableInstances  #-}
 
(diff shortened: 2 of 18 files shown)


nfrisby:
I tried to skim this in a small gap I had in my schedule, but I'm having trouble seeing the essence of the diff with the many stylistic changes. Any chance you could remove the extraneous parts of the diff?

abailly-iohk:
@nfrisby Yes, I will try to clean things up. It's annoying automatic formatting does not work well, I can assure you I ran the script before each commit :)

abailly-iohk:
@nfrisby Seems like the change is much slimmer now :)

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
```diff
@@ -0,0 +1,84 @@
...
+-- HFBlock even with a single function would be evidence we are doing
+-- the dispatching right => we don't test the actual crypto functions,
+-- only the "dispatching" logic that requires different instances for
+-- different eras.
+--
+prop_VRFCryptoDependsOnBlockEra :: CardanoHeader StandardCrypto -> Property
+prop_VRFCryptoDependsOnBlockEra h =
+  case h of
+    HeaderBabbage ShelleyHeader {shelleyHeaderRaw} ->
+      sizeCertVRF (pTieBreakVRFValue shelleyHeaderRaw) === fromIntegral certSizeVRF
```
dnadales: Is `certSizeVRF` different accross eras? Shouldn't we assert this as well here? In this way we would make this fact clear in this test, and as some extra safety (even though the sizes are not very likely to change).

dnadales (commented):
Looks great so far!

ouroboros-consensus-cardano-test/src/Test/Consensus/Cardano/MockCrypto.hs
```diff
@@ -10,8 +10,8 @@ import           Cardano.Crypto.KES (MockKES)
 import           Cardano.Crypto.VRF (MockVRF)
 
 import           Cardano.Ledger.Crypto (Crypto (..))
+import qualified Cardano.Protocol.TPraos.API as Ledger
```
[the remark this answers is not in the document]
abailly-iohk: Not sure why I called this `Ledger` to be honest 🤔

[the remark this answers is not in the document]
abailly-iohk: Ah I know, it's because it's a package that lives in the ledger, but that ultimately should move to consensus as it's not relevant to the ledger.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -516,7 +514,7 @@ translateLedgerViewByronToShelleyWrapper =
 -------------------------------------------------------------------------------}
 
 translateLedgerStateShelleyToAllegraWrapper ::
-     (PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
+     (PraosCrypto c)
```
Jasagredo, line 517: As there is only one constraint I usually favor removing the parentheses although I'm not sure what the style guide says

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -625,7 +623,7 @@ translateValidatedTxMaryToAlonzoWrapper ctxt = InjectValidatedTx $
 -------------------------------------------------------------------------------}
 
 translateLedgerStateAlonzoToBabbageWrapper ::
-     (Praos.PraosCrypto c, TPraos.PraosCrypto c)
+     (Praos.PraosCrypto c)
```
Jasagredo, line 626: Why this one is qualified? Isn't it the same typeclass as the others?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
```diff
@@ -0,0 +1,94 @@
...
+-- What needs to change is this type:
+--
+-- @@
+-- type CardanoShelleyEras c =
+--   '[ ShelleyBlock (TPraos c) (ShelleyEra c)
+--    , ShelleyBlock (TPraos c) (AllegraEra c)
+--    , ShelleyBlock (TPraos c) (MaryEra c)
+--    , ShelleyBlock (TPraos c) (AlonzoEra c)
+--    , ShelleyBlock (Praos c)  (BabbageEra c)
+--    , ShelleyBlock (Praos c)  (ConwayEra c)
```
Jasagredo, line 64: > Check Babbage and Conway blocks use different VRF crypto.

I don't understand why they use different crypto? To me it seems the same, both use `PraosCrypto` right?

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
abailly-iohk: Indeed, that's the object of the next PR! I just wanted to write a test asserting what the current situation is, and then write a failing test and fix it by setting different cert sizes.

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
```diff
@@ -0,0 +1,94 @@
...
+    HeaderMary ShelleyHeader {shelleyHeaderRaw} ->
+      certVRFHasPraosSize shelleyHeaderRaw & label "Mary"
+    HeaderAlonzo ShelleyHeader {shelleyHeaderRaw} ->
+      certVRFHasPraosSize shelleyHeaderRaw & label "Alonzo"
+    HeaderBabbage ShelleyHeader {shelleyHeaderRaw} ->
+      certVRFHasPraosSize shelleyHeaderRaw & label "Babbage"
+    HeaderConway ShelleyHeader {shelleyHeaderRaw} ->
+      -- TODO: this is were we need to change to check we use in the Conway case
+      -- Cardano.Crypto.VRF.PraosBatchCompat.certSizevrf
+      certVRFHasPraosSize shelleyHeaderRaw & label "Conway"
```
Jasagredo, line 90: So right now it must be the case that this test fails, until we resolve this TODO, right?

abailly-iohk: In the next PR, yes. I would like to first merge a PR that does not break anything

ouroboros-consensus-cardano-test/test/Test/Consensus/Cardano/Crypto.hs
abailly-iohk: That's the crux of the issue: They both use `PraosCrypto` and the logic is the same, but the underlying types of VRF and KES are different so we need to find a way to have different _implementations_ for different eras.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
abailly-iohk: Just to have a minimal change

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
abailly-iohk: Seems like the style guide agrees with you :)

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Crypto.hs
```diff
@@ -0,0 +1,39 @@
...
+import           Ouroboros.Consensus.Protocol.Praos.VRF (InputVRF)
+
+class
+    ( KES.KESAlgorithm (KES c)
+    , KES.ContextKES (KES c) ~ ()
+    , KES.Signable (KES c) (HeaderBody c)
+    , VRF.VRFAlgorithm (VRF c)
+    , VRF.Signable (VRF c) InputVRF
+    , DSIGN.Signable (DSIGN c) (OCertSignable c)
+    , DSIGN.Signable (DSIGN c) (Hash.Hash (HASH c) EraIndependentTxBody)
```
Jasagredo, line 32: I _think_ this is almost the same as `DSignable c (Hash c EraIndependentTxBody)` used in `Shelley.Eras`. Maybe they can be merged and only require it here?

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos.hs
```diff
@@ -92,17 +91,6 @@ import           Ouroboros.Consensus.Util.Versioned (VersionDecoder (Decode),
 
 data Praos c
 
-class
-  ( Crypto c,
-    DSIGN.Signable (DSIGN c) (OCertSignable c),
-    DSIGN.Signable (DSIGN c) (SL.Hash c EraIndependentTxBody),
```
Jasagredo, line 98: In particular I'm unsure if this was copied right, or the `HASH c` is now wrong? there seems to be a small inconsistency here.

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
```diff
@@ -516,7 +514,7 @@ translateLedgerViewByronToShelleyWrapper =
 -------------------------------------------------------------------------------}
 
 translateLedgerStateShelleyToAllegraWrapper ::
-     (PraosCrypto c, DSignable c (Hash c EraIndependentTxBody))
```
Jasagredo, line 519: Here you were able to remove it, but not from shelley eras. I think is should be removable also there

Jasagredo (approved):
Looks good, just a comment about a (maybe?) redundant constraint

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos.hs
abailly-iohk: I _think_ these are not exactly the same things: `Hash.Hash` takes a hash type which `HASH c`  provides via the type family in `Crypto` but `SL.Hash` might very well take a `c` and do the "magic" of calling `Hash.Hash` with `HASH c`. This is common pattern in ledger base types which BTW introduces an additional dependency to the ledger everywhere those types are used, which is really annoying becuase it increases the dependency surface. 
I failed to see what real value those aliases provide but 🤷

ouroboros-consensus-cardano/src/Ouroboros/Consensus/Cardano/CanHardFork.hs
abailly-iohk: Will fix when I get back to working dir that builds :)

ouroboros-consensus-protocol/src/Ouroboros/Consensus/Protocol/Praos/Crypto.hs
abailly-iohk: They are nearly similar but not quite, I will leave it as it is now because I barely understand what I am doing

PR #4395, opened 2023-02-24
RawBearer API

tdammers:
# Description

This provides a socket-like API for Snockets (backed by sockets / named pipes, or simulated in IOSim), reading/writing directly to/from raw memory buffers.

This will be needed to securely send KES keys over a network connection or local pipe, because we need to ensure that key data is never stored on disk, nor on the GHC heap, from where it might be swapped to disk.

For context, see also:

- https://github.com/input-output-hk/cardano-base/pull/255: KES Secure Forgetting
- https://github.com/input-output-hk/cardano-base/issues/312: KES Agent
- https://github.com/input-output-hk/cardano-base/pull/317: KES Agent prerequisites

# Checklist

- Branch
    - [X] The documentation has been properly updated
    - [X] New tests are added if needed and existing tests are updated
    - [X] Any changes affecting Consensus packages must have an entry in the appropriate `changelog.d` directory created using [`scriv`](https://github.com/input-output-hk/scriv). If in doubt, see the [Consensus release process](../ouroboros-consensus/docs/ReleaseProcess.md).
    - (n/a) If this branch changes Network and has any consequences for downstream repositories or end users, said changes must be documented in [`interface-CHANGELOG.md`](../docs/interface-CHANGELOG.md)
    - [X] If serialization changes, user-facing consequences (e.g. replay from genesis) are confirmed to be intentional.
- Pull Request
    - [X] Self-reviewed the diff
    - [X] Useful pull request description at least containing the following information:
      - What does this PR change?
    - [X] Reviewer requested

### DIFF
--- a/ouroboros-network-framework/CHANGELOG.md
+++ b/ouroboros-network-framework/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Renamed `requestOutboundConnection` to `acquireOutboundConnection` and
   `unregister{Inbound,Outbound}Connection` to `release{Inbound,Outbound}Connection`.
   `AssertionLocation` constructors were renamed as well.
+* Added `RawBearer` API (see https://github.com/IntersectMBO/ouroboros-network/pull/4395)
 
 ### Non-breaking changes
 
@@ -43,6 +44,7 @@
 * `Ouroboros.Network.ConnectionManager.Core` must be imported qualified.
 * `ConnectionManagerTrace` moved from `Ouroboros.Network.ConnectionManager.Types`
   to the `Core` module & renamed as `Trace`.
+* RawBearer API (typeclass and instances) added.
 
 ### Non-breaking changes
 
--- a/ouroboros-network-framework/changelog.d/20230224_094922_tdammers_raw_bearer.rst
+++ b/ouroboros-network-framework/changelog.d/20230224_094922_tdammers_raw_bearer.rst
@@ -1,6 +1,0 @@
-Added
------
-
-- RawBearer API
-- ToRawBearer typeclass
-- ToRawBearer instances for `Socket`, `LocalSocket`, and `Simulation.Network.Snocket.FD`
--- a/ouroboros-network-framework/changelog.d/20230224_094922_tdammers_raw_bearer.rst
+++ b/ouroboros-network-framework/changelog.d/20230224_094922_tdammers_raw_bearer.rst
@@ -1,0 +1,6 @@
+Added
+-----
+
+- RawBearer API
+- ToRawBearer typeclass
+- ToRawBearer instances for `Socket`, `LocalSocket`, and `Simulation.Network.Snocket.FD`
--- a/ouroboros-network-framework/src/Ouroboros/Network/RawBearer.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/RawBearer.hs
@@ -1,0 +1,46 @@
+{-# LANGUAGE CPP                   #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+
+module Ouroboros.Network.RawBearer where
+
+import           Data.Word (Word8)
+import           Foreign.Ptr (Ptr)
+import           Network.Socket (Socket)
+import qualified Network.Socket as Socket
+
+#if defined(mingw32_HOST_OS)
+import           Foreign.Ptr (castPtr)
+import qualified System.Win32 as Win32
+#endif
+
+-- | Generalized API for sending and receiving raw bytes over a file
+-- descriptor, socket, or similar object.
+data RawBearer m =
+  RawBearer
+    { send :: Ptr Word8 -> Int -> m Int
+    , recv :: Ptr Word8 -> Int -> m Int
+    }
+
+class ToRawBearer m fd where
+  toRawBearer :: fd -> m (RawBearer m)
+
+instance ToRawBearer IO Socket where
+  toRawBearer s =
+    return RawBearer
+      { send = Socket.sendBuf s
+      , recv = Socket.recvBuf s
+      }
+
+#if defined(mingw32_HOST_OS)
+
+-- | We cannot declare an @instance ToRawBearer Win32.HANDLE@, because
+-- 'Win32.Handle' is just a type alias for @Ptr ()@. So instead, we provide
+-- this function, which can be used to implement 'ToRawBearer' elsewhere (e.g.
+-- over a newtype).
+win32HandleToRawBearer :: Win32.HANDLE -> RawBearer IO
+win32HandleToRawBearer s =
+    RawBearer
+      { send = \buf size -> fromIntegral <$> Win32.win32_WriteFile s (castPtr buf) (fromIntegral size) Nothing
+      , recv = \buf size -> fromIntegral <$> Win32.win32_ReadFile s (castPtr buf) (fromIntegral size) Nothing
+      }
+#endif
--- a/ouroboros-network-framework/src/Ouroboros/Network/RawBearer.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/RawBearer.hs
@@ -3,14 +3,14 @@
 
 module Ouroboros.Network.RawBearer where
 
-import           Data.Word (Word8)
-import           Foreign.Ptr (Ptr)
-import           Network.Socket (Socket)
-import qualified Network.Socket as Socket
+import Data.Word (Word8)
+import Foreign.Ptr (Ptr)
+import Network.Socket (Socket)
+import Network.Socket qualified as Socket
 
 #if defined(mingw32_HOST_OS)
-import           Foreign.Ptr (castPtr)
-import qualified System.Win32 as Win32
+import Foreign.Ptr (castPtr)
+import System.Win32 qualified as Win32
 #endif
 
 -- | Generalized API for sending and receiving raw bytes over a file
@@ -21,22 +21,25 @@ data RawBearer m =
     , recv :: Ptr Word8 -> Int -> m Int
     }
 
-class ToRawBearer m fd where
-  toRawBearer :: fd -> m (RawBearer m)
+newtype MakeRawBearer m fd = MakeRawBearer {
+  getRawBearer :: fd -> m (RawBearer m)
+}
 
-instance ToRawBearer IO Socket where
-  toRawBearer s =
-    return RawBearer
+makeSocketRawBearer :: MakeRawBearer IO Socket
+makeSocketRawBearer = MakeRawBearer (return . socketToRawBearer)
+
+socketToRawBearer :: Socket -> RawBearer IO
+socketToRawBearer s =
+    RawBearer
       { send = Socket.sendBuf s
       , recv = Socket.recvBuf s
       }
 
 #if defined(mingw32_HOST_OS)
 
--- | We cannot declare an @instance ToRawBearer Win32.HANDLE@, because
--- 'Win32.Handle' is just a type alias for @Ptr ()@. So instead, we provide
--- this function, which can be used to implement 'ToRawBearer' elsewhere (e.g.
--- over a newtype).
+win32MakeRawBearer :: MakeRawBearer IO Win32.HANDLE
+win32MakeRawBearer = MakeRawBearer (return . win32HandleToRawBearer)
+
 win32HandleToRawBearer :: Win32.HANDLE -> RawBearer IO
 win32HandleToRawBearer s =
     RawBearer
--- a/ouroboros-network-framework/src/Simulation/Network/Snocket.hs
+++ b/ouroboros-network-framework/src/Simulation/Network/Snocket.hs
@@ -49,20 +49,28 @@ import Prelude hiding (read)
 import Control.Applicative (Alternative)
 import Control.Concurrent.Class.MonadSTM qualified as LazySTM
 import Control.Concurrent.Class.MonadSTM.Strict
-import Control.Monad (when)
+import Control.Monad.Class.MonadSay
+import Control.Monad.Class.MonadST
 import Control.Monad.Class.MonadThrow
-import Control.Monad.Class.MonadTime.SI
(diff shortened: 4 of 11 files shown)


ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -528,8 +536,92 @@ instance Show addr => Show (FD_ m addr) where
...
+instance ( MonadST m
+         , MonadThrow m
+         , MonadSay m
+         , MonadLabelledSTM m
+         , Show addr
+         ) => ToRawBearer m (FD m (TestAddress addr)) where
+  toRawBearer = makeRawFDBearer
+
+-- | Make a 'RawBearer' from an 'FD'. Since this is only used for testing, we
+-- can bypass the requirement of moving raw bytes directly between file
```
coot, line 550:
```suggestion
-- can bypas the requirement of moving raw bytes directly between file
```

coot, line 550:
```suggestion
-- can bypas the requirement of moving raw bytes directly between file
```

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -528,8 +536,92 @@ instance Show addr => Show (FD_ m addr) where
...
+                   , Show addr
+                   )
+                => FD m (TestAddress addr)
+                -> m (RawBearer m)
+makeRawFDBearer (FD {fdVar}) = do
+    (bufVar :: StrictTMVar m LBS.ByteString) <- newTMVarIO LBS.empty
+    return RawBearer
+              { send = \src srcSize -> do
+                  labelTVarIO fdVar "sender"
+                  say $ "Sending " ++ show srcSize ++ " bytes"
```
coot, line 569: I would avoid that, because this can be very noisy.  It's better to provide a tracer which the user can fill with `nullTracer` or something better for debugging.

That's another argument to abandon the `ToRawBearer` type class (we probably don't want to have that tracer for real fds).

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
```diff
@@ -528,8 +536,92 @@ instance Show addr => Show (FD_ m addr) where
...
+                      say $ "Checking buffer"
+                      bytesFromBuffer <- atomically $ takeTMVar bufVar
+                      say $ "Buffer: " ++ show bytesFromBuffer
+                      (lhs, rhs) <- if not (LBS.null bytesFromBuffer) then do
+                        say $ "Reading up to " ++ show size ++ " bytes from buffer"
+                        return (LBS.take size64 bytesFromBuffer, LBS.drop size64 bytesFromBuffer)
+                      else do
+                        bytesRead <- acRead (connChannelLocal conn)
+                        say $ "Received " ++ show (LBS.length $ LBS.take size64 bytesRead) ++ " or more bytes"
+                        return (LBS.take size64 bytesRead, LBS.drop size64 bytesRead)
```
coot, line 597: Just a different code layout:
```suggestion
                      (lhs, rhs) <- if not (LBS.null bytesFromBuffer)
                        then do
                          say $ "Reading up to " ++ show size ++ " bytes from buffer"
                          return (LBS.take size64 bytesFromBuffer, LBS.drop size64 bytesFromBuffer)
                        else do
                          bytesRead <- acRead (connChannelLocal conn)
                          say $ "Received " ++ show (LBS.length $ LBS.take size64 bytesRead) ++ " or more bytes"
                          return (LBS.take size64 bytesRead, LBS.drop size64 bytesRead)
```

ouroboros-network-framework/test/Test/Ouroboros/Network/RawBearer.hs
```diff
@@ -0,0 +1,212 @@
...
+      (socketSnocket iomgr)
+      clientAddr serverAddr
+      msg `finally` do
+        cleanUp clientName
+        cleanUp serverName
+  where
+    cleanUp name = do
+        catchJust (\e -> if isDoesNotExistErrorType (ioeGetErrorType e) then Just () else Nothing)
+                  (removeFile name)
+                  (\_ -> return ())
```
coot, line 84: I think that's simpler, but I don't mind `catchJust`.
```suggestion
        catch (removeFile name)
              (\e -> if isDoesNotExistErrorType (ioeGetErrorType e)
                     then return ()
                     else throwIO e)

```

ouroboros-network-framework/test/Test/Ouroboros/Network/RawBearer.hs
```diff
@@ -0,0 +1,212 @@
...
+newtype Message = Message { messageBytes :: ByteString }
+  deriving (Show, Eq, Ord)
+
+instance Arbitrary Message where
+  shrink = filter (not . BS.null . messageBytes) . fmap (Message . BS.pack) . shrink . BS.unpack . messageBytes
+  arbitrary = Message . BS.pack <$> listOf1 arbitrary
+
+newtype TestError =
+  TestError String
+  deriving (Show)
```
coot, line 109:
```suggestion
newtype TestError = TestError String
  deriving Show
```

ouroboros-network-framework/test/Test/Ouroboros/Network/RawBearer.hs
```diff
@@ -0,0 +1,212 @@
...
+                case accepted :: Accepted fd addr of
+                  AcceptFailure err ->
+                    throwIO err
+                  Accepted s' _ -> do
+                    labelThisThread "accept"
+                    say "receiver: connection accepted"
+                    flip finally (say "receiver: closing connection" >> close snocket s' >> say "receiver: connection closed") $ do
+                      bearer <- toRawBearer s'
+                      retval <- bracket (io $ mallocBytes size) (io . free) $ \dstBuf -> do
+                        let go _ 0 = do
```
coot, line 167: Will this be safer?
```suggestion
                        let go _ n | n <= 0 = do
```

ouroboros-network-framework/test/Test/Ouroboros/Network/RawBearer.hs
```diff
@@ -0,0 +1,212 @@
...
+    senderDone <- newEmptyMVar
+    let sender = bracket (openToConnect snocket clientAddr) (close snocket) $ \s -> do
+                    say "sender: connecting"
+                    connect snocket s serverAddr
+                    say "sender: connected"
+                    bearer <- toRawBearer s
+                    bracket (io $ mallocBytes size) (io . free) $ \srcBuf -> do
+                      io $ BS.useAsCStringLen (messageBytes msg)
+                            (uncurry (copyBytes srcBuf))
+                      let go _ 0 = do
```
coot, line 141: Is this safer?
```suggestion
                      let go _ n | n <= 0 = do
```

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
tdammers: I don't think so. https://www.thefreedictionary.com/bypass

ouroboros-network-framework/src/Ouroboros/Network/RawBearer.hs
```diff
@@ -0,0 +1,46 @@
...
+      { send = Socket.sendBuf s
+      , recv = Socket.recvBuf s
+      }
+
+#if defined(mingw32_HOST_OS)
+
+-- | We cannot declare an @instance ToRawBearer Win32.HANDLE@, because
+-- 'Win32.Handle' is just a type alias for @Ptr ()@. So instead, we provide
+-- this function, which can be used to implement 'ToRawBearer' elsewhere (e.g.
+-- over a newtype).
```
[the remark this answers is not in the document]
tdammers: Sure, we can express the typeclass instance here as a plain function, and then express the typeclass instance in terms of that, like we do for Win32.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
tdammers: Ah, but wait - this is the instance for the simulated snocket, which can only be used in `IOSim`. The "real" instances do not demand `MonadSay`, and they do not emit any trace logging at all. And `IOSim` itself already routes `say` through a `Tracer`, so this is actually working just fine - we get traces when running in `IOSim`, but no performance overhead when running in `IO`.

ouroboros-network-framework/changelog.d/20230224_094922_tdammers_raw_bearer.rst
```diff
@@ -0,0 +1,6 @@
+Added
```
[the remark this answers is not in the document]
tdammers: Fixed.

ouroboros-network-framework/test/Test/Ouroboros/Network/RawBearer.hs
coot: So what about:
```hs
let go _ n | n < 0 = error "ups, I was wrong ;)"
```
:wink:

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
tdammers: Indeed, I've also thought of adding it to `Snocket`, but that would add a lot of noise to all snocket code, despite being a fairly niche feature - I expect that the only practical application of this will be for the KES Agent, where we need to send private keys over network sockets without ever leaking them to disk (including swap); anything else can just use the existing networking features and plain old serialization through intermediate values on the Haskell heap.

Your idea of injecting the tracer through `makeRawFDBearer` won't work, because the test code needs to run against both simulated and real snockets; since it's generalized over the `fd` type, it doesn't know to call `makeRawFDBearer`, so unless we give up on the generalization (which means we can no longer test the actual identical code against `IOSim` unchanged, which is the whole point of this), the only ways I can see for this to work would be:

1. Extend the `send` and `recv` methods to accept a `Tracer` as an extra argument. Downside is that we'll be passing tracers to production instances, which will then silently ignore it - a lot of noise, potentially misleading, and there might even be performance overhead to it as well.
2. Pass a function `fd -> m (RawBearer m)` around as an argument; we can now inject the tracer by closing that function over it. This is essentially the same as the typeclass approach, except that we're passing the typeclass dictionary (or the single method, in this case) around explicitly.
3. Same as 2., but make that function part of the `Snocket` data structure.

I'm not a huge fan of either, but if we really feel the need to change the tracer, then option 2 would be my preference.

ouroboros-network-framework/src/Ouroboros/Network/RawBearer.hs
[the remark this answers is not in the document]
tdammers: I've actually gotten rid of the typeclass after all.

ouroboros-network-framework/src/Simulation/Network/Snocket.hs
tdammers: OK, ran with it and implemented a better tracing solution:
- got rid of the typeclass after all
- introduced `MakeRawBearer` type (analogous to `MakeBearer`)
- stringly-typed debug logging calls inside the simulated socket's RawBearer now use a structured trace message type, `FDRawBearerTrace`
- when making a `MakeRawBearer` for simulated sockets, caller must pass a `Tracer m FDRawBearerTrace`

ouroboros-network-framework/test/Test/Ouroboros/Network/RawBearer.hs
tdammers: As above.

coot (approved):
LGTM.

coot:
It can be merged once CI passes, we should merge it after #4571.

ouroboros-network-framework/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Non-breaking changes
 
 * `ghc-9.4` and `ghc-9.6` compatibility.
+* RawBearer API (typeclass and instances) added.
```
coot, line 12: You should add this to a new section which starts with:
```
## next version

### Non-breaking changes
...
```

coot:
There's a conflict in `CHANGELOG.md`; Could you squash / clean the branch a bit before merging?

coot (changes requested):

coot:
@tdammers do we still need this PR?

ouroboros-network-framework/src/Ouroboros/Network/RawBearer.hs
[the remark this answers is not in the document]
coot: The haddock needs to be update, since there's not `ToRawBearer` type class.

ouroboros-network-framework/test/Main.hs
```diff
@@ -0,0 +1,30 @@
...
+  testGroup "ouroboros-network-framework"
+  [ ConnectionManager.tests
+  , Driver.tests
+  , Server2.tests
+  , Socket.tests
+  , Subscription.tests
+  , RateLimiting.tests
+  , Snocket.tests
+  , RawBearer.tests
+  ]
```
coot, line 28: I think you want to add the tests to `ouroboros-network-framework:io-tests`; the `Main.hs` might be an artefact of a rebase error.

coot (changes requested):

ouroboros-network-framework/test/Main.hs
tdammers: Refactored things into the two test suites.

coot:
@tdammers have you pushed your changes?

tdammers:
Everything pushed now.

ouroboros-network-framework/src/Ouroboros/Network/RawBearer.hs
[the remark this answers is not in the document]
coot: ^ ping

Also the CHANGELOG mentions the type class.

coot (approved):
LGTM

PR #4400, opened 2023-02-24
Add a bechmark for adding transactions to the mempool

dnadales:
Part of #4382 

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus-test/bench-mempool/Main.hs
+++ b/ouroboros-consensus-test/bench-mempool/Main.hs
@@ -1,0 +1,4 @@
+module Main (main) where
+
+main :: IO ()
+main = pure ()
--- a/ouroboros-consensus-test/bench-mempool/Main.hs
+++ b/ouroboros-consensus-test/bench-mempool/Main.hs
@@ -1,15 +1,15 @@
+{-# LANGUAGE FlexibleContexts   #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedLists    #-}
-
+{-# LANGUAGE TupleSections      #-}
 
 module Main (main) where
 
-import           Control.Monad (void)
 import           Data.Set ()
 import           Test.Tasty.HUnit (testCase, (@?=))
 
 import qualified Control.Tracer as Tracer
-import           Test.Tasty.Bench (bench, bgroup, defaultMain, nfIO)
+import           Test.Tasty.Bench (bench, bgroup, defaultMain, env, nfIO)
 
 import           Bench.Consensus.Mempool
 import           Bench.Consensus.Mempool.TestBlock (TestBlock)
@@ -21,40 +21,41 @@ main :: IO ()
 main = defaultMain [ benchmarkJustAddingTransactions ]
   where
     benchmarkJustAddingTransactions =
-        bgroup "Just adding transactions"
-          $  fmap benchAddNTxs [1_000, 10_000, 100_000, 1_000_000]
-          <>
-             [ testCase "Transactions are added" $ do
-                 let
-                   n   = 1000
-                   txs = mkNTryAddTxs 1000
-                 mempool    <- addNTxs txs
-                 mempoolTxs <- getTxs mempool
-                 mempoolTxs @?= txsAddedInCmds txs
-                 length mempoolTxs @?= n
-             ]
+        bgroup "Just adding transactions" $
+            fmap benchAddNTxs [1_000, 10_000, 100_000, 1_000_000]
       where
-        benchAddNTxs n = bench (show n) $ nfIO $ addNTxs' $ mkNTryAddTxs n
-        -- FIXME: check that we need forcing. Forcing the list does not seem to make a difference
+        benchAddNTxs n =
+          env (let txs = mkNTryAddTxs n in fmap (, txs) (openMempoolWithCapacipyFor txs))
+              -- The irrefutable pattern was added to avoid the
+              --
+              -- > Unhandled resource. Probably a bug in the runner you're using.
+              --
+              -- error reported here https://hackage.haskell.org/package/tasty-bench-0.3.3/docs/Test-Tasty-Bench.html#v:env
+              (\ ~(mempool, txs) -> bgroup (show n) [
+                  bench    "benchmark"     $ nfIO $ run        mempool txs
+                , testCase "test"          $        testAddTxs mempool txs
+                , testCase "txs length"    $ length txs @?= n
+                ]
+              )
+          where
+            testAddTxs mempool txs = do
+                run mempool txs
+                mempoolTxs <- getTxs mempool
+                mempoolTxs @?= txsAddedInCmds txs
 
 {-------------------------------------------------------------------------------
   Adding TestBlock transactions to a mempool
 -------------------------------------------------------------------------------}
 
-addNTxs' :: [MempoolCmd TestBlock] -> IO ()
-addNTxs' = void . addNTxs
-
-addNTxs :: [MempoolCmd TestBlock] -> IO (MempoolWithMockedLedgerItf IO TestBlock)
-addNTxs cmds = do
-  let capacityRequiredByCmds = Mempool.mkCapacityBytesOverride  totalTxsSize   -- TODO: add a convenience function in the Mempool API.
-        where totalTxsSize = sum $ fmap TestBlock.txSize $ txsAddedInCmds cmds
-  mempool <- openMempoolWithMockedLedgerItf
-                 capacityRequiredByCmds
-                 Tracer.nullTracer
-                 TestBlock.txSize
-                 TestBlock.sampleMempoolAndModelParams
-  run mempool cmds
-  pure mempool
+openMempoolWithCapacipyFor :: [MempoolCmd TestBlock] ->  IO (MempoolWithMockedLedgerItf IO TestBlock)
+openMempoolWithCapacipyFor cmds =
+    openMempoolWithMockedLedgerItf capacityRequiredByCmds
+                                   Tracer.nullTracer
+                                   TestBlock.txSize
+                                   TestBlock.sampleMempoolAndModelParams
+  where
+    capacityRequiredByCmds = Mempool.mkCapacityBytesOverride  totalTxsSize
+      where totalTxsSize = sum $ fmap TestBlock.txSize $ txsAddedInCmds cmds
 
 mkNTryAddTxs :: Int -> [MempoolCmd TestBlock.TestBlock]
 mkNTryAddTxs 0 = []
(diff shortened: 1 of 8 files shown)


abailly-iohk (approved):
LGTM!  Ideally we would like to dump/publish those results somewhere and enable comparisons across changes, otherwise I am afraid we won't reap the benefits of this approach.

ouroboros-consensus-test/ouroboros-consensus-test.cabal
```diff
@@ -326,3 +326,45 @@ test-suite test-infra
                        -Wredundant-constraints
                        -Wmissing-export-lists
                        -fno-ignore-asserts
+
+benchmark bench-mempool
```
jorisdral, line 330: If we plan on writing more benchmarks for other consensus components, would we want to have a single benchmark executable or one for each component that we target?

ouroboros-consensus-test/ouroboros-consensus-test.cabal
```diff
@@ -326,3 +326,45 @@ test-suite test-infra
...
+  ghc-options:         -Wall
+                       -Wcompat
+                       -Wincomplete-uni-patterns
+                       -Wincomplete-record-updates
+                       -Wpartial-fields
+                       -Widentities
+                       -Wredundant-constraints
+                       -Wmissing-export-lists
+                       -Wno-unticked-promoted-constructors
+                       -fno-ignore-asserts
```
jorisdral, line 363: If the code under benchmark contains relatively costly assertions, not ignoring assertions might inflate the results. This is just something to take into consideration, it is not something that should change now

ouroboros-consensus/changelog.d/20230224_113043_damian.only_4382_add_getSnapshotFor_benchmarks.md
```diff
@@ -0,0 +1,5 @@
+### Non-Breaking
+
+- Added `mkCapacityBytesOverride`, a convenience function to create an override
+  for the mempool capacity using the provided number bytes.
```
jorisdral, line 4: Did we decide on present or past tense? I'm mainly using present tense

ouroboros-consensus-test/src/Test/Util/TestBlock.hs
```diff
@@ -355,6 +359,17 @@ instance PayloadSemantics () where
 
   applyPayload _ _ = Right ()
 
+-- | Apply the payload to a ticked state directly to the payload dependent state
+-- portion of it, leaving the rest of the input ticked state unaltered.
```
jorisdral, line 363:
```suggestion
-- | Apply the payload directly to the payload dependent state
-- portion of a ticked state , leaving the rest of the input ticked state unaltered.
```

ouroboros-consensus-test/src/Test/Util/TestBlock.hs
```diff
@@ -355,6 +359,17 @@ instance PayloadSemantics () where
...
+applyDirectlyToPayloadDependentState ::
+     PayloadSemantics ptype
+  => Ticked (LedgerState (TestBlockWith ptype))
+  -> ptype
+  -> Either (PayloadDependentError ptype)
+            (Ticked (LedgerState (TestBlockWith ptype)))
+applyDirectlyToPayloadDependentState (TickedTestLedger st) tx = do
+    payloadDepSt' <- applyPayload (payloadDependentState st) tx
+    pure $ TickedTestLedger $ st { payloadDependentState = payloadDepSt' }
```
jorisdral, line 372:
```suggestion
    pure $ TickedTestLedger $ st { payloadDependentState = payloadDepSt' }
    
```

ouroboros-consensus-test/bench-mempool/Main.hs
```diff
@@ -0,0 +1,65 @@
...
+                 let
+                   n   = 1000
+                   txs = mkNTryAddTxs 1000
+                 mempool    <- addNTxs txs
+                 mempoolTxs <- getTxs mempool
+                 mempoolTxs @?= txsAddedInCmds txs
+                 length mempoolTxs @?= n
+             ]
+      where
+        benchAddNTxs n = bench (show n) $ nfIO $ addNTxs' $ mkNTryAddTxs n
```
jorisdral, line 37: If the cost of adding the txs is what you want to benchmark, and not opening the mocked mempool or creating the mocked transactions you want to add, I would suggest using [`env`/`envWithCleanup`](https://hackage.haskell.org/package/tasty-bench-0.3.3/docs/Test-Tasty-Bench.html#v:env) to do that work before running the benchmark, and then only calling the function-under-bench with those inputs. The inputs will by then already have been forced to normal form by `env`, such that the benchmark only measures the time spent adding the txs. Using `env` might not be strictly useful if the setup is much cheaper than adding the txs, but I don't think it hurts to use `env` regardless

It would look something like:
```suggestion
        benchAddNTxs n = env (doSetup n) $ \(mempool, txs) ->
           bench (show n) $ nfIO $ addNTxs' mempool 
```

ouroboros-consensus-test/bench-mempool/Main.hs
```diff
@@ -0,0 +1,65 @@
...
+          <>
+             [ testCase "Transactions are added" $ do
+                 let
+                   n   = 1000
+                   txs = mkNTryAddTxs 1000
+                 mempool    <- addNTxs txs
+                 mempoolTxs <- getTxs mempool
+                 mempoolTxs @?= txsAddedInCmds txs
+                 length mempoolTxs @?= n
+             ]
```
jorisdral, line 35: To be sure, I would run this test for each of the `n` you use in the benchmarks above here, e.g., by moving the test into `benchAddNTxs`. Each benchmark then has a sanity check

Another addition would be to add a regular quickcheck test that run the test for arbitrary values of `n`, such that we're also leveraging property testing in addition to sanity checks for the benchmarks

ouroboros-consensus-test/bench-mempool/Main.hs
```diff
@@ -0,0 +1,65 @@
...
+      where
+        benchAddNTxs n = bench (show n) $ nfIO $ addNTxs' $ mkNTryAddTxs n
+        -- FIXME: check that we need forcing. Forcing the list does not seem to make a difference
+
+{-------------------------------------------------------------------------------
+  Adding TestBlock transactions to a mempool
+-------------------------------------------------------------------------------}
+
+addNTxs' :: [MempoolCmd TestBlock] -> IO ()
+addNTxs' = void . addNTxs
```
jorisdral, line 45: I can't check now because the module containing `MemppolWithMockedLedgerItf` is not available on the branch, but could the `void` here be preventing some evaluation inside the mempool that we would want to force?

jorisdral:
> LGTM! Ideally we would like to dump/publish those results somewhere and enable comparisons across changes, otherwise I am afraid we won't reap the benefits of this approach.

One option with respect to this would be to output the txs to a file, and use that file as input to the benchmark using [`env`](https://hackage.haskell.org/package/tasty-bench-0.2.5/docs/Test-Tasty-Bench.html#v:env)

jorisdral (commented):
Added some remarks!

dnadales:
> LGTM! Ideally we would like to dump/publish those results somewhere and enable comparisons across changes, otherwise I am afraid we won't reap the benefits of this approach.

Makes sense. Do you have any references I could look into to see how this is done? We can do this in a following step.

ouroboros-consensus-test/ouroboros-consensus-test.cabal
dnadales: What about one per component? Why? To keep them isolated. But I don't know if the other approach as more advantages. Thoughts?

ouroboros-consensus-test/ouroboros-consensus-test.cabal
dnadales: True! I missed that. Thanks!

ouroboros-consensus/changelog.d/20230224_113043_damian.only_4382_add_getSnapshotFor_benchmarks.md
dnadales: Good point. We have no consistency at the moment. I'll raise this with the team.

ouroboros-consensus-test/src/Test/Util/TestBlock.hs
dnadales: This is much simpler :heart: Thanks!

ouroboros-consensus-test/bench-mempool/Main.hs
dnadales: Absolutely! I overlooked this. Thank you :pray:

dnadales: Hmm, `env`'s documentation recommends using [`withResource`](https://hackage.haskell.org/package/tasty-1.4.3/docs/Test-Tasty.html#v:withResource), which does not seem to perform any normal form evaluation.

abailly-iohk:
We can definitely do this in a following step. The way I do it is usually quite simple: generate some markdown file containing your data, publish it as an artifact as part of the CI, then download it somewhere else (eg. in the docs) for publication.

ouroboros-consensus-test/bench-mempool/Main.hs
dnadales: I hope I never forget to use a function like `env` when benchmarking. 

Before `env`:

image

With `env`:

image

ouroboros-consensus-test/bench-mempool/Main.hs
dnadales: I removed this function in favour of the new approach using `env`.

ouroboros-consensus-test/ouroboros-consensus-test.cabal
jorisdral: I'm not opposed to having one per component. An advantage to one executable for all is that we can reuse code, but I don't know how often we will actually be in a situation where we can reuse code. I think one advantage of having multiple executables is that benchmark running times tend to be on the lengthy side, so not having one monolith executable that takes ages to run is probably a good thing.

ouroboros-consensus-test/bench-mempool/Main.hs
jorisdral: Generating transactions looked relatively cheap, so maybe opening the mempool was the costly part

ouroboros-consensus-test/bench-mempool/Main.hs
dnadales: Good point. In this case, I wanted to avoid the cost of having to test each of the values of `n`. Eventually, we should test the interpretation of commands by means of the mempool property tests, but I think it makes sense to test each case.

jorisdral (approved):

PR #4409, opened 2023-02-27
Add deprecation warnings to old exposed entities

Jasagredo:
# Description

Some entities that were previously exposed have changed their name or moved locations. This PR adds deprecation warnings in all the entities that were present previously, pointing to the right entity to import and are still aliases so code should continue running even if the warnings are dismissed.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
+++ b/ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
@@ -20,6 +20,10 @@
 {-# LANGUAGE UndecidableInstances       #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
+
+-- Remove this when @TraceLedgerEvent@ is gone
+{-# OPTIONS_GHC -Wno-deprecations #-}
+
 -- | Main tests for the chain DB.
 --
 -- These are the main tests for the chain DB. Commands include
@@ -1667,6 +1671,10 @@ traceEventName = \case
     TraceGCEvent                ev    -> "GC."                <> constrName ev
     TraceIteratorEvent          ev    -> "Iterator."          <> constrName ev
     TraceSnapshotEvent          ev    -> "Ledger."            <> constrName ev
+    -- This one is deprecated but unless we match it we get non-exhaustiveness
+    -- warnings. When this constructor is deleted we must remove this match as
+    -- well as re-enable the deprecation warnings at the top of this file.
+    TraceLedgerEvent            ev    -> "Ledger."            <> constrName ev
     TraceLedgerReplayEvent      ev    -> "LedgerReplay."      <> constrName ev
     TraceImmutableDBEvent       ev    -> "ImmutableDB."       <> constrName ev
     TraceVolatileDBEvent        ev    -> "VolatileDB."        <> constrName ev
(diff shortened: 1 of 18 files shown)


ouroboros-consensus/src/Ouroboros/Consensus/Mempool/Impl/Types.hs
```diff
@@ -16,4 +18,17 @@ module Ouroboros.Consensus.Mempool.Impl.Types {-# DEPRECATED "Use Ouroboros.Cons
   , validationResultFromIS
   ) where
 
+import           Ouroboros.Consensus.Ledger.Basics
+import           Ouroboros.Consensus.Ledger.SupportsMempool
+
+import           Ouroboros.Consensus.Mempool.Capacity
 import           Ouroboros.Consensus.Mempool.Impl.Common
+
+{-# DEPRECATED validateIS "This function is gone, it was internal and no longer used" #-}
```
dnadales, line 27: Maybe "this function should not be used (it will throw an error), it was ...".

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
```diff
@@ -1,5 +1,152 @@
-module Ouroboros.Consensus.Storage.LedgerDB.OnDisk {-# DEPRECATED "Use Ouroboros.Consensus.Storage.LedgerDB instead" #-}
```
dnadales, line 1: So these warnings alone were not enough?

dnadales (approved):

ouroboros-consensus/src/Ouroboros/Consensus/Storage/LedgerDB/OnDisk.hs
Jasagredo: So if things only had moved around, they would probably have been. However as a couple of types changes their name, they were undiscoverable.

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
```diff
@@ -1667,6 +1671,10 @@ traceEventName = \case
     TraceGCEvent                ev    -> "GC."                <> constrName ev
     TraceIteratorEvent          ev    -> "Iterator."          <> constrName ev
     TraceSnapshotEvent          ev    -> "Ledger."            <> constrName ev
+    -- This one is deprecated but unless we match it we get non-exhaustiveness
+    -- warnings. When this constructor is deleted we must remove this match as
+    -- well as re-enable the deprecation warnings at the top of this file.
+    TraceLedgerEvent            ev    -> "Ledger."            <> constrName ev
```
abailly-iohk, line 1677: Suggestion: raise an `error` in the LHS of the match to ensure matches don't go unnoticed?

abailly-iohk (approved):
Great job!

ouroboros-consensus-test/test-storage/Test/Ouroboros/Storage/ChainDB/StateMachine.hs
Jasagredo: That would probably be the right thing to do, but in this case the code already compiles without this constructor, this is just to silence the warning of non-exhaustiveness. We will simple remove it eventually

PR #4427, opened 2023-03-03
Add `cabal-test` script

Jasagredo:
A minimal wrapper over `cabal test` with a nicer output


Jasagredo:
https://gist.github.com/Jasagredo/93cc7258de970a784164d4905028e7f6 This will live as a gist instead

PR #4435, opened 2023-03-08
Delete stale golden files

Jasagredo:
# Description

These golden files no longer will be created by new golden test runs. We can remove them as they became stale


amesgen (approved):
Ideally, we would use a golden test library mechanism that is aware that a certain directory is supposed to contain *just* golden test files and can hence warn about this; but as there is no real risk in keeping these stale files, I don't think this has any priority and we can just remove them when we notice them.

PR #4438, opened 2023-03-09
consensus-test: add chain generators

nfrisby:
# Description

These generators are the first step towards a new set of tests, that will target the upcoming Ouroboros Genesis implementation.
One of the earlier milestones, however, will be some tests that apply even before Ouroboros Genesis.

The tip of the iceberg is the `Test.Ouroboros.Consensus.ChainGenerator.Tests` module. You can chase dependencies from there.

# Checklist

- Branch
    - [x] Any changes in the Consensus API (every exposed function, type or module) that has changed its name, has been deleted, has been moved, or altered in some other significant way must leave behind a `DEPRECATED` warning that notifies downstream consumers. If deprecating a whole module, remember to add it to `./scripts/ci/check-stylish.sh` as otherwise `stylish-haskell` would un-deprecate it.
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Counting.hs
+++ b/ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Counting.hs
@@ -1,0 +1,336 @@
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE ViewPatterns #-}
+
+-- | Very strong types for working with indices, counts, etc within sequences.
+module Test.Ouroboros.Consensus.ChainGenerator.Counting (
+  -- * general counts
+  Count (Count),
+  (-),
+  (+),
+  getCount,
+  forgetBase,
+  forgetElem,
+  -- * indices and sizes
+  Preds,
+  Index,
+  Size,
+  Total,
+  forRangeM_,
+  lastIndex,
+  range, 
+  uniformIndex,
+  -- * windows
+  Lbl (Lbl),
+  SomeWindow (SomeWindow),
+  Win,
+  Contains (Contains, UnsafeContains),
+  forgetWindow,
+  frWin,
+  frWinVar,
+  joinWin,
+  toWin,
+  windowLast,
+  windowSize,
+  windowStart,
+  withSuffixWindow,
+  withTopWindow,
+  withWindow,
+  withWindowBetween,
+  -- * vectors
+  Vector (Vector),
+  MVector (MVector),
+  createV,
+  getMVector,
+  getVector,
+  lengthMV,
+  lengthV,
+  modifyMV,
+  readMV,
+  readV,
+  replicateMV,
+  sliceMV,
+  sliceV,
+  unsafeThawV,
+  writeMV,
+  -- * variables
+  Other,
+  Var,
+  joinVar,
+  toIndex,
+  toSize,
+  toVar,
+  ) where
+
+import           Control.Monad.ST (ST)
+import           Data.Coerce (coerce)
+import           Data.Foldable (forM_)
+import           Data.Kind (Type)
+import           Data.Proxy (Proxy (Proxy))
+import qualified Data.Type.Equality as TypeEq
+import qualified Data.Vector.Unboxed as V
+import qualified Data.Vector.Unboxed.Mutable as MV
+import           GHC.OverloadedLabels (IsLabel (fromLabel))
+import           Prelude hiding ((+), (-))
+import qualified Prelude
+import qualified System.Random.Stateful as R
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Some as Some
+import qualified Test.QuickCheck as QC
+
+-----
+
+infixl 6 .+, .-
+
+(.+) :: Int -> Int -> Int
+(.+) = (Prelude.+)
+
+(.-) :: Int -> Int -> Int
+(.-) = (Prelude.-)
+
+-----
+
+-- | Indexed by what you're counting from, what you're counting, and which of them you're counting.
+newtype Count (base :: Type) (elem :: kelem) (which :: kwhich) = Count Int
+  deriving (QC.Arbitrary, Eq, Ord, Read, Show)
+
+getCount :: Count base elem which -> Int
+getCount (Count n) = n
+
+infixl 6 +, -
+
+(+) :: Count base elem which -> Int -> Count base elem which
+(+) (Count i) j = Count (i .+ j)
+
+(-) :: Count base elem which -> Int -> Count base elem which
+(-) (Count i) j = Count (i .- j)
+
+forgetBase :: Count base elem which -> Some.Forgotten (Count () elem which)
+forgetBase (Count x) = Some.forgotten $ Count x
+
+forgetElem :: Count base elem which -> Some.Forgotten (Count base () which)
+forgetElem (Count x) = Some.forgotten $ Count x
+
+-----
+
+-- | How many preceding elements
+data Preds
+
+-- | How many elements in 'Total'
+data Total
+
+type Index base elem = Count base elem Preds
+type Size  base elem = Count base elem Total
+
+-- | The 'Index' of the rightmost element in the sequence of the given 'Size'
+lastIndex :: Size base elem -> Index base elem
+lastIndex (Count n) = Count (n .- 1)
+
+range :: Size base elem -> [Index base elem]
+range (Count n) = coerce [0 .. max 0 n .- 1]
+
+forRangeM_ :: Monad m => Size base elem -> (Index base elem -> m a) -> m ()
+forRangeM_ c = forM_ (range c)
+
+uniformIndex :: R.StatefulGen g m => Size base elem -> g -> m (Index base elem)
+uniformIndex n g = Count <$> R.uniformRM (0, getCount $ lastIndex n) g
+
+-----
+
+-- | A human-readable label for a 'Win'
+data Lbl lbl = Lbl   -- no explicit kind var so that type applications don't need to provide the kind
+instance (lbl TypeEq.~~ s) => IsLabel s (Lbl lbl) where fromLabel = Lbl
+
+-- | A named window within some containing sequence
+data Win (lbl :: klbl) (skolem :: Type)
+
+-- | The correspondence between a window and its containing sequence
+data Contains (elem :: kelem) (outer :: Type) (inner :: Type) =
+    UnsafeContains
+        !(Index outer elem)   -- ^ first slot's offset in containing sequence
+        !(Size  inner elem)   -- ^ size of window (INVARIANT: does not reach past end of containing sequence)
+  deriving (Eq, Read, Show)
+
+pattern Contains :: Index outer elem -> Size inner elem -> Contains elem outer inner
+pattern Contains x y <- UnsafeContains x y
+
+{-# COMPLETE Contains #-}
+
+forgetWindow :: Contains elem outer inner -> Some.Forgotten (Index outer elem, Index outer elem)
+forgetWindow win = Some.forgotten $ (windowStart win, windowLast win)
+
+frWin :: Contains elem outer inner -> Index inner elem -> Index outer elem
+frWin (Contains (Count i) _n) (Count j) = Count (i .+ j)
+
+frWinVar :: Contains elem outer inner -> Var inner x -> Var outer x
+frWinVar _ (Count x) = Count x
+
+toWin :: Contains elem outer inner -> Index outer elem -> Maybe (Index inner elem)
+{-# INLINE toWin #-}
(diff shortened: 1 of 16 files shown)


nfrisby:
Tasks so far planned before merge:

- [ ] `scriv` up a changelog entry
- [ ] Add tests that show that these generators are finding the corner cases (for small parameter values)
- [ ] How much simpler is it to store one byte per bit instead of explicitly managing the chunking?
- [ ] Window offsets without sizes form a `Category`. So does the preceding half of `Packs`. Is there a refactoring that can leverage that?
- [ ] Can we simplify the whole `Polarize` interface so that the types are simpler and the combinators are easier to use?
- [ ] I think there's some fat to trim w.r.t the `Var` type

nfrisby:
Once PR #4439 is merged (or you cherry-pick it for now), you can run these new tests with:

```
cabal run ouroboros-consensus-test:test:test-infra -- -pChainGen --quickcheck-tests=10000
00
```

nfrisby:
I just added commit 2b8dd63cb9909624a9a9f2e44037f90fa9aa5fee, which adds a test requiring that the honest chain generator does not add any unnecessary density: `prop_honestChainMutation`

nfrisby:
I've pushed some commits that add `prop_adversarialChainMutation`, along with some commits that found a couple bugs that test caught 🙌

Caveats:

- I don't understand why the limitations I've added on the generators for this mutation test seem to work. I suspect the magic numbers I've chosen (1/3 and 4/3) are fragile.

- I've disabled mutation of Scg for now, because the limitations that work for Kcp and Delta don't work for Scg. I'll repeat repeat my effort from today tomorrow.

nfrisby:
I simplified the `POL` class. The biggest simplification: remove `polarize`, which now means that no `W` is every actually complemented; that was the most ambiguous aspect.

While I was at it, I split out the new `.Word` module, removed some dead code, and improved some key comments.

nfrisby:
I removed the packing, split `.Tests`, and otherwise polished.

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,638 @@
...
+    nn = toEnum n :: Double
+
+    -- the non-normalized stationary distribution
+    --
+    -- 1) See
+    -- https://stephens999.github.io/fiveMinuteStats/markov_chains_discrete_stationary_dist.html
+    --
+    -- 2) The stationary distribution is the eigenvector with eigenvalue 1.
+    --
+    -- 3) We want the /left/ eigenvector (TODO is this fundamental to transition matrices?).
```
amesgen, line 423: > is this fundamental to transition matrices

It is purely a consequence of the convention that the entry $P_{ij}$ in the $i$-th row and the $j$-the column of the transition matrix $P$ is the probability to transition to state $j$ if we are in state $i$ (instead of the opposite), and the way matrix multiplication is defined.

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
```diff
@@ -0,0 +1,133 @@
...
+    elemseq (V_S v) x y               = VG.elemseq v (coerce x) y
+instance V.Unbox S
+
+-----
+
+genAsc :: R.RandomGen g => Asc -> g -> (S, g)
+genAsc asc g =
+    bool `seq` (S bool, g')
+  where
+    (q, g') = R.random g   -- note 0 <= q <- 1
```
amesgen, line 78:
```suggestion
    (q, g') = R.random g   -- note 0 <= q <= 1
```

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,638 @@
...
+    -- a brute-force way to (hopefully) find the stationary distribution
+    -- (recall the way in which 'stimes' is optimized) sd :: Mat.Vector Double
+    -- sd = (Mat.! 0) $ stimes (2^(10 :: Int) :: Int) p
+-}
+
+    -- Transition matrix of an Markov Chain where each state is the
+    -- (idealized?) number of active slots in an EHCG window.
+    --
+    -- (This is a /stochastic matrix/ because every /row/ sums to 1. And every
+    -- state is reachable from every other, since @0 < asc@.)
```
amesgen, line 460: The second property is called "irreducibility", but it is not part of the definition of a stochastic matrix.

Rather, a Markov chain being irreducible and [aperiodic](https://en.wikipedia.org/wiki/Discrete-time_Markov_chain#Periodicity) (trivially satisfied, as actually we have non-zero probability of staying at a state for all states) is a sufficient criterion for there being a unique stationary distribution.

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,638 @@
...
+
+{-
+
+Derivation of the transition matrix used in 'stationaryWindowSize'.
+The x/d summand is the chance we're sliding the window off of an active slot.
+The (d - x)/d summand is the chance we're sliding the window off of an empty slot.
+
+One reflecting barrier at n.
+
+    P( n -> n     ) = (n/d) * 1         +   ((d - n)/d) * 1
```
amesgen, line 513: Something is wrong here, probably replace the last $1$ by $1-f$.

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
```diff
@@ -0,0 +1,638 @@
...
+    -- that somehow a slot's value from /before/ some window is unexpectedly
+    -- influencing the value of the first slot in that window in a way that is
+    -- not accounted for by the number of active slots in the preceding window,
+    -- ie that the Markov Property does not actually hold. I think the seed for
+    -- this doubt was Esgen's earlier insight that a chain with /exactly/ @n@
+    -- actives in every @d@ slot window is periodic. In that case, if there are
+    -- more than @d@ contiguous windows that each have exactly @n@ actives,
+    -- then there is perfect auto-correlation throughout those windows, and it
+    -- spans farther than one window. That's our suspicion for the kind of
+    -- counterexample of the Markov Property of this transition matrix.
```
amesgen, line 479: This should be reworded; any stochastic matrix (such as the one below) corresponds to a Markov chain, but the question is whether that actually models what we are interested in (i.e. whether it is "memoryless").

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Some.hs
```diff
@@ -0,0 +1,114 @@
...
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+{-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
+
+-- | Utilities that make it easier to declare types that merely existential
+-- quantify some type arguments, and actually for any other type that is not
+-- amenable to @deriving@
```
nfrisby, line 12: TODO explain that Hackage's `some` package doesn't work for us here since our data types are quantifying multiple variables at a time and we reject the encoding overheading that would be necessary to use `some` in that context

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
nfrisby: Fixed 👍

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
nfrisby: Fixed 👍

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
```diff
@@ -0,0 +1,133 @@
...
+import qualified System.Random.Stateful as R
+import qualified Test.Ouroboros.Consensus.ChainGenerator.Counting as C
+import           Test.Ouroboros.Consensus.ChainGenerator.Params (Asc, ascVal)
+import qualified Test.QuickCheck as QC
+
+-- | The activeness of some slot
+newtype S = S Bool
+  deriving (QC.Arbitrary, Eq, Ord, Read, Show)
+
+-- these instances adapted from https://github.com/minoki/unboxing-vector/blob/3a152014b9660ef1e2885d6b9c66423064223f63/test/Foo.hs#L36-L63
```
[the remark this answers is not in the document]
nfrisby: I added a TODO comment pointing this out for someone to handle whenever its relevant and our `vector` dep has been bumped. This alone doesn't justify mucking with versions, IMO.

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Honest.hs
nfrisby: Thanks, I added that vocab and stated that implication. 👍

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Slot.hs
nfrisby: Fixed 👍

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Some.hs
```diff
@@ -0,0 +1,114 @@
...
+import qualified Text.ParserCombinators.ReadPrec as Read
+import qualified Text.Read.Lex as Read
+
+-----
+
+type family AbsError (s :: Symbol) (a :: Type) :: Void where AbsError s a = TE.TypeError (TE.Text "You have accidentaly applied `" TE.:<>: TE.Text s TE.:<>: TE.Text "' to a non-concrete type: " TE.:<>: TE.ShowType a)
+
+type family NoFun (s :: Symbol) (a :: Type) (absError :: Void) :: Constraint where
+    NoFun s (a -> b) abs = TE.TypeError (TE.Text "You have accidentaly applied `" TE.:<>: TE.Text s TE.:<>: TE.Text "' to a function type: " TE.:<>: TE.ShowType (a -> b))
+    NoFun s t        abs = ()
```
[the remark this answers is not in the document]
nfrisby: Fixed 👍

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/RaceIterator.hs
```diff
@@ -0,0 +1,136 @@
...
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+{-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
+
+module Test.Ouroboros.Consensus.ChainGenerator.RaceIterator (
+  RaceLbl,
+  Race (Race, UnsafeRace),
```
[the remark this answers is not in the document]
nfrisby: I  intentionally export `Unsafe*` in general; it's the user's fault for using it without understanding it.

I think it's worse to make it more difficult for someone to explore some additional behavior that would otherwise necessitate patching our code.

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Counting.hs
```diff
@@ -0,0 +1,336 @@
...
+  Total,
+  forRangeM_,
+  lastIndex,
+  range, 
+  uniformIndex,
+  -- * windows
+  Lbl (Lbl),
+  SomeWindow (SomeWindow),
+  Win,
+  Contains (Contains, UnsafeContains),
```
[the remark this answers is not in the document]
nfrisby: I  intentionally export `Unsafe*` in general; it's the user's fault for using it without understanding it.

I think it's worse to make it more difficult for someone to explore some additional behavior that would otherwise necessitate patching our code.

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Some.hs
nfrisby: Done.

ouroboros-consensus-test/src/Test/Ouroboros/Consensus/ChainGenerator/Params.hs
```diff
@@ -0,0 +1,94 @@
...
+-- | The /active slot coefficient/
+--
+-- Algebraic data type representing a binary fraction greater than 0 and less than 1.
+data Asc =
+    Half   -- ^ @0.5@ in decimal, ie @0.1@ in binary
+  |
+    HalfOf Asc   -- ^ @x \/ 2@ in decimal, ie @x \/ 10@ in binary
+  |
+    HalfOfOnePlus Asc   -- ^ @(x + 1) \/ 2@ in decimal, ie @0.1 + (x \/ 10)@ in binary
+  deriving (Eq, Read, Show)
```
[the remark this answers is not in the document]
nfrisby: Fixed 👍

(Back when I was generating words instead of just bits, the algebraic `Asc` could capture more precise probabilities than could `Double`. (Not that we'd need that, but I try to favor exact types when they're not too unwieldy.)

When I remove the bitpacking, I dropped that generator, and now it just routes through `Double`. So simplifying here does make all the sense.)

PR #4442, opened 2023-03-09
Format cabal files with `cabal format`

Jasagredo:
# Description

Apply `cabal format` to all the consensus `.cabal` files. Also add it to CI.

### DIFF
--- a/ouroboros-consensus/ouroboros-consensus.cabal
+++ b/ouroboros-consensus/ouroboros-consensus.cabal
@@ -1,357 +1,320 @@
-cabal-version: 3.0
-
-name:                   ouroboros-consensus
-version:                0.3.1.0
-synopsis:               Consensus layer for the Ouroboros blockchain protocol
-description:            Consensus layer for the Ouroboros blockchain protocol.
-license:                Apache-2.0
-license-files:          LICENSE
-                        NOTICE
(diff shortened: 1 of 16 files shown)


ouroboros-consensus/ouroboros-consensus.cabal
```diff
@@ -1,357 +1,320 @@
...
+        TemplateHaskell TupleSections TypeApplications TypeFamilies
+        TypeFamilyDependencies TypeInType TypeOperators
+        UndecidableInstances UndecidableSuperClasses ViewPatterns
 
-                       -- Strict wrapper around SOP
-                       Data.SOP.Strict
-  if os(windows)
-     exposed-modules:
-                       -- TODO Delete this module once all our projects have upgraded to at least
-                       -- ghc-8.10.7, but not before
```
Jasagredo, line 257: We are loosing this comment!

nfrisby (changes requested):
I agree with other reviewers that PR #4443 seems ultimately preferable to this, simply because that formatter doesn't have a reputation for being lossy. Using the "official" one would be nice, but losing eg comments or other non-syntactic structure seems too painful.

(And at least today, `phadej` is a notable Cabal expert, so `cabal-fmt` is more like 2nd-party software rather than 3rd-party...)

Jasagredo:
Close in favor of #4443

PR #4443, opened 2023-03-09
Format cabal files with `cabal-fmt`

Jasagredo:
# Description

Apply `cabal-fmt -i` to all the consensus `.cabal` files. Also add it to CI.


amesgen (approved):
We might want to mention it in `CONTRIBUTING.md`, but the error message `cabal-fmt not found` should already be rather indicative.

PR #4467, opened 2023-03-23
Fix NonP2P to P2P interop problem

### LINKED ISSUE #4465: NonP2P relays fail to keep an established connection to a P2P node
After `1.35.6` release of `cardano-node`, we received some reports where BPs weren't able to keep an established connection to a P2P node.

@karknu was able to reproduce the issue and capture a `pcap` trace, this was interpreted to create the following diagram (where BP is assumed to be a non P2P node):

```mermaid
sequenceDiagram
    BP->>P2P: connect (producer) - A
    P2P->>BP: connect (local root peer) - B
    note right of P2P: P2P is using a fixed PORT
    BP->>P2P: (timeout - tear down by BP - A) [BP no outgoing diffussion]
    note over BP,P2P: BP belives existing connection - no BP outbound connection attempts
    P2P->>BP: after 79 sec: P2P tears down due to non-respose of keep alive protocol[observed]<br>Suggests BP response to timer teardown "captured B" first connection in some way
    note over BP,P2P: time reference T
    loop until BP 'bad peer' timer expires - trace retries at T+10 and T+18
      P2P--xBP: attempts new connection - TCP connections but<br>BP resets before process handshake (assumed still in 'sin bin')
    end
    note over BP,P2P: some seconds later (21s after T in trace)
    BP->>P2P:connect (producer) - A [success]
```

We believe this is a bug in the legacy non-p2p code, which is triggered by P2P nodes which bind to their advertised addressed for outbound connections.  The non-p2p component which tracks the state of inbound & outbound connections can be confused by this and depending on which nodes connects first (e.g. `P2P -> NonP2P` or `NonP2P -> P2P`) this can either result in minutes long or hour long (or more) problems for a non-p2p node to establish a connection to a p2p node.

karknu:
# Description
P2P nodes uses the server port number for outbound connections. This can confuse NonP2P nodes into thinking that they already have a connection to them.

This PR changes the NonP2P connection table to take connection direction into account, similar to @dcoutts  patches in https://github.com/input-output-hk/ouroboros-network/compare/dcoutts/old-connection-table .
The P2P nodes reuse of the server port also means that incoming connections from a P2P node can be refused by a NonP2P node if the peer is "suspended". This suspension never affected NonP2P nodes since they reconnect from a new ephemeral port each time. This PR lets NonP2P nodes differentiate between if it should refuse to accept new connections from a peer or if it should avoid creating new connections to a peer.

Fixes #4465 .

# Checklist

- Branch
    - [ ] Any changes in the Consensus API (every exposed function, type or module) that has changed its name, has been deleted, has been moved, or altered in some other significant way must leave behind a `DEPRECATED` warning that notifies downstream consumers. If deprecating a whole module, remember to add it to `./scripts/ci/check-stylish.sh` as otherwise `stylish-haskell` would un-deprecate it.
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-network-framework/CHANGELOG.md
+++ b/ouroboros-network-framework/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## next version
 
+### Non breaking
+
+- Fix interop problems between NonP2P and P2P nodes (PR #4465)
+
 ### Breaking
 
 - Removed `TrImpossibleConnection` trace (PR #4385)
--- a/ouroboros-network-framework/src/Ouroboros/Network/Server/ConnectionTable.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Server/ConnectionTable.hs
@@ -8,6 +8,7 @@
 module Ouroboros.Network.Server.ConnectionTable
   ( ConnectionTable
   , ConnectionTableRef (..)
+  , ConnectionDirection (..)
   , ValencyCounter
   , newConnectionTableSTM
   , newConnectionTable
@@ -40,7 +41,7 @@ import           Text.Printf
 -- through this structure.
 --
 data ConnectionTable m addr = ConnectionTable {
-    ctTable     :: StrictTVar m (M.Map addr (ConnectionTableEntry m addr))
+    ctTable     :: StrictTVar m (M.Map (addr,ConnectionDirection) (ConnectionTableEntry m addr))
   , ctLastRefId :: StrictTVar m Int
   }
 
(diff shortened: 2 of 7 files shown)


coot:
One thing, could you add an entry to `ouroboros-network-framework` and `ouroboros-network`,  `CHANGELOG` files e.g.

```
## next version

### Non breaking

* Fixed ...
```

coot:
bors merge

PR #4474, opened 2023-03-28
Enable benchmark builds in GitHub Actions by default using `benchmarks: True`

jorisdral:
# Description

The Consensus Mempool benchmark is the only benchmark executable currently on the main branch, but benchmark builds are disabled by default by cabal. As such, our GitHub Actions workflow also does not build benchmark executables. This PR enables benchmark building in GitHub Actions by default, by adding `benchmarks: True` in the relevant `cabal.project.local` file. 

# Checklist

- Branch
    - [ ] Any changes in the Consensus API (every exposed function, type or module) that has changed its name, has been deleted, has been moved, or altered in some other significant way must leave behind a `DEPRECATED` warning that notifies downstream consumers. If deprecating a whole module, remember to add it to `./scripts/ci/check-stylish.sh` as otherwise `stylish-haskell` would un-deprecate it.
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/.github/workflows/cabal.project.local.Windows
+++ b/.github/workflows/cabal.project.local.Windows
@@ -2,6 +2,7 @@ max-backjumps: 5000
 reorder-goals: True
 ignore-project: False
 tests: True
+benchmarks: True
 
 program-options
   ghc-options: -Werror
--- a/cabal.project
+++ b/cabal.project
@@ -49,6 +49,7 @@ packages: ./cardano-ping
           ./cardano-client
 
 tests: True
+benchmarks: True
 
 package cardano-ping
   flags: +asserts


Jasagredo:
I don't think we need the one in `cabal.project.local`, the one in `cabal.project` will be used which is fine

jorisdral:
> I don't think we need the one in cabal.project.local, the one in cabal.project will be used which is fine

Removing the field from `cabal.project` might have the unfortunate side effect that CI does not build the benchmarks anymore, so might it be better to duplicate it to prevent that case?

Jasagredo (approved):
I think it is fine as is, what you said about duplicating it sounds good. Thanks!

PR #4494, opened 2023-04-05
Fix issue #4491

### LINKED ISSUE #4491: Issue with local roots which use more than one domain name
The node fails to maintain a hot connection to all configured local root peers, if more than 1 DNS name is used in a single group.

Splitting the relays up into different access points (groups) works as a workaround.

bolt12:
Updates the result local root groups TVar with up to date information instead of static one.

Also adds & fixes tests to check for this particular case.

Closes #4491

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
@@ -158,7 +158,7 @@ localRootPeersProvider tracer
       -- Launch DomainAddress monitoring threads and wait for threads to error
       -- or for local configuration changes.
       domainsGroups' <-
-        withAsyncAll (monitorDomain rr rootPeersGroups `map` domains) $ \as -> do
+        withAsyncAll (monitorDomain rr `map` domains) $ \as -> do
           res <- atomically $
                   -- wait until any of the monitoring threads errors
                   ((\(a, res) ->
@@ -212,12 +212,9 @@ localRootPeersProvider tracer
 
     monitorDomain
       :: Resource m (DNSorIOError exception) resolver
-      -> Seq (Int, Map peerAddr PeerAdvertise)
-      -- ^ Static configuration, this always comes from the source
-      -- STM transaction 'readDomainGroups'.
       -> (Int, DomainAccessPoint, PeerAdvertise)
       -> m Void
-    monitorDomain rr0 rpgStatic (index, domain, advertisePeer) =
+    monitorDomain rr0 (index, domain, advertisePeer) =
         go rr0 0
       where
         go :: Resource m (DNSorIOError exception) resolver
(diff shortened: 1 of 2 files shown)


ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -707,6 +728,18 @@ prop_local_updatesDomainsCorrectly mockRoots@(MockRoots lrp _ _ _)
...
+  prop_local_updatesDomainsCorrectly
+    (MockRoots [ (2, Map.fromList [(RelayAccessDomain "dns1" 00001, DoAdvertisePeer)
+               , (RelayAccessDomain "dns2" 00000, DoAdvertisePeer)])
+               ]
+               (singletonScript (Map.fromList [ ("dns1", [("0.0.0.0", 0)])
+                                              , ("dns2", [("0.0.0.1", 0)])]))
+               Map.empty
+               (singletonScript Map.empty))
+    (singletonScript (DNSTimeout 1000))
+    (singletonScript (DNSLookupDelay 0))
```
coot, line 742: Is this covered by `prop_local_updatesDomainsCorrectly`? If so, then do we need this?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -246,7 +243,7 @@ localRootPeersProvider tracer
                     -- configured IPs, that's why we get the entry from the
                     -- statically configured rootPeersGroups list.
                     --
-                let (target, entry)  = rpgStatic `Seq.index` index
+                let (target, entry)  = rootPeersGroups `Seq.index` index
```
coot, line 246: What about the comment above?

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: ugh I responded via email and it ended up in a different place. Actual reply:

It is now, I just used it to make sure we were covering the case that would
lead to the bug. I can remove it!

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: Needs to be removed!

bolt12: Done!

ouroboros-network/test/Test/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
bolt12: Done!

bolt12:
bors merge

coot:
bors merge

PR #4502, opened 2023-04-12
Ensure CHaP builds with the released version of consensus

Jasagredo:
# Description

Ensures CHaP will be able to build the newly released version before telling the user to push the updated version.

### DIFF
--- a/scripts/check-consensus-builds-with-chap.sh
+++ b/scripts/check-consensus-builds-with-chap.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
+# remove network packages from the list of local packages
+
+sed 's/\.\/.*network.*//g' cabal.project \
+    | sed 's/\.\/monoidal.*//g' \
+    | sed 's/\.\/ntp.*//g' \
+    | sed 's/\.\/cardano.*//g' \
+    | sed 's/packages:.*/packages:/g' \
+    > cabal.project.consensus
+
+cabal build --project-file cabal.project.consensus --builddir dist-chap all 2>&1 | ts "> "
+
+rm cabal.project.consensus
--- a/scripts/check-consensus-builds-with-chap.sh
+++ b/scripts/check-consensus-builds-with-chap.sh
@@ -2,15 +2,106 @@
 
 set -o pipefail
 
-# remove network packages from the list of local packages
+function last_version {
+  grep "<a id=" $1/CHANGELOG.md  | cut -d\' -f2 | cut -d- -f2 | head -n1
+}
 
-sed 's/\.\/.*network.*//g' cabal.project \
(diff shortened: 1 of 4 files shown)


scripts/check-consensus-builds-with-chap.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
+# remove network packages from the list of local packages
+
+sed 's/\.\/.*network.*//g' cabal.project \
```
nfrisby, line 7: This `sed` line seems practical, but it also seems like something that will often cause confusion.

For example, I think a better solution would be:

- enumerate the packages we want to exclude
- enumerate the packages we want to include
- fail if the packages we find in the actual file are not equal the union of the above
- generate the new file by removing the ones we want to exclude

scripts/consensus-release.sh
```diff
@@ -258,6 +258,16 @@ else
...
-echo "Succesfully created release. You should now inspect the current branch (${branch%?}) and if everything looks right, push it to GitHub, open a PR and get it merged!"
+echo "Checking that CHaP would be able to build these packages (this will take a while!)"
+./scripts/check-consensus-builds-with-chap.sh
+
+if [[ -z $? ]]; then
+  echo "Succesfully created release. You should now inspect the current branch (${branch%?}) and if everything looks right, push it to GitHub, open a PR and get it merged!"
+else
+  echo "Your release seem to depend on changes that are not visible in CHaP/Hackage"
+  echo "Maybe you added a function to a separate package or are using a package that is not in those indices?"
+  echo "I created the branch ${branch%?} but it is probably wrong, so you should delete it, check the error above and re-run this script. I will leave the branch just in case."
```
nfrisby, line 272: Is this line a copy-paste error?

Jasagredo: I don't know specifically what you mean. But my intention was to get that message on the screen, having the script talking as "I" :smile:

scripts/check-consensus-builds-with-chap.sh
Jasagredo: Reasonable, let me rework it

amesgen: Another alternative: move the `packages:` part to two standalone files (one for network, one for consensus) and [import them](https://cabal.readthedocs.io/en/latest/cabal-project.html#conditionals-and-imports), then you can replace the network file with an empty file here temporarily.

scripts/consensus-release.sh
nfrisby: Ah, OK. That seemed odd to me in the PR diff, but I don't think it would confuse me on the command-line output.

dnadales (approved):

PR #4522, opened 2023-04-24
Migrate `ouroboros-consensus` out of this repository

Jasagredo:
# Description

The counterpart of https://github.com/input-output-hk/ouroboros-consensus/pull/34


nix/ouroboros-network.nix
```diff
@@ -57,14 +57,6 @@ let
...
-          packages.ouroboros-consensus-byron-test.components.tests.test.testFlags =
-            lib.mkForce [ "--no-create" ];
-          packages.ouroboros-consensus-shelley-test.components.tests.test.testFlags =
-            lib.mkForce [ "--no-create" ];
-          packages.ouroboros-consensus-cardano-test.components.tests.test.testFlags =
-            lib.mkForce [ "--no-create" ];
-
           packages.cardano-crypto-praos.components.library.pkgconfig =
             lib.mkForce [ [ pkgs.libsodium-vrf ] ];
           packages.cardano-crypto-class.components.library.pkgconfig =
```
coot, line 62: We don't need `cardano-crpyto-praos` and `cardano-crypto-class`; Network doesn't relay on any crypto stuff right now.

.github/workflows/checks.yml
```diff
@@ -1,56 +0,0 @@
...
-
-    defaults:
-      run:
-        shell: bash
-
-    steps:
-
-    - uses: actions/checkout@v3
-
-    - run: ./scripts/ci/check-git-dependencies.sh
```
coot, line 31: This check should stay.

Jasagredo: Done

nix/ouroboros-network.nix
Jasagredo: Done

Jasagredo: Actually this is not true. `cardano-client` -> `ouroboros-consensus-diffusion` -> `ouroboros-consensus` -> crypto

Jasagredo: It seems nix is able to do without this.

Jasagredo: I meant that `nix` is able to build the shell and all packages without these lines

coot (approved):
LGTM

PR #4525, opened 2023-04-25
Introduce PolyKinds to Hash related entities

Jasagredo:
# Description

UTxO-HD will bring the need to not only have kinds `*` on some entities. This benign PR brings these changes in.


coot (approved):
LGTM

PR #4528, opened 2023-04-26
Bring back the separate step for deps in GHA

Jasagredo:
# Description

This step was deleted because of the dependency on `ouroboros-consensus` but as `cardano-client` no longer depends on it, we should be able to bring it back.


coot (approved):
Thanks!

coot:
bors merge

PR #4551, opened 2023-05-12
Use iog-contra-tracer instead of contra-tracer

erikd:
There are two incompatible versions of the later, one on Hackage and one in CHaP. Switching to the former fixes this.


coot (approved):
LGTM

coot:
I think you need to run to update `nix/sources.json`:
```
niv update hackage.nix
niv update CHaP
```

erikd:
This is a bad idea.

PR #4556, opened 2023-05-15
ouroboros-network-protocols: added bounds for `ouroboros-network-api`

coot:


bolt12 (approved):

coot:
The bound was updated in a6c72ab9e38.

PR #4558, opened 2023-05-16
build-with-chap.sh: comment on the PR

coot:
This makes it possible to notice that one did run the script when doing
a release.


bolt12 (approved):
Very cool!

PR #4561, opened 2023-05-16
Make it build with ghc-9.6 and update CI

erikd:


cabal.project
```diff
@@ -49,3 +49,8 @@ package network-mux
 
 package ouroboros-network
   flags: +asserts +cddl
+
+if impl(ghc >= 9.6)
+  allow-newer:
+    , *:base
+    , protolude:*
```
coot, line 56: We should create an issue in the protolude repo to support `ghc-9.6` (its `master` branch doesn't support it).

coot (approved):
LGTM

cabal.project
erikd: There already is one.

Issue #4563, opened 2023-05-19
ActivePeers on Node 8.0.0 (P2P-Mode) getting lower over time


gitmachtl:
Upgraded normally from cardano-node 1.35.7 to cardano-node 8.0.0 on one of my Mainnet-Relais.

The P2P-Settings in the config.json are:
``` json
"EnableP2P": true,
  "TargetNumberOfActivePeers": 30,
  "TargetNumberOfEstablishedPeers": 50,
  "TargetNumberOfKnownPeers": 100,
  "TargetNumberOfRootPeers": 100,
  "TraceAcceptPolicy": true,
  "TraceInboundGovernor": true,
  "TraceLedgerPeers": true,
  "TraceLocalRootPeers": true,
  "TraceConnectionManager": true,
```

After a fresh start, the node gets 30 active peers, all seems normal. But after a few minutes and over the following hours, the number of active peers is reduced one by one. Sent some logs to @coot and @karknu, looks like the node did not start to demote the peers in case an active one gets decreased.

From the logs:
```
2023-05-16T06:09:05.20Z Normal Churn starts
2023-05-16T06:09:05.20Z targetActivePeers is decreased from 30 to 24
2023-05-16T06:09:06.20Z targetActivePeers increased to 30 , no demotion of hot peers has started
2023-05-16T06:09:07.21Z targetEstablishedPeers is decreased to from 50 to 40
                        targetKnownPeers is decreased from 100 to 80
            targetRootPeers is decreased from 100 to 80
2023-05-16T06:09:07.21Z 20 warm peers are picked for demotion, actualEstablished=100 and targetEstablished=40
2023-05-16T06:11:08.20Z Churn Done
2023-05-16T06:11:08.21Z targetEstablished is increased to 50
                        targetKnownPeers is increased to 100
            targetRootPeers is increased to 100
2023-05-16T06:11:08.21Z 20 cold peers are picked for promotion
```

> Karl: The governor has 1s to detect the decrease in targetActivePeers and start to demote them.
> It is apparently busy with other things so the churn doesn't happen.

The machine is a decent one with 12 cores, also did some tests with only the node running. Its chillin all the time.
Test was done with a self-compiled cardano-node binary and also with the pre-built binary from the 8.0.0 release.

Will repeat a test with only 20 targetActivePeers and also Karl sent me a special commit to try out. Such an issue did not happen with node 1.35.7. For a test i reverted the node back to 1.35.7 and the connections were stable. 

[...]

PR #4571, opened 2023-05-25
Changed how DNS Lookup for local root peers works

bolt12:
# Description

Part one of fix for #4563 . This fixes the way we store DNS lookup results.


coot:
Please include `CHANGELOG` entry, not that that your change requires a major bump of `ouroboros-network`.

Issue #4585, opened 2023-06-06
Estimate the number of file descriptors used by ouroboros-network


coot:
This issue is related to:
* input-output-hk/cardano-node#305
* input-output-hk/ouroboros-consensus#20

coot:
The network layer is using:

* 1 socket per upstream / downstream peer using `node-to-node` protocol.  In P2P mode the socket is shared for upstream / downstream connections for the same peer, in non P2P mode 2 separate sockets are used;
* 1 socket per accepted `node-to-client` protocol connection;
* at most 2 listening sockets (one per `IPv4` / `IPv6`);
* one listening Unix domain socket;
* the network layer occasionally reads `/etc/resolv.conf` file.

Note:
The node will re-read the topology file on `SIGHUP` signal.

coot:
Updated limits:

* 1 socket per upstream / downstream peer using node-to-node protocol. In P2P mode the socket is shared for upstream / downstream connections for the same peer, in non P2P mode 2 separate sockets are used;
* 1 socket per accepted node-to-client protocol connection;
* at most 2 listening sockets (one per IPv4 / IPv6);
* one listening Unix domain socket;
* the network layer occasionally reads /etc/resolv.conf file.
* at most 10 file descriptors for dns name resolutions (as of #4596).

The upper bound is `n + m + 15` where

* `n` - number of upstream peers
* `m` - number of downstream peers

PR #4596, opened 2023-06-14
dns: limit concurrency when resolving dns names

coot:
We use two semaphores.  One to limit concurrency of dns name resolution of local root peers to at most `2`; and another one to limit concurrency for public & ledger peers to at most `8`.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
@@ -14,6 +14,9 @@ module Ouroboros.Network.PeerSelection.RootPeersDNS
     -- ** DNSActions IO
   , ioDNSActions
   , LookupReqs (..)
+    -- * DNS semaphore
+  , DNSSemaphore
+  , newLocalAndPublicRootDNSSemaphore
     -- * DNS based provider for local root peers
   , localRootPeersProvider
   , DomainAccessPoint (..)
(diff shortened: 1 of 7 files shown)


coot:
Affects: #4585.

karknu:
LGTM.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -85,6 +89,36 @@ data TraceLocalRootPeers peerAddr exception =
...
+maxDNSLocalRootConcurrency = 2
+
+-- | A semaphore used to limit concurrency of dns names resolution.
+--
+newtype DNSSemaphore m = DNSSemaphore (TSem m)
+
+-- | Create a `DNSSemaphore` for root and ledger peers.
+--
+newDNSSemaphore :: MonadSTM m => m (DNSSemaphore m)
+newDNSSemaphore = DNSSemaphore <$> atomically (newTSem maxDNSConcurrency)
```
bolt12, line 110: Maybe be explicit in the name as well: `newDNSPublicRootSemaphore`

bolt12 (changes requested):
LGTM just small note

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
coot: It's used for public roots and ledger peers, so Irenamed it to `newPublicAndLedgerRootDNSSemaphore` (pretty long name though).

bolt12 (approved):

PR #4603, opened 2023-06-20
Fixed handshake query delay

coot:
After handshake query the responder side is supposed to close the
connection after 20s.  We used wrong `threadDelay` function, and it
waited only 20μs instead.

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake.hs
@@ -15,6 +15,7 @@ module Ouroboros.Network.Protocol.Handshake
   , HandshakeResult (..)
   , RefuseReason (..)
   , Accept (..)
+  , handshake_QUERY_SHUTDOWN_DELAY
   ) where
 
 import           Control.Monad.Class.MonadAsync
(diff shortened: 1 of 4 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake.hs
```diff
@@ -182,3 +183,9 @@ runHandshakeServer bearer
           haTimeLimits
           (fromChannel (muxBearerAsChannel bearer handshakeProtocolNum ResponderDir))
           (handshakeServerPeer haVersionDataCodec haAcceptVersion haQueryVersion versions))
+
+-- | A 20s delay after query result was send back, before we close the
+-- connection.  After that delay we close the connection.
+--
+handshake_QUERY_SHUTDOWN_DELAY :: DiffTime
+handshake_QUERY_SHUTDOWN_DELAY = 20
```
bolt12, line 191: Shouldn't this be `20_000_000` in order to be 20s ?

bolt12 (changes requested):

ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake.hs
bolt12: NVM just noticed you changed the import

bolt12 (approved):

PR #4618, opened 2023-07-10
Makes KnownPeers.insert idempotent

### LINKED ISSUE #4616: Make sure we always use `combinePeerInformation` when inserting PeerSharing information
Currently there's this function: https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/PeerSelection/KnownPeers.hs?plain=1#L219 which is used when [we promote a peer to warm](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs#L306).

In that case, we learn about a potential new vale for `PeerSharing` and we are overwriting it ignoring its `PeerAdvertise` value. We should really use `insert` instead and make sure it is idempotent and using `combinePeerSharingInformation` function

* [ ] include a test that `KnonwPeers.insert` is idempotent

bolt12:
# Description

Closes #4616 

- Make sure we always insert the `PeerSharing` value using `combinePeerInformation` function
- Add test for `KnownPeers.insert` to check if it is idempotent

### DIFF
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Breaking changes
 
+- Removes `updatePeerSharing` from `KnownPeers` module API
+
 ### Non-breaking changes
 
 ## 0.9.0.0 -- 2023-08-09
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
@@ -35,10 +35,6 @@ import           Ouroboros.Network.PeerSelection.Governor.ActivePeers
                      (jobDemoteActivePeer)
 import           Ouroboros.Network.PeerSelection.Governor.Types hiding
                      (PeerSelectionCounters (..))
-import           Ouroboros.Network.PeerSelection.LedgerPeers (IsLedgerPeer (..))
-import           Ouroboros.Network.PeerSelection.PeerAdvertise
-                     (PeerAdvertise (..))
-import           Ouroboros.Network.PeerSelection.PeerSharing (PeerSharing (..))
 import qualified Ouroboros.Network.PeerSelection.State.EstablishedPeers as EstablishedPeers
 import qualified Ouroboros.Network.PeerSelection.State.KnownPeers as KnownPeers
 import qualified Ouroboros.Network.PeerSelection.State.LocalRootPeers as LocalRootPeers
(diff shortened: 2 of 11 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/KnownPeers.hs
```diff
@@ -194,12 +195,15 @@ insert peeraddrs
...
+          kPS' = combinePeerInformation kPS kPA
+       in KnownPeerInfo {
         knownPeerFailCount = 0
       , knownPeerTepid     = False
-      , knownPeerSharing   = ps
-      , knownPeerAdvertise = pa
-      , knownLedgerPeer    = lp
+      , knownPeerSharing   = kPS'
+      , knownPeerAdvertise = kPA
+      , knownLedgerPeer    = fromMaybe IsNotLedgerPeer lp
```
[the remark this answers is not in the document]
bolt12: Done!

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -122,7 +118,7 @@ inboundPeers PeerSelectionActions{
     return $ \_ ->
       let -- If peer happens to already be present in the Known Peer set
           -- 'insert' is going to do its due diligence before adding.
-          newEntry    = Map.singleton addr (ps, DoAdvertisePeer, IsNotLedgerPeer)
+          newEntry    = Map.singleton addr (Just ps, Nothing, Nothing)
```
coot, line 121: Actually when the connection manager writes to the `infoChannel` the connection is already negotiated.  So we should know the advertise flag by now.

ouroboros-network/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Breaking changes
 
+- Removes `updatePeerSharing` from `KnownPeers` module API
```
coot, line 7: We need to define what is public API, and what isn't (not export all the internals as we do right now). `KnownPeers` is an internal API, nobody needs to know about it, so this isn't really a breaking change.

coot (approved):
LGTM

ouroboros-network/CHANGELOG.md
coot: https://github.com/input-output-hk/ouroboros-network/issues/4646

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
coot: https://github.com/input-output-hk/ouroboros-network/issues/4647

PR #4619, opened 2023-07-10
Adds ledger peers in the diffusion simulation

### LINKED ISSUE #4611: Introduce ledger peers in the diffusion simulation.
Currently we don't use ledger peers, [ref](https://github.com/input-output-hk/ouroboros-network/blob/91072496673174213722fd442d8b56f3b1a0578e/ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs#L880-L882).

bolt12:
Closes #4611

### DIFF
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -8,8 +8,9 @@
 
 * Less aggresive churning of established and known peers.
 * Add peer sharing to wireshark dissector.
+* Adds ledger peers to diffusion simulation
 
-- Fix diffusion tests
+* Fix diffusion tests
 
 ## 0.9.1.0 -- 2023-08-22
 
--- a/ouroboros-network/sim-tests/Test/Ouroboros/Network/LedgerPeers.hs
+++ b/ouroboros-network/sim-tests/Test/Ouroboros/Network/LedgerPeers.hs
@@ -90,19 +90,27 @@ instance Arbitrary StakePool where
                                        (NonEmpty.toList spRelay)
       ]
 
-newtype LedgerPools = LedgerPools [(PoolStake, NonEmpty RelayAccessPoint)]
+newtype LedgerPools =
+  LedgerPools { getLedgerPools :: [(PoolStake, NonEmpty RelayAccessPoint)] }
   deriving Show
 
 instance Arbitrary LedgerPools where
     arbitrary = LedgerPools . calculateRelativeStake <$> arbitrary
 
-      where
-        calculateRelativeStake :: [StakePool]
-                               -> [(PoolStake, NonEmpty RelayAccessPoint)]
-        calculateRelativeStake sps =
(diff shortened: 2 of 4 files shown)


ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -910,9 +917,13 @@ diffusionSimulation
...
-              , NodeKernel.iLedgerPeersConsensusInterface
-                                        = LedgerPeersConsensusInterface
-                                        $ \_ -> return Nothing
+              , NodeKernel.iLedgerPeersConsensusInterface =
+                  LedgerPeersConsensusInterface $
+                    \_ -> return
+                        $ Just
+                        $ Map.elems
+                        $ accPoolStake
+                        $ getLedgerPools ledgerPeers
```
coot, line 926: These are static ledger pools, it wouldn't be difficult to have a script which makes them changing over time.  Please add an issue, so we'll do that in the future.

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
```diff
@@ -387,6 +390,8 @@ genNodeArgs raps minConnected genLocalRootPeers (ntnAddr, rap) = do
 
   peerSharing <- arbitrary
 
+  ledgerPeers <- arbitrary
```
coot, line 393: I am a bit suspicious if this is what we want.  The ledger peers need to run so they can be promoted.  Since we have to be careful not to start too many nodes, we could divide `relays` into two parts: local roots and ledger peers.

It would be also good to add a test label which computes how many ledger peers a node promoted as a hot peer - this way we can be sure we are making a difference.

coot (changes requested):
Two comments...

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: Done!

ouroboros-network/test/Test/Ouroboros/Network/Testnet/Simulation/Node.hs
bolt12: done!

bolt12: No, we don't have coverage for diffusion tests, yet

coot: Please create an issue for it then.

bolt12: https://github.com/input-output-hk/ouroboros-network/issues/4683

coot (approved):
LGTM

PR #4627, opened 2023-07-17
Release ouroboros-network

bolt12:
# Description

Release `ouroboros-network` 0.8.2.0


coot:
I think we need also to release at least `ouroboros-network-framework` as well.  You can use `./scripts/release-to-chap.sh -r` to report what has changed since last releases.

bolt12:
I see! Thanks @coot

njd42 (approved):

PR #4632, opened 2023-07-27
Add -fno-full-laziness to protocols

jasagredo:
We observed a manifestation of the Conduit problem ([see](https://www.well-typed.com/blog/2016/09/sharing-conduit/)) which can be circumvented with this temporary fix.


coot:
Yes, I think it's worth it.

coot (approved):
LGTM

PR #4633, opened 2023-07-27
Use NoThunks to verify no memory leaks in PeerMetrics

coot:
- ouroboros-network-framework: added NoThunk orphaned instances
- peer-metrics: use NoThunks in PeerMetrics


bolt12 (approved):

coot:
Some test were failing because we also need to evaluate arguments of `newPeerMetrics'` to normal form.  This fixes it adding a bunch of `NFData` & `NoThunk` instances.

We should add `NFData` instance to `IP` in the `iproute` package (and `NoThunk` instance too).

cabal.project
```diff
@@ -42,7 +42,7 @@ package io-classes
   flags: +asserts
 
 package strict-stm
-  flags: +asserts
+  flags: +asserts +checktvarinvariant
```
bolt12, line 45: should this be 

```suggestion
  flags: +asserts +checktvarinvariants
```

nix/ouroboros-network.nix
```diff
@@ -50,8 +50,8 @@ let
           });
         }
         {
-          packages.strict-stm.components.library.configureFlags = lib.mkForce
-            (if checkTVarInvariant then [ "-f checktvarinvariant" ] else [ ]);
+          packages.strict-checked-vars.components.library.configureFlags = lib.mkForce
+            (if checkTVarInvariants then [ "-f+checktvarinvariants" ] else [ ]);
```
bolt12, line 54:
```suggestion
            (if checkTVarInvariants then [ "-f +checktvarinvariants" ] else [ ]);
```

cabal.project
coot: In that commit we're still using `strict-stm` which used `checktvarinvariant`; unfortunatelly the `strict-checkd-vars` package renamed the flag to `checktvarinvariants`.

PR #4638, opened 2023-08-08
Use any in cddl specs / tests.

coot:
- TxSubmission: removed the Hello transformer
- cddl: updated documentation             
- cddl: removed unused definitions        
- cddl: use Any in the local-state-query mini-protocol 
- cddl: use Any in the chain-sync mini-protocol 

TODO:

* [x] cddl: use Any in the `local-tx-monitor` mini-protocol - not needed, we instantiate `SlotNo` to `Word64` in `CDDL` which is the right type.


bolt12 (approved):
LGTM

PR #4640, opened 2023-08-10
nix.flake: simplify nix setup

coot:
This PR provides `nix.flake` which builds on
* `x86_64-linux`,
* `x86_64-darwin`, and
* `aarch64-darwin`

on `x86_64-linux` it cross compoiles to Windows.  All project settings are set in `./scripts/ci/cabal.project.local.{Windows,Windows.CrossCompile,Linux}.  They are also used for GitHub actions, and that's the preferred way to change configuration option.

The tests & benchmarks are disabled in the `cabal.project.local.Windows.CrossCompile`.

The `cabal.project.local.Linux` is used on all the above platforms; `cabal.project.local.Windows` is only used on GitHub actions.

Nix is using two compilers:
* `ghc-8.10.7` and
* `ghc-9.6.2`

GitHub Actions, only build with  `ghc-8.10.7` on `windows-latest`.

There is a shell available with `nix develop` which gives access to `ghc-9.6.2` and some other build tools.

Based on `cardano-cli` `nix.flake`.


bolt12 (approved):

coot:
I explicitly disabled cross compilation for now.  Once the devx team will find out what's wrong we will re-enable it.

PR #4654, opened 2023-08-22
Split tests into simulation & IO

coot:
Split `ouroboros-network-framework:test` & `ouroboros-network:test` into `io-tests` & `sim-tests`.   The first will contain all the tests that need to run natively on all platforms (speccifically on Windows).


bolt12 (approved):
LGTM!

PR #4660, opened 2023-08-26
Error inspection

### LINKED ISSUE #3836: Inspect use of `error`
Unrecoverable errors, such as raised by error, might fall into various categories (at least)
1. use in ’unreachable code’ / impossible to reach.
2. invariants between various functions & modules have been violated.
3. unexpected input from users, the environment, etc.
Upon first glance, it appears all the instances of error fall into category 1. Suggestions:

* When feasible, rewrite to eliminate the call to error. Of course, this is sometimes much more trouble than it’s worth.
* When rewriting away is infeasible, use a panic (a la GHC source code, for practical purposes the same as error) to indicate this is ”category 1”. And even better, insert a comment as to ”why we believe it’s impossible vs why ghc doesn’t think it’s impoossible” (so the next coder who looks
at the code doesn’t need to figure this all out for himself).

The result of the above would be the following:
* No calls to error in program.
* Unreachable/impossible code is clearly marked so.
* Category 2 errors are caught with assert
* Category 3 errors are all embedded into SomeException (as appears to be
the case).

samcowger:
# Description

This PR contains results of Galois's inspection of the use of `error` in `ouroboros-network` packages. It consists of an enumeration of `error` uses and my notes on each. I was able to document, modify, or remove some errors immediately; others would benefit from more information or clarification from someone more familiar with this codebase.

Viewers/reviewers may examine the diff first and my notes (currently at `3836/3836.md`) second, or vice versa. Everything in the diff should have a corresponding entry in the notes, but not every entry in the notes has a corresponding diff.

This is marked as a draft because there are a couple of changes that may warrant additional unit tests, and because I recognize my notes will need to be relocated or removed before merging, but I wanted to get eyes on this work sooner rather than later. I'll be going through this to annotate (in Github) items that I feel especially warrant another set of eyes.

Fixes #3836.

cc @coot @mtullsen 

# Checklist

- Branch
  - [N/A] If serialization changes, user-facing consequences (e.g. replay from genesis) are confirmed to be intentional.
- Pull Request
  - [x] Self-reviewed the diff
    - What does this PR change?
  - [x] Reviewer requested


ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
```diff
@@ -290,9 +290,9 @@ local_TIME_WAIT_TIMEOUT = 0
 
 
 socketAddressType :: Socket.SockAddr -> Maybe AddressType
-socketAddressType Socket.SockAddrInet {}  = Just IPv4Address
-socketAddressType Socket.SockAddrInet6 {} = Just IPv6Address
-socketAddressType addr                    =
+socketAddressType Socket.SockAddrInet {}      = Just IPv4Address
+socketAddressType Socket.SockAddrInet6 {}     = Just IPv6Address
+socketAddressType addr@Socket.SockAddrUnix {} =
   error ("socketAddressType: unexpected address " ++ show addr)
```
samcowger, line 296: Can this function return `Nothing` instead of erroring when it is provided a `SockAddrUnix`?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
```diff
@@ -110,60 +110,60 @@ txSubmissionOutbound tracer maxUnacked TxSubmissionMempoolReader{..} _version co
...
-                -- Our reply type is different in the blocking vs non-blocking cases
-                in case blocking of
-                    TokNonBlocking -> SendMsgReplyTxIds (NonBlockingReply txs') client'
-                    TokBlocking    -> SendMsgReplyTxIds (BlockingReply   txs'') client'
-                      where
-                        txs'' = case NonEmpty.nonEmpty txs' of
-                          Just  x -> x
-                          Nothing -> error "txSubmissionOutbound: empty transaction's list"
-                        -- Assert txs is non-empty: we blocked until txs was non-null,
-                        -- and we know reqNo > 0, hence take reqNo txs is non-null.
```
samcowger, line 165: I'd appreciate a second look at my refactor here, in particular w.r.t. whether or not I preserved the existing strictness properties.

network-mux/src/Network/Mux/Types.hs
```diff
@@ -66,6 +70,23 @@ newtype RemoteClockModel
...
+sampleIntervalPicoseconds = check 10
+  where
+    check n
+     | n > 0 && n < wrapInterval
+       = let picos = diffTimeToPicoseconds n
+         in  [|picos|]
+     | otherwise
+       = fail $ "sampleIntervalPicoseconds: infeasible sampleInterval: "<>show n
+    wrapInterval
+      = remoteClockPrecision * (fromIntegral $ unRemoteClockModel maxBound)
```
samcowger, line 88: I've used Template Haskell to lift computation (or failure to compute) of this constant to compile time. Is this too much, do you think?

ouroboros-network-mock/src/Ouroboros/Network/Mock/Chain.hs
```diff
@@ -225,14 +225,17 @@ rollback p (c :> b) | blockPoint b == p = Just (c :> b)
 rollback p Genesis  | p == genesisPoint = Just Genesis
                     | otherwise         = Nothing
 
+-- | Find the provided point's successor on the provided chain. If the point is
+-- not on the chain, this function's behavior is undefined. Use 'pointOnChain'
+-- to check whether a point is on a chain.
```
samcowger, line 230: This is, more or less, the function's de facto contract already. Use sites of this function take care to guard against the `error` by checking membership before calling it, so I've codified calling it without doing so as undefined behavior and then redefined the behavior to eliminate the `error`. This warrants attention, though, since it does technically represent a change in semantics.

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
```diff
@@ -162,7 +162,7 @@ anchorToPoint (Anchor s h _b) = BlockPoint s h
 -- In this case, we must also be given the 'BlockNo'. This only makes sense
 -- for points that aren't genesis.
 anchorFromPoint :: Point block -> BlockNo -> Anchor block
-anchorFromPoint GenesisPoint _     = error "anchorFromPoint: genesis point"
+anchorFromPoint GenesisPoint _     = error "anchorFromPoint: genesis point" -- TODO
```
samcowger, line 165: Two points here:
- It seems, to my untrained eye, like the natural result of this function might be `AnchorGenesis`, rather than an error.
- This function appears unused across your org's repos. It may be a candidate for deprecation and/or deletion.

Thoughts?

ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
```diff
@@ -807,5 +807,5 @@ filterWithStopSpec p stop = goNext []
           -> AnchoredSeq v a b
           -> AnchoredSeq v a b
     join' a b =
-        fromMaybe (error "could not join sequences") $
+        fromMaybe (error "could not join sequences") $ -- TODO
         join (\_ _ -> True) a b
```
samcowger, line 811: If desired, this error could be eliminated by creating a total/irrefutable version of `join` that glues its provided sequences together without checking their compatibility.

ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Version.hs
```diff
@@ -70,7 +70,7 @@ foldMapVersions :: (Ord vNum, Foldable f, HasCallStack)
                 -> f x
                 -> Versions vNum extra r
 foldMapVersions f fx = case toList fx of
-    [] -> error "foldMapVersions: precondition violated"
+    [] -> error "foldMapVersions: precondition violated" -- TODO
     xs -> foldl1 (<>) (map f xs)
```
samcowger, line 74: There is documentation of this precondition, but not of the rationale for it, as far as I can tell - and I note that the underlying structure can easily represent an empty collection of versions. This function, or the structure itself, may benefit from such documentation.

ouroboros-network-testing/src/Ouroboros/Network/Testing/Data/Script.hs
```diff
@@ -218,9 +218,9 @@ interpretPickScript :: (MonadSTM m, Ord peeraddr)
                     -> STM m (Set peeraddr)
 interpretPickScript scriptVar available pickNum
   | Set.null available
-  = error "interpretPickScript: given empty map to pick from"
+  = error "interpretPickScript: given empty map to pick from" -- TODO
   | pickNum <= 0
-  = error "interpretPickScript: given invalid pickNum"
+  = error "interpretPickScript: given invalid pickNum" -- TODO
```
samcowger, line 223: `interpretPickScript` is used primarily to initialize `PickPolicy` fields of a `PeerSelectionPolicy` (in `ouroboros-network:Test.Ouroboros.Network.PeerSelection.MockEnvironment`). If `pickPeers` can reliably be expected to serve as the applier of `PickPolicy`s, then I wonder if these checks are duplicative of those in `pickPeers`.

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientRegistry.hs
```diff
@@ -298,7 +298,7 @@ setFetchClientContext :: MonadSTM m
 setFetchClientContext (FetchClientRegistry ctxVar _ _ _ _ _) tracer mkPolicy =
     atomically $ do
       ok <- tryPutTMVar ctxVar (tracer, mkPolicy)
-      unless ok $ error "setFetchClientContext: called more than once"
+      unless ok $ error "setFetchClientContext: called more than once" -- TODO
```
samcowger, line 301: This function could afford to be documented with the contract implicit in this error - namely, that it be called exactly (or perhaps at most) once. Conversely, if this contract is incorrect, then perhaps this error could be eliminated.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+entire list is intended to be reduced via a fold, just fold the structure
+directly bypassing the list." This change would affect strictness: `foldMap'` is
+left-associative and strict in its accumulator, while `foldl1` is
+left-associative but lazy in its accumulator. In principle, this should be a net
+gain for efficiency, since lazy left-folds are generally undesirable.
+
+**Fix: document**
+#unfixed
+
+#input-desired
```
coot, line 445: This is a precondition; passing an empty list would mean that the handshake will always fail since one doesn't offer any choice to the remote side.

I agree that we should just use `null` instead, although I would still use `foldMap' f fx` and relay on the `Versions` semigroup instance - this is a good example how to use it.

Alternatively, we could use non-empty lists and [`foldMap1`](https://hackage.haskell.org/package/foldable1-classes-compat-0.1/docs/Data-Foldable1.html#v:foldMap1-39-).  `foldMap1` has been introduced to `base-4.18`.  I hope soon we can switch to `ghc-9.6` (`base-4.18`) and forget about `ghc-8.10`.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+
+This error is triggered when a developer tries to provide an unsuitable sample
+interval. The existing interval is suitable, so I assume this variable is
+somewhat volatile during some development process. The suitability of the
+provided interval can be determined statically, so I've lifted this computation
+to be compile-time via Template Haskell, so that the package fails to compile
+with an unsuitable interval.
+
+TH is polarizing, I realize, so I would be happy to discuss this in greater
+detail and revert it if there's no appetite for such a change.
```
coot, line 524: yes, I'd rather avoid it. I think we don't have any TH so far.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+a `Just`-constructed slot. `ptclArray` is constructed by using every
+`miniProtocolNum` in `ptcls` to index `pnumArray`, and will trigger this error
+on encountering `Nothing` as a result of that indexing. Therefore, by
+construction, the error will never trigger.
+
+Leveraging this principle, I've reworked the list comprehension to extract the
+`Just`-constructed value required using a view pattern. The semantics of this
+are now slightly different: if `pnumArray` fails to index, `ptclArray` will
+silently skip the offending element instead of loudly failing. I've included
+commentary to this effect.
```
coot, line 551: I'd prefer to avoid a view pattern, an explicit pattern match is more readable in my opinion.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+
+Regardless, I've changed the error message to describe the function enclosing
+it, which is `mkPartialBlockHeader`, not `mkPartialBlock`.
+
+**Fix: message**
+#fixed
+**Fix: refactor**
+#unfixed
+
+#input-desired
```
coot, line 600: It's fine, this is just testing code, so any misuse will be uncovered quickly.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+therefore claim that an idiom has emerged: using `successorBlock` to find the
+successor of a nonexistent point is considered undefined behavior. I propose to
+codify this idiom in the function's contract, then redefine the undefined
+behavior to be `Nothing`, a non-`error` result that happens to inhabit the
+function's return type.
+
+**Fix: refactor**
+#fixed
+
+#input-desired
```
coot, line 625: I agree.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+It's my understanding that `PickPolicy`s aren't called directly, but are always
+invoked via `pickPeers`. If this is true, then we can remove these error checks
+from `interpretPickScript` entirely, trusting that the checks in `pickPeers`
+will catch the same failures. If this isn't true, then we could either leave
+this code as-is or change it to the `assert` style favored by `pickPeers`.
+
+**Fix: uncertain**
+#unfixed
+
+#input-desired
```
coot, line 650: I agree, if they are redundant we can remove them.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -651,6 +655,7 @@ pickPeers PeerSelectionState{localRootPeers, publicRootPeers, knownPeers}
                available numClamped)
   where
     precondition         = not (Set.null available) && num > 0
+                        && available `Set.isSubsetOf` KnownPeers.toSet knownPeers
```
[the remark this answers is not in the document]
coot: You're right it's only evaluated if assertions are on which is not in production systems. I think it is a valid one.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
coot: Strictness looks fine, but I am not sure if the logic is preserved. Previously we scrutinised `mbtxs` in `TokBlocking` case and in `TokNonBlocking` (we only errored if it was `Nothing`); now we don't scrutinise it the `ToknonBlocking` one at all.

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientRegistry.hs
coot: It is only called by `blockFetchLogic`; it is defacto an internal function of the `block-fetch` application.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
```diff
@@ -38,6 +38,11 @@ instance Show peer => Show (TraceKeepAliveClient peer) where
     show (AddSample peer rtt gsv) = "AddSample " ++ show peer ++ " sample: " ++ show rtt
         ++ " gsv: " ++ show gsv
 
+-- | A version of `CM.ControlMessage` restricted to its `Continue` and
+-- `Terminate` constructors, suitable for use in `decisionSTM`, a function local
+-- to `keepAliveClient` that never returns `Quiesce`.
+data ControlMessage = Continue | Terminate
```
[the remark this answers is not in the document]
coot: We pass the `ControlMessageSTM` in `ExpandedInitiatorContext` to all the mini-protocols.  If we have different types then we will need different context for established and hot mini-protocols, which might complicate things.

I suspect this change only moves the problem to the `ouroboros-consensus-diffusion` package which will need to contain a partial function from the original `ControlMessage` to the restricted one defined here.

coot (changes requested):

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
nfrisby: The final sentence of the Haddock explains, I think.

> `This only makes sense for points that aren't genesis`

Shot in the dark: It may be the case, then, that the first argument should instead be a `RealPoint`?

coot: I am not sure if it makes sense to not allow anchors from `Genesis`, in principle it could make things more uniform (one doesn't need to treat `Genesis` in a special way), but maybe it makes sense from `consensus` perspective.

I don't know what `RealPoint` is, maybe it's something in `ouroboros-consensus`.

nfrisby: My apologies. It's the last _two_ sentences, not just the last sentence.

> In this case, we must also be given the 'BlockNo'. This only makes sense for points that aren't genesis.

The caller of this function is providing a `Point` and a `BlockNo` that refer to the same block. Notably, the genesis "block" does not have a block no; block no 0 identifies the first _real_ block, ie the successor of the genesis block. Hence, the existence of the given `BlockNo` implies that the given `Point` is _not_ `GenesisPoint`.

The quotient type "`Point` but not Genesis" is isomorphic to what Consensus defines to be the `RealPoint` type.

ouroboros-network-mock/src/Ouroboros/Network/Mock/Chain.hs
samcowger: I can do that. However, another option would be to change this function's type to reflect this contract. Instead of
```hs
successorBlock :: HasHeader block => Point block -> Chain block -> Maybe block
```
we would have
```hs
successorBlock :: HasHeader block => Point block -> Chain block -> Maybe (Maybe block)
```
or some other three-pronged type. I gamed this out and don't think it would introduce any new `error`-like behavior at existing call sites to catch the new case - though one site already guards against this case with an `assert`. What do you think?

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
samcowger: I see - thanks for the clarification, @nfrisby. I can imagine a few paths forward here, ordered roughly by how much turbulence I think they would entail:
1. Leave everything alone.
2. Make this function explicitly partial - `Point block -> Maybe BlockNo -> Maybe (Anchor block)`, for instance, could reflect a semantics that respects these genesis/`BlockNo` concerns.
3. Deprecate and delete the function.
4. Make the argument a `RealPoint` - which, I think, would entail moving `RealPoint` from `ouroboros-consensus` to `ouroboros-network-api`.

Do y'all have thoughts or preferences among these options?

ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Version.hs
samcowger: I don't seem to be able to follow that link. Would you mind naming the file/files you're pointing me to?

ouroboros-network-testing/src/Ouroboros/Network/Testing/Data/Script.hs
samcowger: Can do.

3836/3836.md
samcowger: You commented after this, inline in the file in question, that we might ought to leave these checks in place, so I'll assume that later opinion governs.

3836/3836.md
samcowger: I don't think we can use `foldMap` or `foldMap'` and at the same time rely on the `Semigroup Versions` instance with `f`. `foldMap` and `foldMap'` rely on a `Monoid` instance, which `Versions` doesn't (and, as I can now see, can't) have, so we'd need to deconstruct it and rely on the `Semigroup Map` instance with `f'` instead.

Maybe I'll leave things as-is for now, but record an issue referring to this code and this discussion that can be revisited when y'all switch to 9.6. How does that sound?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -558,11 +554,27 @@ peerSelectionGovernorLoop tracer
...
+    -- `GuardedSkip`-constructed value. At least one value,
+    -- `Monitor.connections`, yields a `Guarded`-constructed value, and that
+    -- value will dominate the construction of the overall `<>` concatenation,
+    -- according to the `Semigroup Guarded` instance.
+    guardedDecisions' :: Time
                      -> PeerSharing
                      -> PeerSelectionState peeraddr peerconn
                      -> Guarded (STM m) (TimedDecision m peeraddr peerconn)
-    guardedDecisions blockedAt peerSharing st =
+    guardedDecisions' blockedAt peerSharing st =
```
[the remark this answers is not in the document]
samcowger: Fair enough, I'll revert it.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+match a pattern of other record fields (`diNtnAddressType`, `cmAddressType`,
+perhaps inter alia) being similarly partial. I hesitantly assume that the
+program logic would change if the `error` call were replaced with a
+`Nothing`-constructed result, but
+- if that is the case, then
+  - there should probably be documentation to that effect on this function to
+    warn future maintainers from accidentally making such a change, or
+  - the function should be made total in type (`SockAddr -> AddressType`) and
+    composed with `Just` at use sites
+- if that **isn't** the case, then the error should be replaced with `Nothing`
```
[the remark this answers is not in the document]
samcowger: It seems like the biggest impact of that change would be new partialities at instance definitions for `GlobalAddressScheme`. See, for instance, [`Test.Ouroboros.Network.Diffusion.Node.NodeKernel`](https://github.com/input-output-hk/ouroboros-network/blob/cdd54f98c6743f5fd7bfe0d5a9d0b0f8a4a10b73/ouroboros-network/sim-tests/Test/Ouroboros/Network/Diffusion/Node/NodeKernel.hs#L126-L134). Does that seem like the tip of an iceberg, or like a manageable set of changes?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
samcowger: In the unmodified code, there's no path that allows constructing `mbtxs` via `Nothing` if `blocking` is `TokNonBlocking`, so I don't think it actually needs to be scrutinized. That's the crux of my decision to refactor. Does that strike you as valid?

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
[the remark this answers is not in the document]
samcowger: It seems to me like `peer-sharing` doesn't actually need this new structure - whether it's established or hot, it seems capable of relying on the existing `ControlMessage` structure. Would it be sound to simply not change the existing protocols?

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
coot: You might need to ping @nfrisby on slack.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
coot: You're right, it looks good.

ouroboros-network-mock/src/Ouroboros/Network/Mock/Chain.hs
coot: `Maybe (Maybe ...)` is something confusing, it's not obvious which `Nothing` case is which one, and thus a bit error prone.  I'd leave it as is.

ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
coot: @nfrisby do you think it's worth it?

ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Version.hs
coot: I would like to use `NonEmpty` list for this, but I've been waiting for the `Foldable1` type class, which we can use once we drop `ghc-8.10`. It's enough for now to create an issue for it or leave a `TODO` note.

ouroboros-network-mock/src/Ouroboros/Network/Mock/Chain.hs
samcowger: Ok, fair enough, I'll leave it. Regarding making an issue for updating the call sites, I don't think that's necessary - there appear to be precisely three, and all of them guard against calling `successorBlock` in cases where the point isn't on the chain.

Would you prefer that the `error` behavior remain, or that we keep my change to overloading `Nothing`?

ouroboros-network-framework/src/Ouroboros/Network/Protocol/Handshake/Version.hs
samcowger: I've opened #4677 for this.

3836/3836.md
samcowger: I've opened #4677  for this.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+is already full. I believe this error is not impossible to trigger, but instead
+represents a failure of program logic - accidentaly resetting some state instead
+of initializing it from scratch - severe enough to warrant program termination.
+
+I say "I believe" because the surrounding function (`setFetchClientContext`) is
+undocumented. It should probably have some commentary attesting to this logic
+and the rationale for this error.
+
+**Fix: document**
+#unfixed
```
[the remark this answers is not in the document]
samcowger: I've opened #4678 for this, requesting documentation of the function.

network-mux/src/Network/Mux/Ingress.hs
```diff
@@ -173,17 +174,16 @@ setupDispatchTable ptcls =
             | MiniProtocolState {
                 miniProtocolInfo =
                   MiniProtocolInfo {
-                    miniProtocolNum,
+                    -- This pattern happens to be irrefutable - note that
+                    -- `pnumArray` is constructed to ensure that every
+                    -- `miniProtocolNum` in `ptcls` indexes to a `Just` value.
+                    miniProtocolNum = ((pnumArray !) -> Just pix),
```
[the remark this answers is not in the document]
samcowger: Fair enough. I've reverted the change and included this documentation at the original `error` site.

3836/3836.md
samcowger: Reverted.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+`waitAnyCatchSTM`, then, will always yield an asynchronous action that was part
+of the original list, so not finding one should necessitate a `panic`. (This
+analysis assumes a coherent instance of `Eq (Async m Void)`, one that will
+"correctly" compare an action as equal with itself.)
+
+Because this pattern is somewhat inverted, in that threads are expected to throw
+exceptions rather than run to completion, it's difficult to factor out the use
+of `error` here, even though it seems like it ought to be possible with a
+judicious application of `zip` to associate the domains with their async
+results.
```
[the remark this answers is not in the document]
samcowger: Aha, good call! I implemented this in https://github.com/input-output-hk/ouroboros-network/pull/4660/commits/2b0428483dd817d275d86fe91432e1faa8fbc28b; if you could take a look, that'd be great.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+to being logical inverses of one another, if this is desirable.
+
+I note that this function is exported but does not appear to be used elsewhere
+within the IOG org (per an org-wide Github search). Perhaps it could be marked
+deprecated and/or removed?
+
+**Fix: uncertain**
+#unfixed
+
+#input-desired
```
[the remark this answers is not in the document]
samcowger: Discussion about this has largely taken place in another conversation on this issue, so I'll resolve this one.

ouroboros-network-mock/src/Ouroboros/Network/Mock/Chain.hs
coot: Let's leave it as it was.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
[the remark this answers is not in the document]
coot: I'd leave this as it used to be. Otherwise I suspect we'll get problems integrating with `ouroboros-consensus-diffusion`.

ouroboros-network-mock/src/Ouroboros/Network/Mock/Chain.hs
samcowger: Can do.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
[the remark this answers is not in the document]
samcowger: Ok, I can revert it, but out of curiosity, what sorts of problems? The fact that it's an internal convenience type, not exported from `KeepAlive`, makes me think that it's unlikely to significantly impact integration.

ouroboros-network-mock/src/Ouroboros/Network/Mock/Chain.hs
samcowger: I restored the function's original error behavior, but I think its error condition still ought to be documented. I want to make sure I'm using chain/block/tip vocabulary appropriately - what do you think of:
```hs
-- | A 'Just' result holds the provided point's successor block. A 'Nothing'
-- result means the provided point was the tip.
--
-- The function will error if the point is not on the chain - callers can use
-- 'pointOnChain' to check point membership on a chain in advance of calling
-- this function.
```

coot: LGTM

3836/3836.md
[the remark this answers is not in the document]
coot: Very nice.

ouroboros-network/src/Ouroboros/Network/KeepAlive.hs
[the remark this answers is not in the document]
coot: In  `ouroboros-consensus-diffusion` we have to write a function which accepts `STM Ouroboros.Network.ControlMessage.ControlMessage` and we will need to write a partial function anyway, but there this partial function will depend on a property which is no longer local.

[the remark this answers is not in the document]
coot: In particular we will need to fill this [`ClientApp`](https://github.com/input-output-hk/ouroboros-consensus/blob/main/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/Network/NodeToNode.hs#L421-L426) type, here's the [`ExpandedInitatorContext`](https://github.com/input-output-hk/ouroboros-network/blob/master/ouroboros-network-framework/src/Ouroboros/Network/Context.hs#L24).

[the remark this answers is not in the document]
samcowger: Ok, it sounds like reverting is appropriate. I'll do so.

3836/3836.md
```diff
@@ -0,0 +1,666 @@
...
+superset of all other peer sets, so this error is quite unlikely to be triggered
+as long as any extant `PeerSelectionState` adheres to its invariant.
+
+I've added documentation at this `error` site of the rationale I've derived, and
+have included a reference to the invariant-documenting function at the
+declaration site of `PeerSelectionState`.
+
+I've also added a precondition that checks whether `available` is a subset of
+`knownPeers`. This may offer a poor utility to expense ratio in practice, and
+I'd welcome input and discussion on this change.
```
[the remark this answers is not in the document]
samcowger: What do you mean by log, exactly? I haven't looked much into the testing facilities for the `PeerSelectionState` API.

3836/3836.md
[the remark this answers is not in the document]
samcowger: I've opened #4681 to discuss this.

3836/3836.md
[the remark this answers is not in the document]
samcowger: @coot

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
nfrisby: > I am not sure if it makes sense to not allow anchors from Genesis

Are you suggesting removing the `AnchorGenesis` constructor? I anticipate that that one _is_ used in the codebase.

---

My intuition favors options 3 or 4. If it's unused, removing it as dead-code seems wise. On the other hand, it's a very reasonable smart constructor to expect for the data type---in that case, I would favor renaming it to `anchorFromRealPoint`, eg.

---

I would I don't know if `RealPoint` would be useful in the `ouroboros-network` codebase---@marcin?

https://github.com/input-output-hk/ouroboros-consensus/blob/71e6cbb304d89ba7abe32c8d84239014ca24c94e/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Block/RealPoint.hs#L41C1-L43

ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
nfrisby: > creating a total/irrefutable version of join that glues its provided sequences together without checking their compatibility

Just for context: I would not call such a function "total", since the resulting anchored seq is potentially invalid. Along those lines, if such a function were to exist, I think its name should start with the prefix `unsafe*`.

If the new function has the `unsafe*` prefix, then I would only have a very slight preference for leaving this call site as it is compared to defining and using the new function.

---

I don't know the context within this function, but I recall an idea elsewhere to avoid an impossible case by enriching the interface of a preceding call (something like replacing `... checkIntersect ... join ...` with `... checkIntersectAndIfSoJoin ...`).

It looks like the context here is some low-level pattern matching, so maybe that's not immediately obvious. Though perhaps the joining could be happening simultaneously in a new accumulator argument, or something like that?

samcowger: > I would only have a very slight preference for leaving this call site as it is compared to defining and using the new function.

That's reasonable.

> something like replacing `... checkIntersect ... join ...` with `... checkIntersectAndIfSoJoin ...`

A function like that seems like it'd be partial as well (the implied `checkIntersectAndIfNot` case), so we'd have to figure out how to propagate that partiality through or out of this function.

> the joining could be happening simultaneously in a new accumulator argument

I guess an approach like that would eliminate the error, but it would also be indirectly duplicative of `join`, and would introduce the potential to diverge from `join`'s semantics, if they were to change.

Given all this, it seems to me like the existing `error`-based approach might be a local minimum/maximum of complexity/safety, so I'm minded to recommend that we just leave it as-is. Thoughts?

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
samcowger: To be clear, my suggestion that `RealPoint` be moved derives from the dependency of `ouroboros-consensus` on `ouroboros-network-api`, though I imagine it could also be defined in another package entirely. At any rate, I've opened #4690 to continue this discussion, thinking that conversations on moving structures between packages or deprecation of functions warrant more visibility.

coot: > I would I don't know if RealPoint itself would otherwise be useful in the ouroboros-network codebase---@marcin?

Other than the use case here, we don't need `RealPoint`, but I'd be fine with moving it to  `Ouroboros.Network.Block`.

3836/3836.md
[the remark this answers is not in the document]
coot: This would actually be much more involved, it would require to identify all log messages which could  also carry the information about available peers, like `TraceForgetColdPeers`.  Then in a test we can collect all these traces and verify that `available` is a subset of `known` peers.  But this would also pollute the log produced by `cardano-node`, which is not ideal; so we'd need to create another tracer for these values.  This seems to be quite some work, while assertions give us this quite cheaply, so let's do it this way.

ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
coot: Let's leave it as is then.

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
coot, line 165:
```suggestion
anchorFromPoint GenesisPoint _     = error "anchorFromPoint: genesis point" -- TODO #4690
```

ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
```diff
@@ -807,5 +807,5 @@ filterWithStopSpec p stop = goNext []
           -> AnchoredSeq v a b
           -> AnchoredSeq v a b
     join' a b =
-        fromMaybe (error "could not join sequences") $
+        fromMaybe (error "could not join sequences") $ -- TODO
```
coot, line 810:
```suggestion
        fromMaybe (error "could not join sequences") $
```

coot (approved):
LGTM, thank you for a thorough review :).

coot:
@samcowger I rebased your PR.

PR #4670, opened 2023-09-05
UTxO-HD for node 8.2.1

jasagredo:
# Description

Needed for input-output-hk/ouroboros-consensus#205

# Checklist

- Branch
    - [ ] Updated changelog files.
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Codec.hs
+++ b/ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Codec.hs
@@ -37,10 +37,10 @@ codecLocalStateQuery
      )
   => (point -> CBOR.Encoding)
   -> (forall s . CBOR.Decoder s point)
-  -> (forall result . query result -> CBOR.Encoding)
-  -> (forall s . CBOR.Decoder s (Some query))
-  -> (forall result . query result -> result -> CBOR.Encoding)
-  -> (forall result . query result -> forall s . CBOR.Decoder s result)
+  -> (forall fp result . SingI fp => query fp result -> CBOR.Encoding)
+  -> (forall s . CBOR.Decoder s (SomeQuery query))
+  -> (forall fp result . query fp result -> result -> CBOR.Encoding)
+  -> (forall fp result . query fp result -> forall s . CBOR.Decoder s result)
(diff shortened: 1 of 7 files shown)


ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Codec.hs
```diff
@@ -37,10 +37,10 @@ codecLocalStateQuery
      )
   => (point -> CBOR.Encoding)
   -> (forall s . CBOR.Decoder s point)
-  -> (forall result . query result -> CBOR.Encoding)
-  -> (forall s . CBOR.Decoder s (Some query))
-  -> (forall result . query result -> result -> CBOR.Encoding)
-  -> (forall result . query result -> forall s . CBOR.Decoder s result)
+  -> (forall fp result . SingI fp => query fp result -> CBOR.Encoding)
```
coot, line 40: Will the encoding depend on the `fp` parameter?

ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Type.hs
```diff
@@ -212,3 +224,39 @@ instance (ShowQuery query, Show point)
...
+-- queries.
+data QueryFootprint =
+    -- | The query doesn't need ledger tables, thus can be answered only with
+    -- the ledger state.
+    QFNone
+    -- | The query needs some tables, but doesn't need to traverse the whole
+    -- UTxO set.
+  | QFOne
+    -- | The query needs to traverse the whole UTxO set.
+  | QFAll
```
coot, line 250: I have an impression that this flag is query specific, isn't it?  It seems it only matters for the server side, not the client side.  If this is the case I'd prefer to keep in on the consensus side, because here we are complicating things also for clients, e.g. `cardano-wallet`, `cardano-cli`, etc.

If we move it to consensus, the queries still have the type and it will affect clients, so the question is if it is worth to encode this as a type parameter? Removing the type only make sense if each query is only done in one of the mode, rather than the client side choosing the mode of operation of a query.

jasagredo:
This PR is no longer needed :tada:

PR #4674, opened 2023-09-14
AcceptedConnectionsLimit: add json instances

smelc:
Contributes to fixing https://github.com/input-output-hk/cardano-node/issues/5470

# Description

We are getting rid of some orphan instances we have [here in cardano-node](https://github.com/input-output-hk/cardano-node/blob/396739ca42a3b09fc5e3df698f34cec391f795aa/cardano-node/src/Cardano/Node/Orphans.hs#L43): that's PR https://github.com/input-output-hk/cardano-node/pull/5466. That is why we need to add Json instances for `AcceptedConnectionsLimit`.

This PR adds `aeson` as a dependency to `ouroboros-network-framework`. At the same time I'm writing this, aeson 2.1.2.1 is pulled: please let me know which version bound you want to specify (if any) in the cabal project. Right now, I've  omitted specifying the version.

# Checklist

- Branch
    - [X] Updated changelog files.
    - NA The documentation has been properly updated
    - NA New tests are added if needed and existing tests are updated
    - NA If serialization changes, user-facing consequences (e.g. replay from genesis) are confirmed to be intentional.
- Pull Request
    - [X] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
@@ -1,4 +1,5 @@
-{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE NamedFieldPuns    #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 -- | Rage limiting of accepted connections
 --
@@ -18,6 +19,8 @@ import           Control.Tracer (Tracer, traceWith)
 import           Data.Typeable (Typeable)
 import           Data.Word
 import           Text.Printf
+
+import           Data.Aeson.Types
 
 
 -- | Policy which governs how to limit the number of accepted connections.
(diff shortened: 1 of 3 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
```diff
@@ -41,6 +44,28 @@ data AcceptedConnectionsLimit = AcceptedConnectionsLimit {
   }
   deriving (Eq, Ord, Show)
 
+instance ToJSON AcceptedConnectionsLimit where
+  toJSON AcceptedConnectionsLimit
+          { acceptedConnectionsHardLimit
+          , acceptedConnectionsSoftLimit
+          , acceptedConnectionsDelay
+          } =
+    object [ "AcceptedConnectionsLimit" .=
```
coot, line 53: I'd avoid the top level wrapper, e.g. instead of
```json
{ "AcceptedConnectionsLimit": { "hardLimit": 500, "softLimit": 30, "delay": 60} }
```
I think it's better to just have:
```json
{ "hardLimit": 500, "softLimit": 300, "delay": 60}
```

coot (commented):
@smelc could you say why do you need this instance.  Don't we have an instance in `cardano-node` for it?

smelc:
> @smelc could you say why do you need this instance. Don't we have an instance in cardano-node for it?

@coot> indeed I'm doing this PR because the cli team wants to get rid of the orphan instance we have in cardano-node :+1:

ouroboros-network-framework/src/Ouroboros/Network/Server/RateLimiting.hs
Jimbo4350: I would avoid breaking anything.

smelc: Let's keep it like this then. @coot> Good for you?

smelc:
Ran `stylish` locally to fix CI

coot:
We should have a conversation what to do with the `ToJSON` / `FromJSON` instances.  They are in `cardano-node` for a reason.

Jimbo4350:
> We should have a conversation what to do with the `ToJSON` / `FromJSON` instances. They are in `cardano-node` for a reason.

Why? It hasn't been modified in 21 months.

smelc:
@coot> so what's your take on this PR after @Jimbo4350's comment? Anything I could do to move this one forward?

coot:
@smelc sorry I missed your reply.

The `aeson` dependency does not belong to `ouroboros-network`, we deal with binary data and we provide `CBOR` codecs for mini-protocols. I want to move the instances which were added here to `cardano-node` (where we do maintain them). There's also nothing wrong in having orphaned instances like these in `cardano-node` (or in a separate package).

smelc:
Thanks @coot for the additional context. Hence closing this PR :+1:

PR #4707, opened 2023-10-30
Refactor Resource: retryResource

coot:
# Description

Write `retryResource` (formelly `withResource'`) using open recursion style
loosly inspired by "The Different Aspects of Monads and Mixins" by Bruno C.
d S. Oliveira.

# Checklist

- Branch
    - [ ] Updated changelog files.
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS/DNSActions.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS/DNSActions.hs
@@ -1,5 +1,5 @@
-{-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE CPP                 #-}
+{-# LANGUAGE DeriveFunctor       #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -13,14 +13,16 @@ module Ouroboros.Network.PeerSelection.RootPeersDNS.DNSActions
     -- * Utils
     -- ** Resource
   , Resource (..)
-  , withResource'
+  , retryResource
   , constantResource
     -- ** Error type
   , DNSorIOError (..)
   ) where
 
+#if !defined(mingw32_HOST_OS)
+import           Data.Function (fix)
+#endif
 import           Data.List.NonEmpty (NonEmpty (..))
-import qualified Data.List.NonEmpty as NonEmpty
 
 import           Control.Exception (IOException)
 import           Control.Monad.Class.MonadAsync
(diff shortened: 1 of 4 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS/DNSActions.hs
```diff
@@ -59,40 +59,55 @@ instance Exception exception => Exception (DNSorIOError exception) where
...
-
-constantResource :: Applicative m => a -> Resource m err a
-constantResource a = Resource (pure (Right a, constantResource a))
-
+    -- NOTE: this is more general recursion than what B. Oliveira is using. We
+    -- need 'super' as part of the recursion.  The reason is that we want to
+    -- modify `Resource` itself not a traversal.
+    go :: (NonEmpty DiffTime -> Resource m (Either e a) -> Resource m a)
+       -> (NonEmpty DiffTime -> Resource m (Either e a) -> Resource m a)
+    go this ds@(d :| _) super = Resource $ do
```
bolt12, line 96: The way you arranged the parentheses could see this as
essentially of type `Mixin (NonEmpty DiffTime -> Resource
m (Either e a) -> Resource m a)`, at least according to
the paper. Shouldn't then this be `go super ds@(d :| _) this`  ?

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS/DNSActions.hs
```diff
@@ -59,40 +59,55 @@ instance Exception exception => Exception (DNSorIOError exception) where
...
-          withResource' tracer (dropHead delays) resource'
-        (Right r, resource') ->
-          pure (r, resource')
-
-constantResource :: Applicative m => a -> Resource m err a
-constantResource a = Resource (pure (Right a, constantResource a))
-
+    -- NOTE: this is more general recursion than what B. Oliveira is using. We
+    -- need 'super' as part of the recursion.  The reason is that we want to
+    -- modify `Resource` itself not a traversal.
```
bolt12, line 93: I am not sure if it is accurate to say that this is more general. It seems more ad-hoc to me, you are pulling the `super` out of your types instead of using it combinator style. I'd rephrase and say this is inspired in what B. Oliveira is using.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS/DNSActions.hs
coot:
```hs
type Mixin a = a -> a -> a

mix :: Mixin a -> Mixin a -> Mixin a
mix f g = \super this -> f (g super this) this

retryMixin :: forall m e a. Monad m
           => Tracer m e
           -> NonEmpty DiffTime
           -> Mixin (NonEmpty DiffTime -> Resource m (Either e a) -> Resource m a)
retryMixin tracer ds0 = \_super this ds super -> Resource $ do
    ea <- withResource super
    case ea of
      (Left e, super') -> do
        traceWith tracer e
        -- threadDelay
        withResource (this ds super')
      (Right a, super') -> do
        return (a, this ds0 super')


```

the thing is `_super` is useless here, and I don't know if there's a way to implement through inheritance (using `_super` rather than `super` above).

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS/DNSActions.hs
coot: This can fit into `Mixin` style, we're just ignoring the first argument: https://github.com/input-output-hk/ouroboros-network/pull/4707#discussion_r1376005467

Maybe more general is a wrong word as it fits into the same pattern, but it's a different way of modifying a predefined behaviour so I wouldn't say it's ad-hoc.

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS/DNSActions.hs
coot: > Shouldn't then this be go super ds@(d :| _) this ?

I think I should call `super` something else, since it plays a different role than `super` in the paper.

bolt12: What itches me about this is that it seems to me one is just trying to make the
type tetris work and fit this particular mixin type signatures, when in truth
what the function is actually doing is not related to mixins at all. This
function will be retrying a monadic until it succeeds, it there isn't any
inheritance going on. It `fix` works great here because it does exactly what we
want.

I think this rewrite and the mention of mixins paper is a little bit odd and could be confusing for someone reading this for the first time

coot: Indeed, that's why I avoided the `Mixin` in the PR, we don't need inheritance in our cases.  I posted the code only to show that this fits into the `Mixin` pattern (in a trivial way).

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS/DNSActions.hs
```diff
@@ -59,40 +61,52 @@ instance Exception exception => Exception (DNSorIOError exception) where
...
-          withResource' tracer (dropHead delays) resource'
-        (Right r, resource') ->
-          pure (r, resource')
-
-constantResource :: Applicative m => a -> Resource m err a
-constantResource a = Resource (pure (Right a, constantResource a))
-
+    step :: (NonEmpty DiffTime -> Resource m (Either e a) -> Resource m a)
+         -> (NonEmpty DiffTime -> Resource m (Either e a) -> Resource m a)
+    step ref ds@(d :| _) resource = Resource $ do
```
bolt12, line 95:
```suggestion
    step rec ds@(d :| _) resource = Resource $ do
```

bolt12 (approved):
I watched the unfoldr episode but still think this particular example doesn't quite fit nor benefit from applying this open recursion technique. The reasons being what we want to change is the tail call value of `Difftime` and not exactly the resource itself. However I am approving since the diff has been massaged well enough to become more readable!

coot:
> The reasons being what we want to change is the tail call value of Difftime and not exactly the resource itself.

The resource evolves and that's why we need to pass it through the recursive call, rather than accept it as constant argument.

PR #4728, opened 2023-11-24
Disable PeerSharing completely for NTNV_11-12

bolt12:
# Description

Disable PeerSharing completely for NTNV_11-12. I overlooked the fact that the handshake response is also encoded so locally the node will agree on something and send something different over the wire. The fix is to completely disable PeerSharing for older versions.

This also fixes a bug in the pool of peers available to share.

### DIFF
--- a/ouroboros-network-api/CHANGELOG.md
+++ b/ouroboros-network-api/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Non-breaking changes
 
-* Fixed Codec to disable `PeerSharing` with buggy versions 11 and 12.
+* Fixed Codec to disable `PeerSharing` completely with buggy NTNt versions 11 and 12.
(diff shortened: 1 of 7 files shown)


ouroboros-network-api/CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 ### Non-breaking changes
 
-* Fixed Codec to disable `PeerSharing` with buggy versions 11 and 12.
+* Fixed Codec to disable `PeerSharing` completely with buggy NTNt versions 11 and 12.
```
coot, line 9:
```suggestion
* Fixed handshake codec: disabled `PeerSharing` for `node-to-node` versions 11 and 12.
```

jasagredo (approved):

PR #4734, opened 2023-11-29
Fixes async demotions to Cold not being noticed

bolt12:
# Description

This fixes the bug where when a peer disconnected it could no longer be connected to due to the governor not picking it up.

### DIFF
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -7,7 +7,9 @@
 
 ### Non-breaking changes
 
-* Fix random selection of peers to peershare with.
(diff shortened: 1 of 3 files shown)


coot (approved):
This looks good to me.  Thanks @bolt12 for the fix.

I think we should also analyse why the tests we have didn't capture the bug in the first place. This could be because we don't have `prop_governor_target_known_above` in the simulated testnet, which in theory should be able to capture such a bug.  However, this can be done as a follow up.

bolt12:
Yes there's an issue to write all PeerSelection tests into Testnet cases

PR #4739, opened 2023-11-30
KnownPeers insert function bug fix and PeerSharing improvements

bolt12:
# Description

- Introduction of `knownSuccessfulConnection` Flag: added a new flag to the `KnownPeerInfo` data type. This flag serves an essential purpose in tracking the peers a node has successfully connected to previously. It plays a vital role in filtering peers for subsequent sharing operations.

- Resolution of the `KnownPeers` Insert Function Bug: fix of a semantic bug in the `KnownPeers` insert function. Previously, the function's implementation led to unintended overwriting of  information, notably the `PeerAdvertise` flag. This caused an issue where no peers were available for sharing, which has now been effectively addressed.

- Delay in Peer Sharing for New Connections: implemented a five-minute delay in peer sharing for newly established connections. This delay ensures a more stable and reliable integration of new peers into the network's sharing mechanism.

- Optimization of Peer Share Retry Time: adjusted the peer share retry time to 900 seconds.

- Enabling of `PeerAdvertise` Flag: enabled the `PeerAdvertise` flag for peer sharing results and inbound connections. This enhancement is applied unless the peers in question are already known by the outbound governor, in which case their original `PeerAdvertise` status is preserved.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
@@ -71,7 +71,7 @@ belowTargetLocal :: forall peeraddr peerconn m.
                  => PeerSelectionActions peeraddr peerconn m
                  -> MkGuardedDecision peeraddr peerconn m
 belowTargetLocal actions
-                 PeerSelectionPolicy {
+                 policy@PeerSelectionPolicy {
                    policyPickColdPeersToPromote
                  }
                  st@PeerSelectionState {
@@ -127,7 +127,7 @@ belowTargetLocal actions
                           inProgressPromoteCold = inProgressPromoteCold
                                                <> selectedToPromote
                         },
-        decisionJobs  = [ jobPromoteColdPeer actions peer IsNotBigLedgerPeer
+        decisionJobs  = [ jobPromoteColdPeer actions policy peer IsNotBigLedgerPeer
                         | peer <- Set.toList selectedToPromote ]
       }
 
@@ -167,7 +167,7 @@ belowTargetOther :: forall peeraddr peerconn m.
                  => PeerSelectionActions peeraddr peerconn m
                  -> MkGuardedDecision peeraddr peerconn m
 belowTargetOther actions
-                 PeerSelectionPolicy {
+                 policy@PeerSelectionPolicy {
                    policyPickColdPeersToPromote
                  }
                  st@PeerSelectionState {
@@ -217,7 +217,7 @@ belowTargetOther actions
                           inProgressPromoteCold = inProgressPromoteCold
                                                <> selectedToPromote
                         },
-        decisionJobs  = [ jobPromoteColdPeer actions peer IsNotBigLedgerPeer
+        decisionJobs  = [ jobPromoteColdPeer actions policy peer IsNotBigLedgerPeer
                         | peer <- Set.toList selectedToPromote ]
       }
 
@@ -245,7 +245,7 @@ belowTargetBigLedgerPeers :: forall peeraddr peerconn m.
                           => PeerSelectionActions peeraddr peerconn m
                           -> MkGuardedDecision peeraddr peerconn m
 belowTargetBigLedgerPeers actions
-                          PeerSelectionPolicy {
+                          policy@PeerSelectionPolicy {
                             policyPickColdPeersToPromote
                           }
                           st@PeerSelectionState {
@@ -296,7 +296,7 @@ belowTargetBigLedgerPeers actions
                           inProgressPromoteCold = inProgressPromoteCold
                                                <> selectedToPromote
                         },
-        decisionJobs  = [ jobPromoteColdPeer actions peer IsBigLedgerPeer
+        decisionJobs  = [ jobPromoteColdPeer actions policy peer IsBigLedgerPeer
                         | peer <- Set.toList selectedToPromote ]
       }
 
@@ -336,13 +336,16 @@ maxColdPeerRetryBackoff = 5
 jobPromoteColdPeer :: forall peeraddr peerconn m.
                        (Monad m, Ord peeraddr)
                    => PeerSelectionActions peeraddr peerconn m
+                   -> PeerSelectionPolicy peeraddr m
                    -> peeraddr
                    -> IsBigLedgerPeer
                    -> Job () m (Completion m peeraddr peerconn)
 jobPromoteColdPeer PeerSelectionActions {
                      peerStateActions = PeerStateActions {establishPeerConnection},
                      peerConnToPeerSharing
-                   } peeraddr isBigLedgerPeer =
+                   }
+                   PeerSelectionPolicy { policyPeerShareActivationDelay }
+                   peeraddr isBigLedgerPeer =
     Job job handler () "promoteColdPeer"
   where
     handler :: SomeException -> m (Completion m peeraddr peerconn)
@@ -411,8 +414,9 @@ jobPromoteColdPeer PeerSelectionActions {
                                            targetNumberOfEstablishedBigLedgerPeers
                                          }
                              }
-                             _now ->
+                             now ->
         let establishedPeers' = EstablishedPeers.insert peeraddr peerconn
+                                                        (addTime policyPeerShareActivationDelay now)
                                                         establishedPeers
             -- Update PeerSharing value in KnownPeers
             knownPeers'       = KnownPeers.insert (Map.singleton peeraddr (Just peerSharing, Nothing, Nothing))
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE LambdaCase          #-}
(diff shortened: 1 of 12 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -419,8 +422,16 @@ jobPromoteColdPeer PeerSelectionActions {
...
+            advertise = case peerSharing of
+                          PeerSharingEnabled  -> DoAdvertisePeer
+                          PeerSharingDisabled -> DoNotAdvertisePeer
             -- Update PeerSharing value in KnownPeers
-            knownPeers'       = KnownPeers.insert (Map.singleton peeraddr (Just peerSharing, Nothing, Nothing))
+            knownPeers'       = KnownPeers.insert (Map.singleton peeraddr ( Just peerSharing
+                                                                          , newDefaultValue peeraddr
+                                                                                            advertise
+                                                                                            knownPeers
+                                                                          , Nothing))
```
coot, line 433: we do two `KnownPeers` lookup here, one to `insert` another in `newDefaultValue`, why not implement `KnownPeers.alter` similar to [`Map.alter`](https://hackage.haskell.org/package/containers-0.7/docs/Data-Map-Internal.html#v:alter).

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -338,6 +359,12 @@ setTepidFlag :: Ord peeraddr
              -> KnownPeers peeraddr
 setTepidFlag = setTepidFlag' True
 
+setSuccessfulConnectionFlag :: Ord peeraddr
+                            => peeraddr
+                            -> KnownPeers peeraddr
+                            -> KnownPeers peeraddr
+setSuccessfulConnectionFlag = setSuccessfulConnectionFlag' True
```
coot, line 366: Why not inline `setSuccessfulConnectionFlag'` here? It's not used anywhere else.

coot (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -188,39 +198,32 @@ insert peeraddrs
     let allPeersAddrs = Map.keysSet peeraddrs
         knownPeers' = knownPeers {
           allPeers =
-              let (<+>) = Map.unionWith mergePeerInfo in
-              allPeers
-          <+> Map.map newPeerInfo peeraddrs,
+              let (<+>) = Map.unionWith (flip const) in
```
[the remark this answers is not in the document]
bolt12: Great!

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
bolt12: I thought I was following the style of the other functions, but realise now its not it

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -121,6 +123,8 @@ data PeerSelectionPolicy peeraddr m = PeerSelectionPolicy {
        policyPeerShareOverallTimeout    :: !DiffTime,
        -- ^ Amount of time the overall batches of peer sharing requests are
        -- allowed to take
+       policyPeerShareActivationDelay   :: !DiffTime,
+       -- ^ Delay until we consider a peer suitable for peersharing
```
coot, line 127:
```suggestion
       -- ^ Delay until we consider a peer suitable for peer sharing.
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -918,3 +923,19 @@ deriving instance (Ord peeraddr, Show peeraddr)
...
+-- special cases of inbound peers and peer share result peers.
+--
+newDefaultValue :: Ord peeraddr
+                => peeraddr
+                -> v
+                -> KnownPeers peeraddr
+                -> Maybe v
+newDefaultValue a value knownPeers
+  | KnownPeers.member a knownPeers = Nothing
+  | otherwise                      = Just value
```
coot, line 940: I thought that this won't be needed once we have the `KnownPeers.alter` method.

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -187,39 +236,31 @@ insert peeraddrs
...
+          , availableToConnect = availableToConnect
+          } =
+  let newAllPeers = Map.alter f k allPeers
+   in knownPeers {
+        allPeers = newAllPeers
+      , availableToConnect =
+          if Map.member k newAllPeers
+             then Set.insert k availableToConnect
+             else Set.delete k availableToConnect
       }
```
coot, line 264:
```suggestion
   let (isMember, allPeers') = Map.alterF (\a -> let a' = f a in
                                                 (isJust a', a'))
                                          k allPeers
   in knownPeers {
        allPeers = allPeers'
      , availableToConnect =
          if isMember
             then Set.insert k availableToConnect
             else Set.delete k availableToConnect
      }
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
```diff
@@ -192,8 +194,9 @@ jobPeerShare PeerSelectionActions{requestPeerShare}
                                                (Map.fromList
                                                 $ map (\a -> ( a
                                                              , ( Nothing
+                                                               , newDefaultValue a DoAdvertisePeer (knownPeers st)
                                                                , Nothing
-                                                               , Nothing))
+                                                               ))
                                                       )
```
coot, line 200: it seems a more useful method would be `KnownPeers.alter` since we are looking up `knownPeers` for the existing value.  For this it would be more convenient to modify `KnownPeers.alter` to act on multiple peers at once:

```hs
alter :: Ord peeraddr
      => (Maybe KnownPeerInfo -> Maybe KnownPeerInfo)
      -> Set peeraddr
      -> KnownPeers peeraddr
      -> KnownPeers peeraddr
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
```diff
@@ -300,7 +302,7 @@ jobPeerShare PeerSelectionActions{requestPeerShare}
                                             (Map.fromList
                                              $ map (\a -> ( a
                                                           , ( Nothing
-                                                            , Nothing
+                                                            , newDefaultValue a DoAdvertisePeer (knownPeers st)
                                                             , Nothing))
                                                    )
                                                    newPeers)
```
coot, line 308: The same here.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/RootPeers.hs
```diff
@@ -118,7 +117,7 @@ jobReqPublicRootPeers PeerSelectionActions{ requestPublicRootPeers
             knownPeers'      = KnownPeers.insert
                                  -- When we don't know about the PeerSharing information
                                  -- we default to NoPeerSharing
-                                 (Map.map (\(a, b) -> (Just PeerSharingDisabled, Just a, Just b)) newPeers)
+                                 (Map.map (\(a, b) -> (Nothing, Just a, Just b)) newPeers)
```
coot, line 120: Do we need to update the comment above?

coot (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/RootPeers.hs
bolt12: Not really :)

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -7,10 +7,13 @@ module Ouroboros.Network.PeerSelection.State.KnownPeers
...
+    -- * KnownPeerInfo operations
+  , alterKnownPeerInfo
     -- * Basic container operations
   , empty
   , size
   , insert
+  , alter
   , delete
   , toSet
   , member
```
coot, line 19: I think this order will make more sense for a Haddock reader:

```suggestion
    -- * Basic container operations
  , empty
  , size
  , insert
  , alter
  , delete
  , toSet
  , member
    -- * KnownPeerInfo operations
  , alterKnownPeerInfo
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -149,6 +162,32 @@ invariant KnownPeers{..} =
         (Set.fromList (PSQ.keys nextConnectTimes)))
 
 
+-------------------------------
+-- KnownPeerInfo manipulation
+--
+
+alterKnownPeerInfo
```
coot, line 169: Add haddocks, maybe include a use case. Describe where the arguments are coming from.

coot: I would also move the definition after `alter` function, as they are related.

coot (approved):
LGTM

Issue #532, opened 2019-05-20
Make `ioToFsError` OS specific


kderme:
Since windows and unix sometimes throw different kind of Exceptions, on https://github.com/input-output-hk/ouroboros-network/issues/285 we made the equality check too lenient (always true). This can be improved. @Edsko commented https://github.com/input-output-hk/ouroboros-network/pull/530#discussion_r285454081:
```
It would be nice to be a little bit more precise here (as precise as we can be without making MockFS OS 
dependent I guess). But this can be done in a separate PR.

Concretely: let's replace FsIllegalOperation, FsResourceInappropriateType and FsInvalidArgument by a 
more general FsInvalidOperation, and then introduce a Posix and Windows specific ioToFsError function. 
Ideally we'd not bundle FsInsufficientPermissions in there though, so if you do try this and find some 
problems with that, let's think about the specific problems there.
```

edsko:
Closed in #557.

edsko:
Uh, no, sorry, Not

mrBliss:
Apparently, on darwin (macOS), we also get differing `FsError`s.
The test failure below was reported by @kantp:

```
Up to date
ouroboros-storage
  Storage
    HasFS
      HasFS
        q-s-m:
Model {knownHandles = RefEnv []
      ,knownPaths = RefEnv []
      ,mockFS = MockFS {mockFiles =
                       Folder (Map.fromList [])
                       ,mockHandles = Map.fromList []
                       ,mockNextHandle = Handle 0}}

   == At (CreateDirIfMissing False (PExpPath ["z"])) ==> At (Resp {getResp = Right (Path (Reference (Concrete ["z"])) ())}) [ 0 ]

Model {knownPaths = RefEnv [+_×_
                               (Reference ["z"]) ["z"]]
      ,mockFS = MockFS {mockFiles =
                       Folder (Map.fromList [+_×_ "z"
                                                (Folder (Map.fromList []))])
                       ,mockHandles = ...}
      ,knownHandles = ...}

[...]

PR #4749, opened 2023-12-13
SizeInBytes: Derive Serialise

RenateEilers:
# Description

This PR adds lets `SizeInBytes` derive a `Serialise` instance, which we need to fix issue [#532](https://github.com/input-output-hk/ouroboros-consensus/issues/532) in  `ouroboros-consensus`.


coot:
@RenateEilers we avoid to use `Serialise` in codecs to avoid situation where encoding changes underneath us without our knowledge. It's unlikely for a `Word32` type, but if we could avoid it that would be better.

RenateEilers:
@coot

PR #4757, opened 2023-12-18
Adds unit test to check for phantom connections

bolt12:
# Description

This PR adds a unit test that is able to find the p2p bug if one adds a `threadDelay` to the outbound governor code

### DIFF
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
@@ -119,6 +119,8 @@ tests =
                  prop_diffusion_target_active_public
   , testProperty "target established local"
                  prop_diffusion_target_established_local
+  , testProperty "unit reconnect"
+                 prop_unit_reconnect
   , testProperty "target active local"
                  prop_diffusion_target_active_local
   , testProperty "target active root"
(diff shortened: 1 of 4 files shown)


ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -2614,6 +2616,140 @@ prop_unit_4258 =
...
+                (\case
+                   DiffusionConnectionManagerTransitionTrace tr -> Just tr
+                   _                                            -> Nothing
+                ) events
+
+       in conjoin
+        $ map (\(E ts a) -> case a of
+                TransitionTrace addr (Transition _ TerminatedSt) ->
+                  eventually ts (Set.notMember addr) govEstablishedPeersSig
+                _ -> True -- TODO: Do the opposite
```
coot, line 2750: What do you mean?

coot (approved):
LGTM

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
bolt12: That we should/could check that whenever there's a `Unnegotiated -> Negotiated Outbound` transition the peer is eventually added to the established set

coot: Please create an issue for it (an mention it in the todo).

coot: Then you can add it back to the merge queue.

PR #3699, opened 2022-04-06
Remove NodeToClientV_8 and below

coot:
[NodeToClientV_9](https://input-output-hk.github.io/ouroboros-network/ouroboros-network/Ouroboros-Network-NodeToClient.html#t:NodeToClientVersion)
was introduced before the Alonzo hard fork.

PR #4765, opened 2024-01-05
LocalStateQuery: let the client acquire the immutable tip

nfrisby:
# Description

We will use this to make it easier for the CLI to only output answers that are not subject to rollback.

This change to `LocalStateQuery` is gated by `NodeToClient_Version16`, since that is still mutable (it enables Conway et al).

It is a breaking change for any code that involves `MsgAcquire` or `MsgReAcquire`, since the type of the argument has changed to a new (more perspicuous and expressive) ADT.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/docs/network-spec/miniprotocols.tex
+++ b/docs/network-spec/miniprotocols.tex
@@ -1282,25 +1282,36 @@ ledger state.
 \end{figure}
 
 \paragraph{Protocol messages}
-% TODO description of each message
+
+See Figure~\ref{fig:lsq-messages}, in which $AcquireFailure$ is either
+$AcquireFailurePointTooOld$ or $AcquireFailurePointNotOnChain$, and $Target$ is
+either $ImmutableTip$, $VolatileTip$, or $SpecificPoint pt$.
+
+The primary motivation for being able to acquire the $ImmutableTip$ is that
+it's the most recent ledger state that the node will never abandon: the node
+will never rollback to prefix of that immutable chain (unless the on-disk
+ChainDB is corrupted/manipulated). Therefore, answers to queries against the
+$ImmutableTip$ are necessarily not subject rollback.
+
+% TODO natural language description of each message
+
 \begin{figure}[h]
   \begin{tabular}{|l|l|l|l|}
     \hline
     \multicolumn{4}{|c|}{Transition table} \\ \hline
     from state          & message             & parameters          & to state \\ \hline\hline
-    \StIdle             & \MsgAcquire         & $Maybe\ point$      & \StAcquiring \\\hline
+    \StIdle             & \MsgAcquire         & $Target\ point$     & \StAcquiring \\\hline
     \StAcquiring        & \MsgFailure         & $AcquireFailure$    & \StIdle      \\\hline
     \StAcquiring        & \MsgAcquired        &                     & \StAcquired  \\\hline
     \StAcquired         & \MsgQuery           & $query$             & \StQuerying  \\\hline
     \StQuerying         & \MsgResult          & $result$            & \StAcquired  \\\hline
-    \StAcquired         & \MsgReAcquire       & $Maybe\ point$      & \StAcquiring \\\hline
+    \StAcquired         & \MsgReAcquire       & $Target\ point$     & \StAcquiring \\\hline
     \StAcquired         & \MsgRelease         &                     & \StIdle      \\\hline
     \StIdle             & \MsgDone            &                     & \StDone        \\\hline
   \end{tabular}
   \caption{Local State Query mini-protocol messages.}
+  \label{fig:lsq-messages}
 \end{figure}
-where $AcquireFailure$ is either $AcquireFailurePointTooOld$ or
-$AcquireFailurePointNotOnChain$.
 
 \subsection{CDDL encoding specification}
 \lstinputlisting[style=cddl]{../../ouroboros-network-protocols/test-cddl/specs/local-state-query.cddl}
--- a/ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Codec.hs
+++ b/ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Codec.hs
@@ -24,6 +24,7 @@ import           Text.Printf
 
 import           Network.TypedProtocol.Codec.CBOR
 
+import qualified Ouroboros.Network.NodeToClient.Version as V
 import           Ouroboros.Network.Protocol.LocalStateQuery.Type
 
 
@@ -35,18 +36,23 @@ codecLocalStateQuery
      ( MonadST m
      , ShowQuery query
      )
-  => (point -> CBOR.Encoding)
+  => V.NodeToClientVersion
+     -- ^ eg whether to allow 'ImmutableTip' in @'MsgAcquire'
+  -> (point -> CBOR.Encoding)
   -> (forall s . CBOR.Decoder s point)
   -> (forall result . query result -> CBOR.Encoding)
   -> (forall s . CBOR.Decoder s (Some query))
   -> (forall result . query result -> result -> CBOR.Encoding)
   -> (forall result . query result -> forall s . CBOR.Decoder s result)
   -> Codec (LocalStateQuery block point query) CBOR.DeserialiseFailure m ByteString
-codecLocalStateQuery encodePoint  decodePoint
+codecLocalStateQuery version
+                     encodePoint  decodePoint
                      encodeQuery  decodeQuery
                      encodeResult decodeResult =
     mkCodecCborLazyBS encode decode
   where
+    canAcquireImmutable = version >= V.NodeToClientV_16
+
     encodeFailure :: AcquireFailure -> CBOR.Encoding
     encodeFailure AcquireFailurePointTooOld     = CBOR.encodeWord8 0
     encodeFailure AcquireFailurePointNotOnChain = CBOR.encodeWord8 1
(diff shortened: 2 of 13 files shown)


ouroboros-network-protocols/testlib/Ouroboros/Network/Protocol/LocalStateQuery/Test.hs
```diff
@@ -266,32 +284,9 @@ instance ( Arbitrary point
...
-        , pure $ AnyMessageAndAgencyWithResult $ AnyMessageAndAgency (ClientAgency TokIdle)
-            (MsgAcquire Nothing)
-
-        , pure $ AnyMessageAndAgencyWithResult $ AnyMessageAndAgency (ClientAgency TokAcquired)
-            (MsgReAcquire Nothing)
-        ]
-
--- Newtype wrapper which generates only valid data for 'NodeToClientV7' protocol.
---
-newtype AnyMessageAndAgencyV7 block point query result = AnyMessageAndAgencyV7 {
```
nfrisby, line 281: The lowest possible version is now 8. I'm removing this now because I suspect PR #3699 should have, and the code is simpler to maintain this way.

nfrisby:
The CI complains that "./ouroboros-network-api was modified but its CHANGELOG was not updated". But the change just a comment that does not deserve mention in the CHANELOG.

Do you have advise for how to proceed?

ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Codec.hs
```diff
@@ -35,14 +35,16 @@ codecLocalStateQuery
      ( MonadST m
      , ShowQuery query
      )
-  => (point -> CBOR.Encoding)
+  => Bool   -- ^ allow 'ImmutableTip' in @'MsgAcquire'
```
Jimbo4350, line 38: Is it possible to just pass the version instead? I see it's called in consensus and has access to the `NodeToClientVersion` or the `NodeToNodeVersion`.

ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Type.hs
```diff
@@ -68,6 +73,23 @@ instance ( ShowProxy block
       , showProxy (Proxy :: Proxy query)
       ]
 
+data Target point = -- | The tip of the volatile chain
```
Jimbo4350, line 76: Nice :+1:

ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Codec.hs
nfrisby: I suppose so. I primarily chose the `Bool`-approach because that's what was removed in PR 3699--just trying to match existing (albeit removed) patterns, since I'm not a codeowner here.

coot: I'd actually prefer to pass the version too.

Until now, when we modified a protocol we provided a new codec function or duplicated the protocol itself in some way.  Here we could generalise it and parametrise `LocalStateQuery` over the `Maybe` / `Target` type. Although this can be packaged so that it's type safe for the users, I am not sure if it's worth it.

coot (changes requested):
Very nice, thank you @nfrisby.

ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Codec.hs
nfrisby: Done! Thanks @Jimbo4350 and @coot .

I doubt parameterizing over `Maybe` versus `Target` is worth the noise, so I've skipped that. Moreover, as Jordan commented earlier: `Nothing` and `Just` were not great names :)

nfrisby:
For "If serialization changes, user-facing consequences (e.g. replay from genesis) are confirmed to be intentional.", here's some quotes from a Slack discussion with @ch1bo :

> [T]he serialization is backwards-compatible (I used a fresh tag for the new option, whereas the new data structure still has constructors that match the old `Nothing`/`Just` constructors and therefore use the same CBOR tag)

> My general assumption is that no one downstream of Consensus is caseing over these messages (other than maybe formatting log messages, etc). So your experience should mainly be: there's a new constructor you could choose to call, but you shouldn't if the handshake version negotiated with the server (ie the local node) is too low. (Like Marcin said, if you construct it even if the negotiated version is too low, you'll get an _encoder_ failure on the _client side_ (eg the wallet). If a buggy custom encoder ignores the version, the node will simply disconnect due to the ill-formed message.)

Summary:

| Server (ie node) | Client (eg wallet) | Notes |
| - | - | - |
| new | new | Both know about `ImmutableTip` ✔️ |
| old | old | Neither knows about `ImmutableTip` ✔️ |
| new | old | Server could handle `ImmutableTip`, but client won't send it ✔️ |
| old | new | Our encoders error on the client side if the client logic ignores the negotiated version ✔️; if a buggy custom encoder sent it through, the node would simply disconnect ✔️ |

Also: the existing messages' serialization wasn't changed, but their data constructor names were improved.

nfrisby:
I'm assuming the panoply of `darwin` failures is Hydra's fault, not mine.

ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Type.hs
```diff
@@ -68,6 +73,23 @@ instance ( ShowProxy block
...
+                    --
+                    -- Fails to be acquired if the point is not between
+                    -- 'VolatileTip' and 'ImmutableTip' (inclusive).
+                    SpecificPoint point
+                    -- | The tip of the immutable chain
+                    --
+                    -- Cannot fail to be acquired.
+                    --
+                    -- Requires at least 'NodeToClientV_16'.
+                  | ImmutableTip
```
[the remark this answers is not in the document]
coot: Good point, yet the `LocalStateQuery` mini-protocol is documented [here](https://github.com/IntersectMBO/ouroboros-network/blob/master/docs/network-spec/miniprotocols.tex#L1229).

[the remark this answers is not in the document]
nfrisby: Done, I think. Is there an easy way to see the built PDF in the CI outputs?

coot:
> I'm assuming the panoply of darwin failures is Hydra's fault, not mine.

It seems so, I restarted them.

docs/network-spec/miniprotocols.tex
```diff
@@ -1282,25 +1282,36 @@ \subsection{State machine}
...
     \StAcquiring        & \MsgAcquired        &                     & \StAcquired  \\\hline
     \StAcquired         & \MsgQuery           & $query$             & \StQuerying  \\\hline
     \StQuerying         & \MsgResult          & $result$            & \StAcquired  \\\hline
-    \StAcquired         & \MsgReAcquire       & $Maybe\ point$      & \StAcquiring \\\hline
+    \StAcquired         & \MsgReAcquire       & $Target\ point$     & \StAcquiring \\\hline
     \StAcquired         & \MsgRelease         &                     & \StIdle      \\\hline
     \StIdle             & \MsgDone            &                     & \StDone        \\\hline
   \end{tabular}
   \caption{Local State Query mini-protocol messages.}
+  \lable{fig:lsq-messages}
```
coot, line 1313: Could you provide either a definition of `Target` or a link it (it doesn't exist yet, but it will be `https://ouroboros-network.cardano.intersectmbo.org/ouroboros-network-protocols/Ouroboros-Network-Protocol-LocalStateQuery-Type.html#t:Target`)?

ouroboros-network-protocols/src/Ouroboros/Network/Protocol/LocalStateQuery/Type.hs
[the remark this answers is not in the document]
coot: I think so, but I don't know how to find it in `Hydra`.

ouroboros-network-protocols/test-cddl/Main.hs
```diff
@@ -429,6 +429,7 @@ localStateQueryCodec :: Codec (LocalStateQuery Block BlockPoint Query)
                               CBOR.DeserialiseFailure IO BL.ByteString
 localStateQueryCodec =
     codecLocalStateQuery
+      maxBound
```
coot, line 432: It's better if we use a random value.  This is not useful for the changes in this PR, but rather for possible future changes.

docs/network-spec/miniprotocols.tex
nfrisby: I did so in the text, lines 1287 and 1288. Is that not enough?

ouroboros-network-protocols/test-cddl/Main.hs
nfrisby: I don't know the architecture of that `test-cddl` code (would the properties also depend on the chosen version?), so I'd like to leave it as `maxBound` for now, kicking the can down the road to whoever introduces protocol versions for which this would be useful.

docs/network-spec/miniprotocols.tex
coot: I probably missed that, yes that's fine.

ouroboros-network-protocols/test-cddl/Main.hs
coot: I created an issue for it: https://github.com/IntersectMBO/ouroboros-network/issues/4772

coot (approved):

nfrisby:
I rebased and squashed my commits (and removed the merge commit that Marcin pushed).

PR #4778, opened 2024-01-17
Limit the rate with which one can discover all nodes

karknu:
# Description

The commented out function sampleAdvertisedPeers outlined an idea of limiting the rate with which one can discover all nodes in the network.

This implements that behaviour within computePeerSharingPeers. Given how cheap it is to create a new connection that can ask for peers the node will generate one list of peers regardless of who is asking. The list of peers will change after `policyPeerShareRetryTime` seconds. The list of peers shared does at most change by the number of peers added or removed. That is a newly added or removed peer can at most lead to one new peer beeing shared.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE KindSignatures      #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeOperators       #-}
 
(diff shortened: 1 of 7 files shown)


bolt12 (approved):

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
```diff
@@ -1217,20 +1221,39 @@ run tracers tracersExtra args argsExtra apps appsExtra = do
...
-computePeerSharingPeers readPublicState genVar amount = do
+computePeerSharingPeers readPublicState genVar reSaltAtVar PeerSelectionPolicy{..} amount = do
+  now <- getMonotonicTime
   publicState <- atomically readPublicState
-  gen <- atomically $ stateTVar genVar split
+  salt <- atomically $ do
+    reSaltAt <- readTVar reSaltAtVar
+    if reSaltAt <= now
+       then do
+         writeTVar reSaltAtVar $ addTime policyPeerShareRetryTime now
```
karknu, line 1247: The use of `policyPeerShareRetryTime ` interacts poorly with how we use it on the gosping side. Basically nodes will request new peers too early because the retry time is set when we send a request, not when the node gets a reply.
Will switch to a different timeout that is less than `policyPeerShareRetryTime ` .

coot: Could you create an issue for it.

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
```diff
@@ -1217,20 +1221,39 @@ run tracers tracersExtra args argsExtra apps appsExtra = do
...
+       else do
+         gen <- readTVar genVar
+         return $ fst $ random gen
 
   let availableToShareSet = availableToShare publicState
-      is = nub
-         $ take (fromIntegral amount)
-         $ randomRs (0, length availableToShareSet - 1) gen
-      randomList = map (`elemAt` availableToShareSet) is
+      randomList = take (min (fromIntegral policyPeerShareMaxPeers) (fromIntegral amount))
```
coot, line 1254: less brackets :) 
```suggestion
      randomList = take (fromIntegral policyPeerShareMaxPeers `min` fromIntegral amount)
```

coot (approved):
LGTM, just a minor comment.

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
karknu: No, I prefer to fix it in this PR.

PR #4788, opened 2024-01-23
Peersharing improvements

karknu:
# Description

Some Peersharing improvements. 
This PR fixes a bug where a peer with a `minPeerShareTime` could prevent us from peersharing with any other peer.
This PR changes the behavior so that we can start to gossip about incoming peers directly. Previously had to promote them to warm before we would let anyone else know about them.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
@@ -106,7 +106,7 @@ belowTarget actions
         decisionTrace = [TracePeerShareRequests
                           targetNumberOfKnownPeers
                           numKnownPeers
-                          availableForPeerShare
+                          canAsk
                           selectedForPeerShare],
         decisionState = st {
                           inProgressPeerShareReqs = inProgressPeerShareReqs
@@ -190,7 +190,7 @@ jobPeerShare PeerSelectionActions{requestPeerShare}
                -- sharing results is welcome.
                newPeers    = [ p | Right (PeerSharingResult ps) <- totalResults
                                  , p <- ps
-                                 , not (Set.member p (PublicRootPeers.toAllLedgerPeerSet (publicRootPeers st))) ]
+                                 , not (KnownPeers.member p (knownPeers st)) ]
             in Decision { decisionTrace = [ TracePeerShareResults peerResults
                                           , TracePeerShareResultsFiltered newPeers
                                           ]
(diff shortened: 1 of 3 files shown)


bolt12 (approved):
LGTM

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
```diff
@@ -112,13 +112,6 @@ belowTarget actions
           [jobPeerShare actions policy numPeersToReq (Set.toList selectedForPeerShare)]
       }
 
-    -- If we could peer share except that there are none currently available
-    -- then we return the next wakeup time (if any)
-  | numKnownPeers < targetNumberOfKnownPeers
-  , numPeerShareReqsPossible > 0
-  , Set.null availableForPeerShare
-  = GuardedSkip (EstablishedPeers.minPeerShareTime establishedPeers)
```
coot, line 120: Could you explain why this is not desirable?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
```diff
@@ -182,7 +175,7 @@ jobPeerShare PeerSelectionActions{requestPeerShare}
                -- sharing results is welcome.
                newPeers    = [ p | Right (PeerSharingResult ps) <- totalResults
                                  , p <- ps
-                                 , not (isKnownLedgerPeer p (knownPeers st)) ]
+                                 , not (KnownPeers.member p (knownPeers st)) ]
```
coot, line 178: nice catch

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
karknu: As I mentioned in the commit message _any_ peer with a peershare timer can block all peersharing. This meant that a node would only peershare 4 times per hour (15min timeout) even if it was below its targets and had more peers to ask.

coot: Makes sense. Probably the intention was to not peer share too often if we tried all the peers. But that's not very useful if we already limit how often we peer share with each peer (do we do that?).

karknu: We peershare at most once every 15 minutes with each individual peer.
But that only happens when we are bellow our targets, so ideally it should only be needed during churning.

PR #4806, opened 2024-02-15
Fix 'any Cold async demotion' test

bolt12:
# Description

CI found a test failure. I captured the seed:

```
cabal run ouroboros-network:sim-tests -- -p '/any Cold async demotion/' --quickcheck-replay=62992
```

It turns out the logic was flawed and needed a fix


ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -828,15 +828,17 @@ prop_track_coolingToCold_demotions defaultBearerInfo diffScript =
           decreasing [_] = True
           decreasing (x : y : t)
             | x > y     = decreasing (y : t)
+            | x == 0    = decreasing (y : t)
```
coot, line 831: This means we allow the set to increase its size from empty to non-empty, I'd add a comment.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -828,15 +828,17 @@ prop_track_coolingToCold_demotions defaultBearerInfo diffScript =
...
             | otherwise = False
 
        in property
         . decreasing
         . dropWhile (== 0)
```
coot, line 836: We don't need to `dropWhile (== 0)`, since `decreasing` will deal with it.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -828,15 +828,17 @@ prop_track_coolingToCold_demotions defaultBearerInfo diffScript =
...
-        . mapMaybe (fmap length . snd)
+        . mapMaybe snd
         . Signal.eventsToList
         . Signal.toChangeEvents
         $ Signal.stable
+        . fmap (fmap length)
```
coot, line 841: This change is only cosmetic, isn't it?

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
coot: I actually don't understand why `inProgressDemoteToCold` should only shrink in general? Is this test only applied to specific arguments / scenario?

bolt12: This test checks that the node doesn't keep any peers stuck in `inProgressDemoteToCold`. When a node in the testnet disconnects it is going to be added to the `inProgressDemoteToCold` set and the set will only grow after we are able to connect to it again and it disconnects again. We kind of rely on the sequential nature of IOSim but I couldn't find a simpler way to write the test

bolt12: The fix here makes sure that we ignore when the set goes to 0 elements and then grows back again.

coot (approved):
LGTM

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
coot: > This test checks that the node doesn't keep any peers stuck in inProgressDemoteToCold.

If that's what we want to test, maybe we should redesign the test to make sure none of the peers stays in the `inProgressDemoteToCold` for too long.  Is this expressible with the events API?

bolt12: It is definitely more tricky, but I believe it could be done

coot: Could you create & label an issue for this.

PR #4807, opened 2024-02-16
Adds targetNumberOfBootstrapPeers

bolt12:
This target is used when the node is in fallback state


ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -169,6 +169,9 @@ peerChurnGovernor tracer deadlineChurnInterval bulkChurnInterval psOverallTimeou
             decrease (targetNumberOfEstablishedBigLedgerPeers base -
                       targetNumberOfActiveBigLedgerPeers base)
             + targetNumberOfActiveBigLedgerPeers base
+        , targetNumberOfBootstrapPeers =
+            decrease (targetNumberOfBootstrapPeers base - targetNumberOfEstablishedPeers base)
```
karknu, line 173: This doesn't look right.
Plain `decrease (targetNumberOfBootstrapPeers base)` should do.
Target is 5, decrease by 20% gives 4.

bolt12: Nice catch :) I just mostly mimicked what `targetNumberRootPeers` was doing

karknu (approved):
LGTM

bolt12:
I'll close this in favor of #4810 since we no longer need a target for bootstrap peers

PR #4816, opened 2024-02-28
Fix redundant tracing if LSJ is not different

bolt12:


coot (commented):
LGTM

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -502,11 +502,10 @@ localRoots actions@PeerSelectionActions{ readLocalRootPeers
...
 
         $ Decision {
-            decisionTrace = [TraceLocalRootPeersChanged localRootPeers
-                                                        localRootPeers']
-                         ++ if not hasOnlyBootstrapPeers'
-                               then [TraceLedgerStateJudgementChanged YoungEnough]
-                               else [],
+            decisionTrace = TraceLocalRootPeersChanged localRootPeers localRootPeers'
+                          : [ TraceLedgerStateJudgementChanged YoungEnough
+                            | not hasOnlyBootstrapPeers' && ledgerStateJudgement /= ledgerStateJudgement'
```
coot, line 507: It should be `bootstrapPeersFlag && not hasOnlyBoostrapPeers' && ...` (or something similar) so that it works in `Genesis` too.

coot (changes requested):

PR #4835, opened 2024-03-19
Honour policyPeerShareActivationDelay timeout

karknu:
# Description
The policyPeerShareActivationDelay was effectivly ignored.
Commit 194ea32b removed the timeout logic because it was broken. This
was because availableForPeerShare contained a mix of peers that supported
and didn't support peersharing. As soon as a peer that doesn't support
peersharing is more than 900s old it would prevent any timeouts from
taking effect.
    
This commit change it so that only peers that support peersharing are
added to nextPeerShareTimes/availableForPeerShare. With that the
Guarskip related minPeerShareTime can be restored.

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
@@ -837,7 +837,7 @@ prop_track_coolingToCold_demotions defaultBearerInfo diffScript =
           notInProgressDemoteToColdForTooLong =
             map (\addr ->
                   Signal.keyedTimeoutTruncated
-                    60
+                    120
                     (\case
                         Just s | Set.member addr s -> Set.singleton addr
                         _                          -> Set.empty
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
@@ -78,10 +78,6 @@ belowTarget actions
     -- We can only ask ones where we have not asked them within a certain time.
   , not (Set.null availableForPeerShare)
 
-    -- Are there any peers from the availableForPeerShare set we can really ask
-    -- (i.e. have correct PeerSharing permissions)?
-  , not (Set.null canAsk)
-
     -- No peer share requests should be issued when the node is in a sensitive
     -- state
   , not (requiresBootstrapPeers bootstrapPeersFlag ledgerStateJudgement)
@@ -89,7 +85,7 @@ belowTarget actions
       -- Max selected should be <= numPeerShareReqsPossible
       selectedForPeerShare <- pickPeers st
                               policyPickKnownPeersForPeerShare
-                              canAsk
+                              availableForPeerShare
                               numPeerShareReqsPossible
 
       let -- Should be <= numPeerShareReqsPossible
@@ -109,7 +105,7 @@ belowTarget actions
         decisionTrace = [TracePeerShareRequests
                           targetNumberOfKnownPeers
                           numKnownPeers
-                          canAsk
+                          availableForPeerShare
                           selectedForPeerShare],
         decisionState = st {
                           inProgressPeerShareReqs = inProgressPeerShareReqs
@@ -125,6 +121,13 @@ belowTarget actions
              (Set.toList selectedForPeerShare)]
       }
 
+    -- If we could peer share except that there are none currently available
+    -- then we return the next wakeup time (if any)
+  | numKnownPeers < targetNumberOfKnownPeers
+  , numPeerShareReqsPossible > 0
+  , Set.null availableForPeerShare
+  = GuardedSkip $ EstablishedPeers.minPeerShareTime establishedPeers
+
   | otherwise
   = GuardedSkip Nothing
   where
@@ -132,11 +135,9 @@ belowTarget actions
                                    Set.\\ PublicRootPeers.getBigLedgerPeers publicRootPeers
     numPeerShareReqsPossible = policyMaxInProgressPeerShareReqs
                              - inProgressPeerShareReqs
+    -- Only peer which permit peersharing are available
     availableForPeerShare    = EstablishedPeers.availableForPeerShare establishedPeers
                              Set.\\ inProgressDemoteToCold
-    -- Only ask peers which have the correct willingness permission flags
-    canAsk                   =
-      KnownPeers.getPeerSharingRequestPeers availableForPeerShare knownPeers
 
 ---------------------------
 -- Peer sharing job
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
@@ -27,6 +27,7 @@ module Ouroboros.Network.PeerSelection.State.EstablishedPeers
   , minPeerShareTime
   , setPeerShareTime
   , availableForPeerShare
+  , nextPeerShareTimes
   , invariant
   ) where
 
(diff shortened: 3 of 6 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
```diff
@@ -78,18 +78,14 @@ belowTarget actions
     -- We can only ask ones where we have not asked them within a certain time.
   , not (Set.null availableForPeerShare)
 
-    -- Are there any peers from the availableForPeerShare set we can really ask
-    -- (i.e. have correct PeerSharing permissions)?
-  , not (Set.null canAsk)
-
```
bolt12, line 84: Why was this removed? the `availableForPeerShare` set contains peers which might have `PeerSharingDisabled` so we can't issue requests to them.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
```diff
@@ -78,18 +78,14 @@ belowTarget actions
...
     -- No peer share requests should be issued when the node is in a sensitive
     -- state
   , not (requiresBootstrapPeers bootstrapPeersFlag ledgerStateJudgement)
   = Guarded Nothing $ do
       -- Max selected should be <= numPeerShareReqsPossible
       selectedForPeerShare <- pickPeers st
                               policyPickKnownPeersForPeerShare
-                              canAsk
+                              availableForPeerShare
```
bolt12, line 88: This might pick peers which don't have peer sharing enabled resulting in a protocol failure

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
```diff
@@ -97,10 +97,11 @@ invariant :: Ord peeraddr
...
           -> Bool
 invariant EstablishedPeers {..} =
-       -- The combo of the peer share set + psq = the whole set of peers
-       availableForPeerShare
+       -- The combo of the peer share set + psq = a subset of all peers
+ (  availableForPeerShare
     <> Set.fromList (PSQ.keys nextPeerShareTimes)
-    == Map.keysSet allPeers
+ )
+    `Set.isSubsetOf` Map.keysSet allPeers
```
bolt12, line 104: I don't like the fact that this was weakened. Why do we need it?

bolt12 (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
```diff
@@ -162,13 +163,16 @@ member peeraddr = Map.member peeraddr . allPeers
...
-insert peeraddr peerconn peerShareAt ep@EstablishedPeers { allPeers } =
+insert peeraddr peerconn peerShareAt_m ep@EstablishedPeers { allPeers } =
    -- Ask newly established peers for its peers after the specified delay.
-   setPeerShareTime (Set.singleton peeraddr) peerShareAt $
-     ep { allPeers = Map.insert peeraddr peerconn allPeers }
+   let ep' = ep { allPeers = Map.insert peeraddr peerconn allPeers } in
+   case peerShareAt_m of
+        Nothing          -> ep'
+        Just peerShareAt ->
+          setPeerShareTime (Set.singleton peeraddr) peerShareAt $ ep'
```
coot, line 175: I think this is a bit cleaner (at the expense of code duplication, which could be eliminated with a recursive call):
```suggestion
insert peeraddr peerconn Nothing ep@EstablishedPeers { allPeers } =
    ep { allPeers = Map.insert peeraddr peerconn allPeers }   
insert peeraddr peerconn (Just peerShareAt) ep@EstablishedPeers { allPeers } =
    -- Ask newly established peers for its peers after the specified delay.
    setPeerShareTime (Set.singleton peeraddr) $
    ep { allPeers = Map.insert peeraddr peerconn allPeers } 
```

coot (approved):
LGTM, just a minor comment.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
coot: As I understand it it cannot, because we only add peers which enabled peer sharing to `availableForPeerShare`.

karknu: I changed it so that `availableForPeerShare` only contains peers that actually support peersharing. This removes the need for `canAsk`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
```diff
@@ -162,13 +163,16 @@ member peeraddr = Map.member peeraddr . allPeers
 insert :: Ord peeraddr
        => peeraddr
        -> peerconn
-       -> Time
+       -> Maybe Time
```
coot, line 166: Could you add a haddock; it's important to use this argument right.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/KnownPeers.hs
karknu: No, see previous comment regarding `availableForPeerShare`.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -837,7 +837,7 @@ prop_track_coolingToCold_demotions defaultBearerInfo diffScript =
           notInProgressDemoteToColdForTooLong =
             map (\addr ->
                   Signal.keyedTimeoutTruncated
-                    60
+                    120
```
coot, line 840: Could you explain why we need to increase it?

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
karknu: It is needed because `allPeers` contains both peers that support and doesn't support peersharing, but `availableForPeerShare` and `nextPeerShareTimes` only contains peers that support peersharing.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
bolt12: There was a test failure. I looked into it and it was due to the timeout just barely making it.

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
bolt12: I would add a clause to the invariant that checks that all peers in the `availableForPeerShare` set have peer sharing enabled then.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
bolt12: It isn't

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
karknu: :+1: Will do.

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
karknu: I've added a check which makes sure that only peers which support peersharing are in `availableForPeerShare <> nextPeerShareTimes` .

bolt12 (approved):

PR #4836, opened 2024-03-19
Extend and expose PeerSelectionCounters

### LINKED ISSUE #4815: Inform Consensus Layer whether a bootstrap peer should be considered remote
@karknu had raised a concern about the Bootstrap State Machine (ie https://github.com/IntersectMBO/ouroboros-consensus/pull/808) that would be most naturally mitigated by the Diffusion Layer annotating new peers with whether or not that peer should be considered remote.

EG SPOs separately configure the "local trusted" portion of the topology from the "bootstrap peers source" of the topology. If that distinction would be preserved deep enough into the code, then the Bootstrap State Machine would simply refuse to transition from TooOld to YoungEnough while it only has local peers.

bolt12:
# Description

Closes #4815

### DIFF
--- a/ouroboros-network-api/CHANGELOG.md
+++ b/ouroboros-network-api/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Non-Breaking changes
 
+* Added `OnlyLocalOutboundConnections` data type
+
 ## 0.7.1.0 -- 2024-03-14
 
 ### Breaking changes
--- a/ouroboros-network-api/src/Ouroboros/Network/PeerSelection/Bootstrap.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/PeerSelection/Bootstrap.hs
@@ -6,6 +6,7 @@ module Ouroboros.Network.PeerSelection.Bootstrap
   , isBootstrapPeersEnabled
   , requiresBootstrapPeers
   , isNodeAbleToMakeProgress
+  , OnlyLocalOutboundConnections (..)
   ) where
 
 import GHC.Generics (Generic)
@@ -41,3 +42,7 @@ requiresBootstrapPeers ubp lsj =
 isNodeAbleToMakeProgress :: UseBootstrapPeers -> LedgerStateJudgement -> Bool -> Bool
 isNodeAbleToMakeProgress ubp lsj hasOnlyBootstrapPeers =
     not (requiresBootstrapPeers ubp lsj) || hasOnlyBootstrapPeers
+
+data OnlyLocalOutboundConnections = ConnectedToExternalOutboundPeers
+                                  | ConnectedToOnlyLocalOutboundPeers
+  deriving (Eq, Show)
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## next version
 
+### Breaking changes
+
+* Added `daUpdateOnlyLocalConnections :: OnlyLocalOutboundConnections -> STM m ()`
+  to `Diffusion.Common.Applications`. This callback is to be provided by
+  consensus and is propagated all the way to the peer selection governor.
+
+### Non-Breaking changes
+
+* Extended and exposed `PeerSelectionCounters` to the Churn governor.
+* Improved Churn governor by synchronizing according to the counters instead
+  of relying on `threadDelay`.
+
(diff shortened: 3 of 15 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Context.hs
```diff
@@ -22,7 +22,8 @@ import Ouroboros.Network.PeerSelection.LedgerPeers.Type
 data ExpandedInitiatorContext addr m = ExpandedInitiatorContext {
     eicConnectionId    :: !(ConnectionId addr),
     eicControlMessage  :: !(ControlMessageSTM m),
-    eicIsBigLedgerPeer :: !IsBigLedgerPeer
+    eicIsBigLedgerPeer :: !IsBigLedgerPeer,
+    eicIsLocalRootPeer :: !Bool
```
coot, line 26: If I recall we are interested in marking local trusted peers, not just local ones. Am I right?  If that's true the field should be `eicIsLocalTrustedRootPeer`.

We should also provide a data type, instead a boolean:
```
data IsTrustedLocalRootPeer
 ...
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/ActivePeers.hs
```diff
@@ -115,8 +116,10 @@ belowTargetBigLedgerPeers actions
                           inProgressPromoteWarm = inProgressPromoteWarm
                                                <> selectedToPromote
                         },
-        decisionJobs  = [ jobPromoteWarmPeer actions policy peeraddr IsBigLedgerPeer peerconn
-                        | (peeraddr, peerconn) <- Map.assocs selectedToPromote' ]
+        decisionJobs  = [ jobPromoteWarmPeer actions policy peeraddr IsBigLedgerPeer isLocalRootPeer peerconn
+                        | (peeraddr, peerconn) <- Map.assocs selectedToPromote'
+                        , let isLocalRootPeer = LocalRootPeers.member peeraddr localRootPeers
```
coot, line 121: We need to check if the peer is a trusted local root peer; the same in other places.

coot (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/ActivePeers.hs
bolt12: I think checking local root is enough because when the information is going to be used the peer will guaranteed be a trusted one. As far as consensus cares it just wants to know if it is local root or external peer. So I think it is enough to check if it is local root

coot: You're right it's enough for us to check that it's a local root peer.

coot (changes requested):
The `withDelayCancellable` & `modifyPeersBasedOnCounter` contain a bug: they update the targets and wait on the changes in the same stm transaction - changes in a single stm tx are not visible to the outside world until the transaction is committed, as a result we never update the targets.

I think it will be much better to write this code in a different way.  We need a function which:
* update the targets to the specified value
* then block on the current counters until they reach the
So I suggest that we implement:
```hs
updateTargets :: MonadSTM m
              => Tracer m (TracePeerSelection peeraddr)
              -> StrictTVar m PeerSelectionTargets
              -> StrictTVar m PeerSelectionCounters
              -> DiffTime
              -- ^ timeout
              -> (PeerSelectionTargets -> PeerSelectionTargets)
              -- ^ update counters function
              -> m ()
```
it should look something like (Haskell pseudocode):
```hs
updateTargets tracer peerSelectionVar countersVar timeout modifyTargets = do
 -- update targets
 targets <- atomically $ do modifyTVar peerSelectionVar modifyTargets
                            readTVar peerSelectionVar
 -- create timeout
 delayVar <- registerDelay timeout
 -- block until counters reached the targets, or the timeout fires
 atomically $ runFirstToFinish $
   FirstToFinish (readTVar countersVar >>= check . didCountersReachedTargets targets)
   <>
   FIrstToFinish (readTVar delayVar >>= check)               
```

Note that it runs in two `STM` transactions, so the outbound governor will have a chance to notice and act on modified targets.

coot:
Thanks @bolt12, I am closing it in favour of #4846 which was based on this PR.

Issue #4845, opened 2024-04-05
PeerSharing counters


coot:
We need to make peer sharing more visible.  For that we should add counters how many known / warm / hot peers were received through peer sharing.

PR #4846, opened 2024-04-06
Expose OutboundConnectionState for consensus

### LINKED ISSUE #4815: Inform Consensus Layer whether a bootstrap peer should be considered remote

coot:
# Description

This PR subsumes #4836.  It provides API for Genesis which allow consensus to
know if the node is connected just to local root peers or has any external
peers.  Fixes #4815.

TODO:
* [x] - add a basic churn tests
* [x] - #4845

# Checklist

- Branch
- Pull Request
    - [x] Self-reviewed the diff
      - What does this PR change?
    - [x] Reviewer requested

### DIFF
--- a/ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
@@ -1,0 +1,21 @@
+{-# LANGUAGE DeriveGeneric #-}
+
+module Ouroboros.Network.PeerSelection.LocalRootPeers (OutboundConnectionsState (..)) where
+
+import GHC.Generics
+import NoThunks.Class
+
+data OutboundConnectionsState =
+    TrustedStateWithExternalPeers
+    -- ^
+    -- * /in the Praos mode/: connected only to trusted local
+    --   peers and at least one bootstrap peer or public root;
+    -- * /in the Genesis mode/: meeting target of active big ledger peers;
+    -- * or it is in `Unrestricted` mode
+    --   (see `Ouroboros.Network.PeerSelection.Governor.AssociationMode`).
+
+  | UntrustedState
+    -- ^ catch all other cases
+  deriving (Eq, Show, Generic)
+
+instance NoThunks OutboundConnectionsState
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
@@ -3487,8 +3487,14 @@ _governorFindingPublicRoots :: Int
                             -> StrictTVar IO OutboundConnectionsState
                             -> IO Void
 _governorFindingPublicRoots targetNumberOfRootPeers readDomains readUseBootstrapPeers readLedgerStateJudgement peerSharing olocVar = do
+    countersVar <- newTVarIO emptyPeerSelectionCounters
+    publicStateVar <- makePublicPeerSelectionStateVar
+    debugStateVar <- newTVarIO $ emptyPeerSelectionState (mkStdGen 42)
     dnsSemaphore <- newLedgerAndPublicRootDNSSemaphore
     let interfaces = PeerSelectionInterfaces {
+            countersVar,
+            publicStateVar,
+            debugStateVar,
             readUseLedgerPeers = return DontUseLedgerPeers
           }
     publicRootPeersProvider
@@ -3498,16 +3504,10 @@ _governorFindingPublicRoots targetNumberOfRootPeers readDomains readUseBootstrap
       DNS.defaultResolvConf
       readDomains
       (ioDNSActions LookupReqAAndAAAA) $ \requestPublicRootPeers -> do
-        publicStateVar <- makePublicPeerSelectionStateVar
-        debugVar <- newTVarIO $ emptyPeerSelectionState (mkStdGen 42)
-        countersVar <- newTVarIO emptyPeerSelectionCounters
         peerSelectionGovernor
           tracer tracer tracer
           -- TODO: #3182 Rng seed should come from quickcheck.
           (mkStdGen 42)
-          countersVar
-          publicStateVar
-          debugVar
           actions
             { requestPublicRootPeers = \_ ->
                 transformPeerSelectionAction requestPublicRootPeers }
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
@@ -190,6 +190,7 @@ tests =
       , testProperty "node uses ledger peers in non-sensitive mode"
                      prop_governor_uses_ledger_peers
       ]
+    , testProperty "association mode" prop_governor_association_mode
     ]
   , testGroup "issues"
     [ testProperty "3233" prop_issue_3233
@@ -761,6 +762,7 @@ envEventCredits  TraceEnvActivatePeer {}        = 0
 envEventCredits  TraceEnvDeactivatePeer {}      = 0
 envEventCredits  TraceEnvCloseConn {}           = 0
 
+envEventCredits  TraceEnvUseLedgerPeers {}      = 30
 envEventCredits  TraceEnvSetLedgerStateJudgement {} = 30
 
 envEventCredits  TraceEnvSetUseBootstrapPeers {} = 30
@@ -3426,6 +3428,79 @@ prop_governor_uses_ledger_peers env =
      in counterexample (intercalate "\n" $ map show $ usesLedgerPeers)
       $ all snd usesLedgerPeers
 
+
+prop_governor_association_mode :: GovernorMockEnvironment -> Property
+prop_governor_association_mode env =
+    let events = Signal.eventsFromListUpToTime (Time (10 * 60 * 60))
+               . selectPeerSelectionTraceEvents
+               . runGovernorInMockEnvironment
+               $ env
+
+        counters :: Signal (PeerSelectionSetsWithSizes PeerAddr)
+        counters =
+          selectGovState peerSelectionStateToView events
+
+        -- accumulate local roots
+        localRoots :: Signal (Set PeerAddr)
+        localRoots =
+            Signal.keyedUntil
+              (\case
+                Just (GovernorEvent (TraceLocalRootPeersChanged a _)) -> LocalRootPeers.keysSet a
+                Just (MockEnvEvent (TraceEnvSetLocalRoots a)) -> LocalRootPeers.keysSet a
+                _ -> Set.empty
+              )
+              (\_ -> Set.empty)
+              (\_ -> False)
+          . Signal.fromChangeEvents Nothing
+          . fmap Just
+          $ events
+
+        publicRoots :: Signal (Set PeerAddr)
+        publicRoots =
+            Signal.keyedUntil
+              PublicRootPeers.toSet
+              (\_ -> Set.empty)
+              (\_ -> False)
+          . selectGovState Governor.publicRootPeers
+          $ events
+
+        associationMode :: Signal AssociationMode
+        associationMode =
+            Signal.fromChangeEvents Unrestricted
+          $ selectGovAssociationMode events
+
+    in counterexample (intercalate "\n" $ show <$> Signal.eventsToList events)
+     $ signalProperty 20 show
+        (\(cs, localRootSet, publicRootSet, am) ->
+          case am of
+            LocalRootsOnly ->
+              -- we need to remove local and public roots.  They are changing
+              -- over time, and a node might keep using them, event though the
+              -- node is configured as an `Unrestricted` node.
+              --
+              -- This makes this test only effective if a node starts in
+              -- `LocalRootsOnly` mode, until it is reconfigured.  This can
+              -- discover some bugs in `readAssociationMode` but certainly not
+              -- all.
+              --
+              -- TODO: write a more effective test.
+                 Set.null (fst (viewKnownBootstrapPeers cs)
+                            Set.\\ localRootSet
+                            Set.\\ publicRootSet)
+              && Set.null (fst (viewKnownBigLedgerPeers cs)
+                            Set.\\ localRootSet
+                            Set.\\ publicRootSet)
+              && Set.null (fst (viewKnownNonRootPeers cs)
+                            Set.\\ localRootSet
+                            Set.\\ publicRootSet)
+
+            Unrestricted -> True
+        )
+        ((,,,) <$> counters
+               <*> localRoots
+               <*> publicRoots
+               <*> associationMode)
+
 --
 -- Utils for properties
 --
@@ -3443,6 +3518,18 @@ selectGovEvents :: Events TestTraceEvent
 selectGovEvents = Signal.selectEvents
                     (\case GovernorEvent e -> Just $! e
                            _               -> Nothing)
+
+selectGovCounters :: Events TestTraceEvent
+                  -> Events PeerSelectionCounters
+selectGovCounters = Signal.selectEvents
+                      (\case GovernorCounters e -> Just $! e
+                             _                  -> Nothing)
+
+selectGovAssociationMode :: Events TestTraceEvent
+                         -> Events AssociationMode
+selectGovAssociationMode = Signal.selectEvents
+                             (\case GovernorAssociationMode e -> Just $! e
+                                    _                         -> Nothing)
 
 selectGovState :: Eq a
                => (forall peerconn. Governor.PeerSelectionState PeerAddr peerconn -> a)
@@ -3601,6 +3688,7 @@ prop_issue_3550 = prop_governor_target_established_below defaultMaxTime $
       pickColdPeersToForget = Script (PickFirst :| []),
       peerSharingFlag = PeerSharingEnabled,
       useBootstrapPeers = Script ((DontUseBootstrapPeers, NoDelay) :| []),
+      useLedgerPeers = Script ((UseLedgerPeers Always, NoDelay) :| []),
       ledgerStateJudgement = Script ((YoungEnough, NoDelay) :| [])
     }
 
@@ -3637,6 +3725,7 @@ prop_issue_3515 = prop_governor_nolivelock $
       pickColdPeersToForget = Script (PickFirst :| []),
       peerSharingFlag = PeerSharingEnabled,
       useBootstrapPeers = Script ((DontUseBootstrapPeers, NoDelay) :| []),
+      useLedgerPeers = Script ((UseLedgerPeers Always, NoDelay) :| []),
       ledgerStateJudgement = Script ((YoungEnough, NoDelay) :| [])
     }
 
@@ -3673,6 +3762,7 @@ prop_issue_3494 = prop_governor_nofail $
       pickColdPeersToForget = Script (PickFirst :| []),
       peerSharingFlag = PeerSharingEnabled,
       useBootstrapPeers = Script ((DontUseBootstrapPeers, NoDelay) :| []),
+      useLedgerPeers = Script ((UseLedgerPeers Always, NoDelay) :| []),
       ledgerStateJudgement = Script ((YoungEnough, NoDelay) :| [])
     }
 
@@ -3725,6 +3815,7 @@ prop_issue_3233 = prop_governor_nolivelock $
       pickColdPeersToForget = Script (PickFirst :| []),
       peerSharingFlag = PeerSharingEnabled,
       useBootstrapPeers = Script ((DontUseBootstrapPeers, NoDelay) :| []),
+      useLedgerPeers = Script ((UseLedgerPeers Always, NoDelay) :| []),
       ledgerStateJudgement = Script ((YoungEnough, NoDelay) :| [])
     }
 
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
@@ -31,6 +31,7 @@ module Test.Ouroboros.Network.PeerSelection
 
 import Control.Concurrent.Class.MonadSTM.Strict
 import Control.Exception (AssertionFailed (..), catch, evaluate)
+import Control.Monad (when)
 import Control.Monad.Class.MonadTime.SI
 import Control.Monad.Class.MonadTimer.SI
 import Control.Tracer (Tracer (..))
@@ -62,6 +63,7 @@ import Ouroboros.Network.PeerSelection.Governor hiding (PeerSelectionState (..),
            peerSharing)
 import Ouroboros.Network.PeerSelection.Governor qualified as Governor
 import Ouroboros.Network.PeerSelection.LedgerPeers
+import Ouroboros.Network.PeerSelection.LocalRootPeers (OutboundConnectionsState)
 import Ouroboros.Network.PeerSelection.PeerAdvertise
 import Ouroboros.Network.PeerSelection.PeerSharing (PeerSharing (..))
 import Ouroboros.Network.PeerSelection.PeerTrustable (PeerTrustable (..))
@@ -3482,9 +3484,13 @@ _governorFindingPublicRoots :: Int
                             -> STM IO UseBootstrapPeers
                             -> STM IO LedgerStateJudgement
                             -> PeerSharing
+                            -> StrictTVar IO OutboundConnectionsState
                             -> IO Void
-_governorFindingPublicRoots targetNumberOfRootPeers readDomains readUseBootstrapPeers readLedgerStateJudgement peerSharing = do
+_governorFindingPublicRoots targetNumberOfRootPeers readDomains readUseBootstrapPeers readLedgerStateJudgement peerSharing olocVar = do
     dnsSemaphore <- newLedgerAndPublicRootDNSSemaphore
+    let interfaces = PeerSelectionInterfaces {
+            readUseLedgerPeers = return DontUseLedgerPeers
+          }
     publicRootPeersProvider
       tracer
       (curry IP.toSockAddr)
@@ -3506,6 +3512,7 @@ _governorFindingPublicRoots targetNumberOfRootPeers readDomains readUseBootstrap
             { requestPublicRootPeers = \_ ->
                 transformPeerSelectionAction requestPublicRootPeers }
           policy
+          interfaces
   where
     tracer :: Show a => Tracer IO a
     tracer  = Tracer (BS.putStrLn . BS.pack . show)
@@ -3527,8 +3534,12 @@ _governorFindingPublicRoots targetNumberOfRootPeers readDomains readUseBootstrap
                   closePeerConnection      = error "closePeerConnection"
                 },
                 readUseBootstrapPeers,
-                readLedgerStateJudgement
-              }
+                readLedgerStateJudgement,
+                updateOutboundConnectionsState = \a -> do
+                  a' <- readTVar olocVar
+                  when (a /= a') $
+                    writeTVar olocVar a
+                              }
 
     targets :: PeerSelectionTargets
     targets = nullPeerSelectionTargets {
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
@@ -1,8 +1,14 @@
 {-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE MultiWayIf          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+
+#if __GLASGOW_HASKELL__ < 904
+{-# OPTIONS_GHC -Wno-name-shadowing #-}
+#endif
 
 -- | This subsystem manages the discovery and selection of /upstream/ peers.
 --
@@ -14,10 +20,13 @@ module Ouroboros.Network.PeerSelection.Governor
     PeerSelectionPolicy (..)
   , PeerSelectionTargets (..)
   , PeerSelectionActions (..)
+  , PeerSelectionInterfaces (..)
   , PeerStateActions (..)
   , TracePeerSelection (..)
   , ChurnAction (..)
   , DebugPeerSelection (..)
+  , AssociationMode (..)
+  , readAssociationMode
   , DebugPeerSelectionState (..)
   , peerSelectionGovernor
     -- * Peer churn governor
@@ -43,6 +52,7 @@ module Ouroboros.Network.PeerSelection.Governor
 
 import Data.Foldable (traverse_)
 import Data.Hashable
+import Data.Set qualified as Set
 import Data.Void (Void)
 
 import Control.Applicative (Alternative ((<|>)))
@@ -56,6 +66,7 @@ import Control.Monad.Class.MonadTimer.SI
 import Control.Tracer (Tracer (..), traceWith)
 import System.Random
 
+import Ouroboros.Network.PeerSelection.Bootstrap (UseBootstrapPeers (..))
 import Ouroboros.Network.PeerSelection.Churn (ChurnCounters (..),
            peerChurnGovernor)
 import Ouroboros.Network.PeerSelection.Governor.ActivePeers qualified as ActivePeers
@@ -65,8 +76,13 @@ import Ouroboros.Network.PeerSelection.Governor.KnownPeers qualified as KnownPee
 import Ouroboros.Network.PeerSelection.Governor.Monitor qualified as Monitor
 import Ouroboros.Network.PeerSelection.Governor.RootPeers qualified as RootPeers
 import Ouroboros.Network.PeerSelection.Governor.Types
+import Ouroboros.Network.PeerSelection.LedgerPeers.Type (UseLedgerPeers (..))
+import Ouroboros.Network.PeerSelection.LocalRootPeers
+           (OutboundConnectionsState (..))
+import Ouroboros.Network.PeerSelection.PeerSharing (PeerSharing (..))
 import Ouroboros.Network.PeerSelection.State.EstablishedPeers qualified as EstablishedPeers
 import Ouroboros.Network.PeerSelection.State.KnownPeers qualified as KnownPeers
+import Ouroboros.Network.PeerSelection.State.LocalRootPeers qualified as LocalRootPeers
 
 {- $overview
 
@@ -459,8 +475,11 @@ peerSelectionGovernor :: ( Alternative (STM m)
                       -> StrictTVar m (PeerSelectionState peeraddr peerconn)
                       -> PeerSelectionActions peeraddr peerconn m
                       -> PeerSelectionPolicy  peeraddr m
+                      -> PeerSelectionInterfaces m
                       -> m Void
-peerSelectionGovernor tracer debugTracer countersTracer fuzzRng countersVar publicStateVar debugStateVar actions policy =
+peerSelectionGovernor tracer debugTracer countersTracer fuzzRng
+                      countersVar publicStateVar debugStateVar
+                      actions policy interfaces =
     JobPool.withJobPool $ \jobPool ->
       peerSelectionGovernorLoop
         tracer
@@ -471,6 +490,7 @@ peerSelectionGovernor tracer debugTracer countersTracer fuzzRng countersVar publ
         debugStateVar
         actions
         policy
+        interfaces
         jobPool
         (emptyPeerSelectionState fuzzRng)
 
@@ -507,6 +527,7 @@ peerSelectionGovernorLoop :: forall m peeraddr peerconn.
                           -> StrictTVar m (PeerSelectionState peeraddr peerconn)
                           -> PeerSelectionActions peeraddr peerconn m
                           -> PeerSelectionPolicy  peeraddr m
+                          -> PeerSelectionInterfaces m
                           -> JobPool () m (Completion m peeraddr peerconn)
                           -> PeerSelectionState peeraddr peerconn
                           -> m Void
@@ -518,6 +539,7 @@ peerSelectionGovernorLoop tracer
                           debugStateVar
                           actions
                           policy
+                          interfaces
                           jobPool
                           pst = do
     loop pst (Time 0) `catch` (\e -> traceWith tracer (TraceOutboundGovernorCriticalFailure e) >> throwIO e)
@@ -558,13 +580,23 @@ peerSelectionGovernorLoop tracer
       -- get the current time after the governor returned from the blocking
       -- 'evalGuardedDecisions' call.
       now <- getMonotonicTime
-      let Decision { decisionTrace, decisionJobs, decisionState } =
+
+      let Decision { decisionTrace, decisionJobs, decisionState = st'' } =
             timedDecision now
 
       mbCounters <- atomically $ do
+        -- Update outbound connections state
+        let peerSelectionView = peerSelectionStateToView st''
+        associationMode <- readAssociationMode (readUseLedgerPeers interfaces)
+                                               (peerSharing actions)
+                                               (bootstrapPeersFlag st'')
+        updateOutboundConnectionsState
+          actions
+          (outboundConnectionsState associationMode peerSelectionView st'')
+
         -- Update counters
         counters <- readTVar countersVar
-        let !counters' = peerSelectionStateToCounters decisionState
+        let !counters' = snd <$> peerSelectionView
         if counters' /= counters
           then writeTVar countersVar counters'
             >> return (Just counters')
@@ -572,11 +604,11 @@ peerSelectionGovernorLoop tracer
 
       -- Trace counters
       traverse_ (traceWith countersTracer) mbCounters
-
+      -- Trace peer selection
       traverse_ (traceWith tracer) decisionTrace
 
       mapM_ (JobPool.forkJob jobPool) decisionJobs
-      loop decisionState dbgUpdateAt'
+      loop st'' dbgUpdateAt'
 
     evalGuardedDecisions :: Time
                          -> PeerSelectionState peeraddr peerconn
@@ -667,3 +699,88 @@ wakeupDecision st _now =
     decisionState = st,
     decisionJobs  = []
   }
+
+
+-- | Classify if a node is in promiscuous mode.
+--
+-- A node is not in promiscuous mode only if: it doesn't use ledger peers, peer
+-- sharing, the set of bootstrap peers is empty.
+--
+readAssociationMode
+  :: MonadSTM m
+  => STM m UseLedgerPeers
+  -> PeerSharing
+  -> UseBootstrapPeers
+  -> STM m AssociationMode
+readAssociationMode
+  readUseLedgerPeers
+  peerSharing
+  useBootstrapPeers
+  =
+  do useLedgerPeers <- readUseLedgerPeers
+     pure $
+       case (useLedgerPeers, peerSharing, useBootstrapPeers) of
+         (DontUseLedgerPeers, PeerSharingDisabled, DontUseBootstrapPeers)
+           -> LocalRootsOnly
+         (DontUseLedgerPeers, PeerSharingDisabled, UseBootstrapPeers config)
+           |  null config
+           -> LocalRootsOnly
+         _ -> Unrestricted
+
+
+outboundConnectionsState
+    :: Ord peeraddr
+    => AssociationMode
+    -> PeerSelectionSetsWithSizes peeraddr
+    -> PeerSelectionState peeraddr peerconn
+    -> OutboundConnectionsState
+outboundConnectionsState
+    associationMode
+    PeerSelectionView {
+      viewEstablishedPeers          = (viewEstablishedPeers, _),
+      viewEstablishedBootstrapPeers = (viewEstablishedBootstrapPeers, _),
+      viewActiveBootstrapPeers      = (viewActiveBootstrapPeers, _)
+    }
+    PeerSelectionState {
+      localRootPeers,
+      bootstrapPeersFlag
+    }
+    =
+    case (associationMode, bootstrapPeersFlag) of
+      {-
+       -- genesis mode
+       -- TODO: issue #4846
+      (LocalRootsOnly, _)
+        |  numberOfActiveBigLedgerPeers >= targetNumberOfActiveBigLedgerPeers
+        -> TrustedStateWithExternalPeers
+
+        |  otherwise
+        -> UntrustedState
+      -}
+
+      (LocalRootsOnly, _)
+        |  -- we are only connected to trusted local root
+           -- peers
+           viewEstablishedPeers `Set.isSubsetOf` trustableLocalRootSet
+        -> TrustedStateWithExternalPeers
+
+        |  otherwise
+        -> UntrustedState
+
+       -- bootstrap mode
+      (Unrestricted, UseBootstrapPeers {})
+        |  -- we are only connected to trusted local root
+           -- peers or bootstrap peers
+           viewEstablishedPeers `Set.isSubsetOf` (viewEstablishedBootstrapPeers <> trustableLocalRootSet)
+           -- there's at least one active bootstrap peer
+        ,  not (Set.null viewActiveBootstrapPeers)
+        -> TrustedStateWithExternalPeers
+
+        |  otherwise
+        -> UntrustedState
+
+       -- praos mode with public roots
+      (Unrestricted, DontUseBootstrapPeers)
+        -> UntrustedState
+  where
+    trustableLocalRootSet = LocalRootPeers.trustableKeysSet localRootPeers
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
@@ -470,24 +470,17 @@ peerSelectionGovernor :: ( Alternative (STM m)
                       -> Tracer m (DebugPeerSelection peeraddr)
                       -> Tracer m PeerSelectionCounters
                       -> StdGen
-                      -> StrictTVar m PeerSelectionCounters
-                      -> StrictTVar m (PublicPeerSelectionState peeraddr)
-                      -> StrictTVar m (PeerSelectionState peeraddr peerconn)
                       -> PeerSelectionActions peeraddr peerconn m
                       -> PeerSelectionPolicy  peeraddr m
-                      -> PeerSelectionInterfaces m
+                      -> PeerSelectionInterfaces peeraddr peerconn m
                       -> m Void
 peerSelectionGovernor tracer debugTracer countersTracer fuzzRng
-                      countersVar publicStateVar debugStateVar
                       actions policy interfaces =
     JobPool.withJobPool $ \jobPool ->
       peerSelectionGovernorLoop
         tracer
         debugTracer
         countersTracer
-        countersVar
-        publicStateVar
-        debugStateVar
         actions
         policy
         interfaces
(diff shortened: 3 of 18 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -61,180 +77,369 @@ peerChurnGovernor tracer deadlineChurnInterval bulkChurnInterval psOverallTimeou
...
+         numberOfKnownPeers <= targetNumberOfKnownPeers
+      && numberOfKnownBigLedgerPeers <= targetNumberOfKnownBigLedgerPeers
+
+
+    increaseKnownPeers
+      :: ModifyPeerSelectionTargets
+    increaseKnownPeers targets =
+      targets {
+          targetNumberOfRootPeers =
+            decrease (targetNumberOfRootPeers base - targetNumberOfEstablishedPeers base)
```
crocodile-dentist, line 338: oops ;) I think this should be targetNumberOfRootPeers = targetNumberOfRootPeers base
and analogously below:
targetNumberOfKnownPeers = targetNumberOfKnownPeers base
targetNumberOfKnownBigLedgerPeers = targetNumberOfKnownBigLedgerPeers base

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -61,180 +77,369 @@ peerChurnGovernor tracer deadlineChurnInterval bulkChurnInterval psOverallTimeou
...
+      -- ^ update counters function
+      -> CheckPeerSelectionCounters
+      -- ^ check counters
+      -> m ()
+    updateTargets churnActions timeoutDelay modifyTargets checkCounters = do
+      -- update targets, and return the new targets
+      targets <- atomically $ stateTVar peerSelectionVar ((\a -> (a, a)) . modifyTargets)
+
+      -- create timeout and block on counters
+      bracketOnError (registerDelayCancellable timeoutDelay)
```
crocodile-dentist, line 114: I was wondering if bracket can be used here instead to slightly simply things. We could then avoid calling cancelTimeout in the true branch below, but then the question is what would happen in the finalizer if timeout expired, ie. we called cancelTimeout on it. It would be nice if it was silently ignored ;)

coot: Our `cancelTimeout` is implemented in terms of [`unregisterTimeout`](https://hackage.haskell.org/package/base-4.19.1.0/docs/GHC-Event.html#v:unregisterTimeout) which does a lookup in a shared map (which should be cheap).

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,202 @@ toPublicState PeerSelectionState { knownPeers
...
+                                     numberOfActiveBootstrapPeers      = numberOfHotBootstrapPeers,
+
+                                     numberOfKnownLocalRootPeers       = numberOfColdLocalRootPeers,
+                                     numberOfEstablishedLocalRootPeers = numberOfWarmLocalRootPeers,
+                                     numberOfActiveLocalRootPeers      = numberOfHotLocalRootPeers,
+                                     localRoots                        = localRootsHWC,
+
+                                     numberOfKnownSharedPeers          = numberOfColdSharedPeers,
+                                     numberOfEstablishedSharedPeers    = numberOfWarmSharedPeers,
+                                     numberOfActiveSharedPeers         = numberOfHotSharedPeers })
```
coot, line 661: @karknu if you think we should trace `known` / `established` / `active` rather than `cold` / `warm` and `hot` I'll remove this pattern synonym.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -61,180 +77,363 @@ peerChurnGovernor tracer deadlineChurnInterval bulkChurnInterval psOverallTimeou
...
+            _  -> targetNumberOfEstablishedPeers base
+      }
+
+    checkEstablishedPeersIncreased :: CheckPeerSelectionCounters
+    checkEstablishedPeersIncreased
+      PeerSelectionCounters { numberOfEstablishedPeers }
+      PeerSelectionTargets { targetNumberOfEstablishedPeers }
+      =
+      numberOfEstablishedPeers >= targetNumberOfEstablishedPeers
+
```
karknu, line 197: Similar to the problem with `checkActivePeersDecreased`. 10s timeout the governor noticing the change should be enough.
```numberOfEstablishedPeers + inProgressPromoteCold >= targetNumberOfEstablishedPeers ```

File: ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs [hunk 5]
```diff
@@ -61,180 +77,363 @@ peerChurnGovernor tracer deadlineChurnInterval bulkChurnInterval psOverallTimeou
...
+
+    increaseEstablishedBigLedgerPeers
+      :: ModifyPeerSelectionTargets
+    increaseEstablishedBigLedgerPeers
+      targets
+      =
+      targets { targetNumberOfEstablishedBigLedgerPeers = targetNumberOfEstablishedBigLedgerPeers base }
+
+    checkEstablishedBigLedgerPeersIncreased
```
karknu, line 206: see `checkEstablishedPeersIncreased` comment.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -61,180 +77,363 @@ peerChurnGovernor tracer deadlineChurnInterval bulkChurnInterval psOverallTimeou
...
+          case mode of
+            ChurnModeNormal ->
+              targetNumberOfActiveBigLedgerPeers base
+            ChurnModeBulkSync ->
+              min 1 (targetNumberOfActiveBigLedgerPeers base)
+      }
+
+    checkActiveBigLedgerPeersIncreased
+      :: CheckPeerSelectionCounters
+    checkActiveBigLedgerPeersIncreased
```
karknu, line 256: see `checkActivePeersDecreased` comment.

karknu (changes requested):
More can be done by having churn look at the governor's counters.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -61,180 +77,363 @@ peerChurnGovernor tracer deadlineChurnInterval bulkChurnInterval psOverallTimeou
...
+            ChurnModeBulkSync ->
+              min 1 (targetNumberOfActivePeers base - 1)
+      }
+
+    checkActivePeersDecreased :: CheckPeerSelectionCounters
+    checkActivePeersDecreased
+      PeerSelectionCounters { numberOfActivePeers }
+      PeerSelectionTargets { targetNumberOfActivePeers }
+      =
+      numberOfActivePeers <= targetNumberOfActivePeers
```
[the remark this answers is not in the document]
crocodile-dentist: Perhaps the next step, ie. to increase number of active peers, should take into account that hot demotions can take a long time, and increase the target not to targetNumberOfActivePeers, but to targetNumberOfActivePeers + numberOfHotDemotions. After a while, the demoted peers will fade, and we will eventually end up with targetNumberOfActivePeers. Does this make sense?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: I'll keep it so we keep logging `cold` / `warm` / `hot`.  We'll remove it in a future version.  Otherwise all SPOs would need to adapt their grafana configuration in this version.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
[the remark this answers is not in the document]
coot: Once peers are already being demoted we are guaranteed to pick the missing number of peers at some point.  The demotion action has a timeout itself, so it can't take forever.  This timeout is much shorter than the churn interval, so I think  it should be fine as is.

We are also exposing additional counters, which should give us better clue how it works in a running system.  Let's see if we can spot something strange.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
coot: Done.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
coot: Done.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
coot: Done.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,335 @@ toPublicState PeerSelectionState { knownPeers
...
+      numberOfEstablishedBootstrapPeers          = numberOfEstablishedBootstrapPeers
+                                                 - numberOfActiveBootstrapPeers,
+      numberOfWarmBootstrapPeersDemotions,
+      numberOfWarmBootstrapPeersPromotions,
+      numberOfActiveBootstrapPeers,
+      numberOfActiveBootstrapPeersDemotions
+    }
+
+
 peerStateToCounters :: Ord peeraddr => PeerSelectionState peeraddr peerconn -> PeerSelectionCounters
```
coot, line 763: The `peerStateToCounters` function is simple but rather easy to make a sill mistake which would lead to wrong churn behaviour or misleading tracing output.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,335 @@ toPublicState PeerSelectionState { knownPeers
...
+      numberOfKnownPeers                         = Set.size   knownPeersSet,
+      numberOfColdPeersPromotions                = Set.size $ knownPeersSet `Set.intersection` inProgressPromoteCold,
+      numberOfEstablishedPeers                   = Set.size   establishedPeersSet,
+      numberOfWarmPeersDemotions                 = Set.size $ establishedPeersSet `Set.intersection` inProgressDemoteWarm,
+      numberOfWarmPeersPromotions                = Set.size $ establishedPeersSet `Set.intersection` inProgressPromoteWarm,
+      numberOfActivePeers                        = Set.size   activePeersSet,
+      numberOfActivePeersDemotions               = Set.size $ activePeersSet `Set.intersection` inProgressDemoteHot,
+
+      numberOfKnownBigLedgerPeers                = Set.size   knownBigLedgerPeersSet,
+      numberOfColdBigLedgerPeersPromotions       = Set.size $ knownBigLedgerPeersSet `Set.intersection` inProgressPromoteWarm,
```
crocodile-dentist, line 786: Should it be inProgressPromotoeCold, similiar to other such lines here?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,335 @@ toPublicState PeerSelectionState { knownPeers
...
+      numberOfActiveSharedPeers                  = Set.size   activeSharedPeersSet,
+      numberOfActiveSharedPeersDemotions         = Set.size $ activeSharedPeersSet `Set.intersection` inProgressDemoteHot
     }
   where
-    knownPeersSet = KnownPeers.toSet (knownPeers st)
-    establishedPeersSet = EstablishedPeers.toSet (establishedPeers st)
-    coldPeersSet  = knownPeersSet Set.\\ establishedPeersSet
-    warmPeersSet  = establishedPeersSet Set.\\ activePeers
-    hotPeersSet   = activePeers
+    -- convention: only `{knwon,established,active}.*PeersSet` and
```
crocodile-dentist, line 819: known

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,335 @@ toPublicState PeerSelectionState { knownPeers
...
+    bigLedgerSet   = PublicRootPeers.getBigLedgerPeers publicRootPeers
+
+    -- non big ledger peers
+    knownPeersSet       = knownSet Set.\\ bigLedgerSet
+    establishedPeersSet = establishedSet Set.\\ bigLedgerSet
+    activePeersSet      = activePeers Set.\\ bigLedgerSet
+
+    -- big ledger peers
+    knownBigLedgerPeersSet       = knownSet `Set.intersection` bigLedgerSet
+    establishedBigLedgerPeersSet = establishedPeersSet `Set.intersection` bigLedgerSet
```
crocodile-dentist, line 835: establishedPeerSet -> establishedSet

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,335 @@ toPublicState PeerSelectionState { knownPeers
...
       ]
-    bigLedgerPeers = PublicRootPeers.getBigLedgerPeers publicRootPeers
+    localRootSet                  = LocalRootPeers.keysSet localRootPeers
+    knownLocalRootPeersSet        = knownPeersSet `Set.intersection` localRootSet
+    establishedLocalRootsPeersSet = establishedPeersSet `Set.intersection` localRootSet
+    activeLocalRootsPeersSet      = activePeers `Set.intersection` localRootSet
+
+    -- bootstrap peers
+    bootstrapSet                 = PublicRootPeers.getBootstrapPeers publicRootPeers
+    knownBootstrapPeersSet       = knownSet `Set.intersection` bootstrapSet
```
crocodile-dentist, line 850: knownPeersSet?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,335 @@ toPublicState PeerSelectionState { knownPeers
...
+    establishedBootstrapPeersSet = establishedPeersSet `Set.intersection` bootstrapSet
+    activeBootstrapPeersSet      = activePeers `Set.intersection` bootstrapSet
```
crocodile-dentist, line 852: activePeersSet? - or should the middle one be establishedSet instead (making the second comment above irrelevant)?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,335 @@ toPublicState PeerSelectionState { knownPeers
...
+    -- local roots
     localRoots =
       [ (hot, warm)
       | (hot, warm, _) <- LocalRootPeers.toGroupSets localRootPeers
       ]
-    bigLedgerPeers = PublicRootPeers.getBigLedgerPeers publicRootPeers
+    localRootSet                  = LocalRootPeers.keysSet localRootPeers
+    knownLocalRootPeersSet        = knownPeersSet `Set.intersection` localRootSet
+    establishedLocalRootsPeersSet = establishedPeersSet `Set.intersection` localRootSet
+    activeLocalRootsPeersSet      = activePeers `Set.intersection` localRootSet
```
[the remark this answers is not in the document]
coot: I think this doesn't matter, because a local root peers are disjoint from big ledger peers: [ref](https://github.com/intersectmbo/ouroboros-network/blob/77cb6fee1ed7788e7202df17b2273226d0e86388/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs#L1030).

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: bootstrap peers and big ledger peers are disjoint: [ref](https://github.com/intersectmbo/ouroboros-network/blob/77cb6fee1ed7788e7202df17b2273226d0e86388/ouroboros-network/src/Ouroboros/Network/PeerSelection/PublicRootPeers.hs#L67)

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: Right, otherwise `establishedBigLedgerPeersSet` will be empty.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: as above, it doesn't matter since `bootstrapSet` is disjoint from big ledge peers.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -565,49 +577,335 @@ toPublicState PeerSelectionState { knownPeers
...
+    knownBootstrapPeersSet       = knownSet `Set.intersection` bootstrapSet
+    establishedBootstrapPeersSet = establishedPeersSet `Set.intersection` bootstrapSet
+    activeBootstrapPeersSet      = activePeers `Set.intersection` bootstrapSet
+
+    -- shared peers
+    rootSet                    = PublicRootPeers.toSet publicRootPeers
+                              <> localRootSet
+    knownSharedPeersSet        = knownPeersSet Set.\\ rootSet
+    establishedSharedPeersSet  = establishedPeersSet Set.\\ rootSet
+    activeSharedPeersSet       = activePeers Set.\\ rootSet
```
crocodile-dentist, line 859: activePeersSet?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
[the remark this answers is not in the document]
coot: I'll follow your suggestion, because `activePeersSet` is smaller.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: `knownPeersSet` is smaller, so let's use it.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
karknu: Yeah, that was intended for `checkActiveBigLedgerPeersDecreased`

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
karknu: Yeah, it was meant for `checkEstablishedBigLedgerpersDecreased`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
karknu: It doesn't look like you addressed the comment.

coot: Really done :wink: :

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -555,11 +559,22 @@ peerSelectionGovernorLoop tracer
...
+            if activePeers st
+               `Set.isSubsetOf`
+               LocalRootPeers.keysSet (localRootPeers st)
+            then ConnectedToOnlyLocalOutboundPeers
+            else ConnectedToExternalOutboundPeers
+
       atomically $ do
+        -- Update outbound connections state
+        updateOutboundConnectionsState actions outboundConnectionsState
+
```
karknu: I think this logic may cause trouble.
We don't check for trusted peers here, is is possible that we could respond with `ConnectedToExternalOutboundPeers` before we even managed to disconnect from all untrusted peers when entering the TooOld state?

karknu (changes requested):
Some feedback on `OutboundConnectionsState`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -549,16 +556,36 @@ peerSelectionGovernorLoop tracer
...
       now <- getMonotonicTime
+
       let Decision { decisionTrace, decisionJobs, decisionState } =
             timedDecision now
-          !newCounters = peerStateToCounters decisionState
+
+          !outboundConnectionsState =
+            if activePeers st
+               `Set.isSubsetOf`
+               LocalRootPeers.keysSet (localRootPeers st)
```
coot, line 566: We should use `decisionState` not `st`.

* `st` is the initial state of this loop state
* `decisionState` is the state after executing a peer selection action

coot: Fixed.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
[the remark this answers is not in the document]
coot: > I think that the check should be changed to:
> numberOfActivePeers - numberOfDemoteHot <= targetNumberOfActivePeers
> where numberOfDemoteHot = Set.size $ inProgressDemoteHot and should be added to PeerSelectionCounters.
> This means that we can move on as soon as the governor has picked some peers to demote.

I oppose this change.  It's hard to make a churn counter which can make sense: if we optimistically trace that we decreased the peers in progress, then the counter which logs how many peers we promoted will be smaller by that number, which will be confusing.

Churn can take some time, so I don't mind to have a 300s timeout on decreasing the number of active peers.  If churn takes 10 minutes every hour that should be fine.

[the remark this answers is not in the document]
coot: This is how churn graphs looks without this change, like it was done initially:
image

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
```diff
@@ -0,0 +1,5 @@
+module Ouroboros.Network.PeerSelection.LocalRootPeers (OutboundConnectionsState (..)) where
+
+data OutboundConnectionsState = ConnectedToExternalOutboundPeers
+                              | ConnectedToOnlyLocalOutboundPeers
```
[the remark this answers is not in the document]
coot: Good point, yes we need the third state.  However, we don't know if the node is a BP or not, that's something that `consensus` is aware of.

coot:
```
cabal run ouroboros-network:sim-tests -- -p '/PeerSelection/' --quickcheck-tests=10000
All 80 tests passed (20871.28s)
```

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
```diff
@@ -0,0 +1,35 @@
...
+data OutboundConnectionsState =
+    TrustedStateWithExternalPeers
+    -- ^
+    -- * /in bootstrap mode/: connected only to trusted local peers and at least
+    --   one bootstrap peer.
+    -- * /in genesis mode/: meeting target of active big ledger peers, ledger
+    --   state judgement must be `TooOld`.
+
+  | EnteringTrustedStateWithExternalPeers
+    -- ^ entering trusted state with eternal peers.
```
karknu, line 17: s/eternal/external

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
```diff
@@ -0,0 +1,35 @@
...
+    --   at least on bootstrap peer.
+    -- * /in genesis mode/: ledger state judgement returns `TooOld`
+
+  | HiddenRelayOrBP
+    -- ^ A node is classified as a hidden relay or a BP if it is configured so
+    -- that it can only have a chance to be connected to local roots. This is
+    -- true in one of the two ways:
+    --
+    -- * it's configured with `DontUseBootstrapPeers` and `DontUseLedgerPeers`; or
+    -- * the targets are set so the node can only be connected to local roots.
```
coot, line 29: We can collapse these three into one `TrustedStateWithExternalPeers`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +751,53 @@ wakeupDecision st _now =
...
+         ledgerPeers <- getLedgerPeers ledgerPeersConsensusInterface afterSlot
+         return $ case ledgerPeers of
+           BeforeSlot -> True
+           LedgerPeers {} ->
+                -- there are no big ledger peers
+                targetNumberOfKnownBigLedgerPeers == 0
+             && (Set.null knownBigLedgerPeers)
+                -- all known peers are local roots; this implies that there no
+                -- bootstrap peers, nor ledger peers or shared peers
+             && knownPeers `Set.isSubsetOf` knownLocalRootPeers
```
coot:
```suggestion
             return False
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +751,53 @@ wakeupDecision st _now =
...
   }
+
+
+-- | A node is classified as a hidden relay or a BP if it is configured so that
+-- it can only have a chance to be connected to local roots. See `HiddenRelayOrBP`.
+--
+-- In the second case mentioned in `HiddenRelayOrBP` we are slightly more strict.
+-- We not only check the targets but also current `PeerSelectionView`.
+--
+readIsHiddenRelayOrBP
```
coot, line 755: `promiscuousRelay`

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -558,25 +581,86 @@ peerSelectionGovernorLoop tracer
...
+               (False, UseBootstrapPeers bootstrapPeers)
+                         |  -- we are only connected to trusted local root peers or bootstrap peers
+                            viewEstablishedPeers `Set.isSubsetOf` (viewEstablishedBootstrapPeers <> trustableLocalRootSet)
+                            -- there's at least one active bootstrap peer
+                         ,  not (Set.null viewActiveBootstrapPeers)
+                         -> TrustedStateWithExternalPeers
+
+                         |  requiresBootstrapPeers (bootstrapPeersFlag st'') (ledgerStateJudgement st'')
+                         ,  not (null bootstrapPeers)
+                         -> EnteringTrustedStateWithExternalPeers
```
coot, line 621: `UntrustedState`

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -558,25 +581,86 @@ peerSelectionGovernorLoop tracer
...
+
+        isHiddenRealyOrBP <- readIsHiddenRelayOrBP interfaces
+                                                   (bootstrapPeersFlag st'')
+                                                   (targets st'')
+                                                   (peerSelectionStateToView st'')
+
+        let !outboundConnectionsState =
+              case (isHiddenRealyOrBP, bootstrapPeersFlag st'') of
+
+               (True, _) -> HiddenRelayOrBP
```
coot, line 609:
```suggestion
                         |  -- we are only connected to trusted local root peers or bootstrap peers
                            viewEstablishedPeers `Set.isSubsetOf` (trustableLocalRootSet)
                            -- there's at least one active bootstrap peer
                         ,  not (Set.null viewActiveBootstrapPeers)
                         -> TrustedStateWithExternalPeers
                         
                         | otherwise
                         -> UntrustedState
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -558,25 +581,86 @@ peerSelectionGovernorLoop tracer
...
+                         -> TrustedStateWithExternalPeers
+
+                         |  requiresBootstrapPeers (bootstrapPeersFlag st'') (ledgerStateJudgement st'')
+                         ,  not (null bootstrapPeers)
+                         -> EnteringTrustedStateWithExternalPeers
+
+                         |  otherwise
+                         -> UntrustedState
+
+                -- genesis mode
```
coot, line 626: TODO: do it for praos without bootstrap peers, leave a todo for genesis.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
```diff
@@ -0,0 +1,35 @@
...
+
+  | UntrustedState
+    -- ^ catch all other cases
+  deriving (Eq, Show, Generic)
```
coot, line 33: We also need `TrustedUnreachableState` for peers not configured with any trusted peers (bootstrap or local roots).

coot: We decided to map the promiscuous mode to `TrustedStateWithExternalPeers` to simplify things for `consensus`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +766,31 @@ wakeupDecision st _now =
...
     decisionJobs  = []
   }
+
+
+-- | A node is classified as a hidden relay or a BP if it is configured so that
+-- it can only have a chance to be connected to local roots. See `HiddenRelayOrBP`.
+--
+-- In the second case mentioned in `HiddenRelayOrBP` we are slightly more strict.
+-- We not only check the targets but also current `PeerSelectionView`.
+--
```
karknu, line 767: Comment should talk about `PromiscuousMode` not `HiddenRelayOrBP`

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +766,31 @@ wakeupDecision st _now =
...
+  -> UseBootstrapPeers
+  -> PeerSelectionSetsWithSizes peeraddr
+  -> STM m Bool
+readPromiscuousMode
+  PeerSelectionInterfaces { readUseLedgerPeers }
+  useBootstrapPeers
+  PeerSelectionView {
+    viewKnownBootstrapPeers = (knownBootstrapPeers, _)
+  }
+  =
```
karknu, line 780: `readPromiscuousMode` also must take peersharing into account.
One can run a node without any localroots, bootstrap peers, and ledger peers . And then only depend on peersharing.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -558,25 +579,103 @@ peerSelectionGovernorLoop tracer
...
+
+            trustableLocalRootSet = LocalRootPeers.trustableKeysSet (localRootPeers st'')
+
+        isPromiscuous <- readPromiscuousMode interfaces
+                                             (bootstrapPeersFlag st'')
+                                             (peerSelectionStateToView st'')
+
+        let !outboundConnectionsState =
+              case (isPromiscuous, bootstrapPeersFlag st'') of
+                (False, _) -> TrustedStateWithExternalPeers
```
karknu, line 602: I think this case must consider `trustableLocalRootSet` . The node may not be promiscuous but it may still be talking to untrusted localroots.

karknu (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
coot: Done

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
coot: Done

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
coot: Done

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
```diff
@@ -0,0 +1,28 @@
...
+    -- * /in praos mode/: connected only to trusted local
+    --   peers and at least one bootstrap peer or public root;
+    -- * /in genesis mode/: meeting target of active big ledger peers;
+    -- * or it is in promiscuous mode (see below).
+    --
+    -- A node is classified as a non promiscuous node if it is a hidden relay or
+    -- a BPa, e.g. if it is configured such that it can only have a chance to be
+    -- connected to local roots. This is true in one of the two ways:
+    --
+    -- * it's configured with `DontUseBootstrapPeers` and `DontUseLedgerPeers`;
```
karknu, line 20: and with peersharing disabled.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
bolt12: FWIW if one has a weird configuration like specifying node A as a local root and also adding node A to `UseBootstrapPeers` won't be a problem since we prioritize local roots so the node will respect the invariant that the bootstrap peers and local root peers should be disjoint sets

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +766,34 @@ wakeupDecision st _now =
...
+  =
+  do useLedgerPeers <- readUseLedgerPeers
+     pure $
+       case (useLedgerPeers, peerSharing, useBootstrapPeers) of
+         (DontUseLedgerPeers, PeerSharingDisabled, DontUseBootstrapPeers)
+           -> NonPromiscuousMode
+         (DontUseLedgerPeers, PeerSharingDisabled, UseBootstrapPeers {})
+           |  Set.null knownBootstrapPeers
+           -> NonPromiscuousMode 
+         _ -> PromiscuousMode
```
bolt12, line 799: Isn't knowing about this mode only useful when in `TooOld` state? If so I don't think checking for `PeerSharingDisabled` or `DontUseLedgerPeers` is right. When in `TooOld` state we are only going to be connected to trusted peers, so the node can effectively have those flags enabled.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LocalRootPeers.hs
coot: I forgot to update the comment.  I added `PromiscuousMode` type in `Governor.Types` where this comment lives now.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
bolt12: But the way I understand it, this won't work in bootstrap mode, am I wrong?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -558,25 +582,108 @@ peerSelectionGovernorLoop tracer
...
+            PeerSelectionTargets {
+                targetNumberOfActiveBigLedgerPeers = _b
+              } = targets st''
+
+            trustableLocalRootSet = LocalRootPeers.trustableKeysSet (localRootPeers st'')
+
+        promiscuousMode <- readPromiscuousMode (readUseLedgerPeers interfaces)
+                                               (peerSharing actions)
+                                               (bootstrapPeersFlag st'')
+                                               (peerSelectionStateToView st'')
```
bolt12, line 603:
```suggestion
                                               peerSelectionView
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
bolt12: I just saw the logic above and I agree with it. However, I think that logic should be done in this function and not pollute the `peerSelectionGovernorLoop` It's still quite a large chunk of code.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -558,25 +582,108 @@ peerSelectionGovernorLoop tracer
...
+                            let establishedPublicRootPeers =
+                                  viewEstablishedPeers
+                                  `Set.intersection`
+                                  Map.keysSet (getPublicConfigPeers (publicRootPeers st''))
+                          , viewEstablishedPeers `Set.isSubsetOf` (establishedPublicRootPeers <> trustableLocalRootSet)
+                          , not (null establishedPublicRootPeers)
+                          -> TrustedStateWithExternalPeers
+
+                          |  otherwise
+                          -> UntrustedState
```
bolt12, line 660: Can this logic be moved to `readPromiscuousMode` ?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -386,9 +389,29 @@ data PeerSelectionActions peeraddr peerconn m = PeerSelectionActions {
...
      }
 
+-- | Interfaces required by the peer selection governor, which do not need to
+-- be shared with actions and thus are not part of `PeerSelectionActions`.
+--
+-- TODO: add other `TVar`s which outbound governor is using.
+--
+data PeerSelectionInterfaces m = PeerSelectionInterfaces {
+      readUseLedgerPeers :: STM m UseLedgerPeers
+    }
```
bolt12: Nice! Perhaps move `peerSharing`, `readLocalRootPeers`, `readPeerSelectionTargets`, `readUseBootstrapPeers` and `readLedgerStateJudgement` from `PeerSelectionActions` to here since they are not actions as well?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -536,6 +559,24 @@ data PeerSelectionState peeraddr peerconn = PeerSelectionState {
      }
   deriving Show
 
+-- | A node is classified as a non-promiscuous node if it is a hidden relay or
+-- a BPa, e.g. if it is configured such that it can only have a chance to be
```
bolt12, line 570:
```suggestion
-- a BP, e.g. if it is configured such that it can only have a chance to be
```

is this a typo?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -536,6 +559,24 @@ data PeerSelectionState peeraddr peerconn = PeerSelectionState {
...
+-- * `UseBootstrapPeers`, `DontUseLedgerPeers` and
+--   `PeerSharingDisabled`, but it's not using any bootstrap peers (i.e. it is
+--   synced).
+--
+-- See `Ouroboros.Network.PeerSelection.Governor.readPromiscuousMode`.
+--
+data PromiscuousMode =
+     NonPromiscuousMode
+   | PromiscuousMode
+  deriving Show
```
bolt12, line 585: I think it would be worth to make even further distinctions here and explain what's classified as promiscuous or not depending on bootstrap/praos/genesis mode.

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/LocalRootPeers.hs
```diff
@@ -72,6 +73,12 @@ newtype WarmValency = WarmValency { getWarmValency :: Int }
   deriving (Show, Eq, Ord)
   deriving Num via Int
 
+-- | Data available from topology file.
+--
+type Config peeraddr =
+     [(HotValency, WarmValency, Map peeraddr ( PeerAdvertise, PeerTrustable))]
```
bolt12, line 79: I am not a fan of type alias in general, I like this idea though so maybe having a newtype is better so we make sure we always use it and not make the code be more confusing by using `Config` in one place and not others.

Perhaps other idea is to have :

```haskell
data Config = Config { hotValency  :: HotValency
                     , warmValency :: WarmValency
                     , peers       :: Map peeraddr ( PeerAdvertise, PeerTrustable)) 
                     }
```

and then use `[LocalRootPeers.Config]` instead.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
```diff
@@ -3426,6 +3428,77 @@ prop_governor_uses_ledger_peers env =
...
+               Set.null (fst (viewKnownBootstrapPeers cs)
+                          Set.\\ localRootSet
+                          Set.\\ publicRootSet)
+            && Set.null (fst (viewKnownBigLedgerPeers cs)
+                          Set.\\ localRootSet
+                          Set.\\ publicRootSet)
+            && Set.null (fst (viewKnownSharedPeers cs)
+                          Set.\\ localRootSet
+                          Set.\\ publicRootSet)
+          else True)
```
bolt12, line 3496:
```suggestion
          pm 
              -- we need to remove local and public roots.  They are changing
              -- over time, and a node might keep using them, event though the
              -- node is configured as a non-promiscuous node.
              --
              -- This makes this test only effective if a node starts in
              -- non-promiscuous mode, until it is reconfigured.  This can
              -- discover some bugs in `readPromiscuousMode` but certainly not
              -- all.
              --
              -- TODO: write a more effective test.
          ||   (Set.null (fst (viewKnownBootstrapPeers cs)
                          Set.\\ localRootSet
                          Set.\\ publicRootSet)
            && Set.null (fst (viewKnownBigLedgerPeers cs)
                          Set.\\ localRootSet
                          Set.\\ publicRootSet)
            && Set.null (fst (viewKnownSharedPeers cs)
                          Set.\\ localRootSet
                          Set.\\ publicRootSet))
```

bolt12 (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
coot: I'll move it to it's own function, `readAssociationMode` (how `readPromiscuousMode` was renamed) it's good as it stands (small, easy to understand and useful for testing).

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: but it doesn't depend on `praos / genesis` mode, or maybe I just don't understand your comment.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
coot: I added `outboundConnectionsState` function which compute the `OutboundConnectionsState`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
bolt12: I noticed the following commit added more things but not the ones I mentioned here.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
bolt12: Yes, I was thinking about `UntrustedState` and `TrustedStateWithExternalPeers`

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/LocalRootPeers.hs
bolt12: I know, but I thought the idea was to type less 😛

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: I am confused. It's the other way around. `OutboundConnectionsState` depends on `AssociationMode`, so we should rather expand the docs of `OutboundConenctionsState` if needed. Do you have any suggestions?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: They are useful in guarded decisions, so I don't think this is the right move.  Also, I didn't add `readUseLedgerPeers` and the other `StrictTVar`s to `PeerSelectionActions` because they are unnecessary in guarded decisions.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +702,110 @@ wakeupDecision st _now =
...
+    viewKnownBootstrapPeers = (knownBootstrapPeers, _)
+  }
+  =
+  do useLedgerPeers <- readUseLedgerPeers
+     pure $
+       case (useLedgerPeers, peerSharing, useBootstrapPeers) of
+         (DontUseLedgerPeers, PeerSharingDisabled, DontUseBootstrapPeers)
+           -> LocalRootsOnly
+         (DontUseLedgerPeers, PeerSharingDisabled, UseBootstrapPeers {})
+           |  Set.null knownBootstrapPeers
```
karknu, line 724: This looks racy to me.
Isn't it possible for us to have no bootstrap peers while we're in the process of doing a DNS lookup on the bootstrap peers?

I think that this check should just be removed.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +702,110 @@ wakeupDecision st _now =
...
+                -> TrustedStateWithExternalPeers
+
+                |  otherwisej
+                -> UntrustedState
+      -}
+
+      (LocalRootsOnly, _)
+        |  -- we are only connected to trusted local root
+           -- peers or bootstrap peers
+           viewEstablishedPeers `Set.isSubsetOf` (viewEstablishedBootstrapPeers <> trustableLocalRootSet)
```
karknu, line 762: If we are configured to use `LocalRootsOnly` we shouldn't have any established bootstrap peers so they shouldn't be included in the check.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +702,110 @@ wakeupDecision st _now =
...
+                |  numberOfActiveBigLedgerPeers >= targetNumberOfActiveBigLedgerPeers
+                -> TrustedStateWithExternalPeers
+
+                |  otherwisej
+                -> UntrustedState
+      -}
+
+      (LocalRootsOnly, _)
+        |  -- we are only connected to trusted local root
+           -- peers or bootstrap peers
```
karknu, line 761: Remove `bootstrap peers`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +702,110 @@ wakeupDecision st _now =
...
+    =
+    case (associationMode, bootstrapPeersFlag) of
+      {-
+       -- genesis mode
+       -- TODO: issue #4846
+      (LocalRootsOnly, _)
+                |  numberOfActiveBigLedgerPeers >= targetNumberOfActiveBigLedgerPeers
+                -> TrustedStateWithExternalPeers
+
+                |  otherwisej
```
karknu, line 755: `otherwisej` should this be `otherwise`?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -667,3 +702,110 @@ wakeupDecision st _now =
...
+                  -- condition below likely will not be satisfied,
+                  -- and thus `UntrustedState` will be returned.
+                  let establishedPublicRootPeers =
+                        viewEstablishedPeers
+                        `Set.intersection`
+                        Map.keysSet (getPublicConfigPeers publicRootPeers)
+                , viewEstablishedPeers `Set.isSubsetOf` (establishedPublicRootPeers <> trustableLocalRootSet)
+                , not (null establishedPublicRootPeers)
+                -> TrustedStateWithExternalPeers
+
```
karknu, line 798: Perhaps we should just simplify this and always return `Untrusted` state.
The node is not configured to use bootstrap peers and while it is possible that it may be in a trusted state out of pure luck I don't think it is worth the effort to check.

This could also confuse operators when we have exposed the counters for this. The operator thinks he has disabled bootstrap peers but once in a blue moon it is active anyway.

karknu (changes requested):
more feedback.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
coot: And as soon as we will resolve it, the node will switch to `Unrestricted` mode. Is this a problem?
On the other hand, if we remove the whole case, the property will become static unless one changes the topology file.

coot: This covers the case of empty `bootstrap` peers in the topology file with at least one trustable local root peer.  Then `bootstrap` mode will be used ([ref](https://github.com/intersectmbo/cardano-node/blob/388038b8a356aa08e7ff137d255708175036137e/cardano-node/src/Cardano/Node/Configuration/TopologyP2P.hs#L306-L320)), and the node will switch to only trustable local roots when it's far from the tip.

We could match on `UseBootstrapPeers [RelayAccessPoint]`, since the list contains unresolved addresses - directly what's written in the topology file.
I think it's a mistake to include `[RelayAccessPoint]` there, but we could have a `UseBootstrapPeers Bool` inside the outbound governor instead.

karknu: > And as soon as we will resolve it, the node will switch to Unrestricted mode. Is this a problem?

Yes, that is the original case for this hole modification. We where worried that we would signal to consensus that we're in a safe state so fast that we hadn't even had time to connect to an external bootstrap peer.

Changing this be a static config check using `[RelayAccessPoint]` seems fine to me.

coot: I am fine with making the property static.  But maybe I misunderstand something, so just for the sake of the argument... If we resolve bootstrap peer and now `readAssociationMode` returns `UnrestrictedMode` instead of `LocalRootsOnly`, we will land in `(Unrestricted, UseBoostrapPeers {})` in `outboundConnectionState`; we will return `TrustedStateWithExternalPeers` only if there's an active bootstrap peer - so how's that too early?

coot: oh, I think I got it.  You're worried that we return `LocalRootsOnly` early and then return `TrustedStateWithExternalPeers` if we are connected to a trusted local root, while we should instead wait for an active bootstrap peer.  You're right we actually have to use the configuration value.

coot: Fixed.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
coot: Yes, that's fine.  We really need to handle just the genesis and bootstrap modes, the praos mode can stay unchanged.

bolt12 (approved):

karknu (approved):
LGTM

PR #4850, opened 2024-04-11
Feed peer selection governor with big ledger peers obtained from a snapshot

crocodile-dentist:
# Description

Consensus Genesis syncing mode relies on having at least one honest peer to ensure a syncing node will finish on a good chain.  Largest stake pools are a good proxy for honesty, and therefore these peers are the big ledger peers, which a syncing node may not be aware of when its ledger state may still be too stale. A snapshot file of these peers can be provided in a node's topology configuration, and this set will be provided to peer selection governor so it has an opportunity to make these connections.

This change modifies ledgerPeersThread to leverage this data, and appropriately randomly pick from this set and respond to peer selection governor when it makes a request for big ledger peers.

# Checklist

- Branch
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Utils.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Utils.hs
@@ -3,8 +3,8 @@
 
 module Ouroboros.Network.PeerSelection.LedgerPeers.Utils
   ( bigLedgerPeerQuota
-  , accBigPoolStake
-  , reRelativeStake
+  , accumulateBigLedgerStake
+  , recomputeRelativeStake
   , AccPoolStake (..)
   , PoolStake (..)
   , RelayAccessPoint (..)
@@ -29,13 +29,13 @@ bigLedgerPeerQuota = 0.9
 -- and tag each one with cumulative stake, with a cutoff
 -- at 'bigLedgerPeerQuota'
 --
-accBigPoolStake :: [(PoolStake, NonEmpty RelayAccessPoint)]
-                -> [(AccPoolStake, (PoolStake, NonEmpty RelayAccessPoint))]
-accBigPoolStake =
+accumulateBigLedgerStake :: [(PoolStake, NonEmpty RelayAccessPoint)]
+                         -> [(AccPoolStake, (PoolStake, NonEmpty RelayAccessPoint))]
+accumulateBigLedgerStake =
     takeWhilePrev (\(acc, _) -> acc <= bigLedgerPeerQuota)
     . go 0
     . sortOn (Down . fst)
-    . reRelativeStake BigLedgerPeers
+    . recomputeRelativeStake BigLedgerPeers
   where
     takeWhilePrev :: (a -> Bool) -> [a] -> [a]
     takeWhilePrev f as =
@@ -55,10 +55,10 @@ accBigPoolStake =
 -- | Not all stake pools have valid \/ usable relay information. This means that
 -- we need to recalculate the relative stake for each pool.
 --
-reRelativeStake :: LedgerPeersKind
+recomputeRelativeStake :: LedgerPeersKind
                 -> [(PoolStake, NonEmpty RelayAccessPoint)]
                 -> [(PoolStake, NonEmpty RelayAccessPoint)]
-reRelativeStake ledgerPeersKind pl =
+recomputeRelativeStake ledgerPeersKind pl =
     let pl'   = first adjustment <$> pl
         total = List.foldl' (+) 0 (fst <$> pl')
         pl''  = first (/ total) <$> pl'
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
@@ -22,6 +22,7 @@ import Control.Monad.IOSim hiding (SimResult)
 import Control.Tracer (Tracer (..), nullTracer, traceWith)
 import Data.Aeson
 import Data.Aeson.Types as Aeson
+import Data.ByteString.Char8 qualified as BS
 import Data.IP qualified as IP
 import Data.List as List (foldl', intercalate, isPrefixOf, nub, sortOn)
 import Data.List.NonEmpty qualified as NonEmpty
@@ -55,7 +56,8 @@ tests = testGroup "Ouroboros.Network.LedgerPeers"
   , testProperty "Pick" prop_pick
   , testProperty "accBigPoolStake" prop_accBigPoolStake
   , testProperty "getLedgerPeers invariants" prop_getLedgerPeers
-  , testProperty "LedgerPeerSnapshot encode/decode version 1" prop_ledgerPeerSnapshotV1
+  , testProperty "LedgerPeerSnapshot CBOR version 1" prop_ledgerPeerSnapshotCBORV1
+  , testProperty "LedgerPeerSnapshot JSON version 1" prop_ledgerPeerSnapshotJSONV1
   ]
 
 newtype ArbitraryPortNumber = ArbitraryPortNumber { getArbitraryPortNumber :: PortNumber }
@@ -377,34 +379,87 @@ prop_getLedgerPeers (ArbitrarySlotNo curSlot)
                   (pure lsj)
                   (pure (Map.elems (accPoolStake lps)))
 
--- | Tests if the CBOR encoding is valid, and whether a round
--- trip results in the original peer snapshot value.
+-- | Checks validity of LedgerPeerSnapshot CBOR encoding, and whether
+--   round trip cycle is the identity function
 --
-prop_ledgerPeerSnapshotV1 :: ArbitrarySlotNo
-                          -> LedgerPools
-                          -> Property
-prop_ledgerPeerSnapshotV1 (ArbitrarySlotNo slot)
-                          (LedgerPools pools) =
+prop_ledgerPeerSnapshotCBORV1 :: ArbitrarySlotNo
+                              -> LedgerPools
+                              -> Property
+prop_ledgerPeerSnapshotCBORV1 slotNo
+                              ledgerPools =
   counterexample (show snapshot) $
-    conjoin [counterexample "Invalid CBOR encoding" $ validFlatTerm encoded,
-             either ((`counterexample` False) . ("JSON decode failed: " <>))
-                    (("CBOR round trip failed" `counterexample`) . (snapshot ==))
-                    decoded,
-             either ((`counterexample` False) . ("JSON decode failed: " <>))
-                    (("JSON round trip failed" `counterexample`) . (snapshot ==))
-                    decodedJSON]
+         counterexample ("Invalid CBOR encoding" <> show encoded)
+                        (validFlatTerm encoded)
+    .&&. either ((`counterexample` False) . ("CBOR decode failed: " <>))
+                (counterexample . ("CBOR round trip failed: " <>) . show <*> (snapshot ==))
+                decoded
+  where
+    snapshot = snapshotV1 slotNo ledgerPools
+    encoded = toFlatTerm . toCBOR $ snapshot
+    decoded = fromFlatTerm fromCBOR encoded
+
+-- | Tests if LedgerPeerSnapshot JSON round trip is the identity function
+--
+prop_ledgerPeerSnapshotJSONV1 :: ArbitrarySlotNo
+                              -> LedgerPools
+                              -> Property
+prop_ledgerPeerSnapshotJSONV1 slotNo
+                              ledgerPools =
+  counterexample (show snapshot) $
+     either ((`counterexample` False) . ("JSON decode failed: " <>))
+            (counterexample . ("JSON round trip failed: " <>) . show <*> nearlyEqualModuloFullyQualified snapshot)
+            roundTrip
+  where
+    snapshot = snapshotV1 slotNo ledgerPools
+    roundTrip = case fromJSON . toJSON $ snapshot of
+                  Aeson.Success s -> Right s
+                  Error str       -> Left str
+
+    nearlyEqualModuloFullyQualified snapshotOriginal snapshotRoundTripped =
+      let LedgerPeerSnapshotV1 (wOrigin, relaysWithAccStake) = snapshotOriginal
+          strippedRelaysWithAccStake = stripFQN <$> relaysWithAccStake
+          LedgerPeerSnapshotV1 (wOrigin', relaysWithAccStake') = snapshotRoundTripped
+          strippedRelaysWithAccStake' = stripFQN <$> relaysWithAccStake'
+      in
+             wOrigin === wOrigin'
+        .&&. counterexample "fully qualified name"
+                            (strippedRelaysWithAccStake === strippedRelaysWithAccStake')
+        .&&. counterexample "approximation error"
+                            (compareApprox relaysWithAccStake relaysWithAccStake')
+
+    stripFQN (_, (_, relays)) = step <$> relays
+    step it@(RelayAccessDomain domain port) =
+      case BS.unsnoc domain of
+        Just (prefix, '.') -> RelayAccessDomain prefix port
+        _otherwise         -> it
+    step it = it
+
+    compareApprox left right =
+      let left' = [(accStake, relativeStake)
+                  | (AccPoolStake accStake, (PoolStake relativeStake, _)) <- left]
+          right' = [(accStake, relativeStake)
+                   | (AccPoolStake accStake, (PoolStake relativeStake, _)) <- right]
+          go (accStake, relativeStake)
+             (accStake', relativeStake') =
+               accStake' / accStake > 999999 % 1000000
+            && accStake' / accStake < 1000001 % 1000000
+            && relativeStake' / relativeStake > 999999 % 1000000
+            && relativeStake' / relativeStake < 1000001 % 1000000
+
+      in all (uncurry go) (zip left' right')
+
+-- | helper functions for ledgerpeersnapshot encoding tests
+--
+snapshotV1 :: ArbitrarySlotNo
+           -> LedgerPools
+           -> LedgerPeerSnapshot
+snapshotV1 (ArbitrarySlotNo slot)
+           (LedgerPools pools) = LedgerPeerSnapshotV1 (originOrSlot, poolStakeWithAccumulation)
   where
     poolStakeWithAccumulation = Map.assocs . accPoolStake $ pools
     originOrSlot = if slot == 0
                    then Origin
                    else At slot
-    snapshot = LedgerPeerSnapshotV1 (originOrSlot, poolStakeWithAccumulation)
-    encoded = toFlatTerm . toCBOR $ snapshot
-    decoded = fromFlatTerm fromCBOR encoded
-    encodedJSON = toJSON snapshot
-    decodedJSON = case fromJSON encodedJSON of
-                    Aeson.Success s -> Right s
-                    Error str       -> Left str
 
 -- TODO: Belongs in iosim.
 data SimResult a = SimReturn a [String]
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
@@ -194,7 +194,8 @@ prop_pick100 seed (NonNegative n) (ArbLedgerPeersKind ledgerPeersKind) (MockRoot
                 WithLedgerPeersArgs { wlpRng = rng,
                                       wlpConsensusInterface = interface,
                                       wlpTracer = verboseTracer,
-                                      wlpGetUseLedgerPeers = pure $ UseLedgerPeers Always }
+                                      wlpGetUseLedgerPeers = pure $ UseLedgerPeers Always,
+                                      wlpGetLedgerPeerSnapshot = pure Nothing }
                 (\request _ -> do
                   threadDelay 1900 -- we need to invalidate ledger peer's cache
                   resp <- request (NumberOfPeers 1) ledgerPeersKind
@@ -254,7 +255,8 @@ prop_pick (LedgerPools lps) (ArbLedgerPeersKind ledgerPeersKind) count seed (Moc
                 WithLedgerPeersArgs { wlpRng = rng,
                                       wlpConsensusInterface = interface,
                                       wlpTracer = verboseTracer,
-                                      wlpGetUseLedgerPeers = pure $ UseLedgerPeers (After 0) }
+                                      wlpGetUseLedgerPeers = pure $ UseLedgerPeers (After 0),
+                                      wlpGetLedgerPeerSnapshot = pure Nothing }
                 (\request _ -> do
                   threadDelay 1900 -- we need to invalidate ledger peer's cache
                   resp <- request (NumberOfPeers count) ledgerPeersKind
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE DerivingVia         #-}
-{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RankNTypes          #-}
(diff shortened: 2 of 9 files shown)


ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Utils.hs
```diff
@@ -0,0 +1,81 @@
...
+bigLedgerPeerQuota :: AccPoolStake
+bigLedgerPeerQuota = 0.9
+
+-- | Sort ascendingly a given list of pools with stake,
+-- and tag each one with cumulative stake, with a cutoff
+-- at 'bigLedgerPeerQuota'
+--
+accBigPoolStake :: [(PoolStake, NonEmpty RelayAccessPoint)]
+                -> [(AccPoolStake, (PoolStake, NonEmpty RelayAccessPoint))]
+accBigPoolStake =
```
coot, line 34: Sine `accBigPoolStake` nand `reRelativeState` are now public API, it's better to give them better name, maybe:
* `accumulateBigLedgerState`
* `recomputeRelativeStake`

coot, line 34: We have some code duplication between `accPoolStake` and `accBigPoolStake`; could you create an issue to address this in the future?

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +57,40 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
...
+instance ToCBOR RelayAccessPoint where
+  toCBOR = mconcat . (encodeListLen 3 :) . \case
+    RelayAccessDomain domain port ->
+      [encodeWord8 0, serialize' port, toCBOR domain]
+    RelayAccessAddress (IP.IPv4 ip4) port ->
+      [encodeWord8 1, serialize' port, toCBOR (IP.fromIPv4 ip4)]
+    RelayAccessAddress (IP.IPv6 ip6) port ->
+      [encodeWord8 2, serialize' port, toCBOR (IP.fromIPv6 ip6)]
+    where
+      serialize' = toCBOR . toInteger
```
coot, line 73: It will be cleaner to avoid `mconcat`;
```hs
toCBOR (RelayAccessDomain domain port) =
     encodeListLen 3
  <> encodeWord8 0
  <> serialize' port
  <> toCBOR domain
toCBOR (RelayAccessAddress (IP.IPv4 ipv4) port) =
     encodeListLen 3
  <> encodeWord8 1
  <> serialize' port
  <> toCBOR (IP.fromIPv4 ipv4)
toCBOR (RelayAccessAddress (IP.IPv6 ip6) port) =
     encodeListLen 3
  <> encodeWord8 2
  <> serialize' port
  <> toCBOR (IP.fromIPv4 ip6)
```

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +57,40 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
...
+
+instance FromCBOR RelayAccessPoint where
+  fromCBOR = do
+    codeTag <- decodeListLen
+    when (codeTag /= 3) . fail $    "Unrecognized RelayAccessPoint encoding tag "
+                                 <> show codeTag
```
coot, line 79: it's an unrecognised list length rather than a tag.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
```diff
@@ -95,7 +98,7 @@ newtype ArbitrarySlotNo =
 -- of the tests we run.
 instance Arbitrary ArbitrarySlotNo where
     arbitrary =
-      ArbitrarySlotNo . fromInteger <$> arbitrary
+      ArbitrarySlotNo . SlotNo <$> arbitrarySizedBoundedIntegral
```
coot, line 101: :+1: I would just use `arbitrary`, since `Wor64` is using `arbitrarySizedBoundedIntegral` underneath.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Utils.hs
coot, line 34: Do we have any properties for `accBigPoolStake` and `reRelativeState`; it's a good opportunity to add some if we don't.

coot (commented):
We also need to add some tests where the simulation nodes are using ledger snapshot rather as a source of big ledger peers.  Network simulation tests are located [here](https://github.com/IntersectMBO/ouroboros-network/blob/master/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs).

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Utils.hs
crocodile-dentist: We had a test for the first one, and I've added a test for recomputeRelativeStake - the properties I've considered are that
1. It should not add/remove/modify any relay access points
2. Each relative stake should be non-negative
3. Total normalized stake adds up to unity

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -255,20 +259,33 @@ ledgerPeersThread PeerActionsDNS {
...
+                                     <*> getLedgerPeers wlpConsensusInterface ula
+                                     <*> wlpGetLedgerPeerSnapshot)
+
+                 -- we have to assess which of, if any, peers we get from consensus vs.
+                 -- peers we may have from the snapshot file is more recent, and use that
+                 let (accPoolStake -> peersStake, bigPeersStakeMap) =
+                       case (consensusSlotNo, consensusPeers, peerSnapshot) of
+                         (At t, LedgerPeers _ lp, Just (LedgerPeerSnapshot (At t', sp {- snapshot peer-})))
+                           | t' > t -> (lp, Map.fromAscList sp)
+                           | otherwise -> (lp, accBigPoolStakeMap lp)
```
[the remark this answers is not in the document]
bolt12: I wish there would be some symmetry between this type and `LedgerPeers` that encapsulates the Before/After slot semantics. You could perhaps add the slot info and have a semigroup operation that would just keep the most recent :)

bolt12 (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -255,20 +259,33 @@ ledgerPeersThread PeerActionsDNS {
...
-                     bigPeersStakeMap = accBigPoolStakeMap peers
+                 (ledgerStateJudgement, consensusSlotNo, consensusPeers, peerSnapshot) <-
+                   atomically ((,,,) <$> lpGetLedgerStateJudgement
+                                     <*> lpGetLatestSlot
+                                     <*> getLedgerPeers wlpConsensusInterface ula
+                                     <*> wlpGetLedgerPeerSnapshot)
+
+                 -- we have to assess which of, if any, peers we get from consensus vs.
+                 -- peers we may have from the snapshot file is more recent, and use that
+                 let (accPoolStake -> peersStake, bigPeersStakeMap) =
```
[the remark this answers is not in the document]
crocodile-dentist: You're right, and I had this fixed in the next commit in the sequence, but I'll move it to this patch.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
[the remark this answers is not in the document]
crocodile-dentist: That's a neat idea, but will it work? I have to treat the snapshot peers a little differently from than the peers I get from consensus, because the snapshot peers were already computed to contain just the big ledger peers, and I don't want to compute big ledger peers of the big ledger peers - I use the function Map.fromAscList to retrieve the snapshot peers vs. accPoolStake/accBigPoolStake for the ones I get from the ledger. Maybe if I throw some tag in there as well to distinguish... On the other hand, do you have an idea whether we could use this Semigroup instance or if we could generalize over this pattern somewhere else in the code? It seems like it is just one-off and so having it 'inlined' here like I did may be actually clearest, but maybe I'm not seeing something.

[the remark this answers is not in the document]
bolt12: Ahh that's right I forgot that small detail, given that I don't think it is worth then

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Utils.hs
crocodile-dentist: I've also added test for idempotency

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,6 +28,15 @@ import GHC.Generics
...
 import Ouroboros.Network.PeerSelection.RelayAccessPoint (RelayAccessPoint)
 
+-- | Peer snapshot type used to serialize peers from the
+-- current state of a node's ledger:
+-- LocalStateQuery app server running in ouroboros-consensus-diffusion
+-- receives a query (eg. cardano-cli) demanding a dump of this data
+-- from the current ledger. Consensus uses CBOR.
+newtype LedgerPeerSnapshot = LedgerPeerSnapshot
+  { unLedgerPeerSnapshot :: (WithOrigin SlotNo, [(AccPoolStake, (PoolStake, NonEmpty RelayAccessPoint))]) }
+  deriving newtype (Show, ToCBOR, FromCBOR)
```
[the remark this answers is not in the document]
bolt12: hmm I see, I wonder if getting those from cardano-binary is okay, at the time I was working with peer sharing I remember we were striving to avoid automatically derived encoders for the reasons I said above. Not sure if the situation changed. cc @coot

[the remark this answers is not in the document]
crocodile-dentist: The WithOrigin and SlotNo types have automatically derived instances of Serialise, and then in cardano-slotting there's a hand-cranked ToCBOR and FromCBOR that calls out to encode/decode. I suppose I could propose a patch there so that we have manual instances for everything.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -15,17 +22,137 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+                      fail "Invalid ReStake value")
+                  . readMaybe
+            relays <- poolV .: "Relays"
+            return (accStake, (reStake, relays))) <?> Key "BigLedgerPools"
+
+          return $ LedgerPeerSnapshotV1 (slot, bigPools')
+        _ -> fail $ "Network.LedgerPeers.Type: parseJSON: failed to parse unsupported version " <> show vNum
+    case migrateLedgerPeerSnapshot parsedSnapshot of
+      Just payload -> return $ LedgerPeerSnapshot payload
+      Nothing      -> fail "Network.LedgerPeers.Type: parseJSON: failed to migrate big ledger peer snapshot"
```
bolt12, line 110: We also wouldn't need this migration function since the codecs would have to be backward compatible and indexed by `LedgerSnapshotVersion`

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -15,17 +22,137 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+  fromCBOR = do
+    Codec.decodeListLenOf 3
+    version <- Codec.decodeWord8
+    case version of
+      1 -> LedgerPeerSnapshotV1 <$> do
+             (WithOriginWrapper slot, pools) <- (,) <$> fromCBOR <*> fromCBOR
+             let pools' = [(AccPoolStake accStake, (PoolStake reStake, relays))
+                          | (accStake, reStake, relays) <- pools]
+             return (slot, pools')
+      _ -> fail $ "LedgerPeers.Type: no decoder could be found for version " <> show version
```
bolt12, line 154: Have this be indexed by `LEdgerSnapshotVersion`

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -248,32 +250,53 @@ ledgerPeersThread PeerActionsDNS {
...
+                 (consensusSlotNo, consensusPeers, peerSnapshot) <-
+                   atomically ((,,) <$> lpGetLatestSlot wlpConsensusInterface
+                                    <*> getLedgerPeers wlpConsensusInterface ula
+                                    <*> wlpGetLedgerPeerSnapshot)
+
+                 -- we have to assess which of, if any, peers we get from consensus vs.
+                 -- peers we may have from the snapshot file is more recent, and use that
+                 (peersStakeMap, bigPeersStakeMap, lastSnapshotSlot'') <-
+                   case (consensusSlotNo, consensusPeers, peerSnapshot) of
+                     (At t, LedgerPeers _ lp, Just (LedgerPeerSnapshot (At t', sp {- snapshot peers -})))
```
bolt12, line 271:
```suggestion
                     (At t, LedgerPeers _ lp, Just (LedgerPeerSnapshot (At t', snapshotPeers)))
```

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
```diff
@@ -161,6 +172,122 @@ instance Arbitrary ArbLedgerPeersKind where
...
+                                    AllLedgerPeers -> ledgerSet
+                                    BigLedgerPeers -> bigLedgerSet
+            in if peers `Set.isSubsetOf` source
+               then return pass
+               else Left $ intercalate "\n"
+                             ["Counterexample:", "Requested: "++ show numRequested,
+                              "Type requested: " ++ show lpk,
+                              "Ledger slot: " ++ show ls, "Snapshot slot? " ++ show (ss <$ spool),
+                              show peers, "========================", "violates Set.isSubsetOf:",
+                              "========================", show source]
```
bolt12, line 288: Why is this test so complicated? Why don't you factor out the request logic and then just test that single function instead of running a simulation?

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
```diff
@@ -342,6 +470,31 @@ prop_accBigPoolStake  (LedgerPools lps@(_:_)) =
...
+    -- compare relay access points in both lists for equality
+    -- where we assume that recomputerelativestake doesn't change
+    -- the order, and sum up relative stake to make sure it adds up to 1
+    go ((normPoolStake, raps):rest) ((_, raps'):rest') (accStake, _) =
+      if raps == raps' && normPoolStake >= 0
+      then go rest rest' (accStake + normPoolStake, True)
+      else (accStake + normPoolStake, False)
+    go [] (_:_) (accStake, _) = (accStake, False)
+    go (_:_) [] (accStake, _) = (accStake, False)
+    go _ _ (accStake, relayAccessPointsUnchangedNonNegativeStake) = (accStake, relayAccessPointsUnchangedNonNegativeStake)
```
bolt12, line 495: What does this test? why was it added?

bolt12 (changes requested):

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
crocodile-dentist: This is used when we compute big ledger peers, I wanted to increase the confidence that they are correct for Genesis ofc, and so some simple properties I came up with are:
1. The accumulated relative stake adds up to unity
2. No pool relative stake can be less than 0
3. The relays aren't mangled
4. Running this function multiple times always produces the same result

bolt12: I would add some haddocks explaining so its easier to understand

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
crocodile-dentist: Done

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
crocodile-dentist: I've added a comment.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +58,49 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
...
+      0 -> do
+        domain <- fromCBOR
+        return $ RelayAccessDomain domain port
+      1 -> do
+        ip4 <- IP.IPv4 . IP.toIPv4 <$> fromCBOR
+        return $ RelayAccessAddress ip4 port
+      2 -> do
+        ip6 <- IP.IPv6 . IP.toIPv6 <$> fromCBOR
+        return $ RelayAccessAddress ip6 port
+      _ -> fail $ "Unrecognized RelayAccessPoint tag: " <> show constructorTag
```
coot, line 102: @bolt12 don't we have functions which do the serialisation / deserialisation of IP addresses for the peer sharing mini-protocol?

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
coot: I am not in particular happy about using `{To,From}CBOR` type classes. We avoided using them in other parts of the codecs (e.g. mini-protocol codecs), but if we're careful, we can avoid that.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
coot: If we encode `Rational`s as `Scientific` values, we'll lose the property that they add up to `1`, but maybe that's fine.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
bolt12: Yes I think you are right! We can reuse them

coot: They depend on `NodeToNodeVersion`. Unfortunately, we don't have access to them here, so we cannot reuse them.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -15,17 +22,137 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+           , "BigLedgerPools" .= [ object [ "AccStake" .= show accStake
+                                          , "ReStake"  .= show stake
+                                          , "Relays"   .= relays]
+                                 | (AccPoolStake accStake, (PoolStake stake, relays)) <- pools]]
+
+instance FromJSON LedgerPeerSnapshot where
+  parseJSON = withObject "LedgerPeerSnapshot" $ \v -> do
+    vNum <- v .: "Version"
+    parsedSnapshot <-
+      case (vNum :: Int) of
```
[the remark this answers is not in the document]
crocodile-dentist: Done

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -51,14 +178,14 @@ isLedgerPeersEnabled UseLedgerPeers {}  = True
 --
 newtype PoolStake = PoolStake { unPoolStake :: Rational }
   deriving (Eq, Ord, Show)
-  deriving newtype (Fractional, Num, NFData)
+  deriving newtype (Fractional, Num, NFData, Read)
```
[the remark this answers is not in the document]
crocodile-dentist: Serialisation is in floating per per last meeting.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -51,14 +178,14 @@ isLedgerPeersEnabled UseLedgerPeers {}  = True
...
 
 -- | The accumulated relative stake of a stake pool, like PoolStake but it also includes the
 -- relative stake of all preceding pools. A value in the range [0, 1].
 --
 newtype AccPoolStake = AccPoolStake { unAccPoolStake :: Rational }
     deriving (Eq, Ord, Show)
-    deriving newtype (Fractional, Num)
+    deriving newtype (Fractional, Num, Read)
```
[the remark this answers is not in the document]
crocodile-dentist: Done

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/LedgerPeers.hs
coot: We encode them as `Doubles`, which we have already been using internally.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -248,32 +253,36 @@ ledgerPeersThread PeerActionsDNS {
...
-                 return (peersStake, bigPeersStakeMap, now)
+                 (ledgerSlotNo, ledgerPools, peerSnapshot) <-
+                   atomically ((,,) <$> lpGetLatestSlot wlpConsensusInterface
+                                    <*> getLedgerPeers wlpConsensusInterface ula
+                                    <*> wlpGetLedgerPeerSnapshot)
+
+                 let (peersStakeMap, bigPeersStakeMap, lastSnapshotSlot'') =
+                       stakeMapWithSlotOverSource StakeMapOverSource {
+                                                    ledgerSlotNo, ledgerPools, peerSnapshot,
+                                                    lastSnapshotSlot, peerMap, bigPeerMap, ula}
```
bolt12:
```suggestion
               let (peersStakeMap, bigPeersStakeMap, lastSnapshotSlot'') =
                       stakeMapWithSlotOverSource 
                          StakeMapOverSource { 
                            ledgerSlotNo
                          , ledgerPools
                          , peerSnapshot
                          , lastSnapshotSlot
                          , peerMap
                          , bigPeerMap
                          , ula
                          }
```

bolt12: or

```suggestion
               let stakeMapOverSource = 
                     StakeMapOverSource { 
                        ledgerSlotNo
                      , ledgerPools
                      , peerSnapshot
                      , lastSnapshotSlot
                      , peerMap
                      , bigPeerMap
                      , ula
                     }
                   (peersStakeMap, bigPeersStakeMap, lastSnapshotSlot'') = 
                     stakeMapWithSlotOverSource stakeMapOverSource
                          
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -367,32 +368,34 @@ data StakeMapOverSource = StakeMapOverSource {
...
+    (_, LedgerPeers _ ledgerRelays, Nothing) -> (accPoolStake ledgerRelays, accBigPoolStakeMap ledgerRelays, Nothing)
+
+    (_, _, Just (LedgerPeerSnapshot (At snapshotSlotNo, accSnapshotRelays)))
+      | After slot <- ula, snapshotSlotNo >= slot -> do
+        case cachedSlot of
+          Just thatSlot | thatSlot == snapshotSlotNo ->
+                          (peerMap, bigPeerMap, cachedSlot)
+          _otherwise -> (accPoolStake (map snd accSnapshotRelays), Map.fromAscList accSnapshotRelays, Just snapshotSlotNo)
+
+    _otherwise -> (Map.empty, Map.empty, Nothing)
```
bolt12, line 400: Try to keep to 80 character width lines

bolt12 (approved):
LGTM just stylistic comments!

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
crocodile-dentist: Done

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
crocodile-dentist: Done

PR #4851, opened 2024-04-11
Introduction of serialization instances in support of ledger peer snapshot

crocodile-dentist:
# Description

This change adds support for serialization of ledger peers by defining instances for its subcomponent types. The snapshot type will be introduced in a future commit.

# Checklist

- Branch
- Pull Request
    - [ ] Self-reviewed the diff
      - What does this PR change?
    - [ ] Reviewer requested

### DIFF
--- a/ouroboros-network-api/CHANGELOG.md
+++ b/ouroboros-network-api/CHANGELOG.md
@@ -9,6 +9,11 @@
 * Transplanted `accBigPoolStake` and `reRelativeStake` from ouroboros-network
   `LedgerPeers` module to expose functionality that facilitates serializing
   of big ledger peers via LocalStateQuery miniprotocol.
+* Introduced `LedgerPeerSnapshot` type for values of big ledger peers obtained
+  from the ledger at a particular volatile tip.
+  * New type supports CBOR & JSON for serialisation purposes.
+  * Ledger peer snapshot is versioned in case changes need to be made to the
+    encoding format in the future.
 
 * Added `Measure` and `BoundedMeasure` instances for `SizeInBytes`.
 
--- a/ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
@@ -1,7 +1,15 @@
 {-# LANGUAGE DeriveAnyClass             #-}
 {-# LANGUAGE DeriveGeneric              #-}
-{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE DerivingVia                #-}
+{-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE PatternSynonyms            #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE TypeApplications           #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans       #-}
 
 -- | Various types related to ledger peers.  This module is re-exported from
 -- "Ouroboros.Network.PeerSelection.LedgerPeers".
@@ -15,16 +23,130 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
   , UseLedgerPeers (..)
   , AfterSlot (..)
   , LedgerPeersKind (..)
+  , LedgerPeerSnapshot (.., LedgerPeerSnapshot)
   , isLedgerPeersEnabled
   ) where
 
-import Cardano.Slotting.Slot (SlotNo (..), WithOrigin)
+import Control.Monad (forM)
+import Data.ByteString.Char8 qualified as BS
+import Data.List.NonEmpty (NonEmpty)
+import Data.Text.Encoding (decodeUtf8)
+import GHC.Generics (Generic)
+import Text.Read (readMaybe)
+
+import Cardano.Binary (FromCBOR (..), ToCBOR (..))
+import Cardano.Binary qualified as Codec
+import Cardano.Slotting.Slot (SlotNo (..), WithOrigin (..))
 import Control.Concurrent.Class.MonadSTM
 import Control.DeepSeq (NFData (..))
-import Data.List.NonEmpty (NonEmpty)
-import GHC.Generics
+import Data.Aeson
+import Data.Aeson.Types
 import NoThunks.Class
-import Ouroboros.Network.PeerSelection.RelayAccessPoint (RelayAccessPoint)
+import Ouroboros.Network.PeerSelection.RelayAccessPoint
+
+-- |The type of big ledger peers that is serialised or later
+-- provided by node configuration for the networking layer
+-- to connect to when syncing.
+--
+data LedgerPeerSnapshot =
+  LedgerPeerSnapshotV1 (WithOrigin SlotNo, [(AccPoolStake, (PoolStake, NonEmpty RelayAccessPoint))])
+  -- ^ Internal use for version 1, use pattern synonym for public API
+  deriving (Eq, Show)
+
+-- |Public API to access snapshot data. Currently access to only most recent version is available.
+-- Nonetheless, serialisation from the node into JSON is supported for older versions via internal
+-- api so that newer CLI can still support older node formats.
+--
+pattern LedgerPeerSnapshot :: (WithOrigin SlotNo, [(AccPoolStake, (PoolStake, NonEmpty RelayAccessPoint))])
+                           -> LedgerPeerSnapshot
+pattern LedgerPeerSnapshot payload <- LedgerPeerSnapshotV1 payload where
+  LedgerPeerSnapshot payload = LedgerPeerSnapshotV1 payload
+
+{-# COMPLETE LedgerPeerSnapshot #-}
+
+-- | In case the format changes in the future, this function provides a migration functionality
+-- when possible.
+--
+migrateLedgerPeerSnapshot :: LedgerPeerSnapshot
+                          -> Maybe (WithOrigin SlotNo, [(AccPoolStake, (PoolStake, NonEmpty RelayAccessPoint))])
+migrateLedgerPeerSnapshot (LedgerPeerSnapshotV1 lps) = Just lps
+
+instance ToJSON LedgerPeerSnapshot where
+  toJSON (LedgerPeerSnapshotV1 (slot, pools)) =
+    object [ "version" .= (1 :: Int)
+           , "slotNo" .= slot
+           , "bigLedgerPools" .= [ object [ "accumulatedStake" .= fromRational @Double accStake
+                                          , "relativeStake"  .= fromRational @Double relStake
+                                          , "relays"   .= relays']
+                                 | (AccPoolStake accStake, (PoolStake relStake, relays)) <- pools
+                                 , let relays' = fmap RelayAccessPointCoded relays]]
+
+instance FromJSON LedgerPeerSnapshot where
+  parseJSON = withObject "LedgerPeerSnapshot" $ \v -> do
+    vNum :: Int <- v .: "version"
+    parsedSnapshot <-
+      case vNum of
+        1 -> do
+          slot <- v .: "slotNo"
+          bigPools <- v .: "bigLedgerPools"
+          bigPools' <- (forM bigPools . withObject "bigLedgerPools" $ \poolV -> do
+            AccPoolStakeCoded accStake <- poolV .: "accumulatedStake"
+            PoolStakeCoded reStake <- poolV .: "relativeStake"
+            relays <- fmap unRelayAccessPointCoded <$> poolV .: "relays"
+            return (accStake, (reStake, relays))) <?> Key "bigLedgerPools"
+
+          return $ LedgerPeerSnapshotV1 (slot, bigPools')
+        _ -> fail $ "Network.LedgerPeers.Type: parseJSON: failed to parse unsupported version " <> show vNum
+    case migrateLedgerPeerSnapshot parsedSnapshot of
+      Just payload -> return $ LedgerPeerSnapshot payload
+      Nothing      -> fail "Network.LedgerPeers.Type: parseJSON: failed to migrate big ledger peer snapshot"
+
+-- | cardano-slotting provides its own {To,From}CBOR instances for WithOrigin a
+-- but to pin down the encoding for CDDL we provide a wrapper with custom
+-- instances
+--
+newtype WithOriginCoded = WithOriginCoded (WithOrigin SlotNo)
+
+-- | Hand cranked CBOR instances to facilitate CDDL spec
+--
+instance ToCBOR WithOriginCoded where
+  toCBOR (WithOriginCoded Origin) = Codec.encodeListLen 1 <> Codec.encodeWord8 0
+  toCBOR (WithOriginCoded (At slotNo)) = Codec.encodeListLen 2 <> Codec.encodeWord8 1 <> toCBOR slotNo
+
+instance FromCBOR WithOriginCoded where
+  fromCBOR = do
+    listLen <- Codec.decodeListLen
+    tag <- Codec.decodeWord8
+    case (listLen, tag) of
+      (1, 0) -> pure $ WithOriginCoded Origin
+      (1, _) -> fail "LedgerPeers.Type: Expected tag for Origin constructor"
+      (2, 1) -> WithOriginCoded . At <$> fromCBOR
+      (2, _) -> fail "LedgerPeers.Type: Expected tag for At constructor"
+      _      -> fail "LedgerPeers.Type: Unrecognized list length while decoding WithOrigin SlotNo"
+
+instance ToCBOR LedgerPeerSnapshot where
+  toCBOR (LedgerPeerSnapshotV1 (wOrigin, pools)) =
+       Codec.encodeListLen 2
+    <> Codec.encodeWord8 1
+    <> toCBOR (WithOriginCoded wOrigin, pools')
+    where
+      pools' =
+        [(AccPoolStakeCoded accPoolStake, (PoolStakeCoded relStake, neRelayAccessPointCoded))
+        | (accPoolStake, (relStake, relays)) <- pools
+        , let neRelayAccessPointCoded = fmap RelayAccessPointCoded relays]
+
+instance FromCBOR LedgerPeerSnapshot where
+  fromCBOR = do
+    Codec.decodeListLenOf 2
+    version <- Codec.decodeWord8
+    case version of
+      1 -> LedgerPeerSnapshotV1 <$> do
+             (WithOriginCoded wOrigin, pools) <- fromCBOR
+             let pools' = [(accStake, (relStake, relays'))
+                          | (AccPoolStakeCoded accStake, (PoolStakeCoded relStake, relays)) <- pools
+                          , let relays' = unRelayAccessPointCoded <$> relays]
+             return (wOrigin, pools')
+      _ -> fail $ "LedgerPeers.Type: no decoder could be found for version " <> show version
 
 -- | Which ledger peers to pick.
 --
@@ -53,12 +175,18 @@ newtype PoolStake = PoolStake { unPoolStake :: Rational }
   deriving (Eq, Ord, Show)
   deriving newtype (Fractional, Num, NFData)
 
+newtype PoolStakeCoded = PoolStakeCoded PoolStake
+  deriving (ToCBOR, FromCBOR, FromJSON, ToJSON) via Rational
+
 -- | The accumulated relative stake of a stake pool, like PoolStake but it also includes the
 -- relative stake of all preceding pools. A value in the range [0, 1].
 --
 newtype AccPoolStake = AccPoolStake { unAccPoolStake :: Rational }
     deriving (Eq, Ord, Show)
     deriving newtype (Fractional, Num)
+
+newtype AccPoolStakeCoded = AccPoolStakeCoded AccPoolStake
+  deriving (ToCBOR, FromCBOR, FromJSON, ToJSON) via Rational
 
 -- | A boolean like type.  Big ledger peers are the largest SPOs which control
 -- 90% of staked stake.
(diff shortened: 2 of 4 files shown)


ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +57,35 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
...
+    if | mode == 0 ->  do
+         domain <- fromCBOR
+         return $ RelayAccessDomain domain port
+       | mode == 1 -> do
+         ip4 <- IP.IPv4 . IP.toIPv4 <$> fromCBOR
+         return $ RelayAccessAddress ip4 port
+       | mode == 2 -> do
+         ip6 <- IP.IPv6 . IP.toIPv6 <$> fromCBOR
+         return $ RelayAccessAddress ip6 port
+
```
coot, line 88: This PR should also include a test that `fromCBOR` and `toCBOR` are compatible, e.g.
```hs
 fromCBOR . toCBOR = id
```

But also check that the `toCBOR` produces valid CBOR.  This is especially important if we don't have CDDL spec for this definition.   This can be done by encoding and then decoding with [`decodeTermToken`](https://hackage.haskell.org/package/cborg-0.2.10.0/docs/Codec-CBOR-FlatTerm.html#v:decodeTermToken) and validating with ` [`validFlatTerm`](https://hackage.haskell.org/package/cborg-0.2.10.0/docs/Codec-CBOR-FlatTerm.html#v:validFlatTerm); check `prop_codec_valid_cbor_encoding`.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +57,35 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
...
+    port <- fromInteger <$> fromCBOR @Integer
+    if | mode == 0 ->  do
+         domain <- fromCBOR
+         return $ RelayAccessDomain domain port
+       | mode == 1 -> do
+         ip4 <- IP.IPv4 . IP.toIPv4 <$> fromCBOR
+         return $ RelayAccessAddress ip4 port
+       | mode == 2 -> do
+         ip6 <- IP.IPv6 . IP.toIPv6 <$> fromCBOR
+         return $ RelayAccessAddress ip6 port
```
coot, line 87: Two things:

* `case` expressions are a better fit (one extension less to use)
* We need a `fail` case so one will get a nice exception when the wrong data is received.

```suggestion
    case mode of
      0 -> do
         domain <- fromCBOR
         return $ RelayAccessDomain domain port
      1 -> do
         ip4 <- IP.IPv4 . IP.toIPv4 <$> fromCBOR
         return $ RelayAccessAddress ip4 port
      2 -> do
         ip6 <- IP.IPv6 . IP.toIPv6 <$> fromCBOR
         return $ RelayAccessAddress ip6 port
      a -> do
         fail ... 
```

coot, line 87: Two things:

* `case` expressions are a better fit (one extension less to use)
* We need a `fail` case so one will get a nice exception when the wrong data is received.

```suggestion
    tag <- decodeWord8
    port <- fromInteger <$> fromCBOR @Integer 
    case tag of
      0 -> do
         domain <- fromCBOR
         return $ RelayAccessDomain domain port
      1 -> do
         ip4 <- IP.IPv4 . IP.toIPv4 <$> fromCBOR
         return $ RelayAccessAddress ip4 port
      2 -> do
         ip6 <- IP.IPv6 . IP.toIPv6 <$> fromCBOR
         return $ RelayAccessAddress ip6 port
      _ -> do
         fail ("RelayAccessPoint decoder: invalid tag " ++ show tag)
```

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +57,35 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
...
+-- for all its query responses. It appears they provide some improved
+-- debugging diagnostics over Serialize instances.
+instance ToCBOR RelayAccessPoint where
+  toCBOR = mconcat . \case
+    RelayAccessDomain domain port ->
+      [encodeWord8 0, serialize' port, toCBOR domain]
+    RelayAccessAddress (IP.IPv4 ip4) port ->
+      [encodeWord8 1, serialize' port, toCBOR (IP.fromIPv4 ip4)]
+    RelayAccessAddress (IP.IPv6 ip6) port ->
+      [encodeWord8 2, serialize' port, toCBOR (IP.fromIPv6 ip6)]
```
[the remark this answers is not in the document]
coot: Isn't it a bit odd to encode the port before the IP address, domain name? We're all used to `IP:PORT`s.  Although, it's binary encoding so maybe it doesn't really matter.

coot (changes requested):
Very nice, a few comments and a request to write some tests.

ouroboros-network-api/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 * Transplanted `accBigPoolStake` and `reRelativeStake` from ouroboros-network
   `LedgerPeers` module to expose functionality that facilitates serializing
   of big ledger peers via LocalStateQuery miniprotocol.
+* Added `ToCBOR` and `FromCBOR` instance definitions for `RelayAccessPoint`
+  in support of upcoming serialization of (big) ledger peers.
```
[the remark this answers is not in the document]
crocodile-dentist: Yes, consensus answers all queries with a toCBOR eventually.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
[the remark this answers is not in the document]
crocodile-dentist: Yes, but when port comes first, the FromCBOR instance looks slightly more elegant I think.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
crocodile-dentist: You're right, I've missed this.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +57,35 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
                       | RelayAccessAddress !IP.IP      !Socket.PortNumber
   deriving (Eq, Ord)
 
+-- | These instances are used to serialize 'LedgerPeerSnapshot'
+-- consensus LocalStateQuery server which uses these instances
+-- for all its query responses. It appears they provide some improved
+-- debugging diagnostics over Serialize instances.
+instance ToCBOR RelayAccessPoint where
```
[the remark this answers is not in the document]
crocodile-dentist: Unfortunately, I think that's the case indeed.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
crocodile-dentist: Done.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +57,49 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
...
+      0 -> do
+        domain <- fromCBOR
+        return $ RelayAccessDomain domain port
+      1 -> do
+        ip4 <- IP.IPv4 . IP.toIPv4 <$> fromCBOR
+        return $ RelayAccessAddress ip4 port
+      2 -> do
+        ip6 <- IP.IPv6 . IP.toIPv6 <$> fromCBOR
+        return $ RelayAccessAddress ip6 port
+      _ -> fail $ "Unrecognized RelayAccessPoint tag: " <> show constructorTag
```
bolt12, line 101: Is this according to the current CDDL spec?

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -51,7 +52,7 @@ isLedgerPeersEnabled UseLedgerPeers {}  = True
 --
 newtype PoolStake = PoolStake { unPoolStake :: Rational }
   deriving (Eq, Ord, Show)
-  deriving newtype (Fractional, Num, NFData)
```
bolt12, line 48: I think it would be better to write the instance by hand and document it on the CDDL spec.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
bolt12: Also CBOR codecs in ouroboros-network-protocols where the cddl spec is defined and the codecs are defined use:

```
import Codec.CBOR.Decoding qualified as CBOR
import Codec.CBOR.Encoding qualified as CBOR
```

and the `Codec` type. Take a look at how things are done there. I am not sure about defining the CDDL, though because this project doesn't deal with it... hmm

bolt12 (changes requested):

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
crocodile-dentist: I haven't seen the CDDL spec for this, but I could create one if this is a requirement. I guess we should establish who should own this spec in the first place, us, node or consensus?

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: One wrinkle here is that is you look at [Slot](https://github.com/IntersectMBO/cardano-base/blob/master/cardano-slotting/src/Cardano/Slotting/Slot.hs#L40) in cardano-slotting, the SlotNo type, which is a component of the snapshot, has a newtype automatically derived instance for Serialise, which is leveraged by the ToCBOR instance. Same for WithOrigin. And so I might as well bite the bullet and keep my auto derived instances as well I think.

crocodile-dentist: So I've issued a PR to cardano-base so that hand written instances can be made for the WithOrigin type for recording the slot number of the snapshot.

crocodile-dentist: So I've hand written all instances necessary to serialise the snapshot.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
crocodile-dentist: The path to writing a spec is opened now that all serialisation instances are hand written.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -15,17 +22,134 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+
+          return $ LedgerPeerSnapshotV1 (slot, bigPools')
+        _ -> fail $ "Network.LedgerPeers.Type: parseJSON: failed to parse unsupported version " <> show vNum
+    case migrateLedgerPeerSnapshot parsedSnapshot of
+      Just payload -> return $ LedgerPeerSnapshot payload
+      Nothing      -> fail "Network.LedgerPeers.Type: parseJSON: failed to migrate big ledger peer snapshot"
+
+-- | Hand cranked CBOR instances to facilitate CDDL spec
+--
+instance {-# OVERLAPPING #-} ToCBOR (WithOrigin SlotNo) where
```
bolt12, line 114: Why do you need OVERLAPPING pragma?

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -15,17 +22,134 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+    case listLen of
+      1 -> do
+        tag <- Codec.decodeWord8
+        if tag == 0 then return Origin else fail "LedgerPeers.Type: Expected tag for Origin constructor"
+      2 -> do
+        tag <- Codec.decodeWord8
+        if tag == 1
+          then At . SlotNo <$> fromCBOR
+          else fail "LedgerPeers.Type: Expected tag for At constructor"
+      _ -> fail "LedgerPeers.Type: Unrecognized list length while decoding WithOrigin SlotNo"
```
bolt12, line 130:
```suggestion
instance {-# OVERLAPPING #-} FromCBOR (WithOrigin SlotNo) where
  fromCBOR = do
    listLen <- Codec.decodeListLen
    tag <- Codec.decodeWord8
    case (listLen, tag) of
      (1, 0) -> pure Origin
      (1, _) -> fail "LedgerPeers.Type: Expected tag for Origin constructor"
      (2, 1) -> At . SlotNo <$> fromCBOR
      (2, _) -> fail "LedgerPeers.Type: Expected tag for At constructor"
      _      -> fail "LedgerPeers.Type: Unrecognized list length while decoding WithOrigin SlotNo"
```

bolt12 (approved):
Looks good to me! I just left some small comments

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: There is a ```ToCBOR (WithOrigin a)``` instance in cardano-slotting - I've made a patch PR to make it OVERLAPPABLE there. I'll hold off on merging this one until it gets approved. If not, I guess I can just create a newtype wrapper.

bolt12: If you already have an instance for `WithOrigin` than it is not our responsibility to provide one since that's done in a different place. We should provide instance for types we own, so in this case we can't do much the way I see it! Sorry if I was misleading regarding CBOR instances, but I don't think we should go through the trouble of making these overlappable. They provide the `WithOrigin` instance and we just either need to provide the `SlotNo` instance or get it from somewhere if we don't own that type

crocodile-dentist: Their definition looks like this:
`instance {-# OVERLAPPABLE #-} (Serialise t, Typeable t) => ToCBOR (WithOrigin t) where
  toCBOR = encode`
I guess to have full control of the encoding this is not really useable. I've created a newtype wrapper instead.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
```diff
@@ -53,6 +58,49 @@ data RelayAccessPoint = RelayAccessDomain  !DNS.Domain !Socket.PortNumber
                       | RelayAccessAddress !IP.IP      !Socket.PortNumber
   deriving (Eq, Ord)
 
+-- | These instances are used to serialize 'LedgerPeerSnapshot'
+-- consensus LocalStateQuery server which uses these instances
+-- for all its query responses. It appears they provide some improved
+-- debugging diagnostics over Serialize instances.
+instance ToCBOR RelayAccessPoint where
```
coot, line 65: Do we have a round trip tests for `{To,From}CBOR` instances?

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -15,17 +22,137 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+    listLen <- Codec.decodeListLen
+    tag <- Codec.decodeWord8
+    case (listLen, tag) of
+      (1, 0) -> pure $ WithOriginWrapper Origin
+      (1, _) -> fail "LedgerPeers.Type: Expected tag for Origin constructor"
+      (2, 1) -> WithOriginWrapper . At . SlotNo <$> fromCBOR
+      (2, _) -> fail "LedgerPeers.Type: Expected tag for At constructor"
+      _      -> fail "LedgerPeers.Type: Unrecognized list length while decoding WithOrigin SlotNo"
+
+instance ToCBOR LedgerPeerSnapshot where
```
coot, line 135: Do we have round trip tests for `{To,From}CBOR` instances?

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -15,17 +22,137 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+{-# COMPLETE LedgerPeerSnapshot #-}
+
+-- | In case the format changes in the future, this function provides a migration functionality
+-- when possible.
+--
+migrateLedgerPeerSnapshot :: LedgerPeerSnapshot
+                          -> Maybe (WithOrigin SlotNo, [(AccPoolStake, (PoolStake, NonEmpty RelayAccessPoint))])
+migrateLedgerPeerSnapshot (LedgerPeerSnapshotV1 lps) = Just lps
+
+instance ToJSON LedgerPeerSnapshot where
```
coot, line 71: Do we have round trip tests for `{To,From}JSON` instances?

coot (changes requested):

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/RelayAccessPoint.hs
crocodile-dentist: Yes.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: Yes.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: Yes.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -15,17 +22,137 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+      _      -> fail "LedgerPeers.Type: Unrecognized list length while decoding WithOrigin SlotNo"
+
+instance ToCBOR LedgerPeerSnapshot where
+  toCBOR (LedgerPeerSnapshotV1 (slot, pools)) =
+       Codec.encodeListLen 3
+    <> Codec.encodeWord8 1
+    <> toCBOR (WithOriginWrapper slot)
+    <> toCBOR pools'
+    where
+      pools' = map (\(AccPoolStake accPoolStake, (PoolStake reStake, relays)) -> (accPoolStake, reStake, relays)) pools
```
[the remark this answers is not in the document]
crocodile-dentist: It encodes it as a two element list of Int's: the numerator and denominator.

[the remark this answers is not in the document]
coot: Let's document the encoding in the `LedgerPeerSnapshot` type.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -51,14 +178,14 @@ isLedgerPeersEnabled UseLedgerPeers {}  = True
 --
 newtype PoolStake = PoolStake { unPoolStake :: Rational }
   deriving (Eq, Ord, Show)
-  deriving newtype (Fractional, Num, NFData)
+  deriving newtype (Fractional, Num, NFData, Read)
```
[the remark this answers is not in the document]
crocodile-dentist: Done

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -51,14 +178,14 @@ isLedgerPeersEnabled UseLedgerPeers {}  = True
...
 
 -- | The accumulated relative stake of a stake pool, like PoolStake but it also includes the
 -- relative stake of all preceding pools. A value in the range [0, 1].
 --
 newtype AccPoolStake = AccPoolStake { unAccPoolStake :: Rational }
     deriving (Eq, Ord, Show)
-    deriving newtype (Fractional, Num)
+    deriving newtype (Fractional, Num, Read)
```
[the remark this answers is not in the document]
crocodile-dentist: Done

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: There's a test that checks if:
1. CBOR encoding is valid
2. CBOR round trip = identity
3. JSON decodes with the following stipulations:

   - since I've fully qualified every domain name, round trip equality test is modulo this
   - since serialisation of stake is in floating point, comparison to original Rational values is approximate

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: Yes, see above.

ouroboros-network-api/src/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -85,3 +212,40 @@ data LedgerPeersConsensusInterface m = LedgerPeersConsensusInterface {
...
+               Just addr ->
+                 return . RelayAccessPointCoded $ RelayAccessAddress addr port
+            . readMaybe
+
+      Just domain'
+        | Just (_, '.') <- BS.unsnoc domain' ->
+          return . RelayAccessPointCoded $ RelayAccessDomain domain' port
+        | otherwise ->
+          let fullyQualified = domain' `BS.snoc` '.'
+          in return . RelayAccessPointCoded $ RelayAccessDomain fullyQualified port
```
coot, line 251: So this is only to not parse `1.1.1.1.` as an `IP` address.  For that we'd also need a newer version of `iproute`.

I  would avoid this at all. This should be in the ledger rules.

crocodile-dentist: No, IP addresses are handled by the domain constructor. This is a check to ensure that domain name we have read is fully qualified, and if not, add a period at the end. I realize now that this is redundant since I serialise all domain names in fully qualified format - maybe I wrote the former before I added the latter.

coot (approved):
LGTM

PR #4861, opened 2024-04-25
Rebranded SharedPeers as NonRootPeers

coot:
This is less misleading, especially since ex-local root peers might fall
into that category, and they were never received through peer sharing.


crocodile-dentist (approved):
LGTM

karknu (approved):
LGTM

PR #4872, opened 2024-05-06
Adds IOSimPOR diffusion tests

bolt12:
# Description

Refactors the diffusion test suite to be easily run on IOSim and IOSimPOR.

Needs: https://github.com/input-output-hk/io-sim/pull/153

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [x] Added labels.

### DIFF
--- a/network-mux/demo/mux-demo.hs
+++ b/network-mux/demo/mux-demo.hs
@@ -19,7 +19,7 @@ import Control.Exception (finally)
 import Control.Monad
 import Control.Tracer (Tracer (..), nullTracer, showTracing)
 
-import System.Environment
+import System.Environment qualified as SysEnv
 import System.Exit
 import System.IO
 
@@ -43,7 +43,7 @@ import Test.Mux.ReqResp
 
 main :: IO ()
 main = do
-    args <- getArgs
+    args <- SysEnv.getArgs
     case args of
       ["server"]         -> server
       ["client", n, msg] -> client (read n) msg
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
@@ -97,6 +97,7 @@ import Ouroboros.Network.PeerSelection.State.LocalRootPeers (HotValency (..),
 import Ouroboros.Network.PeerSharing (PeerSharingResult (..))
 import Test.Ouroboros.Network.LedgerPeers (LedgerPools (..))
 
+import Control.Monad.Class.MonadTest (exploreRaces)
 import Ouroboros.Network.PeerSelection.Bootstrap (requiresBootstrapPeers)
 import Ouroboros.Network.PeerSelection.LedgerPeers
 
@@ -109,8 +110,12 @@ tests =
     , testProperty "diffusionScript command script valid"
                    prop_diffusionScript_commandScript_valid
     ]
+  , testGroup "IOSimPOR"
+    [ nightlyTest $ testProperty "no failure"
+                      (testWithIOSimPOR prop_diffusion_nofail 1000)
+    ]
   , testProperty "no failure"
-                 prop_diffusion_nofail
+                 (testWithIOSim prop_diffusion_nofail 125000)
   , testProperty "no livelock"
                  prop_diffusion_nolivelock
   , testProperty "dns can recover from fails"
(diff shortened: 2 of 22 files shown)


network-mux/demo/mux-demo.hs
```diff
@@ -28,7 +32,7 @@ import Data.Bits
 import System.IOManager
 import System.Win32
 import System.Win32.Async qualified as Win32.Async
-import System.Win32.NamedPipes
+import "Win32-network" System.Win32.NamedPipes
```
coot, line 35: Why do we need package imports here?

bolt12: Because some of your changes got merged onto most recent "Win32" and there's conflicts on the import names

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -203,6 +208,35 @@ tests =
...
+              -> AbsBearerInfo
+              -> DiffusionScript
+              -> Property
+testWithIOSim f traceNumber bi ds =
+  let sim :: forall s . IOSim s Void
+      sim = diffusionSimulation (toBearerInfo bi)
+                                ds
+                                iosimTracer
+   in labelDiffusionScript ds
+    $ f (runSimTrace sim) traceNumber
```
coot, line 272: You could simplify the type signature to:
```
testWithIOSim :: (SimTrace Void -> Property)
              -> AbsBearerInfo
              -> DiffusionScript
              -> Property
```
and have the `traceNumber` defined: either inside `f` or passed to `f` if different values are used whether we run it with `IOSim` or `IOSimPOR`.

The same with `testWithIOSimPOR`.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
```diff
@@ -3312,17 +3158,26 @@ unit_peer_sharing =
...
+               -- * first peer sharing request will be issued after
+               --   `policyPeerSharAcitvationDelay = 300`
+               -- * this request will not bring any new peers, because non of the peers
+               --    are yet mature
+               -- * inbound connections become mature at 900s (15 mins)
+               -- * next peer share request happens after 900s, e.g. around 1200s.
+               . Trace.takeWhile (\se -> case se of
+                                          SimEvent    {seTime} -> seTime < Time 1250
+                                          SimPOREvent {seTime} -> seTime < Time 1250
+                                          _                    -> False
```
coot, line 3178: We could redefine `SimEvent` type and cheat that there's not other constructor with pattern matches and `COMPLETE` pragma; the `SimRacesFound` constructor is internal to `IOSimPOR` if I recall correctly.

coot (commented):
Very nice PR! Thanks @bolt12.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
bolt12: I know, I made the tradeoff of being explicit about the `traceNumber` when defining the test because some tests do require a different number of tests and it is easier to tweak in the future if we want. They are also available at the top level instead of having to go to each function manually change the value inside `f`

network-mux/demo/mux-demo.hs
coot: So proper way forward is to update `win32-network` to a newer `Win32` version.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
coot: It's still manual, just in a different place.

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
coot: I would hide it (since it makes a difference for all test authors), but this requires some care, e.g. if we define:
```
data SimEvent = SimeEvent_ {..} | SimPOREvent_ {..} | SimRacesFound_ {..}

pattern SimEvent = ...
pattern SimPOREvent = ...
{-# COMPLETE SimEvent, SimPOREvent #-}
```
Then we need to make sure that in `io-sim` use use the constructors not patterns (at least in `IOSimPOR`) to avoid strange bugs / warnings.

bolt12: Yes that would be better UI/UX but it should be done on io-sim side

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
bolt12: Yes but the place it is at now is more upfront in the sense that they are all in the same place: the test tree.

crocodile-dentist (approved):
I think it looks good.

PR #4884, opened 2024-05-22
Implement AnchoredSeq.splitAtMeasure

facundominguez:
This PR implements a function that allows to split an `AnchoredSeq` at an arbitrary measure boundary.

It is helpful in the Genesis implementation to split `AnchoredFragment`s. We need to count the amount of headers that a fragment contains in a window of slots (the genesis window). We currently do this with `dropWhileNewest`.

We use `dropWhileNewest` to cut the upper end of the fragment because we don't know the slot of the headers before or after the end of the window, which rules out `splitBeforePoint`, `splitBeforeMeasure`, etc.

Also tried `filter` to keep the headers in the window, but it was a bit slower than `dropWhileNewest`.

The implementation of `splitAtMeasure` provided here is at least as fast as `dropWhileNewest` in our measuring.

### Alternatives

Expose a more general function `split :: (v -> Bool) -> AnchoredSeq v a b -> AnchoredSeq v a b` which delegates to `FingerTree.split`.

Or expose the implementation of `AnchoredSeq`, so we can implement this function in a "Utils" module without having to upgrade or fork `ouroboros-network-api`.

### DIFF
--- a/ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
@@ -304,8 +304,10 @@ splitAt i (AnchoredSeq a ft) = case FT.split (\v -> measureSize v > i) ft of
 -- > anchor      before == anchor s
 -- > headAnchor  before == anchor after
 -- > headAnchor  after  == headAnchor s
--- > [before] == filter ((< v) . getElementMeasure . MeasuredWith) s
--- > [after] == filter ((v <=) . getElementMeasure . MeasuredWith) s
+-- > toOldestFirst before ==
+-- >   filter ((< v) . getAnchorMeasure @v @a (Proxy @b) . asAnchor) (toOldestFirst s)
+-- > toOldestFirst after ==
+-- >   filter ((v <=) . getAnchorMeasure @v @a (Proxy @b) . asAnchor) (toOldestFirst s)
 splitAtMeasure ::
       Anchorable v a b
    => v
--- a/ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
@@ -25,6 +25,7 @@ module Ouroboros.Network.AnchoredSeq
   , fromNewestFirst
   , fromOldestFirst
   , splitAt
+  , splitAtMeasure
   , dropNewest
   , takeOldest
   , dropWhileNewest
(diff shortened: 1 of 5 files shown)


ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
```diff
@@ -295,6 +296,26 @@ splitAt i (AnchoredSeq a ft) = case FT.split (\v -> measureSize v > i) ft of
...
+--  measure.
+--
+-- POSTCONDITION: @(before, after) = splitAtMeasure v s@, then:
+--
+-- > anchor      before == anchor s
+-- > headAnchor  before == anchor after
+-- > headAnchor  after  == headAnchor s
+-- > [before] == filter ((< v) . getElementMeasure . MeasuredWith) s
+-- > [after] == filter ((v <=) . getElementMeasure . MeasuredWith) s
+splitAtMeasure ::
```
amesgen, line 309: What about having a `AnchoredFragment`-specific wrapper (eg `splitAtSlot`), just like `splitAfterPoint` is a wrapper around `splitAfterMeasure`?

ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
```diff
@@ -295,6 +296,26 @@ splitAt i (AnchoredSeq a ft) = case FT.split (\v -> measureSize v > i) ft of
...
+-- | \( O(\log(\min(i,n-i)) \). Split the 'AnchoredSeq' at a given
+--  measure.
+--
+-- POSTCONDITION: @(before, after) = splitAtMeasure v s@, then:
+--
+-- > anchor      before == anchor s
+-- > headAnchor  before == anchor after
+-- > headAnchor  after  == headAnchor s
+-- > [before] == filter ((< v) . getElementMeasure . MeasuredWith) s
+-- > [after] == filter ((v <=) . getElementMeasure . MeasuredWith) s
```
amesgen, line 308: It would nice and relatively straightforward to add a property test for this to `Test.AnchoredFragment`.

Also, `getElementMeasure` and `MeasuredWith` are internal. Not a big deal per se; when testing, we can extract the slot number directly. Might be another small advantage of adding `splitAtSlot` as its laws could be tested unmodified.

amesgen, line 308: I think these two properties are false in boring ways when `s` is empty (as `filter` will then return an empty list for every predicate).

facundominguez: Fixed.

facundominguez: Added `splitAtSlot` and the tests.

ouroboros-network-api/src/Ouroboros/Network/AnchoredSeq.hs
facundominguez: I added `splitAtSlot`.

ouroboros-network-protocols/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Non-Breaking changes
 
+* Added `AnchoredFragment.splitAtSlot`
```
amesgen, line 11: This should go to the `ouroboros-network-api` changelog, right?

ouroboros-network-protocols/test/Test/AnchoredFragment.hs
```diff
@@ -477,6 +497,72 @@ prop_shrink_TestAnchoredFragmentAndPoint t@(TestAnchoredFragmentAndPoint c _) =
...
+prop_arbitrary_TestAnchoredFragmentAndSlot :: TestAnchoredFragmentAndSlot -> Property
+prop_arbitrary_TestAnchoredFragmentAndSlot (TestAnchoredFragmentAndSlot c slot) =
+  let onAnchoredFragment = slotOnFragment slot c in
+  cover 75 onAnchoredFragment      "slot on fragment"     $
+  cover 1 (not onAnchoredFragment) "slot not on fragment" $
+  AF.valid c
+
+prop_shrink_TestAnchoredFragmentAndSlot :: TestAnchoredFragmentAndSlot -> Property
+prop_shrink_TestAnchoredFragmentAndSlot t@(TestAnchoredFragmentAndSlot c _) =
+  foldr (.&&.) (property True)
```
amesgen, line 560:
```suggestion
  conjoin
```

amesgen (commented):
Looks great, I will let someone from the Network team approve

ouroboros-network-protocols/CHANGELOG.md
facundominguez: Fixed.

ouroboros-network-protocols/test/Test/AnchoredFragment.hs
facundominguez: Adopted.

Thanks for the review!

crocodile-dentist:
I will try to review this, but please bear with me as I need to analyze the code in context as well to make sense of it

coot (approved):
Thanks @facundominguez, LGTM.

facundominguez:
Rebased master to resolve conflicts in a changelog.

facundominguez:
Should I try merging at another time?

PR #4885, opened 2024-05-29
Fix tracing of InboundGovernor Counters

bolt12:
The value was running against a never updated cache, so it was always missing and tracing the same values.


crocodile-dentist (approved):
LGTM

karknu (approved):
LGTM

PR #4887, opened 2024-06-04
TX Submission Logic

coot:
# Description

This is a draft PR (work in progress) which introduces `tx-submission` logic responsible for choosing from which peer to download a tx.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/State.hs
```diff
@@ -241,6 +242,7 @@ tickTimedTxs now st@SharedTxState{ timedTxs
                                  , bufferedTxs } =
     let (expiredTxs, timedTxs') = Map.split now timedTxs
         expiredTxs'             =  -- Map.split doesn't include the `now` entry in any map
+                                   -- TODO: there must be a better way.
```
karknu, line 245: Short of getting the maintainers from Map to change their ways I cant think of one

coot: There's [`splitLookup`](https://hackage.haskell.org/package/containers-0.8/docs/Data-Map-Lazy.html#v:splitLookup) which reduces two traversals into one; subsequent `insert` is cheap, because it just inserts minimum element into the map.

coot: Fixed: https://github.com/IntersectMBO/ouroboros-network/pull/4887/commits/2a581d73851fa7d45c7f2dae4014c65ebefb7a61

coot:
Undrafted the PR, so it can run on our CI.

coot:
Experimental `cardano-node` branch (now based on the coming `cardano-node-10.3`): https://github.com/IntersectMBO/cardano-node/tree/coot/tx-submission-10.3.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/State.hs
```diff
@@ -224,48 +223,47 @@ instance ( NoThunks peeraddr
...
+      , ps { unacknowledgedTxIds    = unacknowledgedTxIds',
+             availableTxIds         = availableTxIds',
+             unknownTxs             = unknownTxs',
+             requestedTxIdsInflight = requestedTxIdsInflight
+                                    + txIdsToRequest }
+      )
+      else
+      ( 0
+      , 0
+      , []
```
coot: @bolt12 I think this is too conservative, why don't we return `txsToMempool` here?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,551 @@
...
+
+    fn :: PeerTxState txid tx -> Bool
+    fn PeerTxState { unacknowledgedTxIds,
+                     requestedTxIdsInflight,
+                     unknownTxs,
+                     downloadedTxs,
+                     requestedTxsInflight
+                   } =
+        --    hasTxIdsToAcknowledge st ps ||
+           requestedTxIdsInflight == 0 -- document why it's not <= maxTxIdsInFlightPerPeer
```
crocodile-dentist: Perhaps remove the first commented out line here? Also, would be nice if some additional documentation was provided here.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -0,0 +1,509 @@
...
+    -> m Void
+decisionLogicThread tracer counterTracer policy txChannelsVar sharedStateVar = do
+    labelThisThread "tx-decision"
+    go
+  where
+    go :: m Void
+    go = do
+      -- We rate limit the decision making process, it could overwhelm the CPU
+      -- if there are too many inbound connections.
+      threadDelay 0.005 -- 5ms
```
crocodile-dentist: Perhaps we should only delay for the difference between .005 and how long we were blocked by the check on line 484 below. Lets also make a top level constant for this value.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -0,0 +1,509 @@
...
+              pts@PeerTxState { score, scoreTs } =
+    let duration = diffTime now scoreTs
+        !drain   = realToFrac duration * scoreRate
+        !drained = max 0 $ score - drain in
+    pts { score   = min scoreMax $ drained + n
+        , scoreTs = now
+        }
+
+
+drainRejectionThread
```
crocodile-dentist, line 418: Is this used anywhere? If so, haddocks should be added.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -0,0 +1,509 @@
...
+                 unacknowledgedTxIds      = StrictSeq.empty,
+                 unknownTxs               = Set.empty,
+                 score                    = 0,
+                 scoreTs                  = Time 0,
+                 downloadedTxs            = Map.empty,
+                 toMempoolTxs             = Map.empty }
+               peerTxStates
+         }
+
+    -- TODO: this function needs to be tested!
```
crocodile-dentist, line 185: review comment todo, is this function tested?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -0,0 +1,444 @@
...
+
+       -- | The number of transaction identifiers that we have requested but
+       -- which have not yet been replied to. We need to track this it keep
+       -- our requests within the limit on the number of unacknowledged txids.
+       --
+       requestedTxIdsInflight   :: !NumTxIdsToReq,
+
+       -- | The size in bytes of transactions that we have requested but which
+       -- have not yet been replied to. We need to track this it keep our
+       -- requests within the limit on the number of unacknowledged txids.
```
crocodile-dentist: s/on the number of unacknowledged txids/of maximum size of all inflight transactions with this peer/

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -0,0 +1,509 @@
...
+    -- PeerTxAPI
+    --
+
+    submitTxToMempool :: Tracer m (TraceTxSubmissionInbound txid tx) -> txid -> tx -> m ()
+    submitTxToMempool txTracer txid tx =
+        bracket_ (atomically $ waitTSem mempoolSem)
+                 (atomically $ signalTSem mempoolSem)
+          $ do
+            res <- addTx
+            start <- getMonotonicTime
```
crocodile-dentist: Shouldn't these two be swapped?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -0,0 +1,509 @@
...
+          !peerTxAPI <-
+            modifyMVar channelsVar
+              \ TxChannels { txChannelMap } -> do
+                chann <- newEmptyMVar
+                let (chann', txChannelMap') =
+                      Map.alterF (\mbChann ->
+                                   let !chann'' = fromMaybe chann mbChann
+                                   in (chann'', Just chann''))
+                                 peeraddr
+                                 txChannelMap
```
crocodile-dentist, line 143: Is this necessary? Under what circumstances would we already have a channel for a peer - connections to multiple interfaces perhaps?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,551 @@
...
+  where
+    fn :: forall a.
+          (a, [(peeraddr, TxDecision txid tx)])
+       -> (a, Map peeraddr (TxDecision txid tx))
+    fn (a, as) = (a, Map.fromList as)
+
+
+-- | Order peers by how useful the TXs they have provided are.
+--
+-- TXs delivered late will fail to apply because they where included in
```
crocodile-dentist: s/where/were/

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -0,0 +1,571 @@
...
+                             && txid `Set.notMember` requestedTxsInflight
+                      )
+                      unacknowledgedTxIds
+
+    txsToMempool = [ (txid,tx)
+                   | txid <- toList toMempoolTxIds
+                   , tx <- maybeToList $ txid `Map.lookup` downloadedTxs
+                   ]
+    (toMempoolTxIds, _) =
+      StrictSeq.spanl (\txid -> txid `Map.member` downloadedTxs
```
crocodile-dentist: use filter, also the Map.member call is unnecessary since we use Map.lookup in the list comprehension above which references this.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -0,0 +1,444 @@
...
+      --      `unacknowledgedTxIds` sequences;
+      --    * @Map.keysSet bufferedTxs `Set.isSubsetOf` Map.keysSet referenceCounts@;
+      --    * all counts are positive integers.
+      --
+      referenceCounts          :: !(Map txid Int),
+
+      -- | A set of timeouts for txids that have been added to bufferedTxs after being
+      -- inserted into the mempool.
+      -- Every txid entry has a reference count in `referenceCounts`.
+      timedTxs                 :: Map Time [txid],
```
crocodile-dentist: A short comment on why this timeout is useful would be beneficial.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -0,0 +1,444 @@
...
+       -- requests within the limit on the number of unacknowledged txids.
+       --
+       requestedTxsInflightSize :: !SizeInBytes,
+
+       -- | The set of requested `txid`s.
+       --
+       requestedTxsInflight     :: !(Set txid),
+
+       -- | A subset of `unacknowledgedTxIds` which were unknown to the peer.
+       -- We need to track these `txid`s since they need to be acknowledged.
```
crocodile-dentist: The name unknownTxs is slightly confusing - perhaps we can at least enhance the haddock that these reference the transactions which were requested but did not receive in the reply ie. some sort of invalidated or withdrawn tx id's since the time they were announced.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,551 @@
...
+      where
+        -- Split `unacknowledgedTxIds'` into the longest prefix of `txid`s which
+        -- can be acknowledged and the unacknowledged `txid`s.
+        (acknowledgedTxIds, _) =
+          StrictSeq.spanl (\txid -> (txid `Map.member` bufferedTxs
+                                 || txid `Set.member` unknownTxs
+                                 || txid `Map.member` downloadedTxs)
+                                 && txid `Set.notMember` requestedTxsInflight
+                          )
+                          unacknowledgedTxIds
```
crocodile-dentist: This logic is duplicated in State.acknowledgeTxIds and it is a little involved so perhaps we should lift this out into its own definition just in case it has to be adjusted in the future.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,551 @@
...
+        numOfUnacked = fromIntegral (StrictSeq.length unacknowledgedTxIds)
+        numOfAcked   = StrictSeq.length acknowledgedTxIds
+        unackedAndRequested = numOfUnacked + requestedTxIdsInflight
+        txIdsToRequest =
```
crocodile-dentist: same as above

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,551 @@
...
+                     }
+    SharedTxState { peerTxStates,
+                    bufferedTxs,
+                    inflightTxs,
+                    inflightTxsSize,
+                    inSubmissionToMempoolTxs }
+    | inflightTxsSize > maxTxsSizeInflight
+    = Map.filter fn peerTxStates
+    | otherwise
+    = Map.filter gn peerTxStates
```
crocodile-dentist: We should add some haddocks for this distinction otherwise it isn't immediately obvious without looking at the definitions what is being done.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -0,0 +1,444 @@
...
+      --    * are already in the mempool (`Nothing` is inserted in that case),
+      --
+      -- We only keep live `txid`, e.g. ones which `txid` is unacknowledged by
+      -- at least one peer or has a `timedTxs` entry.
+      --
+      -- /Note:/ `txid`s which `tx` were unknown by a peer are tracked
+      -- separately in `unknownTxs`.
+      --
+      -- /Note:/ previous implementation also needed to explicitly tracked
+      -- `txid`s which were already acknowledged, but are still unacknowledged.
```
crocodile-dentist, line 195: This statement is confusing, in any case I think we can remove this comment and instead focus on and expand the haddock for `referenceCounts` and its conceptual relationship with `bufferedTxs`

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -0,0 +1,571 @@
...
+                , ps
+                )
+      where
+        wrongSizedTxs :: [(txid, SizeInBytes, SizeInBytes)]
+        wrongSizedTxs =
+            map (\(a, (b,c)) -> (a,b,c))
+          . Map.toList
+          $ Map.merge
+              (Map.mapMaybeMissing \_ _ -> Nothing)
+              (Map.mapMaybeMissing \_ _ -> Nothing)
```
crocodile-dentist: Map.dropMissing for both - also, the first tactic will not perform a size check if the transaction is gone from the available set, but it will still be added to downloaded transactions set. Is this possible? Perhaps the `collectTxsImpl` function should also take as an argument the `availableTxIds` as observed when the download request was made (or the `requestedTxIds` should be a set of `(txid, size)`? If perhaps a subsequent request for more txids's was made from this peer, and some matching transactions were downloaded from a different peer (end up in the bufferedTxs set), those txid's will be removed from the available set for this peer and hence the size check will not be performed?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -0,0 +1,571 @@
...
+        unknownTxs'  = unknownTxs ps <> notReceived
+
+        requestedTxsInflight' =
+          assert (requestedTxIds `Set.isSubsetOf` requestedTxsInflight ps) $
+          requestedTxsInflight ps  Set.\\ requestedTxIds
+
+        requestedSize = fold $ availableTxIds ps `Map.restrictKeys` requestedTxIds
+        requestedTxsInflightSize' =
+          -- TODO: VALIDATE size of received txs against what was announced
+          -- earlier;
```
crocodile-dentist: This can be removed now since we check transaction sizes above?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -0,0 +1,509 @@
...
+              Map.update (\case 1 -> Nothing; n -> Just $! pred n)
+                         txid inSubmissionToMempoolTxs
+
+            timedTxs' = Map.alter fn (addTime bufferedTxsMinLifetime now) timedTxs
+              where
+                fn :: Maybe [txid] -> Maybe [txid]
+                fn Nothing      = Just [txid]
+                fn (Just txids) = Just $! (txid:txids)
+
+            referenceCounts' = Map.alter fn txid referenceCounts
```
crocodile-dentist, line 348: Does this not break the invariant:
```
      --    * the txid count is equal to multiplicity of txid in all
      --      `unacknowledgedTxIds` sequences;
```
in `Types`?
[edit] It looks like we bump the reference count because we track the transaction in `timedTxs` map, but this breaks the multiplicity invariant, no? But I'm not sure if we rely on it anywhere...

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -0,0 +1,571 @@
...
+
+    gn :: Map txid Int
+       -> txid
+       -> Map txid Int
+    gn m txid = Map.alter af txid m
+
+    af :: Maybe Int
+       -> Maybe Int
+    af Nothing  = Just 1
+    af (Just n) = Just $! succ n
```
crocodile-dentist, line 265: Since we bump this reference count, do we need to be sure that there is still some peer offering this tx and that it hasn't been ackowledged yet?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -0,0 +1,444 @@
...
+      -- Once the count reaches 0, a tx is removed from `bufferedTxs`.
+      --
+      -- The `bufferedTx` map contains a subset of `txid` which
+      -- `referenceCounts` contains.
+      --
+      -- /Invariants:/
+      --
+      --    * the txid count is equal to multiplicity of txid in all
+      --      `unacknowledgedTxIds` sequences;
+      --    * @Map.keysSet bufferedTxs `Set.isSubsetOf` Map.keysSet referenceCounts@;
```
crocodile-dentist, line 213: Under what circumstances would the `bufferedTxs` be a proper subset of the keys of `referenceCounts`? 
[edit] the only place I can imagine if the key is still referred to by `timedTxs`
[edit2] Is there some meaning of a transaction not in the `bufferedTxs` but which has a non-zero reference count, ie. if the above is correct this would be that it is referred to only by `timedTxs`.

crocodile-dentist: To answer my own question, these would be undownloaded transactions, or already downloaded, adopted in a block, then timed out and announced by some late peer.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
crocodile-dentist: Nevermind, this sum is subtracted from the reference counts and potentially dropped if it becomes 0.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -0,0 +1,571 @@
...
+                (\txid _ -> txid `notElem` unacknowledgedTxIds
+                         && txid `Map.notMember` bufferedTxs)
+                availableTxIdsMap)
+
+        -- Add received txids to `unacknowledgedTxIds`.
+        unacknowledgedTxIds' = unacknowledgedTxIds <> txidsSeq
+
+        -- Add ignored `txs` to buffered ones.
+        -- Note: we prefer to keep the `tx` if it's already in `bufferedTxs`.
+        bufferedTxs' = bufferedTxs
```
crocodile-dentist, line 346: Should we update the `timedTxs` map with a new expiration deadline? If we acknowledge this transaction because at this time it is in the `bufferedTxs`, it's reference count will drop by 1. If it times out immediately after, and that is its last reference, we will re-download it if another peer immediately announces it to us again.
[edit] to be more specific, we could in principle update the timeouts after learning that the tx is still in the mempool (so txid is also in `ignoredTxIds`)

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V1.hs
```diff
@@ -361,17 +308,21 @@ txSubmissionInbound tracer (NumTxIdsToAck maxUnacked) mpReader mpWriter _version
...
+        let duration = diffTime end start
+        traceWith tracer $
+          TraceTxInboundAddedToMempool txidsAccepted duration
         let !accepted = length txidsAccepted
 
         traceWith tracer $ TraceTxSubmissionProcessed ProcessedTxCount {
             ptxcAccepted = accepted
-          , ptxcRejected = collected - accepted
+          , ptxcRejected = length txs - accepted
+          , ptxcScore    = 0 -- This implementatin does not track score
```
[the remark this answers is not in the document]
coot: Fixed

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
coot: We cannot request more txids (they are already requested) or txs (protocol cannot do it now) when txids are in-flight.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: `0.005` is quite short; I'd leave it as is.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: @karknu this function is not used anywhere; it's also not used anywhere in the commit which introduced it.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: no it isn't, I'll create an issue for it.

coot: https://github.com/IntersectMBO/ouroboros-network/issues/5151

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: :thinking: we need something like:
```hs
            start <- getMonotonicTime
            res <- addTx
            end <- getMonotonicTime
            atomically $ modifyTVar sharedStateVar (updateBufferedTx end res)
```
@karknu do you agree?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: Right, multiple connections from different interfaces to the same peer.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: Fixed.  I hid both `txDecisionLogicThread` and `drainRejectionThread` and added `txDecisionLogicThreads`, which runs using `concurrently`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
coot: Fixed with:
```hs
    txsToMempool = toList
                 . fmap (second fromJust)
                 . StrictSeq.takeWhileL (isJust . snd)
                 . fmap (\txid -> (txid, Map.lookup txid downloadedTxs))
                 $ acknowledgedTxIds
```

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
coot: This can happen after block adoption, then the mempool is freed of all the tx-s that went into the block, the `tx-submission` outbound side will no longer be able to serve them (it only has access to the mempool).

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
karknu: That function is called in https://github.com/IntersectMBO/ouroboros-consensus/blob/7526b30b02c742fcf5b4c2372e7c66f4df4d6bee/ouroboros-consensus-diffusion/src/ouroboros-consensus-diffusion/Ouroboros/Consensus/NodeKernel.hs#L343-L345 .

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
coot: Well spotted. I'll combine both (i.e. the one from the previous comment) into one function, as they are used together in all three places (the same calculation is performed in `State`).

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
coot: I'd like to keep it here until we remove the V1 implementation.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
crocodile-dentist: Is the `takeWhileL` appropriate here? `acknowledgedTxIds` is already a longest prefix, so we would be taking a prefix of the prefix. Can't we just do
```
    txsToMempool = [ (txid,tx)
                   | txid <- toList acknowledgedTxIds
                   , txid `Map.notMember` bufferedTxs
                   , tx <- maybeToList $ txid `Map.lookup` downloadedTxs
                   ]
```
[edit] This should work if there are gaps between downloaded transactions in order of their acknowledgement.
[edit2] on second thought, this should work as is so marking this as resolved.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
crocodile-dentist: Based on internal discussion, it is probably not worth the effort to update the expiration times for tx's which are still in the mempool.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
coot: > Map.dropMissing for both
:+1:

I think you're right, it could happen that we downloaded a `tx` from somebody else, and the peer will get away with a wrongly sized tx.  I'll go with `requestedTxIds :: Map txid SizeInBytes`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,550 @@
...
+
+    fn :: PeerTxState txid tx -> Bool
+    fn PeerTxState { unacknowledgedTxIds,
+                     requestedTxIdsInflight,
+                     unknownTxs,
+                     downloadedTxs,
+                     requestedTxsInflight
+                   } =
+           requestedTxIdsInflight == 0 -- document why it's not <= maxTxIdsInFlightPerPeer
+        && requestedTxIdsInflight + numOfUnacked <= maxUnacknowledgedTxIds
```
crocodile-dentist, line 438: This is just `numOfUnacked <= maxUnacknowledgedTxIds` now

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: True, the invariant no longer holds if the timeout is involved.

The reference counts are updated by the `drainRejectionThread` (using `tickTimedTxs`).

APIs: `collectTxs`, `acknowledgedTxIds`, `receivedTxsIds` preserve it (e.g. if inputs satisfy it, the output will too), except for this call here, e.g. `submitTxToMempool`.

We should add a test for `submitTxToMempool` and the `tickTimedTxs`.  It's essential to get the reference counts correct, and the invariant ensures this is the case.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
coot: True, though I'll leave it for `GHC` to optimise it :grin:.

coot:
^ rebased the PR

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: Now it's not exposed and called [here](https://github.com/IntersectMBO/ouroboros-network/blob/coot/tx-submission/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs#L519-L540); this way we don't miss calling it also in the diffusion simulation.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
coot: There are two changes in the invariant we're checking in the tests:
* bufferedtxs are no longer a subset of unacknowledged ones, the difference though must be subset of the timed txids, [ref](https://github.com/IntersectMBO/ouroboros-network/blob/coot/tx-submission/ouroboros-network/testlib/Test/Ouroboros/Network/TxSubmission/TxLogic.hs#L152-L160)
* reference counts need to account for the `txid`s in the timed script, [ref](https://github.com/IntersectMBO/ouroboros-network/blob/coot/tx-submission/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs#L519-L540)

I only the first part makes the invariant weaker, since for all tests except when `submitTxToMempool` is used, there are no timed `txid`s.

coot:
^ squashed commits which were ment for it (created through git absorb)

coot:
^ rebased, some minor conflicts were resolved

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,499 @@
...
+        requestedTxIdsInflight
+       } =
+           requestedTxIdsInflight == 0
+           -- if a peer has txids in-flight, we cannot request more txids or txs.
+        && requestedTxIdsInflight + numOfUnacked <= maxUnacknowledgedTxIds
+        && txIdsToRequest > 0
+      where
+        -- Split `unacknowledgedTxIds'` into the longest prefix of `txid`s which
+        -- can be acknowledged and the unacknowledged `txid`s.
+        (txIdsToRequest, _, unackedTxIds) = splitAcknowledgedTxIds policy sharedTxState peerTxState
```
crocodile-dentist, line 443: 'splitAcknowledgedTxIds' is called by 'gn' as well and can be lifted out to eg. below 'unrequestable' binding.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
```diff
@@ -0,0 +1,499 @@
...
+                                 availableTxIds,
+                                 unknownTxs
+                               } =
+          (    requestedTxIdsInflight == 0
+            && requestedTxIdsInflight + numOfUnacked <= maxUnacknowledgedTxIds
+            && txIdsToRequest > 0
+          )
+        || (underSizeLimit && not (Map.null downloadable))
+      where
+        numOfUnacked   = fromIntegral (StrictSeq.length unacknowledgedTxIds)
```
crocodile-dentist, line 460: If the 'splitAcknowledgedTxIds' is lifted out, this here can be removed as the same result is provided by the former.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -0,0 +1,558 @@
...
+                 . StrictSeq.takeWhileL (isJust . snd)
+                 . fmap (\txid -> (txid, Map.lookup txid downloadedTxs))
+                 $ acknowledgedTxIds
+
+    txsToMempoolMap = Map.fromList txsToMempool
+
+    toMempoolTxs' = toMempoolTxs <> txsToMempoolMap
+
+    (downloadedTxs', ackedDownloadedTxs) = Map.partitionWithKey (\txid _ -> txid `Set.member` liveSet) downloadedTxs
+    lateTxs = Map.filterWithKey (\txid _ -> txid `Map.notMember` txsToMempoolMap) ackedDownloadedTxs
```
crocodile-dentist, line 131: This is a little peculiar. So the late tx's are those that aren't going to the mempool from among the acked and downloaded tx's. If we look at the `txsToMempool` definition, we see that the txs on the way to the mempool are from the acknowledged and downloaded set. So we filter the `not (acked and downloaded)` from `acked and downloaded`? Maybe I'm not reading this right, or it can be re-stated differently or if it is correct then a haddock would be nice explaining the corner case this is for.

coot: I pushed a [commit](https://github.com/IntersectMBO/ouroboros-network/pull/4887/commits/f6b9e9dd30b0f62292b44f7bdde9da6bd384c5e2) which restores the logic of `toMempoolTxs` that shouldn't be changed by the commit: `tx-submission: splitAcknowledgedTxIds`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
coot: We cannot push it there because we need to access `peerTxState`, which is available only when evaluating `fn` and/or `gn`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/Decision.hs
coot: As above; I left it out because the type signature of `splitAcknowledgedTxIds` is clearer, while computing it is `O(1)`.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
coot: I also pushed a [commit](https://github.com/IntersectMBO/ouroboros-network/pull/4887/commits/90024f9920143a57fabd26e39897cb2685cf5f5c) which explains `lateTxs` better.

karknu (approved):
LGTM

PR #4888, opened 2024-06-05
Verification of big ledger peer snapshot file

crocodile-dentist:
# Description

A check was added to verify if big ledger stake pools from the peer snapshot file match what is provided by the ledger once syncing reaches the point at which the snapshot was ostensibly taken at. If there's a mismatch, an error is raised.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
@@ -654,6 +654,13 @@ peerSelectionGovernorLoop tracer
 
       <> Monitor.connections          actions st
       <> Monitor.jobs                 jobPool st
+      -- This job monitors for changes in big ledger peer snapshot file (eg. reload)
+      -- and copies it into the governor's private state. When a change is detected,
+      -- it also flips private state LedgerStateJudgement to TooYoung so that it
+      -- can launch the appropriate verification task in the job pool when external
+      -- LedgerStateJudgement is TooOld. If the verification job detects a discrepancy
+      -- vs. big peers on the ledger, it throws and the node is shut down.
+      <> Monitor.ledgerPeerSnapshotChange st actions
       -- In Genesis consensus mode, this is responsible for settings targets on the basis
       -- of the ledger state judgement. It takes into account whether
       -- the churn governor is running via a tmvar such that targets are set
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
@@ -12,11 +12,13 @@
 module Ouroboros.Network.PeerSelection.Governor.Monitor
   ( targetPeers
   , jobs
+  , jobVerifyPeerSnapshot
   , connections
   , localRoots
   , monitorLedgerStateJudgement
   , monitorBootstrapPeersFlag
   , waitForSystemToQuiesce
+  , ledgerPeerSnapshotChange
   ) where
 
 import Data.Map.Strict (Map)
@@ -25,7 +27,7 @@ import Data.Maybe (fromMaybe, isJust)
 import Data.Set (Set)
 import Data.Set qualified as Set
 
-import Control.Concurrent.JobPool (JobPool)
+import Control.Concurrent.JobPool (Job (..), JobPool)
 import Control.Concurrent.JobPool qualified as JobPool
 import Control.Exception (assert)
 import Control.Monad.Class.MonadSTM
@@ -42,7 +44,9 @@ import Ouroboros.Network.PeerSelection.Governor.ActivePeers
 import Ouroboros.Network.PeerSelection.Governor.Types hiding
            (PeerSelectionCounters)
 import Ouroboros.Network.PeerSelection.LedgerPeers.Type
-           (LedgerStateJudgement (..))
+           (LedgerPeerSnapshot (..), LedgerPeersConsensusInterface (..),
+           LedgerStateJudgement (..), compareLedgerPeerSnapshotApproximate)
+import Ouroboros.Network.PeerSelection.LedgerPeers.Utils
 import Ouroboros.Network.PeerSelection.PeerTrustable (PeerTrustable (..))
 import Ouroboros.Network.PeerSelection.PublicRootPeers qualified as PublicRootPeers
 import Ouroboros.Network.PeerSelection.State.EstablishedPeers qualified as EstablishedPeers
@@ -600,7 +604,8 @@ monitorLedgerStateJudgement :: ( MonadSTM m
                             => PeerSelectionActions peeraddr peerconn m
                             -> PeerSelectionState peeraddr peerconn
                             -> Guarded (STM m) (TimedDecision m peeraddr peerconn)
-monitorLedgerStateJudgement PeerSelectionActions{ readLedgerStateJudgement }
+monitorLedgerStateJudgement PeerSelectionActions{ getLedgerStateCtx = ledgerCtx@LedgerPeersConsensusInterface {
+                                                    lpGetLedgerStateJudgement = readLedgerStateJudgement } }
                             st@PeerSelectionState{ bootstrapPeersFlag,
                                                    publicRootPeers,
                                                    knownPeers,
@@ -608,7 +613,8 @@ monitorLedgerStateJudgement PeerSelectionActions{ readLedgerStateJudgement }
                                                    inProgressPromoteCold,
                                                    inProgressPromoteWarm,
                                                    ledgerStateJudgement,
-                                                   consensusMode }
+                                                   consensusMode,
+                                                   ledgerPeerSnapshot }
   | GenesisMode <- consensusMode =
     Guarded Nothing $ do
       lsj <- readLedgerStateJudgement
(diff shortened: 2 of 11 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -479,8 +480,10 @@ peerSelectionGovernor :: ( Alternative (STM m)
                       -> PeerSelectionActions peeraddr peerconn m
                       -> PeerSelectionPolicy  peeraddr m
                       -> PeerSelectionInterfaces peeraddr peerconn m
+                      -> STM m (Maybe LedgerPeerSnapshot)
+                      -> LedgerPeersConsensusInterface m
                       -> m Void
-peerSelectionGovernor tracer debugTracer countersTracer fuzzRng consensusMode actions policy interfaces =
+peerSelectionGovernor tracer debugTracer countersTracer fuzzRng consensusMode actions policy interfaces readBigLedgerSnapshot ledgerPeersCtx =
```
bolt12, line 486: You should not pass `LedgerPeersConsensusInterface` here. You should cascade the function you need through `PeerSelectionActions`, notice that PeerSelectionActions contains `readLedgerStateJudgement`

bolt12: The same for `STM m (Maybe LedgerPeerSnapshot)` put it inside `PeerSelectionInterfaces`

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -645,7 +664,7 @@ peerSelectionGovernorLoop tracer
       --   Check the definition site for more details, but in short, when the
       --   node changes to 'TooOld' state it will go through a purging phase which
       --   the 'waitForTheSystemToQuiesce' monitoring action will wait for.
-      <> Monitor.monitorLedgerStateJudgement actions st
+      <> Monitor.monitorLedgerStateJudgement actions readBigLedgerSnapshot ledgerPeersCtx st
```
bolt12, line 667: See here `actions` would already have everything you need.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -735,3 +745,37 @@ waitForSystemToQuiesce st@PeerSelectionState{
...
+        TooOld -> do
+          maybeLps <- readBigLedgerSnapshot
+          (snapshotSlotNo, snapshotPools) <-
+            case maybeLps of
+              Just (LedgerPeerSnapshot result) -> pure result
+              Nothing                          -> retry
+
+          check . (snapshotSlotNo ==) =<< lpGetLatestSlot
+          result <- (snapshotPools ==) . accumulateBigLedgerStake <$> lpGetLedgerPeers
+          completion Nothing (Just $ TraceVerifyPeerSnapshot result)
```
bolt12, line 778: This job should receive the ledger state judgement and not query for it. The peer selection governor has the current state and jobs should be issued according to it. Otherwise it becomes very hard to reason about. Sure there can be races but it shouldn't really matter if you are just verifying the snapshot is read only.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
```diff
@@ -537,8 +544,16 @@ peerSelectionGovernorLoop tracer
...
-    loop pst (Time 0) `catch` (\e -> traceWith tracer (TraceOutboundGovernorCriticalFailure e) >> throwIO e)
+                          pst
+                          readBigLedgerSnapshot
+                          ledgerPeersCtx =
+    handle (\e -> traceWith tracer (TraceOutboundGovernorCriticalFailure e) >> throwIO e) $ do
+      -- we run the verification job pre-emptively at startup because we may be stuck
+      -- in the TooOld ledger state before we catch up. Once ledger state turns YoungEnough,
+      -- this job ends. If for any reason ledger state turns TooOld from YoungEnough, this job
+      -- is restarted by the monitorLedgerStateJudgement job.
+      JobPool.forkJob jobPool (Monitor.jobVerifyPeerSnapshot readBigLedgerSnapshot ledgerPeersCtx)
```
bolt12, line 555: You don't need to start it preemptively, if the snapshot is invalid you want to shutdown diffusion so I don't think it makes sense for the governor to start without knowing if the snapshot is valid. I would synchronously verify the snapshot and only then start the loop.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
bolt12: You should throw `throwIO BigLedgerPeerSnapshotError` here instead of the `peerSelectionGovernorLoop` body.

bolt12 (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
bolt12: You should have a monitoring job then that is monitoring the change of the snapshot and runs the verification job asynchronously.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
crocodile-dentist: Looks like if I throw here, it will be caught be the handler which won't do what I need. I'd have to use async exceptions, but I think the flow control is more explicit and clearer in my current implementation. One can tell from reading the governor loop body that this condition can stop the node - similar to how we handle bootstrapPeersTimeout.

bolt12: yes and in the handler you can do some cleanup stuff and rethrow the exception in the, right?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
crocodile-dentist: Done

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor.hs
crocodile-dentist: Done

coot: @bolt12 did you mean `readBigLedgerSnapshot` and `ledgerPeersCtx`, or sth else?

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -617,7 +625,10 @@ monitorLedgerStateJudgement PeerSelectionActions{ readLedgerStateJudgement }
       return $ \_now ->
         Decision {
           decisionTrace = [TraceLedgerStateJudgementChanged lsj],
-          decisionJobs  = [],
+          decisionJobs = case (lsj, ledgerPeerSnapshot) of
+                           (TooOld, Just ledgerPeerSnapshot') ->
+                             [jobVerifyPeerSnapshot ledgerPeerSnapshot' ledgerCtx]
+                           _otherwise -> [],
```
coot: This job is fired when `ledgerStateJudgement` changes, it doesn't mean that's when the snapshot was taken.  It's quite possible that the node was started with a valid snapshot taken 3 months behind, which is well past the `ledgerStateJudgement` change.

We have to have to wait for the exact slot in which the ledger snapshot was taken.  Also note that we need to obtain ledger peers in the same `STM` transaction, otherwise if this is done in a separate thread it could happen that we will not be comparing the snapshot with the right ledger.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -744,3 +755,64 @@ waitForSystemToQuiesce st@PeerSelectionState{
...
+    completion queueJob trace = return . Completion $ \st _now ->
+      Decision {
+        decisionTrace = maybeToList trace,
+        decisionState = st,
+        decisionJobs  = maybeToList queueJob }
+
+    job = do
+      result <-
+        atomically $ do
+          check . (slot >=) =<< lpGetLatestSlot
```
coot: This should be in the other part, when `jobVerifyPeerSnapshot` is called.

coot (changes requested):
I don't think it's right to dependon `YoungEnough :: LedgerStateJudgement` when we want to fire a job when the current slot no is equal to snapshot's slot no.  When the snapshot was created has nothing to do with `YoungEnough` - which comes from `consensus` and is of the order of 3hrs, while the ledger peer snapshot can be few weeks if not months old.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
crocodile-dentist: The job which is started doesn't verify immediately, it does wait until we are >= the slot where the snapshot was taken, and then it does the verification. It does obtain the ledger peers in the same transaction as well, so all your concerns are indeed addressed.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/Monitor.hs
crocodile-dentist: I didn't like that the governor would be woken up every time the slot changed and then do it's various checks in the main loop when most of the time there'd be nothing to do but wait. So I moved this check here - there is no race because lpGetLatestSlot is not read anywhere else in the governor code. I think this is a better way.

crocodile-dentist:
I've changed it so it just traces whether verification was successful or not instead of terminating diffusion.

PR #4904, opened 2024-07-03
Full rework of the BlockFetch logic for bulk sync mode

Niols:


ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -19,1133 +19,152 @@ module Ouroboros.Network.BlockFetch.Decision
...
+          case chainSelStarvation of
+            ChainSelStarvationEndedAt time
+              | time < addTime (1 :: DiffTime) peersOrderStart ->
+                pure peersOrder
+            _ -> do
+              let peersOrder' =
+                    PeersOrder
+                      { peersOrderCurrent = Nothing,
+                        peersOrderOthers = msnoc peersOrderOthers peersOrderCurrent,
+                        peersOrderStart
```
facundominguez, line 167: The dynamo is being rotated but the start time is preserved. Unless this is updated elsewhere, I would expect:
```suggestion
                        time
```

facundominguez: Oh, but it is updated by the caller `checkLastChainSelStarvation`. Please, disregard the above comment.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Common.hs
```diff
@@ -0,0 +1,677 @@
...
+We do not maintain any such invariant and so we have to deal with there being
+gaps in the ranges we have already fetched or are yet to fetch. To keep the
+tracking simple we do not track the ranges themselves, rather we track the set
+of individual blocks without their relationship to each other.
+
+-}
+
+-- | Find the fragments of the chain suffix that we still need to fetch, these
+-- are the fragments covering blocks that have not yet been fetched and are
+-- not currently in the process of being fetched from this peer.
```
facundominguez, line 400: I'm failing to see see how the not-inflight blocks are determined in this function.

facundominguez (commented):
Some comments. Still a partial review.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -19,1133 +19,153 @@ module Ouroboros.Network.BlockFetch.Decision
...
 
+import Ouroboros.Network.BlockFetch.Decision.Common (FetchDecisionPolicy (..), PeerInfo, FetchDecision, FetchDecline (..),
+                                                     filterPlausibleCandidates, filterNotAlreadyFetched, filterNotAlreadyInFlightWithPeer,
+                                                     selectForkSuffixes)
+import Ouroboros.Network.BlockFetch.Decision.Deadline (fetchDecisionsDeadline,
+                                                      prioritisePeerChains, fetchRequestDecisions)
+import Ouroboros.Network.BlockFetch.Decision.BulkSync (fetchDecisionsBulkSync)
+import Ouroboros.Network.BlockFetch.ConsensusInterface (ChainSelStarvation(..))
 
 fetchDecisions
```
facundominguez, line 45: I'm not understanding yet which mechanism is ensuring that the dynamo ends up being the peer from which blocks are downloaded eventually.

Do we need a blob comment in this module or in `Decision.BulkSync` explaining what is the strategy to select downloading blocks?

facundominguez (commented):
More comments. Still not finished.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Deadline.hs
```diff
@@ -0,0 +1,450 @@
+{-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE NamedFieldPuns      #-}
+{-# LANGUAGE RankNTypes          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE TypeOperators       #-}
+
+-- | This module contains the part of the block fetch decisions process that is
+-- specific to the deadline mode.
+module Ouroboros.Network.BlockFetch.Decision.Deadline
```
facundominguez, line 10: Are there any changes worth reviewing in this module?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Common.hs
```diff
@@ -0,0 +1,677 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | This module contains the part of the block fetch decisions process that is
+-- common to both the bulk sync and deadline modes.
+module Ouroboros.Network.BlockFetch.Decision.Common (
```
facundominguez, line 8: Are there any changes worth reviewing in this module?

facundominguez (commented):
Ok, now I think this completes a first review of all of the PR.

ouroboros-network/src/Ouroboros/Network/BlockFetch/ClientState.hs
```diff
@@ -231,12 +236,19 @@ data PeerFetchInFlight header = PeerFetchInFlight {
      }
   deriving (Eq, Show)
 
+-- | Information associated to a block in flight.
+data PeerFetchBlockInFlight = PeerFetchBlockInFlight
+  deriving (Eq, Show)
```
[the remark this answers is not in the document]
Niols: It was only in preparation to adding a field to this datatype. Now outdated.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,344 @@
...
+import qualified Ouroboros.Network.AnchoredFragment as AF
+import Ouroboros.Network.Block
+import Ouroboros.Network.BlockFetch.ClientState (FetchRequest (..), PeersOrder (..), mcons)
+import Ouroboros.Network.BlockFetch.ConsensusInterface (FetchMode(FetchModeBulkSync))
+import Ouroboros.Network.BlockFetch.DeltaQ (calculatePeerFetchInFlightLimits)
+
+import Ouroboros.Network.BlockFetch.Decision.Common
+-- REVIEW: We should not import anything from 'Decision.Deadline'; if the need
+-- arises, we should move the interesting piece of code to 'Decision.Common'.
+-- This is to be done on demand.
```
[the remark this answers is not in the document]
Niols: That was more of a reminder to myself. I will remove it.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Common.hs
Niols: This comment comes from before; cf https://github.com/IntersectMBO/ouroboros-network/blob/49150d2679b0cca15ca7238b75b696abad6a53d9/ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs#L571-L576 on `main`. I think it is supposed to apply to both functions `filterNotAlreadyFetched` and `filterNotAlreadyInFlightWithPeer`. I am not necessarily very happy with this either and now is our chance to change it.

Niols: Fixed in d4a453db395bda67b848b8799a8a5fff328cc09a.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,344 @@
...
+    where
+      combineWithDeclined ::
+        MaybeT (WithDeclined peer) (a, peer) ->
+        ( Maybe (a, peer),
+          [(FetchDecline, peer)]
+        )
+      combineWithDeclined = second listConcatToList . runWithDeclined . runMaybeT
+
+-- FIXME: The 'FetchDeclineConcurrencyLimit' should only be used for
+-- 'FetchModeDeadline', and 'FetchModeBulkSync' should have its own reasons.
```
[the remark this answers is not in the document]
Niols: It still needs to be fixed, but I am not sure if I actually think that it would need fixing, I am just wondering about clarity. The peers are indeed declined because we limit ourselves to a concurrency of 1, so it is a pretty degenerate case of this constructor and I wouldn't want it to mixed things up. It is probably reasonable to keep things as-is.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
Niols: > I'm not understanding yet which mechanism is ensuring that the dynamo ends up being the peer from which blocks are downloaded eventually.

I don't think it necessarily happens. It _might_ if we download blocks faster than peers jump, because then the jumpers won't have a candidate big enough to contain the gross request, and then only the dynamo (or disengaged peers) will remain. In normal CSJ without disengaged peers, I would expect that we most of the time see this BlockFetch strategy align on the dynamo.

> Do we need a blob comment in this module or in `Decision.BulkSync` explaining what is the strategy to select peers and to form the next batch?

Sounds good to me. I will see if I can exploit the document written by @nfrisby and put it here.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,344 @@
...
+        candidates
+
+    -- Order the peers according to the peer order that we have been given,
+    -- then separate between declined peers and the others.
+    let peersOrdered =
+          [ (candidate, peerInfo)
+            | (candidate, peerInfo@(_, _, _, peer, _)) <- peers,
+              peer' <- mcons (peersOrderCurrent peersOrder) (peersOrderOthers peersOrder),
+              peer == peer'
+          ]
```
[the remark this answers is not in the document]
Niols: What is intended is to sort the list `peers` using the order defined in `peersOrder`. I think this is what is achieved here, although with quadratic complexity when we could do linear?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Common.hs
Niols: No, everything has just been extracted from `Decision.hs` at the beginning and almost not touched since. The few changes have to do with documentation and sometimes extraction of a helper that is needed in `BulkSync.hs`.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Deadline.hs
Niols: No, everything has just been extracted from `Decision.hs` at the beginning and not touched since.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,344 @@
...
+
+          -- Trim the fragments to the peer's candidate, keeping only blocks that
+          -- they may actually serve.
+          trimmedFragments <- trimFragmentsToCandidate thePeerCandidate (snd fragments)
+
+          -- Try to create a request for those fragments.
+          fetchRequestDecision
+            fetchDecisionPolicy
+            FetchModeBulkSync
+            0 -- bypass all concurrency limits. REVIEW: is this really what we want?
```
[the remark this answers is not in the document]
Niols: They are not cancelled; they do linger until the peers reply or timeout. I think this is what was intended in Nick's design (@nfrisby > do you confim?). I am also not sure there is a way to cancel a request in typed-protocol, although I supposed we could at least stop downloading the batch that we are interested in.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,344 @@
...
+          selectBlocksUpToLimits
+            blockFetchSize
+            0 -- number of request in flight
+            maxBound -- maximum number of requests in flight
+            0 -- bytes in flight
+            (1024 * 1024) -- maximum bytes in flight; one megabyte
+            <$> fragments
+
+    -- For each peer, check whether its candidate contains the gross request
+    -- in its entirety, otherwise decline it.
```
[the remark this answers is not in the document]
Niols: You're correct about the expectation that if the grossRequest can be served, then they will be able to serve the refined request. It does deserve to be set explicitly, though.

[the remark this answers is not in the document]
Niols: Is 0be0889f6f0bc7687556f19edff9e05f4821d62f enough or would you want more?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,344 @@
...
+selectThePeer
+  FetchDecisionPolicy {blockFetchSize}
+  fetchedBlocks
+  fetchedMaxSlotNo
+  peersOrder
+  theCandidate
+  candidates = do
+    -- Filter out from the chosen candidate fragment the blocks that have
+    -- already been downloaded, but keep the blocks that have a request in
+    -- flight.
```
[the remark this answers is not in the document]
Niols: The comment has been changed since then and so has the one of `filterNotAlreadyFetch`.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,344 @@
...
+  fetchedMaxSlotNo
+  theCandidate
+  thePeer@(status, inflight, gsvs, _, _)
+  thePeerCandidate =
+    let theDecision = do
+          -- Keep blocks that have not already been downloaded or that are not
+          -- already in-flight with this peer.
+          fragments <-
+            filterNotAlreadyFetched fetchedBlocks fetchedMaxSlotNo theCandidate
+              >>= filterNotAlreadyInFlightWithPeer inflight
```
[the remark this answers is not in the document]
Niols: I have since then factorised this call, indeed.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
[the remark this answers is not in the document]
facundominguez: My initial observation is wrong. But I think the resulting `peersOrdered` has elements in the same order as `peers`.

```
>  [ x | x<-"1234", y<-"4321", x==y ]
"1234"
```

I don't think the linear search in the result of `mcons` is accomplishing much.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -170,3 +159,17 @@ fetchDecisions
                       }
               traverse_ demoteCSJDynamoAndIgnoreInflightBlocks peersOrderCurrent
               pure peersOrder'
+
+      checkChangeOfCurrentPeer :: Maybe (any, PeerInfo header peer extra) -> PeersOrder peer -> m ()
+      checkChangeOfCurrentPeer theDecision PeersOrder {peersOrderCurrent, peersOrderOthers} =
```
facundominguez, line 168: Maybe rename to `maybeSetCurrentPeer`.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -148,17 +148,20 @@ fetchDecisions
...
-                      { peersOrderCurrent = Nothing,
-                        peersOrderOthers = msnoc peersOrderOthers peersOrderCurrent,
-                        peersOrderStart
-                      }
-              traverse_ demoteCSJDynamoAndIgnoreInflightBlocks peersOrderCurrent
-              pure peersOrder'
+          case peersOrderCurrent of
+            Just peersOrderCurrent_
+              | peerHasBlocksInFlight peersOrderCurrent_
+                  && lastStarvationTime >= addTime bulkSyncGracePeriod peersOrderStart ->
```
facundominguez, line 155: This condition is being reversed! But it looks to me like the new version is the correct one.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -12,7 +13,7 @@ module Ouroboros.Network.BlockFetch.Decision.BulkSync (
 
 import Control.Monad (filterM)
 import Control.Monad.Trans.Maybe (MaybeT (MaybeT, runMaybeT))
-import Control.Monad.Writer.CPS (Writer, runWriter, MonadWriter (tell))
+import Control.Monad.Writer.Strict (Writer, runWriter, MonadWriter (tell))
```
facundominguez, line 16: This is reversing an optimization attempt. I'm not sure if it is relevant.

facundominguez (commented):
Finished a pass over the updates. Left some comments.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
facundominguez: Ah, this a confusion on my side. Please disregard.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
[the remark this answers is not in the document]
Niols: That is a very good catch. I also came to the same conclusion; thank you for shining a light on those lines. Fixed in 8c07b585adedc1b651cac285f0c5369ba1cd1c00.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -307,17 +292,13 @@ fetchTheCandidate ::
...
+  theFragments
   thePeer@(status, inflight, gsvs, _, _)
   thePeerCandidate =
     let theDecision = do
           -- Keep blocks that have not already been downloaded or that are not
           -- already in-flight with this peer.
-          fragments <-
-            filterNotAlreadyFetched fetchedBlocks fetchedMaxSlotNo theCandidate
-              >>= filterNotAlreadyInFlightWithPeer inflight
+          fragments <- filterNotAlreadyInFlightWithPeer inflight =<< theFragments
```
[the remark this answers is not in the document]
Niols: No, because this time we also consider peers that are in-flight and ignored!

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
Niols: It does not just set the current peer, though, it also demotes the previous one. This is only done by calling a callback, though, so it's probably OK if it doesn't appear in the name. I'll go for `maybeSetCurrentPeerAndDemote` but let me know if you think this is too much.

Niols: Done in 7044561f22d538f4bd2511dc08197b5ac5af0aa2.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
Niols: Compared to previous commits, you mean? I actually forgot to flip the sign, initially, and it was caught by the tests.

Niols: I do think it is the right direction.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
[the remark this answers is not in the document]
Niols: Tried to make the comment better in 7dd843423abeae5be34ea55bdc5c4b90d773ef0b.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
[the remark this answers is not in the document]
Niols: Hum, so the fact that the peer has the `grossRequest` means it can definitely create a request of at least one block (as long as the gross request is not empty; maybe I should add an assertion). And then this request can either be included in the `grossRequest` (for instance if the peer has tighter bytes-in-flight limits) but it can also include the `grossRequest` (if the peer has much higher bytes-in-flight limits and also has more blocks than this).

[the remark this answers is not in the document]
Niols: It is very much worth documenting somewhere. I don't think this in-code documentation is the best place, though.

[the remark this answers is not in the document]
Niols: WDYT of 04be7c4a8c7f9093db1e5514a2d83f6f2b8ceed5?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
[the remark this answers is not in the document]
Niols: Removed it. I think `FetchDeclineConcurrencyLimit` is fine. If someone finds it annoying later on in a debugging session, then they can of course refine it.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
Niols: WDYT of f43032b1970c1343d1bd9c3e73f46def29d7fac0?

Niols: And 4205e6853fed5410512fbf9dc7c7aa4bf70f41c3 for the interaction with CSJ.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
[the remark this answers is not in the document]
facundominguez: Ah, why is that in-flight and ignored are discarded here but taken into account to select the peer?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -5,8 +5,86 @@
...
+--   request, as influenced by exactly which blocks are actually already
+--   currently in-flight with the chosen peer.
+--
+-- - Update @currentPeer@ and @currentStart@, if needed. Namely:
+--
+--   - If @thePeer /= currentPeer@, then make @thePeer@ the current peer and the
+--     best according to @peersOrder@, and reset @currentStart@ to now.
+--
+--   - If @thePeer == currentPeer@, but @inflight(thePeer)@ is empty, the reset
+--     @currentStart@ to now.
```
facundominguez, line 84: typo: ... **then** reset

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -5,8 +5,86 @@
...
+-- - If @grossRequest@ is empty, then terminate this iteration. Otherwise, pick
+--   the best peer (according to @peersOrder@) offering all of the blocks in
+--   @grossRequest@. We will call it @thePeer@. Because @currentPeer@, if it
+--   exists, is the best according to @peersOrder@, then it will be our
+--   preferred peer, as long as it can provide the @grossRequest@s.
+--
+-- - If the byte size of @inflight(thePeer)@ is below the low-water mark, then
+--   terminate this iteration. Otherwise, decide and send the actual next batch
+--   request, as influenced by exactly which blocks are actually already
+--   currently in-flight with the chosen peer.
```
facundominguez, line 76: Maybe typo: I would expect requests to be inhibited if the byte count is **above** some threshold.

Can the text explain what the low-water mark is?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -5,8 +5,86 @@
...
+--   - Stop considering the peer “current” and make them the worst according to
+--     the @peersOrder@.
+--
+-- - Let @theCandidate :: AnchoredFragment (Header blk)@ be the best candidate
+--   header chain among the ChainSync clients (eg. best raw tiebreaker among the
+--   longest).
+--
+-- - Let @grossRequest@ be the oldest blocks on @theCandidate@ that have not
+--   already been downloaded, are not in @inflightNonIgnored@, and total less
+--   than 20 mebibytes.
```
facundominguez, line 65: Why 20 mebibytes and not a lower limit?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -5,8 +5,86 @@
...
+--
+-- Natural language specification
+-- ------------------------------
+--
+-- Definitions:
+--
+-- - Let @inflight :: peer -> Set blk@ be the outstanding blocks, those that
+--   have been requested and are expected to arrive but have not yet.
+--
+-- - Let @inflightIgnored :: Set blk@ be the outstanding blocks that have been
```
facundominguez, line 22: `inflightIgnored` is updated below but never used for anything, therefore, why care about it?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -5,8 +5,86 @@
...
+--
+-- - If the byte size of @inflight(thePeer)@ is below the low-water mark, then
+--   terminate this iteration. Otherwise, decide and send the actual next batch
+--   request, as influenced by exactly which blocks are actually already
+--   currently in-flight with the chosen peer.
+--
+-- - Update @currentPeer@ and @currentStart@, if needed. Namely:
+--
+--   - If @thePeer /= currentPeer@, then make @thePeer@ the current peer and the
+--     best according to @peersOrder@, and reset @currentStart@ to now.
```
facundominguez, line 81: Shouldn't `currentStart` be set only if `inflight(thePeer)` is empty? Otherwise, the grace period is extended despite the fact that there are old requests that the peer hasn't responded to.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -5,8 +5,86 @@
...
+--
+-- - Let @grossRequest@ be the oldest blocks on @theCandidate@ that have not
+--   already been downloaded, are not in @inflightNonIgnored@, and total less
+--   than 20 mebibytes.
+--
+-- - If @grossRequest@ is empty, then terminate this iteration. Otherwise, pick
+--   the best peer (according to @peersOrder@) offering all of the blocks in
+--   @grossRequest@. We will call it @thePeer@. Because @currentPeer@, if it
+--   exists, is the best according to @peersOrder@, then it will be our
+--   preferred peer, as long as it can provide the @grossRequest@s.
```
facundominguez, line 71: If `thePeer` has `inflight` requests whose grace period is expired, will it be preferred just the same?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -85,6 +85,26 @@
...
+--
+-- - Because we always require our peers to be able to serve a gross request of
+--   oldest blocks, peers with longer chains have a better chance to pass this
+--   criteria and to be selected as current peer. The CSJ dynamo, being always
+--   ahead of jumpers, has therefore more chances to be selected as the current
+--   peer. It is still possible for a jumper or a disengaged peer to be
+--   selected.
+--
+-- - If the current peer is the CSJ dynamo, but it is a dishonest peer serving
+--   headers fast but retaining headers, it might be able to drastically leash
```
facundominguez, line 102: typo: ... retaining **blocks**

facundominguez (commented):
Reviewed all of the updates. The documentation is quite helpful!

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,613 @@
...
+--
+-- - If @inflight@ is non-empty and the block validation component has idled at
+--   any point after @currentStart@ plus @gracePeriod@, then the peer
+--   @currentPeer@ has failed to promptly serve @inflight(currentPeer)@, and:
+--
+--   - If @currentPeer@ is the ChainSync Jumping dynamo, then it must
+--     immediately be replaced as the dynamo.
+--
+--   - Assume @currentPeer@ will never finish replying to that fetch request and
+--     add all of @inflight(currentPeer)@ to @inflightIgnored@. REVIEW: Nick's
```
facundominguez, line 50: I can't find this piece of the implementation.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,613 @@
...
+          candidatesAndPeers
+
+    -- Step 2: Filter out from the chosen candidate fragment the blocks that
+    -- have already been downloaded, or that have a request in flight (except
+    -- for the requests in flight that are ignored). NOTE: if not declined,
+    -- @theFragments@ is guaranteed to be non-empty.
+    let (theFragments :: FetchDecision (CandidateFragments header)) =
+          pure theCandidate
+            >>= filterNotAlreadyFetched fetchedBlocks fetchedMaxSlotNo
+            >>= filterNotAlreadyInFlightWithAnyPeerNonIgnored candidatesAndPeers
```
facundominguez, line 357: It aches me to read these names. I'd find more natural to write:
```haskell
  >>= dropAlreadyFetched fetchedBlocks fetchedMaxSlotNo
￼ >>= dropAlreadyInFlightWithAnyNonIgnoredPeer candidatesAndPeers
```

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,613 @@
...
+                { peersOrderCurrent = peersOrderCurrent',
+                  peersOrderOthers = peersOrderOthers',
+                  peersOrderStart
+                }
+
+      checkLastChainSelStarvation :: PeersOrder peer -> m (PeersOrder peer)
+      checkLastChainSelStarvation
+        peersOrder@PeersOrder {peersOrderCurrent, peersOrderStart, peersOrderOthers} = do
+          lastStarvationTime <- case chainSelStarvation of
+            ChainSelStarvationEndedAt time -> pure time
```
facundominguez, line 259: I'm not finding any place in the repos where `ChainSelStarvationEndedAt` is created. Is this value produced ever?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
facundominguez: I have found it now.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
facundominguez, line 340: Here's something else that puzzles me. I think the following claims are true:
1. All the in-flight requests of the current peer aren't ignored. The implementation does avoid rerequesting these in-flights.
2. The inflights of all other peers are ignored, as they have been rotated out of the best position in the `peersOrder`.

Therefore, first question: Why are we marking the blocks in inflight requests with a boolean in `PeerFetchBlockInFlight`? We know whether they are ignored or not by looking at which peer is at the front of the `peersOrder`.

Second question: Why bother dropping the inflight requests when preparing to select the peer? Surely we want to select a peer that can offer all the blocks that haven't been downloaded yet.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
Niols: It should be created by `getChainSelMessage` in `ouroboros-consensus`. See the [current state of this function](https://github.com/IntersectMBO/ouroboros-consensus/blob/0c8ef3713343960f2e91b826464be69346d932b0/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Storage/ChainDB/Impl/Types.hs#L519-L548)

Niols: Note: at the point where you said this, as we discussed on Slack, it was indeed missing. I think the new version is correct, but it doesn't look like it from the logs, so please have a close look at it!

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
Niols: I agree very much. The second one has been removed entirely after our discussion in the sync. The first one pre-dates our effort and that is why I haven't renamed it. I can, however, while we're at it.

Niols: I think you're right and we could (should) definitely get rid of the boolean tracking ignored blocks. This would involve changing the way we track the current peer. Currently, it is by definition the only one that has non-ignored in-flight requests. However, we could just add it to a TVar somewhere, erase it on starvations, and say that the current peer is that one _if_ it has requests in flights, and otherwise we may elect a new one. I think this would be much cleaner, and it would drastically simplify the specification by removing this notion of ignored vs. non-ignored in-flight blocks.

Niols: @nfrisby> Are we missing something here?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
Niols: I think this is a question for @nfrisby.

Niols:
Note that, with Friday's changes, the text description is not up-to-date anymore with what the code actually does.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,591 @@
...
+--   ahead of jumpers, has therefore more chances to be selected as the current
+--   peer. It is still possible for a jumper or a disengaged peer to be
+--   selected.
+--
+-- - If the current peer is the CSJ dynamo, but it is a dishonest peer serving
+--   headers fast but retaining blocks, it might be able to drastically leash
+--   us, because its ChainSync client will be stuck behind the forecast horizon
+--   (and therefore not subject to ChainSync punishments such as the Limit on
+--   Patience). This is why we need to consider starvation of ChainSel and
+--   demote peers that let us starve.
```
facundominguez, line 106: I don't understand this bit yet. If the dynamo is blocked on the forecast horizon, BlockFetch will still download blocks from some other peer because ChainSel will starve just the same.

What need is there to rotate the dynamos?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,605 @@
...
+-- Terminate this iteration.
+--
+-- About ignored in-flight requests
+-- --------------------------------
+--
+-- One can note that in-flight requests are ignored when finding a new peer, but
+-- considered when crafting the actual request to a chosen peer. This is by
+-- design. The goal of this algorithm is to keep talking to the same peer unless
+-- it proves to be too weak; in that case, @inflight(p)@ will be empty for all
+-- @p /= currentPeer@.
```
facundominguez, line 84:
```suggestion
-- About the influence of in-flight requests
-- -----------------------------------------
--
-- In-flight requests are ignored when finding a new peer, but
-- considered when crafting the actual request to a chosen peer. We explain the
-- rationale here.
```

I couldn't make sense last sentence of the paragraph. It looks to me like non-current peers can have in-flights regardless of what the current peer is doing. It also looks like the narrative stands without that consideration.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
```diff
@@ -0,0 +1,605 @@
...
+--
+-- - Let @inflight :: peer -> Set blk@ be the outstanding blocks, those that
+--   have been requested and are expected to arrive but have not yet.
+--
+-- - Let @peersOrder@ be an order of preference among the peers. This order is
+--   not set in stone and will evolve as we go.
+--
+-- - Let @currentPeer :: Maybe peer@ be the “current peer” with which we are
+--   interacting. If it exists, this peer must be the best according to
+--   @peersOrder@, and the last fetch request must have been sent to them.
```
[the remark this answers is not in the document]
Niols: Actually, now that I see your comment, I don't think it is necessary for `currentPeer` to be the best according to `peersOrder`, since we will not look at `peersOrder` until this peer has starved us or we have no in-flight request with this peer anymore, probably because it doesn't have any blocks for us anymore.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
Niols: I think @nfrisby is better equipped to answer this.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
Niols: Made irrelevant by recent changes.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/BulkSync.hs
Niols: We got rid of the notion of ignored in-flight requests.

dnadales:
Closing this as stale.

PR #4909, opened 2024-07-10
Small improvements and bug fix in the Signal API

bolt12:
# Description

Cleanup the Signal API, filling it the missing functions and fixing a subtle bug in `keyedTimeout` and `keyedLinger` function


crocodile-dentist (approved):
LGTM

PR #4916, opened 2024-07-18
Error cardano-ping whenever there's a decoding error

bolt12:
# Description

Fixes https://github.com/IntersectMBO/ouroboros-network/issues/4865


coot (changes requested):
Let's define an error type which provides `displayError` in its `Exception` instance, and use it with `throwIO`.

bolt12:
The problem is that the error messages depend on an `IO String` result

crocodile-dentist (approved):
LGTM

coot (approved):
LGTM

PR #4920, opened 2024-08-02
Switch to `input-output-hk/actions/haskell`

jasagredo:
`actions/haskell` is more recently updated than `setup-haskell`. This solves the issues that were happening on CI, see for example https://github.com/IntersectMBO/ouroboros-network/actions/runs/10203312773/job/28229368660


crocodile-dentist (approved):
LGTM

PR #4921, opened 2024-08-05
Renamed default peer selection target values in config module

crocodile-dentist:
# Description

Change maintains consistency with related code.

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Diffusion/Node.hs
```diff
@@ -404,6 +405,8 @@ run blockGeneratorArgs limits ni na tracersExtra tracerBlockFetch =
       , Diff.P2P.daBulkChurnInterval      = 300
       , Diff.P2P.daReadLedgerPeerSnapshot = pure Nothing -- ^ tested independently
       , Diff.P2P.daConsensusMode          = aConsensusMode na
+      , Diff.P2P.daMinBigLedgerPeersForTrustedState
+          = MinBigLedgerPeersForTrustedState 0 -- ^ todo: fix
```
bolt12, line 409: Should this be fixed?

bolt12 (approved):
LGTM!

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Diffusion/Node.hs
crocodile-dentist: What I meant is to actually generate arbitrary but valid values and test this properly -  I should add a separate issue I think.

ouroboros-network/src/Ouroboros/Network/Diffusion/Configuration.hs
```diff
@@ -87,14 +95,28 @@ defaultPraosTargets =
...
+--   Exiting syncing / entering deadline mode is predicated on this
+--   condition. This should be below `targetNumberOfActiveBigLedgerPeers`
+--   in `syncTargets` otherwise untrusted state will never be departed.
+--   This value is lower than the target, because in Genesis we may
+--   demote a big ledger peer for underperformance and not immediately
+--   promote one from the warm set in case there are adversaries
+--   whom are intentionally trying to slow us down.
+--
+defaultMinBigLedgerPeersForTrustedState :: MinBigLedgerPeersForTrustedState
+defaultMinBigLedgerPeersForTrustedState = MinBigLedgerPeersForTrustedState 5
```
coot, line 118: > This should be below `targetNumberOfActiveBigLedgerPeers`
> --   in `syncTargets` otherwise untrusted state will never be departed.


Then we should also check this condition in `sanePeerSelectionTargets`.

coot: I would also add a comment in the `targetNumberOfActiveBigLedgerPeers` in `defaulSyncTargets`.

ouroboros-network/src/Ouroboros/Network/Diffusion/Configuration.hs
```diff
@@ -87,14 +95,28 @@ defaultPraosTargets =
...
-defaultGenesisSyncTargets :: PeerSelectionTargets
-defaultGenesisSyncTargets =
-  defaultPraosTargets {
+defaultSyncTargets :: PeerSelectionTargets
+defaultSyncTargets =
+  defaultDeadlineTargets {
     targetNumberOfActivePeers               = 0,
     targetNumberOfKnownBigLedgerPeers       = 100,
     targetNumberOfEstablishedBigLedgerPeers = 50,
     targetNumberOfActiveBigLedgerPeers      = 30 }
```
coot, line 104: We should reduce the other targets.  Every hot outbound connection is costly both in terms of memory and CPU.  In `defaultDeadlingTargets` we have `20` active peers all together, here we have `45`.  I'd advocate for `targetNumberOfActivePeers = 0`, which will be a 50% increase in active peers where all of them are big ledger peers.

karknu: The big ledger peers target is separate from normal peers so the default config has 25 active peers.

The cost of hot peers is due to chainsync, and consensus is (has?) addressing it with "chain-skip".
In the discussions I've been a part of consensus have said that they need 50 hot connections to big ledger peers to maintain security guarantees. We should test the performance with 50 hot peers to see if "chain-skip" improves things enough.

ouroboros-network/src/Ouroboros/Network/Diffusion/Configuration.hs
crocodile-dentist: The fix is to write tests, it is not broken per se.

PR #4926, opened 2024-08-08
Expose transaction size to tx submission

crocodile-dentist:
# Description

A callback, exposing CBOR-encoded transaction size as it is when transmitted over the network, is provided to 
tx submission inbound and outbound handlers. The size returned by this function does not take into account some additional wrapping that happens when a transaction is sent across the network. Namely, it does not include header data for CBOR-in-CBOR encoding (a tag + it's value to indicate a blob payload, and an encodeBytes tag ~ so roughly 3 words) neither the top level header which consists of EncodeListLen tag + it's value, era word tag + era index word, so ~4 words).

IntersectMBO/cardano-ledger#4521
IntersectMBO/ouroboros-consensus#1211

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
+++ b/ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
@@ -44,6 +44,7 @@ import Network.TypedProtocol.Pipelined (N, Nat (..), natToInt)
 import Ouroboros.Network.NodeToNode.Version (NodeToNodeVersion)
 import Ouroboros.Network.Protocol.Limits
 import Ouroboros.Network.Protocol.TxSubmission2.Server
+import Ouroboros.Network.SizeInBytes
 import Ouroboros.Network.Protocol.TxSubmission2.Type
 import Ouroboros.Network.TxSubmission.Inbound.Types (ProcessedTxCount (..),
            TraceTxSubmissionInbound (..), TxSubmissionMempoolWriter (..),
@@ -135,9 +136,10 @@ txSubmissionInbound
   -> NumTxIdsToAck  -- ^ Maximum number of unacknowledged txids allowed
   -> TxSubmissionMempoolReader txid tx idx m
   -> TxSubmissionMempoolWriter txid tx idx m
+  -> (tx -> SizeInBytes) -- ^ get size of CBOR encoded transaction
   -> NodeToNodeVersion
   -> TxSubmissionServerPipelined txid tx m ()
-txSubmissionInbound tracer (NumTxIdsToAck maxUnacked) mpReader mpWriter _version =
+txSubmissionInbound tracer (NumTxIdsToAck maxUnacked) mpReader mpWriter txSize _version =
     TxSubmissionServerPipelined $ do
 #ifdef TXSUBMISSION_DELAY
       -- make the client linger before asking for tx's and expending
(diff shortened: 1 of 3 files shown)


ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
```diff
@@ -80,10 +81,11 @@ txSubmissionOutbound
   => Tracer m (TraceTxSubmissionOutbound txid tx)
   -> NumTxIdsToAck  -- ^ Maximum number of unacknowledged txids allowed
   -> TxSubmissionMempoolReader txid tx idx m
+  -> (tx -> SizeInBytes) -- ^ get size of CBOR encoded transaction
   -> NodeToNodeVersion
   -> ControlMessageSTM m
   -> TxSubmissionClient txid tx m ()
-txSubmissionOutbound tracer maxUnacked TxSubmissionMempoolReader{..} _version controlMessageSTM =
+txSubmissionOutbound tracer maxUnacked TxSubmissionMempoolReader{..} _txSize _version controlMessageSTM =
```
coot, line 88: Does the outbound side need the callback, or can we use the sizes from the mempool?

Maybe we should change the mempool interface, to give us just `[txid]`s rather than `[(txid, SizeInBytes)]` - so we don't have multiple ways of getting that information, possibly in a non-compatible way.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -181,9 +182,10 @@ txSubmissionInbound
   -> NumTxIdsToAck  -- ^ Maximum number of unacknowledged txids allowed
   -> TxSubmissionMempoolReader txid tx idx m
   -> TxSubmissionMempoolWriter txid tx idx m
+  -> (tx -> SizeInBytes) -- ^ get size of CBOR encoded transaction
   -> NodeToNodeVersion
   -> TxSubmissionServerPipelined txid tx m ()
-txSubmissionInbound tracer (NumTxIdsToAck maxUnacked) mpReader mpWriter _version =
+txSubmissionInbound tracer (NumTxIdsToAck maxUnacked) mpReader mpWriter _txSize _version =
```
coot, line 188: Can we verify the tx sizes using an `assert`-ion? I'd expect that production nodes run without assertions, and this would give us a chance to verify that this works well.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
crocodile-dentist: Essentially, the callback does provide the size of the transaction as it is stored in the mempool, except for the hash because it is not part of the transaction when crossing the network. I think your second remark is accurate.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
crocodile-dentist: Good idea - @karknu commented on slack I think that this maybe should be in an #ifdef.

coot: I don't think we need CPP for it, `assert`-tions are only enforced if one compiled the node with `-fno-ignore-asserts` flag.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Outbound.hs
crocodile-dentist: I've removed the callback from the outbound side. We can keep the interface as-is for now, and if/when the annotated codec gets integrated we can remove `SizeInBytes` from the mempool API.

PR #4932, opened 2024-08-19
Keep the BulkSync implementation when adding the new Genesis fetch mode

facundominguez:


ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Genesis.hs
```diff
@@ -0,0 +1,589 @@
...
+        grossRequest = FetchRequest $ assert (all (not . AF.null) requestBlock) requestBlock
+
+    -- Return the first peer that can serve the gross request and decline
+    -- the other peers.
+    go grossRequest candidates
+  where
+    go grossRequest (c@(candidate, peerInfo) : xs) = do
+      if requestHeadInCandidate candidate grossRequest then do
+        tell $ DList.fromList
+          [(FetchDeclineConcurrencyLimit FetchModeBulkSync 1, pInfo)
```
facundominguez, line 493: Should we use `FetchModeGenesis` instead? This would require some changes in `Decision.hs` to accomodate for the type `GenesisFetchMode`.

ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
```diff
@@ -1,24 +1,35 @@
...
 import GHC.Stack (HasCallStack)
+import NoThunks.Class (NoThunks)
 
 import Ouroboros.Network.AnchoredFragment (AnchoredFragment)
 import Ouroboros.Network.Block
 import Ouroboros.Network.SizeInBytes (SizeInBytes)
 
 
+-- | The fetch mode that the block fetch logic should use.
+data GenesisFetchMode = FetchModeGenesis | PraosFetchMode FetchMode
```
amesgen, line 30: It seems a bit confusing that `PraosFetchMode` is of type `GenesisFetchMode`. One idea would be to have rename stuff like this:
```haskell
data FetchMode = FetchModeGenesis | PraosFetchMode PraosFetchMode
data PraosFetchMode = FetchModeDeadline | FetchModeBulkSynk
```
But this would cause a larger diff of course.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Genesis.hs
amesgen: It indeed seems nicer to use `FetchModeGenesis` here, I think very small changes to `Decision.hs` (especially if they are in a separate commit) are fine.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Genesis.hs
```diff
@@ -0,0 +1,589 @@
...
+          fragments <- dropAlreadyInFlightWithPeer inflight theFragments
+
+          -- Trim the fragments to the peer's candidate, keeping only blocks that
+          -- they may actually serve.
+          trimmedFragments <- trimFragmentsToCandidate thePeerCandidate (snd fragments)
+
+          -- Try to create a request for those fragments.
+          fetchRequestDecision
+            fetchDecisionPolicy
+            FetchModeBulkSync
```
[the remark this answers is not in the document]
amesgen: Yes, this seems ok to me, it seems completely fine for the Praos BulkSync and the Genesis BlockFetch logic to share functionality :+1:

PR #4934, opened 2024-08-21
WIP: Annotated decoder

coot:
This is an experiment to provide `runAnnotatedPeer`, which is like `runPeer' but allows us to run a decoder which has access to bytes used
when decoding a message.   This allows one to record ByteString from
which a piece of data was decoded, e.g. for each `tx` inside
`MsgReplyTxs`.

The `Codec` type in `typed-protocols` was generalised for this purpose. The core functionality is implemented in
`runAnnotatedDecoderWithChannel` which runs `AnnotatedCodec` against a `Channel` which does incremental decoding & recording bytes used so far.  We also expose `runAnnotatedPeer` which runs a `Peer` against `Channel` using an `AnnotatedCodec` (using `annotatedDriverSimple`).

TODO:

* `runAnnotatedPeerWithLimits`
* `runAnnotatedPipelinedPeerWithLimits`

It's actually the last one that we will need in `tx-submission`.

* Add
```
data WithBytes a {
  encoded :: ByteString,
  decoded :: a
```
  and generalise `codecTxSubmission2` so that it can be used to used with annotator and without it - it might require two separate function, but I think it can be generated from one more general function (so we don't need to maintain two codecs).

TODO: design & implement quickcheck properties

# Description

_reasonably detailed description of the pull request_

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


ouroboros-network-framework/src/Ouroboros/Network/Driver/Limits.hs
```diff
@@ -219,6 +314,64 @@ runDecoderWithLimit limit size Channel{recv} =
...
+    go !_ !_ !_ (DecodeFail failure) = return (Left (Just failure))
+
+    go !_ !sz !_ DecodePartial {} | sz > limit =
+      return (Left Nothing)
+
+    go !bytes !sz Nothing (DecodePartial k) = do
+      mbs <- recv
+      let !sz' = sz + maybe 0 size mbs
+      step <- k mbs
+      go (bytes <> fromMaybe mempty mbs) sz' Nothing step
```
crocodile-dentist: The `<>` is problematic, here is it's definition from bytestring:
```
append :: ByteString -> ByteString -> ByteString
append (BS _   0)    b                  = b
append a             (BS _   0)    = a
append (BS fp1 len1) (BS fp2 len2) =
    unsafeCreateFp (checkedAdd "append" len1 len2) $ \destptr1 -> do
      let destptr2 = destptr1 `plusForeignPtr` len1
      memcpyFp destptr1 fp1 len1
      memcpyFp destptr2 fp2 len2
```
So this has O(n^2) time complexity, I think. Perhaps `runAnnotatedDecoderWithLimit` should take an additional parameter with type `bytes -> Builder`, here we just concatenate the builder, and at the end in the `DecodeDone` case match we execute the builder once to get the final bytes?

ouroboros-network-framework/src/Ouroboros/Network/Driver/Limits.hs
```diff
@@ -219,6 +314,64 @@ runDecoderWithLimit limit size Channel{recv} =
...
+
+    go !bytes !sz (Just trailing) (DecodePartial k) = do
+      let sz' = sz + size trailing
+      step <- k (Just trailing)
+      go (bytes <> trailing) sz' Nothing step
```
crocodile-dentist: same as above

ouroboros-network-framework/src/Ouroboros/Network/Driver/Simple.hs
```diff
@@ -209,15 +305,34 @@ runPipelinedPeer tracer codec channel peer =
...
+  -> DecodeStep bytes failure m (bytes -> a)
+  -> m (Either failure (a, Maybe bytes))
+
+runAnnotatedDecoderWithChannel Channel{recv} bs0 = go (fromMaybe mempty bs0) bs0
+  where
+    go :: bytes -> Maybe bytes -> DecodeStep bytes failure m (bytes -> a) -> m (Either failure (a, Maybe bytes))
+    go bytes  _ (DecodeDone f trailing)        = return $ Right (f bytes, trailing)
+    go _bytes _ (DecodeFail failure)           = return (Left failure)
+    go bytes Nothing         (DecodePartial k) = recv >>= \bs -> k bs >>= go (bytes <> fromMaybe mempty bs) Nothing
+    go bytes (Just trailing) (DecodePartial k) = k (Just trailing) >>= go (bytes <> trailing) Nothing
```
crocodile-dentist: same as above

crocodile-dentist (commented):
One remark on efficiency

ouroboros-network-framework/src/Ouroboros/Network/Driver/Limits.hs
coot: The one you quoted is for strict bytestrings.  The one for lazy bytestring is `foldrChunks Chunks`: https://hackage-content.haskell.org/package/bytestring-0.12.2.0/docs/src/Data.ByteString.Lazy.Internal.html#append.

Both are expensive when in the number of chunks, e.g. $a_1 <> a_2 <> ... <> a_n$ then associativity matters, we do the $((a_1 <> a_2) <> ... ) <> a_n$ which is even worse than the other one.

I don't want to overcomplicate the API with a `Builder`, but we can use `mconcat`, which is linear in the number of chunks.

crocodile-dentist (approved):

coot:
I'll merge it once I'll release `typed-protocols-1.1` to `Hackage`.

PR #4951, opened 2024-09-06
io-classes-1.8

coot:
* [x] `forAllProjectPackages` needs to be fixed
* [x] nix infinite recusion error:


jorisdral:
Is there anything in particular that is holding this PR back from being merged? I'd like to use a library in `ouroboros-consensus` that depends on `io-classes-1.7`, but it means that `ouroboros-network` and `typed-protocols` have to be updated

coot:
@jorisdral I updated PR description.

coot:
There's nothing blocking us from merging it right now.  It needs to be reviewed, which I hope will happen in this sprint (which we just started).

geo2a:
I'd greatly appreciate if this PR could be included into the upcoming release of `ouroboros-network`.

crocodile-dentist (approved):

geo2a:
Should we remove the `ci/hydra-build:x86_64-darwin` job from the required checks? In Consensus, we are also considering doing this, as it times out consistently.

PR #4952, opened 2024-09-07
Decode NodeToClientV17

coot:


karknu (changes requested):
You need to add `NodeToClientVersionV17` to `supportedNodeToClientVersions` otherwise it will never be negotiated.

coot:
Thanks, fixed.

karknu (approved):
LGTM

PR #4956, opened 2024-09-10
Fix test failure

### LINKED ISSUE #4942: Test failure: node never uses non-trustable peers in clean state
CI test failure:
```
sim-tests --quickcheck-replay="(SMGen 1055325783478325567 18154453325318164197,92)" -p '/node never uses non-trustable peers in clean state/'
```

crocodile-dentist:
# Description

A problem with test logic was identitied when 'targetNumberOfKnownPeers' is decreased causing a trusted local root peer to be dropped immediately.
This peer lingers in the known peers set until another action in the governor sweeps it out, but for a brief while a failure is signaled when known peers /= trusted peers.
The fix adds trusted local root peers from initial configuration into the enriched known set such that this glitch doesn't get detected.

resolves #4942 
# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


coot (approved):

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
```diff
@@ -3483,11 +3494,15 @@ prop_governor_only_bootstrap_peers_in_clean_state env =
              )
 
      in signalProperty 20 show
-          (\(b, (kp, tp), t) -> (b && (Set.null (Set.difference kp tp) || targetNumberOfKnownPeers t <= 0))
-                             || not b)
-          ((,,) <$> isInCleanState
-                <*> govKnownAndTrustedPeers
-                <*> govTargets
+          (\(b, (kp, tp), fromConfigTrustedLocalRoots, t) ->
```
coot, line 3497:
```suggestion
             -- adjust with configured local roots see
             -- https://www.github.com/IntersectMBO/ouroboros-network/pulls/4956
             let kp' = kp `Set.union` fromConfigTrustedLocalRoots
```
or describe it (as you did in the PR).

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
```diff
@@ -3453,6 +3453,17 @@ prop_governor_only_bootstrap_peers_in_clean_state env =
...
+        configTrustedLocalRoots :: Signal (Set PeerAddr)
+        configTrustedLocalRoots =
+            Signal.fromChangeEvents Set.empty
+          . Signal.selectEvents
+              (\case TraceEnvSetLocalRoots (LocalRootPeers peerMap _) ->
+                       Just . Map.keysSet . Map.filter (\case (_, IsTrustable) -> True
+                                                              _ -> False) $ peerMap
+                     _ -> Nothing)
+          . selectEnvEvents
+          $ events
```
bolt12, line 3465: Could you better align this, this was super hard for me to read 😆

ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/PeerSelection.hs
crocodile-dentist: I've added this note:
             -- the governor logic has two separate monitoring actions for dealing with local roots
             -- and keeping (known) peer targets in check. A corner case can arise when peer targets
             -- are reduced and a trusted local root peer is selected to be dropped immediately. A trace
             -- will show that this trusted local root is no longer among the localRoots field in the
             -- governor state, but until the second monitoring action mentioned above sweeps it out
             -- from the knownPeers set, a discrepancy occurs which can cause a naive version of this test
             -- to signal a failure. In reality, this is just a glitch and the governor doesn't leave
             -- clean state because semantically this peer, which is still among the knownPeers, is
             -- trusted by assumption. The fix is to permanently add all trusted local roots to the known
             -- set via union below since this doesn't brake the invariant. On the other hand,
             -- if some other peer which is not from the known trusted local root set somehow lingers
             -- in the known set, the test will fail, as it should.
             -- cf. https://github.com/IntersectMBO/ouroboros-network/pull/4956

PR #4963, opened 2024-09-22
check-stylish: better support for -c & -u switches

coot:
* Use `fd` & ignore file list, in the same way when no option is passed.
* Reviewed `CODEOWNERS` file.


crocodile-dentist (approved):
lgtm

PR #4975, opened 2024-10-01
Fix churn target

crocodile-dentist:
# Description

This change rolls back one hunk from the previous fix which was erroneous.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
+++ b/ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
@@ -268,6 +268,10 @@ peerChurnGovernor PeerChurnArgs {
           case regime of
             ChurnBootstrapPraosSync -> min (targetNumberOfActivePeers targets + 1)
                                            (targetNumberOfEstablishedPeers base)
+            -- ^ In this mode, we are only connected to a handful of bootstrap peers.
+            -- The original churn strategy was to increase the targets by small
+            -- fixed amount (e.g. 1). Therefore, we use
+            -- targets to calculate the upper bound, ie. active + 1 here.
             _otherwise -> targetNumberOfEstablishedPeers base }
 
     checkEstablishedPeersIncreased :: CheckPeerSelectionCounters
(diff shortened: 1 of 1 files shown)


ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -305,7 +305,7 @@ peerChurnGovernor PeerChurnArgs {
       targets {
         targetNumberOfEstablishedPeers =
           case regime of
-            ChurnBootstrapPraosSync -> min (targetNumberOfActivePeers base)
+            ChurnBootstrapPraosSync -> min (targetNumberOfActivePeers targets)
```
coot, line 308: Could you add a comment why `targets` should be used here.

coot (approved):
LGTM

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
crocodile-dentist: I didn't give it too much thought but realized that this is what we do in increaseEstablishedPeers as well even since before Genesis patches were applied. I suppose the only benefit would be if some re-ordering of churn actions was made, eg. we had decreaseActive -> decreaseEstablished -> increaseEstablished -> increaseActive instead of the current decreaseActive -> increaseActive -> decreaseEstablished -> increaseEstablished. In the former case, having the `targets` argument in the patch would perhaps provide the correct upper bound? But this alternate sequence would not be desirable because we would have the node running with lower number of active peers for a little longer than absolutely necessary. Other than that, this change was about going back to how it was before. So maybe it is increaseEstablished that should be changed to reflect decreaseEstablished when we use the current sequence of churn actions?

bolt12 (approved):

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
coot: In the `ChurnBootstrapPraosSync` case we are only connected to a handful of bootstrap peers.  The original churn strategy was to decrease the targets by small fixed amount (e.g. `1`) and then increase it back.  I think it's the right thing to do; it would just be good to document it in both places so we don't make the same mistake again.

crocodile-dentist: Done.

PR #4984, opened 2024-10-04
Let the correct thread add TXs to the mempool

karknu:
# Description

Previously all peers that had the TXid of a downloaded TX would attempt to add it to the mempool (a rather costly operation).

This patch moves the addition to the mempool to the Server's CollectTxs state. Which means that the downloaded TXs is only added by the thread that actually downloaded it.

The legacy TX submission protocol has been left as is.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


crocodile-dentist (approved):
lgtm

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound.hs
```diff
@@ -314,13 +315,18 @@ txSubmissionInbound tracer (NumTxIdsToAck maxUnacked) mpReader mpWriter _version
...
+        !end <- getMonotonicTime
+        let duration = diffTime end start
+        traceWith tracer $
+          TraceTxInboundAddedToMempool txidsAccepted duration
         let !accepted = length txidsAccepted
 
         traceWith tracer $ TraceTxSubmissionProcessed ProcessedTxCount {
             ptxcAccepted = accepted
           , ptxcRejected = collected - accepted
+          , ptxcScore    = 0 -- This implementatin does not track score
```
coot, line 329: Should it be a `TODO`?

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/Registry.hs
```diff
@@ -210,8 +221,43 @@ withPeer tracer
...
+      st <- readTVar sharedStateVar
+      case Map.lookup peeraddr (peerTxStates st)  of
+           Nothing -> error "missing peer updated"
+           Just ps -> return $ rejectedTxs ps
+       where
+        cntRejects :: SharedTxState peeraddr txid tx
+                   -> SharedTxState peeraddr txid tx
+        cntRejects st@SharedTxState { peerTxStates } =
+          let peerTxStates' = Map.update (\ps -> Just $! ps { rejectedTxs = min 42 (max (-42) (rejectedTxs ps + n)) }) peeraddr peerTxStates in
+          st {peerTxStates    = peerTxStates'}
```
coot, line 247: It will be more efficient to update it for all peers at once, but it might be harder to implement it. Another way would be to store the counts in `Map peerAddt (TVar ...)`, so that we don't block on taking the whole state to update the count just for one peer.

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/Server.hs
```diff
@@ -70,18 +79,53 @@ txSubmissionInboundV2
...
-
-        traceWith tracer $ TraceTxSubmissionProcessed ProcessedTxCount {
-            ptxcAccepted = accepted
-          , ptxcRejected = collected - accepted
-          }
+        mpSnapshot <- atomically mempoolGetSnapshot
+        let receivedL = [ (txId tx, tx) | tx <- txs ]
+        fetchedSet <- consumeFetchedTxs (Set.fromList (map fst receivedL))
+
+        -- Only attempt to add TXs if we actually has fetched some.
```
coot, line 88:
```suggestion
        -- Only attempt to add TXs if we actually have fetched some.
```

ouroboros-network/src/Ouroboros/Network/TxSubmission/Inbound/Server.hs
```diff
@@ -70,18 +79,53 @@ txSubmissionInboundV2
...
+        when (not $ Set.null fetchedSet) $ do
+            let fetched = filter 
+                            (\(txid, _) -> Set.member txid fetchedSet)
+                            receivedL
+                fetchedS = Set.fromList $ map fst fetched
+
+            -- Note that checking if the mempool contains a TX before
+            -- spending several ms attempting to add it to the pool has
+            -- been judged immoral.
```
coot, line 97: :grin:

PR #4992, opened 2024-10-17
Bump versions for ouroboros-network-0.18 release

coot:


coot:
Good point, I'll add it to `cardano-ping` too.

PR #4995, opened 2024-10-21
github-actions: fixed caching

coot:
Fixed caching of cabal store: https://github.com/IntersectMBO/ouroboros-network/actions/runs/11435190378/job/31810104749

* more robust way of cache key management
* removed `enableCrossArchive` (it doesn't seem to be necessary now)
* only save the cache when necessary


crocodile-dentist (approved):
lgtm

PR #5050, opened 2025-01-24
Label some threads

jasagredo:
# Description

I believe these thread labels are missing.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
+++ b/ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
@@ -38,7 +38,6 @@ import Control.Monad.Class.MonadTimer.SI
 import Control.Tracer (Tracer, contramap, traceWith)
 
 import Data.ByteString.Lazy (ByteString)
-import Data.List as List (foldl')
 import Data.List.NonEmpty (NonEmpty)
 import Data.List.NonEmpty qualified as NonEmpty
 import Data.Void (Void, absurd)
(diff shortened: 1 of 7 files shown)


ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
```diff
@@ -175,6 +177,7 @@ with Arguments {
                           ]
           -- race all `acceptLoops` with `actionThread` and
           -- `inboundGovernorThread`
+          -- TODO is this right?
```
jasagredo, line 180: This seems to be racing the new action vs the accumulated action, but for that it spawns a new thread that runs the accumulated action. Isn't here an unneeded intermediate thread? Say I have
the list `[a, b]` and the initial element `e`, then this will end up in:
```
(e `race` a) `race` b
```

which will have:
- a racing thread for `b`
- a racing thread for `e race a`
  - a racing thread for `e`
  - a racing thread for `a`

No?

jasagredo: We can either try to solve this here or if you prefer leave the `TODO` for later investigation?

 
cc @bolt12

bolt12: Lets leave it, and do it in a different PR

coot: Can I do that?

coot (changes requested):
Thanks, @jasagredo.

ouroboros-network-framework/src/Ouroboros/Network/Server2.hs
jasagredo: In fact this intermediate thread is the one that is unlabelled! This was a nightmare to find.

coot (approved):
LGTM, thank you @jasagredo.

jasagredo:
Let's see if we can make Hydra re-run those timing-out jobs.

PR #5056, opened 2025-01-29
Add cardano-blueprint submodule and move handshake cddl to it

ch1bo:
# Description

This demonstrates how we could move the CDDL to a common repository and tested against from multiple implementations of the ouroboros protocols.

Git submodules are obviously only one way of managing this dependency.

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


ch1bo:
@coot I'd love to use this PR as a basis for discussion on how we could move the CDDLs into a place where they are more easily consumed by other implementations.

coot:
@ch1bo, such an approach will work, but first, this should be discussed in a wider forum, e.g. with consensus & ledger teams.  We should not corner them with a decision like that.

I haven't used git-submodules extensively, but enough to know it can be painful.  Are there other possible approaches which avoid git-submodules?

Can one use tags instead of git-hashes?  It would be much nicer to say `node-to-node-v14`, isn't it?

PR #5059, opened 2025-01-30
Fix testnet handshake

bolt12:


crocodile-dentist (approved):
lgtm

karknu (approved):
LGTM.

Do you still see the tracing errors due to other bugs?

bolt12:
@karknu  I didn't check, how can I do that quickly?

karknu:
> @karknu I didn't check, how can I do that quickly?

Modify a testcase to fail, then just examine the resulting tracer. That is how I noticed it.

Proper way would be to have a testcase that fails if there are any `UnknownMiniProtocol` errors.

karknu:
Don't merge yet. Test case will sometime fails so there are other issues, for sure in the real code and possibly with the Testnet environment.

coot:
Closing in favour of #5062. Thanks @bolt12 & @karknu.

PR #5070, opened 2025-02-06
Diffusion improvements

coot:
# Description

- **diffusion: removed daExtraArgs from ArgumentsExtra**
- **diffusion: removed extraActions**

There will be more...

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


bolt12 (commented):
I don't think my PR is leaking network details. It is setting up a different API. Currently, `cardano-node` is only one possible client of our diffusion so we must not entirely tailor our API to it, we need to be sensible for others too. For instance, I'd prefer that the extension points were all in the same top level/entry data types and then the diffusion library would make the plumbing itself, also from a type signature perspective it is more clear to me where I should have my extension data types.

Removing `extraActions` just because they can be computed from `extraArgs` _*in the cardano node case*_ doesn't mean it is going to be the same case for all other applications. So we might be simplifying things for cardano-node case and making things difficult for the others...

The same for the `extraAPI` ... for `cardano-node` it is easy to optimise for, but I don't like removing the extension point from the main Consensus API data type `LedgerConsensusInterface` because it makes it harder (imo) to see where one would extend the API.

I don't disagree with these changes but I think they are missing the bigger picture. We can improve our API without making it worse for non-cardano clients.

ouroboros-network/src/Ouroboros/Cardano/Network/PeerSelection/Governor/Types.hs
```diff
@@ -214,11 +212,15 @@ cardanoPeerSelectionGovernorArgs extraActions readUseLedgerPeers peerSharing upd
...
+  , updateWithState = \PeerSelectionInterfaces { readUseLedgerPeers }
+                       PeerSelectionActions { getLedgerStateCtx,
+                                              peerSharing }
+                       psv st -> do
       associationMode <- readAssociationMode readUseLedgerPeers
                                              peerSharing
                                              (Cardano.bootstrapPeersFlag (extraState st))
-      updateOutboundConnectionsState
+      Cardano.updateOutboundConnectionsState
+        (lpExtraAPI getLedgerStateCtx)
```
bolt12, line 223: nice!

bolt12 (approved):
LGTM! Thank you @coot

PR #5075, opened 2025-02-10
Increased short timeout for Churn timeouts

### LINKED ISSUE #5072: Churn no timeouts test failure
Use --quickcheck-replay="(SMGen 6777981991749325316 6764231823468127323,3)" to reproduce.
Use -p '/IOSim.Churn.no timeouts/' to rerun this test only.

bolt12:
Closes #5072

### DIFF
--- a/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
+++ b/ouroboros-network/sim-tests-lib/Test/Ouroboros/Network/Testnet.hs
@@ -157,7 +157,7 @@ tests =
                       (testWithIOSimPOR prop_no_non_trustable_peers_before_caught_up_state 10000)
     , testGroup "Churn"
       [ nightlyTest $ testProperty "no timeouts"
-                        (testWithIOSimPOR prop_churn_notimeouts 10000)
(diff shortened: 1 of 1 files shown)


crocodile-dentist:
The defaultBulkChurnInterval is 15 mins - it appears with these extended 'short' timeouts we could in principle be non-stop in churning mode, and at the same time be very slow at it.

crocodile-dentist:
That is true. However, IIRC the purpose was to wait a reasonable time to get most accurate counters in the logs. Even if the timeout is too short, the targets will of course be eventually met but the relevant traces will not be printed. Maybe these timeouts could be exposed to our mocking framework and we set higher values there to make sure everything works, but keep things snappy for actual use?

bolt12:
Yeah, either way sounds fine, but for the test in question we don't look at the timeout values but at the traces, so the quickest way to fix is this small patch :P

bolt12:
@coot  what do you think?

bolt12:
for the churn test this is the haddock:

> -- | This property verifies that when nodes are running without network
> -- attenuation, decreasing numbers by churn never timeouts.

and indeed it was running with attenuations so it was flaky. I changed it to run with no attenuation and reverted the timeout changee

PR #5081, opened 2025-02-20
peer selection exports

coot:
- **peer-selection: import and export lists**
- **sim-tests-lib: code style**
- **ouroboros-network: renamed sim-test-lib as testlib**

Depends on #5007


ouroboros-network/src/Ouroboros/Network/PeerSelection.hs
```diff
@@ -0,0 +1,42 @@
...
+           newPeerMetric, newPeerMetric', nullMetric, reportMetric)
+import Ouroboros.Network.PeerSelection.PeerSelectionActions as PeerSelection.PeerSelectionActions
+import Ouroboros.Network.PeerSelection.PeerSharing as PeerSelection.PeerSharing
+import Ouroboros.Network.PeerSelection.PeerStateActions as PeerSelection.PeerStateActions
+import Ouroboros.Network.PeerSelection.PublicRootPeers as PeerSelection.PublicRootPeers
+           (PublicRootPeers)
+import Ouroboros.Network.PeerSelection.RelayAccessPoint as PeerSelection.RelayAccessPoint
+           (DomainAccessPoint (..), IP (..), PortNumber, RelayAccessPoint (..))
+import Ouroboros.Network.PeerSelection.Types as PeerSelection
+import Ouroboros.Network.PeerSelection.Types as PeerSelection.Types
```
bolt12, line 42: I don't like this... I prefer for the imports to come from the place where the expression is defined. Meta modules like these are confusing for large codebases

ouroboros-network/src/Ouroboros/Network/PeerSelection/RootPeersDNS.hs
```diff
@@ -1,13 +1,14 @@
...
---
-data PeerActionsDNS peeraddr resolver exception m = PeerActionsDNS {
-  paToPeerAddr :: IP -> PortNumber -> peeraddr,
-  paDnsActions :: DNSActions resolver exception m
-  }
+import Ouroboros.Network.PeerSelection.RootPeersDNS.DNSActions as DNSActions
+import Ouroboros.Network.PeerSelection.RootPeersDNS.DNSSemaphore as DNSSemaphore
+import Ouroboros.Network.PeerSelection.RootPeersDNS.LedgerPeers as LedgerPeers
+import Ouroboros.Network.PeerSelection.RootPeersDNS.LocalRootPeers as LocalRootPeers
+import Ouroboros.Network.PeerSelection.RootPeersDNS.PublicRootPeers as PublicRootPeers
```
bolt12, line 14: Same here. We can move `PeerActionsDNS` to a better module but we should import it from there.

bolt12 (commented):
I am not entirely sold on this import changes. It is nice to clean up imports and simplify them but I think meta modules are a bit of a stretch. I usually import the module that owns the expression definition or data type, I don't usually import from top level modules like these, I think it makes it clearer where things are coming from.

ouroboros-network/src/Ouroboros/Network/PeerSelection.hs
coot: Another solution could be to have explicit imports in the meta modules, but I am not sure if it's worth it in times of `hls` / `ghc-tags-plugin` etc.

bolt12: Right but, I also use HLS to manage automated imports for me and I usually import from the modules which defined the expression/function. So my worry is that we'll have inconsistencies still in the imports. At least I think the name should be more obvious so I know I should import from the `Meta`/`General`/`Imports`/`All` modules

bolt12 (approved):

PR #5104, opened 2025-03-27
Inbound governor performance improvement

crocodile-dentist:
# Description

The change is expected to reduce CPU load and is designed to scale with the number of inbound connections served by a node. In particular, bootstrap relays and Genesis peers are mostly affected, but all relays in general should benefit from these improvements.

Profiling a real loaded node revealed a CPU bottleneck in the inbound governor loop, which essentially single-threads the application. Specifically, the firstPeerTo* `FirstToFinish` actions are CPU intensive in the amount of state checking performed for all active connections, repeatedly for each such action in a linear fashion. Casting aside the STM costs, there was in particular also duplication of filtering of hot/warm responder miniprotocols for each firstToFinish* activity. In terms of the the STM costs, especially the embedded `LastToFinish` computations are expensive because of lack of short-circuiting of STM `retry` semantics. For instance, when `LastToFinish 1 <> 2 <> 3` is attempted, and `1` and `2` retry, but `3` succeeds, then `2` is attempted again, and if it succeeds then `1` is attempted again. From a performance perspective, it would be better to abort that peer's firstToFinish activity on the first retry and go try the next peer, but I'm not sure if that is supported. 

This PR introduces an event based system where each connection's muxer thread can concurrently notify the inbound governor  of interesting peer activity such that it can perform its tasks. The inbound governor creates a tracer which enqueues muxer events from each peer that it knows about into its information channel queue. This queue is processed in batch where all ordering of events is preserved. In particular, muxer start, stop, miniprotocol clean & exception exits are tracked.  Some additional state is also  introduced which tracks the number of hot and non-hot (warm & established) responder miniprotocols such that remote promotions and demotions can be identified efficiently.

The commit history is sequenced in a bottom up fashion, and while showing the 'guts' of how the new event system operates, it masks at first sight how this is folded into the program as a whole and what other reorganizations are needed from the following commits. In particular, a difficulty manifested because currently the CM is instantiated at the front, but the idea behind the proposed change is that the new tracer should be baked into the connection handler by the inbound governor, which should be provided to the CM as the penultimate step before starting the server. When a new outbound or incoming connection is made, the CM forks this augmented connection handler, which then adjoins this new tracer to the mux tracer before calling `Mux.run`. In this manner, the new tracing is only performed for connections served by the inbound handler, or the outbound handler only when a duplex data flow was negotiated. This reorganization is best viewed in the commit of `Diffusion` module refactoring.

Other changes include:
* Fixes and improvements to simulation shrinkers
* Testnet simulation tracing UX improvements
* Refactors of testing code, with a bugfix to `prop_mux_starvation` test
* Updates to network-spec

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [x] Added labels.
* [ x Updated changelog files.


ouroboros-network-framework/src/Ouroboros/Network/ConnectionHandler.hs
```diff
@@ -337,8 +364,25 @@ makeConnectionHandler muxTracer singMuxMode
...
+                                     mux bearer
+                        _notResponder ->
+                              -- If this is InitiatorOnly, or a server where unidirectional flow was negotiated
+                              -- the IG will never be informed of this remote for obvious reasons. We should not pass
+                              -- the IG tracer to mux, and must not in the latter case as the muxer will
+                              -- deadlock itself before it launches any miniprotocols from the command queue.
+                              -- It will be stuck when reaches mature state, forever waiting for the incoming
+                              -- peer handle.
+                          Mx.run (Mx.WithBearer connectionId `contramap` muxTracer)
+                                 mux bearer
```
bolt12, line 385:
```suggestion
                  let tracer = 
                      case inResponderMode of
                        InResponderMode (inboundGovChannelTracer, connectionDataFlow)
                          | Duplex <- connectionDataFlow agreedOptions ->
                              muxTracer <> inboundGovChannelTracer
                        _notResponder ->
															muxTracer
                  Mx.run (Mx.WithBearer connectionId `contramap` tracer)
                         mux bearer
 ```
 
 this looks a bit cleaner. I have omitted the comments

File: ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs [hunk 2]
```diff
@@ -139,35 +156,165 @@ with :: forall (muxMode :: Mux.Mode) socket initiatorCtx peerAddr versionData ve
...
+          -- the problematic case is when the IG tracer registers responder
+          -- miniprotocol launch command before the CM informs the IG of the
+          -- new connection. However unlikely, this will lead to
+          -- incoherent transitions between the two components.
+          -- The muxer blocks itself here to yield right-of-way
+          -- to the CM.
+          check . Map.member peer . connections =<< readTVar stateVar
+          modifyTVar countersVar $ Map.insert peer (ResponderCounters 0 0)
+
+        (startWithID -> Just mid) -> atomically do
```
bolt12, line 224: I don't think you need view patterns here, hence can avoid adding a language extension.

If you do `_ | Just mid <- startWithID ->` I think it works.

Then you can better organise all your other clauses under the same match

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
```diff
@@ -775,44 +778,45 @@ instance Arbitrary DiffusionScript where
...
+                                                  : (JoinNetwork <$> shrinkDelay d)
+                                                 <> [JoinNetwork 0]
+                                        | otherwise = []
+          shrinkCommand (Kill d) | d > 1 = Kill 0 : (Kill <$> shrinkDelay d)
+                                 | otherwise = []
+          shrinkCommand (Reconfigure d lrp) | d > 1 =    Kill 0
+                                                      : (Reconfigure <$> shrinkDelay d
+                                                                     <*> pure lrp)
+                                                     <> [Reconfigure 0 lrp]
+                                            | otherwise = []
```
bolt12, line 819: Why do you add a Kill at the beginning?

ouroboros-network-framework/src/Ouroboros/Network/ConnectionHandler.hs
```diff
@@ -115,6 +118,27 @@ data Handle (muxMode :: Mx.Mode) initiatorCtx responderCtx versionData bytes m a
...
+    -> MuxConnectionHandler Mx.ResponderMode socket initiatorCtx responderCtx peerAddr
+                                             versionNumber versionData bytes m a b
+
+  MkMuxConnectionHandler Mx.InitiatorResponderMode socket initiatorCtx responderCtx peerAddr
+                                                   versionNumber versionData bytes m a b =
+       (versionData -> DataFlow)
+    -> Tracer m (WithBearer (ConnectionId peerAddr) Trace)
+    -> MuxConnectionHandler Mx.InitiatorResponderMode socket initiatorCtx responderCtx
+                                                      peerAddr versionNumber versionData
+                                                      bytes m a b
```
bolt12, line 141: Can't we get away without the type family? What about a GADT?

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -165,6 +165,121 @@ with
       \thread ->
         k thread (readTVarIO var)
   where
+    -- the responder IG info channel tracer is embedded with the mux tracer unconditionally
+    -- by the conn handler on the inbound side and conditionally for 'InitiatorResponderMode'
+    -- on the outbound side only when the latter has negotiated a 'Duplex' data flow - otherwise
+    -- for unidirectional mode the IG is never provided with the peer handle since the inbound
+    -- is inactive.
```
bolt12, line 203: Do I get this right that this code will run every time something is logged, which means we have to be very careful when to log things?

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
crocodile-dentist: We discussed at the mtg, but it's worthwhile to leave this here as well. I thought that it's an easier trace to analyze if a particular node is killed/not running and the test still fails. So if a node is not running, and a `JoinNetwork` command is next, then 'killing it' essentially leaves it still dead and not contributing to the failure if that still turns out to be the outcome. If it's running and reconfigure command is up, let's kill it instead and see if the test still fails - this means that the reconfiguring step is not contributing to the failure either.

crocodile-dentist: Although for the Kill case, the Kill 0 should probably be put at the back of the list.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
crocodile-dentist: We discussed at the mtg, but yes - every time the muxer (not the bearer, which has the same trace type) traces, some extra code will run. The upside is that no change needs to be done to be muxer as the api/implementation is general enough. It's a philosophical question of whether we want to be very explicit or not. This approach took the path of least resistance, at the cost of comprehensibility - but I've left comments to give a hint of what is going on for the reader.

karknu:
> Some initial comments, I am not sure I fully understood the solution but I don't think I like the added complexity. Tracers shouldn't be used in this way. I don't think it is clearer what happens when the tracer is a null one, for example. Implicit behaviours hidden in tracers sounds like a nightmare to debug as well. Did I understand it right?

We use tracers in this way for ranking peers. I think this is what tracers excel at.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionHandler.hs
crocodile-dentist: Hmm, I tried with a GADT and I think the type family is neater. By going with a GADT, it is essentially duplicating the SingMuxMode GADT with constructors like `MkInitiatorConnectionHandler'. Type families allow me to share that singleton.

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
crocodile-dentist: I've reworked this and introduced a Skip tag which works better.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionHandler.hs
crocodile-dentist: The commit on top shows how this looks like with a GADT.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionHandler.hs
```diff
@@ -120,29 +120,22 @@ data Handle (muxMode :: Mx.Mode) initiatorCtx responderCtx versionData bytes m a
...
+  MuxInitiatorConnectionHandler :: MkMuxConnectionHandler
+                                     Mx.InitiatorMode socket initiatorCtx responderCtx
+                                     peerAddr versionNumber versionData bytes m a b
+  MuxResponderConnectionHandler :: (   StrictTVar m (Maybe ResponderCounters)
+                                    -> Tracer m (WithBearer (ConnectionId peerAddr) Trace))
+                                -> MkMuxConnectionHandler
+                                     Mx.ResponderMode socket initiatorCtx responderCtx
+                                     peerAddr versionNumber versionData bytes m a b
+  MuxInitiatorResponderConnectionHandler
+    :: (versionData -> DataFlow)
```
bolt12, line 134: I think this commit looks much better, no need to carry around a singleton when we can do it with pretty idiomatic Haskell! :)

bolt12 (approved):
LGTM just some minor comments

ouroboros-network-framework/sim-tests/Test/Ouroboros/Network/Server/Sim.hs
```diff
@@ -2318,7 +2336,7 @@ multiNodeSim stdGen serverAcc dataFlow defaultBearerInfo
                    acceptedConnLimit events attenuationMap = do
   multiNodeSimTracer stdGen serverAcc dataFlow defaultBearerInfo acceptedConnLimit
                      events attenuationMap dynamicTracer dynamicTracer dynamicTracer
-                     (Tracer traceM) dynamicTracer
+                     (Tracer traceM) dynamicTracer dynamicTracer --debugTracerG
```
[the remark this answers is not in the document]
crocodile-dentist: Intentional, I left it here so it's handy to toggle for debugging.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -341,59 +342,53 @@ runM Interfaces
...
+              inboundGovernorArgs   =
+                IG.Arguments {
+                  tracer                = dtLocalInboundGovernorTracer,
+                  transitionTracer      = nullTracer,
+                  debugTracer           = nullTracer,
+                  connectionDataFlow    = ntcDataFlow,
+                  idleTimeout           = Nothing,
+                  withConnectionManager = localWithConnectionManager localInbInfoChannel,
+                  mkConnectionHandler   = mkLocalConnectionHandler,
+                  infoChannel           = localInbInfoChannel } }
```
coot, line 390: I find it more readable to put the final brackets in separate lines, so it's easier to see when a given record finishes (especially in nested records).
```suggestion
                  infoChannel           = localInbInfoChannel
                }
              }
```

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -766,45 +737,77 @@ runM Interfaces
...
+                        tracer = dtInboundGovernorTracer,
+                        transitionTracer      = dtInboundGovernorTransitionTracer,
+                        debugTracer           = nullTracer,
+                        connectionDataFlow    = diNtnDataFlow,
+                        idleTimeout           = Just daProtocolIdleTimeout,
+                        withConnectionManager =
+                          withConnectionManagerInitiatorAndResponderMode inboundInfoChannel,
+                        mkConnectionHandler   = mkConnectionHandler
+                                                  SingInitiatorResponderMode diNtnDataFlow,
+                        infoChannel           = inboundInfoChannel } }
```
coot, line 763:
```suggestion
                        infoChannel           = inboundInfoChannel
                      }
                    }
```

ouroboros-network-framework/sim-tests/Test/Ouroboros/Network/Server/Sim.hs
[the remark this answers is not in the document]
coot: Doesn't the commented tracer need to be `(dynamicTracer <> debugTracerG)` instead?

ouroboros-network-framework/sim-tests/Test/Ouroboros/Network/Server/Sim.hs
```diff
@@ -532,17 +532,20 @@ instance Arbitrary req =>
   -- we could miss which change actually introduces the failure, and be lift
   -- with a larger counter example.
   shrink (MultiNodePruningScript
-            (AcceptedConnectionsLimit hardLimit softLimit delay)
+            acl@(AcceptedConnectionsLimit hardLimit softLimit delay)
             events
             attenuationMap) =
-    MultiNodePruningScript
-        <$> (AcceptedConnectionsLimit
+    let a = AcceptedConnectionsLimit
```
coot, line 538:
```suggestion
   let acls = AcceptedConnectionsLimit
```

ouroboros-network-framework/sim-tests/Test/Ouroboros/Network/Server/Sim.hs
```diff
@@ -532,17 +532,20 @@ instance Arbitrary req =>
...
               <$> shrink hardLimit
               <*> shrink softLimit
-              <*> pure delay)
-        <*> (makeValid
-            <$> shrinkList shrinkEvent events)
-        <*> shrink attenuationMap
+              <*> pure delay in
+      [MultiNodePruningScript a' events attenuationMap
+      | a' <- a] <>
```
coot, line 543:
```suggestion
      [MultiNodePruningScript acl' events attenuationMap
      | acl' <- acls] <>
```

ouroboros-network-framework/sim-tests/Test/Ouroboros/Network/Server/Sim.hs
```diff
@@ -226,7 +226,7 @@ data ConnectionEvent req peerAddr
     -- ^ Close an outbound connection.
   | ShutdownClientServer DiffTime peerAddr
     -- ^ Shuts down a client/server (simulates power loss)
-  deriving (Show, Functor)
+  deriving (Eq, Show, Functor)
```
coot, line 229: apropos `Improve Simulation testnet shrinker`: could you explain where the clue of the fix is?

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
```diff
@@ -1297,90 +1303,100 @@ diffusionSimulation
...
                                                  . tracerWithName ntnAddr
                                                  . tracerWithTime
-                                                 $ nodeTracer
+                                                 $ nodeTracer' -- <> sayTracer'
         , Diff.dtTraceLedgerPeersTracer          = contramap
                                                   DiffusionLedgerPeersTrace
                                                  . tracerWithName ntnAddr
                                                  . tracerWithTime
-                                                 $ nodeTracer
+                                                 $ nodeTracer' -- <> sayTracer'
```
coot, line 1354: we don't need the commented tracer, it's enough to change the hard coded `True` value in the `tracer'`, isn't it?

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
```diff
@@ -1297,90 +1303,100 @@ diffusionSimulation
...
                       NtCAddr NtCVersion NtCVersionData
                       SomeException Cardano.ExtraState
                       Cardano.ExtraState PeerTrustable
                       (Cardano.ExtraPeers NtNAddr)
                       (Cardano.ExtraPeerSelectionSetsWithSizes NtNAddr) m
-    mkTracers ntnAddr =
+    mkTracers ntnAddr i =
+      let sayTracer' = Tracer \msg -> say $ "(node-" <> show i <> ")" <> show msg
+          -- toggle and uncomment interesting sayTracer' below
+          nodeTracer' = if True then nodeTracer <> sayTracer' else nodeTracer in
```
coot, line 1332: why not just
```suggestion
          nodeTracer' = nodeTracer <> sayTracer' in
```

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
```diff
@@ -383,9 +385,12 @@ genNodeArgs relays minConnected localRootPeers self = flip suchThat hasUpstream
...
   -- Generating an InitiatorResponderMode node is 3 times more likely since we
   -- want our tests to cover more this case.
-  diffusionMode <- frequency [ (1, pure InitiatorOnlyDiffusionMode)
-                             , (3, pure InitiatorAndResponderDiffusionMode)
-                             ]
+  -- diffusionMode <- frequency [ (1, pure InitiatorOnlyDiffusionMode)
+  --                            , (3, pure InitiatorAndResponderDiffusionMode)
+  --                            ]
+  -- TODO: 'cm & ig enforce timeouts' fails in 'InitiatorOnlyDiffusionMode'
+  -- so we pin it to this
```
coot, line 392: Is this fixed in this PR?

Do you know what's the reason for the failure?

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -370,339 +370,340 @@ with
...
+            return (Just connId, state')
+
+          MiniProtocolTerminated
+            Terminated {
+                tConnId,
+                tMux,
+                tMiniProtocolData = mpd@MiniProtocolData { mpdMiniProtocol = miniProtocol },
+                tResult
+              } -> do
+            tResult' <- atomically tResult
```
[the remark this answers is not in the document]
crocodile-dentist: I had some issues with deadlocking the muxer thread since it would wait for some TVars that it itself would set later. By offloading this work into the IG, we can avoid that problem.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionHandler.hs
```diff
@@ -359,11 +367,32 @@ makeConnectionHandler muxTracer forkPolicy
...
+                          case inResponderMode of
+                            InResponderMode (inboundGovChannelTracer, connectionDataFlow)
+                              | Duplex <- connectionDataFlow agreedOptions -> do
+                                  -- In this case, following the Mx.run call below, the muxer begins racing with
+                                  -- the CM to write to the information channel queue. The tracer of mux activity,
+                                  -- while the CM of new connection notification. The latter *should* come first.
+                                  -- The IG tracer will block the muxer on a TVar when it reaches mature state
+                                  -- until the CM informs the former of new peer connection to ensure
+                                  -- proper sequencing of events.
+                                  countersVar <- newTVarIO $ Just $ ResponderCounters 0 0
```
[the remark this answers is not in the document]
crocodile-dentist: I think it's better than the alternative of creating a map and inserting counters in there. You get stm contention, and have to manage the peers in there as well as the connections in the IG state. If we forgot to delete a peer, it will stay there forever. Here, when the connection drops, the state is garbage collected.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -139,32 +155,181 @@ with :: forall (muxMode :: Mux.Mode) socket initiatorCtx peerAddr versionData ve
...
+                          _orNot -> writeTVar countersVar $ Just rc { numTraceWarmResponders = succ numTraceWarmResponders }
+
+                  case csRemoteState of
+                    -- we retry on expired because we let the IG
+                    -- loop handle this peer. If the connection is released,
+                    -- and CM reports CommitTr, this peer will disappear
+                    -- from the connections so on retry we will hit the
+                    -- _otherwise clause instead and promotion will fail,
+                    -- as it should. Otherwise, if KeepTr is returned,
+                    -- we can handle 'AwakeRemote' from this peer.
```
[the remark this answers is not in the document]
crocodile-dentist: In that case, after retry, the peer will be gone from the connections, so the map lookup in the case will return Nothing and we will take the _otherwise path next time, also replacing the counters with Nothing for good measure.

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
crocodile-dentist: I haven't investigated, but it fails on main as well. Except on main, this is pinned elsewhere, so when you look at the trace of a failed test, at the top it would show that a node is in InitiatorAndResponder mode, then in the trace you would see some error about invalid duplex mode and you would scratch your head to find out that the node is actually InitiatorOnly. I've moved the pin here so it's clear.

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
crocodile-dentist: I like it my may, because if you're trying to debug a test failure that you know doesn't concern, say dns resolution, you can cut out a lot of the noise by flipping the main switch to False and just uncomment the relevant sayTracers, which may be sometimes just 1 or 2 - some low number. This will generate really short trace.

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
crocodile-dentist: If we are sure apriori that all we need is this tracer to quickly debug, we set the flag to False and just uncomment here. The trace will be very succinct.

ouroboros-network/testlib/Test/Ouroboros/Network/Diffusion/Testnet/Cardano/Simulation.hs
crocodile-dentist: So for CI, we leave all say tracers on, this is mostly for local debugging where you don't want to stare at 10,000 lines of traces.

ouroboros-network-framework/sim-tests/Test/Ouroboros/Network/Server/Sim.hs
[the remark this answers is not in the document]
crocodile-dentist: True, I've changed the definition of `debugTracerG` to:
```
debugTracerG :: (Show a, Typeable a) => Tracer (IOSim s) a
debugTracerG =    Tracer (\msg -> getCurrentTime >>= say . show . (,msg))
               <> Tracer traceM
            -- <> Tracer Debug.traceShowM
```

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -497,7 +497,10 @@ with
               Just err -> traceWith tracer (TrMuxErrored connId err)
 
             -- the connection manager does should realise this on itself.
-            let state' = unregisterConnection connId state
+            -- we bypass the assertion check since MuxFinished could have been
+            -- placed on the queue before we managed to remove the connection from
+            -- the private state.
+            let state' = unregisterConnection True connId state
```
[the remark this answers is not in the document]
crocodile-dentist: It's just what is in the stateVar which is not shared with the outside world in any way.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -413,6 +411,9 @@ with args@Arguments {
        }
      classifyHandleError
      inboundGovernorInfoChannel
+     ConnectionHandler {
+       connectionHandler
+     }
```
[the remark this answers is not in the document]
crocodile-dentist: Maybe something got mixed up in the rebase.

network-mux/src/Network/Mux/Trace.hs
```diff
@@ -208,3 +210,20 @@ instance Show Trace where
...
+data MuxTracerBundle m = MuxTracerBundle {
+  channelTracer :: Tracer m Trace,
+  -- ^ the tracer for individual miniprotocol messages
+  -- to not spam the inbound governor information channel with
+  -- tags which are (a) frequent (b) not interesting
+  muxTracer     :: Tracer m Trace
+  -- ^ the main tracer, which contains an embedded inbound governor
+  -- information channel tracer for inbound and (outbound duplex)
+  -- connections only (see `ConnectionHandler`)
+  }
```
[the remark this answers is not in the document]
crocodile-dentist: But the bearer is a different component and its traces are distinct from the ones emitted by the `channelTracer` in my patch and so this is still confusing. Neither one of these fields will trace the low level bearer tags.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionHandler.hs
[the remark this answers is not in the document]
crocodile-dentist: Initially, I've had one map for all the counters but I've moved away from it for the reasons I gave. Then I tried an STRef which was explicit in the sense that it was some state cell not shared with any other threads, but it required some additional imports and was not convenient to work with - the TVar here can go in one atomic action in the IG tracer code so it is slick, doesn't require any more imports anywhere.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -139,32 +155,181 @@ with :: forall (muxMode :: Mux.Mode) socket initiatorCtx peerAddr versionData ve
...
+          -- handshake, the muxer begins racing with the connection
+          -- manager on informing the IG of the new connection
+          -- and responder miniprotocols' activity. In principle,
+          -- the problematic case is when the IG tracer registers responder
+          -- miniprotocol launch command before the CM informs the IG of the
+          -- new connection. However unlikely, this will lead to
+          -- incoherent transitions between the two components.
+          -- The muxer blocks itself here to yield right-of-way
+          -- to the CM.
+          check . Map.member peer . connections =<< readTVar stateVar
```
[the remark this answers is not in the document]
crocodile-dentist: Did I miss something? I was convinced that the IG is informed about new inbound connections by the CM and that it doesn't know 'by itself'.

Other than that, I think it is doable the way you propose. When I wrote this part, I didn't have `connectionDataFlow` in the handler which I've had to later add because the problematic part turned out to be `Unidirectional` mode on the outbound side. Now that I have that callback there, the handler has enough information to decide whether to notify the IG of the new connection. So my initial conclusions are no longer pertinent after I made more changes, but it slipped my attention to go back to this because it worked.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -139,32 +155,181 @@ with :: forall (muxMode :: Mux.Mode) socket initiatorCtx peerAddr versionData ve
...
+                                InfoChannel.writeMessage infoChannel $ RemotePromotedToHot peer
+                          (Hot, _, 0) ->
+                                InfoChannel.writeMessage infoChannel $ RemotePromotedToHot peer
+                          (Hot, _, _) -> pure ()
+                          (_orNot, 0, 0) -> do
+                            InfoChannel.writeMessage infoChannel $ AwakeRemote peer
+                          _otherwise -> pure ()
+                        case miniProtocolTemp of
+                          Hot    -> writeTVar countersVar $ Just rc { numTraceHotResponders = succ numTraceHotResponders }
+                          _orNot -> writeTVar countersVar $ Just rc { numTraceWarmResponders = succ numTraceWarmResponders }
```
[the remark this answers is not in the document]
crocodile-dentist: I think that is indeed simpler and should work.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE BlockArguments           #-}
```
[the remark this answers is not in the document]
crocodile-dentist: Sure, I can try to minimize the damage.

ouroboros-network-framework/src/Ouroboros/Network/Server.hs
```diff
@@ -127,10 +106,18 @@ with :: forall muxMode socket initiatorCtx peerAddr versionData versionNumber m
...
-    => Arguments muxMode socket initiatorCtx peerAddr versionData versionNumber ByteString m a b
+    => Arguments muxMode socket peerAddr initiatorCtx responderCtx handle handlerTrace
+                 handleError versionNumber versionData bytes m a b x
     -- ^ record which holds all server arguments
-    -> (Async m Void -> m (InboundGovernor.PublicState peerAddr versionData) -> m x)
+    -> (   Async m Void
+        -> m (InboundGovernor.PublicState peerAddr versionData)
+        -> ConnectionManager
+              muxMode socket peerAddr handle handleError m
+        -> m x)
```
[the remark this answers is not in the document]
crocodile-dentist: But it is not, at least I'm not sure what you mean. The connection manager is available once the inbound governor provides the tracer.

ouroboros-network-framework/src/Ouroboros/Network/Server.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE BlockArguments      #-}
```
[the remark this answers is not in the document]
crocodile-dentist: Which part - the Arguments type and the with signature are adapted. Some fields from the record were moved into the fields of the InboundGovernor.Arguments record where they belong (eg. trTracer, inboundIdleTimeout, etc). The with signature had to be adapted and in particular the continuation now takes a `ConnectionManager` parameter since that is available once the IG provides the tracer. The BlockArguments is just sugar for places I've had to edit anyway.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1055,8 +1056,8 @@ with args@Arguments {
                     case inboundGovernorInfoChannel of
                       InResponderMode infoChannel ->
                         atomically $ InfoChannel.writeMessage
-                                       infoChannel
-                                       (NewConnectionInfo provenance connId dataFlow handle)
+                                       infoChannel $
+                                       NewConnection (NewConnectionInfo provenance connId dataFlow handle)
```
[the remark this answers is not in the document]
crocodile-dentist: If nothing else, it does have a dedicated show instance which makes reading debug logs a little more pleasant.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -578,7 +578,7 @@ with args@Arguments {
             -- time and making us go above timeout schedules.
             traverse
               (\thread -> do
-                throwTo (asyncThreadId thread) AsyncCancelled
+                throwTo (asyncThreadId thread) Mx.ColdBlooded
```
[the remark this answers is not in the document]
crocodile-dentist: There is another comment here that we can remove the queue from the CM since mux can in principle notify the IG of a new connection. This would simplify the CM interface, at the cost of not being able to perform queue draining during shutdown.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -98,17 +103,26 @@ data Arguments muxMode socket initiatorCtx peerAddr versionNumber versionData m
       -- ^ debug inbound governor tracer
       connectionDataFlow :: versionData -> DataFlow,
       -- ^ connection data flow
-      infoChannel        :: InboundGovernorInfoChannel muxMode initiatorCtx peerAddr versionData ByteString m a b,
+      infoChannel        :: InboundGovernorInfoChannel muxMode peerAddr initiatorCtx versionData ByteString m a b,
```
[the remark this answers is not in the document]
crocodile-dentist: This was accidental.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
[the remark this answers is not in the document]
crocodile-dentist: I've changed the patch so that the connection handler informs the IG of a new connection, and to avoid circular dependencies this type is useful. I've moved it to the ConnectionHandler module.

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -766,45 +737,77 @@ runM Interfaces
...
-                                          -- wait for any thread to fail:
-                                          snd <$> Async.waitAny [ ledgerPeersThread
-                                                                , localRootPeersProvider
-                                                                , governorThread
-                                                                , churnGovernorThread
-                                                                , inboundGovernorThread
-                                                                ]
+          inboundInfoChannel <- newInformationChannel
+          let mkConnectionHandler =
+                makeConnectionHandler' daApplicationInitiatorResponderMode
```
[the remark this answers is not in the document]
crocodile-dentist: This is re the optional commit: do we want the GADT or is the type family ok to stay?

ouroboros-network/src/Ouroboros/Network/Diffusion.hs
```diff
@@ -766,45 +737,77 @@ runM Interfaces
...
+                      IG.Arguments {
+                        tracer = dtInboundGovernorTracer,
+                        transitionTracer      = dtInboundGovernorTransitionTracer,
+                        debugTracer           = nullTracer,
+                        connectionDataFlow    = diNtnDataFlow,
+                        idleTimeout           = Just daProtocolIdleTimeout,
+                        withConnectionManager =
+                          withConnectionManagerInitiatorAndResponderMode inboundInfoChannel,
+                        mkConnectionHandler = mkConnectionHandler
+                                            . MuxInitiatorResponderConnectionHandler diNtnDataFlow,
```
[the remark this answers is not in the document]
crocodile-dentist: See comment below.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -182,154 +182,23 @@ with
...
+                                                         active
     withConnectionManager connectionHandler \connectionManager ->
       withAsync
         (  labelThisThread "inbound-governor-loop" >>
            forever (inboundGovernorStep connectionManager stateVar >> yield)
-         `catch`
-           handleError stateVar)
+         `catch` \e -> do
+           atomically $ writeTVar active False
+           _ <- atomically $ InfoChannel.readMessages infoChannel
```
coot, line 197: Could you leave a comment for future us why we need to flush the info channel.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
```diff
@@ -967,3 +945,7 @@ data Trace peerAddr
 
 data Debug peerAddr versionData = forall muxMode initiatorCtx m a b.
     Debug (State muxMode initiatorCtx peerAddr versionData m a b)
+
+data LoopDecision state peer = TraceOnly
```
coot, line 949: Do we need the `TraceOnly` constructor? If we do maybe there's a more suitable name for it.

coot (commented):
Just minor comments.

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
crocodile-dentist: We can use `Maybe`

ouroboros-network-framework/src/Ouroboros/Network/ConnectionHandler.hs
```diff
@@ -391,28 +437,30 @@ makeConnectionHandler muxTracer singMuxMode
...
+                       hVersionData    = agreedOptions
+                     }
+               atomically $ writePromise (Right $ HandshakeConnectionResult handle (versionNumber, agreedOptions))
+               withBuffer \buffer -> do
+                 bearer <- mkMuxBearer sduTimeout socket buffer
+                 countersVar <- newTVarIO . SJust $ ResponderCounters 0 0
+                 let traceWithBearer = contramap $ Mx.WithBearer connectionId
+                 unmask $ Mx.run Mx.Tracers {
+                            muxTracer     = traceWithBearer (muxTracer <> inboundGovernorMuxTracer countersVar),
+                            channelTracer = traceWithBearer muxTracer }
```
coot, line 462: Is this change related to this branch or more with #5121?

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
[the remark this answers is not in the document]
coot: I presume the other place you added it was the mux tracer. It deadlocked the muxer because while logging, it was waiting for its own termination and thus not being able to make progress.

It's worth commenting in the haddock of the mux tracer that one needs to be careful not to deadlock the muxer as the tracer runs in its context rather than the context of the inbound governor.

ouroboros-network-framework/src/Ouroboros/Network/Server.hs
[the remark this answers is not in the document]
coot: :thinking: is there a way to organise the code so that the connection manager is set up outside the inbound governor as in `main`?

ouroboros-network-framework/src/Ouroboros/Network/InboundGovernor.hs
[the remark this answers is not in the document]
crocodile-dentist: Yes, there's a comment at the top of the tracer `--hello from muxer main thread` :)

ouroboros-network-framework/src/Ouroboros/Network/Server.hs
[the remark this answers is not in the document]
crocodile-dentist: A different way of looking at this is that this continuation is the same as:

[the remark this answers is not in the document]
coot: Thanks, I might need to do it in #5046.

coot (approved):
LGTM

PR #5110, opened 2025-04-16
Bugfix lingering outbound connection

crocodile-dentist:
A race condition was identified which would manifest as an exception raised in the CM when attempting to acquire an outbound connection to a peer which is in a cooling state. This could happen when for eg. when churn demotes a warm peer and then immediately promote the same peer. Currently, the peer being demoted is removed too early from the established & inProgressDemoteWarm sets in the outbound governor's private state. The proposed fix ensure that the peer is completely demoted to Cold before it is removed from those sets.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -962,6 +963,7 @@ withPeerStateActions PeerStateActionsArguments {
         Nothing -> do
           Mux.stop pchMux
           atomically (writeTVar pchPeerStatus PeerCooling)
+          atomically $ check . (== PeerCold) =<< readTVar pchPeerStatus
```
bolt12, line 966: whenever we throwIO here we don't need the `check` because `establishConnection` error handler is going to do the right thing

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -973,6 +975,7 @@ withPeerStateActions PeerStateActionsArguments {
           -- relay on mux property: if any of the mini-protocols errors, mux
           -- throws an exception as well.
           atomically (writeTVar pchPeerStatus PeerCooling)
+          atomically $ check . (== PeerCold) =<< readTVar pchPeerStatus
```
bolt12, line 978: same here, not needed

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -998,7 +1001,7 @@ withPeerStateActions PeerStateActionsArguments {
              then traceWith spsTracer (PeerStatusChanged (HotToWarm pchConnectionId))
              else do
                  traceWith spsTracer (PeerStatusChangeFailure
-                                      (WarmToHot pchConnectionId)
+                                      (HotToWarm pchConnectionId)
```
bolt12, line 1004: good catch!

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE BlockArguments      #-}
```
bolt12, line 1: I don't think we gain much by adding this

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -1012,7 +1015,7 @@ withPeerStateActions PeerStateActionsArguments {
             pchAppHandles,
             pchMux
           } = do
-      atomically $ do
+      atomically do
```
bolt12, line 1018: Not quite necessary, i'd say

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -1033,11 +1036,11 @@ withPeerStateActions PeerStateActionsArguments {
           -- timeout fired
           Mux.stop pchMux
           wasWarm <- atomically (updateUnlessCoolingOrCold pchPeerStatus PeerCooling)
-          when wasWarm $
+          atomically $ check . (== PeerCold) =<< readTVar pchPeerStatus
```
bolt12, line 1039: Do we need this here? I think it is enough to have the check in the `jobDemoteEstablishedPeers`

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -1046,10 +1049,11 @@ withPeerStateActions PeerStateActionsArguments {
           -- rely on mux property: if any of the mini-protocols errors, mux
           -- throws an exception as well.
           wasWarm <- atomically (updateUnlessCoolingOrCold pchPeerStatus PeerCooling)
-          when wasWarm $
+          when wasWarm do
             traceWith spsTracer (PeerStatusChangeFailure
                                   (WarmToCooling pchConnectionId)
                                   (ApplicationFailure errs))
+          atomically $ check . (== PeerCold) =<< readTVar pchPeerStatus
```
bolt12, line 1056: same here, we throw so we don't need the `check`

bolt12 (changes requested):

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
crocodile-dentist: I think you're right, I've put this here before I realized I really need that check in the 'jobDemoteEstablishedPeers'.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
crocodile-dentist: But I do like this check here, because I can remove the peer from the warm and demotewarm sets in the handler safely.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
crocodile-dentist: It's such a lovely extension, but ok I'll wait for a better patch someday.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
crocodile-dentist: Maybe we can make use of this in the handlers to remove the peers safely from the various sets, as per my comment below?

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
bolt12: I am not a fan to be honest 😝

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/ActivePeers.hs
```diff
@@ -946,18 +947,19 @@ jobDemoteActivePeer
...
     Job job handler () "demoteActivePeer"
   where
     handler :: SomeException -> m (Completion m extraState extraDebugState extraFlags extraPeers peeraddr peerconn)
-    handler e = return $
-      -- It's quite bad if demoting fails. The peer is cooling so
-      -- we can't remove it from the set of established and hot peers.
-      --
-      Completion $ \st@PeerSelectionState {
+    handler e = do
+      atomically $ monitorPeerConnection peerconn >>= check . (== PeerCold) . fst
```
bolt12, line 957: 👍🏼

crocodile-dentist:
I'll squash the commits before pulling the trigger.

crocodile-dentist:
closes #5111

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -877,19 +878,18 @@ jobDemoteEstablishedPeer
...
     handler :: SomeException
             -> m (Completion m extraState extraDebugState extraFlags extraPeers
                              peeraddr peerconn)
-    handler e = return $
-      -- It's quite bad if closing fails. The peer is cooling so
-      -- we can't remove it from the set of established peers.
-      --
-      Completion $ \st@PeerSelectionState {
+    handler e = do
+      atomically $ monitorPeerConnection peerconn >>= check . (== PeerCold) . fst
```
coot, line 891: Could you add a haddock explaining why we have to block on the peer to become `cold`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -854,7 +854,7 @@ withPeerStateActions PeerStateActionsArguments {
     monitorPeerConnection PeerConnectionHandle { pchPeerStatus, pchAppHandles } =
          p  <$> readTVar pchPeerStatus
             <*> (g <$> traverse f pchAppHandles)
-        `orElse` throwSTM MonitorPeerConnectionBlocked
+        `orElse` throwSTM MonitorPeerConnectionBlocked -- ^ looks like throwSTM here is dead code?
```
coot, line 857:
```suggestion
        `orElse` throwSTM MonitorPeerConnectionBlocked
```
it is, and if we introduce a bug in `monitorPeerConnection` we'll discover it in a failing test.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/ActivePeers.hs
coot: add some haddocs here too, a link to the issue / PR might also be helpful.

ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -909,21 +936,33 @@ jobDemoteEstablishedPeer PeerSelectionActions {
...
 
     job :: m (Completion m extraState extraDebugState extraFlags extraPeers peeraddr peerconn)
     job = do
-      closePeerConnection peerconn
+      -- cf. issue #5111 - we wait for the connection to be forgotten by the CM
+      -- before we return. Otherwise, if the same peer is promoted too quickly by
+      -- eg. churn, the connection will fail.
+      res <-
+        timeout c_PEER_DEMOTION_TIMEOUT do
+          closePeerConnection peerconn
```
coot, line 959: `closePeerConnection` has it's own timeout (set to `120s`); it will be cearer if we just wrap the atomic action below in its own timeout.

PR #5112, opened 2025-04-18
Split mux's Tracer type

coot:
- **network-mux: removed handshake traces**
- **network-mux: split mux tracer into three independent parts**


crocodile-dentist (approved):
lgtm

PR #5121, opened 2025-05-13
CM bugfixes

crocodile-dentist:
# Description

Fix async exception unmasking
Properly unmask job pool jobs

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/network-mux/src/Control/Concurrent/JobPool.hs
+++ b/network-mux/src/Control/Concurrent/JobPool.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE BangPatterns        #-}
+{-# LANGUAGE BlockArguments      #-}
 {-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -76,14 +77,14 @@ forkJob' :: forall group m a.
             ( MonadAsync m, MonadMask m
             , Ord group
             )
-         => (m () -> m (Async m ()))
+         => (((forall x. m x -> m x) -> m ()) -> m (Async m ()))
          -- ^ how to fork a thread, e.g. `async`, `asyncOn`.
          -> JobPool group m a
          -> Job     group m a
          -> m ()
 forkJob' doFork JobPool{jobsVar, completionQueue} (Job action handler group label) =
-    mask $ \restore -> do
-      jobAsync <- doFork $ do
+    mask_ do
+      jobAsync <- doFork \restore -> do
         tid <- myThreadId
         io tid restore
           `onException`
@@ -120,7 +121,7 @@ forkJob :: forall group m a.
         => JobPool group m a
         -> Job     group m a
         -> m ()
-forkJob = forkJob' async
+forkJob = forkJob' asyncWithUnmask
 
 
 -- | Fork a `Job` using `asyncOn`.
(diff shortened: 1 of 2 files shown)


ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -869,7 +869,7 @@ with args@Arguments {
                                  hardLimit
                                  socket
                                  connId = do
-        r <- modifyTMVar stateVar $ \state -> do
+        r <- mask \unmask -> modifyTMVar unmask stateVar \state -> do
```
coot, line 872: Wouldn't it be cleaner if we leave `modifyTMVar` as it used to be, and do

```
mask \unmask -> modifyTMVar stateVar \state -> unmask do ...
```

At least `modifyTMVar` would be close to something standard.

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -2064,14 +2059,16 @@ with args@Arguments {
...
+                                 return . Right $ (connState', trTrace)
                              )
           case r of
-            Right connState -> do
-              let connState' = TerminatingState connId connThread Nothing
-              atomically $ writeTVar connVar connState'
-              traceWith trTracer (TransitionTrace connStateId
-                                   (mkTransition connState connState'))
+            Right (connState', trace') -> do
+              traceWith trTracer trace'
```
coot, line 2071: Is this necessary? It seems it's not, if so, I'd leave it as it was. It was nicely structured into reading / pattern-matching / writing sections - which is common in this module.

crocodile-dentist: The racy risk is that if the timeout expires, the connection can be reactivated just before it is set to `TerminatingState`. In the atomic block, this can't happen.

network-mux/src/Control/Concurrent/JobPool.hs
```diff
@@ -76,14 +77,14 @@ forkJob' :: forall group m a.
             ( MonadAsync m, MonadMask m
             , Ord group
             )
-         => (m () -> m (Async m ()))
+         => (((forall x. m x -> m x) -> m ()) -> m (Async m ()))
```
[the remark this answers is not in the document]
crocodile-dentist: Sure, but wasn't this the risk as well before?

ouroboros-network-framework/src/Ouroboros/Network/ConnectionManager/Core.hs
crocodile-dentist: Yes, that's better

network-mux/src/Control/Concurrent/JobPool.hs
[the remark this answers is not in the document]
crocodile-dentist: To expand a little bit, unless I misunderstand, this is clear that this function must receive a callback which provides an `unmask` continuation and we use it the same as before. Prior to the patch, the thread also started in a masked state so there is no difference besides that using the old restore relied on undefined behavior. Or do you mean we want to give the user the ability to permanently start the new thread in a masked state without the ability to unmask itself, which wasn't possible earlier either?

coot (approved):
LGTM

PR #5130, opened 2025-05-28
Fix GenesisMode syncing from origin

crocodile-dentist:
# Description

- **fixes retrieval of big ledger peers from the snapshot when `useLedgerAfterSlot = 0` and syncing from the origin**
- **fixes initial churning targets for Genesis and Praos modes**

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/cardano-client/cardano-client.cabal
+++ b/cardano-client/cardano-client.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: cardano-client
-version: 0.5.1.0
+version: 0.5.1.1
 synopsis: An API for ouroboros-network
 description: An API for ouroboros-network.
 license: Apache-2.0
--- a/network-mux/CHANGELOG.md
+++ b/network-mux/CHANGELOG.md
@@ -6,7 +6,15 @@
 
 ### Non-breaking changes
 
-## 0.8.0.0 -- 205-05-13
+## 0.8.0.1 -- 2025-06-02
+
+### Breaking changes
+
+### Non-breaking changes
+
+* Properly unmask async exceptions in the job pool
+
+## 0.8.0.0 -- 2025-05-13
 
 ### Breaking changes
 
--- a/network-mux/network-mux.cabal
+++ b/network-mux/network-mux.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: network-mux
-version: 0.8
+version: 0.8.0.1
 synopsis: Multiplexing library
 description: Multiplexing library.
 license: Apache-2.0
--- a/ouroboros-network-framework/CHANGELOG.md
+++ b/ouroboros-network-framework/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 ### Non-breaking changes
 
+## 0.18.0.1 -- 2025-06-02
+
+### Breaking changes
+
+### Non-breaking changes
+
+* properly unmask async exceptions in the CM
+
 ## 0.18.0.0 -- 2025-05-13
 
 ### Breaking changes
--- a/ouroboros-network-framework/ouroboros-network-framework.cabal
+++ b/ouroboros-network-framework/ouroboros-network-framework.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: ouroboros-network-framework
-version: 0.18.0.0
+version: 0.18.0.1
 synopsis: Ouroboros network framework
 description: Ouroboros network framework.
 license: Apache-2.0
@@ -89,7 +89,7 @@ library
     monoidal-synchronisation ^>=0.1.0.6,
     mtl,
     network ^>=3.2.7,
-    network-mux ^>=0.8,
+    network-mux ^>=0.8.0.1,
     nothunks,
     nothunks ^>=0.1.4 || ^>=0.2,
     ouroboros-network-api ^>=0.14,
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -6,6 +6,21 @@
 
 ### Non-breaking changes
 
+## 0.21.2.0 -- 2025-06-02
+
+### Breaking changes
+
+### Non-breaking changes
+
+* Bugfix retrieve ledger peers from snapshot when `useLedgerPeers: 0`
+  in the topology file
+* Explicitly provide the following peer selection sync target defaults:
+  targetNumberOfRootPeers, targetNumberOfKnownPeers, targetNumberOfEstablishedPeers,
+  targetNumberOfActivePeers
+* Update peer selection deadline default targets to match agreed upon
+  values which the node has been released with in the configuration files
+* minor churn bugfixes and improvements
+
 ## 0.21.1.0 -- 2025-05-26
 
 ### Non-breaking changes
--- a/ouroboros-network/ouroboros-network.cabal
+++ b/ouroboros-network/ouroboros-network.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: ouroboros-network
-version: 0.21.1.0
+version: 0.21.2.0
 synopsis: A networking layer for the Ouroboros blockchain protocol
 description: A networking layer for the Ouroboros blockchain protocol.
 license: Apache-2.0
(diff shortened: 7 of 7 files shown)


ouroboros-network/src/Ouroboros/Cardano/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -111,6 +111,10 @@ targetPeers Cardano.ExtraPeerSelectionActions {
             } =
     Guarded Nothing $ do
       churnTargets <- readPeerSelectionTargets
+      -- the following is only necessary to support legacy sync mode
+      -- and can be removed once Genesis is fully adopted.
```
coot, line 115: Let's leave it as a TODO, so we can grep it :grin:
```suggestion
      -- TODO: the following is only necessary to support legacy sync mode
      -- and can be removed once Genesis is fully adopted.
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -386,29 +386,30 @@ stakeMapWithSlotOverSource StakeMapOverSource {
...
-        case cachedSlot of
-          Just thatSlot | thatSlot == snapshotSlotNo ->
-                          (peerMap, bigPeerMap, cachedSlot)
-          _otherwise -> ( accPoolStake (map snd accSnapshotRelays)
-                        , Map.fromAscList accSnapshotRelays
-                        , Just snapshotSlotNo)
+    -- check if we can use the snapshot first
+    (ledgerSlotNo, _, Just (LedgerPeerSnapshot (At snapshotSlotNo, accSnapshotRelays)))
+      | snapshotSlotNo >= ledgerSlotNo'
+      , snapshotSlotNo >= ula' ->
```
coot, line 392: :thinking: shouldn't this be:
```hs
      | snapshotSlotNo <  ledgerSlotNo'
      , snapshotSlotNo >= ula' ->
```
e.g. the snapshot must be newer than _use ledger after_ (`ula'`), but not newer than `ledgerSlotNo'` (if it's otherwise we should use the ledger peers, not the snapshot).

Not in this PR, but we should rename `ula` and its incarnations to `useLedgerAfter`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
```diff
@@ -386,29 +386,30 @@ stakeMapWithSlotOverSource StakeMapOverSource {
...
+          -- we cache the peers from the snapshot
+          -- to avoid unnecessary work
+          case cachedSlot of
+            Just thatSlot | thatSlot == snapshotSlotNo ->
+                              (peerMap, bigPeerMap, cachedSlot)
```
coot, line 397: Such conditions should be an inequality (with STM you're not guaranteed to hit `snapshtoSlotNo`).  We want to use the cached data when it's newer than the snapshot.
```suggestion
            Just thatSlot | thatSlot >= snapshotSlotNo ->
                              (peerMap, bigPeerMap, cachedSlot)
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
karknu: I think that the `snapshotSlotNo >= ledgerSlotNo'` check is correct. 

However if we want to have the usage of the snapshot file depend on `useLedgerAfter` cardano-node should issue a warning when one provides a snapshotfile with that is older than `useLedgerAfter` since it will never be used.

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
```diff
@@ -961,7 +961,7 @@ runM Interfaces
 
       localRootsVar <- newTVarIO mempty
 
-      peerSelectionTargetsVar <- newTVarIO daPeerSelectionTargets
+      peerSelectionTargetsVar <- newTVarIO nullPeerSelectionTargets --daPeerSelectionTargets
```
karknu, line 964: Remove the trailing  `--daPeerSelectionTargets`

karknu (approved):
LGTM

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
crocodile-dentist: There is somewhat of a risk that what at some point was a big ledger peer is no longer in the new snapshot, but we may have a big ledger peer connection to that peer that gives a false sense of security. However, a check is performed for each reload once we reach the slot of a snapshot and check the stake distribution - it must match what was on that particular snapshot that was loaded at some point, otherwise an error is traced but the node continues.

ouroboros-network/src/Ouroboros/Network/PeerSelection/LedgerPeers.hs
crocodile-dentist: > However if we want to have the usage of the snapshot file depend on useLedgerAfter cardano-node should issue a warning when one provides a snapshotfile with that is older than useLedgerAfter since it will never be used.

I've added this check to my integration branch at the node level.

ouroboros-network/src/Ouroboros/Cardano/PeerSelection/Churn.hs
```diff
@@ -127,18 +127,25 @@ peerChurnGovernor PeerChurnArgs {
...
   -- to become warm.
   startTs0 <- getMonotonicTime
-  -- TODO: revisit the policy once we have local root peers in the governor.
-  -- The intention is to give local root peers give head start and avoid
-  -- giving advantage to hostile and quick root peers.
-  threadDelay 3
+
+  -- if this code here is removed, then the initial peer churn targets
+  -- in the TVar must not be nullPeerselectionTargets otherwise
+  -- targetPeers in Monitor will not work due to a check!
```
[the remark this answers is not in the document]
crocodile-dentist: The issue is a race at startup between churn and peer selection in Praos mode. @karknu and Nick were talking about when does the node startup in TooOld state, but at least in my experience the answer is usually or almost always, or at least probably often enough. So the ledger state will immediately turn to `TooOld` and `targetPeers` will set the targets away from the initial `nullPeerSelectionTargets`. But in churn, the idea is that we want to give a head start to local root peers. So the check to prevent this action is to not do this when the churn targets are `nullPeerSelectionTargets`. For PraosMode, churn will take the praos targets, set the root peers to 0, and write to the tvar, which will unblock targetPeers.

ouroboros-network/src/Ouroboros/Cardano/Network/Diffusion/Configuration.hs
```diff
@@ -29,8 +28,11 @@ defaultNumBootstrapPeers = DefaultNumBootstrapPeers 30
 --
 defaultSyncTargets :: PeerSelectionTargets
 defaultSyncTargets =
-  defaultDeadlineTargets {
-    targetNumberOfActivePeers               = 0,
+  PeerSelectionTargets {
+    targetNumberOfRootPeers                 = 0,
+    targetNumberOfKnownPeers                = 85,
```
karknu, line 33: This should be 150.

ouroboros-network/src/Ouroboros/Cardano/PeerSelection/Churn.hs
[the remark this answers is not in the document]
crocodile-dentist: I've deployed a node in both Genesis and Praos consensus modes to verify that this does work as intended by observing the traces.

ouroboros-network/src/Ouroboros/Network/Diffusion/P2P.hs
```diff
@@ -961,7 +961,7 @@ runM Interfaces
 
       localRootsVar <- newTVarIO mempty
 
-      peerSelectionTargetsVar <- newTVarIO daPeerSelectionTargets
+      peerSelectionTargetsVar <- newTVarIO nullPeerSelectionTargets
```
coot, line 965:
```suggestion
      -- churn will set initial targets
      peerSelectionTargetsVar <- newTVarIO nullPeerSelectionTargets
```

ouroboros-network/src/Ouroboros/Cardano/PeerSelection/Churn.hs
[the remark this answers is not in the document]
coot: Thanks. Let's include a brief comment that code is to provide a head start for local roots and big ledger peers.

Issue #5114, opened 2025-04-24
Incremental decoding of CBOR-in-CBOR


coot:
[`unwrapCBORinCBOR`](https://ouroboros-network.cardano.intersectmbo.org/ouroboros-network-api/Ouroboros-Network-Block.html#v:unwrapCBORinCBOR) is not incremental;  it could be using https://github.com/well-typed/cborg/pull/286.

lehins:
Instance for `Serialised` also needs to be fixed, in particular the `TODO`s needs to be take care of. https://github.com/IntersectMBO/ouroboros-network/blob/86299622851df87089b16ac3f53d2879bca70b55/ouroboros-network-api/src/Ouroboros/Network/Block.hs#L472-L488

PR #5138, opened 2025-06-10
leverage lazy bytestring for 'Serialise' CBOR-in-CBOR.

KtorZ:
# Description

The main use case for this type lies within the StateQuery protocol, where it can be particularly useful to obtain a plain CBOR response. This is particularly useful for large query results such as DebugNewEpochState (multiple GB on mainnet). The network library makes a great effort at trying to serialise and deserialise bytes lazily throughout; and these efforts are unfortunately destroyed by this implementation that would here evaluate the entire ByteString when decoding and encoding. So on a machine that would have both a server and client using this library, we would pay twice the cost of fully evaluating in memory the entire response, instead of leverage lazy IO as, I believe, is originally intended.

Note that 'decode' and 'encode' here rely on the default implementation for lazy ByteString in Codec.Serialise, which do the right thing: encode lazy bytestrings as indefinite sequences of byte chunks; effectively preserving the laziness in both directions.

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network-api/src/Ouroboros/Network/Block.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/Block.hs
@@ -473,16 +473,13 @@ fromSerialised dec (Serialised payload) =
 --
 -- TODO: replace with encodeEmbeddedCBOR from cborg-0.2.4 once
 -- it is available, since that will be faster.
---
--- TODO: Avoid converting to a strict ByteString, as that requires copying O(n)
--- in case the lazy ByteString consists of more than one chunks.
 instance Serialise (Serialised a) where
   encode (Serialised bs) = mconcat [
         Enc.encodeTag 24
-      , Enc.encodeBytes (Lazy.toStrict bs)
+        encode bs
       ]
 
   decode = do
       tag <- Dec.decodeTag
       when (tag /= 24) $ fail "expected tag 24 (CBOR-in-CBOR)"
-      Serialised . Lazy.fromStrict <$> Dec.decodeBytes
+      Serialised <$> decode


ouroboros-network-api/src/Ouroboros/Network/Block.hs
```diff
@@ -473,16 +473,13 @@ fromSerialised dec (Serialised payload) =
...
 -- TODO: replace with encodeEmbeddedCBOR from cborg-0.2.4 once
 -- it is available, since that will be faster.
---
--- TODO: Avoid converting to a strict ByteString, as that requires copying O(n)
--- in case the lazy ByteString consists of more than one chunks.
 instance Serialise (Serialised a) where
   encode (Serialised bs) = mconcat [
         Enc.encodeTag 24
-      , Enc.encodeBytes (Lazy.toStrict bs)
+        encode bs
```
coot, line 479: Nice, so we avoid lazy to strict conversion, which indeed is costly.

The only problem with it is that we're modifying serialisation, which won't work across different versions (backwards compatibility).

I think we should:
* add `SerialisedV2` newtype wrapper,
* bump version of the `NodeToClient` mini-protocol
* use it in `LocalQueryLedger` mini-protocol if the negotiated version allows for it

An alternative is to modify the instance without changing the encoding (so it remains backwards compatible).

@nfrisby do you agree?  I think, this type is mostly used in `ouroboros-conesnsus`.

coot (changes requested):

ouroboros-network-api/src/Ouroboros/Network/Block.hs
amesgen: FTR the ticket for making decoding in this instance lazy is #5114 which we recently talked about in the context of incremental ledger serialization (EDIT: encoding indeed is difficult to do non-incremental as it is only possible if one knows the size upfront[^1]).

Also note that this instance (via `(un)wrapCBORinCBOR`) is used for sending eg txs/headers/blocks via N2C/N2N protocols, so all of these also would have to be patched for backwards-compatibility (across all implementations that do anything with txs/headers/blocks). Therefore,

> An alternative is to modify the instance without changing the encoding (so it remains backwards compatible).

sounds simpler to me.

[^1]: It doesn't really matter for txs/headers/blocks as we do know the size there, but it is annoying for the LSQ stuff which motivates this PR.

KtorZ: > The only problem with it is that we're modifying serialisation

Indeed, but that's also a desired feature IMO as it should prevent evaluating the entire bytestring in one chunk on the server; which can instead stream the response to clients. (I am really seeing this in the context of the state-query protocol).

I like the idea of making this version-specific; although the _version_ here should be driven by the NodeToClient version in the case of the state query protocols. I believe the Serialised type is also used elsewhere, where version-constaints may be different but all that can very likely be resolved through a type-class or a type family.

KtorZ: > An alternative is to modify the instance without changing the encoding (so it remains backwards compatible).

While this might be possible for `decode`, I don't think it would be possible for `encode` as we can't know the length of the bytestring when we begin serialising. So we need to at least partially evaluate it to know how many chunks are there if we want to use definite CBOR structures. 

For decoding, we might still be able to decode by chunks once we have parsed the CBOR header type and know the expected size of the ByteString. 

Having said that, doing it for even only just `decode` at least solves the problem on the receiving end. So that's half of the problem already solved :)

coot: From the RFC (section `3.4.5.1`):

> Sometimes it is beneficial to carry an embedded CBOR data item that is not meant to be decoded immediately at the time the enclosing data item is being decoded. Tag number 24 (CBOR data item) can be used to tag the embedded byte string as a single data item encoded in CBOR format. Contained items that aren't byte strings are invalid.

It seems what you're proposing in this patch is not a valid CBOR, see [`encodeChunked`](https://hackage.haskell.org/package/serialise-0.2.6.1/docs/src/Codec.Serialise.Class.html#encodeChunked).

KtorZ: I don't think you're reading this right. `encodeChunked` produces a single data item: a bytestring (major type 2).

Whether it is indefinite or definite doesn't really change the fact that it's a single CBOR data item.

coot: Yes, you're right.

I think we all agree on adding `SerialisedV2` to `ouroboros-newtork` and a new `NodeToClientVersion`.

PR #5146, opened 2025-06-20
Trace connection durations

crocodile-dentist:
# Description

It may be useful for some diagnostic purposes to know how long lived
an outbound hot connection has been. This patch introduces tracing
when a hot peer is demoted or errors out with the active time in seconds.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -1,10 +1,14 @@
 # Revision history for ouroboros-network
 
-## next release
+## 0.22.1.0 -- 2025-07-
 
 ### Breaking changes
 
 ### Non-breaking changes
+
+- Added `pchPromotedHotVar` to `PeerConnectionHandle` to track when a peer has been promoted to hot
+- Added tag `PeerHotDuration` to `PeerSelectionActionsTrace` to indicate how long a remote
+  peer has been in hot mode until it was either demoted or closed.
 
 ## 0.22.0.0 -- 28.06.2025
 
@@ -50,6 +54,10 @@
   of ledger peer.
 * Added `dispatchLookupWithTTL`
 * Lower the time to cache DNS errors to at most 15min.
+
+## 0.21.3.0
+
+Consult changelog on main
 
 ## 0.21.2.0 -- 2025-06-02
 


crocodile-dentist:
Are we actually interested in knowing the connection duration at the handler level? I thought about that and didn't find the metric useful enough in general and so decided against implementing re your first point. If we want to track outbound duration to big ledger peers, the connection duration might not give an accurate result if the connection was used via the inbound side strictly at least some of the time. This implementation will give the connection durations separately for outbound and inbound sides via patches in the inbound governor and peer state actions (via peer connection handle, so the duration can be exposed to peer selection governor easily in the future if we do decide to implement a max duration policy).

karknu (changes requested):
For this to be usable in the way we discussed I think that the lifetime for the hot connections must be included in the DebugSt data too.

We are looking for the life time of the longest lived connections, and they may still be active. That is emitting a trace event for hot demotions won't capture these connections.

ouroboros-network/cardano-diffusion/Cardano/Network/Diffusion/Handlers.hs
```diff
@@ -40,6 +41,7 @@ sigUSR1Handler
   -> STM IO UseBootstrapPeers
   -> STM IO LedgerStateJudgement
   -> PeerMetrics IO ntnAddr
+  -> (peerconn -> STM IO (Maybe Time))
```
coot, line 44: If I understand correctly, the callback is only called on currently active peers. Add a haddock comment:
```suggestion
  -> (peerconn -> STM IO (Maybe Time))
  -- ^ return time when an active peer was promoted to a hot peer.
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -620,14 +634,10 @@ withPeerStateActions PeerStateActionsArguments {
...
 
-    isNotCoolingOrCold :: StrictTVar m PeerStatus -> STM m Bool
-    isNotCoolingOrCold stateVar =
-      (> PeerCooling) <$> readTVar stateVar
-
     peerMonitoringLoop
       :: PeerConnectionHandle muxMode responderCtx peerAddr versionData ByteString m a b
       -> m ()
-    peerMonitoringLoop pch@PeerConnectionHandle { pchConnectionId, pchPeerStatus, pchAppHandles } = do
+    peerMonitoringLoop pch@PeerConnectionHandle { pchConnectionId, pchPeerStatus, pchAppHandles, pchPromotedHotVar } = do
```
coot: The line is pretty long:
```suggestion
    peerMonitoringLoop pch@PeerConnectionHandle { 
                             pchConnectionId,
                             pchPeerStatus,
                             pchAppHandles,
                             pchPromotedHotVar
                           } = do
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -669,10 +679,8 @@ withPeerStateActions PeerStateActionsArguments {
 
           Just (WithSomeProtocolTemperature (WithHot MiniProtocolError{})) -> do
             -- current `pchPeerStatus` must be 'HotPeer'
-            state <- atomically $ do
-              peerState <- readTVar pchPeerStatus
-              _  <- updateUnlessCoolingOrCold pchPeerStatus PeerCooling
-              return peerState
+            state <- atomically do
+              readTVar pchPeerStatus <* updateUnlessCoolingOrCold pchPeerStatus PeerCooling
```
coot: I'd live the original, monadic expressions are easier to read (or we're just more used to them).

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -687,10 +695,9 @@ withPeerStateActions PeerStateActionsArguments {
           Just (WithSomeProtocolTemperature (WithEstablished MiniProtocolError{})) -> do
             -- update 'pchPeerStatus' and log (as the two other transition to
             -- cold state.
-            state <- atomically $ do
-              peerState <- readTVar pchPeerStatus
-              _  <- updateUnlessCoolingOrCold pchPeerStatus PeerCooling
-              pure peerState
+            state <- atomically do
+              readTVar pchPeerStatus <* updateUnlessCoolingOrCold pchPeerStatus PeerCooling
+
```
coot: as above

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -913,106 +934,95 @@ withPeerStateActions PeerStateActionsArguments {
...
+            writeTVar pchPeerStatus PeerHot
+            return $ do
+              startProtocols SingHot isBigLedgerPeer connHandle
+              atomically . writeTVar pchPromotedHotVar . (Just $!) =<< getMonotonicTime
+              traceWith spsTracer (PeerStatusChanged (WarmToHot pchConnectionId))
+          _otherwise -> return $ do
+            traceWith spsTracer (PeerStatusChangeFailure
+                                  (WarmToHot pchConnectionId)
+                                  (ActiveCold peerStatus))
+            throwIO $ ColdActivationException pchConnectionId
```
coot: We can simplify this to avoid `join`.  It is there because there are two branches, but the `_otherwise` branch always throws an error, so it's the same as this one:
```suggestion
      atomically $ do
        peerStatus <- readTVar pchPeerStatus
        case peerStatus of
          PeerWarm -> do
            writeTVar (getControlVar SingHot pchAppHandles) Continue
            writeTVar (getControlVar SingWarm pchAppHandles) Quiesce
            writeTVar pchPeerStatus PeerHot

          _otherwise -> return $ do
            traceWith spsTracer (PeerStatusChangeFailure
                                  (WarmToHot pchConnectionId)
                                  (ActiveCold peerStatus))
            throwIO $ ColdActivationException pchConnectionId
        
      startProtocols SingHot isBigLedgerPeer connHandle
      atomically . writeTVar pchPromotedHotVar . (Just $!) =<< getMonotonicTime
      traceWith spsTracer (PeerStatusChanged (WarmToHot pchConnectionId))    
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -913,106 +934,95 @@ withPeerStateActions PeerStateActionsArguments {
...
+                  when trace do
+                    traceWith spsTracer (PeerStatusChangeFailure
+                                          (HotToCooling pchConnectionId)
+                                          TimeoutError)
+                  throwIO (DeactivationTimeout pchConnectionId)
+                Just _ -> return ()
+
+          -- we could genuinly hit this case due to a race between 'peerMonitoringLoop'
+          -- and peer selection demotion job
+          PeerWarm -> return $ pure ()
```
coot: Are you sure? Don't `hot -> warm` demotions only happen through an explicit call to `deactivatePeerConnection`?

coot (commented):
I like that we're getting rid of this `isWarm` stuff (this would fit into a separate PR or commit at least).  I'll need to look at the new logic for a bit longer when I feel better though.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
crocodile-dentist: Yes, there are two call sites. One is in the outbound governor, and the other in the peer monitoring loop. Consider when the OG demotes a peer, this function will write to the control var to terminate the miniprotocols and wait for them to finish. Once the first hot miniprotocol finishes, peer monitoring loop will notice this and enter this function as well. Potentially both of them can go into the first case to wait for all hot miniprotocols to finish - if the OG context didn't manage to flip this in time:
```
                              stateTVar pchPeerStatus \case
                                PeerHot -> (traceWith spsTracer (PeerStatusChanged (HotToWarm pchConnectionId))
                                           ,  PeerWarm)
                                x       -> (pure () , x)
```
The x -> case is for the loser to not trace twice, and not change the peer status var if it is <= `PeerCooling` because the winner was so fast that it started the demotion to cold.

If the winner managed to finish the first case, then the loser of the race will hit the `PeerWarm` case, but that should be fine.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -913,106 +934,95 @@ withPeerStateActions PeerStateActionsArguments {
...
-           throwIO $ ColdActivationException pchConnectionId
+            pchAppHandles,
+            pchPromotedHotVar } = do
+      join . atomically $ do
+        peerStatus <- readTVar pchPeerStatus
+        case peerStatus of
+          PeerWarm -> do
+            writeTVar (getControlVar SingHot pchAppHandles) Continue
+            writeTVar (getControlVar SingWarm pchAppHandles) Quiesce
+            writeTVar pchPeerStatus PeerHot
```
crocodile-dentist: I flip the `pchPeerStatus` to `PeerHot` (edit: before actually starting the protocols - they are still in `NotStarted` state), which should be fine because if we immediately try to async demote then we will block in `awaitAllResults` (either for them to start running just to be shutdown asap, or straight into `NotRunning`), except if Karl's proposed PR #5118 is merged, which will make it an error then. But I want to do it in one transaction to ensure that we only try to promote from `PeerWarm`.

ouroboros-network/cardano-diffusion/Cardano/Network/Diffusion/Handlers.hs
crocodile-dentist: That's the intention, but it is not an error to call it on a warm peer.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
crocodile-dentist: paging @karknu

karknu: The changes in #5118 doesn't fix the original bug it was intended for. I've closed the PR so you don't need to worry about it.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -796,9 +801,15 @@ withPeerStateActions PeerStateActionsArguments {
...
-                                        atomically $ do
+                                        promotedHot <- atomically $ do
                                           waitForOutboundDemotion spsConnectionManager connId
                                           writeTVar peerStateVar PeerCold
+                                          stateTVar pchPromotedHotVar (, Nothing)
+                                        case promotedHot of
+                                          Just t1 -> do
+                                            dt <- diffTime <$> getMonotonicTime <*> pure t1
+                                            traceWith spsTracer (PeerHotDuration connId dt)
+                                          Nothing -> pure ()
```
karknu, line 834: You're repeating this code in a few places. I think it could be worth turning it into a function.

crocodile-dentist:
I've decided to split out the refactoring of peer monitoring loop & friends to a separate commits, and merge just the tracing of connection durations into the release branch of o-n 0.22 so just these minor changes will go into 10.6.

ouroboros-network/src/Ouroboros/Network/PeerSelection/PeerStateActions.hs
crocodile-dentist: I've addressed this in #5167

PR #5157, opened 2025-07-11
Speed up `AF.intersect`

amesgen:
Just dumping this here to get this out of my mind for now.

When $n_1,n_2$ are the lengths of the two fragments, this PR speeds up `AF.intersect` from $O(n_2 \log(n_1))$ to $O(\log(n_1)\log(n_2))$ (where $n_2 \ge n_1$) via a binary search. I think it should also be possible to further speed this up to $O(\log(n_2) + \log(n_1)^2)$, but this doesn't seem that major.

This *seems* to also be faster in practice, eg the test
```
cabal run ouroboros-network-protocols:test:test -- -p 'AnchoredFragment.intersect' --quickcheck-tests=10000 --quickcheck-replay=702628
```
runs faster: 2.23s before this PR, 1.35s after this PR. In reality, I expect the speedup to be more significant (with fragments with 2160 or even 8640 headers), but this should actually be benchmarked.

### DIFF
--- a/ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
@@ -580,8 +580,8 @@ join :: HasHeader block
 join = AS.join $ \aOrB a ->
     either anchorToPoint blockPoint aOrB == anchorToPoint a
 
--- | \( O(n_2 \log(n_1)) \). Look for the most recent intersection of two
--- 'AnchoredFragment's @c1@ and @c2@.
+-- | \( O(\log(\min(n_1, n_2))\log(\max(n_1, n_2))) \). Look for the most recent
+-- intersection of two 'AnchoredFragment's @c1@ and @c2@.
 --
 -- The fragments need not have the same anchor point.
 --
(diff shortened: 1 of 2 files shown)


coot:
Nice improvement.

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
```diff
@@ -580,8 +580,8 @@ join :: HasHeader block
 join = AS.join $ \aOrB a ->
     either anchorToPoint blockPoint aOrB == anchorToPoint a
 
--- | \( O(n_2 \log(n_1)) \). Look for the most recent intersection of two
--- 'AnchoredFragment's @c1@ and @c2@.
+-- | \( O(\log(\min(n_1, n_2))\log(\max(n_1, n_2))) \). Look for the most recent
```
coot: why not just `O(log(n_1)log(n_2))`?

ouroboros-network-protocols/test/Test/AnchoredFragment.hs
```diff
@@ -295,21 +295,31 @@ prop_intersect :: TestAnchoredFragmentFork -> Property
 prop_intersect (TestAnchoredFragmentFork origP1 origP2 c1 c2) =
```
coot, line 295: Could you move these tests to `ouroboros-network-api`?

coot (approved):
LGTM

ouroboros-network-api/src/Ouroboros/Network/AnchoredFragment.hs
amesgen: Yes :sweat_smile: :facepalm:

ouroboros-network-protocols/test/Test/AnchoredFragment.hs
amesgen: You mean the entire module? Note that this means that a test suite of `ouroboros-network-api` will depend on `ouroboros-network-mock`/`ouroboros-network-protocols` (or even more stuff needs to be moved around).

amesgen: Ah, there already is this TODO: https://github.com/IntersectMBO/ouroboros-network/blob/556931f91f37554217b66222b66e054b283207ff/ouroboros-network-protocols/ouroboros-network-protocols.cabal#L212-L215

Would it be fine to do this in a separate PR?

amesgen: We just talked about this, and this doesn't have to happen in this PR; @coot also has some general restructuring plans.

PR #5161, opened 2025-07-15
Support weighted chain comparisons

amesgen:
With Ouroboros Peras, comparing chains now also depends on *certificates* which can give extra weight to certain blocks. Concretely, a particular candidate chain might first not be preferred to the current selection, but this can change when a particular Peras certificate arrives.

This PR adds support for this while keeping the exposure of Peras to the Network layer as minimal as possible in this area (in particular, Peras/certificates/weights are only mentioned in comments).

This PR is needed for https://github.com/tweag/cardano-peras/issues/62

### Dynamic chain comparisons

The first commit allows chain comparisons to be *dynamic*, ie change over time, modelling the arrival of new Peras certificates.

Previously, chain comparisons were modeled by the following two functions in the `BlockFetchConsensusInterface`:
```haskell
data BlockFetchConsensusInterface peer header block m = BlockFetchConsensusInterface {
    ...
    plausibleCandidateChain :: AnchoredFragment header -> AnchoredFragment header -> Bool,
    compareCandidateChains  :: AnchoredFragment header -> AnchoredFragment header -> Ordering
    ...
  }
```
This PR replaces this with
```haskell
  readChainComparison :: STM m (WithFingerprint (ChainComparison header))
```
where
```haskell
data ChainComparison header = ChainComparison {
    plausibleCandidateChain :: AnchoredFragment header -> AnchoredFragment header -> Bool,
    compareCandidateChains  :: AnchoredFragment header -> AnchoredFragment header -> Ordering
  }
```
Here, `WithFingerprint` is a [micro-abstraction copied from Consensus](https://ouroboros-consensus.cardano.intersectmbo.org/haddocks/ouroboros-consensus/Ouroboros-Consensus-Util-STM.html#t:WithFingerprint) used to indicate when new Peras certificates arrive (and therefore the results of the chain comparison functions might change), allowing the BlockFetch decision logic to be run again.

### Ensuring that arguments to `compareCandidateChains` intersect

With unweighted chain comparisons, `compareCandidateChains` [required](https://github.com/IntersectMBO/ouroboros-consensus/blob/252f4f0b7380ec666d0428e1275d40e474611cbf/ouroboros-consensus/src/ouroboros-consensus/Ouroboros/Consensus/Util/AnchoredFragment.hs#L79) that one of the following two criteria is true for its two fragment arguments:

 1. The fragments intersect.
 2. Both fragments are non-empty.

This latter criterion sufficed as we only need to get the block number and potentially further tiebreaker information (such as the VRF) from the tip header; the rest of the fragment is not relevant. However, this is no longer true for *weighted* chain comparisons: Here, we need the suffixes of both fragments after their common intersection.

The second commit ensures that this stronger precondition is met by enriching the `ChainSuffix` type by also storing the "full" candidate, each with the same anchor as the current chain, which directly yields that they must intersect pairwise.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
+++ b/ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
@@ -1,15 +1,22 @@
-{-# LANGUAGE DeriveAnyClass #-}
-{-# LANGUAGE DeriveFunctor  #-}
-{-# LANGUAGE DeriveGeneric  #-}
-{-# LANGUAGE LambdaCase     #-}
-{-# LANGUAGE RankNTypes     #-}
+{-# LANGUAGE DeriveAnyClass             #-}
+{-# LANGUAGE DeriveFunctor              #-}
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE RankNTypes                 #-}
 
 module Ouroboros.Network.BlockFetch.ConsensusInterface
   ( PraosFetchMode (..)
   , FetchMode (..)
   , BlockFetchConsensusInterface (..)
   , ChainSelStarvation (..)
+  , ChainComparison (..)
   , mkReadFetchMode
+    -- * Utilities
+  , WithFingerprint (..)
+  , Fingerprint (..)
+  , initialWithFingerprint
   ) where
 
 import Control.Monad.Class.MonadSTM
@@ -18,6 +25,7 @@ import Control.Monad.Class.MonadTime.SI (Time)
 import Data.Functor ((<&>))
 
 import Data.Map.Strict (Map)
+import Data.Word (Word64)
 import GHC.Generics (Generic)
 import GHC.Stack (HasCallStack)
 import NoThunks.Class (NoThunks)
@@ -129,24 +137,9 @@ data BlockFetchConsensusInterface peer header block m =
        -- have been downloaded anyway.
        readFetchedMaxSlotNo    :: STM m MaxSlotNo,
 
-       -- | Given the current chain, is the given chain plausible as a
-       -- candidate chain. Classically for Ouroboros this would simply
-       -- check if the candidate is strictly longer, but for Ouroboros
-       -- with operational key certificates there are also cases where
-       -- we would consider a chain of equal length to the current chain.
-       --
-       plausibleCandidateChain :: HasCallStack
-                               => AnchoredFragment header
-                               -> AnchoredFragment header -> Bool,
-
-       -- | Compare two candidate chains and return a preference ordering.
-       -- This is used as part of selecting which chains to prioritise for
-       -- downloading block bodies.
-       --
-       compareCandidateChains  :: HasCallStack
-                               => AnchoredFragment header
-                               -> AnchoredFragment header
-                               -> Ordering,
+       -- | Compare chain fragments. This might involve further state, such as
+       -- Peras certificates (which give certain blocks additional weight).
+       readChainComparison     :: STM m (WithFingerprint (ChainComparison header)),
 
        -- | Much of the logic for deciding which blocks to download from which
        -- peer depends on making estimates based on recent performance metrics.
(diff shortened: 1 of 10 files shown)


ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
```diff
@@ -199,6 +199,8 @@ data ChainComparison header =
        -- This is used as part of selecting which chains to prioritise for
        -- downloading block bodies.
        --
+       -- PRECONDITION: The two fragments must intersect.
```
amesgen, line 204: An alternative here would be to modify `compareCandidateChains` to also take the
current selection as an argument and require that the other two fragments
intersect with it (but don't necessarily intersect themselves). This would avoid modifying `ChainSuffix` in the rest of this commit, but I don't see a huge advantage in that, especially as it would also be slightly less efficient.

ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
```diff
@@ -184,3 +177,56 @@ data ChainSelStarvation
...
+  deriving stock (Show, Eq, Generic)
+  deriving newtype (Enum)
+  deriving anyclass (NoThunks)
+
+-- | Store a value together with its 'Fingerprint'.
+data WithFingerprint a = WithFingerprint
+  { forgetFingerprint :: !a
+  , getFingerprint    :: !Fingerprint
+  }
+  deriving stock (Show, Eq, Functor, Generic)
```
coot: Should it have a custom `Eq` instance which only compares `getFingerprint`s? (or even better only expose a function: `eqFingerprints :: WithFingerprint a -> WithFingerprint a -> Bool`)?

ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
```diff
@@ -184,3 +177,56 @@ data ChainSelStarvation
...
+  deriving anyclass (NoThunks)
+
+-- | Attach @'Fingerprint' 0@ to the given value.
+mkWithFingerprint :: a -> WithFingerprint a
+mkWithFingerprint a = WithFingerprint a (Fingerprint 0)
```
coot: It looks a bit dangerous to have this as an API.  Is it for testing?

What about this instead:
```hs
mkWithFingerprint :: (a -> Fingerprint) -> a -> WithFingerprint a
mkWithFingerprint getFingerprint a = WithFingerprint {
    forgetFingerprint = a,
    getFingerprint = getFingerprint a
  }
```

Then, for tests, one can use `mkWithFingerprint (const 0)`, and the real thing can compute the fingerprint.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -783,7 +800,7 @@ prioritisePeerChains FetchModeDeadline salt compareCandidateChains blockFetchSiz
       | EQ <- compareCandidateChains chain1 chain2 = True
       | otherwise                                  = False
 
-    chainHeadPoint (_,ChainSuffix c,_) = AF.headPoint c
+    chainHeadPoint (_,ChainSuffix c _,_) = AF.headPoint c
```
coot:
```suggestion
    chainHeadPoint (_,ChainSuffix {getChainSuffix = c},_) = AF.headPoint c
```

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Genesis.hs
```diff
@@ -457,13 +462,16 @@ selectTheCandidate
...
+            -- 'getFullCandidate's intersect pairwise (due to having the same
+            -- anchor as our current chain).
             let maxChainOn f c0 c1 = case compareCandidateChains (f c0) (f c1) of
                   LT -> c1
                   _  -> c0
                 -- maximumBy yields the last element in case of a tie while we
                 -- prefer the first one
                 chainSfx = fst $
-                  List.foldl1' (maxChainOn (getChainSuffix . fst)) inRace
+                  List.foldl1' (maxChainOn (getFullCandidate . fst)) inRace
```
coot, line 474: In the old version we're using `getChainSuffix`, in the new one `getFullCandidate`, is this intentional?

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
```diff
@@ -480,8 +473,19 @@ empty fetch range, but this is ok since we never request empty ranges.
...
-    ChainSuffix { getChainSuffix :: AnchoredFragment header }
+--
+-- Additionally, we store the full candidate (with the same anchor as our
+-- current chain), as this is needed for comparing different candidates via
+-- 'compareCandidateChains'.
+data ChainSuffix header = ChainSuffix {
+   -- | The suffix of the candidate after the intersection with the current
+   -- chain.
+   getChainSuffix   :: !(AnchoredFragment header)
+ , -- | The full candidate, characterized by having the same tip as
```
coot: According to our style-guide :grin::
```suggestion
   getChainSuffix   :: !(AnchoredFragment header),
   -- | The full candidate, characterized by having the same tip as
```

ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
amesgen: I simply removed the `Eq` instance :+1:

ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
amesgen: FTR one idea is that the fingerprint does not have to be computed from the data `a` that is stored in the `WithFingerprint`. Instead, suppose that the data is maintained in a `TVar`, and every time the `TVar` is updated, the `Fingerprint` will be increased. Does that make sense?

In this context, a better name for this function might be `initialWithFingerprint`.

I expanded the haddocks of `WithFingerprint` somewhat.

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision.hs
amesgen: Adopted (though it doesn't doesn't matter which of the two fragments we use, they always have the same tip)

ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Genesis.hs
amesgen: Yes, see the comment added here: https://github.com/IntersectMBO/ouroboros-network/blob/d62c9285a2bf87c8696c845c1aa2483046daf0ec/ouroboros-network/src/Ouroboros/Network/BlockFetch/Decision/Genesis.hs#L465-L467

ouroboros-network-api/src/Ouroboros/Network/BlockFetch/ConsensusInterface.hs
```diff
@@ -184,3 +177,58 @@ data ChainSelStarvation
...
+       --  * Tiebreakers (e.g. based on the opcert numbers and VRFs) for chains
+       --    of equal length.
+       --
+       --  * Weight in the context of Ouroboros Peras, due to a boost from a
+       --    Peras certificate.
+       --
+       plausibleCandidateChain :: HasCallStack
+                               => AnchoredFragment header
+                               -> AnchoredFragment header -> Bool,
+
```
coot: Let's fix code style
```suggestion
       plausibleCandidateChain :: HasCallStack
                               => AnchoredFragment header
                               -> AnchoredFragment header
                               -> Bool,

```

coot (approved):
Just a minor comments.

PR #5164, opened 2025-07-18
Default options for relays / BPs

coot:
# Description

- **Default values that depend on node role (BP/Relay)**
- **Revert "nix: use haskell.nix from PR#2390"**
- **nix: Updated hackageNix in flake.lock**

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network/src/Ouroboros/Network/Diffusion/Configuration.hs
+++ b/ouroboros-network/src/Ouroboros/Network/Diffusion/Configuration.hs
@@ -9,6 +9,7 @@ module Ouroboros.Network.Diffusion.Configuration
   , defaultDeadlineTargets
   , defaultDeadlineChurnInterval
   , defaultBulkChurnInterval
+  , BlockProducerOrRelay (..)
     -- re-exports
   , AcceptedConnectionsLimit (..)
   , DiffusionMode (..)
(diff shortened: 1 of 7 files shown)


ouroboros-network/cardano-diffusion/Cardano/Network/Diffusion/Configuration.hs
```diff
@@ -43,11 +45,13 @@ defaultNumBootstrapPeers = DefaultNumBootstrapPeers 30
...
-defaultSyncTargets :: PeerSelectionTargets
-defaultSyncTargets =
+defaultSyncTargets :: IsBlockProducer
+                   -- ^ block producer or relay node
+                   -> PeerSelectionTargets
+defaultSyncTargets bp =
   PeerSelectionTargets {
-    targetNumberOfRootPeers                 = 0,
-    targetNumberOfKnownPeers                = 150,
+    targetNumberOfRootPeers                 = case bp of { BlockProducer -> 100; Relay -> 60  },
```
crocodile-dentist: @karknu suggested we set this target to 0 such that we bias the active sync to non-big ledger peers.

ouroboros-network/src/Ouroboros/Network/Diffusion/Configuration.hs
```diff
@@ -72,10 +73,15 @@ defaultAcceptedConnectionsLimit =
     acceptedConnectionsSoftLimit = 384,
     acceptedConnectionsDelay     = 5 }
 
+
+data IsBlockProducer = BlockProducer | Relay
```
crocodile-dentist: Maybe a short haddock explaining that we eg. use this type to adjust peer selection targets based on node configuration  , allowing us to also reduce the number of configuration artifacts in the deployments.

ouroboros-network/cardano-diffusion/Cardano/Network/Diffusion/Configuration.hs
coot: For a BP, isn't it?

coot: Updated.

crocodile-dentist: I think it was 0 for edge node eg. daedalus or other such clients that occasionally start up and begin syncing (these sync targets are only applied when Genesis mode is enabled). Would a BP ever need to leverage public root peers/ledger peers in a real deployment where they are also syncing from them? I could imagine SPO's syncing their relays, and then copy the db to the block producer so these targets aren't really applicable to BP's and for consistency and simplicity when can also set this specific target to 0. 
[edit] for synthetic testing environment, this can be set to an arbitrary value anyway in the node config file so we don't lose anything.

coot: Oh, I updated the wrong targets :facepalm:  I should have updated the deadline targets for IntersectMBO/cardano-node#6286.

crocodile-dentist (approved):

PR #5184, opened 2025-08-08
tx-sub: added mempoolLastTicket to MempoolSnapshot

crocodile-dentist:
# Description

This is just to expose what consensus already tracks so we don't have to traverse the list of transactions to find the last tx index/ticket. This will also come in handy in dmq node to client code.

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network/testlib/Test/Ouroboros/Network/TxSubmission/Types.hs
+++ b/ouroboros-network/testlib/Test/Ouroboros/Network/TxSubmission/Types.hs
@@ -131,7 +131,8 @@ getMempoolReader (Mempool mempool) =
             \idx -> zipWith f [idx + 1 ..] (toList $ Seq.drop idx seq),
           -- why do I need to use `pred`?
           mempoolLookupTx   = flip Seq.lookup seq . pred,
-          mempoolHasTx      = \txid -> isJust $ find (\tx -> getTxId tx == txid) seq
+          mempoolHasTx      = \txid -> isJust $ find (\tx -> getTxId tx == txid) seq,
+          mempoolLastTicket = Just $ Seq.length seq
        }
 
     f :: Int -> Tx txid -> (txid, Int, SizeInBytes)
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -14,6 +14,7 @@
 * `Ouroboros.Network.TxSubmission.Inbound.V1.txSubmissionInbound` takes extra argument: `TxSubmissionInitDelay` (previously configurable through `cabal` flags).
 * Removed the `txsubmission-delay` cabal flag.
 * `ProtocolErrorRequestedTooManyTxids` includes number of unacked txids.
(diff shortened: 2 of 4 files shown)


ouroboros-network/testlib/Test/Ouroboros/Network/TxSubmission/Types.hs
```diff
@@ -131,7 +131,8 @@ getMempoolReader (Mempool mempool) =
             \idx -> zipWith f [idx + 1 ..] (toList $ Seq.drop idx seq),
           -- why do I need to use `pred`?
           mempoolLookupTx   = flip Seq.lookup seq . pred,
-          mempoolHasTx      = \txid -> isJust $ find (\tx -> getTxId tx == txid) seq
+          mempoolHasTx      = \txid -> isJust $ find (\tx -> getTxId tx == txid) seq,
+          mempoolLastTicket = Just $ Seq.length seq
```
coot, line 135: At least in `main` branch, we'll need to subtract `1`, as the indexes are `0`-based, but not in this branch. So this is not the reason why the test fails.

Should we actually return `Nothing` if the list is empty?
```suggestion
          mempoolLastTicket = if Seq.length seq > 0 then
                              then Seq.length seq - 1
                              else Nothing
```

crocodile-dentist: I've tried this and it didn't work. I've also added a counter to the mempool state because perhaps relying on the length is not correct, but that also didn't work. I will have to look at the logs more carefully as I've ran out of obvious simple tweaks.

crocodile-dentist: The problem is that this `mempoolLastTicket` moves the window too much, so we skip tx's. This doesn't matter for tx submission that much where we only ask for a few at a time, but the local dmq protocols will probably transmit more between the peers. We could change `mempoolTxIdsAfter` to `mempoolTxIdsSlice` to do what we want.

~As an aside, I've noticed that the tx submission server could be more eager in requesting tx id's. The server will often make a request for 1 txid when it could ask for all that the protocol allows. For eg, here is an example of an initial request:
`TraceTxInboundDecision (TxDecision {txdTxIdsToAcknowledge = NumTxIdsToAck 0, txdTxIdsToRequest = NumTxIdsToReq 1, txdPipelineTxIds = False, txdTxsToRequest = fromList [], txdTxsToMempool = []})`
It only asked for one, when it could have requested for the maximum unacknowledged limit.
Later, after receving the single lone tx, it will send another request for one tx id while acknowledging that prior tx id.
`TraceTxInboundDecision (TxDecision {txdTxIdsToAcknowledge = NumTxIdsToAck 1, txdTxIdsToRequest = NumTxIdsToReq 1, txdPipelineTxIds = False, txdTxsToRequest = fromList [], txdTxsToMempool [...]= `~
[edit] nevermind, this is just from the arbitrary instance for tx decision policy.

PR #5199, opened 2025-09-17
Update LedgerPeerSnapshot

crocodile-dentist:
# Description

- `LedgerPeerSnapshot` is extended to contain either all ledger peers or big ledger peers
- Added `Point` (slot w/ block hash) and network magic
- new snapshot encoding functions to move away from CBOR instances when NtC V22 support is dropped
- test updates

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/dmq-node/changelog.d/20251030_162643_crocodile-dentist_ledgerpeersnapshot_hash.md
+++ b/dmq-node/changelog.d/20251030_162643_crocodile-dentist_ledgerpeersnapshot_hash.md
@@ -1,0 +1,17 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+
+### Non-Breaking
+
+- Obtain all ledger peers via GetLedgerPeerSnapshot AllLedgerPeers
--- a/ouroboros-network/changelog.d/20251028_160648_crocodile-dentist_ledgerpeersnapshot_hash.md
+++ b/ouroboros-network/changelog.d/20251028_160648_crocodile-dentist_ledgerpeersnapshot_hash.md
@@ -1,0 +1,24 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+### Breaking
+
+ouroboros-network:api:
+- Added `LedgerBigPeerSnapshotV23` and `LedgerAllPeerSnapshotV23` constructors to `LedgerPeerSnapshot` type
+  to identify network magic and tip block hash when snapshot was recorded
+- removed compareLedgerPeerSnapshotApproximate
+
+### Non-Breaking
+
+ouroboros-network:api:
+- Added {To,From}JSON instances to `Point` and `Block`
+- added {encode,decode}LedgerPeerSnapshotPoint
+- added {encode,decode}StakePools
+
+ouroboros-network:
+- Removed cardano-slotting dependency
+- moved `jobVerifyPeerSnapshot` to cardano-diffusion
--- a/ouroboros-network/tests/lib/Test/Ouroboros/Network/LedgerPeers.hs
+++ b/ouroboros-network/tests/lib/Test/Ouroboros/Network/LedgerPeers.hs
@@ -566,7 +566,7 @@ prop_ledgerPeerSnapshotCBORV2 srvSupport slotNo
       _otherwise -> error "impossible"
 
 
--- TODO: move to `ouroboros-network-api:test`
+-- TODO: move to `ouroboros-network:api-tests`
 prop_ledgerPeerSnapshotCBORV3 :: SlotNo -> Word32 -> LedgerPools -> Bool -> Property
 prop_ledgerPeerSnapshotCBORV3 slotNo magic ledgerPools big =
   counterexample (show someSnapshot) $
--- a/cardano-diffusion/changelog.d/20251028_162450_crocodile-dentist_ledgerpeersnapshot_hash.md
+++ b/cardano-diffusion/changelog.d/20251028_162450_crocodile-dentist_ledgerpeersnapshot_hash.md
@@ -1,0 +1,16 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+### Breaking
+
+cardano-diffusion:
+- added `getBlockHash` to `LedgerPeersConsensusInterface`
+
+### Non-Breaking
+
+cardano-diffusion:
+- moved `jobVerifyPeerSnapshot` from ouroboros-network
--- a/cardano-ping/changelog.d/20251028_162800_crocodile-dentist_ledgerpeersnapshot_hash.md
+++ b/cardano-ping/changelog.d/20251028_162800_crocodile-dentist_ledgerpeersnapshot_hash.md
@@ -1,0 +1,11 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+
+### Breaking
+
+- Added `NodeToClientVersionV23`


cardano-diffusion/lib/Cardano/Network/LedgerPeerConsensusInterface.hs
```diff
@@ -31,4 +36,7 @@ data LedgerPeersConsensusInterface m =
     -- it only has local peers.
     --
   , updateOutboundConnectionsState :: OutboundConnectionsState -> STM m ()
+
+  , getBlockHash
+      :: forall r. SlotNo -> (forall a. STM m (Block SlotNo (Hash Blake2b_256 a)) -> r) -> r
```
coot: We should not instantiate to a concrete `Hash` type. Large part of consensus is parametric over `blk` type, e.g. `ouroboros-cosnensus-diffusion` is - and that's where we plug `cardano-diffusion`.

cardano-diffusion/lib/Cardano/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -675,6 +683,36 @@ waitForSystemToQuiesce st@PeerSelectionState{
   | otherwise = GuardedSkip Nothing
 
 
+-- |This job, which is initiated by monitorLedgerStateJudgement job,
+-- verifies whether the provided big ledger pools match up with the
+-- ledger state once the node catches up to the slot at which the
+-- snapshot was ostensibly taken
+--
+jobVerifyPeerSnapshot :: (MonadSTM m)
```
coot:
```suggestion
jobVerifyPeerSnapshot :: MonadSTM m
```

cardano-diffusion/lib/Cardano/Network/PeerSelection/Governor/Monitor.hs
```diff
@@ -675,6 +683,36 @@ waitForSystemToQuiesce st@PeerSelectionState{
...
+                      -> Job () m (Completion m extraState extraDebugState extraFlags extraPeers extraTrace peeraddr peerconn)
+jobVerifyPeerSnapshot (slotNo, theHash)
+                      Cardano.LedgerPeersConsensusInterface { getBlockHash }
+  = Job job (const (completion False)) () "jobVerifyPeerSnapshot"
+  where
+    completion result = return . Completion $ \st _now ->
+      Decision {
+        decisionTrace = [TraceVerifyPeerSnapshot result],
+        decisionState = st,
+        decisionJobs  = [] }
```
coot, line 707: We should log the error in the error handler rather than just a `False` value.
```suggestion
    errorHandler err = return . Completion $ \st _now ->
      Decision {
        decisionTrace = [TraceVerifyPeerSnapshotError err],
        decisionState = st,
        decisionJobs  = [] }
```
or ignore it altogether with:
```suggestion
    errorHandler _ = return . Completion $ \st _now ->
      Decision {
        decisionTrace = [],
        decisionState = st,
        decisionJobs  = [] }
```
The `getBlockHash` is an IO operation that could fail, but if it fails, we'll be in greater trouble than not being able to verify a hash in the snapshot.

dmq-node/changelog.d/20251030_162643_crocodile-dentist_ledgerpeersnapshot_hash.md
```diff
@@ -0,0 +1,17 @@
...
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+
+### Non-Breaking
+
+- update to new LedgerPeerSnapshot definition
```
coot: Could you be more explicit.

ouroboros-network/changelog.d/20251028_160648_crocodile-dentist_ledgerpeersnapshot_hash.md
```diff
@@ -0,0 +1,23 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+### Breaking
+
+o-n-api:
```
coot:
```suggestion
ouroboros-network:api
```

ouroboros-network/changelog.d/20251028_160648_crocodile-dentist_ledgerpeersnapshot_hash.md
```diff
@@ -0,0 +1,23 @@
...
+- Added tag `LedgerPeerSnapshotV3`
+- removed compareLedgerPeerSnapshotApproximate
+
+### Non-Breaking
+
+o-n-api:
```
coot:
```suggestion
ouroboros-network:api
```

ouroboros-network/changelog.d/20251028_160648_crocodile-dentist_ledgerpeersnapshot_hash.md
```diff
@@ -0,0 +1,23 @@
...
+- Added {To,From}JSON instances to `Point` and `Block`
+- added {encode,decode}LedgerPeerSnapshotPoint
+- added {encode,decode}StakePools
+
+o-n:
```
coot:
```suggestion
ouroboros-network
```

ouroboros-network/lib/Ouroboros/Network/PeerSelection/PeerMetric.hs
```diff
@@ -50,7 +50,7 @@ import GHC.Generics
 import NoThunks.Class
 import NoThunks.Class.Orphans ()
 
-import Cardano.Slotting.Slot (SlotNo (..))
+import Ouroboros.Network.Block (SlotNo (..))
```
coot, line 53: It doesn't seem necessary :smile:.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +28,383 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
-data LedgerPeerSnapshot =
-  LedgerPeerSnapshotV2 (WithOrigin SlotNo, [(AccPoolStake, (PoolStake, NonEmpty LedgerRelayAccessPoint))])
-  -- ^ Internal use for version 2, use pattern synonym for public API
-  deriving (Eq, Show)
+data LedgerPeerSnapshot (a :: LedgerPeersKind) where
+  LedgerPeerSnapshotV2
+    :: (WithOrigin SlotNo, [(AccPoolStake, (PoolStake, NonEmpty LedgerRelayAccessPoint))])
+    -> LedgerPeerSnapshot BigLedgerPeers
+  LedgerBigPeerSnapshotV23
+    :: !(Point (LedgerPeerSnapshot BigLedgerPeers))
```
coot: `Point (LedgerPeerSnapshot BigLedgerPeers)` is a bit strange thing, since `Point` points to some chain...,  what's the challenge with just a `Point blk`?

cardano-diffusion/lib/Cardano/Network/LedgerPeerConsensusInterface.hs
crocodile-dentist: I think the issue is that we parse the ledger peer snapshot at node startup and so I must instantiate to a specific type.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/PeerMetric.hs
crocodile-dentist: But it did allow me to drop `cardano-slotting` from ouroboros-network dependencies.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: I would make the argument that the snapshot is made at some Point on a chain. I think if we make it polymorphic here, the decision which concrete type to use will have to be made in the JSON parser, so we just move it there and have to deal with type parameters everywhere else.

cardano-diffusion/lib/Cardano/Network/LedgerPeerConsensusInterface.hs
crocodile-dentist: I think I see your point now - this should be doable here.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/PeerMetric.hs
coot: cool :).

cardano-diffusion/lib/Cardano/Network/LedgerPeerConsensusInterface.hs
```diff
@@ -31,4 +34,7 @@ data LedgerPeersConsensusInterface m =
     -- it only has local peers.
     --
   , updateOutboundConnectionsState :: OutboundConnectionsState -> STM m ()
+
+  , getBlockHash
+      :: forall r. SlotNo -> (forall blk. (StandardHash blk) => STM m (Point blk) -> r) -> r
```
coot: not needed brackets
```suggestion
      :: forall r. SlotNo -> (forall blk. StandardHash blk => STM m (Point blk) -> r) -> r
```

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet/Simulation.hs
```diff
@@ -1245,6 +1245,10 @@ diffusionSimulationM
                     Cardano.LedgerPeersConsensusInterface {
                       Cardano.readFetchMode = pure (PraosFetchMode FetchModeDeadline)
                     , Cardano.getLedgerStateJudgement = pure TooOld
+                    , Cardano.getBlockHash =
+                        let dummy :: forall r s. SlotNo -> (STM m (Point (LedgerPeerSnapshot s)) -> r) -> r
```
coot: break the line:
```suggestion
                        let dummy :: forall r s. SlotNo
                                  -> (STM m (Point (LedgerPeerSnapshot s)) -> r)
                                  -> r
```

cardano-diffusion/tests/lib/Test/Cardano/Network/PeerSelection.hs
```diff
@@ -4374,6 +4376,11 @@ _governorFindingPublicRoots targetNumberOfRootPeers readDomains readUseBootstrap
                     lpExtraAPI = Cardano.LedgerPeersConsensusInterface {
                       readFetchMode = pure (PraosFetchMode FetchModeDeadline),
                       getLedgerStateJudgement = readLedgerStateJudgement,
+                      getBlockHash =
+                        let dummy :: forall r kind.
+                                     SlotNo -> (STM IO (Point (LedgerPeerSnapshot kind)) -> r) -> r
```
coot:
```suggestion
                        let dummy :: forall r kind.
                                     SlotNo
                                  -> (STM IO (Point (LedgerPeerSnapshot kind)) -> r)
                                  -> r
```

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
coot: The node that parses the JSON will use the `blk` that it is instantiated with, so we won't need any conversions, or am I missing something?

crocodile-dentist: Isn't `blk` a parameter of the consensus layer? The parsing of this file happens earlier in the pipeline.

coot: I think `blk` is instantiated in `cardano-node`, so using `blk` should be doable - but we won't really know until we try.

crocodile-dentist: What do you think about my latest pushes here and just treat the hash directly as a `ShortByteString`? So the `Point` type will carry the block's slot number and it's hash directly as a SBS wrapped in a newtype.

coot: From `Cardano.Node.Run`
```hs
handleSimpleNode
  :: forall blk .
    ( Api.Protocol IO blk
    )
  => Api.BlockType blk
  -> Api.ProtocolInfoArgs blk
  -> Tracers RemoteAddress LocalAddress blk IO
  -> NodeConfiguration
  -> (NodeKernel IO RemoteAddress LocalConnectionId blk -> IO ())
  -- ^ Called on the 'NodeKernel' after creating it, but before the network
  -- layer is initialised.  This implies this function must not block,
  -- otherwise the node won't actually start.
  -> IO ()
```
`handleSimpleNode` has access to `blk`.  `blk` is instatiated by passing [`Api.BlockType blk`](https://github.com/IntersectMBO/ouroboros-consensus/blob/22773e62030913fe6cd974563b9bfe847f1ab493/ouroboros-consensus-cardano/src/unstable-cardano-tools/Cardano/Api/Protocol/Types.hs/#L134-L137).  Since parsing of the topology file happens within `handleSimpleNode`, we're fine.

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet/Simulation.hs
```diff
@@ -1245,6 +1245,7 @@ diffusionSimulationM
                     Cardano.LedgerPeersConsensusInterface {
                       Cardano.readFetchMode = pure (PraosFetchMode FetchModeDeadline)
                     , Cardano.getLedgerStateJudgement = pure TooOld
+                    , Cardano.getBlockHash = \_slotNo k -> k $ atomically retry
```
coot: if it's not used, it might be better to just `error "getBlockHash: not implemented"` than to deadlock.

cardano-diffusion/tests/lib/Test/Cardano/Network/PeerSelection/MockEnvironment.hs
```diff
@@ -485,6 +485,7 @@ mockPeerSelectionActions' tracer
           lpExtraAPI = Cardano.LedgerPeersConsensusInterface {
             readFetchMode = pure (PraosFetchMode FetchModeDeadline),
             getLedgerStateJudgement = readLedgerStateJudgement,
+            getBlockHash = \_slot k -> k $ atomically retry,
```
coot: as above

cardano-diffusion/tests/lib/Test/Cardano/Network/PeerSelection.hs
```diff
@@ -4374,6 +4375,7 @@ _governorFindingPublicRoots targetNumberOfRootPeers readDomains readUseBootstrap
                     lpExtraAPI = Cardano.LedgerPeersConsensusInterface {
                       readFetchMode = pure (PraosFetchMode FetchModeDeadline),
                       getLedgerStateJudgement = readLedgerStateJudgement,
+                      getBlockHash = \_slotNo k -> k $ atomically retry,
```
coot: as above

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,456 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+
+type instance HeaderHash SomeHashableBlock = SomeHashableBlock
+
+instance StandardHash SomeHashableBlock
+
+instance Eq SomeHashableBlock where
+  (==) (SomeHashableBlock (_ :: Proxy blk1) h1) (SomeHashableBlock (_ :: Proxy blk2) h2) =
+    case eqT :: Maybe (HeaderHash blk1 :~: HeaderHash blk2) of
+      Just Refl -> h1 == h2
+      Nothing   -> show h1 == show h2 -- nicer than error
```
coot: We should use a coercion function rather than `eqT` and then dealing with different types via a `Show` instance.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,456 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+
+instance Show SomeHashableBlock where
+  show (SomeHashableBlock _ sbh) = show sbh
+
+instance Ord SomeHashableBlock where
+  compare (SomeHashableBlock (_ :: Proxy blk1) h1) (SomeHashableBlock (_ :: Proxy blk2) h2) =
+    case eqT :: Maybe (blk1 :~: blk2) of
+      Just Refl -> compare h1 h2
+      Nothing   -> compare (show h1) (show h2)
```
coot: As above.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,456 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
                                    ]]
+  toJSON (LedgerAllPeerSnapshotV23 pt magic pools) =
+    object [ "NodeToClientVersion" .= (23 :: Int)
+           , "Point" .= toJSON pt
+           , "NetworkMagic" .= unNetworkMagic magic
+           , "allLedgerPools" .= [ object
+                                   [ "relativeStake" .= fromRational @Double relStake
+                                   , "relays"        .= relays]
+                                   | (PoolStake relStake, relays) <- pools
+                                   ]]
```
coot: let's make it a bit more obvious where each list ends...
```suggestion
           , "allLedgerPools" .= [ object
                                   [ "relativeStake" .= fromRational @Double relStake
                                   , "relays"        .= relays
                                   ]
                                 | (PoolStake relStake, relays) <- pools
                                 ]
           ]
```

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,456 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+  toJSON (LedgerBigPeerSnapshotV23 pt magic pools) =
+    object [ "NodeToClientVersion" .= (23 :: Int)
+           , "Point" .= toJSON pt
+           , "NetworkMagic" .= unNetworkMagic magic
+           , "bigLedgerPools" .= [ object
+                                   [ "accumulatedStake" .= fromRational @Double accStake
+                                   , "relativeStake"    .= fromRational @Double relStake
+                                   , "relays"           .= relays]
+                                   | (AccPoolStake accStake, (PoolStake relStake, relays)) <- pools
+                                   ]]
```
coot: as above, just to make it more readable
```suggestion
           , "bigLedgerPools" .= [ object
                                   [ "accumulatedStake" .= fromRational @Double accStake
                                   , "relativeStake"    .= fromRational @Double relStake
                                   , "relays"           .= relays
                                   ]
                                 | (AccPoolStake accStake, (PoolStake relStake, relays)) <- pools
                                 ]
           ]
```

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,456 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+
+instance ( StandardHash blk
+         , ToJSON (HeaderHash blk)
+         , FromJSON (HeaderHash blk)
+         , FromCBOR (HeaderHash blk)
+         , ToCBOR (HeaderHash blk)
```
coot, line 216: Could you explain why do we need `{From,To}CBOR` in `FromJSON` instance?  This looks suspicious 👀.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,456 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+       return . LedgerPeerSnapshotWithBlock $ LedgerAllPeerSnapshotV23 point (NetworkMagic magic) allPools'
+      _ ->
+        fail $ "Network.LedgerPeers.Type: parseJSON: failed to parse unsupported version "
+            <> show vNum
+
+instance ( StandardHash blk
+         , FromJSON (HeaderHash blk)
+         , ToJSON (HeaderHash blk)
+         , FromCBOR (HeaderHash blk)
+         , ToCBOR (HeaderHash blk)
```
coot, line 247: as above

ouroboros-network/changelog.d/20251028_160648_crocodile-dentist_ledgerpeersnapshot_hash.md
```diff
@@ -0,0 +1,24 @@
...
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+### Breaking
+
+ouroboros-network:api:
+- Added tags `LedgerBigPeerSnapshotV23` and `LedgerAllPeerSnapshotV23`
```
coot:
```suggestion
- Added `LedgerBigPeerSnapshotV23` and `LedgerAllPeerSnapshotV23` constructors to `LedgerPeerSnapshot` type
```

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -356,7 +357,7 @@ data PeerSelectionActions extraState extraFlags extraPeers extraAPI extraCounter
 
        -- | Read the current state of ledger peer snapshot
        --
-       readLedgerPeerSnapshot :: STM m (Maybe LedgerPeerSnapshot)
+       readLedgerPeerSnapshot :: STM m (Maybe (LedgerPeerSnapshot BigLedgerPeers))
```
coot, line 360: Shouldn't this be `LedgerPeerSnapshot AllLedgerPeers` (from it we can always recover `LedgerPeerSnapshot BigLedgerPeers`?

ouroboros-network/tests/lib/Test/Ouroboros/Network/LedgerPeers.hs
```diff
@@ -511,59 +530,111 @@ prop_ledgerPeerSnapshotCBORV2 :: LedgerPeerSnapshotSRVSupport
...
+                           _ -> True
+                        )
+                        relays
+                , not (null relays')
+                ]
+              )
+      _otherwise -> error "impossible"
+
+
+-- TODO: move to `ouroboros-network-api:test`
```
coot:
```suggestion
-- TODO: move to `ouroboros-network:api-tests`
```

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/Types.hs
crocodile-dentist: No, because the snapshot file is already limited to just the big ledger peers.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: Unfortunately, I need those for `SomeHashableBlock`, which hides the block type variable `blk` so we do not have to parameterize `ouroboros-network` with it. This type is also used to CBOR encode/decode the `Block` type, which includes the hash.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: I think the underlying type for the hash does not export the newtype data constructor, and coercion will not work.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,467 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+        bigPools' <- forM (zip [0 :: Int ..] bigPools) \(idx, poolO) -> do
+                       let f poolV = do
+                               accStake <- poolV .: "accumulatedStake"
+                               reStake  <- poolV .: "relativeStake"
+                               relays   <- poolV .: "relays"
+                               return (accStake, (reStake, relays))
+                       withObject ("bigLedgerPools[" <> show idx <> "]") f (Object poolO)
+
+        return . LedgerPeerSnapshotWithBlock $ LedgerPeerSnapshotV2 (slot, bigPools')
+      23 -> do
```
coot, line 279: When we write the version to the snapshot file, where do we take the `NodeToClientVersion` from?

Should we be more relaxed here to allow future versions, e.g. `n | n >= 23`?

coot: If the `NodeToClientVersion` comes from handshake negotiation done by `cardano-cli`, then the suggestion might be worth it. This will prevent parsing errors when a new `NodeToClientVersion` is introduced.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,467 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+         , FromCBOR (HeaderHash blk)
+         , ToCBOR (HeaderHash blk)
+         , Typeable blk
+         ) => FromJSON (LedgerPeerSnapshotWithBlock blk AllLedgerPeers) where
+  parseJSON = withObject "LedgerPeerSnapshot" \v -> do
+    -- TODO: remove "version" key after NtC V22 support is removed
+    vNum :: Int <- v .: "version" <|> v .: "NodeToClientVersion"
+    allPools    <- v .: "allLedgerPools"
+    case vNum of
+      23 -> do
```
coot, line 224: See the comment below.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
```diff
@@ -27,228 +31,467 @@ module Ouroboros.Network.PeerSelection.LedgerPeers.Type
...
+      2 -> Codec.decodeListLenOf 3 >>
+             SomeLedgerPeerSnapshot <$>
+             (LedgerBigPeerSnapshotV23 <$> decodeLedgerPeerSnapshotPoint proxy
+                                       <*> (NetworkMagic <$> Codec.decodeWord32)
+                                       <*> decodeBigStakePools)
+      3 -> Codec.decodeListLenOf 3 >>
+             SomeLedgerPeerSnapshot <$>
+             (LedgerAllPeerSnapshotV23 <$> decodeLedgerPeerSnapshotPoint proxy
+                                       <*> (NetworkMagic <$> Codec.decodeWord32)
+                                       <*> decodeAllStakePools)
```
coot, line 381: `3` or `23` as in the `FromJSON` instances?

coot: I think here `3` is ok, as it is an internal thing of this module, but it's nice for you @crocodile-dentist to confirm it.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: Yes, good catch.
[edit] Actually, the version number is hard coded. If we change the format in the future, we will just bump it here as well.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
crocodile-dentist: Yes, this is just internal.

ouroboros-network/api/lib/Ouroboros/Network/PeerSelection/LedgerPeers/Type.hs
coot: I am not concerned with that case, but when we introduce a new `NodeToClientVersion` for whatever reason, and then we will get parsing errors if we forget to update the codec here.  We have a similar situation with `cardano-ping` (the most recent one: #5281).

crocodile-dentist: I don't think api/cli serialises the node to client version, this is hardcoded to version 23 and that indicates the cbor serialisation format. We only need to bump this version if some future node to client version changes the cbor, otherwise it can stay at v23 even if we bump ntc version for other reasons.

coot: If we automatically set the version in cardano-cli, we won't need to update it when we change it - one less thing to avoid a mistake, especially easy when real integration is done by a random release engineer from another team.

PR #5201, opened 2025-09-24
Register PerasVoteDiffusion mini-protocol in `NodeToNode.hs`

tbagrel1:
Similarly to what has been done for PerasCertDiffusion, this PR plugs-in PerasVoteDiffusion as a new mini-protocol in NodeToNode.


tbagrel1:
I'm sorry, I've mistakenly created this PR as a normal one instead of a draft, but it is intended only for internal review by the Tweag Peras team at the moment.

PR #5202, opened 2025-09-29
Initial mini-protocols for Peras

amesgen:
# Description

### Add `NodeToNodeV_16` for Peras, and depend on `Cardano.Base.FeatureFlag`

We add a new (experimental) `NodeToNodeVersion` for Peras: `NodeToNodeV_16`.

Additionally, we add a function

```haskell
getLocalPerasSupport :: Set CardanoFeatureFlag -> NodeToNodeVersion -> PerasSupport
```

which additionally checks that the `PerasFlag` is set, cf https://github.com/IntersectMBO/cardano-base/pull/547. Eventually, once Peras is non-experimental, the `PerasFlag` will be removed again.

Peras enablement status is negociated through the field `perasSupport` in  `NodeToNodeVersionData`. The `decoder` for `NodeToNodeVersionData` when `version < v16` has been updated to set `perasSupport` to value `PerasUnsupported` (since this field is not part of the CBOR representation for earlier versions).

### Add generic ObjectDiffusion mini-protocol

This protocol is going to be used for Peras vote and certificate diffusion. It is essentially a copy of TxSubmission, except that `MsgInit` and `MsgDone` are sent by the client/inbound side, as the flow of information is the reverse of that of TxSubmission.

We also add codec roundtrip tests, CDDL tests, direct, connect, and channel-based tests.

 - `MsgInit`: currently pointless, but will get a payload indicating that we are only interested in objects after some indicator.

 - The representation of object ids currently always is a list of object ids. We are potentially going to make this customizable to optimize for cases where a more succinct representation is possible (such as requesting "ranges" of ids).

 - We might make some changes to how IDs are requested, such as allowing the server to reply with `MsgAwaitReply` when we do a blocking request, cf https://github.com/tweag/cardano-peras/issues/144.

### Register Peras vote and certificate diffusion mini-protocols

In `Cardano.Network.NodeToNode`, make the Network layer aware of Peras certificate and vote diffusion, gated behind `perasSupportStatus` from the negociated `NodeToNodeVersionData`.

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [x] Added labels.

### DIFF
--- a/cardano-diffusion/lib/Cardano/Network/NodeToNode.hs
+++ b/cardano-diffusion/lib/Cardano/Network/NodeToNode.hs
@@ -22,6 +22,8 @@ module Cardano.Network.NodeToNode
   , txSubmissionProtocolLimits
   , keepAliveProtocolLimits
   , peerSharingProtocolLimits
+  , perasCertDiffusionProtocolLimits
+  , perasVoteDiffusionProtocolLimits
   , defaultMiniProtocolParameters
   , NodeToNodeVersion (..)
   , NodeToNodeVersionData (..)
@@ -70,12 +72,15 @@ module Cardano.Network.NodeToNode
   , txSubmissionMiniProtocolNum
   , keepAliveMiniProtocolNum
   , peerSharingMiniProtocolNum
+  , perasCertDiffusionMiniProtocolNum
+  , perasVoteDiffusionMiniProtocolNum
   ) where
 
 import Control.DeepSeq (NFData)
 import Control.Exception (SomeException)
 
 import Data.ByteString.Lazy qualified as BL
+import Data.Set (Set)
 import Data.Word
 
 import Network.Mux qualified as Mx
@@ -85,6 +90,8 @@ import Network.Socket (Socket, StructLinger (..))
 import Network.Socket (Socket)
 #endif
 import Network.Socket qualified as Socket
+
+import Cardano.Base.FeatureFlags (CardanoFeatureFlag (..))
 
 import Cardano.Network.NodeToNode.Version
 import Cardano.Network.PeerSelection.PeerTrustable (PeerTrustable)
@@ -103,6 +110,8 @@ import Ouroboros.Network.PeerSelection.PeerAdvertise (PeerAdvertise (..))
 import Ouroboros.Network.PeerSelection.PeerSharing (PeerSharing (..))
 import Ouroboros.Network.Protocol.Handshake.Type
 import Ouroboros.Network.Protocol.Handshake.Version hiding (Accept)
+import Ouroboros.Network.Protocol.ObjectDiffusion.Type
+           (NumObjectsUnacknowledged)
 import Ouroboros.Network.Protocol.TxSubmission2.Type (NumTxIdsToAck (..))
 import Ouroboros.Network.Server.RateLimiting
 import Ouroboros.Network.SizeInBytes
@@ -115,25 +124,32 @@ import Ouroboros.Network.TxSubmission.Inbound.V2.Policy (TxDecisionPolicy (..),
 data NodeToNodeProtocols appType initiatorCtx responderCtx bytes m a b = NodeToNodeProtocols {
     -- | chain-sync mini-protocol
     --
-    chainSyncProtocol    :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
+    chainSyncProtocol          :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
 
     -- | block-fetch mini-protocol
     --
-    blockFetchProtocol   :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
+    blockFetchProtocol         :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
 
     -- | tx-submission mini-protocol
     --
-    txSubmissionProtocol :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
+    txSubmissionProtocol       :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
+
+    -- | Peras certificate diffusion mini-protocol
+    --
+    perasCertDiffusionProtocol :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
+
+    -- | Peras vote diffusion mini-protocol
+    --
+    perasVoteDiffusionProtocol :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
 
     -- | keep-alive mini-protocol
     --
-    keepAliveProtocol    :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
+    keepAliveProtocol          :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b,
 
     -- | peer sharing mini-protocol
     --
-    peerSharingProtocol  :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b
-
-  }
+    peerSharingProtocol        :: RunMiniProtocol appType initiatorCtx responderCtx bytes m a b
+}
 
 type NodeToNodeProtocolsWithExpandedCtx appType ntnAddr bytes m a b =
     NodeToNodeProtocols appType (ExpandedInitiatorContext ntnAddr PeerTrustable m) (ResponderContext ntnAddr) bytes m a b
@@ -142,11 +158,11 @@ type NodeToNodeProtocolsWithMinimalCtx  appType ntnAddr bytes m a b =
 
 
 data MiniProtocolParameters = MiniProtocolParameters {
-      chainSyncPipeliningHighMark :: !Word16,
+      chainSyncPipeliningHighMark                :: !Word16,
       -- ^ high threshold for pipelining (we will never exceed that many
       -- messages pipelined).
 
-      chainSyncPipeliningLowMark  :: !Word16,
+      chainSyncPipeliningLowMark                 :: !Word16,
       -- ^ low threshold: if we hit the 'chainSyncPipeliningHighMark' we will
       -- listen for responses until there are at most
       -- 'chainSyncPipeliningLowMark' pipelined message
@@ -156,19 +172,51 @@ data MiniProtocolParameters = MiniProtocolParameters {
       -- Note: 'chainSyncPipeliningLowMark' and 'chainSyncPipeliningLowMark'
       -- are passed to 'pipelineDecisionLowHighMark'.
 
-      blockFetchPipeliningMax     :: !Word16,
+      blockFetchPipeliningMax                    :: !Word16,
       -- ^ maximal number of pipelined messages in 'block-fetch' mini-protocol.
 
-      txDecisionPolicy            :: !TxDecisionPolicy
+      txDecisionPolicy                           :: !TxDecisionPolicy,
       -- ^ tx submission protocol decision logic parameters
+
+      perasCertDiffusionMaxObjectsUnacknowledged :: !NumObjectsUnacknowledged,
+      -- ^ Maximum number of PerasCerts in the outbound peer's outstanding FIFO.
+      --
+      -- This indirectly limits the number of pipelined requests from the inbound peer:
+      -- the inbound peer can only request @n@ new IDs if the execution of preceding
+      -- requests would result in at least @n@ empty seats in the FIFO.
+      --
+      -- In the worst case:
+      --
+      --   * The inbound peer requests IDs and objects one by one.
+      --   * The inbound peer is aware of @perasCertDiffusionMaxObjectsUnacknowledged@ IDs for objects
+      --     it hasn't requested yet (i.e., the FIFO is full).
+      --
+      -- Then, the inbound peer can pipeline at most @perasCertDiffusionMaxObjectsUnacknowledged@
+      -- requests for one object each (with a known ID), and up to
+      -- @perasCertDiffusionMaxObjectsUnacknowledged@ requests for one new ID each.
+      --
+      -- So, the theoretical maximum pipeline size is
+      -- @2 * perasCertDiffusionMaxObjectsUnacknowledged@, but in practice the pipeline size will
+      -- be much smaller, as the inbound peer typically batches requests.
+
+      perasVoteDiffusionMaxObjectsUnacknowledged :: !NumObjectsUnacknowledged
+      -- ^ Maximum number of PerasVotes in the outbound peer's outstanding FIFO.
+      -- See comment on 'perasCertDiffusionMaxObjectsUnacknowledged' for more details to
+      -- understand why this indirectly limits the number of pipelined requests.
     }
 
 defaultMiniProtocolParameters :: MiniProtocolParameters
 defaultMiniProtocolParameters = MiniProtocolParameters {
-      chainSyncPipeliningLowMark  = 200
-    , chainSyncPipeliningHighMark = 300
-    , blockFetchPipeliningMax     = 100
-    , txDecisionPolicy            = defaultTxDecisionPolicy
+      chainSyncPipeliningLowMark      = 200
+    , chainSyncPipeliningHighMark     = 300
+    , blockFetchPipeliningMax         = 100
+    , txDecisionPolicy                = defaultTxDecisionPolicy
+    -- | TODO: this value is still being discussed.
+    -- See https://github.com/tweag/cardano-peras/issues/97 for reference.
+    , perasCertDiffusionMaxObjectsUnacknowledged = 10
+    -- | TODO: this value is still being discussed.
+    -- See https://github.com/tweag/cardano-peras/issues/97 for reference.
+    , perasVoteDiffusionMaxObjectsUnacknowledged = 50
   }
 
 -- | Make an 'OuroborosApplication' for the bundle of mini-protocols that
@@ -190,15 +238,16 @@ defaultMiniProtocolParameters = MiniProtocolParameters {
 -- both protocols, e.g.  wireshark plugins.
 --
 nodeToNodeProtocols
-  :: MiniProtocolParameters
+  :: Set CardanoFeatureFlag
+  -> MiniProtocolParameters
   -> NodeToNodeProtocols muxMode initiatorCtx responderCtx bytes m a b
   -> NodeToNodeVersion
   -- ^ negotiated version number
   -> NodeToNodeVersionData
   -- ^ negotiated version data
   -> OuroborosBundle muxMode initiatorCtx responderCtx bytes m a b
-nodeToNodeProtocols miniProtocolParameters protocols
-                    _version NodeToNodeVersionData { peerSharing }
+nodeToNodeProtocols _featureFlags miniProtocolParameters protocols
+                    _version NodeToNodeVersionData { peerSharing, perasSupport }
                     =
     TemperatureBundle
       -- Hot protocols: 'chain-sync', 'block-fetch' and 'tx-submission'.
@@ -206,7 +255,9 @@ nodeToNodeProtocols miniProtocolParameters protocols
         case protocols of
           NodeToNodeProtocols { chainSyncProtocol,
                                 blockFetchProtocol,
-                                txSubmissionProtocol
+                                txSubmissionProtocol,
+                                perasCertDiffusionProtocol,
+                                perasVoteDiffusionProtocol
                               } ->
             [ MiniProtocol {
                 miniProtocolNum    = chainSyncMiniProtocolNum,
@@ -226,7 +277,25 @@ nodeToNodeProtocols miniProtocolParameters protocols
                 miniProtocolLimits = txSubmissionProtocolLimits miniProtocolParameters,
                 miniProtocolRun    = txSubmissionProtocol
               }
-            ])
+            ]
+              <> case perasSupport of
+                   PerasSupported   -> perasMiniProtocols
+                   PerasUnsupported -> mempty
+           where
+             perasMiniProtocols =
+               [ MiniProtocol {
+                   miniProtocolNum    = perasCertDiffusionMiniProtocolNum,
+                   miniProtocolStart  = StartOnDemand,
+                   miniProtocolLimits = perasCertDiffusionProtocolLimits miniProtocolParameters,
+                   miniProtocolRun    = perasCertDiffusionProtocol
+                 }
+               , MiniProtocol {
+                   miniProtocolNum    = perasVoteDiffusionMiniProtocolNum,
+                   miniProtocolStart  = StartOnDemand,
+                   miniProtocolLimits = perasVoteDiffusionProtocolLimits miniProtocolParameters,
+                   miniProtocolRun    = perasVoteDiffusionProtocol
+                 }
+               ])
 
       -- Warm protocols: reserved for 'tip-sample'.
       (WithWarm [])
@@ -262,7 +331,9 @@ chainSyncProtocolLimits
   , blockFetchProtocolLimits
   , txSubmissionProtocolLimits
   , keepAliveProtocolLimits
-  , peerSharingProtocolLimits :: MiniProtocolParameters -> MiniProtocolLimits
+  , peerSharingProtocolLimits
+  , perasCertDiffusionProtocolLimits
+  , perasVoteDiffusionProtocolLimits :: MiniProtocolParameters -> MiniProtocolLimits
 
 chainSyncProtocolLimits MiniProtocolParameters { chainSyncPipeliningHighMark } =
   MiniProtocolLimits {
@@ -381,6 +452,26 @@ peerSharingProtocolLimits _ =
   maximumIngressQueue = 4 * 1440
   }
 
+perasCertDiffusionProtocolLimits MiniProtocolParameters { perasCertDiffusionMaxObjectsUnacknowledged } =
+  MiniProtocolLimits {
+      -- The reasoning here is very similar to the 'txSubmissionProtocolLimits'.
+      --
+      -- Peras certificates will definitely be smaller than 20 kB; potentially
+      -- even much smaller.
+      -- See https://github.com/tweag/cardano-peras/issues/97
+      maximumIngressQueue = addSafetyMargin $
+        fromIntegral perasCertDiffusionMaxObjectsUnacknowledged * 20_000
+    }
+
+perasVoteDiffusionProtocolLimits MiniProtocolParameters { perasVoteDiffusionMaxObjectsUnacknowledged } =
+  MiniProtocolLimits {
+      -- Peras votes are expected to be much smaller than Peras certificates.
+      -- We assume an upper bound of 1 kB per vote.
+      -- See https://github.com/tweag/cardano-peras/issues/97
+      maximumIngressQueue = addSafetyMargin $
+        fromIntegral perasVoteDiffusionMaxObjectsUnacknowledged * 1_000
+    }
+
 chainSyncMiniProtocolNum :: MiniProtocolNum
 chainSyncMiniProtocolNum = MiniProtocolNum 2
 
@@ -395,6 +486,12 @@ keepAliveMiniProtocolNum = MiniProtocolNum 8
 
 peerSharingMiniProtocolNum :: MiniProtocolNum
 peerSharingMiniProtocolNum = MiniProtocolNum 10
+
+perasCertDiffusionMiniProtocolNum :: MiniProtocolNum
+perasCertDiffusionMiniProtocolNum = MiniProtocolNum 16
+
+perasVoteDiffusionMiniProtocolNum :: MiniProtocolNum
+perasVoteDiffusionMiniProtocolNum = MiniProtocolNum 17
 
 -- | A specialised version of @'Ouroboros.Network.Socket.connectToNode'@.
 --
--- a/cardano-diffusion/changelog.d/20250929_093633_alexander.esgen_peras_diffusion.md
+++ b/cardano-diffusion/changelog.d/20250929_093633_alexander.esgen_peras_diffusion.md
@@ -3,6 +3,11 @@
 - Added `NodeToNodeV_16` and updated `NodeToNodeVersionData` with a new `PerasSupport` field that is used only for `NodeToNodeV_16` version and later. This field indicates whether the node supports Peras protocols, and is negotiated during the version handshake.
   - CDDL specification for `NodeToNodeVersionData` has been updated in a backward-compatible way.
 
+- `Cardano.Network.NodeToNode`: Added support for certificate and vote diffusion mini-protocols for Ouroboros Peras.
+  - New fields in `NodeToNodeProtocols`.
+  - Protocol limits `perasCertDiffusionProtocolLimits`/`perasCertVoteProtocolLimits`.
+  - Mini-protocol numbers `perasCertDiffusionMiniProtocolNum`/`perasVoteDiffusionMiniProtocolNum`.
+
 ### Non-Breaking
 
 - `Cardano.Network.OrphanInstances`: Added support for `NodeToNodeV_16`.


ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
```diff
@@ -0,0 +1,298 @@
...
+  -- requested in any order, but at most once. Object identifiers are
+  -- acknowledged in the same FIFO order they were provided in. The
+  -- acknowledgement is included in the same messages used to ask for more
+  -- object identifiers.
+  data Message (ObjectDiffusion objectId object) from to where
+    -- | Initial message. The payload is currently unused; the planned use case
+    -- is to indicate that the inbound side is only interested to receive messages
+    -- newer than a given indicator.
+    MsgInit
+      :: Message (ObjectDiffusion objectId object) StInit StIdle
```
coot: What's the reason for `MsgInit` if in both `StInit` & `StIdle` state the agency is on the `client`/`inbound` side?

Note, in `TxSubmission2`, which this is modeled on, `MsgInit` transfers agency from `client` (outbound side) to `server` (inbound side).

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
```diff
@@ -0,0 +1,298 @@
...
+    -- In the 'StObjectIds' 'Blocking' state the list must be non-empty while in
+    -- the 'StObjectIds' 'NonBlocking' state the list may be empty.
+    --
+    -- These objects are added to the notional FIFO of outstanding object
+    -- identifiers for the protocol.
+    --
+    -- The order in which these object identifiers are returned must be the
+    -- order in which they are submitted to the mempool, to preserve dependent
+    -- objects.
+    MsgReplyObjectIds
```
coot, line 233: Adding a bit of spaces would be welcomed in this module.
```suggestion
    
    MsgReplyObjectIds
```

According to our style guide we use one or two empty lines (between top level definitions).

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
```diff
@@ -0,0 +1,298 @@
...
+
+-- | We have requests for lists of things. In the blocking case the
+-- corresponding reply must be non-empty, whereas in the non-blocking case
+-- and empty reply is fine.
+data BlockingReplyList (blocking :: StBlockingStyle) a where
+  BlockingReply    :: NonEmpty a -> BlockingReplyList StBlocking a
+  NonBlockingReply :: [a]        -> BlockingReplyList StNonBlocking a
+deriving instance (Eq a)   => Eq (BlockingReplyList blocking a)
+deriving instance (Show a) => Show (BlockingReplyList blocking a)
+deriving instance             Foldable (BlockingReplyList blocking)
```
coot: Oh, this type cannot be imported from `Ouroboros.Network.Protocol.TxSubmission2.Type`, since we'd need to also reuse the `StBlockingStyle`.

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
```diff
@@ -0,0 +1,298 @@
...
+instance StateTokenI StInit where stateToken = SingInit
+
+instance StateTokenI StIdle where stateToken = SingIdle
+
+instance (SingI stBlocking) => StateTokenI (StObjectIds stBlocking) where
+  stateToken = SingObjectIds sing
+
+instance StateTokenI StObjects where stateToken = SingObjects
+
+instance StateTokenI StDone where stateToken = SingDone
```
coot, line 106: Let's put these instances together, since they are all very similar 
```suggestion
instance StateTokenI StInit where stateToken = SingInit
instance StateTokenI StIdle where stateToken = SingIdle
instance SingI stBlocking
      => StateTokenI (StObjectIds stBlocking) where
  stateToken = SingObjectIds sing
instance StateTokenI StObjects where stateToken = SingObjects
instance StateTokenI StDone    where stateToken = SingDone
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Codec.hs
```diff
@@ -0,0 +1,287 @@
...
+    encode (MsgReplyObjectIds objIds) =
+         CBOR.encodeListLen 2
+      <> CBOR.encodeWord 2
+      <> CBOR.encodeListLenIndef
+      <> foldr (\objId r -> encodeObjectId objId <> r) CBOR.encodeBreak objIds'
+           where
+             objIds' :: [objectId]
+             objIds' = case objIds of
+               BlockingReply xs    -> NonEmpty.toList xs
+               NonBlockingReply xs -> xs
```
coot: We can use the `Foldable` instance of `BlockingReplyList` and `Monoid` instance of `Encode`:
```suggestion
      <> CBOR.encodeListLenIndef
      <> foldMap encodeObjectId objIds
      <> CBOR.encodeBreak
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Codec.hs
```diff
@@ -0,0 +1,287 @@
...
+    encode (MsgRequestObjects objIds) =
+         CBOR.encodeListLen 2
+      <> CBOR.encodeWord 3
+      <> CBOR.encodeListLenIndef
+      <> foldr (\objId r -> encodeObjectId objId <> r) CBOR.encodeBreak objIds
```
coot:
```suggestion
      <> CBOR.encodeListLenIndef
      <> foldMap encodeObjectId objIds
      <> CBOR.encodeBreak
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Codec.hs
```diff
@@ -0,0 +1,287 @@
...
+    encode (MsgReplyObjects objects) =
+         CBOR.encodeListLen 2
+      <> CBOR.encodeWord 4
+      <> CBOR.encodeListLenIndef
+      <> foldr (\objId r -> encodeObject objId <> r) CBOR.encodeBreak objects
```
coot:
```suggestion
      <> CBOR.encodeListLenIndef
      <> foldMap encodeObject objects
      <> CBOR.encodeBreak
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Codec.hs
```diff
@@ -0,0 +1,287 @@
...
+        (SingObjectIds SingNonBlocking, _, _) ->
+          fail $ printf "codecObjectDiffusion (%s) unexpected key (%d, %d)" (show stok) key len
+        (SingObjects, _, _) ->
+          fail $ printf "codecObjectDiffusion (%s) unexpected key (%d, %d)" (show stok) key len
+        (SingIdle, _, _) ->
+          fail $ printf "codecObjectDiffusion (%s) unexpected key (%d, %d)" (show stok) key len
+
+codecObjectDiffusionId
+  :: forall objectId object m.
+     (Monad m)
```
coot:
```suggestion
     Monad m
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Codec.hs
```diff
@@ -0,0 +1,287 @@
...
+       CodecFailure
+       m
+       (AnyMessage (ObjectDiffusion objectId object))
+codecObjectDiffusionId = Codec {encode, decode}
+  where
+    encode
+      :: forall st st'.
+         (ActiveState st
+         , StateTokenI st
+         )
```
coot, line 246: code style
```suggestion
      :: forall st st'.
         ( ActiveState st
         , StateTokenI st
         )
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Codec.hs
```diff
@@ -0,0 +1,287 @@
...
+      => Message (ObjectDiffusion objectId object) st st'
+      -> AnyMessage (ObjectDiffusion objectId object)
+    encode = AnyMessage
+
+    decode
+      :: forall (st :: ObjectDiffusion objectId object).
+         (ActiveState st)
```
coot:
```suggestion
         ActiveState st
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Inbound.hs
```diff
@@ -0,0 +1,116 @@
...
+        )
+        (inboundRun k)
+inboundRun (CollectPipelined none collect) =
+      Collect
+        (inboundRun <$> none)
+        (inboundRun . collect)
+inboundRun (SendMsgDone done) =
+      Yield MsgDone $ Done done
+inboundRun (WithEffect mNext) =
+      Effect $ inboundRun <$> mNext
```
coot: Could you move this function into the `where` clause of `objectDiffusionInboundPeerPipelined`.  Shorter name could also be more readable like `run` or `go`.  It's just an internal detail of its implementation.

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Outbound.hs
```diff
@@ -0,0 +1,111 @@
...
+-- * a request for a given list of objects
+-- * a termination message
+--
+-- It must be prepared to handle any of these.
+data OutboundStIdle objectId object m a = OutboundStIdle {
+      recvMsgRequestObjectIds :: forall blocking.
+                                  SingBlockingStyle blocking
+                              -> NumObjectIdsAck
+                              -> NumObjectIdsReq
+                              -> m (OutboundStObjectIds blocking objectId object m a),
```
coot, line 53: code style
```suggestion
      recvMsgRequestObjectIds :: forall blocking.
                                 SingBlockingStyle blocking
                              -> NumObjectIdsAck
                              -> NumObjectIdsReq
                              -> m (OutboundStObjectIds blocking objectId object m a),
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Outbound.hs
```diff
@@ -0,0 +1,111 @@
...
+              Yield
+                (MsgReplyObjectIds objectIds)
+                (outboundRun k)
+    MsgRequestObjects objectIds -> Effect $ do
+      SendMsgReplyObjects objects k <- recvMsgRequestObjects objectIds
+      return $
+        Yield
+          (MsgReplyObjects objects)
+          (outboundRun k)
+    MsgDone -> Effect $ Done <$> recvMsgDone
```
coot: As with the `inboudnRun`, it can be moved in a `where` clause of `objectDiffusionOtboundPeer`.

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Outbound.hs
```diff
@@ -0,0 +1,111 @@
...
+
+data OutboundStObjects objectId object m a where
+  SendMsgReplyObjects
+    :: [object]
+    -> OutboundStIdle objectId object m a
+    -> OutboundStObjects objectId object m a
+
+outboundRun
+  :: forall objectId object m a.
+     (Monad m)
```
coot:
```suggestion
     Monad m
```

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Outbound.hs
```diff
@@ -0,0 +1,111 @@
...
+      return $
+        Yield
+          (MsgReplyObjects objects)
+          (outboundRun k)
+    MsgDone -> Effect $ Done <$> recvMsgDone
+
+-- | A non-pipelined 'Peer' representing the 'ObjectDiffusionOutbound'.
+objectDiffusionOutboundPeer
+  :: forall objectId object m a.
+     (Monad m)
```
coot:
```suggestion
     Monad m
```

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
```diff
@@ -0,0 +1,266 @@
...
+{-# LANGUAGE GADTs                      #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PolyKinds                  #-}
+{-# LANGUAGE QuantifiedConstraints      #-}
+{-# LANGUAGE RankNTypes                 #-}
+{-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE TypeFamilies               #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+{-# LANGUAGE DeriveGeneric              #-}
```
coot: It should be moved to other `LANGUAGE` pragmas.

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Codec.hs
```diff
@@ -0,0 +1,287 @@
+{-# LANGUAGE DataKinds           #-}
```
coot, line 1: You should also add a CBOR spec for this codec, and test this implementation against it.  See `cardano-diffusion/protocols/cddl` (the `cardano-diffusion:cddl` component).

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
```diff
@@ -0,0 +1,266 @@
...
+--
+-- Test cases
+--
+
+
+tests :: TestTree
+tests =
+  testGroup "Ouroboros.Network.Protocol"
+    [ testGroup "ObjectDiffusion"
+        [ testProperty "codec"               prop_codec
```
coot: In the `TxSubmission2` tests we also have `prop_direct`, could you add it (including the `Ouroboros.Network.ObjectDiffusion.Direct` module)?

cardano-diffusion/protocols/cddl/specs/handshake-node-to-node-v14.cddl
```diff
@@ -15,7 +15,7 @@ msgQueryReply      = [3, versionTable]
 ; The codec only accepts definite-length maps.
 versionTable = { * versionNumber_v14 => v14.nodeToNodeVersionData }
 
-versionNumber_v14 = 14 / 15
+versionNumber_v14 = 14 / 15 / 16
```
coot: Could you add `v16` support for `cardano-ping`.

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
```diff
@@ -170,3 +182,8 @@ nodeToNodeCodecCBORTerm =
 
 
 data ConnectionMode = UnidirectionalMode | DuplexMode
+
+isPerasEnabled :: Set CardanoFeatureFlag -> NodeToNodeVersion -> Bool
+isPerasEnabled featureFlags v =
+       Set.member PerasFlag featureFlags
+    && v >= NodeToNodeV_16
```
coot: It is not sound to decide if Peras is enabled / disabled using configuration & version number.  One side might enable it, while the other one has it disabled, and the connection will error once one side sends their Peras traffic to the other side.

Pares must be negotiated between the two sides (without accessing local configuration, only based on `NodeToNodeVersionData` and `VersionData`).  You can add a boolean-like flag to `NodeToNodeVersionData` (just don't use another boolean :pray:) on which both sides agree.  Only nodes that are configured to use _Pares_ (I guess it will be behind a feature flag at the beginning) should set it to a true value in their advertised `NodeToNodeVersionData`.

Once `Pares` is production-ready, we can enable it based solely on a version, and the flag in `NodeToNodeVersionData` can be dropped.

You also need to make sure that after the flag in `NodeToNodeVersionData` is negotiated, only if it's true, then the mini-protocol is passed to the diffusion layer.  This can be done inside of `Ouroboros.Network.Diffusion.Applications` (i.e. inside the `Versions` which have access to the negotiated version data).

cardano-diffusion/lib/Cardano/Network/NodeToNode.hs
```diff
@@ -149,19 +164,51 @@ data MiniProtocolParameters = MiniProtocolParameters {
...
+      chainSyncPipeliningLowMark      = 200
+    , chainSyncPipeliningHighMark     = 300
+    , blockFetchPipeliningMax         = 100
+    , txDecisionPolicy                = defaultTxDecisionPolicy
+    -- | TODO: this value is still being discussed.
+    -- See https://github.com/tweag/cardano-peras/issues/97 for reference.
+    , perasCertDiffusionMaxFifoLength = 10
+    -- | TODO: this value is still being discussed.
+    -- See https://github.com/tweag/cardano-peras/issues/97 for reference.
+    , perasVoteDiffusionMaxFifoLength = 10_000
```
coot: This value seems excessive, why it is so large?

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
tbagrel1: It seems the `BlockingReplyList` type is defined here, not imported. So I'm not sure I understand what you mean by that _it cannot be imported from TXSubmission2_?

Is it a problem to reuse elements from TXSubmission2? I don't mind reimplementing the same types here, but I thought it was better to reuse existing code as much as possible.

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
tbagrel1: This implem follows closely the [design document](https://tweag.github.io/cardano-peras/peras-design.pdf#section.2.5.2) (section 2.5.2). Are we allowed to deviate from the design document and drop this init message then?

coot: There should be a process of updating it.  Recently, we were working on `dmq-node`, and during development, we changed an associated CIP that was already published.

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
coot: > I thought it was better to reuse existing code as much as possible.

That's what I wanted.  Giving it a second thought, it's just the `StBlockingStyle` that we'd share between the protocols, so it should be fine to do so. For us, the maintainers, it's less confusing when we have a single type for a given purpose.

cardano-diffusion/lib/Cardano/Network/NodeToNode.hs
tbagrel1: Do you have any idea on what could be a better value here?
We expect to see about 1000 different votes at each round of voting (~= the size of the voting committee), so a more sensible value for the FIFO would be somewhere between 10 and 1000 I guess? At the moment, we lack information to take a more principled decision. We will run benchmarks at a later point to have metrics on what is the observed FIFO occupation in practice.

cardano-diffusion/protocols/cddl/specs/handshake-node-to-node-v14.cddl
tbagrel1: @nbacquey could you link the related commit here once it has been added to `cardano-ping`? Thanks!

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
nbacquey: The reason `MsgInit` exists is also to eventually transmit a payload during the handshake. Typically, if the inbound peer doesn't start from Genesis, but from a later point in time, and only want to catch up with what it has missed, there will be some sort of `lastKnownPoint` payload to `MsgInit`.

cardano-diffusion/protocols/cddl/specs/handshake-node-to-node-v14.cddl
nbacquey: Done in a907a777e67123564d53416f13208f2f58404ae2

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
tbagrel1: I was working on that, but it appears that "direct" tests still need a client and server actual implementation. But in our case, for ObjectDiffusion, the client/server definition (via typed protocols) is in `ouroboros-network`, while the client/server implementation is in `ouroboros-consensus`.

With this code organisation, I don't think I can implement direct tests? Or I would need to define mock/testing client+server implems in `ouroboros-network`?

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
tbagrel1: Follow-up: https://github.com/IntersectMBO/ouroboros-network/pull/5202#issuecomment-3612171636

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
tbagrel1: FTR, I started working on direct tests in https://github.com/IntersectMBO/ouroboros-network/commit/61a3724252606cff639fc116aa0d2d89757bead6 , but I reverted it as I don't think we can have direct tests without an implementation of ObjectDiffusion in this repo

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
coot: Thanks, that makes sense.  Could you leave a note in the `MsgInit` haddocks and if you have an issue where it's planned/discussed please include it.

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
coot: You'll need to bring only the GADT wrappers to `ouroboros-network`; do not bring the actual client and server implementations.


The direct test is just a version of `connect`, but at the level of the GADT wrappers - no need for special implementation of `ObjectDiffusion` in this repo, besides some simple example client & server.  Check out how this is done for `tx-submission` or `chain-sync` (which production implementation lives in `ouroboros-consensus`).

coot: :thinking: 61a3724 is a good start.  In `tx-submission2` we have:

* [`txSubmissionServer`](https://github.com/intersectmbo/ouroboros-network/blob/227f9d7699281e3ff66d15dfc63b3a94d8517024/ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/TxSubmission2/Examples.hs#L215-L397)
* [`txSubmissionClient`](https://github.com/intersectmbo/ouroboros-network/blob/227f9d7699281e3ff66d15dfc63b3a94d8517024/ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/TxSubmission2/Examples.hs#L56-L153)

both are example implementations. Probably something very similar can be done in this case, so it's not as much work as it looks.

cardano-diffusion/lib/Cardano/Network/NodeToNode.hs
coot: Do I understand correctly that this is the size of the outstanding FIFO  which is mentioned [here](https://github.com/intersectmbo/ouroboros-network/blob/227f9d7699281e3ff66d15dfc63b3a94d8517024/ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs#L158-L164)?

_In `tx-submission` we call them unacknowledged txs - it would be good to use a similar language in the haddocs, since the concept is analogous._

If the answer is yes, then the buffer of unacknowledged objects can be much smaller than the voting round.  It shouldn't be used for storing all the votes, but just as a buffer which allows to request a number of votes, add them to the storage layer, acknowledge them and request more.  It's the storage layer that needs to hold all the 1000s votes.  In `tx-submission` a the unacknowledged buffer can hold `10` txs.

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Codec.hs
tbagrel1: Done in https://github.com/IntersectMBO/ouroboros-network/pull/5202/commits/d31da1777554d7ad81e3cea035073196cd309eaa :)

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Type.hs
tbagrel1: Done in https://github.com/IntersectMBO/ouroboros-network/pull/5202/commits/c1b89e9ed90e57f0c6b88432808d9dc89985a0c0 !

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
tbagrel1: As agreed in network team meeting of 2025/12/10, this PR will only include the `directPipelined` function (see https://github.com/IntersectMBO/ouroboros-network/pull/5202/commits/f86e3e4b68c07e1fdd25a240ead377078a0ac231), and the actual direct tests will be added in a future PR.

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
tbagrel1: Done in https://github.com/IntersectMBO/ouroboros-network/pull/5202/commits/5a03c9a3f3ea5c21e2e6544683267f3a1378dd6b

cardano-diffusion/lib/Cardano/Network/NodeToNode.hs
tbagrel1: Addressed in https://github.com/IntersectMBO/ouroboros-network/pull/5202/commits/e483e1739ea1949f7c431126f93d44158e9dd136 :)

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
agustinmista: Hey @coot! We started implementing direct tests for ObjectDiffusion (inspired by those in TxSubmission) on top of this PR on #5267. Most of the work is done but, as there's no easy way to terminate the outbound side, we are a bit stuck with no way to populate one of the blocking handlers. This will hopefully be easier to fix after we address https://github.com/tweag/cardano-peras/issues/144  soon.

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
```diff
@@ -123,50 +140,81 @@ instance Acceptable NodeToNodeVersionData where
...
+              []                   | version <  NodeToNodeV_16 -> pure PerasUnsupported
+              [CBOR.TBool perasSupportStatus] | version >= NodeToNodeV_16 -> pure $
+                if perasSupportStatus
+                  then PerasSupported
+                  else PerasUnsupported
+              l -> die $ "invalid encoding for perasSupportStatus given the version " <> show version <> ": " <> show l
+
+      other -> die $ "unexpected encoding when decoding NodeToNodeVersionData: " <> show other
+
+    die = Left . T.pack
```
coot: Let's use something more appropriate, e.g. `left`; `die` usually means we're throwing something (e.g. `System.Exit.die`) but we're only returning an error.

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
```diff
@@ -123,50 +140,81 @@ instance Acceptable NodeToNodeVersionData where
...
+            decodePeerSharing ps
+              | ps == 0   = pure PeerSharingDisabled
+              | ps == 1   = pure PeerSharingEnabled
+              | otherwise = die $ "peerSharing is out of bound: " <> show ps
+
+            decodeQuery = pure
+
+            decodePerasSupportStatusOptional = \case
+              []                   | version <  NodeToNodeV_16 -> pure PerasUnsupported
+              [CBOR.TBool perasSupportStatus] | version >= NodeToNodeV_16 -> pure $
```
coot: code style:
```suggestion
              []                              | version <  NodeToNodeV_16 -> pure PerasUnsupported
              [CBOR.TBool perasSupportStatus] | version >= NodeToNodeV_16 -> pure $
```

cardano-diffusion/api/tests/Test/Cardano/Network/NodeToNode/Version.hs
```diff
@@ -29,32 +33,47 @@ instance Arbitrary NodeToNodeVersion where
...
-    where
-      CodecCBORTerm { encodeTerm, decodeTerm } = nodeToNodeCodecCBORTerm ntnVersion
+        Right ntnData' -> ntnData' === ntnData
+        Left err       -> counterexample (show err) False
+  where
+    CodecCBORTerm { encodeTerm, decodeTerm } = nodeToNodeCodecCBORTerm ntnVersion
+
+prop_nodeToNodeCodecHandleInvalidData :: NodeToNodeVersion -> NodeToNodeVersionData -> Property
+prop_nodeToNodeCodecHandleInvalidData ntnVersion ntnData =
+    not (isValidNtnVersionDataForVersion ntnVersion ntnData) ==> ioProperty $ do
```
coot: It's better to use a generator that explores the spaces of `not (isValidNtnVersionDataForVersion ...)` rather than ignoring cases with `==>`.

cardano-diffusion/lib/Cardano/Network/NodeToNode.hs
```diff
@@ -219,7 +270,23 @@ nodeToNodeProtocols miniProtocolParameters protocols
                 miniProtocolLimits = txSubmissionProtocolLimits miniProtocolParameters,
                 miniProtocolRun    = txSubmissionProtocol
               }
-            ])
+            ]
+              <> concat [perasMiniProtocols | perasSupportStatus == PerasSupported]
```
coot:
```suggestion
              <> case perasSupportStatus of
                   PerasSupported   -> perasMiniProtocols
                   PerasUnsupported -> mempty
```

cardano-diffusion/protocols/cddl/Main.hs
```diff
@@ -395,14 +417,27 @@ validateCBOR (CDDLSpec spec) blob =
...
+      <*> arbitrary
+      <*> elements [ PerasUnsupported
+                   , PerasSupported
+                   ]
+
+genValidNtnVersionDataForVersion
+    :: NodeToNodeVersion
+    -> Gen NodeToNodeVersionData
+genValidNtnVersionDataForVersion version =
+       arbitrary `suchThat` (isValidNtnVersionDataForVersion version)
```
coot: Instead of throwing away generated data, we can fix it:
```suggestion
genValidNtnVersionDataForVersion version =
       fix version <$> arbitrary
  where
    fix version versionData | version < NodeToNodeV_16
                            = versionData { perasSupportsStatus = PerasUnsupported }
                            | otherwise
                            = versionData
```

cardano-diffusion/protocols/cddl/Main.hs
```diff
@@ -875,6 +913,21 @@ unit_decodeTxSubmission2 spec =
...
+    -> Assertion
+unit_decodeObjectDiffusion spec =
+    validateDecoder (Just indefiniteListFix)
+      spec objectDiffusionCodec
+      [ SomeAgency ObjectDiffusion.SingInit
+      , SomeAgency $ ObjectDiffusion.SingObjectIds ObjectDiffusion.SingBlocking
+      , SomeAgency $ ObjectDiffusion.SingObjectIds ObjectDiffusion.SingNonBlocking
+      , SomeAgency   ObjectDiffusion.SingObjects
+      , SomeAgency   ObjectDiffusion.SingIdle
+      ]
```
coot, line 909: code style:
```suggestion
      [ SomeAgency   ObjectDiffusion.SingInit
      , SomeAgency $ ObjectDiffusion.SingObjectIds ObjectDiffusion.SingBlocking
      , SomeAgency $ ObjectDiffusion.SingObjectIds ObjectDiffusion.SingNonBlocking
      , SomeAgency   ObjectDiffusion.SingObjects
      , SomeAgency   ObjectDiffusion.SingIdle
```

cardano-diffusion/protocols/tests-lib/Cardano/Network/Protocol/Handshake/Test.hs
```diff
@@ -174,8 +171,9 @@ instance Show ArbitraryNodeToNodeVersions where
 
 instance Arbitrary ArbitraryNodeToNodeVersions where
     arbitrary = do
+      let compatibleWith v = isValidNtnVersionDataForVersion v . getNodeToNodeVersionData
       vs <- listOf (getNodeToNodeVersion <$> arbitrary)
-      ds <- vectorOf (length vs) arbitrary
+      ds <- traverse (\v -> arbitrary `suchThat` compatibleWith v) vs
```
coot: Instead of throwing away generated result, it's better to provide a fixup function (as in the cddl test suite).

cardano-ping/src/Cardano/Network/Ping.hs
```diff
@@ -399,6 +404,8 @@ handshakeReqEnc versions query =
     encodeVersion (NodeToNodeVersionV12 magic mode) = encodeWithMode 12 magic mode
     encodeVersion (NodeToNodeVersionV13 magic mode _) = encodeWithMode 13 magic mode
     encodeVersion (NodeToNodeVersionV14 magic mode _) = encodeWithMode 14 magic mode
+    encodeVersion (NodeToNodeVersionV15 magic mode _) = encodeWithMode 15 magic mode
+    encodeVersion (NodeToNodeVersionV16 magic mode _) = encodeWithMode 16 magic mode
```
coot: For `V16` you need to support the `paresSupportStatus` field.  If you want you can always encode `False` (aka `ParesUnsupported`).

cardano-ping/src/Cardano/Network/Ping.hs
```diff
@@ -513,6 +520,8 @@ handshakeDec = do
         (12, False) -> decodeWithModeAndQuery NodeToNodeVersionV12
         (13, False) -> decodeWithModeQueryAndPeerSharing NodeToNodeVersionV13
         (14, False) -> decodeWithModeQueryAndPeerSharing NodeToNodeVersionV14
+        (15, False) -> decodeWithModeQueryAndPeerSharing NodeToNodeVersionV15
+        (16, False) -> decodeWithModeQueryAndPeerSharing NodeToNodeVersionV16
```
coot: For `V16` we need to support `paresSupportStatus` field, while`decodeWithModeQueryAndPerSharing` doesn't, we need a new function `decodeWithModeQueryPeerShairingAndPeras`.

ouroboros-network/api/lib/Ouroboros/Network/PerasSupport.hs
```diff
@@ -0,0 +1,16 @@
...
+-- then Peras protocol can be used.
+
+data PerasSupportStatus = PerasUnsupported | PerasSupported
+  deriving (Eq, Show)
+
+instance Ord PerasSupportStatus where
+  PerasUnsupported <= PerasUnsupported = True
+  PerasSupported <= PerasUnsupported   = False
+  PerasUnsupported <= PerasSupported   = True
+  PerasSupported <= PerasSupported     = True
```
coot: The instance here seems wrong to me but the derived on would be good:
```suggestion
data PerasSupportStatus = PerasUnsupported | PerasSupported
  deriving (Eq, Ord, Show)
```

We use `min` in [`Acceptable`](https://github.com/IntersectMBO/ouroboros-network/pull/5202/changes#diff-db808dd3ac63b1a326ca23961fc2096aad39911c46fbe7f58d7e04099857588eR127) instance, `PerasSuported` should only be used if both sides agree on it, not just one, isn't it?

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Direct.hs
```diff
@@ -0,0 +1,66 @@
...
+directPipelined (ObjectDiffusionOutbound mOutbound)
+                (ObjectDiffusionInboundPipelined inbound) = do
+    outbound <- mOutbound
+    directSender EmptyQ inbound outbound
+  where
+    directSender :: forall (n :: N).
+                    Queue          n (Collect objectId object)
+                 -> InboundStIdle n objectId object m a
+                 -> OutboundStIdle    objectId object m a
+                 -> m a
```
coot:
```suggestion
                    Queue         n (Collect objectId object)
                 -> InboundStIdle n objectId object m a
                 -> OutboundStIdle    objectId object m a
                 -> m a
```

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
```diff
@@ -0,0 +1,234 @@
...
+        , testProperty "codec id"            prop_codec_id
+        , testProperty "codec 2-splits"    $ withMaxSize 50
+                                             prop_codec_splits2
+        , testProperty "codec 3-splits"    $ withMaxSize 10
+                                             prop_codec_splits3
+        , testProperty "codec cbor"          prop_codec_cbor
+        , testProperty "codec valid cbor"    prop_codec_valid_cbor
```
coot: There are no tests that are using `connectPipelined` and by connecting client & server with channels (e.g. `createConnectedChannels`), see `Ouroboros.Network.Protocol.ChainSync.Test` for some examples.

ouroboros-network/tests/lib/Test/Ouroboros/Network/NodeToNode/Version.hs
```diff
@@ -1,5 +1,7 @@
-{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE NamedFieldPuns      #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
```
coot:
```suggestion
{-# LANGUAGE NamedFieldPuns      #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TypeApplications    #-}

{-# OPTIONS_GHC -Wno-orphans #-}
```

ouroboros-network/tests/lib/Test/Ouroboros/Network/NodeToNode/Version.hs
```diff
@@ -28,21 +32,38 @@ instance Arbitrary NodeToNodeVersion where
...
+    isValidNtnVersionDataForVersion ntnVersion ntnData ==>
+        case decodeTerm (encodeTerm ntnData) of
+            Right ntnData' -> ntnData' === ntnData
+            Left err       -> counterexample (show err) False
     where
       CodecCBORTerm { encodeTerm, decodeTerm } = nodeToNodeCodecCBORTerm ntnVersion
+
+prop_nodeToNodeCodecHandleInvalidData :: NodeToNodeVersion -> NodeToNodeVersionData -> Property
+prop_nodeToNodeCodecHandleInvalidData ntnVersion ntnData =
+    not (isValidNtnVersionDataForVersion ntnVersion ntnData) ==> ioProperty $ do
```
coot: As commented elsewhere, it's better to fix the generated data than throwing it away.

coot (changes requested):
Most of the review comments are benine.  One important one is the `Ord` instance for `PerasSupportStatus`.

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
```diff
@@ -93,27 +108,29 @@ nodeToNodeVersionCodec = CodecCBORTerm { encodeTerm, decodeTerm }
...
 data NodeToNodeVersionData = NodeToNodeVersionData
-  { networkMagic  :: !NetworkMagic
-  , diffusionMode :: !DiffusionMode
-  , peerSharing   :: !PeerSharing
-  , query         :: !Bool
+  { networkMagic       :: !NetworkMagic
+  , diffusionMode      :: !DiffusionMode
+  , peerSharing        :: !PeerSharing
+  , query              :: !Bool
+  , perasSupportStatus :: !PerasSupportStatus
```
coot: Could this be just called `perasSupport`, and the same for the type?
```suggestion
  , perasSupport :: !PerasSupport
```

ouroboros-network/api/lib/Ouroboros/Network/PerasSupport.hs
tbagrel1: I think the mental model here is that `PerasUnsupported` is morally 0, `PerasSupported` is morally 1, then the `<=` operator implem should follow that idea, and this way, we should be able to use `min` to get `PerasSupported` only when both operands are `PerasSupported`.

But maybe there is an error in the instance we wrote? (I'm not seeing it atm)

tbagrel1: I suppose the derived instance is doing exactly the same? Treating the variant as their index in the enum, and using these indices for comparisons?

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
tbagrel1: Done in bb05cce8b2a1556ceb474598ab02478509ff8635

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
tbagrel1: Done in 434234c746a8e167a9bc410c8848e21ea6f2ca40

cardano-diffusion/api/tests/Test/Cardano/Network/NodeToNode/Version.hs
tbagrel1: I've tried to improve on all of such cases in 8271ed4932c9996af7fa9511a6ce5fa4d7f4598e

It's a bit ugly, especially for the _invalid data_ case.

Indeed, we don't want the concern of knowing whether data is valid/invalid to leak outside of the lib file `cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs`.   But we can't introduce a proper generator of invalid cases in this lib file (because we don't depend on Quickcheck AFAIK).  
So instead, I've introduced the `fixNtnVersionDataForVersion` and `invalidateNtnVersionData` functions, so that the multiple test suites can share these.

Let us know what you think! We can always revert to the precondition-based test structure if you think it's preferable :)

cardano-diffusion/api/lib/Cardano/Network/NodeToNode/Version.hs
tbagrel1: Addressed with all other styling suggestions in 0fad0ef9e09bb74ca72a9fb30ebe6ed49aa2512e :)

ouroboros-network/api/lib/Ouroboros/Network/PerasSupport.hs
tbagrel1: FTR, I've changed for the derived one in 14575646c33f9538957ea9c122fa9df03a9f95cd

cardano-ping/src/Cardano/Network/Ping.hs
tbagrel1: Done in 63fbdd7484b9e6ace8e29bba569717759f489c34 :)

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
tbagrel1: Indeed, we don't have direct tests as part of this PR, but we have prepared some in https://github.com/IntersectMBO/ouroboros-network/pull/5267 (I think this is what you are asking us to implement?).

Unfortunately, we are still stuck on an issue in implementing the direct tests: we don't have a way for the inbound side (= client for ObjectDiffusion) to know it is caught up, so we can't exit gracefully when all objects have been exchanged. See https://github.com/IntersectMBO/ouroboros-network/pull/5267#discussion_r2614504221

Let us know if you want us to try to fix that and include the code as part of this PR instead of a separate one.

tbagrel1: Actually I found a solution to finish the direct tests, so I've just pushed it and merged https://github.com/IntersectMBO/ouroboros-network/pull/5267 into this PR branch. Lemme know if that is what you wanted :)

cardano-diffusion/api/tests/Test/Cardano/Network/NodeToNode/Version.hs
coot: The two functions:

* `fixNtnVersionDataForVersion`
* `invalidateNtnVersionData`

Do not belong to `ouroboros-netowrk:api`, but either `ouroboros-network:api-tests-lib` (if they are used in more than one test suite) or `ouroboros-network:api-tests`.

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Test.hs
coot: It's not just the direct test that is missing, but also other ones e.g. using `connect` and connecting via a channel.

cardano-diffusion/api/tests/Test/Cardano/Network/NodeToNode/Version.hs
tbagrel1: Hey, I've tried to extract the `fixNtnVersionDataForVersion` and `invalidateNtnVersionData` into `ouroboros-network:api-tests-lib` as you suggested, but I ran into a cyclic dependency issue:
- `NodeToNode{Version,VersionData}` are defined in `cardano-diffusion`, which depends on `ouroboros-network`
- so importing `NodeToNode{Version,VersionData}` into `ouroboros-network:api-tests-lib` to write the functions there creates a cycle `ouroboros-network` -> `cardano-diffusion` -> `ouroboros-network` since Cabal does package-wide resolution.

So instead, I've introduced a new sublib `cardano-diffusion:api-tests-lib` in commit [3811d30](https://github.com/IntersectMBO/ouroboros-network/commit/3811d304193a212f37b1510ede908ed3c764ffb7), with a single module `Cardano.Network.NodeToNode.Version.TestUtils` that contains the test helper functions; effectively breaking the cycle.

Since we can depend on `QuickCheck` now, I've replaced `invalidateNtnVersionData :: NodeToNodeVersionData -> [(NodeToNodeVersion, NodeToNodeVersionData)]` by the better-looking `genInvalidNtnVersionAndDataPair :: Gen (NodeToNodeVersion, NodeToNodeVersionData)`.

Since we now have a common place for test helpers for `NodeToNode`-related stuff, I've also made a couple of changes:
- define genNodeToNodeVersion{,Data} and shrinkNodeToNodeVersion{,Data} in this new module, and export them, so that we can avoid reimplementing the generators in each `instance Arbitrary NodeToNodeVersion{,Data}` in the various test modules.
- make sure the shrinking logic for `NodeToNodeVersionData` consistently shrinks one field at a time

It seems the main user of the shrinking logic was `ArbitraryNodeToNodeVersionData` in `cardano-diffusion/protocols/tests-lib/Cardano/Network/Protocol/Handshake/Test.hs`. Now it delegates to the new `shrinkNodeToNodeVersionData` function with the consistent shrinking logic. Let me know if that is a problem/if there was a reason for the previous non-uniform shrinking logic. I've also wondered why do we need the wrapper `ArbitraryNodeToNodeVersionData` around `NodeToNodeVersionData`, since we already disable warning for orphan instances in that module. But for now, I've left it as-is. Let me know if you think we can remove it.

These changes account for a bit more line diff than I expected, but I think since we introduce a common place for test helpers, it's a good opportunity to improve the repetitive test structure we had for `NodeToNodeVersion{,Data}`. We can of course revert that if you don't like this approach.

coot: > I've also wondered why do we need the wrapper ArbitraryNodeToNodeVersionData around NodeToNodeVersionData, since we already disable warning for orphan instances in that module. But for now, I've left it as-is. Let me know if you think we can remove it.

Probably because various people along the years had different preference for `Arbitrary` instances.

> These changes account for a bit more line diff than I expected, but I think since we introduce a common place for test helpers, it's a good opportunity to improve the repetitive test structure we had for NodeToNodeVersion{,Data}. We can of course revert that if you don't like this approach.

Thanks, I much appreciate it.

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Direct.hs
```diff
@@ -0,0 +1,467 @@
...
+
+
+objectDiffusionOutbound
+  :: forall objectId object m.
+     (Ord objectId, Show objectId, Monad m)
+  => Tracer m (TraceObjectDiffusionDirect objectId object)
+  -> (object -> objectId)
+  -> Word16  -- ^ Maximum number of unacknowledged object IDs allowed
+  -> [object]
+  -> ObjectDiffusionOutbound objectId object m ()
```
coot: These are only for testing purposes. I guess the client & server used in production will live in `ouroboros-consensus`, isn't it?  It's still important for them to be tested in a similar way as these implementations are.

ouroboros-network/protocols/tests-lib/Ouroboros/Network/Protocol/ObjectDiffusion/Direct.hs
```diff
@@ -0,0 +1,467 @@
...
+objectDiffusionInbound
+  :: forall objectId object m.
+     (Ord objectId, MonadCatch m, MonadEvaluate m)
+  => Tracer m (TraceObjectDiffusionDirect objectId object)
+  -> (object -> objectId)
+  -> Word16  -- ^ Maximum number of unacknowledged object IDs allowed
+  -> Word16  -- ^ Maximum number of object IDs to request in any one go
+  -> Word16  -- ^ Maximum number of objects to request in any one go
+  -> ObjectDiffusionInboundPipelined objectId object m [object]
+objectDiffusionInbound
```
coot: Have you considered drafting an API which would allow both `objectDiffusion{Inbound,Outbound}` to actually be used in production (so we can test it in rigorous ways here), and then the API needed to be run in production be defined else where (e.g. in consensus).  This would be a bit similar to `tx-submission`.  The [client](https://github.com/intersectmbo/ouroboros-network/blob/peras-diffusion/ouroboros-network/lib/Ouroboros/Network/TxSubmission/Outbound.hs#L78), [servers](https://github.com/intersectmbo/ouroboros-network/blob/peras-diffusion/ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs#L44) and the API ([mempool reader](https://github.com/intersectmbo/ouroboros-network/blob/peras-diffusion/ouroboros-network/lib/Ouroboros/Network/TxSubmission/Mempool/Reader.hs), [mempool writer](https://github.com/intersectmbo/ouroboros-network/blob/3595c9b74ca57667e096f27333ed137439e6f418/ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs#L358-L374)) are defined in `ouroboros-network`, but the production API is defined in `ouroboros-consensus`.

tbagrel1:
Hi @coot,

With https://github.com/IntersectMBO/ouroboros-network/commit/ad4cbb8a5e2006c704e9177dddacf91564ddc8a7, I introduced and fixed the `connectPipelined` and channel-based tests for ObjectDiffusion, in addition to the existing `directPipelined` ones.

I believe this was the last item requested in the latest review round, but please let me know if I missed anything.

Regarding comments  
https://github.com/IntersectMBO/ouroboros-network/pull/5202#discussion_r2932376350 and  
https://github.com/IntersectMBO/ouroboros-network/pull/5202#discussion_r2932448016:

We currently have a test client/server implementation in `ouroboros-network` and a production one in `ouroboros-consensus`. Both share the `typed-protocol`-based ObjectDiffusion spec defined in `ouroboros-network`.

We now have `direct`, `connect`, and channel-based tests for the test implementation in `ouroboros-network`, and smoke tests (equivalent to channel-based tests) for the production implementation in `ouroboros-consensus`.

Not sure if this fully answers your questions. In any case, we plan to join this week’s network meeting on Wednesday to discuss this live with you (if that works), so we can hopefully reach a mergeable state soon. If you agree, could you please send us an invite (cc @agustinmista @nbacquey )?

Thanks for your help and feedback.

cardano-diffusion/api/tests/Test/Cardano/Network/NodeToNode/Version.hs
```diff
@@ -18,43 +21,47 @@ tests :: TestTree
...
     ]
 
 instance Arbitrary NodeToNodeVersion where
-    arbitrary = arbitraryBoundedEnum
-
-    shrink v
-      | v == minBound = []
-      | otherwise     = [pred v]
+    arbitrary = genNodeToNodeVersion
+    shrink = shrinkNodeToNodeVersion
```
tbagrel1: Need some reformatting

cardano-diffusion/protocols/cddl/Main.hs
```diff
@@ -187,8 +197,10 @@ tests CDDLSpecs { cddlChainSync
       , testProperty "PeerSharing V14 to Last" (prop_encodePeerSharingV14ToLast
                                                   cddlPeerSharingNodeToNodeV14ToLast)
 
-      , testProperty "NodeToNodeVersionData V14 to Last" (prop_encodeNodeToNodeVersionDataV14ToLast
-                                                            cddlNodeToNodeVersionDataV14ToLast)
+      , testProperty "NodeToNodeVersionData V14 to V15" (prop_encodeNodeToNodeVersionDataV14ToV15
```
tbagrel1: alignement

cardano-diffusion/protocols/cddl/Main.hs
```diff
@@ -216,8 +230,10 @@ tests CDDLSpecs { cddlChainSync
       , testCase "PeerSharing V14 to Last" (unit_decodePeerSharingV14ToLast
                                                 cddlPeerSharingNodeToNodeV14ToLast)
 
-      , testCase "NodeToNodeVersionData V14 to Last" (unit_decodeNodeToNodeVersionDataV14ToLast
-                                                        cddlNodeToNodeVersionDataV14ToLast)
+      , testCase "NodeToNodeVersionData V14 to V15" (unit_decodeNodeToNodeVersionDataV14ToV15
+                                                        cddlNodeToNodeVersionDataV14ToV15)
```
tbagrel1, line 233: alignment

cardano-diffusion/protocols/cddl/Main.hs
```diff
@@ -240,7 +257,8 @@ data CDDLSpecs = CDDLSpecs {
 
     cddlPeerSharingNodeToNodeV14ToLast :: CDDLSpec (PeerSharing.PeerSharing SockAddr),
 
-    cddlNodeToNodeVersionDataV14ToLast :: CDDLSpec NodeToNodeVersionData
+    cddlNodeToNodeVersionDataV14ToV15 :: CDDLSpec NodeToNodeVersionData,
```
tbagrel1: alignment

cardano-diffusion/protocols/cddl/Main.hs
```diff
@@ -265,27 +283,31 @@ readCDDLSpecs = do
...
+    objectDiffusion       <- cddlc (dir </> "object-diffusion.cddl")
     keepAlive             <- cddlc (dir </> "keep-alive.cddl")
     localTxSubmission     <- cddlc (dir </> "local-tx-submission.cddl")
     localTxMonitor        <- cddlc (dir </> "local-tx-monitor.cddl")
     localStateQuery       <- cddlc (dir </> "local-state-query.cddl")
 
     peerSharingNodeToNodeV14ToLast <- cddlc (dir </> "peer-sharing-v14.cddl")
-    nodeToNodeVersionDataV14ToLast <- cddlc (dir </> "node-to-node-version-data-v14.cddl")
+    nodeToNodeVersionDataV14ToV15 <- cddlc (dir </> "node-to-node-version-data-v14.cddl")
+    nodeToNodeVersionDataV16ToLast <- cddlc (dir </> "node-to-node-version-data-v16.cddl")
```
tbagrel1, line 293: alignment

cardano-diffusion/protocols/cddl/Main.hs
```diff
@@ -265,27 +283,31 @@ readCDDLSpecs = do
...
+        cddlObjectDiffusion              = CDDLSpec objectDiffusion,
         cddlKeepAlive                    = CDDLSpec keepAlive,
         cddlLocalTxSubmission            = CDDLSpec localTxSubmission,
         cddlLocalTxMonitor               = CDDLSpec localTxMonitor,
         cddlLocalStateQuery              = CDDLSpec localStateQuery,
 
         cddlPeerSharingNodeToNodeV14ToLast = CDDLSpec peerSharingNodeToNodeV14ToLast,
-        cddlNodeToNodeVersionDataV14ToLast = CDDLSpec nodeToNodeVersionDataV14ToLast
+        cddlNodeToNodeVersionDataV14ToV15 = CDDLSpec nodeToNodeVersionDataV14ToV15,
+        cddlNodeToNodeVersionDataV16ToLast = CDDLSpec nodeToNodeVersionDataV16ToLast
```
tbagrel1, line 309: alignment

cardano-diffusion/protocols/cddl/Main.hs
```diff
@@ -395,14 +417,10 @@ validateCBOR (CDDLSpec spec) blob =
         Left err -> Left err
         Right _  -> Right ()
 
+instance Arbitrary NodeToNodeVersionData where
+    arbitrary = genNodeToNodeVersionData
+    shrink = shrinkNodeToNodeVersionData
```
tbagrel1: alignment

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Inbound.hs
```diff
@@ -0,0 +1,114 @@
...
+  SendMsgRequestObjectIdsBlocking
+    :: NumObjectIdsAck -- ^ number of objectIds to acknowledge
+    -> NumObjectIdsReq -- ^ number of objectIds to request
+    -> (NonEmpty objectId -> InboundStIdle Z objectId object m a)
+    -> InboundStIdle Z objectId object m a
+  SendMsgRequestObjectIdsPipelined
+    :: NumObjectIdsAck
+    -> NumObjectIdsReq
+    -> InboundStIdle (S n) objectId object m a
+    -> InboundStIdle n objectId object m a
```
tbagrel1: alignment

ouroboros-network/protocols/lib/Ouroboros/Network/Protocol/ObjectDiffusion/Inbound.hs
```diff
@@ -0,0 +1,114 @@
...
+  SendMsgRequestObjectsPipelined
+    :: [objectId]
+    -> InboundStIdle (S n) objectId object m a
+    -> InboundStIdle n objectId object m a
```
tbagrel1: alignment

coot (approved):
LGTM

PR #5232, opened 2025-10-24
peer selection: fixed a false positive tests failure

coot:
# Description

I stumbled upon a test failure in peer-selection test: https://ci.iog.io/build/9817878/nixlog/1

The governor has a target of 2 established peers, but there were 3 of them:

* one local root peer
* two public root peers

One of the public root peers was already being demoted, which was not handled
correctly by th test.  The patch that fixes this issue follows the logic of the
governor to subtract the number of demotions in progress from the number of
established peers when checking against the target.  Set comparison logic is
not enough, as in this case we could demote the other public root peer as well.

- **cardano-diffusion: added export list to Cardano.Network.PeerSelection.PublicRootPeers**
- **peer-selection: code alignement**
- **peer-selection: fixed a false positive tests**
- **peer-selection: addressed an old TODO**

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [x] Added labels.

### DIFF
--- a/cardano-diffusion/changelog.d/20251024_151019_coot_peer_selection_false_positive_test_failure.md
+++ b/cardano-diffusion/changelog.d/20251024_151019_coot_peer_selection_false_positive_test_failure.md
@@ -1,0 +1,6 @@
+### Non-Breaking
+
+- Added export list to `Cardano.Network.PeerSelection.PublicRootPeers`, also
+  re-exported `Ouroboros.Network.PeerSelection.PublicRootPeers` module.
+- Fixed a false positive test failure in peer selection test: `prop_governor_target_established_above`
+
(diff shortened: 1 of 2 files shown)


ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -1021,43 +1021,43 @@ peerSelectionCountersHWC PeerSelectionCounters {..} =
     PeerSelectionCounters {
       numberOfRootPeers,
 
-      numberOfKnownPeers                         = numberOfKnownPeers
-                                                 - numberOfEstablishedPeers,
+      numberOfKnownPeers                        = numberOfKnownPeers
+                                                - numberOfEstablishedPeers,
```
crocodile-dentist, line 1025: This could be confusing, perhaps rename to `numberOfFreeKnownPeers` and similar for ones below?

coot: This is an internal part of a pattern synonym, outside it's called `numberOfColdPeers`, check `PeerSelectionCountersHWC`.  `peerSelectionCountersHWC` is not exported, only used as a view pattern in the definition of `PeerSelectionCountersHWC`.

PR #5238, opened 2025-11-04
Enforce a min churn - ouroboros-network-0.22

### LINKED ISSUE #5237: Established peers can become a graveyard of rejected peers
If the number of churned established peers is less than the number of churned active peers the set of established peers will be filled with tried and rejected peers. This makes churning less efficient as the node risk selecting from mostly the same set of previously tried peers.

For example with the default config of 20 peers 4 active peers are churned out every hour, but only 2 established peers.

karknu:
# Description

If the number of established peers replaced is lower than the number of active peers replaced the set of established peers risk becoming a graveyard of already tried peers.

This change enforces a min number of changed established and known peers based on the number of replaced active and established peers.

Fixes #5237 .

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [x] Added labels.

### DIFF
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Revision history for ouroboros-network
 
-## 0.22.3.1 -- 04.11.2025
+## 0.22.4.0 -- 04.11.2025
 
 ### Non-breaking changes
 


ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -275,12 +280,13 @@ peerChurnGovernor
...
+    decreaseEstablishedBigLedgerPeers :: Int
+                                      -> PeerSelectionTargets
                                       -> ModifyPeerSelectionTargets
-    decreaseEstablishedBigLedgerPeers base targets =
+    decreaseEstablishedBigLedgerPeers minDecrease base targets =
       targets {
         targetNumberOfEstablishedBigLedgerPeers =
-          decrease (targetNumberOfEstablishedBigLedgerPeers base -
+          decreaseWithMin minDecrease (targetNumberOfEstablishedBigLedgerPeers base -
                     targetNumberOfActiveBigLedgerPeers base)
```
coot:
```suggestion
          decreaseWithMin minDecrease
                          (targetNumberOfEstablishedBigLedgerPeers base -
                    targetNumberOfActiveBigLedgerPeers base)
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -319,12 +327,13 @@ peerChurnGovernor
...
+      :: Int
+      -> PeerSelectionTargets
       -> ModifyPeerSelectionTargets
-    decreaseKnownBigLedgerPeers base targets =
+    decreaseKnownBigLedgerPeers minDecrease base targets =
       targets {
         targetNumberOfKnownBigLedgerPeers =
-          decrease (targetNumberOfKnownBigLedgerPeers base -
+          decreaseWithMin minDecrease (targetNumberOfKnownBigLedgerPeers base -
                     targetNumberOfEstablishedBigLedgerPeers base)
```
coot:
```suggestion
          decreaseWithMin minDecrease
                          (targetNumberOfKnownBigLedgerPeers base -
                    targetNumberOfEstablishedBigLedgerPeers base)
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/Churn.hs
```diff
@@ -497,4 +506,8 @@ peerChurnGovernor
 
     -- Replace 20% or at least one peer every churnInterval.
     decrease :: Int -> Int
-    decrease v = max 0 $ v  - max 1 (v `div` 5)
+    decrease = decreaseWithMin 1
+
+    -- Replace 20% or at least `u` or at least one peer every churnInterval.
+    decreaseWithMin :: Int -> Int -> Int
+    decreaseWithMin u v =  max 0 $ v - max u (max 1 (v `div` 5))
```
coot, line 515: If `20%` of active peers is more than all warm peers, we'll churn all warm peers plus some established.  A sane configuration should be such that `0.2 * active <= (established - active)`, an example of insane :wink: configuration: `55` active peers, `65` established, each turn we'd churn `11` established peers.

Does this deserve a warning?

karknu: We would actually churn 10 established peers. since `max 0` protects us from churning more than the number of established - active.

So every churn we'll replace all established peers. I am not sure it would warrant a warning

coot: You're right.

ouroboros-network/ouroboros-network.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 3.4
 name: ouroboros-network
-version: 0.22.3.0
+version: 0.22.3.1
```
coot:
```suggestion
version: 0.22.4.0
```
We usually use patch releases just for changes of package constraints.

coot (approved):

ouroboros-network/ouroboros-network.cabal
coot: According to [PVP](https://pvp.haskell.org/):

> 2.Non-breaking change. Otherwise, if only new bindings, types, classes, non-orphan instances or modules (but see below) were added to the interface, then A.B MAY remain the same but the new C MUST be greater than the old C. Note that modifying imports or depending on a newer version of another package may cause extra non-orphan instances to be exported and thus force a minor version change.
> 3. Other changes. Otherwise, e.g. if change consist only of corrected documentation, non-visible change to allow different dependency range etc. A.B.C MAY remain the same (other version components may change).

we should increase the version to `0.22.4.0`

ouroboros-network/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Revision history for ouroboros-network
 
+## 0.22.3.1 -- 04.11.2025
```
coot:
```suggestion
## 0.22.4.0 -- 04.11.2025
```

PR #5241, opened 2025-11-08
Peers behind a firewall

### LINKED ISSUE #4381: Optimised connectivity to peers behind firewall
Currently the P2P mode supports bidirectional connections (e.g. `InitiatorAndResponderDiffusionMode`), however if a remote peer which is in its configuration file (e.g. a local root peer) is behind a firewall, the outbound governor will try to poll until the remote peer connects.  The idea is to add a new peer category to the network topology file: peers which we do not connect to, but wait until they connect to us before we start use them as upstream peers.

the-headless-ghost:
# Description

Closes #4381

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [x] Added labels.


ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -471,6 +474,7 @@ jobPromoteColdPeer
   -> Job () m (Completion m extraState extraDebugState extraFlags extraPeers
                             extraTrace peeraddr peerconn)
 jobPromoteColdPeer PeerSelectionActions {
+                     readLocalRootPeers,
```
crocodile-dentist: We normally try to isolate reading tvar's from the environment in their own dedicated monitoring actions to preserve  invariants and isolate state changes to keep consistency. For instance, the localRoots action clamps them to the target of known peers iirc, and also to trustable only in certain scenarios.
I think an improved approach is to pass a flag to `jobPromoteColdPeer` that this peer is behind a firewall.
The connection manager's `Types` module exports `Provenance` which I think is clearer than a boolean, and we can avoid creating new data types.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -541,7 +545,21 @@ jobPromoteColdPeer PeerSelectionActions {
...
+            (\LocalRootConfig {localRootBehindFirewall} -> not localRootBehindFirewall)
+            (Map.lookup peeraddr m)
+
+      connectionMode <- atomically $
+        bool RequireInbound CreateNewIfNoInbound
+        -- ^ We require inbound if peer is behind a firewall
+        . foldr (\(_, _, m) b -> b && isNotBehindFirewall m) True
+        -- ^ True if peer is not behind a firewall
+        <$> readLocalRootPeers
+
```
crocodile-dentist: If the provenance value is passed to `jobPromoteColdPeer` we will not need this block.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -541,7 +545,21 @@ jobPromoteColdPeer PeerSelectionActions {
...
+      peerconn <- establishPeerConnection isBigLedgerPeer diffusionMode peeraddr connectionMode
```
crocodile-dentist: rename connectionMode to provenance.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/Types.hs
```diff
@@ -393,7 +394,9 @@ data PeerStateActions peeraddr peerconn m = PeerStateActions {
     --
     establishPeerConnection  :: IsBigLedgerPeer
                              -> DiffusionMode
-                             -> peeraddr -> m peerconn,
+                             -> peeraddr
+                             -> ConnectionMode
```
crocodile-dentist: Use `Provenance` here instead.

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1486,6 +1487,10 @@ with args@Arguments {
 
           -- connection manager does not have a connection with @peerAddr@.
           Right Nowhere -> do
+            -- Only proceed if creating a new connection is allowed
+            when (inboundRequired connectionMode)  $
+              throwIO (withCallStack $ InboundConnectionNotFound peerAddr)
+
```
crocodile-dentist: The damage is done by the time you do this check. A few lines above in the `Nothing` case we have updated the `stateVar` by inserting this outbound connection into the CM's connection table.

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -497,9 +499,23 @@ data Connected peerAddr handle handleError =
...
+-- conection.
+--  - 'RequireInbound': Strictly require an inbound connection; fail if none
+-- exists.
+data ConnectionMode
+    = CreateNewIfNoInbound
+    | RequireInbound
+
+inboundRequired :: ConnectionMode -> Bool
+inboundRequired RequireInbound = True
+inboundRequired _other         = False
```
crocodile-dentist: I don't think we need a new data type. We can use `Provenance` for our purposes.

ouroboros-network/lib/Ouroboros/Network/Diffusion/Topology.hs
```diff
@@ -32,7 +32,7 @@ newtype LocalRootPeersGroups extraFlags = LocalRootPeersGroups
 -- will attempt to maintain. By default this value will be equal to 'hotValency'.
 --
 data LocalRootPeersGroup extraFlags = LocalRootPeersGroup
-  { localRoots        :: RootConfig
+  { localRoots        :: LocalRoots
```
crocodile-dentist, line 36: I don't think we need a wrapper. Consider inserting a field `provenance :: Provenance`. Maybe this is also a better more general name than 'behindFirewall'. Our topology file can have this field, with a default value of Outbound, but if it is set explicitly to 'Inbound' then it will indicate that we have to piggy back on it to get that upstream as a local root. What do you think?

ouroboros-network/lib/Ouroboros/Network/Diffusion/Topology.hs
```diff
@@ -63,9 +69,16 @@ data RootConfig = RootConfig
...
-rootConfigToRelayAccessPoint RootConfig { rootAccessPoints, rootAdvertise  } =
+rootConfigToRelayAccessPoint RootConfig { rootAccessPoints, rootAdvertise } =
     [ (ap, rootAdvertise) | ap <- rootAccessPoints ]
 
+localRootsToRelayAccessPoint
+  :: LocalRoots
+  -> [(RelayAccessPoint, PeerAdvertise, Bool)]
+localRootsToRelayAccessPoint LocalRoots {rootConfig, behindFirewall} =
+      (\(accessPoint, advertise) -> (accessPoint, advertise, behindFirewall))
+  <$> rootConfigToRelayAccessPoint rootConfig
```
crocodile-dentist, line 80: I don't think it's needed.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/LocalRootPeers.hs
```diff
@@ -44,9 +44,10 @@ import Ouroboros.Network.PeerSelection.PeerAdvertise (PeerAdvertise)
 --
 
 data LocalRootConfig extraFlags = LocalRootConfig {
-    peerAdvertise       :: !PeerAdvertise,
-    diffusionMode       :: !DiffusionMode,
-    extraLocalRootFlags :: !extraFlags
+    peerAdvertise           :: !PeerAdvertise,
+    diffusionMode           :: !DiffusionMode,
+    localRootBehindFirewall :: !Bool,
```
crocodile-dentist: Change this type to `Provenance`?

cardano-diffusion/lib/Cardano/Network/Diffusion/Topology.hs
```diff
@@ -30,5 +30,6 @@ isValidTrustedPeerConfiguration
               IsTrustable    -> not
                               . null
                               . rootAccessPoints
+                              . rootConfig
```
crocodile-dentist, line 33: I don't think it's needed - see my other comments.

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -497,9 +499,23 @@ data Connected peerAddr handle handleError =
...
 
 type AcquireOutboundConnection peerAddr handle handleError m
-    = DiffusionMode -> peerAddr -> m (Connected peerAddr handle handleError)
+    = DiffusionMode -> peerAddr -> ConnectionMode -> m (Connected peerAddr handle handleError)
```
crocodile-dentist: s/ConnectionMode/Provenance/

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -723,6 +739,11 @@ data ConnectionManagerError peerAddr
     --
     | ForbiddenConnection   !(ConnectionId peerAddr) !CallStack
 
+    -- | No matching inbound connection found and creating new connection is
+    -- not allowed.
+    --
+    | InboundConnectionNotFound !peerAddr !CallStack
```
crocodile-dentist, line 724: This can be a typical case where an inbound is not found for behind firewall peers. Maybe instead create the opposite trace where we inform that we are establishing a connection back to a peer of 'inbound' provenance?

ouroboros-network/lib/Ouroboros/Network/Diffusion/Topology.hs
the-headless-ghost: The "wrapper" is here to separate it from public root peers. We do not want to specify the `behindFirewall` option for public root peers. Our goal is simply to indicate that the peer is unreachable. As noted below in the comment, a remote peer is always permitted to connect to you. This is not an either/or scenario.

crocodile-dentist: I have not looked too closely, but can we not just lift our behindFirewall field into the `LocalRootPeersGroup` record itself? This is what I meant above. If your point is still valid just let me know that and I'll look again to convince myself.

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -497,9 +499,24 @@ data Connected peerAddr handle handleError =
...
+--  - 'RequireInbound': Strictly require an inbound connection; fail if none
+-- exists.
+data ConnectionMode
+    = CreateNewIfNoInbound
+    | RequireInbound
+    deriving Show
+
+inboundRequired :: ConnectionMode -> Bool
+inboundRequired RequireInbound = True
+inboundRequired _other         = False
```
coot: This function is unnecessary indirection, we can directly pattern match on the constructors.

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1463,20 +1464,30 @@ with args@Arguments {
...
-                     )
+
+              -- Only proceed if creating a new connection is allowed
+              if inboundRequired connectionMode
+              then do
+                return ( Just (Right (TrInboundConnectionNotFound connectionMode peerAddr))
+                   , mutableConnState
+                   , Left (withCallStack
+                            (InboundConnectionNotFound connectionMode peerAddr))
+                   )
```
coot: code style:
```suggestion
                return ( Just (Right (TrInboundConnectionNotFound connectionMode peerAddr))
                       , mutableConnState
                       , Left (withCallStack
                                (InboundConnectionNotFound connectionMode peerAddr))
                       )
```

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1463,20 +1464,30 @@ with args@Arguments {
...
-                                      }))
-                     , mutableConnState
-                     , Right Nowhere
-                     )
+
+              -- Only proceed if creating a new connection is allowed
+              if inboundRequired connectionMode
+              then do
+                return ( Just (Right (TrInboundConnectionNotFound connectionMode peerAddr))
+                   , mutableConnState
```
coot: There's not point in returning `mutableConnState` (this branch of `if` should not executed `newMutableConnState` as it must not be used, which requries a bit more refactoring.

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Core.hs
```diff
@@ -1463,20 +1464,30 @@ with args@Arguments {
...
+                   , Left (withCallStack
+                            (InboundConnectionNotFound connectionMode peerAddr))
```
coot: I was trying to find out if throwing an exception in this case is right or should we return an explicit value.  After a deliberation, I think it's fine to do so (and thus throw an exception).  The outbound governor is not supposed to call `establishPeerConnection` unless there already is an inbound connection for such a peer.  So if we find `InboundConnectionNotFound` in the logs, it's actually a signal for us to debug it.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/LocalRootPeers.hs
coot: This will also avoid boolean blindness :smile:, but it still should be a `Boolean` value in json.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/Types.hs
coot: I'd swap the arguments, so the connection configuration goes together
```suggestion
                             -> ConnectionMode
                             -> peeraddr
```

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -723,6 +740,11 @@ data ConnectionManagerError peerAddr
     --
     | ForbiddenConnection   !(ConnectionId peerAddr) !CallStack
 
+    -- | No matching inbound connection found and creating new connection is
+    -- not allowed.
+    --
```
coot:
```suggestion
    -- | No matching inbound connection found while establishing a new connection is
    -- not allowed.
    --
```

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -864,6 +877,19 @@ aboveTargetBigLedgerPeers actions@PeerSelectionActions {
...
+    )
+  => peeraddr
+  -> LocalRootPeers.LocalRootPeers extraFlags peeraddr
+  -> ConnectionMode
+connectionMode peer localRootPeers =
+      maybe
+        CreateNewIfNoInbound
+        (\LocalRootConfig {localRootBehindFirewall} ->
+           bool CreateNewIfNoInbound RequireInbound localRootBehindFirewall)
+        (Map.lookup peer $ LocalRootPeers.toMap localRootPeers)
```
coot: When you change `ConnectionMode` to `Provenance`, this function should be renamed to `localRootConnectionProvenance`.

ouroboros-network/lib/Ouroboros/Network/Diffusion/Topology.hs
coot: I like the `LocalRoots` wrapper; it's more explicit than hiding things in deserialization (where the default value needs to be set) and then not using the value for public roots.  Once public roots are gone, we can remove the indirection.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -79,6 +79,9 @@ data KnownPeers peeraddr = KnownPeers {
        -- establish a connection to now. This is because we have not connected
        -- with them before or because any failure backoff time has expired.
        --
+       -- Note: Some peers may be behind a firewall. Local root peers marked as
+       -- behind a firewall are not excluded from this list.
+       --
```
coot: Thanks for the note.

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet.hs
```diff
@@ -2846,6 +2851,96 @@ prop_diffusion_target_established_local_iosim
...
+          govUnreachablePeersSig :: Signal (Set NtNAddr)
+          govUnreachablePeersSig =
+            (\local ->
+               let
+                 isUnreachablePeer (LocalRootConfig {localRootBehindFirewall}) =
+                   localRootBehindFirewall
+                 unreachablePeers =
+                   Map.keysSet
+                     $ Map.filter isUnreachablePeer
+                     $ LocalRootPeers.toMap local
```
coot:
```suggestion
                 unreachablePeers =
                   Map.keysSet
                     $ Map.filter localRootBehindFirewall
                     $ LocalRootPeers.toMap local
```

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet.hs
```diff
@@ -2846,6 +2851,96 @@ prop_diffusion_target_established_local_iosim
...
+                )
+            $ events
+
+      in counterexample
+          ("\nSignal key: (local root peers, unreachables, inbounds") $
+        signalProperty 100 show
+          (\(_,unreachables,inbounds) -> unreachables `Set.disjoint` inbounds)
+          ((,,) <$> govLocalRootPeersSig
+                   <*> govUnreachablePeersSig
+                   <*> govInboundConnectionsSig)
```
coot: code style
```suggestion
          ((,,) <$> govLocalRootPeersSig
                <*> govUnreachablePeersSig
                <*> govInboundConnectionsSig)
```

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet.hs
```diff
@@ -2846,6 +2851,96 @@ prop_diffusion_target_established_local_iosim
...
+                 unreachablePeers =
+                   Map.keysSet
+                     $ Map.filter isUnreachablePeer
+                     $ LocalRootPeers.toMap local
+               in
+                 unreachablePeers
+            ) <$> govLocalRootPeersSig
+
+          govInboundConnectionsSig :: Signal (Set NtNAddr)
+          govInboundConnectionsSig =
```
coot:
```suggestion
          -- all connections which triggered `TrConnectionNotFound Inbound`
          govNotFoundInboundConnSig :: Signal (Set NtNAddr)
          govNotFoundInboundConnSig =
```

ouroboros-network/orphan-instances/Ouroboros/Network/OrphanInstances.hs
```diff
@@ -97,6 +97,21 @@ kindObject k fields = object $ ("kind" .= String k) : fields
 
 -- FromJSON Instances
 
+instance FromJSON LocalRoots where
+  parseJSON = withObject "LocalRoots" $ \o ->
+                LocalRoots
+                  <$> (RootConfig
+                        <$> o .:  "accessPoints"
+                        <*> o .:? "advertise" .!= DoNotAdvertisePeer)
```
coot: Don't we have a `FromJSON` instance for `RootConfig` that can be used here?

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Core.hs
crocodile-dentist: There could be a race condition here, but maybe throwing an exception is still a convenient way of dealing with it anyway.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -282,7 +288,10 @@ belowTargetOther actions@PeerSelectionActions {
                           inProgressPromoteCold = inProgressPromoteCold
                                                <> selectedToPromote
                         },
-        decisionJobs  = [ jobPromoteColdPeer actions policy peer IsNotBigLedgerPeer InitiatorAndResponderDiffusionMode
+        decisionJobs  = [ jobPromoteColdPeer
+                            actions policy peer IsNotBigLedgerPeer
+                            InitiatorAndResponderDiffusionMode
+                            (localRootConnectionProvenance peer localRootPeers)
```
crocodile-dentist: I think this should just be `Outbound`

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -413,7 +423,9 @@ belowTargetBigLedgerPeers enableAction
                           inProgressPromoteCold = inProgressPromoteCold
                                                <> selectedToPromote
                         },
-        decisionJobs  = [ jobPromoteColdPeer actions policy peer IsBigLedgerPeer InitiatorAndResponderDiffusionMode
+        decisionJobs  = [ jobPromoteColdPeer actions policy peer IsBigLedgerPeer
+                          InitiatorAndResponderDiffusionMode
+                          (localRootConnectionProvenance peer localRootPeers)
```
crocodile-dentist: also just `Outbound`?

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
```diff
@@ -864,6 +877,19 @@ aboveTargetBigLedgerPeers actions@PeerSelectionActions {
...
+    )
+  => peeraddr
+  -> LocalRootPeers.LocalRootPeers extraFlags peeraddr
+  -> Provenance
+localRootConnectionProvenance peer localRootPeers =
+      maybe
+        Outbound
+        (\LocalRootConfig {localRootBehindFirewall} ->
+           bool Outbound Inbound localRootBehindFirewall)
+        (Map.lookup peer $ LocalRootPeers.toMap localRootPeers)
```
crocodile-dentist: If this is to be used only in one place then we can make this a local binding.

ouroboros-network/lib/Ouroboros/Network/Diffusion/Topology.hs
```diff
@@ -44,6 +44,12 @@ newtype PublicRootPeers = PublicRootPeers
   { publicRoots :: RootConfig
   } deriving (Eq, Show)
 
+data LocalRoots = LocalRoots
+  { rootConfig     :: RootConfig
+  , behindFirewall :: Bool
```
crocodile-dentist: s/behindFirewall/demandInboundProvenance/ ? This way we explicitly state the actual precondition in the name itself - someone might want to use this in this way and perhaps we don't have to muddy the waters with notions of firewalls and whatnot? Or at least a haddock here what it actually does.

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -497,9 +497,14 @@ data Connected peerAddr handle handleError =
     --
   | Disconnected !(ConnectionId peerAddr) !(DisconnectionException handleError)
 
-
+-- | Acquire outbound connection
+-- 'Inbound' provenance restricts acquiring an outbound connection to those
+-- initiated by a remote peer.
+-- 'Outbound' provenance allows acquiring an outbound connection regardless of
+-- who initiated it.
```
crocodile-dentist: Strictly speaking this is misleading since one cannot acquire an outbound for which there already is an outbound.
Maybe instead the wording should be:
'Outbound' provenance doesn't require any additional pre-conditions.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
the-headless-ghost: Yes, I think you are right.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/Governor/EstablishedPeers.hs
the-headless-ghost: Agree, after the fix this will only be used in `belowTargetLocal`.

ouroboros-network/lib/Ouroboros/Network/Diffusion/Topology.hs
the-headless-ghost: I agree, but Im somewhat hesitant because I dont know how we are naming it in the config file. Should we update it as well? I guess It should be something that the SPOs understand.

ouroboros-network/framework/lib/Ouroboros/Network/ConnectionManager/Types.hs
```diff
@@ -723,6 +727,11 @@ data ConnectionManagerError peerAddr
     --
     | ForbiddenConnection   !(ConnectionId peerAddr) !CallStack
 
+    -- | No matching inbound connection found while establishing a new connection is
+    -- not allowed.
+    --
+    | InboundConnectionNotFound !Provenance !peerAddr !CallStack
```
coot: Isn't the provenance always `Outbound`?

the-headless-ghost: Right, we can remove the provenance here. It is always `Inbound`.

PR #5265, opened 2025-12-11
More robust diffusion (ouroboros-network-0.21)

karknu:
# Description

This is a series a changes that makes the node more robust.

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [x] Added labels.

### DIFF
--- a/ouroboros-network/CHANGELOG.md
+++ b/ouroboros-network/CHANGELOG.md
@@ -6,6 +6,13 @@
 
 ### Non-breaking changes
 
+* Limit the number of faulures to 5 before a peer that isn't a localroot, bootstrap peer or public root peer is forgotten.
+* Decrease the time blockfetch waits for chainsync to exit in case of an error
+* Increase the timeout for chainsync in state StMustReply to between 601 and 911 seconds.
+* Ensure timeout to enter safe mode when enabling bootstrap peers is respected
+* bugfix async demotion
+* Add hot connection duration traces and sigusr1 debug handler
+
 ## 0.21.4.0 -- 2025-10-05
 
 ### Non-breaking changes


ouroboros-network/src/Ouroboros/Network/PeerSelection/Governor/ActivePeers.hs
```diff
@@ -481,20 +486,23 @@ jobPromoteWarmPeer PeerSelectionActions{peerStateActions = PeerStateActions {act
...
-                                                 else
+                          (knownPeers', publicRootPeers') =
+                              if peeraddr `KnownPeers.member` knownPeers
+                                 then KnownPeers.reportFailures
+                                           now
+                                           policyMaxConnectionRetries
+                                           (Set.singleton peeraddr)
+                                           (\p -> LocalRootPeers.member p localRootPeers ||
+                                            (memberExtraPeers p (PublicRootPeers.getExtraPeers publicRootPeers)))
+                                           (const . const delay)
```
coot:
```suggestion
                                           (\_ _ -> delay)
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -405,30 +425,101 @@ minConnectTime KnownPeers { nextConnectTimes } fn =
...
-    let knownPeers' = knownPeers {
+reportFailures :: Ord peeraddr
+               => Time
+               -- ^ current time
+               -> Int
+               -- ^ Max number of failures before a peer is forgotten
+               -> Set peeraddr
+               -- ^ peers with failure
+               -> (peeraddr -> Bool)
+               -- ^ do we have to remember the peer?
```
coot, line 444:
```suggestion
               -- ^ do we have to remember the fail count for a peer?
```

could you also add which peers are rememberd, e.g. local roots & extra root peers aka bootstrap peers.

### unrelated rant

This is another reason why I think bootstrap peers should actually be part of `ouroboros-network` rather than `cardano-diffusion`, e.g. it's awkward for us do this part in `cardano-diffusion` - which would be the proper way in the current split between `ouroboros-network` and `cardano-diffusion`.

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -405,30 +421,101 @@ minConnectTime KnownPeers { nextConnectTimes } fn =
...
+               -> (peeraddr -> Int -> DiffTime)
+               -- ^ calculate delay from failure count
+               -> (Set peeraddr -> a)
+               -- ^ callback for forgotten peers
+               -> KnownPeers peeraddr
+               -> (KnownPeers peeraddr, a)
```
coot: It seems the callback is only used on the returned value so we can leave that to the caller and just return the forgotten peers.
```suggestion
               -> KnownPeers peeraddr
               -> (KnownPeers peeraddr, Set peeraddr)
```

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -405,30 +421,101 @@ minConnectTime KnownPeers { nextConnectTimes } fn =
...
-                -> KnownPeers peeraddr
-setConnectTimes times
-                knownPeers@KnownPeers {
-                  allPeers,
-                  availableToConnect,
-                  nextConnectTimes
-                } =
-    assert (all (`Map.member` allPeers) (Map.keysSet times)) $
-    let knownPeers' = knownPeers {
+reportFailures :: Ord peeraddr
```
coot, line 436: Maybe a more explicit name would be `setConnectTimesAndFailCount`

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
karknu: I am not committing the suggestion. The purpose of the function is to control if we can forget the peer or not. Not simply if we need to track its fail count.

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
karknu: The real description would be `setConnectionTimesAndFailCountAndPossiblyForgetPeers` ;)

I prefer `reportFailures` to that, but I've added a comment to the function to make it clear what it is and what it does.

ouroboros-network/src/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
karknu: :+1: Great suggestion, will make that change

PR #5276, opened 2026-01-02
Allow using `QuickCheck < 2.16`

geo2a:
This change is necessary for Consensus to build with the current Network

### DIFF
--- a/ouroboros-network/ouroboros-network.cabal
+++ b/ouroboros-network/ouroboros-network.cabal
@@ -55,9 +55,16 @@ common ghc-options
 -- in tests librararies redundant constraints are sometimes useful (e.g.
 -- by default truned off debug tracing might require extra constraints like
 -- `Show` or `MonadSay`).
+--
+-- TODO: '-Wno-redundant-constraints' remove when we support
+-- `quickcheck-monoids`. For now, it is needed to prevent
+-- false-positives that lead to build errors with `-Werror`.
+-- See https://gitlab.haskell.org/ghc/ghc/-/issues/24173 for details.
 common ghc-options-tests
   import: ghc-options
-  ghc-options: -Wno-redundant-constraints
+  ghc-options:
+    -Wno-redundant-constraints
+    -Wno-unused-packages
 
 library api
   import: ghc-options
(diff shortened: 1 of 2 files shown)


ouroboros-network/ouroboros-network.cabal
```diff
@@ -57,7 +57,9 @@ common ghc-options
 -- `Show` or `MonadSay`).
 common ghc-options-tests
   import: ghc-options
-  ghc-options: -Wno-redundant-constraints
+  ghc-options:
+    -Wno-redundant-constraints
+    -Wno-unused-packages
```
coot, line 67: Could you add a comment, that this is only required until we support `quickcheck-monoids`.

cardano-diffusion/tests/lib/Test/Cardano/Network/PeerSelection.hs
```diff
@@ -88,7 +88,9 @@ import Cardano.Network.PeerSelection.Governor.Types qualified as Cardano.ExtraSi
 import Cardano.Network.PeerSelection.State.LocalRootPeers qualified as LocalRootPeers
 import Ouroboros.Network.BlockFetch (FetchMode (..), PraosFetchMode (..))
 import Test.QuickCheck
-import Test.QuickCheck.Monoids
+#if !MIN_VERSION_QuickCheck(2,16,0)
+import "quickcheck-monoids" Test.QuickCheck.Monoids
+#endif
```
coot, line 93: You can squash this on top of the first commit, which contain similar changes.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -10,7 +10,7 @@
 {-# LANGUAGE TypeApplications    #-}
 {-# LANGUAGE TypeOperators       #-}
 
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP                 #-}
```
coot: Remove the empty line, sort the pragmas and squash it onto the commit where it was introduced.

PR #5307, opened 2026-02-09
RNG handling in SRV lookups for public root peers

Savissy:
**Motivation**:

The public root peers provider used the same StdGen seed for all DNS lookups, causing SRV weighted-sampling to reuse the same RNG and produce correlated/identical selections across domains.
The change aims to provide independent RNGs per domain lookup so SRV selection becomes varied and correct.

**Description:**

Add a mutable RNG StrictTVar (rngVar) initialized from the supplied StdGen and thread it into the request path for publicRootPeersProvider by changing the local requestPublicRootPeers signature to accept the RNG var.
Generate one split RNG per domain with replicateM + stateTVar rngVar split and zip those RNGs with the domain list, then pass each domain-specific RNG into dnsLookupWithTTL so lookups (including srvRecordLookupWithTTL) use independent RNGs.
Import replicateM from Control.Monad and create rngVar via newTVarIO near the resolver resource initialization.

### DIFF
--- a/ouroboros-network/lib/Ouroboros/Network/PeerSelection/RootPeersDNS/PublicRootPeers.hs
+++ b/ouroboros-network/lib/Ouroboros/Network/PeerSelection/RootPeersDNS/PublicRootPeers.hs
@@ -18,6 +18,7 @@ import Data.Word (Word32)
 import System.Random
 
 import Control.Concurrent.Class.MonadSTM.Strict
+import Control.Monad (replicateM)
 import Control.Monad.Class.MonadAsync
 import Control.Monad.Class.MonadFork
 import Control.Monad.Class.MonadThrow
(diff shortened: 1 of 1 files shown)


crocodile-dentist:
I don't see a way this can be viably exploited or what actual risk this will mitigate.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/RootPeersDNS/PublicRootPeers.hs
```diff
@@ -73,13 +74,15 @@ publicRootPeersProvider tracer
...
+    rngVar <- newTVarIO rng
+    action (requestPublicRootPeers resourceVar rngVar)
   where
     requestPublicRootPeers
       :: StrictTVar m (Resource m (Either DNSorIOError resolver))
+      -> StrictTVar m StdGen
       -> Int
       -> m (Map peerAddr PeerAdvertise, DiffTime)
-    requestPublicRootPeers resourceVar _numRequested = do
+     requestPublicRootPeers resourceVar rngVar _numRequested = do
```
coot, line 85: Indentation
```suggestion
    requestPublicRootPeers resourceVar rngVar _numRequested = do
```

PR #5308, opened 2026-02-09
Handle missing peer in keepalive client updates

Onyinyechi46:
Motivation

Fix a crash/partial-state risk in keepAliveClient where code assumed a peer entry existed and used fromJust, by ensuring samples are recorded and the map is updated when the peer is missing.

Description
Updated ouroboros-network/lib/Ouroboros/Network/KeepAlive.hs to remove the fromJust usage and atomically perform a safe M.updateLookupWithKey that inserts the sample if the peer key was absent while preserving the existing update logic and asserting the invariant before writing the TVar.

Testing
No automated tests were run for this change; change is narrowly scoped and preserves existing behavior when entries exist while adding safe handling for missing entries.


crocodile-dentist:
It is an error if the peer is missing, this patch will mask an issue so it's probably better to let the peer crash. Our property tests should catch such a failure in pure code.

crocodile-dentist:
The current logic guarantees that the previous client has cleaned up before restarting, and there aren't any other components involved so no coordination issues can arise.

PR #5315, opened 2026-02-16
tx-submission robustness and testing enhancements

### LINKED ISSUE #5312: TxSubmission simulation failure
It can be reproduced with
```
cabal run ouroboros-network:ouroboros-network-sim-tests -- --quickcheck-replay="(SMGen 15534779231887965963 12670053090749425933,3)" -p '/AppV2.inflight/'
```

[sim.log](https://github.com/user-attachments/files/25235964/sim.log)

crocodile-dentist:
# Description

**Ensure all eligible downloaded tx's will be submitted to the mempool**
**Enforce that no transaction is enqueued to the mempool more than once by the same peer before it is acknowledged**
**Improve testcase generation and inflight test**
**Remove global size limit for inflight tx's**

resolves #5312 

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -116,9 +116,8 @@ acknowledgeTxIds
                    , txid `Map.notMember` bufferedTxs sharedTxState
                    , tx <- maybeToList $ txid `Map.lookup` downloadedTxs
                    ]
-    (toMempoolTxIds, _) =
-      StrictSeq.spanl (`Map.member` downloadedTxs) acknowledgedTxIds
-
+    toMempoolTxIds =
+      StrictSeq.filter (`Map.member` downloadedTxs) acknowledgedTxIds
```
coot, line 126:
```suggestion
    -- Select downloaded txs from the prefix of `acknowledgedTxIds`, ignoring
    -- unknown and buffered txs.
    toMempoolTxIds =
      StrictSeq.filter (`Map.member` downloadedTxs) acknowledgedTxIds
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -111,10 +111,14 @@ acknowledgeTxIds
...
-    txsToMempool = [ (txid, tx)
+    txsToMempool = [ (txid, downloadedTxs Map.! txid)
                    | txid <- toList toMempoolTxIds
                    , txid `Map.notMember` bufferedTxs sharedTxState
-                   , tx <- maybeToList $ txid `Map.lookup` downloadedTxs
+                   -- without the guard below we could potentially enqueue
+                   -- the same tx into the mempool multiple times over
+                   -- several decision loop iterations before the tx
+                   -- is finally in the mempool, or rejected.
+                   , txid `Map.notMember` toMempoolTxs
```
coot, line 121: :+1:

It's worth adding a test that captures properties of `txsToMempool` in `Test.Ouroboros.Network.TxSubmission.TxLogic` module:

1. `txsToMempool` contains only downloaded `txs`
2. `txsToMempool` is disjoint from `unacknowledgedTxIds` after the decision is taken
3. The diff between previous `txsToMempool` and the new one is a subset of the intersection `acknowledgedTxIds ∩ downloadedTxs \ bufferedTxs` (it's more fine-grained than 1., but requires multiple decisions to be made).

It's easier to analyse a failing property like that than a whole simulation.  These tests might come in handy when we start making bigger changes to the tx-logic.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/Types.hs
```diff
@@ -225,7 +229,7 @@ verboseTracer :: forall a m.
 verboseTracer = threadAndTimeTracer $ showTracing $ Tracer say
 
 debugTracer :: forall a s. Show a => Tracer (IOSim s) a
-debugTracer = threadAndTimeTracer $ showTracing $ Tracer (traceM . show)
+debugTracer = threadAndTimeTracer $ show >$< Tracer traceM
```
coot: 🤔 Btw, using `traceM` combined with `show` is strange to me.  It will just capture strings in `Dynamic` which is not visible in the `SimTrace`, `say` is better for debugging since it's visible in the trace and it can be recovered from the trace using `selectTraceSay`.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -212,15 +219,17 @@ runTxSubmission tracer tracerTxLogic st0 txDecisionPolicy = do
...
-                                                                     api
+                                  let server =
+                                        txSubmissionInboundV2 sayTracer --verboseTracer
+                                                              NoTxSubmissionInitDelay
+                                                              (getMempoolWriter duplicateTxIdsVar
+                                                                                inboundMempool)
+                                                              api
                                   runPipelinedPeerWithLimits
-                                    (("INBOUND " ++ show addr,) `contramap` verboseTracer)
+                                    (("INBOUND " ++ show addr,) `contramap` sayTracer {-verboseTracer-})
```
coot:
```suggestion

                                    (("INBOUND " ++ show addr,) `contramap` sayTracer)
```

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -299,9 +310,10 @@ txSubmissionSimulation (TxSubmissionState state txDecisionPolicy) = do
         atomically (traverse_ (`writeTVar` Terminate) controlMessageVars)
     ) \_ -> do
       let tracer :: forall a. (Show a, Typeable a) => Tracer (IOSim s) a
-          tracer = verboseTracer
-                <> debugTracer
-                <> Tracer traceM
+          tracer = sayTracer
```
coot: Don't we need `Tracer traceM` to get the events for the property validation?

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -344,6 +356,12 @@ prop_txSubmission st@(TxSubmissionState peers _) =
              counterexample (ppTrace tr)
            $ conjoin (validate inmp `map` outmps)
   where
+    checkMempools :: [Tx Int] -> [Tx Int] -> Bool
+    checkMempools consumer producer =
+      let producer' = Set.fromList $ getTxId <$> producer
+          consumer' = Set.fromList $ getTxId <$> consumer
+      in producer' `Set.isSubsetOf` consumer'
```
coot, line 361: :+1:

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -416,88 +430,102 @@ prop_txSubmission_inflight st@(TxSubmissionState state _) =
...
+    fn tx r' -- | getTxAdvSize tx /= getTxSize tx
+             -- = empty
+             -- ^ that is too severe
+            | Map.member (getTxId tx) r'
+            , getTxValid tx
+            = Map.update (Just . succ @Int) (getTxId tx) r'
+            | getTxValid tx
+            = Map.insert (getTxId tx) 1 r'
+            | otherwise
+            = r'
```
coot: Let's also simplify the logic with `Map.alter`:
```suggestion
    fn :: Tx TxId -> Map TxId Int -> Map TxId Int
    fn tx r' -- | getTxAdvSize tx /= getTxSize tx
             -- = empty
             -- ^ that is too severe
             | getTxValid tx
             = Map.alter (Just . maybe 1 succ) (getTxId tx) r'
             | otherwise
             = r'            
```

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -416,88 +430,102 @@ prop_txSubmission_inflight st@(TxSubmissionState state _) =
...
+          . counterexample ("hasInvalidSize: " <> show hasInvalidSize)
+          . counterexample ("Result valid [(txid, repeated)]:\n" <> show resultRepeatedValidTxs)
+          . counterexample ("Testcase max valid [(txid, repeated)]:\n" <> show maxRepeatedValidTxs)
+          . conjoin . Map.elems $ if hasInvalidSize
+              then merge (mapMissing \_txid _left  -> error "impossible")
+                         (mapMissing \_txid _right -> True)
+                         (zipWithMatched \_txid left right ->
+                             if txInflightMultiplicity policy >= right
+                               then left <= right
+                               else left <= txInflightMultiplicity policy)
```
coot: Could you add a comment on what this condition checks?

A simpler form:
```suggestion
                             left <= right `min` txInflightMultiplicity policy
```

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -416,88 +430,102 @@ prop_txSubmission_inflight st@(TxSubmissionState state _) =
...
+                             .&&. let inflight = Map.keysSet $ inflightTxs st
+                                      buffered = Map.keysSet $ bufferedTxs st
+                                      inflightBack = Map.keysSet $ inflightTxs stBack
+                                  in
+                                    -- here we account for a very slow peer from whom we requested
+                                    -- a transaction, but it didn't arrive until we have also requested
+                                    -- it from another peer, received it, and placed it into the mempool,
+                                    -- and so it ended up in bufferedTxs when the first one is still
+                                    -- in flight. It is an error when the opposite happens.
+                                    null $ inflight `Set.intersection` (buffered Set.\\ inflightBack)
```
coot: This is the same, but maybe a bit better, conveys the meaning, e.g. newly added inflight txs are not buffered.
```suggestion
                                    null $ (inflight Set.\\ inflightBack) `Set.intersection` buffered
```

ouroboros-network/bench/Main.hs
```diff
@@ -30,30 +36,36 @@ main =
...
+                    -- pPrint a
+                    performMajorGC
                     traceMarkerIO "evaluated decision context"
                     return a
                 )
-                (\a ->
+                (\(~a@(_policy, state)) ->
                      bench "makeDecisions: 10"
-                   $ nf (uncurry Tx.makeDecisions) a
+                   $ nf (flip (uncurry Tx.makeDecisions) (peerTxStates state)) a
```
coot: If we add a type signature, this will not break when we add more arguments to `makeDecisions`:
```suggestion
                   $ let f :: ( Tx.TxDeicisionPolicy
                              , SharedTxState PeerAddr TxId (Tx TxId)
                              )
                          ->  ( SharedTxState PeerAddr TxId (Tx TxId)
                              , Map PeerAddr (TxDecision TxId (Tx TxId))
                              )
                         f = flip (uncurry Tx.makeDecisions) (peerTxStates state)
                     in nf f a

```

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
crocodile-dentist: This tightens the check for `txInflightMultiplicity` - if the number of peers offering a tx is lower than the multiplicity cap, then there shouldn't be more of those tx's inflight than the number of associated peers.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -318,7 +314,11 @@ instance Ord txid => Semigroup (TxDecision txid tx) where
                    txdTxIdsToRequest     = txdTxIdsToRequest + txdTxIdsToRequest',
                    txdPipelineTxIds      = txdPipelineTxIds',
                    txdTxsToRequest       = txdTxsToRequest <> txdTxsToRequest',
-                   txdTxsToMempool       = txdTxsToMempool <> txdTxsToMempool'
+                   txdTxsToMempool       =
+                     let left  = Set.fromList . fmap fst $ listOfTxsToMempool txdTxsToMempool
+                         right = Set.fromList . fmap fst $ listOfTxsToMempool txdTxsToMempool'
+                         shared = Set.intersection left right
+                     in assert (Set.null shared) $ txdTxsToMempool <> txdTxsToMempool'
```
karknu: I don't think that this assert belongs here.
It works now because we only call (<>) on decisions for the same peer.

But different peers may decide to send the same TX to the mempool if `txInflightMultiplicity > 1`.
This means that we would no longer be able to use (<>) for decisions made by different peers.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/Types.hs
crocodile-dentist: Should we just remove this binding since what we seem to want here - include thread and time - is done with `verboseTracer`?

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
crocodile-dentist: The `txdTxsToMempool` have only the tx's enqueued by that particular peer so there is no risk of collision.

karknu: Only because our current use only ever does (<>) for decisions related to the same peer.

karknu: You can still have the assert, but move it to `decisionLogicThread` on `d` and `d'`.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
crocodile-dentist: I've created an issue so (3) can be added later.

PR #5350, opened 2026-04-02
Removed AcquireConnectionError

coot:
# Description

A more detailed trace is provided by ConnectionManager - `TrConnectError`.
I also expanded it to cover three other cases which `AcquireConnectionError`
covered.

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [x] Added labels.

### DIFF
--- a/ouroboros-network/lib/Ouroboros/Network/PeerSelection/PeerStateActions.hs
+++ b/ouroboros-network/lib/Ouroboros/Network/PeerSelection/PeerStateActions.hs
@@ -53,7 +53,6 @@ import Data.ByteString.Lazy (ByteString)
 import Data.Functor (void, ($>))
 import Data.Map.Strict (Map)
 import Data.Map.Strict qualified as Map
-import Data.Maybe (isNothing)
 import Data.Typeable (Typeable, cast)
 
 import Network.Mux qualified as Mux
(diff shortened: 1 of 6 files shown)


karknu (approved):
LGTM

ouroboros-network/lib/Ouroboros/Network/PeerSelection/PeerStateActions.hs
```diff
@@ -1233,6 +1230,5 @@ data PeerSelectionActionsTrace peerAddr vNumber =
     | PeerStatusChangeFailure (PeerStatusChangeType peerAddr) (FailureType vNumber)
     | PeerMonitoringError     (ConnectionId peerAddr) SomeException
     | PeerMonitoringResult    (ConnectionId peerAddr) (Maybe (WithSomeProtocolTemperature FirstToFinishResult))
-    | AcquireConnectionError  SomeException
```
crocodile-dentist, line 1236: One drawback of removing it here is that to get this trace one has to run with connection manager tracer on. Maybe we should keep both, and possibly make this one something like `DMaximal` verbosity for peer state actions tracer and/or Debug level trace?

[edit] also possibly keeping it can be more convenient for some property test, now or in the future.

crocodile-dentist (commented):
What do you think about keeping both?

ouroboros-network/lib/Ouroboros/Network/PeerSelection/PeerStateActions.hs
coot: I thought about this and I think it's ok to require `ConnectionManager` on, even at `Info` level it's not very verbose.  The outbound governor trace doesn't say which connection error, so one still needs the connection manager tracer to be on to get a meaningful information.

We didn't needed it for testing until now, so there's a good chance we won't need it in the future.  If I am proven wrong we can `git revert` and then avoid tracing it by filtering trace output in the node, rather than removing it.

crocodile-dentist (approved):
lgtm

PR #5354, opened 2026-04-16
tx-submission demo

### LINKED ISSUE #5337: TxSubmissionV2 demo
In order to measure traffic that `tx-submission` can support we are going to implement a `tx-submission` demo with the following requirements:

* single inbound side connected to multiple outbound tx-submission applications
* connect over a socket, re-use traffic shaping done in #5261
* a transaction in this experiment will just be a `ByteString` blob of fixed size (or range of sizes), `tx-id` a hash of that blob.
* we can simulate CPU demand on tx validation by means of `threadDelay`

There are two extreme scenarios in which we'd be interested in:

* all upstream peers offer the same set of txs
* all upstream peers offer different set of txs

Tx-Submission-V2 aim is to improve the bandwidth in the first case, while not disrupt the performance in the second case (versus V1) too much (extra CPU cycles are expected, since V1 is optimised the other way around).

coot:
# Description

`tx-submission-demo`

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [x] Added labels.

### DIFF
--- a/network-mux/wireshark-plugin/mux.lua
+++ b/network-mux/wireshark-plugin/mux.lua
@@ -11,12 +11,12 @@
 
 local default_settings = {
   -- the port used by `./network-mux/demo/mux-leios-demo.sh`
-  port = 9001,
+  port = 4000,
   -- a range of mini-protocol numbers is used to check if we parsed a valid
   -- MuxSDU header.  This is important when TCP segments are lost, and we need
   -- to skip bytes until we find a begining of a new MuxSDU.
-  min_mini_protocol_num = 2,
-  max_mini_protocol_num = 3,
+  min_mini_protocol_num = 1,
+  max_mini_protocol_num = 15,
 }
 
 local MUX_HDR_LEN = 8
--- a/ouroboros-network/demo/tx-submission/main.hs
+++ b/ouroboros-network/demo/tx-submission/main.hs
@@ -44,6 +44,7 @@ import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks (..))
 import Options.Applicative
 import Statistics.Quantile qualified as Stat
+import System.IO (hPutStrLn, stderr)
 import System.Random qualified as Random
 
 import Network.Mux qualified as Mx
@@ -74,6 +75,7 @@ import Test.QuickCheck.Gen (Gen (..))
 import Test.QuickCheck.Gen qualified as QC
 import Test.QuickCheck.Instances.ByteString ()
 import Test.QuickCheck.Random (newQCGen)
+import GHC.IO.Exception (IOException(..), IOErrorType (..))
 
 main :: IO ()
 main =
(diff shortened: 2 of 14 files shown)


network-mux/wireshark-plugin/mux.lua
```diff
@@ -11,12 +11,12 @@
 
 local default_settings = {
   -- the port used by `./network-mux/demo/mux-leios-demo.sh`
-  port = 9001,
+  port = 4000,
```
karknu, line 14: Please, lets not do this.

coot: You want the port `9001` or none?

network-mux/wireshark-plugin/mux.lua
```diff
@@ -11,12 +11,12 @@
...
   -- a range of mini-protocol numbers is used to check if we parsed a valid
   -- MuxSDU header.  This is important when TCP segments are lost, and we need
   -- to skip bytes until we find a begining of a new MuxSDU.
-  min_mini_protocol_num = 2,
-  max_mini_protocol_num = 3,
+  min_mini_protocol_num = 1,
+  max_mini_protocol_num = 15,
```
crocodile-dentist, line 19: Actually, this is not necessary and manually scanning for lost/OOO packets can be very slow. Wireshark can do this automatically, if you go to Edit->Preferences and select Protocols->TCP in the list on the left then ensure that 'Allow subdissector to reassemble TCP streams' and 'Reassemble out-of-order segments' are checked. Then some of the previous changes to this file can be reverted, or we can essentially re-use the ouroboros-network.lua dissector from @karknu, just including the new miniprotocols numbers for this demo.

ouroboros-network/demo/tx-submission/main.hs
```diff
@@ -0,0 +1,781 @@
...
+         (command "analyse-txs"
+         $ info analyseParser
+         $    fullDesc
+           <> progDesc "analyze tx file")
+  where
+    defaultPort :: PortNumber
+    defaultPort = 4000
+
+    defaultAddr :: IP
+    defaultAddr = read "127.0.0.1"
```
crocodile-dentist, line 166: Is this not supposed to be from some namespace, eg. 10.0.0.1?

crocodile-dentist (commented):
When running `cabal run demo-tx-submission` it outputs:
>Missing: (COMMAND | COMMAND | COMMAND | COMMAND)
>
>Usage: demo-tx-submission (COMMAND | COMMAND | COMMAND | COMMAND)

instead of one of the expected `inbound`, `outbound`, etc.

network-mux/wireshark-plugin/mux.lua
karknu: I'd say leave it unchanged. If you need this to match to make your demo easier move the demo to use pot 9001.

ouroboros-network/demo/tx-submission/main.hs
coot: It is supported, but the demo itself doesn't assume we run it in a special environment (like one provided by the included script).

crocodile-dentist (approved):
lgtm

PR #5357, opened 2026-04-24
tx-submission v2: return results in submitTxToMempool

coot:
I need this for IntersectMBO/dmq-node#53.


crocodile-dentist (approved):
lgtm

Issue #5336, opened 2026-03-12
Performance improvement for tx-submission v2


karknu:
V2 tx-submission was designed to reduce network load by limiting the number of peers the node requests the same TX from. However benchmarks shows that the V2 tx-submission consumes more CPU and allocs more than V1.

This needs to be addressed before it can replace V1. The network load reduction that V2 brings is needed for Leios.

ch1bo:
The undecision improvements were merged and should help with Leios-levels of throughput. We'll have a follow-up if we need more improvements.

PR #5375, opened 2026-05-19
V2 Tx-submission without a decision thread

karknu:
# Description

Refactoring of the V2 Tx-submission implementation without a decision thread.
Implements #5336. 

# Checklist

### Quality
* [ ] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.


ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1475 @@
...
+  , pacPeerInFlight ctx
+  , pacSharedState ctx
+  )
+
+-- | Select downloaded transactions that this peer may submit to the mempool.
+pickSubmitAction
+  :: PeerActionContext peeraddr txid tx
+  -> Maybe [TxKey]
+pickSubmitAction PeerActionContext { pacPeerState, pacPeerInFlight, pacSharedState } =
+  let txsToSubmit = pickBufferedTxsToSubmit in
```
coot: This binding is not neccessary

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -31,324 +44,608 @@ module Ouroboros.Network.TxSubmission.Inbound.V2.Types
...
+
+    -- | Keys this peer currently counts toward 'txAttempt'.  Spans
+    -- the @claim -> download -> buffer@ phases; the key moves to
+    -- 'pifSubmitting' on the @submit@ transition.
+    pifAttempting  :: !IntSet,
+
+    -- | Keys this peer currently counts toward 'txInSubmission'.
+    -- Set on 'PeerSubmitTxs' / 'markSubmittingTxs', cleared on accept
+    -- or reject.
+    pifSubmitting  :: !IntSet,
```
coot, line 210: Is `pifSubmitting` disjoined from `pifAttempting`?

It might be worth writing down what are the invariants for these sets.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1475 @@
...
+                       else go seen selectedRev selectedSize txTable rest
+                  -- Entry resolved or swept since this peer's local state last synced; skip.
+                  Nothing -> go seen selectedRev selectedSize txTable rest
+
+        exceedsBudget selectedSize txSize
+          | selectedSize + txSize <= sizeBudget = False
+          | selectedSize /= 0 = True
+          -- The inflight size limit is soft by up to one tx size, so a peer with
+          -- spare capacity may still request its first tx in a batch even when
+          -- that single tx exceeds the remaining byte budget.
```
coot: I'd put that comment in front of `exceedsBudget`, as it captures well the whole function.  It's a bit less permissive than a simple `selectedSize > sizeBudget` (which would allow to exceed by one tx not only if it's the first tx in the batch).

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1475 @@
...
+        -- in which the peer sent us the txids.  Walking the queue in that
+        -- order aligns our fetch order with the peer's advertisement order,
+        -- which preserves whatever submission-validity ordering the peer
+        -- intended within the batch.
+        candidates =
+          [ (k, txSize)
+          | TxKey k <- toList (peerUnacknowledgedTxIds pacPeerState)
+          , IntSet.notMember k (peerRequestedTxs pacPeerState)
+          , IntMap.notMember k (peerDownloadedTxs pacPeerState)
+          , Just txSize <- [IntMap.lookup k (peerAvailableTxIds pacPeerState)]
```
coot, line 436: Aren't we guaranteed that `k` is in `peerAvailableTxIds` (since it's in `peerUnacknowledgedTxIds`, e.g. this should be fine even if it is partial:
```hs
let txSize = peerAvailableTxIds pacPerState IntMap.! k
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+    -- Split the unacknowledged txid queue into acknowledged and remaining portions.
+    --
+    -- acknowledgedTxIds is the longest prefix of "ackable" txids.
+    -- unacknowledgedTxIds is the remaining txids
+    (acknowledgedTxIds, txIdsToAcknowledge, txIdsToRequest, unacknowledgedTxIds') =
+      ( toList acknowledgedTxIdsSeq
+      , fromIntegral numOfAcked
+      , txIdsToRequest'
+      , unacknowledgedTxIds
+      )
```
coot: we don't need to rebind unmodified `unacknowledgedTxIds`
```suggestion
    (acknowledgedTxIds, txIdsToAcknowledge, txIdsToRequest) =
      ( toList acknowledgedTxIdsSeq
      , fromIntegral numOfAcked
      , txIdsToRequest'
      )
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+        ackablePrefix = StrictSeq.takeWhileL (txIdAckable ctx) (peerUnacknowledgedTxIds
+                                             pacPeerState)
+
+        numOfUnacked = StrictSeq.length (peerUnacknowledgedTxIds pacPeerState)
+        numOfRequested = fromIntegral (peerRequestedTxIds pacPeerState) :: Int
+        hasOutstandingBodyReplies =
+          not (StrictSeq.null (peerRequestedTxBatches pacPeerState))
+        keepOneUnackedForPipelinedRequest =
+          txIdRequestMode == AllowPipelinedTxIdRequests
+            && (numOfRequested > 0 || hasOutstandingBodyReplies)
```
coot: If we never acknowledge all txs, the other side will not have a change to terminate the protocol cleanly.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+  -- in pipelined mode, so wait for a later wake.
+  | txIdRequestMode == AllowPipelinedTxIdRequests
+  , StrictSeq.null (peerUnacknowledgedTxIds pacPeerState) = Nothing
+
+  -- Backoff after an empty txid reply: if our last request came back
+  -- with zero txids and we still have nothing to ack, suppress the
+  -- next 'ack=0, req=N' pipelined wire message (V2.hs sends pipelined
+  -- whenever the unacked queue is non-empty).  The flag is cleared by
+  -- a non-empty txid reply or by the mempool accepting one of this
+  -- peer's txs - both signs the peer may have new work for us.
```
coot:
```suggestion
  -- peer's txs - both signal the peer may have new work for us.
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+                  Just txEntry ->
+                    if txSelectable ctx (TxKey k) txEntry
+                       then
+                         go (IntSet.insert k seen)
+                            (TxKey k : selectedRev)
+                            (selectedSize + txSize)
+                            (IntMap.insert k (claimTx (pacPeerAddr ctx) leaseUntil txEntry)
+                                           txTable)
+                            rest
+                       else go seen selectedRev selectedSize txTable rest
```
coot, line 465: I think we can add it to the `seen` set:  next time we'll pass the same `ctx`, also `txEntry` is not changed, so the outcome of `txSelectable` will be the same.
```suggestion
                       else go (IntSet.insert k seen) selectedRev selectedSize txTable rest
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+                  -- Entry resolved or swept since this peer's local state last synced; skip.
+                  Nothing -> go seen selectedRev selectedSize txTable rest
```
coot, line 467: We could use `alterF` to traverse `txTable` just once:
```suggestion
              let fn :: Maybe (TxEntry peeraddr) -> (Bool, Maybe (TxEntry peeraddr))
                  fn Nothing = (False, Nothing)
                  fn (Just txEntry) =
                    if txSelectable ctx (TxKey k) txEntry
                       then (True, Just (claimTx (pacPeerAddr ctx) leaseUntil txEntry))
                       else (False, Just txEntry)
              in
              case IntMap.alterF fn k txTable of
                (True, txTable') ->
                  go (IntSet.insert k seen)
                     (TxKey k : selectedRev)
                     (selectedSize + txSize)
                     txTable'
                     rest
                (False, txTable') ->
                  go seen selectedRev selectedSize txTable' rest
```

note: previous comment is not taken into account in this snippet.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Policy.hs
```diff
@@ -41,37 +41,64 @@ data TxDecisionPolicy = TxDecisionPolicy {
...
+      scoreAcceptDecrement           :: !Double,
+      -- ^ amount subtracted from the peer's score for each tx body the peer delivered that the mempool accepted.
+
+      interTxSpace                   :: !DiffTime,
+      -- ^ space between actual requests for the same TX.
+
+      inflightTimeout                :: !DiffTime,
+      -- ^ Maximum time a peer's attempt may sit between claim and
+      -- entering submission before the per-entry inflight-multiplicity
+      -- cap is bumped, allowing another peer to attempt in parallel.
```
coot, line 75: Shouldn't `inflightTimeout` be smaller than `interTxSpace` (see `stuckBumpReadyAt`).  It might be good to document it here.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+
+    futureClaimWake txEntry =
+      let claimWake = txClaimReadyAt pacClaimDelay txEntry in
+      if claimWake > pacNow then Just claimWake
+                            else Nothing
+
+    minMaybe :: Ord a => Maybe a -> Maybe a -> Maybe a
+    minMaybe Nothing y         = y
+    minMaybe x Nothing         = x
+    minMaybe (Just x) (Just y) = Just (min x y)
```
coot, line 599: or `min <$> ma <*> mb`, `liftA2 min`...

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+        numOfAcked
+          -- A pipelined txid request becomes a non-blocking protocol message
+          -- while any txid or body reply is still in flight. The outbound side
+          -- requires at least one txid to remain unacknowledged in that case.
+          | keepOneUnackedForPipelinedRequest =
+              min numOfAcked0 (max 0 (numOfUnacked - 1))
+          | otherwise = numOfAcked0
+
+        acknowledgedTxIdsSeq = StrictSeq.take numOfAcked ackablePrefix
+        unacknowledgedTxIds = StrictSeq.drop numOfAcked (peerUnacknowledgedTxIds pacPeerState)
```
coot: I think we can compute it with just a single traversal:
```suggestion
        (acknowledgedTxIdsSeq, unacknowledgedTxIds) = StrictSeq.splitAt numOfAcked (peerUnacknowledgedTxIds pacPeerState)
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+        txidsAndSizes
+
+    peerUnacknowledgedTxIds' =
+      peerUnacknowledgedTxIds peerState <> StrictSeq.fromList (reverse receivedTxKeysRev)
+
+    peerState'' = peerState {
+        peerUnacknowledgedTxIds = peerUnacknowledgedTxIds',
+        peerRequestedTxIds = fromIntegral $
+            max 0 ( fromIntegral (peerRequestedTxIds peerState) -
+                    fromIntegral requestedTxIds :: Int ),
```
coot: Maybe use an assertion that `requestedTxIds <= peerRequestedTxIds peerState`?

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -1,549 +1,624 @@
...
+applyReceivedTxIdsImp :: ( MonadSTM m
+                         , HasRawTxId txid )
+                      => TxDecisionPolicy
+                      -> STM m (MempoolSnapshot txid tx idx)
+                      -> SharedTxStateVar m peeraddr txid
+                      -> PeerTxInFlightVar m
+                      -> TxSubmissionCountersVar m
+                      -> Time
+                      -> NumTxIdsToReq
+                      -> [(txid, SizeInBytes)]
```
coot, line 496: It's kind of obvious, but still I think it's good to include it.
```suggestion
                      -> NumTxIdsToReq
                      -- ^ number of requested txid's
                      -> [(txid, SizeInBytes)]
                      -- ^ actually received txid's with sizes
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+        go IntSet.empty [] (toList (peerUnacknowledgedTxIds pacPeerState))
+      where
+        go _    acc [] = reverse acc
+        go seen acc (txKey@(TxKey k) : rest)
+          | IntSet.member k seen = go seen acc rest
+          | otherwise =
+              case IntMap.lookup k (sharedTxTable pacSharedState) of
+                Just txEntry
+                  | txBufferedByPeer pacPeerState k
+                  , not (txSubmittingByOther pacPeerInFlight k txEntry) ->
```
coot: It seems the `txSubmittingByOther` which excludes own txs is a defensive measure. Is it even possible that we'll be resubmitting our own txs? I think not, because submittion by a given peer runs sequentially.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -31,324 +44,608 @@ module Ouroboros.Network.TxSubmission.Inbound.V2.Types
...
+--   may simply be slow to claim).
+data PeerTxInFlight = PeerTxInFlight {
+    -- | Keys this peer currently has in its advertised window.  Used
+    -- by the orphan sweep to know the entry is still wanted, and by
+    -- 'nextWakeDelay' to scan for the earliest claim wake.
+    pifAdvertised  :: !IntSet,
+
+    -- | Keys this peer currently holds 'TxLeased' on.  Cleared on
+    -- omit, accept, reject, lease-loss and bracket exit.
+    pifLeased      :: !IntSet,
```
coot, line 200: Isn't there an invariant between `pifLeased` and `sharedTxTable`: e.g. the `peerAddr` in `TxEntry` must match the owner of `PeerTxInFlight`.

If this is an invariant do we verify it in a simulation?

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -31,324 +44,608 @@ module Ouroboros.Network.TxSubmission.Inbound.V2.Types
...
+  | Just key <- Map.lookup rawId sharedTxIdToKey = (rawId, key, st)
+  | otherwise =
+      let key = TxKey sharedNextTxKey in
+      ( rawId
+         , key
+         , st { sharedTxIdToKey = Map.insert rawId key sharedTxIdToKey
+              , sharedKeyToTxId = IntMap.insert sharedNextTxKey txid sharedKeyToTxId
+              , sharedNextTxKey = sharedNextTxKey + 1
+              }
+         )
```
coot: indentation suggetion:
```suggestion
      ( rawId
      , key
      , st { sharedTxIdToKey = Map.insert rawId key sharedTxIdToKey
           , sharedKeyToTxId = IntMap.insert sharedNextTxKey txid sharedKeyToTxId
           , sharedNextTxKey = sharedNextTxKey + 1
           }
      )
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -31,324 +44,608 @@ module Ouroboros.Network.TxSubmission.Inbound.V2.Types
...
-    txdPipelineTxIds      = False,
-    txdTxsToRequest       = Map.empty,
-    txdTxsToMempool       = mempty
+-- Both are kept in lockstep by the 'retained*' helpers below.
+data RetainedTxs = RetainedTxs {
+    retainedQueue :: !(IntPSQ Time ()),
+    retainedSet   :: !IntSet
   }
+  deriving stock    (Eq, Show, Generic)
+  deriving anyclass (NFData, NoThunks)
```
coot, line 576: Could you add a haddock which describes when a key enters the `RetainedTxs` and when it leaves it.  I presume it's when we downloaded a `tx` and the keept a bounded amount of time after it's accepted by the `Mempool` (if it's rejected, it needs to be kept until all peers that know it will try submitting it until first success or all failed).

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -31,324 +44,608 @@ module Ouroboros.Network.TxSubmission.Inbound.V2.Types
...
+    -- | Structural dirty bit.  Bumped on most structural changes; used by
+    -- 'writeSharedStateIfChanged' (paired with 'sharedGeneration') to skip
+    -- redundant TVar writes.  A change to either counter triggers a write.
+    sharedRevision    :: !Word64
+  }
+  deriving stock Generic
 
-    txdTxsToRequest       :: !(Map txid SizeInBytes),
-    -- ^ txid's to download.
+deriving stock instance (Eq peeraddr, Eq txid, HasRawTxId txid) => Eq (SharedTxState peeraddr txid)
```
coot, line 538: Do we need an `Eq` instance on `SharedTxState`? It is expensive on various containers.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1480 @@
...
+      (txid, tx) =
+        case Map.lookup (getRawTxId txid) txidToKey of
+             Nothing ->
+               ( lateCountAcc + 1
+               , pendingKeysAcc
+               , bufferedAcc
+               , sharedAcc
+               , downloadedAcc
+               , wakeAcc
+               )
```
coot, line 1078: Should this branch be protected with `assert False`?

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -1,549 +1,624 @@
...
+                         -> [TxKey]
+                         -> [TxKey]
+                         -> PeerTxLocalState tx
+                         -> m (PeerTxLocalState tx),
+
+    -- | Resolve txids and advertised sizes for a batch of tx keys to request.
+    resolveTxRequest     :: PeerTxLocalState tx
+                         -> [TxKey]
+                         -> m (Map txid SizeInBytes),
+    -- | Resolve buffered tx bodies into full submission records.
```
coot, line 227:
```suggestion
                         -> m (Map txid SizeInBytes),

        -- | Resolve buffered tx bodies into full submission records.
        --
        -- pre-condition: the list of keys is in `peerDownloadedTxs` field of
        -- `PeerTxLocalState`.  It is ensured by `pickSubmitAction`.
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs
```diff
@@ -7,199 +7,434 @@
...
+      let submitted = [ (txKey, txid') | (txKey, txid', _) <- bufferedTxs ]
+          toSubmit  = [ tx | (_, _, tx) <- bufferedTxs ]
+
+      (acceptedTxIds, _) <- if null toSubmit
+                               then pure ([], [])
+                               else mempoolAddTxs toSubmit
+      end <- getMonotonicTime
 
+      let (acceptedTxs, rejectedTxs) =
+            classifySubmittedTxs submitted (Set.fromList acceptedTxIds)
```
coot: We don't need to `classifySubmittedTxs`, instead we can rely on `mempoolAddTxs` returning the list of rejected txs above.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs
```diff
@@ -7,199 +7,434 @@
...
 import Ouroboros.Network.TxSubmission.Inbound.V2.Types as V2
 
+-- The same Stateful types as V1 uses.
+newtype Stateful s n txid tx m = Stateful (s -> ServerStIdle n txid tx m ())
+
+newtype StatefulM s n txid tx m
+  = StatefulM (s -> m (ServerStIdle n txid tx m ()))
+
+newtype StatefulCollect s n txid tx m
+  = StatefulCollect (s -> Collect txid tx -> m (ServerStIdle n txid tx m ()))
```
coot, line 50: I never liked these types.  They just hide the protocol state which is useful when reading the code, and are not a real `StateT` transformer, more of a `ReaderT` type of thing.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs
```diff
@@ -7,199 +7,434 @@
...
+             continueWithStateM (submitBufferedTxs txKeys serverIdle) peerState''
+           PeerRequestTxs txKeys ->
+             continueWithStateM (requestTxBodies Zero txKeys) peerState''
+           PeerRequestTxIds txIdsToAck txIdsToReq ->
+             continueWithStateM (serverReqTxIds Zero txIdsToAck txIdsToReq) peerState''
+
+    -- | Submit buffered transaction bodies to the mempool.
+    submitBufferedTxs :: forall (n :: N).
+                         [TxKey]
+                      -> StatefulM (PeerTxLocalState tx) n txid tx m
```
coot, line 218:
```suggestion
                      -> StatefulM (PeerTxLocalState tx) n txid tx m
                      -- ^ a continuation
```

coot (commented):
Review of the core logic & inbound client.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs
```diff
@@ -7,199 +7,434 @@
...
-      pure $ CollectPipelined
-                Nothing
-                (handleReply serverIdle)
+                     Nat (S n)
+                  -> Stateful (PeerTxLocalState tx) (S n) txid tx m
+    handleReplies (Succ Zero) = Stateful $ \peerState ->
+      CollectPipelined Nothing (collectAndContinueWithState (handleReply Zero) peerState)
+
+    handleReplies (Succ n'@Succ{}) = Stateful $ \peerState ->
+      CollectPipelined Nothing (collectAndContinueWithState (handleReply n') peerState)
```
coot:
```suggestion
    handleReplies (Succ n) = Stateful $ \peerState ->
      CollectPipelined Nothing (collectAndContinueWithState (handleReply n) peerState)
```
_I checked that it still compiles._

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
karknu: Yes, those sets are disjoint. It is checked in `peerTxInFlightInvariant` in TxLogic. 
I will add a comment for it.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: No, txids that that we learn of but are retained or in the mempool are removed from `peerAvailableTxIds`. 
They are not available for download but we still need to track them so that they can be ack:ed.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: The proposed change doesn't compile, but your comment is correct.
I will remove the nested `where` and the unnecessary bind along with it.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: I think both your and my version lacks a `that`, i.e. `both signs that the peer may have..." .
I will add it.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: The other side can only send `MsgDone` as a reply to a `MsgRequestTxIdsBlocking` request.
This means we will be out of pipeline mode and this part won't apply.

If the other side wants to speed up the exiting process it could stop providing new `txids` and omit tx bodies in its replies.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: I don't think this is worth the complication. This will only have an effect in the case of duplicate `txid`s within the same window.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: This change is a win when we end up picking TXs to download. But when we can't select a TX it will rebuild the spine of the IntMap.
The expected use case is fetching TXs from one peer and have 100+ peers piggy back on its fetch. Therefor I don't think this is a net win.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Policy.hs
karknu: No, `inflightTimeout` should be larger than `interTxSpace` . Otherwise one would just keep bumping the inflight limits before the `interTxSpace` timeout triggers.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: I prefer minMaybe since it looks less like perl ;).

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: 👍 Will make the change but your patch doesn't apply since I've made other changes.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: 👍

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: No, we won't be resubmitting our own txs. `seen` would handle that de-duplication.
The `txInSubmission` flag will is a better check for this. I will simplify this code.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
karknu: Yes, there is an invariant tied to `pifLeased`. However it is slightly relaxed since a peer can have its lease stolen by another peer once it expires. That theft doesn't update the victims private `pifLeased` feild.

See combinedStateInvariant in TxLogic.hs

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
karknu: I will add a haddock for the lifecycle.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
karknu: I agree that it is expensive, but it is only used for testing.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: No, this is a common case when `txInflightMultiplicity` > 1. 
Multiple peers where downloading the same tx and this path is taken by the loosing peer.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
karknu: 👍 . Added, but without the extra indentation.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs
karknu: Nice. That was changed since I started to work on this. Will remove `classifySubmittedTxs`.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs
karknu: We need some way to thread local state through the typed-protocol state machine.
But if you have a suggestion we can change it.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
coot: Right, good point.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
coot: > If the other side wants to speed up the exiting process it could stop providing new txids and omit tx bodies in its replies.

Let's do that.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
coot: > `both signs that the peer may have..." .

I am sure you mean: `both signal that ...` :smile:

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
coot: > That theft doesn't update the victims private pifLeased feild.

It would be good to document it somewhere.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
coot: We could you use a newtype wrapper in tests to provide an instance, but maybe it's probably not worth the effort.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs
coot: One can use the left hand side directly, without the newtype wrappers, e.g. just passing the state as an argument.  But it's not essential...

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Policy.hs
coot: Right, it should be larger - could you document it here.  If there are other in-variants like this maybe add `saneTxDecisionPolicy :: TxDecisionPolicy -> Bool` and use an assertion or something similar.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Policy.hs
```diff
@@ -41,37 +41,64 @@ data TxDecisionPolicy = TxDecisionPolicy {
...
 
-      scoreMax               :: !Double
+      scoreMax                       :: !Double,
       -- ^ Maximum number of "rejections". Unit: seconds
 
+      scoreAcceptDecrement           :: !Double,
+      -- ^ amount subtracted from the peer's score for each tx body the peer delivered that the mempool accepted.
+
+      interTxSpace                   :: !DiffTime,
+      -- ^ space between actual requests for the same TX.
```
coot: I'd add that it governs the lease time, just to connect it with the right notion.

coot (commented):
I resolved some comments, you can resolve the ones you already fixed in code.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
karknu: It is documented in for example the haddock for `scrubFromPeerInFlight` in Registry.hs and in the invariant checks in TxLogic.hs.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2.hs
karknu: I will leave it as is then.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: That would be on the server side. I am not expanding this branch to cover changes to that.
This PR is already large enough, but I will create an issue for it.

Created https://github.com/IntersectMBO/ouroboros-network/issues/5390

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
karknu: Yeah, I don't think it is worth it.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/TxLogic.hs
```diff
@@ -1,1622 +1,3573 @@
...
+--
+
+-- | Check that a value has no thunks in its fields.
+checkNoThunks :: NoThunks a => String -> a -> Property
+checkNoThunks name val =
+    val `seq` case unsafeNoThunks val of
+      Nothing   -> property True
+      Just info -> counterexample
+                     (name ++ " contains thunks: " ++ show info)
+                     (property False)
```
coot: btw, a recent version of `nothunks` will include type information and source location, which are available if one is compiling with these two flags:
```
package *
  ghc-options: -finfo-table-map
               -fdistinct-constructor-tables

```
_that's a snippet from my `cabal.project.local` file_.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/TxLogic.hs
```diff
@@ -1,1622 +1,3573 @@
...
+                            peerAvailableTxIds,
+                            peerRequestedTxs,
+                            peerRequestedTxBatches,
+                            peerRequestedTxsSize,
+                            peerDownloadedTxs,
+                            peerScore
+                          } =
+  conjoin
+    [ counterexample "requested keys are not all available"
+        (property (peerRequestedTxs `IntSet.isSubsetOf` availableKeys))
```
coot: The `property` is not needed inside a `counterexample` (similar in the other places in this module):
```suggestion
        (peerRequestedTxs `IntSet.isSubsetOf` availableKeys)
```

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet/MiniProtocols.hs
```diff
@@ -713,30 +722,36 @@ applications debugTracer txSubmissionInboundTracer txSubmissionInboundDebug node
...
+      -> TxSubmissionCountersVar m
+      -> SharedTxStateVar m NtNAddr Int
+      -> PeerTxRegistry m NtNAddr
       -> MiniProtocolCb (ResponderContext NtNAddr) ByteString m ()
-    txSubmissionResponder mempool txChannelsVar txMempoolSem sharedTxStateVar =
+    txSubmissionResponder mempool txCountersVar sharedTxStateVar inFlightRegistry =
       MiniProtocolCb $
         \ ResponderContext { rcConnectionId = connId@ConnectionId { remoteAddress = them }} channel
         -> do
-          withPeer txSubmissionInboundDebug
```
coot, line 727: The new `withPeer` doesn't have the `txSubmissionInboundDebugTracer`.  Either create an issue to provide it or remove the `_txSubmissionInboundDebugTracer` argument from the top of the function.

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet/Simulation.hs
```diff
@@ -240,6 +241,9 @@ data NodeArgs =
     , naChainSyncEarlyExit     :: Bool
     , naFetchModeScript        :: Script FetchMode
     , naTxs                    :: [Tx Int]
+    , naTxImpairment           :: Impairment
+      -- ^ behavioural fault injection on this node's outbound
+      -- 'TxSubmissionClient' (default: 'noImpairment')
```
coot, line 246: 👍

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/Impaired.hs
```diff
@@ -0,0 +1,253 @@
...
+          SendMsgReplyTxs txs k <- recvMsgRequestTxs txids
+          kept <- atomically (filterM (const rollKeep) txs)
+          pure (SendMsgReplyTxs kept (wrapIdle k))
+      }
+
+    wrapTxIds :: ClientStTxIds blocking txid tx m a
+              -> ClientStTxIds blocking txid tx m a
+    wrapTxIds (SendMsgReplyTxIds reply k) = SendMsgReplyTxIds reply (wrapIdle k)
+    wrapTxIds (SendMsgDone a)             = SendMsgDone a
+
```
coot, line 117: We're missing `wrapTxs`:
```suggestion

    wrapTxs :: ClientStTxs txid tx m a -> ClientStTxs txid tx m a
    wrapTxs (SendMsgReplyTxs txs k) = SendMsgReplyTxs txs (wrapIdle k)

```

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/Impaired.hs
```diff
@@ -0,0 +1,253 @@
...
+    -- Append (req + n - length xs) duplicates of the head, so the final
+    -- length exceeds the requested count by exactly n.
+    pad :: NumTxIdsToReq
+        -> BlockingReplyList blocking (txid, SizeInBytes)
+        -> BlockingReplyList blocking (txid, SizeInBytes)
+    pad _   r | n == 0 = r
+    pad req (BlockingReply xs) =
+        let target = fromIntegral (getNumTxIdsToReq req) + fromIntegral n :: Int
+            need   = max 1 (target - NonEmpty.length xs)
```
coot: The comment above sais:

> so the final length exceeds the requested count by exactly n

if `n >= 0` then we don't need `max 1`, since then `n > 0` in this branch.  But maybe `n` can be negative?

The same below.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/Impaired.hs
```diff
@@ -0,0 +1,253 @@
...
+    -- ^ added before each MsgReplyTxs; txid replies are unaffected
+  , impairOmitProb      :: Double
+    -- ^ per-body Bernoulli drop probability, in [0, 1]
+  , impairSeed          :: Int
+    -- ^ seed for the per-peer StdGen used by 'omitBodies'
+  , impairExtraTxIds    :: Word16
+    -- ^ pad each MsgReplyTxIds with this many duplicate entries; 0 = off
+  , impairUnrequestedTx :: Bool
+    -- ^ append one fabricated body whose txid is not in the request
+  } deriving (Eq, Show)
```
coot, line 226: If we have a `QuickCheck` generator for this type, it would be good to put it in this module.

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet.hs
```diff
@@ -1296,6 +1312,239 @@ prop_txSubmission_chainIntegrity_iosim :: ArbTxDecisionPolicy
...
+    let txIdCap = fromIntegral
+                . getNumTxIdsToReq
+                . maxNumTxIdsToRequest
+                $ txScoreImpairmentPolicy
+    n    <- choose (1, txIdCap)
+    bMul <- choose (1.5, 4.0)
+    cMul <- (bMul +) <$> choose (1.0, 6.0)
+    pure (ScoreImpairmentInput n bMul cMul)
+  -- Only shrink the tx count: the delay constraint cMul > bMul > 1 is
+  -- delicate and shrinking either independently could violate it.
```
coot: We can independently shrink `bMul` and `cMul - bMul`.

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet.hs
```diff
@@ -1296,6 +1312,239 @@ prop_txSubmission_chainIntegrity_iosim :: ArbTxDecisionPolicy
...
+  shrink (ScoreImpairmentInput n bMul cMul) =
+    [ ScoreImpairmentInput n' bMul cMul
+    | n' <- shrink n, n' >= 1
+    ]
```
coot, line 1360:
```suggestion
    ]
    ++
    [ ScoreImpairmentInput n' bMul' (bMul' + cMul - bMul)
    | bMul' <- shrink bMul
    , bMul' >= 1
    ]
    ++
    [ ScoreImpairmentInput n' bMul (bMul + diff)
    | diff <- shrink (cMul - bMul)
    , diff > 0
    ]
```

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -71,16 +74,49 @@ import Test.QuickCheck
...
+          prop_TxSubmissionDisconnectState_shrinkSmaller
+      , testProperty "TxSubmissionDisconnectState/shrinkNoDups"
+          prop_TxSubmissionDisconnectState_shrinkNoDups
+      ]
+  , testProperty "txSubmission"          prop_txSubmission
+  , testProperty "inflight"              prop_txSubmission_inflight
+  , testProperty "resilientToImpairment" prop_txSubmission_resilientToImpairment
+  , testProperty "peerDisconnect"        prop_txSubmission_peerDisconnect
   , testProperty "SharedTxState" $ withMaxSize 25
                                  $ withMaxSuccess 25
```
coot: If it's merged after #5379, `withMaxSuccess` needs to be replaced with the shim version of `withNumTests`.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -312,19 +381,235 @@ txSubmissionSimulation (TxSubmissionState state txDecisionPolicy) = do
...
 filterValidTxs
   = filter getTxValid
   . takeWhile (\Tx{getTxSize, getTxAdvSize} -> getTxSize == getTxAdvSize)
 
+-- | Mutator passed to 'applyImpairment' for the @unrequestedTx@ wrapper.
+-- Returns a clone of the given body with its txid replaced by 'maxBound',
+-- which is well outside the @Arbitrary Int@ range used by the generators
+-- and so is guaranteed not to collide with any requested txid.
+mkUnrequested :: [Int] -> Tx Int -> Tx Int
+mkUnrequested _reqs tx = tx { getTxId = maxBound }
```
coot, line 399: Why we don't care that the `txId`s are unique?

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -312,19 +381,235 @@ txSubmissionSimulation (TxSubmissionState state txDecisionPolicy) = do
...
+-- with their simulation time.  The full 'ppTrace' includes scheduling
+-- noise (ThreadDelay, Deschedule, TxCommitted, etc.) that obscures the
+-- protocol-relevant events; this helper restricts the output to the
+-- events that the test infrastructure forwards through 'sayTracer'
+-- (protocol messages, counter snapshots, shared-state changes).
+ppSayTrace :: SimTrace a -> String
+ppSayTrace tr =
+    List.intercalate "\n"
+  $ map (\(Time t, ev) -> show t <> " " <> ev)
+  $ selectTraceEventsSayWithTime' tr
```
coot, line 411: Or
```
ppSayTrace :: SimTrace a -> String
ppSayTrace tr = ppEvents
              [ a
              | a(_, _, _, EventSay {}) <- traceEvents tr
              ]
```
it will include: time, thread id / label and the say event.  Maybe we should include it in the `io-sim` library?

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
```diff
@@ -469,50 +757,622 @@ prop_txSubmission_inflight st@(TxSubmissionState state policy) =
...
+     [ Map.delete k m | Map.size m > 1, k <- Map.keys m ]
+  ++ [ Map.insert k imp' m
+     | (k, imp) <- Map.toList m
+     , imp' <- shrinkImpairment imp
+     ]
+
+shrinkImpairment :: Impairment -> [Impairment]
+shrinkImpairment imp = List.nub $
+     [ imp { impairUnrequestedTx = False } | impairUnrequestedTx imp ]
+  ++ [ imp { impairExtraTxIds = 0 }        | impairExtraTxIds imp > 0 ]
```
coot:
```suggestion
  ++ [ imp { impairExtraTxIds = a }        | a <- shrink (impairExtraTxIds imp)
                                           , a > 0 ]
  ++ [ imp { impairExtraTxIds = 0 }        | impairExtraTxIds imp > 0 ]                                       
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1470 @@
...
-                    peeraddr
-                    peerTxStates of
-      (Right st', peerTxStates') ->
-        Right st' { peerTxStates = peerTxStates' }
-      (Left e, _) ->
-        Left $ ProtocolErrorTxSizeError e
+-- A txid remains ackable after it has been resolved and removed from shared
+-- state. The wire protocol only needs an ack count for the peer-local
+-- unacknowledged prefix, so a late ack is still safe even after the active
+-- entry and retained marker have been pruned.
```
crocodile-dentist: This is a dense haddock.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -1,549 +1,629 @@
...
+  :: MonadSTM m
+  => PeerTxRegistry m peeraddr
+  -> STM m IntSet
+snapshotLiveReferences registry = do
+    peers <- readTVar registry
+    pifs  <- traverse (readTVar . fst) (Map.elems peers)
+    pure $! IntSet.unions
+              [ s
+              | pif <- pifs
+              , s   <- [pifAdvertised pif, pifAcksPending pif]
```
crocodile-dentist: pifAcksPending is a superset of pifAdvertised so we don't need this in the union, ie. pifAcksPending should suffice, no?

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/TxLogic.hs
```diff
@@ -1,1622 +1,3573 @@
...
+           (peerTxLocalStateInvariant policy ps)
+       | (p, (ps, _)) <- Map.toList peers ]
+    ++ [ counterexample ("PeerTxInFlight invariant violated for peer "
+                           ++ show p)
+           (peerTxInFlightInvariant pif)
+       | (p, (_, pif)) <- Map.toList peers ]
+    ++ [ sharedTxStateInvariant sharedState ]
+    ++ [ counterexample ("peer " ++ show p
+                          ++ " advertises keys that are neither active nor retained")
+           (property (pifAdvertised pif `IntSet.isSubsetOf` liveKeys))
```
crocodile-dentist: Do we also want to add
```haskell
      ++ [ counterexample ("peer " ++ show p                                                                                                                                                                                                                  
                            ++ " has pifAdvertised not a subset of pifAcksPending")
             (property (pifAdvertised pif `IntSet.isSubsetOf` pifAcksPending pif))                                                                                                                                                                            
         | (p, (_, pif)) <- Map.toList peers ]
```
here?

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1470 @@
...
+    -- | Address of the peer whose next action is being chosen.
+    pacPeerAddr     :: !peeraddr,
+    -- | Current peer-local state after local pruning has been applied.
+    pacPeerState    :: !(PeerTxLocalState tx),
+    -- | This peer's contribution counters mirroring shared-state writes.
+    pacPeerInFlight :: !PeerTxInFlight,
+    -- | Shared tx-submission state after shared pruning has been applied.
+    pacSharedState  :: !(SharedTxState peeraddr txid),
+    -- | Score-derived delay this peer must wait after a tx becomes claimable.
+    pacClaimDelay   :: !DiffTime
```
crocodile-dentist, line 56: Mathing shows that for 10ms min penalty and drain rate of .001, a single rejection will cause this delay to be applied for 17 minutes, if I understand this correctly. That seems very punitive.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/TxLogic.hs
karknu: Nice, I have added it to my local file.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Policy.hs
karknu: I have added a `saneTxDecisionPolicy` function and its haddock documents the limits.
I've added test for the TxDecisionPolicy generator and shrinker to verify that all policies are sane (including `defaultTxDecisionPolicy`)

Currently the use of `defaultTxDecisionPolicy` is hardcoded but if/when we decide to make it configurable the node can use `saneTxDecisionPolicy` to verify it.

crocodile-dentist (commented):
A few comments for now, I'm still looking

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: There needs to be a noticeable penalty due to how it is used.
The penalty is applied as a delay for when a peer may issue a request for a TX. This makes it possible to pick a winning peer without the need for coordination between them.

And there is reward for good behavior, having a tx accepted by the mempool (scoreAcceptDecrement) will quickly take away the penalty.

A peer that had a sudden latency spike, or just bad luck (we may have just adopted a block that contained the TX it was sending us) will be able to work of its penalty. But for a peer that does nothing it will take a long time to have the penalty to tick down.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -31,324 +44,688 @@ module Ouroboros.Network.TxSubmission.Inbound.V2.Types
...
+      let key = TxKey sharedNextTxKey in
+      ( rawId
+      , key
+      , st { sharedTxIdToKey = Map.insert rawId key sharedTxIdToKey
+           , sharedKeyToTxId = IntMap.insert sharedNextTxKey txid sharedKeyToTxId
+           , sharedNextTxKey = sharedNextTxKey + 1
+           }
+      )
+  where
+    rawId = getRawTxId txid
```
crocodile-dentist, line 708: A single lookup and insert if missing:

```suggestion
internTxId txid st@SharedTxState { sharedTxIdToKey, sharedKeyToTxId, sharedNextTxKey } =
    case Map.alterF f rawId sharedTxIdToKey of
      (Just existing, _) ->
        -- Already interned; map unchanged.
        (rawId, existing, st)
      (Nothing, sharedTxIdToKey') ->
        -- Freshly allocated; map has the new entry.
        ( rawId
        , freshKey
        , st { sharedTxIdToKey = sharedTxIdToKey'
             , sharedKeyToTxId = IntMap.insert sharedNextTxKey txid sharedKeyToTxId
             , sharedNextTxKey = sharedNextTxKey + 1
             }
        )
  where
    rawId    = getRawTxId txid
    freshKey = TxKey sharedNextTxKey
    f Nothing    = (Nothing, Just freshKey)
    f j                  = (j, j)
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
karknu: You're right, will make the change.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/TxLogic.hs
karknu: I've remove the superfluous `property` in the file.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/TxLogic.hs
karknu: I've added that check to `peerTxInFlightInvariant` since that invariant only requires TxInFlight.
I did add a check that pifAcksPending is a subset of key(peerUnacknowledgedTxIds) to combinedStateInvariant

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -31,324 +44,688 @@ module Ouroboros.Network.TxSubmission.Inbound.V2.Types
...
+
+internTxIds :: (Foldable f, HasRawTxId txid)
+            => f txid
+            -> SharedTxState peeraddr txid
+            -> (Map (RawTxId txid) TxKey, SharedTxState peeraddr txid)
+internTxIds txids st0 = List.foldl' step (Map.empty, st0) txids
+  where
+    step (acc, st) txid =
```
crocodile-dentist:
```suggestion
    step (!acc, !st) txid =
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
karknu: Agreed, I've rephrased it.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
```diff
@@ -1,549 +1,629 @@
...
+
+    decAttempt tbl k =
+      IntMap.adjust
+        (\entry -> entry { txAttempt = max 0 (txAttempt entry - 1) })
+        k tbl
+
+    clearSubmission tbl k =
+      IntMap.adjust
+        (\entry -> entry { txInSubmission = False })
+        k tbl
```
crocodile-dentist:
```suggestion
    -- Walk each unique key once and apply all three modifications in a
    -- single 'IntMap.adjust'.  Keys typically appear in two of the
    -- three sets ('pifLeased' is a subset of 'pifAttempting ∪
    -- pifSubmitting').
    allKeys = pifLeased pif
        `IntSet.union` pifAttempting pif
        `IntSet.union` pifSubmitting pif

    sharedTxTable' = IntSet.foldl' scrubOne (sharedTxTable st) allKeys

    scrubOne tbl k = IntMap.adjust (modifyEntry k) k tbl

    modifyEntry k =
        applyIf (IntSet.member k (pifSubmitting pif)) clearSubmission
      . applyIf (IntSet.member k (pifAttempting pif)) decAttempt
      . applyIf (IntSet.member k (pifLeased     pif)) releaseLease

    applyIf True  f = f
    applyIf False _ = id

    releaseLease entry = case txLease entry of
      TxLeased owner _ | owner == peeraddr -> entry { txLease = TxClaimable now }
      _                                    -> entry

    decAttempt entry = entry { txAttempt = max 0 (txAttempt entry - 1) }

    clearSubmission entry = entry { txInSubmission = False }
```
I think it's more comprehensible this way.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
```diff
@@ -31,324 +44,688 @@ module Ouroboros.Network.TxSubmission.Inbound.V2.Types
...
+             => SharedTxState peeraddr txid
+             -> TxKey
+             -> txid
+resolveTxKey SharedTxState { sharedKeyToTxId } (TxKey k) =
+    case IntMap.lookup k sharedKeyToTxId of
+         Just txid -> txid
+         Nothing   -> error $
+           "TxSubmission.V2.resolveTxKey: missing tx key " ++ show k
+
+internTxId :: HasRawTxId txid
```
crocodile-dentist, line 692:
```suggestion
-- | Intern a txid into the shared lookup tables.  Returns the txid's
-- raw form, its stable 'TxKey' (existing if already interned, freshly
-- allocated otherwise), and the resulting shared state.
internTxId :: HasRawTxId txid
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1470 @@
...
-       -> [txid]
-       -> Map txid Int
-    fn m txids = Foldable.foldl' gn m txids
+    (anyBumped, txTable') =
+      IntMap.foldlWithKey' bumpOne (False, sharedTxTable st) downloadedTxs
+    bumpOne (!changed, !tbl) k _tx
+      | Just entry  <- IntMap.lookup k tbl
+      , Just entry' <- bumpCurrentMaxIfStuck entry
+      = (True, IntMap.insert k entry' tbl)
+      | otherwise = (changed, tbl)
```
crocodile-dentist, line 678:
```suggestion
    bumpOne (!changed, !tbl) k _tx =
        let (bumpedHere, tbl') = IntMap.alterF tryBump k tbl
        in (changed || bumpedHere, tbl')
      where
        tryBump Nothing  = (False, Nothing)
        tryBump (Just e) = case bumpCurrentMaxIfStuck policy now e of
                             Just e' -> (True,  Just e')
                             Nothing -> (False, Just e)
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1470 @@
...
+                  sharedTxTable =
+                    IntMap.insert k (releaseLease txEntry)
+                      (sharedTxTable sharedAcc)
+                }
+              , True
+              )
+            Nothing ->
+              (omittedCountAcc + 1, sharedAcc, wakeAcc)
+      | otherwise =
+          (omittedCountAcc + 1, sharedAcc, wakeAcc)
```
crocodile-dentist, line 1146:
```suggestion
          let (entryWasPresent, txTable') =
                IntMap.alterF releaseLeaseIfPresent k (sharedTxTable sharedAcc)
          in ( omittedCountAcc + 1
             , sharedAcc { sharedTxTable = txTable' }
             , wakeAcc || entryWasPresent
             )
      | otherwise =
          (omittedCountAcc + 1, sharedAcc, wakeAcc)

    releaseLeaseIfPresent Nothing  = (False, Nothing)
    releaseLeaseIfPresent (Just e) = (True,  Just (releaseLease e))
```

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
```diff
@@ -1,580 +1,1470 @@
...
+
+    updateRejected sharedAcc k =
+      case IntMap.lookup k (sharedTxTable sharedAcc) of
+        Just txEntry ->
+          sharedAcc {
+            sharedTxTable =
+              IntMap.insert k (markRejected txEntry) (sharedTxTable sharedAcc)
+          }
+        Nothing ->
+          sharedAcc
```
crocodile-dentist:
```suggestion
      sharedAcc {
        sharedTxTable = IntMap.adjust markRejected k (sharedTxTable sharedAcc)
      }
```

crocodile-dentist (commented):
A few more suggestions. Here we can cut down some work of lookup then insert with a single adjust.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
karknu: Done.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
karknu: It shouldn't matter. When we receive the first unrequested tx we will throw an exception.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
karknu: Your suggestion doesn't compile. Will keep my version for now.

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet/MiniProtocols.hs
karknu: done

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/Impaired.hs
karknu: ommitBodies doesn't use a wrapTxs like that since its purpose is to change the body list, not simply delay a `SendMsgReplyTxs`. See its `recvMsgRequestTxs` .

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/Impaired.hs
karknu: `n` is a Word16 so can't be negative. I will remove `max 1` in both places.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
crocodile-dentist: Then again, this will always rebuild the map so maybe it's not worth it, and the other ones in my other suggestions here as well.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/Impaired.hs
karknu: I've moved its generator and its shrinker.

cardano-diffusion/tests/lib/Test/Cardano/Network/Diffusion/Testnet.hs
karknu: The suggestion doesn't compile, but I've incorporated the shrinking and added tests to verify that the generator and shrinker produces valid `ScoreImpairmentInput`s.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
karknu: I moved the `shrinkImpairment` too Impaired.hs based on your suggestions so the patch doesn't apply.
I've simplified the shrinker based on your suggestions, added a `validImpairment` and updated AppV2's `validTxSubmissionImpairmentState` to use it.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
crocodile-dentist: I think this one still stands - adjust returns the original map if the key isn't present, otherwise it updates the value.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/State.hs
crocodile-dentist: This one isn't an improvement after all.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
karknu: `interntTxIds` is only used by testing today, but good to do if it changes in the future.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Registry.hs
karknu: Agreed.

ouroboros-network/lib/Ouroboros/Network/TxSubmission/Inbound/V2/Types.hs
karknu: The problem with this is that `Map.alterF` always allocates memory. The expected common case is that the txid is already interned by another of the 100+ peers.

There are benchmarks which covers this path and they all show an increase in allocation (between 8% - 40%) for the benchmarks with multiple peers.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
coot: Maybe add a comment.

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
coot: there's a simple syntax error, it should be `a@(_, _, _, ...`.  I'll create a PR to `io-sim`.

coot (approved):
LGTM

ouroboros-network/tests/lib/Test/Ouroboros/Network/TxSubmission/AppV2.hs
coot: https://github.com/IntersectMBO/io-sim/pull/261

crocodile-dentist (approved):
lgtm

coot (approved):
LGTM

PR #5381, opened 2026-06-11
Efficient common case for setCurrentTime

karknu:
# Description

KnownPeers.setCurrentTime and EstablishedPeers.setCurrentTime run on every governor loop iteration. Add a fast path to KnownPeers for the common case where nothing is scheduled or the earliest scheduled time is still in the future; previously it always ran the general path, rebuilding the record and (with +asserts) checking the invariant.

In EstablishedPeers use findMin instead of minView in the existing fast-path guard: an O(1) peek that avoids allocating the deletion of the minimum. The fast path now also covers the empty-queue case, which previously fell through to the general path.

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [x] Added labels.

### DIFF
--- a/ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
+++ b/ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
@@ -38,7 +38,6 @@ import Data.Map.Strict (Map)
 import Data.Map.Strict qualified as Map
 import Data.OrdPSQ (OrdPSQ)
 import Data.OrdPSQ qualified as PSQ
-import Data.Semigroup (Min (..))
 import Data.Set (Set)
 import Data.Set qualified as Set
 
@@ -220,13 +219,14 @@ setCurrentTime :: Ord peeraddr
 setCurrentTime now ep@EstablishedPeers { nextPeerShareTimes
                                        , nextActivateTimes
                                        }
- -- Efficient check for the common case of there being nothing to do:
-    | Just (Min t) <- (f <$> PSQ.minView nextPeerShareTimes)
-                   <> (f <$> PSQ.minView nextActivateTimes)
-    , t > now
+    -- Efficient check for the common case of there being nothing to do: either
+    -- nothing is scheduled, or the earliest scheduled time is still in the
+    -- future.
+    | noneDue nextPeerShareTimes
+    , noneDue nextActivateTimes
     = ep
   where
-    f (_,t,_,_) = Min t
+    noneDue q = maybe True (\(_,t,_) -> t > now) (PSQ.findMin q)
 
 setCurrentTime now ep@EstablishedPeers { nextPeerShareTimes
                                        , availableForPeerShare
--- a/ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
+++ b/ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
@@ -309,6 +309,19 @@ setCurrentTime :: Ord peeraddr
                -> KnownPeers peeraddr
                -> KnownPeers peeraddr
 setCurrentTime now knownPeers@KnownPeers {
+                     nextConnectTimes,
+                     clearFailCountTimes
+                   }
+    -- Efficient check for the common case of there being nothing to do: either
+    -- nothing is scheduled, or the earliest scheduled time is still in the
+    -- future.
+    | noneDue nextConnectTimes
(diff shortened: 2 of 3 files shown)


ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
```diff
@@ -220,13 +219,14 @@ setCurrentTime :: Ord peeraddr
...
-    , t > now
+    -- Efficient check for the common case of there being nothing to do: either
+    -- nothing is scheduled, or the earliest scheduled time is still in the
+    -- future.
+    | noneDue nextPeerShareTimes
+    , noneDue nextActivateTimes
     = ep
   where
-    f (_,t,_,_) = Min t
+    noneDue q = maybe True (\(_,t,_) -> t > now) (PSQ.findMin q)
```
coot, line 229: Previously when either `minView` returned `Nothing`, we didn't run this branch, now we do.  I think it should be:
```suggestion
    noneDue q = maybe False (\(_,t,_) -> t > now) (PSQ.findMin q)
```
then when either `nextPeerShareTimes` or `nextActiveTimes` is empty, we'll set it.

crocodile-dentist (approved):
lgtm

ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
```diff
@@ -308,6 +308,19 @@ setCurrentTime :: Ord peeraddr
...
+                     clearFailCountTimes
+                   }
+    -- Efficient check for the common case of there being nothing to do: either
+    -- nothing is scheduled, or the earliest scheduled time is still in the
+    -- future.
+    | noneDue nextConnectTimes
+    , noneDue clearFailCountTimes
+    = knownPeers
+  where
+    noneDue q = maybe True (\(_,t,_) -> t > now) (PSQ.findMin q)
```
coot, line 322: The same as above.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
karknu: We don't really set anything. We returned the established peers unchanged. And we should only do that if both nextPeerShareTimes and nextActiveTimes are empty or has min time in the future.

ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/KnownPeers.hs
karknu: Same reply as above ;)

ouroboros-network/lib/Ouroboros/Network/PeerSelection/State/EstablishedPeers.hs
coot: The original logic makes more sense to me, e.g. update the time if:
* both times are set and either of them is past `now`, or
* either of them is not set

versus the current:
* both times are set and both are past `now`, or
* both are not set

My proposition is also wrong :grin:, but at least it's matches the original in the second case.

coot: I think we want:
```
    |  noneDue nextPeerShareTimes
    || noneDue nextActivateTimes   -- use OR
    = ep
  where
    noneDue q = maybe False (\(_,t,_) -> t > now) (PSQ.findMin q)
```

karknu: I think a table will make this clear.

| state | needed | current (`maybe True`, `&&`) | suggested (`maybe False`, `\|\|`) |
|---|---|---|---|
| both empty | fast path | fast path | general path (harmless no-op) |
| one empty, other all-future | fast path | fast path | fast path |
| one all-future, others has due entry | general path | general path | fast path (bug) |
| both all-future | fast path | fast path | fast path |
| both have due entries | general path | general path | general path |

karknu: I did mention the changed behavior for established in the commit message: 
"The fast path now also covers the empty-queue case, which previously fell through to the general path."

coot: You're right, I got a wrong mental model. We want `&&` since it's guarding the fast path.

coot (approved):

PR #5402, opened 2026-07-14
Rearrange severity query for new tracing

f-f:
# Description

This is to satisfy the tests in `cardano-node` that check for the namespace severity being defined - e.g. see [this failing test run](https://github.com/IntersectMBO/cardano-node/actions/runs/28873986470/job/85643976289?pr=6604)

# Checklist

### Quality
* [x] Self-reviewed the PR.

### Maintenance
* [ ] Added labels.

### DIFF
--- a/ouroboros-network/changelog.d/20260715_120250_crocodile-dentist_fix_tracing_severity.md
+++ b/ouroboros-network/changelog.d/20260715_120250_crocodile-dentist_fix_tracing_severity.md
@@ -1,0 +1,22 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Breaking
+
+- A bullet item for the Breaking category.
+
+-->
+<!--
+### Non-Breaking
+
+- A bullet item for the Non-Breaking category.
+
+-->
+### Patch
+
+- Tracing instances cleanup
--- a/ouroboros-network/tracing/Ouroboros/Network/Tracing/PeerSelection.hs
+++ b/ouroboros-network/tracing/Ouroboros/Network/Tracing/PeerSelection.hs
@@ -76,7 +76,7 @@ instance (Show ntnAddr, Show ntcAddr) =>
     , "mode" .= String (pack (showString "0o" . showOct mode $ ""))
     ]
   forMachine _dtal (Diff.InsecureLocalSocketPermissions localAddress mode) = mconcat
-    [ "kind" .= String "InsecureLocalSocketDirectory"
+    [ "kind" .= String "InsecureLocalSocketPermissions"
     , "path" .= String (pack . show $ localAddress)
     , "mode" .= String (pack (showString "0o" . showOct mode $ ""))
     ]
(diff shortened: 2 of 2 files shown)


crocodile-dentist (approved):

coot:
Forcing pattern matching on list of strings rather than constructors doesn't look right (and could be error prone).  @mgmeier could the test be changed?

mgmeier:
The tracing system does not currently represent namespaces on the type level. There's an ongoing discussion on the pros and cons of doing so.

With that being said: The severity can be defined by matching on a namespace or by inspecting a run-time value. Obviously, the former approach has less performance overhead, as the tracing system can optimize better based on a specific config (i.e. filtering a trace before an unnecessary value is even constructed and then immediately fizzled out).

The tracing system does not prevent you from mixing both strategies, though. There are various examples in the Node that use a mixed approach.

So I want to highlight that the failing test that motivated this PR is deliberate design decision inside `cardano-node` - and its test suite. The tests guarantee that *at least* a namespace-based severity definition exists. It's not a design decision of the tracing system.

If I'm correct, there's another thing you'd sacrifice, which is exactly what the test failure points to: You can't use namespace-based severity filtering on that trace message anymore - it will instead use the filter value from its nearest ancestor that *does* have a value via the config - or am I wrong @jutaro ? So it's not only a tiny bit of performance you give up, but also granularity to configure your tracing output.

And now to answer the question - could the test be changed? Yes, of course - by removing `Startup.DiffusionInit: <...>` from the test data. This would imply "We do not guarantee that Startup.DiffusionInit provides configurable namespace-based filtering - instead, we'll throw a warning when you try that in your config."

jutaro:
Thanks for tracking this down! I fully agree this is the right fix. The tracing system should resolve severity from the namespace — full stop. Matching on the runtime value should only be a fallback for the genuinely rare cases where the same namespace needs different severities depending on the payload (e.g. a Decline event with different rejection reasons). I wasn't aware that code like severityFor _ (Just Diff.RunServer{}) = Just Info existed in DiffusionTracer — every one of those equations maps one-to-one to a distinct namespace anyway, so this was never actually exploiting the value-dependent case, it was just an accidental (and needlessly fragile) way of expressing a namespace-based mapping. Good catch, and thanks for the fix.

jutaro:
Hi Micha, you're right that this is cardano-node's own test enforcing a stricter guarantee than trace-dispatcher itself requires — that's a fair framing, not a misrepresentation.

But I'd push back on treating this as a tradeoff we're sacrificing something for. The reason cardano-node wants that guarantee everywhere is exactly the performance/configurability argument you made yourself: namespace-based severity is strictly more capable — it resolves even with no value in hand (docs generation, static filter elision), and it costs nothing extra in the common case. Value-based severity is only necessary when a single namespace legitimately carries different severities depending on payload — the Decline case elsewhere in the node is a real example of that. DiffusionTracer's old code wasn't that: every Just Diff.X{} pattern corresponded to exactly one namespace, so it wasn't "mixing strategies" on purpose, it was 100% value-based with no namespace path at all. There's no legitimate case being given up here, so I don't think weakening the test to carve out Startup.DiffusionInit is the right move — it would be codifying an oversight as a deliberate exception, not preserving a real design tradeoff.

On the specific technical claim: I went and checked getSeverity/getOption in trace-dispatcher's Configuration.hs — config does resolve by walking up the namespace path (inits/reverse/asum over prefixes) to the nearest ancestor with an explicit entry, but that happens unconditionally for every namespace, regardless of whether severityFor is namespace- or value-based. What value-based severityFor actually costs is different: since severityFor ns Nothing can't resolve without a constructed value, isSilentTracer can't statically prove a namespace falls below the configured threshold before the value even exists, so that early-elision optimization is lost for that whole trace family. That's a real cost, just not quite "you lose namespace-based filtering" — happy to be corrected if I'm missing the mechanism you had in mind.

